PR #6194, opened 2022-06-06
Use helm chart to generate manifests for testing

caseydavenport:
## Description

Fixes a problem where the tests we use in node CI and in-repo e2e
testing aren't the same as the manifests we ship.

The manifest we use in our tests is _slightly_ modified from the ones we ship, so this PR uses helm to manage those minor differences:

- Different image names (to use the locally built code)
- Image pull policy of "Never" to ensure we don't accidentally pull images.
- Don't install CRDs (this is done as part of cluster creation) 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


node/tests/k8st/infra/calico-kdd.yaml
```diff
@@ -160,14 +4102,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: calico-node
 rules:
-  # The tests require that node has access to secrets for BGP peering.
```
caseydavenport, line 163: Need to provide these permissions some other way - e.g. with a second ClusterRole / Binding.

mgleung (approved):

PR #6198, opened 2022-06-07
Add counters to bpf dataplane

mazdakn:
## Description

Add counters to bpf programs and update `calico-bpf` binary with `counters` sub command to support reading and flushing bpf counters.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
eBPF: Add counters to eBPF programs, and add support to calico-bpf to work with those counters.
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/tc.c
```diff
@@ -92,6 +93,12 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 	}
 	__builtin_memset(ctx.state, 0, sizeof(*ctx.state));
 
+	if (!ctx.counters) {
+		CALI_DEBUG("Counters map lookup failed: DROP\n");
+		return TC_ACT_SHOT;
+	}
+	inc_counter(ctx.counters, TOTAL_PKTS);
```
tomastigera, line 100:
```suggestion
	inc(ctx, TOTAL_PKTS)
```

some small macro like that? :thinking:

mazdakn: sure, just if you don't mind I'll uppercase like `INC` just to express that it is macro?

felix/bpf-gpl/tc.c
```diff
@@ -1341,25 +1390,34 @@ int calico_tc_skb_drop(struct __sk_buff *skb)
...
+	CALI_DEBUG("src=%x dst=%x\n", bpf_ntohl(ctx.state->ip_src),
+			bpf_ntohl(ctx.state->ip_dst));
+	CALI_DEBUG("pre_nat=%x:%d\n", bpf_ntohl(ctx.state->pre_nat_ip_dst),
+			ctx.state->pre_nat_dport);
+	CALI_DEBUG("post_nat=%x:%d\n", bpf_ntohl(ctx.state->post_nat_ip_dst), ctx.state->post_nat_dport);
+	CALI_DEBUG("tun_ip=%x\n", ctx.state->tun_ip);
+	CALI_DEBUG("pol_rc=%d\n", ctx.state->pol_rc);
+	CALI_DEBUG("sport=%d\n", ctx.state->sport);
+	CALI_DEBUG("flags=0x%x\n", ctx.state->flags);
+	CALI_DEBUG("ct_rc=%d\n", ctx.state->ct_result.rc);
```
tomastigera, line 1420: why all this change?

felix/bpf-gpl/ut/icmp_port_unreachable.c
```diff
@@ -9,7 +9,12 @@
 static CALI_BPF_INLINE int calico_unittest_entry (struct __sk_buff *skb)
 {
 	struct cali_tc_ctx ctx = {
+		.counters = counters_get(),
```
tomastigera, line 12: you could have just provided a dummy on stack datablock, you would not need to do the checks below. It never calls any bpf map helper after this, right?

felix/bpf/counters/counters.go
```diff
@@ -0,0 +1,146 @@
...
+
+const (
+	HookIngress = "ingress"
+	HookEgress  = "egress"
+)
+
+var Hooks = []string{HookIngress, HookEgress}
+
+type Counters struct {
+	maps     map[string]bpf.Map
```
tomastigera, line 52: wouldn't an array of 2 be enough? You can define constants for 0 and 1 for the code to be equally readable

felix/bpf/counters/map.go
```diff
@@ -0,0 +1,42 @@
...
+func MapForTest(mc *bpf.MapContext) bpf.Map {
+	return mc.NewPinnedMap(bpf.MapParameters{
+		Filename:   "/sys/fs/bpf/tc/globals/test_v1_counters",
+		Type:       "percpu_array",
+		KeySize:    4, // __u32
+		ValueSize:  uint32Size * MaxCounterNumber,
+		MaxEntries: 1,
+		Name:       bpf.CountersMapName(),
+	})
+}
```
tomastigera, line 42: why not in the test file? and why not to just use the Map() above in the test?

felix/bpf-gpl/conntrack.h
```diff
@@ -294,6 +294,7 @@ static CALI_BPF_INLINE bool skb_icmp_err_unpack(struct cali_tc_ctx *ctx, struct
 
 	if (skb_refresh_validate_ptrs(ctx, ICMP_SIZE + sizeof(struct iphdr) + 8)) {
 		ctx->fwd.reason = CALI_REASON_SHORT;
+		INC(ctx, ERR_SHORT_PKTS);
```
tomastigera, line 297: I wonder if we could merge these. We do not really use the reasons for anything except some debugging. And the reasons could be not funny numbers like now, but something that would work for error counters too.  So that we do  not need to remember to set both.

Maybe `DENY_REASON(ctx, xyz)`

felix/bpf-gpl/tc.c
mazdakn: `INC` macro need a `ctx` variable, and to add counter in this part, needed to declare one. Since `ctx` also contain `state` I thought we can use it instead of declaring another one.

felix/bpf/counters/counters.go
mazdakn: We will need another one for `XDP` later. To answer you question, I thought it would be nice to use the key value i.e. `ingress`, and `egress` in the log output. Does it make sense?

felix/bpf/counters/map.go
mazdakn: Right. fixed it.

felix/bpf-gpl/conntrack.h
mazdakn: I was thinking do to so in next step. It seem the pattern for setting reason and counter is repeated many places. So definitely we can do so.

felix/bpf-gpl/ut/icmp_port_unreachable.c
mazdakn: It ends up calling `skb_refresh_validate_ptrs()` which calls `bpf_skb_pull_data()`. If we don't init `.counters` the object file will be rejected by the verifier.

felix/bpf/tc/compiler.go
```diff
@@ -121,3 +124,22 @@ func ProgFilename(epType EndpointType, toOrFrom ToOrFromEp, epToHostDrop, fib, d
...
+const (
+	PinBaseDir = "/sys/fs/bpf/tc/"
+)
+
+func MapPinPath(typ int, name, iface string, hook Hook) string {
+	subDir := "globals"
+	if (typ == unix.BPF_MAP_TYPE_PROG_ARRAY && strings.Contains(name, bpf.JumpMapName())) ||
+		(typ == unix.BPF_MAP_TYPE_PERCPU_ARRAY && strings.Contains(name, bpf.CountersMapName())) {
+		// Remove period in the interface name if any
+		ifName := strings.ReplaceAll(iface, ".", "")
```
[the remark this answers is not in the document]
fasaxc: Hmm, this code was just moved, maybe see if you can dig out the reason and comment it!

fasaxc (changes requested):
Generally looking good, few comments inline.

felix/bpf-gpl/reasons.h
```diff
@@ -1,31 +1,34 @@
...
+	CALI_REASON_UNKNOWN, // This value also used for counting the total number of packets
+	CALI_REASON_ACCEPTED_BY_FAILSAFE,
+	CALI_REASON_ACCEPTED_BY_POLICY,
+	CALI_REASON_DROPPED_BY_POLICY,
+	CALI_REASON_SHORT,
+	CALI_REASON_CSUM_FAIL,
+	CALI_REASON_IP_OPTIONS,
+	CALI_REASON_IP_MALFORMED,
+
+	// The following values are only used for setting reasons, and are not used by
```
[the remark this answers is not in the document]
mazdakn: I'll add more of these to the counters map. 

only `CALI_REASON_ACCEPTED_BY_XDP` is used as an optimisation before `ctx` is initiated in tc programs. I can re-arrange to initialize `ctx` before `CALI_REASON_ACCEPTED_BY_XDP` but I am not sure how it affects performance of optimisations.

felix/bpf/counters/counters.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func NewCounters(iface string) *Counters {
+	cntr := Counters{
+		iface: iface,
+	}
+	var err error
+	cntr.numOfCpu, err = libbpf.NumPossibleCPUs()
+	if err != nil {
+		logrus.WithError(err).Error("failed to get libbpf number of possible cpu. Will use runtime information")
```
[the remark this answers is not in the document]
mazdakn: Fixed.

felix/cmd/calico-bpf/commands/counters.go
```diff
@@ -0,0 +1,152 @@
...
+	cmd.Printf("Accepted by policy: \t\t\t%d\t\t%d\n",
+		values["ingress"][counters.AcceptedByPolicy], values["egress"][counters.AcceptedByPolicy])
+
+	cmd.Printf("Dropped by policy: \t\t\t%d\t\t%d\n",
+		values["ingress"][counters.DroppedByPolicy], values["egress"][counters.DroppedByPolicy])
+	cmd.Printf("Dropped short packets: \t\t\t%d\t\t%d\n",
+		values["ingress"][counters.DroppedShortPacket], values["egress"][counters.DroppedShortPacket])
+	cmd.Printf("Dropped incorrect checksum: \t\t%d\t\t%d\n",
+		values["ingress"][counters.DroppedFailedCSUM], values["egress"][counters.DroppedFailedCSUM])
+	cmd.Printf("Dropped Packets with IP options: \t%d\t\t%d\n",
```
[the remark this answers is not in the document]
mazdakn: Fixed.

felix/cmd/calico-bpf/commands/counters.go
```diff
@@ -0,0 +1,152 @@
...
+		values["ingress"][counters.TotalPackets], values["egress"][counters.TotalPackets])
+
+	cmd.Printf("Accepted by failsafe: \t\t\t%d\t\t%d\n",
+		values["ingress"][counters.AcceptedByFailsafe], values["egress"][counters.AcceptedByPolicy])
+	cmd.Printf("Accepted by policy: \t\t\t%d\t\t%d\n",
+		values["ingress"][counters.AcceptedByPolicy], values["egress"][counters.AcceptedByPolicy])
+
+	cmd.Printf("Dropped by policy: \t\t\t%d\t\t%d\n",
+		values["ingress"][counters.DroppedByPolicy], values["egress"][counters.DroppedByPolicy])
+	cmd.Printf("Dropped short packets: \t\t\t%d\t\t%d\n",
```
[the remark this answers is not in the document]
mazdakn: Fixed.

felix/bpf/maps.go
```diff
@@ -287,24 +291,41 @@ func (b *PinnedMap) Iter(f IterCallback) error {
 func (b *PinnedMap) Update(k, v []byte) error {
 	if b.perCPU {
 		// Per-CPU maps need a buffer of value-size * num-CPUs.
-		logrus.Panic("Per-CPU operations not implemented")
+		numCPU, err := libbpf.NumPossibleCPUs()
```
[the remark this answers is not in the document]
mazdakn: Done. Thanks for the suggestion.

felix/bpf/counters/counters.go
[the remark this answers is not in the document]
mazdakn: Fixed.

felix/bpf/tc/compiler.go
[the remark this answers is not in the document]
mazdakn: This is added in this commit: https://github.com/projectcalico/calico/commit/6369bf2599e0d5bfa61c97b4936db5efa0d61190

@sridhartigera do you remember why `.` is removed from interface name?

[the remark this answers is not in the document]
sridhartigera: Since we use the ifname in the pin path of the jump map, having a period is not allowed. Hence we remove the period. For this case, ```foobar``` and ```foo.bar```, we need to think of some other scheme.

felix/cmd/calico-bpf/commands/counters.go
```diff
@@ -0,0 +1,152 @@
...
+			return fmt.Errorf("Failed to read bpf counters. hook=%s err=%v", hook, err)
+		}
+		if len(val) < counters.MaxCounterNumber {
+			return fmt.Errorf("Failed to read enough data from bpf counters. hook=%s", hook)
+		}
+
+		values[hook] = val
+	}
+
+	cmd.Printf("\t\t\t\t\tingress\t\tegress\n")
```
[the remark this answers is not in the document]
mazdakn: Thanks for suggestion. Now I used `tablewriter` package like calicoctl to print output.

felix/bpf-gpl/reasons.h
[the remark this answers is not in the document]
mazdakn: All counters are added now, except `CALI_REASON_ACCEPTED_BY_XDP` due to the reason stated above.
Do you think initiating `ctx` worth counting the number of packets accepted by XDP?

felix/bpf/counters/counters.go
mazdakn: Fixed.

felix/cmd/calico-bpf/commands/counters.go
```diff
@@ -0,0 +1,151 @@
...
+	Short: "Show and reset counters",
+}
+
+func init() {
+	countersCmd.AddCommand(countersDumpCmd)
+	countersCmd.AddCommand(countersFlushCmd)
+	rootCmd.AddCommand(countersCmd)
+
+	countersDumpCmd.Flags().String("iface", "", "Interface name")
+	countersDumpCmd.Flags().Bool("all", false, "All interfaces")
```
fasaxc, line 42: I think you should drop the `--all` parameter, it just makes two ways to do the same thing, which is usually a bad idea.

felix/cmd/calico-bpf/commands/counters.go
```diff
@@ -0,0 +1,151 @@
...
+			caption,
+			fmt.Sprintf("%v", ingress),
+			fmt.Sprintf("%v", egress),
+		}
+	}
+
+	var rows [][]string
+
+	noOfCounters := len(counters.Descriptions)
+	for c := 0; c < noOfCounters; c++ {
```
fasaxc, line 132: I recommend sorting the values by description somehow.  Then, when we add a new counter it won't have to be stuck at the end.

fasaxc (changes requested):

felix/cmd/calico-bpf/commands/counters.go
fasaxc: (If you're not sure how to do that, I can give you some pointers)

felix/bpf-gpl/reasons.h
```diff
@@ -1,31 +1,31 @@
...
-	CALI_REASON_CSUM_FAIL= 0xcf,
-	CALI_REASON_ENCAP_FAIL = 0xef,
-	CALI_REASON_DECAP_FAIL = 0xdf,
-	CALI_REASON_ICMP_DF = 0x1c,
-	CALI_REASON_IP_OPTIONS = 0xeb,
-	CALI_REASON_IP_MALFORMED = 0xec,
-	CALI_REASON_UNAUTH_SOURCE = 0xed,
-	CALI_REASON_RT_UNKNOWN = 0xdead,
-	CALI_REASON_ACCEPTED_BY_XDP = 0xd9,
+	CALI_REASON_UNKNOWN, // This value also used for counting the total number of packets
```
tomastigera, line 11:
```suggestion
	CALI_REASON_UNKNOWN,
	COUNTER_TOTAL_PACKETS = CALI_REASON_UNKNOWN,
```

in the spirit of 

```
#include <stdio.h>
#include <stdlib.h>

enum someenum {
    UNKNOWN,
    UNKNOWN_ALIAS = UNKNOWN,
    A,
    B,
    C = B,
    D
};


int main() {
    printf("%d\n", UNKNOWN);
    printf("%d\n", UNKNOWN_ALIAS);
    printf("%d\n", A);
    printf("%d\n", B);
    printf("%d\n", C);
    printf("%d\n", D);
}

0
0
1
2
2
3

```

felix/bpf-gpl/tc.c
```diff
@@ -92,6 +93,14 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 	}
 	__builtin_memset(ctx.state, 0, sizeof(*ctx.state));
 
+	if (!ctx.counters) {
+		CALI_DEBUG("Counters map lookup failed: DROP\n");
+		// We don't want to drop packets just because counters initialization fails, but
+		// failing here normally should not happen.
+		return TC_ACT_SHOT;
+	}
+	COUNTER_INC(&ctx, CALI_REASON_UNKNOWN); // This is used to keep the total number of packets
```
tomastigera, line 102:
```suggestion
	COUNTER_INC(&ctx, COUNTER_TOTAL_PACKETS);
```

Let the code to speak for itself ;-)

felix/bpf-gpl/tc.c
```diff
@@ -108,14 +117,16 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 		case CALI_SKB_MARK_BYPASS_FWD:
 			CALI_DEBUG("Packet approved for forward.\n");
 			ctx.fwd.reason = CALI_REASON_BYPASS;
```
tomastigera, line 119: If we do not deny such packets, why do we set the reason? :thinking: Perhaps we should not set it?

tomastigera (commented):
LGTM How does an example of the output for say `all` looks like?

felix/cmd/calico-bpf/commands/counters.go
mazdakn: Fixed.

felix/bpf-gpl/tc.c
mazdakn: Fixed. Thanks

felix/bpf-gpl/reasons.h
mazdakn: Fixed.

felix/bpf-gpl/tc.c
mazdakn: Fixed.

felix/cmd/calico-bpf/commands/counters.go
mazdakn: Fixed.

felix/cmd/calico-bpf/commands/counters.go
mazdakn: Done. Is it OK now?

fasaxc (commented):
Looking good but please an you add some details to the docs, perhaps in the eBPF troubleshooting page?

tomastigera (approved):
LGTM thanks!

Issue #5712, opened 2022-03-05
After K8S cluster run stable for weeks, K8S cannot start POD (randomly) because "error getting ClusterInformation: connection is unauthorized: Unauthorized"


earthlovepython:
K8S & Calico information
====================
HostOS: RHEL 8.2 
K8S:  on-premise cluster; version is v1.21.1;  "IPVS" mode;  IP4/IP6 dual stack; installed using kubespray
Calico: version is v3.18.4; non-BGP mode;  enabled "IP6" DNAT. 
Our docker image is built on top of "RHEL ubi:8"
We do not setup external ETCD cluster.

"kubectl describe" output
====================
[support@node-cont-1-qa conf]$ kubectl describe pod export-job-job-dp8hb
Name:           export-job-job-dp8hb
Namespace:      pio
Priority:       0
Node:           node-df1-1/10.0.156.180
Start Time:     Wed, 23 Feb 2022 05:57:18 -0800
Labels:         app.kubernetes.io/instance=export-job-job
                controller-uid=5d9f3e4b-e74c-4280-a3be-e31d37e92b84
                job-name=export-job-job
Annotations:    cni.projectcalico.org/podIP:
                cni.projectcalico.org/podIPs:
Status:         Pending
IP:
IPs:            <none>
Controlled By:  Job/export-job-job
Containers:
  export-job-job:
    Container ID:
    Image:         10.0.156.250:5000/img-admf:9.3.0.0B038
    Image ID:
    Port:          <none>
    Host Port:     <none>
    Command:
      csh
    Args:
      -c
      source /TT9/configXcp.sh; lis_conf; python2 /etc/pio/APPL/XcdbBackup.py --exportdb --dir /var/tmp; sleep 300
    State:          Waiting
      Reason:       ContainerCreating
    Ready:          False
    Restart Count:  0
    Limits:
      cpu:     500m
      memory:  512Mi
    Requests:
      cpu:        200m
      memory:     256Mi
    Environment:  <none>
    Mounts:
      /TT9/PIO/9.0.0/RUN/config/APPL/DBConMgr.cnfg from db-conf (rw,path="DBConMgr.cnfg")
      /TT9/PIO/9.0.0/RUN/config/feature_conf.json from feature-conf (rw,path="feature_conf.json")
      /TT9/PIO/9.0.0/RUN/license/license.json from license-conf (rw,path="license.json")
[...]

PR #5910, opened 2022-04-14
Use TokenRequest API instead of calico-nodes service account token for CNI kubeconfig.

ScheererJ:
## Description

Use TokenRequest API instead of calico-nodes service account token for CNI kubeconfig.

With projected service account tokens in kubernetes, the service account tokens of pods
are bound to the lifetime of the corresponding pod. Therefore, it may lead to problems
if an external process re-uses the token of a pod.
The CNI binaries used the token of calico-node. However, in case calico-node got stopped
the corresponding token lost its validity and hence could no longer be used for CNI
operations. Usually, this automatically resolves over time, but there are some edge cases
where this is not possible, e.g. if calico-node is autoscaled in terms of resources and
the new resource requests would require preemption/eviction of an existing pod the CNI
operation to delete the network sandbox will fail due to the no longer valid token
(as calico-node was stopped beforehand).
This change switches over to using the TokenRequest API instead, i.e. creating new
tokens with limited validity. It would have been good to bind the token to an object,
e.g. to the corresponding node, but as of now only secret and pod are supported types
for binding tokens. Hence, the tokens are only limited in time and not bound to any
other kubernetes object.

## Related issues/PRs

fixes #4857 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

Issue #5921, opened 2022-04-15
Separate permissions for calico/node and calico/cni


caseydavenport:
## Expected Behavior

As discussed here: https://github.com/projectcalico/calico/pull/5910#discussion_r851364934

We can introduce a service account specifically for our CNI plugin - this would be a nice improvement rather than sharing serviceaccounts between calico/node and the CNI plugin, since they required similar but ultimately different permissions. 

## Current Behavior

serviceaccount and RBAC resources shared beteween calico/node and CNI plugin. 

## Possible Solution

- Add new calico-cni serviceaccount, clusterrole, and binding.
- Split out permissions and tidy up calico-node RBAC resources.
- Make sure upgrade works alright (may need a grace period where both get the superset of permissions)

PR #6218, opened 2022-06-13
[release-v3.23] Auto pick #5910: Use TokenRequest API instead of calico-nodes service account

### LINKED ISSUE #4857: network: error  getting ClusterInformation: connection is unauthorized
## Expected Behavior

Calico CNI plugin tears down Pod in a timely manner.

## Current Behavior

Calico CNI plugin shows errors terminating Pods, and therefore eviction takes too long. Especially relevant in Kubernetes conformance testing.

```
Aug 18 18:19:04.521: INFO: At 2021-08-18 18:18:01 +0000 UTC - event for taint-eviction-a1: {kubelet ip-10-0-8-52} FailedKillPod: error killing pod: failed to "KillPodSandbox" for "0701ef9b-e
17d-43b5-a48f-89fa3ac00999" with KillPodSandboxError: "rpc error: code = Unknown desc = networkPlugin cni failed to teardown pod \"taint-eviction-a1_taint-multiple-pods-4011\" network: error
 getting ClusterInformation: connection is unauthorized: Unauthorized"
```

The natural things to check are RBAC permissions, which match recommendations:

```
- apiGroups:
  - crd.projectcalico.org
  resources:
  - globalfelixconfigs
  - felixconfigurations
  - bgppeers
  - globalbgpconfigs
  - bgpconfigurations
  - ippools
  - ipamblocks
  - globalnetworkpolicies
  - globalnetworksets
  - networkpolicies
  - networksets
  - clusterinformations
  - hostendpoints
  - blockaffinities
  verbs:
  - get
  - list
  - watch
...
```

To be certain, we can use the actual kubeconfig Calico writes to the host's `/etc/cni/net.d`. It does indeed seem to have permission to get clusterinformations. The error above is unusual.

```
./kubectl --kubeconfig /etc/cni/net.d/calico-kubeconfig auth can-i get clusterinformations --all-namespaces
yes
```

```
sonobuoy run --e2e-focus="NoExecuteTaintManager Multiple Pods" --e2e-skip="" \
--plugin-env=e2e.E2E_EXTRA_ARGS="--non-blocking-taints=node-role.kubernetes.io/controller"
```

## Context

This issue affects Kubernetes Conformance tests:

```
Summarizing 1 Failure:

[Fail] [sig-node] NoExecuteTaintManager Multiple Pods [Serial] [It] evicts pods with minTolerationSeconds [Disruptive] [Conformance] 
/workspace/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/github.com/onsi/ginkgo/internal/leafnodes/runner.go:113
```

The test in question creates two Pods that don't tolerate a taint, and expects them to be terminated within certain times. In Kubelet logs, the Calico CNI plugin is complaining with the logs above and termination takes too long.

## Your Environment

* Calico version: v3.19.2 or v3.20.0
* Orchestrator version (e.g. kubernetes, mesos, rkt): Kubernetes v1.22.0
* Operating System and version: Fedora CoreOS
* Link to your project (optional): https://github.com/poseidon/typhoon, Calico manifests https://github.com/poseidon/terraform-render-bootstrap/tree/master/resources/calico

caseydavenport:
Cherry pick of #5910 on release-v3.23.

#5910: Use TokenRequest API instead of calico-nodes service account

# Original PR Body below

## Description

Use TokenRequest API instead of calico-nodes service account token for CNI kubeconfig.

With projected service account tokens in kubernetes, the service account tokens of pods
are bound to the lifetime of the corresponding pod. Therefore, it may lead to problems
if an external process re-uses the token of a pod.
The CNI binaries used the token of calico-node. However, in case calico-node got stopped
the corresponding token lost its validity and hence could no longer be used for CNI
operations. Usually, this automatically resolves over time, but there are some edge cases
where this is not possible, e.g. if calico-node is autoscaled in terms of resources and
the new resource requests would require preemption/eviction of an existing pod the CNI
operation to delete the network sandbox will fail due to the no longer valid token
(as calico-node was stopped beforehand).
This change switches over to using the TokenRequest API instead, i.e. creating new
tokens with limited validity. It would have been good to bind the token to an object,
e.g. to the corresponding node, but as of now only secret and pod are supported types
for binding tokens. Hence, the tokens are only limited in time and not bound to any
other kubernetes object.

## Related issues/PRs

fixes #4857 
fixes https://github.com/projectcalico/calico/issues/5712

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico now uses the TokenRequest API to generate and refresh a token for the CNI plugin.  This ensures that the token remains valid even when the calico-node daemonset is restarted. (@[ScheererJ](https://github.com/ScheererJ))
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/calico/_includes/charts/calico/templates/calico-node-rbac.yaml
+++ b/calico/_includes/charts/calico/templates/calico-node-rbac.yaml
@@ -70,6 +70,14 @@ rules:
       - pods/status
     verbs:
       - patch
+  # Used for creating service account tokens to be used by the CNI plugin
+  - apiGroups: [""]
+    resources:
+      - serviceaccounts/token
+    resourceNames:
+      - calico-node
+    verbs:
+      - create
   # Calico monitors various CRDs for config.
   - apiGroups: ["crd.projectcalico.org"]
     resources:
--- a/app-policy/config/install/05-calico.yaml
+++ b/app-policy/config/install/05-calico.yaml
@@ -96,6 +96,13 @@ rules:
       - get
       - list
       - watch
+  - apiGroups: [""]
+    resources:
+      - serviceaccounts/token
+    resourceNames:
+      - calico-node
+    verbs:
+      - create
 ---
 
 apiVersion: rbac.authorization.k8s.io/v1beta1
--- a/calico/getting-started/kubernetes/hardway/install-node.md
+++ b/calico/getting-started/kubernetes/hardway/install-node.md
@@ -121,6 +121,14 @@ rules:
       - pods/status
     verbs:
       - patch
+  # Used for creating service account tokens to be used by the CNI plugin
+  - apiGroups: [""]
+    resources:
+      - serviceaccounts/token
+    resourceNames:
+      - calico-node
+    verbs:
+      - create
   # Calico monitors various CRDs for config.
   - apiGroups: ["crd.projectcalico.org"]
     resources:
--- a/cni-plugin/Makefile
+++ b/cni-plugin/Makefile
@@ -207,8 +207,8 @@ pkg/install/install.test: pkg/install/*.go
 
 .PHONY: test-install-cni
 ## Test the install
-test-install-cni: image pkg/install/install.test
-	cd pkg/install && CONTAINER_NAME=$(CNI_PLUGIN_IMAGE):latest-$(ARCH) ./install.test
+test-install-cni: run-k8s-apiserver image pkg/install/install.test
+	cd pkg/install && CONTAINER_NAME=$(CNI_PLUGIN_IMAGE):latest-$(ARCH) CERTS_PATH=$(CERTS_PATH) ./install.test
 
 ###############################################################################
 # CI/CD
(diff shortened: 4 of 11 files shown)


calico/_includes/charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -70,6 +70,14 @@ rules:
       - pods/status
     verbs:
       - patch
+  # Used for creating service account tokens to be used by the CNI plugin
+  - apiGroups: [""]
+    resources:
+      - serviceaccounts/token
+    resourceNames:
+      - calico-node
```
mtaufen, line 78: If the role containing this rule is also bound to the `calico-node` service account, it allows `calico-node` to mint tokens for `calico-node`. This creates a potential security issue where an attacker could steal a token and prevent their access from ever expiring by sending periodic TokenRequests for the same service account whose token they stole. Instead, it would be better to grant `calico-node` permission to request tokens for a separate service account that the CNI is intended to run as. For example, grant `calico-node` permission to mint tokens for a `calico-node-cni` service account, and grant `calico-node-cni` the permissions that the CNI needs. That way, no tokens have the ability to self-perpetuate.

caseydavenport: Yep, I agree. I hadn't spotted this particular issue but already have a ticket tracking this enhancement for other reasons as well: https://github.com/projectcalico/calico/issues/5921

PR #6241, opened 2022-06-16
bpf fix attach gap

tomastigera:
## Description

felix/bpf: bpf_ep_mgr.ifStateMap is a cachingmap

This reconciles state after restarts and makes sure that unused entries
do not stay for ever.

felix/bpf: ifstate map contains iface name

For easier debugging and more robustnes in detecting what has changed
and what was the previous state.

felix/bpf: drop packets for disapproved FIB success

felix/bpf: bpfEndpointManager marks weps as ready.

When bpfEndpointManager has ifindex and applies tc programs to a WEP, it
also records in a bpf map that the iface is "ready". Thereafter tc
programs that have successful FIB lookup can redirec straight to the
wep. Otherwise the packet is dropped as it is unsafe to forward it and
it it pointless to let it be dropped by iptables for the same reason.

felix/bpf: approve FIB success

When we have a FIB success, we need to approve it otherwise we could
redirect packets to a local workload that does not have it's tc programs
and a policy applied yet. That leaves security hole.

felix/bpf: remove comment that is no longer true

felix/bpf: CT_RES_CONFIRMED set when both sides approved

CALI_CT_ESTABLISHED_BYPASS still required because that is a special case
for connections that have no NAT.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: we drop packets that are about to be redirected to a workload endpoint that does not have a tc attached program yet, hence is unprotected.
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -274,6 +281,10 @@ func newBPFEndpointManager(
 		rpfStrictModeEnabled: config.BPFEnforceRPF,
 	}
 
+	if err := m.ifStateMap.LoadCacheFromDataplane(); err != nil {
```
fasaxc, line 284: No need to load the cache here, it'll happen on the first Apply call

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -767,6 +801,20 @@ func (m *bpfEndpointManager) updateWEPsInDataplane() {
...
+			if err != nil {
+				if isLinkNotFoundError(err) {
+					// Interface is gone, nothing to do.
+					log.WithField("ifaceName", ifaceName).Debug(
+						"Ignoring request to program interface that is not present.")
+					err = nil
+				}
+			} else {
+				m.ifacesLock.Lock()
+				if iface, ok := m.nameToIface[ifaceName]; !ok || iface.info.ifindex == 0 {
```
fasaxc, line 813: I think this case will match with `applyPolicy`'s existing check on whether the interface is up according to `iface.info.ifaceIsUp` which is true exactly when the index is set.

Previously we handled that quietly because it's 100% expected that felix will see WEPs for pods that are being deleted in the dataplane (so their interfaces are missing) so need to be careful not to log lots of new errors here.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -823,6 +877,25 @@ func (m *bpfEndpointManager) updateWEPsInDataplane() {
 	})
 }
 
+func (m *bpfEndpointManager) wepIfaceReady(name string, ready bool) {
```
fasaxc, line 880: Feels like this method should key off `iface.dpState.isReady` for symmetry with the case in `onInterfaceUpdate`: ideally, whenever we change either flag we should rerun some common code to update the map.  However, `iface.dpState.isReady` never seem to get set back to false, even after an error?  Maybe that's wrong!

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -823,6 +877,25 @@ func (m *bpfEndpointManager) updateWEPsInDataplane() {
...
+		return
+	}
+
+	flgReady := ifstate.FlgReady
+	if !ready {
+		flgReady = 0
+	}
+
+	m.ifStateMap.SetDesired(
+		ifstate.NewKey(uint32(iface.info.ifindex)).AsBytes(),
```
fasaxc, line 892: Please can you add a utility method to actually do the map write.  Best not to have two places that build up the keys/flags in case we add a new flag later.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -897,6 +964,14 @@ func (m *bpfEndpointManager) applyPolicy(ifaceName string) error {
 
 	applyTime := time.Since(startTime)
 	log.WithField("timeTaken", applyTime).Info("Finished applying BPF programs for workload")
+
+	m.ifacesLock.Lock()
+	m.withIface(ifaceName, func(iface *bpfInterface) (forceDirty bool) {
+		iface.dpState.isReady = true
```
fasaxc, line 970: Think this needs to be set back to false somewhere.  It's also critical that `iface` is completely zeroed out once the wep and the interface are both gone.  Otherwise we leak.  The `withIface` method checks if the `iface` is the zero struct and cleans it up if so.

fasaxc (changes requested):

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -826,13 +877,24 @@ func (m *bpfEndpointManager) updateWEPsInDataplane() {
 // applyPolicy actually applies the policy to the given workload.
 func (m *bpfEndpointManager) applyPolicy(ifaceName string) error {
 	startTime := time.Now()
+	isReady := false
+
+	// Update the readiness state when exiting
+	defer func() {
+		m.ifacesLock.Lock()
```
fasaxc, line 884: Just spotted an issue here; if the code inside the withIface block below ever panics then this Lock() will deadlock and we'll never get the stack trace.  Suggest moving the block below into a function so the Unlock can be deferred.  Will probably never happen but trying to debug it would be tough if it did!

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -825,16 +871,16 @@ func (m *bpfEndpointManager) updateWEPsInDataplane() {
 	})
 }
 
-// applyPolicy actually applies the policy to the given workload.
-func (m *bpfEndpointManager) applyPolicy(ifaceName string) error {
+func (m *bpfEndpointManager) doApplyPolicy(ifaceName string, isReady *bool) error {
```
fasaxc, line 874: Could you just return `(bool, error)` instead of an out param?

fasaxc (approved):
Looks good; just two tiny questions.  Don't think I need to see it again though.

felix/daemon/daemon.go
```diff
@@ -987,7 +986,7 @@ func (fc *DataplaneConnector) readMessagesFromDataplane() {
 		log.WithField("payload", payload).Debug("New message from dataplane")
 		switch msg := payload.(type) {
 		case *proto.ProcessStatusUpdate:
-			fc.handleProcessStatusUpdate(ctx, msg)
+			fc.handleProcessStatusUpdate(context.TODO(), msg)
```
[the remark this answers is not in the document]
tomastigera: linter

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: I felt like this made simpler code without that much restructuring.

tomastigera:
Merging with just a few  known flakes in unrelated parts.

Issue #5196, opened 2020-04-06
Allow configuration of MAC address per-pod


lelvisl:
I have some legacy, that wants a permanent Mac address. How I can set MAC address for specific container/pod?

caseydavenport:
I think we could resolve this by introducing a new annotation that can be set for a pod.

```
cni.projectcalico.org/hwAdddress
```

It might get complicated when thinking about multi-network support, but we have other annotations to consider there as well so we'll need to figure that out anyway. I assume you want this MAC to be the veth within the container, yeah?

lelvisl:
Yes)

KlavsKlavsen:
We would love this as well - as we too have dockerized license crap software that checks interface mac before it wants to run (which works fine in docker - but we need to be able to move to kubernetes :)

PR #6249, opened 2022-06-20
Introduce new Pod annotation for assigning specified MAC address to container veth

Josh-Tigera:
## Description

Introduce new Pod annotation "cni.projectcalico.org/hwAddr" that will be used to assign a specified MAC address to the
container-side Veth interface

Added new unit tests that pass, as well as manually verified by creating a container within a cluster and examining the MAC address of eth0 from within the container to make sure it matched the address specified by annotation.

## Related issues/PRs
fixes [5196](https://github.com/projectcalico/calico/issues/5196)

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Add new Pod annotation for assigning specified MAC address to container veth
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
+++ b/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
@@ -152,20 +152,22 @@ func (d *linuxDataplane) DoNetworking(
 
 		// Check if there is an annotation requesting a specific fixed MAC address for the container Veth, otherwise
 		// use kernel-assigned MAC.
-		requestedContVethMac, found := annotations["cni.projectcalico.org/hwAddr"]
-		if found {
+		if requestedContVethMac, found := annotations["cni.projectcalico.org/hwAddr"]; found {
 			tmpContVethMAC, err := net.ParseMAC(requestedContVethMac)
-			if err == nil {
-				err = netlink.LinkSetHardwareAddr(contVeth, tmpContVethMAC)
-			}
 			if err != nil {
-				return fmt.Errorf("failed to set container veth MAC to %v as requested via annotation: %v",
+				return fmt.Errorf("failed to parse MAC address %v provided via cni.projectcalico.org/hwAddr: %v",
 					requestedContVethMac, err)
-			} else {
-				contVethMAC = tmpContVethMAC.String()
-				d.logger.Infof("successfully configured container veth MAC to %v as requested via annotation",
-					contVethMAC)
 			}
+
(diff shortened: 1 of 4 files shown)


cni-plugin/pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -150,8 +150,26 @@ func (d *linuxDataplane) DoNetworking(
...
 		}
 
-		// Fetch the MAC from the container Veth. This is needed by Calico.
-		contVethMAC = contVeth.Attrs().HardwareAddr.String()
+		// Check if there is an annotation requesting a specific fixed MAC address for the container Veth, otherwise
+		// use kernel-assigned MAC.
+		requestedContVethMac, found := annotations["cni.projectcalico.org/hwAddr"]
+		if found {
+			tmpContVethMAC, err := net.ParseMAC(requestedContVethMac)
+			if err == nil {
```
caseydavenport, line 158: This isn't necessarily wrong, but it's not a typical error handling pattern in go. Generally better to handle errors immediately rather than try to condense multiple error generating functions into a single error check. I think for most cases it's more idiomatic to do an `err != nil` check than an `err == nil` check as well, although there are obviously exceptions to that. 

e.g.,

```go
mac, err := parseMAC(..)
if err != nil {
  return err
}

err = netlink.LinkSetHardwareAddr(...)
if err != nil {
  return err
}
```

^ this also lets us tailor specific error messages to the user. e.g., if the annotation is bad, we want to say something specifically like "The MAC address provided via cni.projectcalico.org is invalid: %s"

cni-plugin/pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -150,8 +150,26 @@ func (d *linuxDataplane) DoNetworking(
...
+				err = netlink.LinkSetHardwareAddr(contVeth, tmpContVethMAC)
+			}
+			if err != nil {
+				return fmt.Errorf("failed to set container veth MAC to %v as requested via annotation: %v",
+					requestedContVethMac, err)
+			} else {
```
caseydavenport, line 164: Don't need an else block here since the code returns in the "if" statement.

cni-plugin/pkg/dataplane/linux/dataplane_linux.go
Josh-Tigera: That makes sense specifically about having unique messages for each failure case.

caseydavenport (approved):

PR #6250, opened 2022-06-20
Add input filter rule for wireguard

muff1nman:
## Description

Without this rule, in an environment with a default drop, Wireguard
traffic will not be accepted. This change adds a static input rule in the
same fashion as done for other encapsulation modes (ipip, vxlan) to
ensure these packets are accepted. No drop rule is needed however as
wireguard does its own validation of incoming messages.

## Todos

- [X] Tests
- [ ] Documentation
- [X] Release note

## Release Note

```release-note
Calico will now add an ACCEPT rule for the Wireguard UDP port when enabled to ensure policy doesn't drop Calico Wireguard traffic.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/driver.go
+++ b/felix/dataplane/driver.go
@@ -173,55 +173,6 @@ func StartDataplaneDriver(configParams *config.Config,
 			log.WithError(err).Warning("Unable to assign table index for wireguard")
 		}
 
-		// If wireguard is enabled, update the failsafe ports to include the wireguard port.
-		failsafeInboundHostPorts := configParams.FailsafeInboundHostPorts
-		failsafeOutboundHostPorts := configParams.FailsafeOutboundHostPorts
-		if configParams.WireguardEnabled {
-			found := false
-			for _, i := range failsafeInboundHostPorts {
-				if i.Port == uint16(configParams.WireguardListeningPort) && i.Protocol == "udp" {
-					log.WithFields(log.Fields{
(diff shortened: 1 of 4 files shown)


coutinhop:
Hi @muff1nman,

First of all, thanks for your contribution!

The wireguard port is added to the failsafes (when enabled) in https://github.com/projectcalico/calico/blob/13bbda0aeec290e720fe43e0085d3cf50e2efa7d/felix/dataplane/driver.go#L186, which ultimately should add an allow rule for the wireguard port and be equivalent to what your changes are doing.

I'm just trying to understand, what was the issue you were seeing with default DROP? I'm thinking the fix may be something else (if wireguard is not working properly with default DROP)...

muff1nman:
@coutinhop  Based on our out-of-band conversation I've removed the failsafes logic entirely for wireguard and added any missing rules to the static rules (namely OUTPUT). Feel free to reiterate/summarize our conversation here especially if you feel the latest change doesn't meet your expectations. I'll plan to squash.

coutinhop:
@muff1nman thanks for these changes! That was exactly what I had in mind. @caseydavenport @fasaxc could you take a look and see if you can tell anything that could go wrong by switching from the failsafes to these static rules? Thanks!

/sem-approve

caseydavenport (approved):

coutinhop (approved):

Issue #6255, opened 2022-06-22
flag provided but not defined: -best-effort


nitinjagjivan:
Using calico v3.23.1
Today I am facing an issue while installing it.

kubectl get po -n kube-system
NAME                                                        READY   STATUS                  RESTARTS      AGE
calico-kube-controllers-6b77fff45-djdf5                     0/1     ContainerCreating       0             49m
**calico-node-gcxmw                                           0/1     Init:CrashLoopBackOff   6 (39s ago)   6m16s**
coredns-64897985d-qtt6l                                     0/1     ContainerCreating       0             49m
coredns-64897985d-wmjnw                                     0/1     ContainerCreating       0             49m

kubectl logs -n kube-system calico-node-gcxmw -c mount-bpffs
`flag provided but not defined: -best-effort
Usage of Calico:
  -allocate-tunnel-addrs
flag provided but not defined: -best-effort
    	Configure tunnel addresses for this node
  -allocate-tunnel-addrs-run-once
    	Run allocate-tunnel-addrs in oneshot mode
  -bird-live
    	Run bird liveness checks
  -bird-ready
    	Run BIRD readiness checks
  -bird6-live
    	Run bird6 liveness checks
  -bird6-ready
    	Run BIRD6 readiness checks
  -bpf
    	Run BPF debug tool
  -confd
    	Run confd
  -confd-confdir string
    	Confd configuration directory. (default "/etc/calico/confd")
  -confd-keep-stage-file
    	Keep stage file when running confd
  -confd-run-once
    	Run confd in oneshot mode
  -felix
    	Run Felix
  -felix-live
    	Run felix liveness checks
  -felix-ready
    	Run felix readiness checks
  -hostpath-init
    	Initialize hostpaths for non-root access
  -init
    	Do privileged initialisation of a new node (mount file systems etc).
  -monitor-addresses
    	Monitor change in node IP addresses
  -monitor-token
    	Watch for Kubernetes token changes, update CNI config
  -should-install-windows-upgrade
    	Check if Windows upgrade service should be installed.
  -show-status
    	Print out node status
  -shutdown
    	Do shutdown routine.
  -startup
    	Do non-privileged start-up routine.
  -status-reporter
    	Run node status reporter
  -threshold-time duration
    	Threshold time for bird readiness (default 30s)
  -upgrade-windows
    	Run Windows upgrade service.
  -v	Display version
`

Orchestrator version (e.g. kubernetes, containerd): k8s 1.23.4
[...]

PR #6261, opened 2022-06-22
Use GitHub links for manifests referenced in the docs

caseydavenport:
## Description

I think this is probably something we want to do, so putting this up for
comment. There are a couple of non-trivial downsides that we need to be
aware of, and potentially mitigate before we merge this.

This gives us a few benefits:

- Changes made to the manifests don't immediately go live on the docs site.
- Provides users with a stable URL, contents of the downloaded manifest are guaranteed not to change.

Downside:

- ~~Lose the ability to link to the latest calico.yaml file regardless of what the latest release is.~~
- Unable to roll out bugfixes to manifests without a patch release.

## Related issues/PRs

Example of a change that was rolled out before it was ready: https://github.com/projectcalico/calico/issues/6255

Deploy preview: https://deploy-preview-6261--calico-master.netlify.app/

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Documentation updated to use static per-patch raw.githubuercontent.com manifest links.
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
> Lose the ability to link to the latest calico.yaml file regardless of what the latest release is.

I suppose to mitigate this, we can still host the manifests at their old location with their old behavior. Just update the links in the docs so that there's a (default) option to use pinned manifests.

mgleung (approved):

PR #6276, opened 2022-06-24
support multinet for calicovpp by adding annotations to wep

hedibouattour:
## Description

Pod annotations are passed to a workloadendpoint new annotations field.
This was introduced for multiple networks feature in calico/vpp as a way to get the multinet info in network policies coming from felix, and then support multinet policies (policies applied to specific network interfaces as determined in their rules metadata annotations).


caseydavenport:
@hedibouattour could you share an example of the annotations that you're using? I wonder if they might actually be better represented as labels anyway.

hedibouattour:
There are actually two kinds of annotations, one for policies (rules), this is an example of it:
```
kind: NetworkPolicy
metadata:
  name: pol
spec:
  selector: demo == 'demo'
  ingress:
  - action: Allow
    protocol: TCP
    destination:
      ports:
      - 3434
  - action: Allow
    protocol: TCP
    metadata:
      annotations:
        extensions.projectcalico.org/network: "blue" 
    destination:
      ports:
      - 7809
  egress:
  - action: Allow
    protocol: TCP
    metadata:
      annotations:
        extensions.projectcalico.org/network: "red"    
    destination:
      ports:
      - 5555
```
and one for wep (the one that is meant by this PR). This one is actually automatically created by multus for a pod in multiple networks, it contains the different network interfaces for that pod and their addresses. So it is passed as it is to a wep in this PR. This is an example of it:
```
    k8s.v1.cni.cncf.io/networks-status: |-
      [{
          "name": "k8s-pod-network",
          "ips": [
              "172.16.166.191"
          ],
          "default": true,
          "dns": {}
      },{
          "name": "default/network-blue-conf",
          "ips": [
              "172.19.0.1"
          ],
          "dns": {}
      },{
          "name": "default/network-red-conf",
          "ips": [
              "172.21.0.1"
          ],
          "dns": {}
      }]
```

caseydavenport (changes requested):
My main concern with this is that for some clusters this might result in a fairly large occupancy change on every node, if they use a substantial amount of annotations.

I think we might want to use a filter to opt-in certain annotations, and not pass through annotations that we know our code doesn't use.

For example, perhaps we can only pass through annotations that are prefixed with `extensions.projectcalico.org` or `k8s.v1.cni.cncf.io` and for any pods that don't have such an annotation, we just pass a nil map.

libcalico-go/lib/backend/k8s/conversion/workload_endpoint_default.go
```diff
@@ -253,6 +253,15 @@ func (wc defaultWorkloadEndpointConverter) podToDefaultWorkloadEndpoint(pod *kap
 		AllowSpoofedSourcePrefixes: requestedSourcePrefixes,
 	}
 
+	for key := range pod.Annotations {
+		if strings.HasPrefix(key, "k8s.v1.cni.cncf.io/network-status") {
+			if true {}
```
caseydavenport, line 258: stray "if true" here

hedibouattour: oups, sorry

libcalico-go/lib/backend/k8s/conversion/workload_endpoint_default.go
```diff
@@ -253,6 +253,14 @@ func (wc defaultWorkloadEndpointConverter) podToDefaultWorkloadEndpoint(pod *kap
 		AllowSpoofedSourcePrefixes: requestedSourcePrefixes,
 	}
 
+	for key := range pod.Annotations {
```
caseydavenport, line 256: I think we can actually do this more efficiently, since we're matching on the exact key.

e.g.,

```go
if v, ok := pod.Annotations["k8s.v1.cni.cncf.io/network-status"]; ok {
  	if wep.Annotations == nil {
		wep.Annotations = make(map[string]string)
	}
        wep.Annotations["k8s.v1.cni.cncf.io/network-status"] = v
}
```

^ That way we don't need to loop through the annotations.

caseydavenport (approved):
One comment but otherwise LGTM

caseydavenport:
@hedibouattour I forgot to mention we should definitely have unit tests for this logic.

caseydavenport (approved):

PR #6284, opened 2022-06-27
Node - trim python from lib and usr lib

freecaykes:
## Description

Removing unnecessary python imports on Node
 
## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/Dockerfile.amd64
+++ b/node/Dockerfile.amd64
@@ -219,6 +219,7 @@ RUN /clean-up-filesystem.sh
 # Copy everything into a fresh scratch image so that naive CVE scanners don't pick up binaries and libraries
 # that have been removed in our later layers.
 FROM scratch
+# base ubi image has been cleaned from  clean-up-filesystem.sh of all unecessary imports
 COPY --from=ubi / /
 
 # Add in top-level license file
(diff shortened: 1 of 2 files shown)


node/Dockerfile.amd64
```diff
@@ -219,6 +219,7 @@ RUN /clean-up-filesystem.sh
 # Copy everything into a fresh scratch image so that naive CVE scanners don't pick up binaries and libraries
 # that have been removed in our later layers.
 FROM scratch
+# base ubi image has been cleaned from  clean-up-filesystem.sh of all unecessary imports
```
caseydavenport, line 222: I think this comment isn't needed, it's explained in the two comments above.

caseydavenport: At a minimum, should clean this up so that it starts with a capital letter "Base" and remove the extra space between "from" and "clean-up-filesystem" (there are two)

caseydavenport (approved):
LGTM - I think the comment added isn't really necessary, but if you want to leave it we should at least clean up the couple of formatting issues I called out in the comment. Code looks good though.

node/Dockerfile.amd64
freecaykes: https://github.com/projectcalico/calico/pull/6284/commits/5b19f25c502f9a6d308738d726ec285b4e899199 - removed

PR #6308, opened 2022-07-01
Add command to split IP pools

mgleung:
## Description

Add calicoctl command to split IP pools by CIDR.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add calicoctl command to split IP pools.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/calicoctl/calicoctl/commands/ipam/split.go
+++ b/calicoctl/calicoctl/commands/ipam/split.go
@@ -1,0 +1,248 @@
+// Copyright (c) 2022 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ipam
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"math/big"
+	"net"
+	"strconv"
+	"strings"
+
+	"github.com/docopt/docopt-go"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+
+	"github.com/projectcalico/calico/calicoctl/calicoctl/commands/clientmgr"
+	"github.com/projectcalico/calico/calicoctl/calicoctl/commands/common"
+	"github.com/projectcalico/calico/calicoctl/calicoctl/commands/constants"
+	"github.com/projectcalico/calico/calicoctl/calicoctl/util"
+	cnet "github.com/projectcalico/calico/libcalico-go/lib/net"
+	"github.com/projectcalico/calico/libcalico-go/lib/options"
+)
+
+func Split(args []string) error {
+	doc := constants.DatastoreIntro + `Usage:
+  <BINARY_NAME> ipam split <NUMBER> [--cidr=<CIDR>] [--name=<POOL_NAME>] [--config=<CONFIG>] [--allow-version-mismatch]
+
+Options:
+  -h --help                    Show this screen.
+  -c --config=<CONFIG>         Path to the file containing connection configuration in
+                               YAML or JSON format.
+                               [default: ` + constants.DefaultConfigPath + `]
+     --cidr=<CIDR>             CIDR of the IP pool to split.
+     --name=<POOL_NAME>        Name of the IP pool to split.
+     --allow-version-mismatch  Allow client and cluster versions mismatch.
+
+Description:
+  The ipam split command splits an IP pool specified by the specified CIDR or name
+  into the specified number of smaller IP pools. Each child IP pool will be of equal
+  size. IP pools can only be split into a number of smaller pools that is a power
+  of 2.
+
+Examples:
+  # Split the IP pool specified by 172.0.0.0/8 into 2 smaller pools
+  <BINARY_NAME> ipam split --cidr=172.0.0.0/8 2
+`
+	// Replace all instances of BINARY_NAME with the name of the binary.
+	name, _ := util.NameAndDescription()
+	doc = strings.ReplaceAll(doc, "<BINARY_NAME>", name)
+
+	parsedArgs, err := docopt.ParseArgs(doc, args, "")
+	if err != nil {
+		return fmt.Errorf("Invalid option: 'calicoctl %s'. Use flag '--help' to read about a specific subcommand.", strings.Join(args, " "))
+	}
+	if len(parsedArgs) == 0 {
+		return nil
+	}
+
+	err = common.CheckVersionMismatch(parsedArgs["--config"], parsedArgs["--allow-version-mismatch"])
+	if err != nil {
+		return err
+	}
+
+	// Validate that a name or CIDR is provided to search for the IP pool.
+	var oldPoolCIDR, oldPoolName string
+	if cidr := parsedArgs["--cidr"]; cidr != nil {
+		oldPoolCIDR = parsedArgs["--cidr"].(string)
+	}
+	if name := parsedArgs["--name"]; name != nil {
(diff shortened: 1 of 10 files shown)


calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,208 @@
...
+
+Description:
+  The ipam split command splits an IP pool specified by the specified CIDR into
+  the specified number of smaller IP pools. Each child IP pool will be of equal
+  size. IP pools can only be split into a number of smaller pools that is a power
+  of 2.
+
+Examples:
+  # Split the IP pool specified by 172.0.0.0/8 into 2 smaller pools
+  calicoctl ipam split 172.0.0.0/8 2
```
caseydavenport, line 58: I think "calicoctl" should be "<BINARY_NAME>" here.

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// Find the specified IP pool.
+	poolList, err := client.IPPools().List(ctx, options.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("Unable to list IP pools to find the pool specified by %s", parsedArgs["<CIDR>"])
+	}
+
+	var oldPool *apiv3.IPPool
+	for _, pool := range poolList.Items {
+		if pool.Spec.CIDR == parsedArgs["<CIDR>"] {
```
caseydavenport, line 101: I think we might want to support specification by name as well. 

Perhaps we need arguments:

```
--name: the name of the IP pool to split
--cidr: the CIDR of the IP pool to split
```

Or we just make the <CIDR> argument accept either a name or CIDR?

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,208 @@
...
+	splitCIDRs, err := splitCIDR(oldPool.Spec.CIDR, splitNum)
+	if err != nil {
+		return fmt.Errorf("Error splitting the CIDR %s into %d CIDRs: %v", oldPool.Spec.CIDR, splitNum, err)
+	}
+
+	// Create the new split pools using UnsafeCreate.
+	for i, cidr := range splitCIDRs {
+		splitPool := &apiv3.IPPool{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("split-%s-%d", oldPool.GetObjectMeta().GetName(), i+1),
```
caseydavenport, line 133: Probably good to start the indexing at 0, so `i` instead of `i+1`. Feels more natural (and consistent with other things that do this, like statefulsets)

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	splitCIDRs, err := splitCIDR(oldPool.Spec.CIDR, splitNum)
+	if err != nil {
+		return fmt.Errorf("Error splitting the CIDR %s into %d CIDRs: %v", oldPool.Spec.CIDR, splitNum, err)
+	}
+
+	// Create the new split pools using UnsafeCreate.
+	for i, cidr := range splitCIDRs {
+		splitPool := &apiv3.IPPool{
```
caseydavenport, line 131: I think we need to use a DeepCopy to make a copy here, rather than manually copying fields across. This is fragile when new fields are added, and already doesn't include everything we might want (e.g., labels and annotations from the old pool).

So we probably need:

```
newP := oldPool.DeepCopyObject()
newP.Name = ...
newP.Spec.CIDR = ...
```

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,208 @@
...
+	fmt.Printf("IP Pool %s was successfully split into %d smaller pools. Please refer to the documentation for final steps.", oldPool.GetObjectMeta().GetName(), splitNum)
+
+	return nil
+}
+
+func splitCIDR(oldCIDR string, parts int) ([]string, error) {
+	// Validate that we are trying to split the CIDR into a valid number of child CIDRs.
+	power := math.Log2(float64(parts))
+	if math.IsNaN(power) || power == 0 {
+		return []string{}, fmt.Errorf("Number to split CIDR into is not a valid power of 2: %d", parts)
```
caseydavenport, line 170: instead of `[]string{}` you should return `nil`

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,208 @@
...
+		return []string{}, fmt.Errorf("Error reading CIDR %s before attempting to split it", oldCIDR)
+	}
+
+	ones, bits := cidr.Mask.Size()
+	if int(power)+ones > bits {
+		return []string{}, fmt.Errorf("The CIDR %s is not large enough to be split into %d parts", oldCIDR, parts)
+	}
+
+	// Find the block size to increment over.
+	blockSize := new(big.Int).Exp(big.NewInt(2), big.NewInt(int64(bits-ones-int(power))), nil)
```
caseydavenport, line 185: `blockSize` is perhaps a confusing name for this, given it already has a meaning in the IPAM space. Maybe `newPoolSize` would be better?

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// Mask the IP in case the CIDR isn't formatted nicely.
+	mask := net.CIDRMask(ones, bits)
+	maskedIP := cnet.IP{IP: cidr.IP.Mask(mask)}
+
+	// Create the child CIDRs
+	splitCIDRs := make([]string, parts)
+	i := 0
+	for i < parts {
```
caseydavenport, line 194: Any reason not to do `for i :=0; i < parts; i++ {` ?

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	// Delete the old IP pool.
+	_, err = client.IPPools().Delete(ctx, oldPool.GetObjectMeta().GetName(), options.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("Error removing the IP pool that was split %s: %v", oldPool.GetObjectMeta().GetName(), err)
+	}
+
+	// Output follow-up directions.
+	fmt.Printf("IP Pool %s was successfully split into %d smaller pools. Please refer to the documentation for final steps.", oldPool.GetObjectMeta().GetName(), splitNum)
```
caseydavenport, line 161: We might want to print out more details on the pools that were created - e.g., their names.

I'm imagining some output like this:

```
IP Pool %s was successfully split into %d smaller pools.
Created default-0 with CIDR 192.168.0.0/17
Created default-1 with CIDR 192.168.128.0/17
```

Ideally, we'd also provide a `-o json` or `-o yaml` mode that simply printed out a json of the newly created pools so that this command could be more easily scripted against.

caseydavenport (commented):
Looking good - a few comments / thoughts.

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,246 @@
...
+
+		// Clear out unneeded metadata
+		splitPool.GetObjectMeta().SetUID("")
+		splitPool.GetObjectMeta().SetResourceVersion("")
+		splitPool.GetObjectMeta().SetGeneration(0)
+		splitPool.GetObjectMeta().SetSelfLink("")
+		splitPool.GetObjectMeta().SetCreationTimestamp(metav1.Time{})
+
+		// Set the new name
+		newName := fmt.Sprintf("split-%s-%d", oldPool.GetObjectMeta().GetName(), i)
```
caseydavenport, line 167: We might need to do something smarter here - k8s objects have a name character limit, and adding a prefix / suffix to an existing name might put us above that limit.

We need logic that says "if the name is > limit, then remove some characters and add a uniqueness marker to it".

This can be done in a follow-on PR though so long as you track it.

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -0,0 +1,246 @@
...
+		newGenName := ""
+		if oldPool.GetObjectMeta().GetGenerateName() != "" {
```
caseydavenport, line 169: Generate name doesn't make sense on pools, so I'd just zero this out no matter what.

caseydavenport (approved):
One comment, feel free to do in a follow-up if you'd like.

PR #6078, opened 2022-05-11
Allow calico-node to access root namespace in cgroup v2 to attach ctlb programs properly

mazdakn:
## Description
This PR fixes the issue related to the way cgroupv2 works and the change in the behavior of docker that places each container in its own cgroup name space by default. This issue is explained in https://github.com/projectcalico/calico/issues/5852

PR to cover operator changes: https://github.com/tigera/operator/pull/1957

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
eBPF: Move mount of BPFfs and cgroupv2 to a dedicated init container with elevated privileges; enter the root cgroup namespace to mount cgroupv2 in order to allow the CTLB to be installed system-wide. Reduce the mount privileges of the main calico-node container.
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #6305, opened 2022-06-30
Make sure cgroup2 fs default mount point exists on host

mazdakn:
## Description
This is a fix for https://github.com/projectcalico/calico/issues/6279 . `mountns` needs to make sure the mount point exists before trying to mount cgroup2 at the mount point.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

Issue #6279, opened 2022-06-27
CrashLoopBackoff with mount-bpffs init container, calico v3.23.2, operator v1.27.7


techmunk:
I'm currently spinning up fresh talos linux clusters. I'm install the CNI layer from the manifests at https://projectcalico.docs.tigera.io/archive/v3.23/manifests/tigera-operator.yaml which uses the operator version v1.27.7 since a day or two ago, which install calico v3.23.2.

Along with the CNI manifest above, I'm also adding the following inlineManifests. (Replace kubaapi_ip, and pod_subnet vars.).

```yaml
apiVersion: crd.projectcalico.org/v1
kind: FelixConfiguration
metadata:
  name: default
spec:
  wireguardEnabled: true
---
apiVersion: operator.tigera.io/v1
kind: Installation
metadata:
  name: default
spec:
  calicoNetwork:
    linuxDataplane: BPF
    ipPools:
      - blockSize: 26
        cidr: {{ pod_subnet }}
        encapsulation: VXLANCrossSubnet
        natOutgoing: Enabled
        nodeSelector: all()
---
apiVersion: operator.tigera.io/v1
kind: APIServer
metadata:
  name: default
spec: {}
```

Deploying this results in the following errors in the mount-bpffs container:
```
kubectl logs --container=mount-bpffs -n calico-system calico-node-hmv2l
2022-06-27 03:20:48.241 [INFO][1] init/startup.go 425: Early log level set to info
2022-06-27 03:20:48.242 [INFO][1] init/calico-init_linux.go 58: Checking if BPF filesystem is mounted.
2022-06-27 03:20:48.242 [INFO][1] init/calico-init_linux.go 70: BPF filesystem is mounted.
2022-06-27 03:20:48.243 [INFO][1] init/calico-init_linux.go 95: Checking if Cgroup2 filesystem is mounted.
2022-06-27 03:20:48.243 [ERROR][1] init/calico-init_linux.go 49: Failed to mount cgroup2 filesystem. error=failed to open /initproc/mountinfo: open /initproc/mountinfo: no such file or directory
```

Looking deeper into this, it seems that the mount changes recently introduced by https://github.com/projectcalico/calico/pull/6240/files, have not made it into the release-1.27 branch of the operator, but are on master via https://github.com/tigera/operator/pull/1957.

[...]

Issue #6256, opened 2022-06-22
Kubernetes cluster fails bringing up calico node


Rajalakshmi-Girish:
Our kubernetes clusters have been failing with the below error approximately from June 22nd 3am (IST)
Using Calico v3.23.1 from https://docs.projectcalico.org/manifests/calico.yaml
From calico node:
```
Warning  Failed     10s (x4 over 54s)  kubelet            Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/proc/1" to rootfs at "/initproc": change mount propagation through procfd: mount /initproc (via /proc/self/fd/6), flags: 0x44000: permission denied: unknown
```

From kubelet logs:

```
Jun 22 11:59:53 config1-1655881919-master kubelet[26537]: E0622 11:59:53.317459   26537 pod_workers.go:951] "Error syncing pod, skipping" err="failed to \"StartContainer\" for \"mount-bpffs\" with CrashLoopBackOff: \"back-off 2m40s restarting failed container=mount-bpffs pod=calico-node-465kz_kube-system(74560965-8be7-411c-9e00-a90dcd60f3d4)\"" pod="kube-system/calico-node-465kz" podUID=74560965-8be7-411c-9e00-a90dcd60f3d4
```

## Expected Behavior

The calico node and controller pods must come up `Running` without error.

## Current Behavior

The calico node pods are in `Init:CrashLoopBackOff` state and the calico controller is in `ContainerCreating` state.

## Possible Solution

This seems to be caused by the below change made around the same time:
https://github.com/projectcalico/calico/pull/6240

## Steps to Reproduce (for bugs)

https://prow.ppc64le-cloud.org/view/gs/ppc64le-kubernetes/logs/periodic-kubernetes-containerd-conformance-test-ppc64le/1539370251861364736
Setup kubernetes cluster using the manifest of calico v3.23.1. We see the calico node pos failing with Error.

* Calico version v3.23.1

Rajalakshmi-Girish:
@mazdakn @caseydavenport ^^

mkumatag:
But `v3.23.1` released a month ago 🤔

fasaxc:
The fix was a backport.  The manifests are "live" in the docs, they're not versioned down to the 0.0.x releases (well, they are but you have to download the release tarball to get a patch release manifest).

Rajalakshmi-Girish:
[...]

PR #6316, opened 2022-07-05
Add /proc and /var/run/calico to mount-bpffs volumes

mazdakn:
## Description

This PR fixes two bugs related to cgroupv2 fix: https://github.com/projectcalico/calico/pull/6078

1 - permission denied in runc container runtime, described here: https://github.com/projectcalico/calico/issues/6256 and also internally https://tigera.slack.com/archives/CKEP2NKFG/p1656529558783059

The fix is to change the volume needed by `mount-bpffs` init container from `/proc/1` to `/proc`. Also the volume is renamed from `init-proc` to `nodeproc`.

2 - `/var/run/calico/cgroup` does not exist on nodes, as described here: https://github.com/projectcalico/calico/issues/6279

The issue is already solved by PR: https://github.com/projectcalico/calico/pull/6305, which makes sure the path does exist in `mountns` binary. However, this PR adds an additional step in `nodeinit` package using the same method applied for `/sys/fs/` path.

Operator PR: https://github.com/tigera/operator/pull/2064

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/bpf/bpf.go
```diff
@@ -78,7 +78,7 @@ const (
 	sockmapEndpointsMapName    = "calico_sk_endpoints_" + sockmapEndpointsMapVersion
 
 	DefaultBPFfsPath = "/sys/fs/bpf"
-	CgroupV2Path     = "/run/calico/cgroup"
+	CgroupV2Path     = "/var/run/calico/cgroup"
```
mazdakn, line 81: `/var/run` is a symlink to `/var/` so these two path are identical, but thought to change it to prevent confusion.

fasaxc (approved):
Think you can just assume PID 1 will exist but LGTM otherwise.

PR #6320, opened 2022-07-05
felix/bpf: fix for WG no longer uses failsafes

tomastigera:
Previously, WG traffic was allowed by patching failsafe ports. It was broken by https://github.com/projectcalico/calico/pull/6250

We allow wireguard traffic between our nodes even if a user accidentally
block the traffic by a HEP policy.

We do that only if there was an actuall DENY by a bad policy to save
ourselves constantly testing whether to let it pass or not at the fast
path. The test requires 2 RT lookups.

When WG traffic is blocked by a bad policy, we allow it and only then
create conntrack entry to allow subsequent packets. Those entries are
kept alive by constant traffic (most of the time) or recreated after
they expire after policy is applied again.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: WG traffic is allowed by the HEP programs in case of a conflicting policy.
```

## Reminder for the reviewer

Other optional labels:


mazdakn (approved):
LGTM

marvin-tigera:

PR #6328, opened 2022-07-07
Fix for BPF UT map repinning flake

sridhartigera:
## Description

when repinning maps in UT, maps from the previous test linger around
for sometime. Hence make sure, maps are not listed in ```bpftool map list -j```.
Wait for maps to be completely removed from the kernel after
the fd is closed. Close the FDs, when there is error during
map resizing and upgrade.

Remove UpdatedByBpf flag for mock maps as we dont call the flow which closes the oldfd in all cases.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/mock/multiversion/v5/map.go
+++ b/felix/bpf/mock/multiversion/v5/map.go
@@ -23,14 +23,15 @@ import (
 )
 
 var MockMapParams = bpf.MapParameters{
-	Filename:   "/sys/fs/bpf/tc/globals/cali_mock",
-	Type:       "hash",
-	KeySize:    40,
-	ValueSize:  88,
-	MaxEntries: 1024,
-	Name:       "cali_mock",
-	Flags:      unix.BPF_F_NO_PREALLOC,
-	Version:    5,
+	Filename:     "/sys/fs/bpf/tc/globals/cali_mock",
+	Type:         "hash",
+	KeySize:      40,
+	ValueSize:    88,
+	MaxEntries:   1024,
+	Name:         "cali_mock",
+	Flags:        unix.BPF_F_NO_PREALLOC,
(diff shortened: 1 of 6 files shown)


felix/bpf/mock/multiversion/v5/map.go
```diff
@@ -23,15 +23,14 @@ import (
...
 var MockMapParams = bpf.MapParameters{
-	Filename:     "/sys/fs/bpf/tc/globals/cali_mock",
-	Type:         "hash",
-	KeySize:      40,
-	ValueSize:    88,
-	MaxEntries:   1024,
-	Name:         "cali_mock",
-	Flags:        unix.BPF_F_NO_PREALLOC,
-	Version:      5,
-	UpdatedByBPF: true,
```
tomastigera, line 34: I thought we needed this to emulate the maps being updated by bpf :thinking:  Is it related to the flake? Imo not.

sridhartigera: If we have ```UpdatedByBPF``` set to ```true```, we don't close the oldfd and do that in the other flow. We don't call that flow in the UTs in some of the tests. As a result, there is a reference to the map present even after the test which results in an error. Other option is to close the ```oldfd``` in ```Close()```, so that all the references are closed properly.

tomastigera: That might be the way to go. We want to test the "real thing" rather than to test what is easier to test ;-) Close() can do a test whether the oldfd is still valid.

tomastigera (approved):
LGTM, thanks!

PR #6337, opened 2022-07-07
cachingmap pkg

tomastigera:
## Description
    felix: cachingmap is BPF-independent package

    felix: caching map is BPF independent

    felix: cachingmap test in _test package

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/cachingmap/caching_map.go
```diff
@@ -260,16 +264,16 @@ func (c *CachingMap[K, V]) ApplyUpdatesOnly() error {
...
 	err := c.maybeLoadCache()
 	if err != nil {
 		return err
 	}
 	var errs ErrSlice
 	for k := range c.pendingDeletions {
 		err := c.dpMap.Delete(k)
-		if err != nil && !bpf.IsNotExists(err) {
-			logrus.WithError(err).Warn("Error while deleting from BPF map")
+		if err != nil && err != ErrNotExists {
```
fasaxc, line 275: Annoying little problem dealing with different exception types that mean the same thing.  Wonder if we should pass in an `shouldIgnoreDeletionError(error) bool` func or maybe change the code we use it with to not return a not exists error?

fasaxc (changes requested):
Just one thought that the change to error handling might bite us if we use it with some other underlying map later (or if we refactor the BPF map to return say, a wrapped error).

felix/cachingmap/caching_map.go
tomastigera: This gave me a hard time, and I originally though (and forgot to do) that the bpf.TypedMap would return this error on behalf of the bpf.Map. And that any other user would return this error as well. But I like the idea that the cachingmap.DataplaneMap would have `ShouldIgnoreDeletionError` or `IsEExist()` or something like that.

fasaxc (approved):
Ship it.

PR #5742, opened 2022-03-11
Add option to selectively disable rpf check for workloads

AloysAugustin:
## Description

This PR implements the feature described in #5643 .

This is a new feature, disabled by default and controlled by a felix configuration parameter: WorkloadSourceSpoofing
It has 2 possible values:
- None (default): maintain current behaviour, pods cannot spoof their IP
- Any: pods can ask to send packets from any address

A pod can request IP address spoofing with the following annotation:
cni.projectcalico.org/allowedSourceIPs

The value should be a JSON array of prefixes that the pod wants to use as source, for instance:
```
cni.projectcalico.org/allowedSourceIPs: "[\"10.96.0.0/16\", \"1.2.3.4/32\"]"
```
This feature is currently only implemented in the iptables dataplane.

## Implementation

This feature is implemented by adding a AllowSpoofedSourcePrefixes field in the various WorkloadEndpoint structures (is it really required to update all of them with this new field?). This field contains the list of IP addresses that the pod requests to be able to send traffic from, parsed from the annotation.
The contents of this field are checked in the felix ValidationFilter in order to ensure that they comply to the Felix configuration. If the validation doesn't pass, the WorkloadEndpoint is considered to be deleted (and as a result felix will not handle it any further). If the validation passes, the WorkloadEndpoint is passed to the dataplane by the calculation graph.

The dataplane can then handle this AllowSpoofedSourcePrefixes parameter to allow the pod to bypass the RPF check. This PR implements this in the iptables dataplane, by adding some rules that bypass the rpf check in the raw PREROUTING chain, and by disabling the kernel RPF check (rp_filter sysctl) for the interfaces that have source spoofing enabled for at least one address. This is handled in `felix/dataplane/linux/endpoint_mgr.go`.

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/5643

## Todos

- [x] Tests
- [x] Documentation
- [ ] Release note

## Release Note

Would a release not be required for this?

```release-note
None required
```

PR #6344, opened 2022-07-08
Add option to selectively disable rpf check for all workloads

ottoyiu:
## Description
We use a separate loadbalancing solution outside of Calico that uses IPIP encapsulation to implement a DSR-type loadbalancer.

Currently, the strict rpf checks on the iptables raw table causes response/DSR packets from the pod to the client to be dropped.

This PR introduces a mechanism that allows a list of CIDRs to be specified that are considered "safe", to be excluded from the RPF. In this scenario, the allowlist would be set to the VIP ranges of the loadbalancers.

`WorkloadSourceSpoofing` (#5742) is similarly to what we need, but that is on a workload to workload (pod to pod) basis, and we want this globally on every single workload without opt-in.

I currently have this implemented in the `endpoint_mgr.go` alongside to the `workloadSourceSpoofing` implementation, but these iptables rules maybe better suited programmed in the `DefaultRuleRenderer` in felix/rules/static.go. Before writing tests, I want to gather feedback to see if this is structured in the right location.

Suggestions and feedback is much appreciated! Would be curious to see if there's any appetite for a feature like this to be merged.

## Related issues/PRs

strict RPF check using iptables: https://github.com/projectcalico/calico/commit/60b41a07001a38b94e90e123d34ec535260b1ae1

selective disabling of rpf check for workloads: https://github.com/projectcalico/calico/pull/5742

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
Hey @ottoyiu sorry for the delay on response here - we've been pretty focused on v3.24 work. We'll take a look at this as a candidate for the v3.25 release.

ottoyiu:
@caseydavenport thanks for the reply! I'm still missing tests on those changes, but looking to see if it's the right place for those changes before I invest further time. When you have the chance, can you give a quick glance to see if the change seems sound?

coutinhop:
@ottoyiu I'm sorry for the delay, this PR must have slipped through the cracks... It seems reasonable and we're open to it if you still want to proceed, but the PR is pretty old so I'm going to close it. Please do reopen it if you want to move on with it.

PR #6372, opened 2022-07-12
felix/bpf: service routes are managed by a caching map

tomastigera:
## Description

    The caching map ensures that the desired state does not miss an update
    when, for example, a service IP is reause. That is, a service gets
    deleted and a new service with the same IP gets added in the same calc
    graph iteration, etc.
    
    The caching map updates the routetable.RouteTable which then ensures
    that the desired state is maintained in the dataplane.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1572,6 +1582,7 @@ func (m *bpfEndpointManager) ensureBPFDevices() error {
 	if err != nil {
 		return fmt.Errorf("failed to create nelink: %w", err)
 	}
+	defer nl.Close()
```
fasaxc, line 1585: Does this fix need to be backported?  Looks like we might leak a socket every time through this code

FWIW, if we're not going to mock out netlink for UT I'm not sure it makes sense to use the netlinkshim.  

The idea of the shim was to allow netlink to be swapped for a mock version.  Usually we keep a long-lived shim/handle around so that we avoid the overhead of connecting/disconnecting from netlink.  However, if there's an unexpected error, it's a good idea to close the handle and open a new one.

(Or perhaps we should have those tests that use a shim!)

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -273,7 +274,7 @@ func newBPFEndpointManager(
 		bpfMapContext:           bpfMapContext,
 		ifStateMap: cachingmap.New[ifstate.Key, ifstate.Value](ifstate.MapParams.Name,
 			bpf.NewTypedMap[ifstate.Key, ifstate.Value](
-				bpfMapContext.IfStateMap, ifstate.KeyFromBytes, ifstate.ValueFromBytes,
+				bpfMapContext.IfStateMap.(bpf.MapWithExistsCheck), ifstate.KeyFromBytes, ifstate.ValueFromBytes,
```
fasaxc, line 277: Can the type of `bpfMapContext.IfStateMap` be updated to MapWithExistsCheck?

fasaxc (changes requested):
Maybe I'm missing something but I'm not quite sure how this helps.  The RouteTable already has a cachingmap-like model where updates are cached and written out later; seems like this extra layer might be unneeded?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -664,6 +666,10 @@ func (m *bpfEndpointManager) markExistingWEPDirty(wlID proto.WorkloadEndpointID,
 }
 
 func (m *bpfEndpointManager) CompleteDeferredWork() error {
+	start := time.Now()
+	defer func() {
+		log.Infof("CompleteDeferredWork took %v", time.Since(start))
```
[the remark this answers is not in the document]
tomastigera: This wasn't suppose to be there, just for local debugging, will remove

tomastigera:
> Maybe I'm missing something but I'm not quite sure how this helps. The RouteTable already has a cachingmap-like model where updates are cached and written out later; seems like this extra layer might be unneeded?

Ohhh I know what what we wanted to fix, som eother stuff, closing this then,

PR #6377, opened 2022-07-13
felix/bpf: service routes maintenance fixed

tomastigera:
## Description

    This fix ensures that if a service IP is recycled during the same calc
    graph iteration, then we endup with the IP being present in the route
    table.
    
    Previously, it was possible to be programmed and then deleted by a service
    that was being removed if iterating over the dirtyServices was done in a
    wrong order.
    
    We remove all IPs immediately when they are deleted, while we add IPs
    of dirtyServices that should exist at the end of the iteration.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -669,10 +669,6 @@ func (m *bpfEndpointManager) markExistingWEPDirty(wlID proto.WorkloadEndpointID,
 }
 
 func (m *bpfEndpointManager) CompleteDeferredWork() error {
-	start := time.Now()
-	defer func() {
-		log.Infof("CompleteDeferredWork took %v", time.Since(start))
-	}()
 	// Do one-off initialisation.
 	m.dp.ensureStarted()
 
@@ -1956,7 +1952,7 @@ func (m *bpfEndpointManager) onServiceUpdate(update *proto.ServiceUpdate) {
 
 	ips4 := make([]ip.V4CIDR, 0, len(ips))
 	for _, i := range ips {
-		cidr, err := ip.CIDRFromString(i + "/32")
+		cidr, err := ip.ParseCIDROrIP(i)
 		if err != nil {
 			log.WithFields(log.Fields{"service": key, "ip": i}).Warn("Not a valid CIDR.")
 		} else if cidrv4, ok := cidr.(ip.V4CIDR); !ok {
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -1585,15 +1585,9 @@ func (m *bpfEndpointManager) ensureBPFDevices() error {
 		return nil
 	}
 
-	nl, err := netlink.NewHandle(syscall.NETLINK_ROUTE)
-	if err != nil {
-		return fmt.Errorf("failed to create nelink: %w", err)
-	}
-	defer nl.Close()
-
 	var bpfout, bpfin netlink.Link
 
-	bpfin, err = nl.LinkByName(bpfInDev)
+	bpfin, err := netlink.LinkByName(bpfInDev)
 	if err != nil {
 		nat := &netlink.Veth{
 			LinkAttrs: netlink.LinkAttrs{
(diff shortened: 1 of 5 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -218,8 +217,8 @@ type bpfEndpointManager struct {
 	bpfPolicyDebugEnabled bool
 
 	routeTable    *routetable.RouteTable
-	services      map[serviceKey][]string
-	dirtyServices map[serviceKey][]string
+	services      map[serviceKey][]ip.V4CIDR
+	dirtyServices map[serviceKey]struct{}
```
fasaxc, line 221: `set.Set[serviceKey]`?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -664,16 +669,28 @@ func (m *bpfEndpointManager) markExistingWEPDirty(wlID proto.WorkloadEndpointID,
 }
 
 func (m *bpfEndpointManager) CompleteDeferredWork() error {
+	start := time.Now()
+	defer func() {
+		log.Infof("CompleteDeferredWork took %v", time.Since(start))
```
fasaxc, line 674: See previous review, this will log every time through the main int dataplane loop whether we do any work here or not.  I don't think we can do that.  It's OK to log when the log is triggered by a change of a workload or a policy but logging more often than that is problematic

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1568,10 +1585,11 @@ func (m *bpfEndpointManager) ensureBPFDevices() error {
 		return nil
 	}
 
-	nl, err := netlinkshim.NewRealNetlink()
+	nl, err := netlink.NewHandle(syscall.NETLINK_ROUTE)
```
fasaxc, line 1588: If not shimming this, can you just use netlink's package-level API:
```
netlink.LinkByName()
```
Instead of 
```
Handle.LinkByName()
```
Netlink maintains a package level shared handle.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1931,7 +1949,35 @@ func (m *bpfEndpointManager) onServiceUpdate(update *proto.ServiceUpdate) {
 	ips = append(ips, update.ExternalIps...)
 
 	key := serviceKey{name: update.Name, namespace: update.Namespace}
-	m.dirtyServices[key] = ips
+
+	ips4 := make([]ip.V4CIDR, 0, len(ips))
+	for _, i := range ips {
+		cidr, err := ip.CIDRFromString(i + "/32")
```
fasaxc, line 1955:
```suggestion
		cidr, err := ip.ParseCIDROrIP(i)
```

felix/ip/ip_addr.go
```diff
@@ -255,6 +255,14 @@ func CIDRFromString(cidrStr string) (CIDR, error) {
 	return CIDRFromIPNet(cidr), nil
 }
 
+func CIDRFromStringMust(cidrStr string) CIDR {
```
fasaxc, line 258: Already have MustParseCIDROrIP, in case that's useful.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -331,8 +330,14 @@ func newBPFEndpointManager(
 			254,
 			opReporter,
 		)
-		m.services = make(map[serviceKey][]string)
-		m.dirtyServices = make(map[serviceKey][]string)
+		// Since we do not know what the state of services is at the start, mark
+		// the routes to be all cleaned up. As the service update will arrive,
+		// we will update the state.
+		if err := m.invalidateServiceRoutes(); err != nil {
```
fasaxc, line 336: I don't think you should even need this start-of-day clean up.  The RouteTable logic is built around the idea of "Calico owned routes".  It uses a matcher function to tell if a route belongs to us and if so, it automatically cleans it up if the manager layer doesn't tell it that the route should exist before the first Apply.

It's like how, to clear a caching map, all you need to do is create the caching map and then hit Apply(); it'll load the dataplane state and spot that the desired state is an empty map and clea up accordingly.

The only prereq for that is to have a good matcher function and I think you're in luck here since all your route point at a Calico-owned interface so if it points at that interface, it's ours.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1931,7 +1949,35 @@ func (m *bpfEndpointManager) onServiceUpdate(update *proto.ServiceUpdate) {
...
+			log.WithFields(log.Fields{"service": key, "ip": i}).Debug("Not a valid V4 CIDR.")
+		} else {
+			ips4 = append(ips4, cidrv4)
+		}
+	}
+
+	// Check which IPs have been removed (no-op if we haven't seen it yet)
+	for _, old := range m.services[key] {
+		exists := false
+		for _, svcIP := range ips4 {
```
fasaxc, line 1968: This is only dealing with frontend service IPs, right?  I expect there are only ~10 of those at most so the `O(n^2)` loop is OK.  For backends, it wouldn't be OK since there can be 100k backends (if every pod in the cluster has a prometheus endpoint, for example).

fasaxc (changes requested):
I _think_ you can rely on the RouteTable even more heavily; it should do the start-of-day clean up by mark and sweep: any routes you didn't tell it to add, it'll automatically clean up.

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: ok, wasnt sure about that behaviour, if it is so, great :+1:

mazdakn (approved):
LGTM

PR #6379, opened 2022-07-14
Tune conntrack timeout behaviour

fasaxc:
## Description

Building on #6315 after finding a further issue with the timeouts.  If the pre-retry function takes a long time (for example, cleaning conntrack in every felix...) then we can end up with an effectively short timeout.  For example, suppose the intended timeout was 10s, we might get:

- T=0s Check runs, takes 1-2s
- T=2s Pre-retry runs, takes 2-3 seconds
- T=5s Check attempt 2 runs, takes 1-2s
- T=7s Pre-retry runs, takes 2-3 seconds
- T=10s timeout, even though the last check started at T=5s

This change moves the timeout check to before the pre-retry function and allows extra leeway if the check itself took >2s.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


song-jiang (approved):
LGTM

PR #6381, opened 2022-07-14
felix: FeatureGate - to dis/enable features

tomastigera:
Sometimes we want to be able to include code that may not be finished yet
or 100% tested and we want to keep that code disabled in normal
production environments but we want to have it enabled in tests or in
specific user environments. Or we have a fix that breaks other things but
works for a particular corner case usecase without troubles.

In such cases we want to have the flexibility to configure felix easily
without introducing new config parameters that may disappear soon etc.

FeatureGate workis similar to FeatureDetectOverride which we
previously overloaded for this functionality. However,
FeatureDetectOverride is specifically used to overriding environment
feature detection in cases where the automatic detection mechanism is
not perfect and configures felix in a troublesome way.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
added FeatureGates to Felix
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/api/pkg/apis/projectcalico/v3/felixconfig.go
+++ b/api/pkg/apis/projectcalico/v3/felixconfig.go
@@ -118,12 +118,16 @@ type FelixConfigurationSpec struct {
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
 	// responsive when the lock is contended, but use more CPU. [Default: 50ms]
 	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeInterval,omitempty" configv1timescale:"milliseconds" confignamev1:"IptablesLockProbeIntervalMillis"`
-	// FeatureDetectOverride is used to override the feature detection.
-	// Values are specified in a comma separated list with no spaces, example;
-	// "SNATFullyRandom=true,MASQFullyRandom=false,RestoreSupportsLock=".
-	// "true" or "false" will force the feature, empty or omitted values are
-	// auto-detected.
+	// FeatureDetectOverride is used to override feature detection based on auto-detected platform
+	// capabilities.  Values are specified in a comma separated list with no spaces, example;
+	// "SNATFullyRandom=true,MASQFullyRandom=false,RestoreSupportsLock=".  "true" or "false" will
+	// force the feature, empty or omitted values are auto-detected.
 	FeatureDetectOverride string `json:"featureDetectOverride,omitempty" validate:"omitempty,keyValueList"`
+	// FeatureGates is used to enable or disable tech-preview Calico features.
+	// Values are specified in a comma separated list with no spaces, example;
+	// "BPFConnectTimeLoadBalancingWorkaround=enabled,XyZ=false". This is
+	// used to enable features that are not fully production ready.
+	FeatureGates string `json:"featureGates,omitempty" validate:"omitempty,keyValueList"`
 	// IpsetsRefreshInterval is the period at which Felix re-checks all iptables
 	// state to ensure that no other process has accidentally broken Calico's rules. Set to 0 to
 	// disable iptables refresh. [Default: 90s]
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -323,8 +323,8 @@ func newBPFEndpointManager(
 		m.dp = m
 	}
 
-	if config.FeatureDetectOverrides != nil {
-		m.ctlbWorkaroundEnabled = config.FeatureDetectOverrides["BPFConnectTimeLoadBalancingWorkaround"] == "enabled"
+	if config.FeatureGates != nil {
+		m.ctlbWorkaroundEnabled = config.FeatureGates["BPFConnectTimeLoadBalancingWorkaround"] == "enabled"
 	}
 
 	if m.ctlbWorkaroundEnabled {
--- a/api/pkg/openapi/openapi_generated.go
+++ b/api/pkg/openapi/openapi_generated.go
@@ -1976,7 +1976,14 @@ func schema_pkg_apis_projectcalico_v3_FelixConfigurationSpec(ref common.Referenc
 					},
 					"featureDetectOverride": {
 						SchemaProps: spec.SchemaProps{
-							Description: "FeatureDetectOverride is used to override the feature detection. Values are specified in a comma separated list with no spaces, example; \"SNATFullyRandom=true,MASQFullyRandom=false,RestoreSupportsLock=\". \"true\" or \"false\" will force the feature, empty or omitted values are auto-detected.",
+							Description: "FeatureDetectOverride is used to override feature detection based on auto-detected platform capabilities.  Values are specified in a comma separated list with no spaces, example; \"SNATFullyRandom=true,MASQFullyRandom=false,RestoreSupportsLock=\".  \"true\" or \"false\" will force the feature, empty or omitted values are auto-detected.",
+							Type:        []string{"string"},
+							Format:      "",
+						},
+					},
+					"featureGates": {
+						SchemaProps: spec.SchemaProps{
+							Description: "FeatureGates is used to enable or disable tech-preview Calico features. Values are specified in a comma separated list with no spaces, example; \"BPFConnectTimeLoadBalancingWorkaround=enabled,XyZ=false\". This is used to enable features that are not fully production ready.",
 							Type:        []string{"string"},
 							Format:      "",
 						},
--- a/felix/config/config_params.go
+++ b/felix/config/config_params.go
@@ -247,6 +247,7 @@ type Config struct {
 	IptablesLockTimeoutSecs            time.Duration     `config:"seconds;0"`
 	IptablesLockProbeIntervalMillis    time.Duration     `config:"millis;50"`
 	FeatureDetectOverride              map[string]string `config:"keyvaluelist;;"`
+	FeatureGates                       map[string]string `config:"keyvaluelist;;"`
 	IpsetsRefreshInterval              time.Duration     `config:"seconds;10"`
 	MaxIpsetSize                       int               `config:"int;1048576;non-zero"`
 	XDPRefreshInterval                 time.Duration     `config:"seconds;90"`
--- a/felix/dataplane/driver.go
+++ b/felix/dataplane/driver.go
@@ -352,6 +352,7 @@ func StartDataplaneDriver(configParams *config.Config,
 			KubeClientSet: k8sClientSet,
 
 			FeatureDetectOverrides: configParams.FeatureDetectOverride,
+			FeatureGates:           configParams.FeatureGates,
 
 			RouteSource: configParams.RouteSource,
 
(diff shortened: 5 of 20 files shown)


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -118,12 +118,19 @@ type FelixConfigurationSpec struct {
...
 	// "true" or "false" will force the feature, empty or omitted values are
 	// auto-detected.
 	FeatureDetectOverride string `json:"featureDetectOverride,omitempty" validate:"omitempty,keyValueList"`
+	// FeatureSelector is used to enable or disable Calico features.
+	// Values are specified in a comma separated list with no spaces, example;
+	// "BPFConnectTimeLoadBalancingWorkaround=enabled,XyZ=false". This is
+	// primarily used to enable feature that is not fully production ready or
+	// helps to fix a corner case in certain environments, but is not need most
+	// of the time.
+	FeatureSelector string `json:"featureSelector,omitempty" validate:"omitempty,keyValueList"`
```
caseydavenport, line 133: Kubernetes actually has a concept for this already, but it's called "FeatureGates" - https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

Might be worthwhile to align terminology? Very minor thing.

tomastigera: I like that!

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -118,12 +118,19 @@ type FelixConfigurationSpec struct {
...
 	// "true" or "false" will force the feature, empty or omitted values are
 	// auto-detected.
 	FeatureDetectOverride string `json:"featureDetectOverride,omitempty" validate:"omitempty,keyValueList"`
+	// FeatureGate is used to enable or disable Calico features.
+	// Values are specified in a comma separated list with no spaces, example;
+	// "BPFConnectTimeLoadBalancingWorkaround=enabled,XyZ=false". This is
+	// primarily used to enable feature that is not fully production ready or
+	// helps to fix a corner case in certain environments, but is not need most
+	// of the time.
+	FeatureGate string `json:"featureGate,omitempty" validate:"omitempty,keyValueList"`
```
caseydavenport, line 133: nit, but I think it should be FeatureGates (plural) since there are multiple independent gates that we expose.

Given this is in the v3 API, IMO it should really be a slice format rather than forcing a comma-separated list, which is hard for users to work with. Not a deal-breaker, but if it's not too much of a pain this would be nicer:

```
featureGates:
- name: BPFConnectTimeLoadBalancingWorkaround
  value: false

Comment (caseydavenport) [reply api/pkg/apis/projectcalico/v3/felixconfig.go:133]: I understand we already have a bunch of other slices in this struct represented as comma-separated values, so we don't need to block on this change. It's just too bad that we've sort of built that convention which is a bit awkward comparatively.

File: api/pkg/apis/projectcalico/v3/felixconfig.go [hunk 3]
```diff
@@ -118,12 +118,19 @@ type FelixConfigurationSpec struct {
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
 	// responsive when the lock is contended, but use more CPU. [Default: 50ms]
 	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeInterval,omitempty" configv1timescale:"milliseconds" confignamev1:"IptablesLockProbeIntervalMillis"`
-	// FeatureDetectOverride is used to override the feature detection.
+	// FeatureDetectOverride is used to override the environment feature detection.
```
caseydavenport, line 121: What is the environment feature detection?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -314,8 +314,8 @@ func newBPFEndpointManager(
 		m.dp = m
 	}
 
-	if config.FeatureDetectOverrides != nil {
```
caseydavenport, line 317: Isn't this a breaking change for anyone who might have set this already in their detection overrides?

api/pkg/apis/projectcalico/v3/felixconfig.go
caseydavenport: (just ignore the switch to a slice, but I do think the name should be plural)

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -118,12 +118,19 @@ type FelixConfigurationSpec struct {
...
 	// Values are specified in a comma separated list with no spaces, example;
 	// "SNATFullyRandom=true,MASQFullyRandom=false,RestoreSupportsLock=".
 	// "true" or "false" will force the feature, empty or omitted values are
 	// auto-detected.
 	FeatureDetectOverride string `json:"featureDetectOverride,omitempty" validate:"omitempty,keyValueList"`
+	// FeatureGate is used to enable or disable Calico features.
```
caseydavenport, line 127: "tech-preview Calico features" to be more clear

api/pkg/apis/projectcalico/v3/felixconfig.go
caseydavenport: Maybe "feature detection based on platform features"?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -118,12 +118,19 @@ type FelixConfigurationSpec struct {
...
+	// Values are specified in a comma separated list with no spaces, example;
+	// "BPFConnectTimeLoadBalancingWorkaround=enabled,XyZ=false". This is
+	// primarily used to enable feature that is not fully production ready or
```
caseydavenport, line 130: "to enable features that are not fully production ready"

> helps to fix a corner case in certain environments, but is not need most of the time.

I don't think this is the right use-case for feature gates. Those should be under FeatureDetectOverride, or honestly should probably just have their own config options, right? They are GA features that vary based on platform / environment? 

I think we want to be super clear on the scope of each and the boundary between the two:

- FeatureDetectOverride: for enabling or disabling GA features that are typically auto-configured. Used when you need to do something sort of funky that our auto-detection doesn't cover.
- FeatureGates: for enabling or disabling non-GA features that are in tech-preview or may disappear in the future.

Anything other than those two categories should get its own top-level field in the FelixConfiguraiton object. My expectation is that configuration of features would move from FeatureGates to a top-level field once we decide it is a GA feature.

tomastigera: sure, I think we are in agreement on this :-) Imo the gate should be for something that will be eventually always turned on once it settles.

We would introduce the gate just that the FeatureDetectOverride is used to override something that is detected on the platform.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -118,12 +118,17 @@ type FelixConfigurationSpec struct {
...
+	// FeatureDetectOverride is used to override the environment feature detection.
 	// Values are specified in a comma separated list with no spaces, example;
 	// "SNATFullyRandom=true,MASQFullyRandom=false,RestoreSupportsLock=".
 	// "true" or "false" will force the feature, empty or omitted values are
 	// auto-detected.
 	FeatureDetectOverride string `json:"featureDetectOverride,omitempty" validate:"omitempty,keyValueList"`
+	// FeatureGate is used to enable or disable Calico features.
+	// Values are specified in a comma separated list with no spaces, example;
+	// "BPFConnectTimeLoadBalancingWorkaround=enabled,XyZ=false". This is
+	// used to enable feature that is not fully production ready.
```
caseydavenport, line 130: should be "This is used to enabled or disable feature_s_ that _are_ not fully production ready"

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -118,12 +118,17 @@ type FelixConfigurationSpec struct {
...
+	FeatureGate string `json:"featureGate,omitempty" validate:"omitempty,keyValueList"`
```
caseydavenport, line 131: Maybe my comment got lost on the other commit, but since this is a list, the name should be plural.

"FeatureGates" instead of "FeatureGate"

felix/dataplane/linux/bpf_ep_mgr.go
caseydavenport: @tomastigera I think this is still a problem for anyone who upgrades

api/pkg/apis/projectcalico/v3/felixconfig.go
caseydavenport: @tomastigera I think you missed this suggestion. WDYT?

api/pkg/apis/projectcalico/v3/felixconfig.go
caseydavenport: @tomastigera I think you missed this one too

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: I do not think that this is an issue as that has not been documented anywhere and it is a half baked feature and was pretty much just for ourselves so that we could merge code and not to worry about it to make mess to anyone. The point of this PR is to switch to the gates before we make the ctlb workaround public.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -118,12 +118,16 @@ type FelixConfigurationSpec struct {
...
-	// Values are specified in a comma separated list with no spaces, example;
-	// "SNATFullyRandom=true,MASQFullyRandom=false,RestoreSupportsLock=".
-	// "true" or "false" will force the feature, empty or omitted values are
-	// auto-detected.
+	// FeatureDetectOverride is used to override feature detection based on auto-detected platform
+	// capabilities.  Values are specified in a comma separated list with no spaces, example;
+	// "SNATFullyRandom=true,MASQFullyRandom=false,RestoreSupportsLock=".  "true" or "false" will
+	// force the feature, empty or omitted values are auto-detected.
 	FeatureDetectOverride string `json:"featureDetectOverride,omitempty" validate:"omitempty,keyValueList"`
+	// FeatureGate is used to enable or disable tech-preview Calico features.
```
caseydavenport, line 126: Comment still says "FeatureGate" (singular)

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -118,12 +118,16 @@ type FelixConfigurationSpec struct {
...
+	// Values are specified in a comma separated list with no spaces, example;
+	// "BPFConnectTimeLoadBalancingWorkaround=enabled,XyZ=false". This is
+	// used to enable featureis that are not fully production ready.
```
caseydavenport, line 129: typo: "featureis" -> "features"

caseydavenport (approved):
Two minor typos but otherwise LGTM

marvin-tigera:

PR #6387, opened 2022-07-15
Add IPv6 wireguard to docs

coutinhop:
## Description

Update the docs for IPv6 wireguard.

Also add some missed IPv6 VXLAN updates to the docs.

## Todos

- [ ] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


calico/security/encrypt-cluster-pod-traffic.md
```diff
@@ -186,7 +211,7 @@ With the above command, {{site.prodname}} will not encrypt any of the pod traffi
 To enable encryption for pod traffic on node `my-node` again:
 
 ```bash
-calicoctl patch felixconfiguration node.my-node --type='merge' -p '{"spec":{"wireguardEnabled":true}}'
+calicoctl patch felixconfiguration node.my-node --type='merge' -p '{"spec":{"wireguardEnabled":true,"wireguardEnabledV6":true}}'
```
mgleung, line 214: I think instead of just providing a command to turn IPv4 and IPv6 wireguard on and off here, we should probably refer users to turn on wireguard only for the IP version

coutinhop: Good point! My intention was that this section was to serve as an example (for dual stack in this case), but I can see if that's not clear. I still think we might want to provide commands (to make it clear that we are patching the node-specific FelixConfig), so I could add the commands from v4-only, v6-only and v4+v6, directing the user to use whichever is appropriate? Or do you think that could be too confusing?

calico/security/encrypt-cluster-pod-traffic.md
```diff
@@ -132,22 +132,46 @@ To install WireGuard for OpenShift v4.8:
...
 <%
 
-Enable WireGuard encryption across all the nodes using the following command.
+Enable IPv4 WireGuard encryption across all the nodes using the following command.
 
 ```bash
 kubectl patch felixconfiguration default --type='merge' -p '{"spec":{"wireguardEnabled":true}}'
 ```
 
+Enable IPv6 WireGuard encryption across all the nodes using the following command.
```
caseydavenport, line 141: Remember that Wireguard supports v4 in v6 and v6 in v4 encap, so we should probably have a section here explaining what Calico implements when v6 / dual-stack wireguard is enabled so that it's blindingly obvious to the user what they should expect.

bmckercher123: Agree with Casey. Also, CE docs removed support for IPv6, so now we only have dual stack doc. https://docs.tigera.io/networking/ipv6. Was this a miss on our part to pick this to OS docs, or does OS still support IP v6?

caseydavenport: Calico still supports an IPv6-only mode (in addition to dual-stack)

coutinhop: Good point @caseydavenport! I remember you mentioning making it clear in code comments, I'll make sure I do this in the docs as well. Thanks @caseydavenport and @bmckercher123!

caseydavenport (approved):
LGTM - going to merge for now, but @bmckercher123 / @ctauchen if you still have some comments we can address in a follow up.

PR #6393, opened 2022-07-18
Added IPAMConfiguration to public API

song-jiang:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add IPAMConfiguration to projectcalico.org/v3 API
```

## Reminder for the reviewer

Other optional labels:


libcalico-go/lib/ipam/ipam.go
```diff
@@ -1805,24 +1806,57 @@ func (c ipamClient) GetAssignmentAttributes(ctx context.Context, addr net.IP) (m
...
-			config = &IPAMConfig{
-				AutoAllocateBlocks: true,
-				StrictAffinity:     false,
-				MaxBlocksPerHost:   0,
+	var obj *model.KVPair
+
+	var retries int
+	maxRetry := 5
+
+	// First, trying to get IPAM config (create if not exists) for 5 times.
```
song-jiang, line 1814: This part of the code is the only change I added to IPAM. It tries to create a default object if not exists.

api/pkg/apis/projectcalico/v3/ipamconfig.go
```diff
@@ -0,0 +1,68 @@
...
+
+	Items []IPAMConfig `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// IPAMConfig contains information about a block for IP address assignment.
+type IPAMConfig struct {
```
caseydavenport, line 42: Just musing... maybe this should be `IPAMConfiguration` to match `FelixConfiguration` and `BGPConfiguration`?

caseydavenport: How crazy / disruptive would that be?

libcalico-go/lib/ipam/ipam.go
mgleung: Out of curiosity, how did the IPAM Config creation used to get handled? Did we really just return the default in the absence of anything until someone finally wanted to make changes and saved something different?

api/pkg/apis/projectcalico/v3/ipamconfig.go
song-jiang: It would touch 69 files. How about I update the public API to use `IPAMConfiguration` in this PR and keep the current name for libcalico-go.

libcalico-go/lib/ipam/ipam.go
song-jiang: Currently the only way to create an actual object in the datastore is via calicoctl. It calls `ipam.SetIPAMConfig` function to apply the change.

api/pkg/apis/projectcalico/v3/ipamconfig.go
caseydavenport: That sounds exactly right to me.

api/pkg/apis/projectcalico/v3/ipamconfig.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// IPAMConfigurationSpec contains the specification for an IPPool resource.
+type IPAMConfigurationSpec struct {
+	// When StrictAffinity is true, borrowing IP addresses is not allowed.
+	StrictAffinity bool `json:"strictAffinity" validate:"required"`
+
+	// MaxBlocksPerHost, if non-zero, is the max number of blocks that can be
+	// affine to each host.
+	MaxBlocksPerHost int `json:"maxBlocksPerHost,omitempty"`
```
caseydavenport, line 56: For Kubernetes APIs I think this should be an explicit `int32` rather than `int`, per conventions.

libcalico-go/lib/backend/k8s/k8s.go
```diff
@@ -409,6 +415,7 @@ func (c *KubeClient) Clean() error {
 		apiv3.KindIPReservation,
 		apiv3.KindHostEndpoint,
 		apiv3.KindKubeControllersConfiguration,
+		apiv3.KindIPAMConfiguration,
```
caseydavenport, line 418: I think the `Kind` here needs to be the CRD kind (IPAMConfig) not the apiserver kind (IPAMConfiguration)

libcalico-go/lib/backend/k8s/resources/ipam_config.go
```diff
@@ -79,10 +80,62 @@ func (c ipamConfigClient) toV1(kvpv3 *model.KVPair) (*model.KVPair, error) {
...
+	case *model.IPAMConfig:
+		strictAffinity = obj.StrictAffinity
+		autoAllocateBlocks = obj.AutoAllocateBlocks
+		maxBlocksPerHost = obj.MaxBlocksPerHost
+	case *libapiv3.IPAMConfig:
+		strictAffinity = obj.Spec.StrictAffinity
+		autoAllocateBlocks = obj.Spec.AutoAllocateBlocks
+		maxBlocksPerHost = obj.Spec.MaxBlocksPerHost
+
+		// // For V3 resource update, creationTimestamp has to be presented.
```
caseydavenport, line 133: `// //`

caseydavenport: Also, we probably need to copy across other metadata as well - e.g., resourceVersion, labels, annotations, UID. 

We do that for our other CRDs here: https://github.com/projectcalico/calico/blob/7f586a208dfdac565a7733ef13415a77919f31ed/libcalico-go/lib/backend/k8s/resources/resources.go#L202-L205

libcalico-go/lib/backend/k8s/resources/ipam_config.go
```diff
@@ -79,10 +80,62 @@ func (c ipamConfigClient) toV1(kvpv3 *model.KVPair) (*model.KVPair, error) {
...
+	default:
+		log.Panic("ipamConfigClient : wrong key interface type")
+	}
+	return false
+}
+
+// For the first point, toV3 takes the given v1 KVPair and converts it into a v3 representation, suitable
 // for writing as a CRD to the Kubernetes API.
+//
+// Also note the name of the resource are hard coded to "default".
```
caseydavenport, line 116: This comment / function is a little different now, because it's possible that it will receive a v3 version of the object on input, right?

Should we just short-circuit if given an already assembled v3 object?

libcalico-go/lib/backend/k8s/resources/ipam_config.go
```diff
@@ -113,6 +167,12 @@ func (c *ipamConfigClient) Create(ctx context.Context, kvp *model.KVPair) (*mode
 	if err != nil {
```
caseydavenport, line 167: Here's an example - if `kvp` is already a v3 object, then `c.toV3` is a no-op right?

I think this code is getting a little bit (more) confusing with this change, since a few things are not clear:

- What does v3 mean? Is it "projectcalico.org/v3", "crd.projectcalico.org/v1", or both? 
- What is the expected input to this function? Looks like it can take multiple formulations of the data now.

I think at a minimum we should clearly document the expected types of data, but potentially we should have separate functions to handle this and split it:

```
func (c *client) Create(ctx, context.Context, kvp *model.KVPair) (kvp, error) {
  switch kvp {
    case fromIPAMCode:
      // From the IPAM code - we need to convert to CRD format.
      return c.createV1()
    case fromClientV3:
      // From the v3 client - it's already in CRD format.
      return c.createV3()
    default:
      panic("Bad type!") 
  }
}
```

libcalico-go/lib/ipam/ipam.go
```diff
@@ -1806,24 +1807,57 @@ func (c ipamClient) GetAssignmentAttributes(ctx context.Context, addr net.IP) (m
...
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
+
+		obj, err = c.client.Get(ctx, model.IPAMConfigKey{}, "")
+		if err != nil {
+			// Create the default config if it doesn't already exist.
+			if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
+				// Write to datastore.
+				new := &model.KVPair{
```
caseydavenport, line 1826: nit: shouldn't use the `new` keyword here. Probably OK to just use `kvp` instead.

libcalico-go/lib/ipam/ipam.go
```diff
@@ -1806,24 +1807,57 @@ func (c ipamClient) GetAssignmentAttributes(ctx context.Context, addr net.IP) (m
...
+				}
+
+				obj, err = c.client.Create(ctx, new)
+				if err != nil {
+					if _, ok := err.(cerrors.ErrorResourceAlreadyExists); ok {
+						log.Info("Failed to create global IPAM config; another node got there first.")
+						time.Sleep(1 * time.Second)
+						continue
+					}
+					log.WithError(err).Errorf("Error creating IPAM config")
```
caseydavenport, line 1842: We should make sure our CNI plugin and calico/node RBAC is updated to include `create` permissions on this, if it doesn't have it already.

libcalico-go/lib/clientv3/interface.go
```diff
@@ -56,6 +56,8 @@ type Interface interface {
 	KubeControllersConfiguration() KubeControllersConfigurationInterface
 	// CalicoNodeStatus returns an interface for managing CalicoNodeStatus resources.
 	CalicoNodeStatus() CalicoNodeStatusInterface
+	// IPAMConfig returns an interface for managing IPAMConfig resources.
+	IPAMConfig() IPAMConfigInterface
```
mgleung, line 60: nit: should this be `IPAMConfiguration` to match the other naming conventions? It doesn't particularly matter that it's abbreviated other than it doesn't quite match.

song-jiang: We try to keep libcalico-go files using the old name so that the name update has limited impact. Please see comment https://github.com/projectcalico/calico/pull/6393#discussion_r927814045

libcalico-go/lib/backend/k8s/resources/ipam_config.go
song-jiang: - v3 means the KVP is passed from `libcalico-go/lib/clientv3` with the format:
key: model.ResourceKey
value: *libapiv3.IPAMConfig

- v1 means the KVP is passed from IPAM code with the format (I presume it is the original format used by v1 backend client? )
key: model.IPAMConfigKey
value: *model.IPAMConfig

I think using separate functions is a much clearer approach. I'll update the code.

Issue #6074, opened 2022-05-11
BGP advertising doesn't honor externalTrafficPolicy Local for LoadBalancer services


cubagabriel:
For service type `LoadBalancer` and `externalTrafficPolicy: Local`, Calico will advertise routes from all nodes, regardless of whether valid endpoints exist.

A calico-node running in an host which has endpoints for that service will learn the endpoint IP address:

```
ubuntu@deployer:~$ klo -n kube-system calico-node-nsjfq  | grep 'default/my-service'
2022-05-03 22:16:24.089 [DEBUG][78] confd/routes.go 171: getEndpointsForService: service for endpoint not found, passing key="default/my-service"
2022-05-03 22:16:24.135 [DEBUG][78] confd/routes.go 171: getEndpointsForService: service for endpoint not found, passing key="default/my-service"
2022-05-03 22:16:24.200 [DEBUG][78] confd/routes.go 417: Advertising local service svc="default/my-service"
2022-05-03 22:16:24.200 [DEBUG][78] confd/routes.go 206: Checking routes for service advertise=true svc="default/my-service"
2022-05-03 22:16:24.200 [DEBUG][78] confd/routes.go 294: Setting routes for key key="default/my-service" routes=[]string{"10.100.19.72/32"}
```

On the other hand, a calico-node that has no endpoints for that service will recognize it correctly, as seen in the logs:

```
ubuntu@deployer:~$ kubectl logs -n kube-system calico-node-zhl2r  | grep 'default/my-service'
2022-05-03 22:16:24.088 [DEBUG][69] confd/routes.go 171: getEndpointsForService: service for endpoint not found, passing key="default/my-service"
2022-05-03 22:16:24.122 [DEBUG][69] confd/routes.go 171: getEndpointsForService: service for endpoint not found, passing key="default/my-service"
2022-05-03 22:16:24.190 [DEBUG][69] confd/routes.go 422: Skipping service with no local endpoints svc="default/my-service"
2022-05-03 22:16:24.190 [DEBUG][69] confd/routes.go 206: Checking routes for service advertise=false svc="default/my-service"

```

However, both calico-nodes have the same bird static configuration so both of them advertise all the prefixes:

```
[root@node4 /]# cat /etc/calico/confd/config/bird_aggr.cfg 
# Generated by confd

protocol static {
   # IP blocks for this host.
   route 10.233.111.0/24 blackhole;
   # Static routes.
   route 10.100.18.237/32 blackhole;
[...]

PR #6416, opened 2022-07-22
Properly refcount routes, fix accidental withdrawal of routes

caseydavenport:
## Description

Discovered in https://github.com/projectcalico/calico/issues/6074 and
https://github.com/projectcalico/calico/pull/6282.

BGPConfiguration and Services both provide static routes. Should they
provide the same route, we need to properly reference count them in
order to prevent accidental route withdrawals.

Example:

- BGPConfiguration contributes 10.0.0.1/32 
- Service A contributes 10.0.0.1/32 
- BGPConfiguration withdraws 10.0.0.1/32, replaces with 10.0.0.0/24 

We need to make sure we keep advertising 10.0.0.1/32, as contributed by the Service, even 
though BGPConfiguration no longer contributes that prefix. 

We actually have two layers of reference counting involved here - the
RouteGenerator reference counts prefixes since multiple services can
contribute the same CIDR, and the client further reference counts
prefixes from the RouteGenerator and BGPConfiguration. 

It's possible we could consolidate to a single layer, but it seems
conceptually correct and easier to test if we do this in two stages.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix occasional incorrect withdrawal of Service IPs over BGP when changing BGPConfiguration.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/confd/pkg/backends/calico/client.go
+++ b/confd/pkg/backends/calico/client.go
@@ -80,6 +80,21 @@ type backendClientAccessor interface {
 	Backend() api.Client
 }
 
+func NewRouteIndex() *RouteIndex {
+	return &RouteIndex{
+		programmedRoutes:       make(map[string]bool),
+		programmedRejectRoutes: make(map[string]bool),
+	}
+}
+
+// RouteIndex is a helper type for tracking which routes have been programmed
+// for a given route type (ExternalIP, LoadBalancer, ClusterIP) based on what we've
+// read from the BGPConfiguration API. Used to ensure we don't send duplicate routes.
+type RouteIndex struct {
+	programmedRoutes       map[string]bool
+	programmedRejectRoutes map[string]bool
+}
+
 func NewCalicoClient(confdConfig *config.Config) (*client, error) {
 	// Load the client clientCfg.  This loads from the environment if a filename
 	// has not been specified.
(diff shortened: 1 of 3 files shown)


aaaaaaaalex (approved):
I wouldn't be as aware of potential edge cases in this area as some others, so definitely seek another review other than mine, but at surface level, LGMT :smile: left one typo suggestion

caseydavenport:
Realized that the client code wasn't properly sending single updates for routes and reject routes, so working on some new commits to make sure it only sends a route if it needs to do so, and withdraws any that shouldn't be there in a more reliable way.

confd/pkg/backends/calico/client.go
```diff
@@ -121,18 +121,28 @@ func NewCalicoClient(confdConfig *config.Config) (*client, error) {
...
+		bgpPeers:                make(map[string]*apiv3.BGPPeer),
+		sourceReady:             make(map[string]bool),
+		nodeListenPorts:         make(map[string]uint16),
+		globalBGPConfig:         cfg,
+		nodeIPs:                 make(map[string]struct{}),
+		programmedRouteRefCount: make(map[string]int),
+
+		// Track which routes we have sent, and which we have not. We need maps for
+		// each of the three types of routes we track.
+		programmedRejectRoutesExt: make(map[string]bool),
```
caseydavenport, line 140: This is a bit of a hack, but I think what we need to do in order to track the routes we've sent more precisely. 

Would be better to encap this within a sub-component, which I'll do in a cleanup commit soon.

confd/pkg/backends/calico/client.go
```diff
@@ -1367,38 +1387,67 @@ func (c *client) GetLoadBalancerIPs() []*net.IPNet {
...
-	}
-
 	if !c.ExcludeServiceAdvertisement() {
-		// Withdraw the old CIDRs and add the new.
 		log.Info("Advertise global service ranges from this node")
-		c.addRoutesLockHeld(rejectKeyPrefix, rejectKeyPrefixV6, new)
-		c.addRoutesLockHeld(routeKeyPrefix, routeKeyPrefixV6, new)
-		c.deleteRoutesLockHeld(rejectKeyPrefix, rejectKeyPrefixV6, withdraws)
-		c.deleteRoutesLockHeld(routeKeyPrefix, routeKeyPrefixV6, withdraws)
+		for _, r := range routes {
```
caseydavenport, line 1401: This can all be simplified. It works, but the current form is more verbose than it needs to be ( can combine some branches)

mgleung (approved):
LGTM

PR #6420, opened 2022-07-22
Add block affinity API

mgleung:
## Description

Adds the block affinity object to the v3 API.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add the API for accessing block affinities with read-only permissions.
```

## Reminder for the reviewer

Other optional labels:


apiserver/pkg/storage/calico/blockAffinity_storage.go
```diff
@@ -0,0 +1,129 @@
...
+		converter:         BlockAffinityConverter{},
+	}, Codec: opts.RESTOptions.StorageConfig.Codec}
+	return dryRunnableStorage, func() {}
+}
+
+type BlockAffinityConverter struct {
+}
+
+func (gc BlockAffinityConverter) convertToLibcalico(aapiObj runtime.Object) resourceObject {
+	aapiBlockAffinity := aapiObj.(*aapi.BlockAffinity)
```
caseydavenport, line 86: Strictly speaking, this function should never be called, right? Since we only ever do read operations?

If so, we should probably just stub it out and log an error if it is called.

apiserver/pkg/storage/calico/blockAffinity_storage.go
```diff
@@ -0,0 +1,129 @@
...
+	lcgBlockAffinity.Spec.Deleted = false
+
+	return lcgBlockAffinity
+}
+
+func (gc BlockAffinityConverter) convertToAAPI(libcalicoObject resourceObject, aapiObj runtime.Object) {
+	lcgBlockAffinity := libcalicoObject.(*libapi.BlockAffinity)
+	aapiBlockAffinity := aapiObj.(*aapi.BlockAffinity)
+	// Copy spec but ignore internal field Deleted.
+	aapiBlockAffinity.Spec.State = lcgBlockAffinity.Spec.State
```
caseydavenport, line 106: The more I think about it, the more I think we might want to share the same struct here to avoid the need to copy fields individually. Will be a huge pain when we forget to do this when we add a new field.

I think we can do it in a Part II PR, though.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -138,6 +138,10 @@ func (c blockAffinityClient) parseKey(k model.Key) (name, cidr, host string) {
 // toV3 takes the given v1 KVPair and converts it into a v3 representation, suitable
 // for writing as a CRD to the Kubernetes API.
 func (c blockAffinityClient) toV3(kvpv1 *model.KVPair) *model.KVPair {
+	if !isV1Key(kvp1.Key) {
```
caseydavenport: I say remove this code, and let's make sure we never call this function with a v3 object.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -210,6 +255,44 @@ func (c *blockAffinityClient) DeleteKVP(ctx context.Context, kvp *model.KVPair)
 	return c.toV1(kvp)
 }
 
+func (c *blockAffinityClient) deleteKVPV3(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
+	// We need to mark as deleted first, since the Kubernetes API doesn't support
+	// compare-and-delete. This update operation allows us to eliminate races with other clients.
+	var err error
+	log.Debugf("This is the value: %+v: bool: %t", kvp.Key, kvp.Value == nil)
```
caseydavenport, line 262: can remove this log now :grin:

caseydavenport: Actually, `deleteKVPV3` can really just be a "NotImplemented" error, right? We don't support delete on the v3 API. And given we don't have the Value in hand... I don't think we ever will :laughing: :sob:

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -246,15 +329,55 @@ func (c *blockAffinityClient) Get(ctx context.Context, key model.Key, revision s
...
+
+func (c *blockAffinityClient) listV1(ctx context.Context, list model.BlockAffinityListOptions, revision string) (*model.KVPairList, error) {
 	l := model.ResourceListOptions{Kind: libapiv3.KindBlockAffinity}
 	v3list, err := c.rc.List(ctx, l, revision)
 	if err != nil {
 		return nil, err
 	}
 
-	host := list.(model.BlockAffinityListOptions).Host
-	requestedIPVersion := list.(model.BlockAffinityListOptions).IPVersion
```
caseydavenport, line 257: Huh, I think this is a pre-existing bug... bug we include affinities with Deleted=true when listing them, but not when performing a Get(). Oh well.

libcalico-go/lib/backend/k8s/k8s.go
```diff
@@ -217,6 +217,12 @@ func NewKubeClient(ca *apiconfig.CalicoAPIConfigSpec) (api.Client, error) {
 		libapiv3.KindIPAMConfig,
 		resources.NewIPAMConfigClient(cs, crdClientV1),
 	)
+	kubeClient.registerResourceClient(
```
caseydavenport, line 220: I am realizing now that this would have been a bit simpler if we simply had two different backend clients - e.g.,

```
registerResourceClient( NewV3BlockAffinityClient )
registerResourceClient (NewV1BlockAffinityClient )
```

But that's a project for another day.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
caseydavenport: Might actually need to fix this one.

apiserver/pkg/registry/projectcalico/rest/storage_calico.go
```diff
@@ -386,6 +387,28 @@ func (p RESTStorageProvider) NewV3Storage(
...
+			NewListFunc:   calicoblockaffinity.NewList,
+			GetAttrsFunc:  calicoblockaffinity.GetAttrs,
+			Trigger:       nil,
+		},
+		calicostorage.Options{
+			RESTOptions: blockAffinityRESTOptions,
+		},
+		p.StorageType,
+		authorizer,
+		[]string{"blockaffinity"},
```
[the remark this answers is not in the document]
mgleung: I thought it was everything that didn't match the resource name (in this case everything outside of `blockaffinities`). I put the singular name here, but I can also add some short names e.g. `affinity` and `affinities`.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
mgleung: I think we need to have an implementation here since using this backs the `clientv3` for the Kubernetes datastore. I can get rid of it if we do not plan on eventually standardizing IPAM code on the `clientv3` library, but if we do, then we will probably need this.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -210,6 +251,43 @@ func (c *blockAffinityClient) DeleteKVP(ctx context.Context, kvp *model.KVPair)
 	return c.toV1(kvp)
 }
 
+func (c *blockAffinityClient) deleteKVPV3(ctx context.Context, kvp *model.KVPair) (*model.KVPair, error) {
+	// We need to mark as deleted first, since the Kubernetes API doesn't support
+	// compare-and-delete. This update operation allows us to eliminate races with other clients.
+	var err error
+	nkvp := kvp
+	if kvp.Value == nil {
+		// Need to check if a value is given since V3 deletes are done by key only.
```
caseydavenport, line 260: > 		// Need to check if a value is given since V3 deletes are done by key only.
>		// Look up missing values with the provided key.

What do you mean by this comment, then? It seems to suggest that `kvp.Value` will always be nil in this function. What is the scenario where it isn't going to be nil, if any?

If we are going to do this, we need to ensure that we continue to use the original `kvp.Revision` when calling `Update` below, otherwise we're losing the compare-and-swap behavior that we're guaranteeing to the client.

mgleung: I believe we get to this logic because v3 deletes take in just the name in order to perform their deletes (I think the call that translates that to a KVP is here: https://github.com/projectcalico/calico/blob/master/libcalico-go/lib/clientv3/resources.go#L183). I don't think there should be a case _now_ where they will not be nil, but if we are going to use the `clientv3` calls eventually in our IPAM code (as opposed to the v1 calls we use now), then non-nil values can be passed into this logic.

I agree about the `kvp.Revision` portion. I filled it out without too much thought so I can fix that part up.

caseydavenport: > I don't think there should be a case now where they will not be nil

We should update the comment to say this then - the code and comment right now suggests that there is a path by which this might be `nil`. 

Either we always query it (with a comment explaining why) or we conditionally query it, but include a comment that says "We expect this to always be nil but we're just being safe"

caseydavenport: Actually, I think this is all irrelevant anyway - `DeleteKVP` isn't an operation on the v3 client. `v3` is always going to call `Delete`.

caseydavenport: Which I think is probably enough of a problem to say the IPAM code isn't going to use the clientv3 logic, ever. Adding an additional Get() is just not really worth it, imo.

mgleung: Doesn't the v3 client call `Delete` which then in turn calls `DeleteKVP` here: https://github.com/projectcalico/calico/blob/master/libcalico-go/lib/clientv3/resources.go#L183 ? At least my tests through the v3 client (which makes calls using `Delete`) also ran into these issues because of it.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -264,8 +382,10 @@ func (c *blockAffinityClient) List(ctx context.Context, list model.ListInterface
 		}
 		if host == "" || v1kvp.Key.(model.BlockAffinityKey).Host == host {
 			cidr := v1kvp.Key.(model.BlockAffinityKey).CIDR
-			cidr2 := &cidr
-			if requestedIPVersion == 0 || requestedIPVersion == cidr2.Version() {
+			cidrPtr := &cidr
+			// TODO: If we fix this inconsistency between Get and List, we break Watch in that it will not pick up Deleted events anymore.
+			//if (requestedIPVersion == 0 || requestedIPVersion == cidrPtr.Version()) && !v1kvp.Value.(*model.BlockAffinity).Deleted {
+			if requestedIPVersion == 0 || requestedIPVersion == cidrPtr.Version() {
```
mgleung, line 388: This is really unfortunate. Since `Watch` depends on this logic (and when we modify for deletion, we set the `Deleted` field), this actually removes any return value from any operation done with `Deleted` set to `"true"`. We might be able to do a hacky fix and make the return values from `Update` and `Delete` return values with the field forced to `"true"` just on the return value (not on what is actually stored), but that just seems wrong to return an object that wasn't what was written.

mgleung: Actually this is fixed by a separate bug in my code. Disregard.

api/pkg/apis/projectcalico/v3/blockaffinity.go
```diff
@@ -0,0 +1,84 @@
...
+	// The state of the block affinity with regard to any referenced IPAM blocks.
+	State string `json:"state"`
+
+	// The node that this block affinity is assigned to.
+	Node string `json:"node"`
+
+	// The CIDR range this block affinity references.
+	CIDR string `json:"cidr"`
+
+	// Deleted indicates that this block affinity is deleted.
```
caseydavenport, line 50: I think we should make this a `bool` on the v3 API and convert it when storing in v1 CRDs, since a bool is way more usable.

I think we should make this more clear what clients should do with this.

e.g.,

```
Deleted indicates whether or not this block affinity is disabled and is used as part of race-condition prevention. When set to true, clients should treat this block as if it does not exist.
```

apiserver/test/integration/clientset_test.go
```diff
@@ -1349,3 +1355,109 @@ func testIPAMConfigClient(client calicoclient.Interface, name string) error {
...
+			CIDR:    "10.0.0.0/24",
+			Node:    "node1",
+			State:   "pending",
+			Deleted: "false",
+		},
+	}
+	ctx := context.Background()
+
+	// Calico libv3 client instantiation in order to get around the API create restrictions
+	// TODO: Currently these tests only run on a Kubernetes datastore since profile creation
```
caseydavenport, line 1403: "profile"

apiserver/test/integration/clientset_test.go
```diff
@@ -1349,3 +1355,109 @@ func testIPAMConfigClient(client calicoclient.Interface, name string) error {
...
+	if err != nil {
+		return fmt.Errorf("error creating the object through the Calico v3 API '%v' (%v)", libV3BlockAffinity, err)
+	}
+
+	blockAffinityNew, err := blockAffinityClient.Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("error getting object %s (%s)", name, err)
+	}
+
+	blockAffinityList, err := blockAffinityClient.List(ctx, metav1.ListOptions{})
```
caseydavenport, line 1437: The primary use-case for this API involves watches, so I think we want a test that ensures Watches behave correctly as well.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -210,6 +251,43 @@ func (c *blockAffinityClient) DeleteKVP(ctx context.Context, kvp *model.KVPair)
...
+		nkvp, err = c.getV3(ctx, kvp.Key.(model.ResourceKey), kvp.Revision)
+		if err != nil {
+			if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
+				return nil, fmt.Errorf("Unable to find block affinity. Block affinity may have already been deleted.")
+			}
+			return nil, fmt.Errorf("Error retrieving block affinity for deletion: %s", err)
+		}
+	}
+
+	nkvp.Value.(*libapiv3.BlockAffinity).Spec.Deleted = fmt.Sprintf("%t", true)
```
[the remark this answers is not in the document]
caseydavenport: I would explicitly set it:

```
nkvp.Revision = kvp.Revision
```

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -264,8 +382,8 @@ func (c *blockAffinityClient) List(ctx context.Context, list model.ListInterface
 		}
 		if host == "" || v1kvp.Key.(model.BlockAffinityKey).Host == host {
 			cidr := v1kvp.Key.(model.BlockAffinityKey).CIDR
-			cidr2 := &cidr
-			if requestedIPVersion == 0 || requestedIPVersion == cidr2.Version() {
+			cidrPtr := &cidr
+			if (requestedIPVersion == 0 || requestedIPVersion == cidrPtr.Version()) && !v1kvp.Value.(*model.BlockAffinity).Deleted {
```
caseydavenport, line 386: I think this might be problematic. Think of this situation:

- Perform a List()
- Start a Watch()
- Receive a deleted event for a block that wasn't present in the List(). 

Might just be better to show "Deleted" affinities in the list.

caseydavenport (commented):
Looks good! A few minor comments though

api/pkg/apis/projectcalico/v3/blockaffinity.go
mgleung: If we do that, then we won't be able to eventually change the API to use the same Spec objects (which makes it easier to do the conversion since all we need to do is copy the spec objects in). I know you mentioned wanting to go this route in follow-up PRs earlier, do we still want to follow that?

apiserver/test/integration/clientset_test.go
mgleung: I meant to use the "profile" here since profiles are the actual reason that we only run KDD tests and not etcd tests. It's a tech debt piece we haven't really addressed yet. I can remove mentioning profiles if it's confusing but then might not be as clear when these eventually start failing due to the profile issue being cleaned up.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
mgleung: The whole `Get`, `List`, `Watch` flow is a little strange. If we make `List` and `Watch` consistent, then `Get` and `List` are not consistent (you can try to `Get` an affinity you just deleted and not see it, but then might notice it showing up in a `List`). My take is that `Watch` is always going to look a little funny because unless you understand what is happening in the backend and why, every single time you delete an affinity, you'll see 2 events: an update and a delete. I feel like it is probably more useful for `List` and `Get` to be more consistent since I picture them being more run by users from the command line vs `Watch` which I envision will be some process or logging that is monitoring affinity activity.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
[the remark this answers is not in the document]
mgleung: There is a little bit of an unintended consequence of this as well. Since the call for `Delete` from the v3 client can be only a name, the `ResourceVersion/Revision` may not be available in this call (though it always is in calls from the v1 client. Since the `ResourceVersion/Revision` is needed for the race condition checks (and the use of this call is intended to be internal anyways), I made `ResourceVersion/Revision` required if this is called from the v3 client on a Kubernetes datastore cluster. This should be only be used internally anyways so it's not a huge deal, but it's something to make note of.

apiserver/pkg/storage/calico/converter.go
```diff
@@ -111,6 +111,13 @@ func convertToAAPI(libcalicoObject runtime.Object) (res runtime.Object) {
 		aapi := &aapi.IPAMConfiguration{}
 		IPAMConfigConverter{}.convertToAAPI(lcg, aapi)
 		return aapi
+	// BlockAffinity works off of the libapi objects since
+	// the v3 client is used for mostly internal operations.
+	case *libapi.BlockAffinity:
```
mgleung, line 116: This is a consequence of not writing the v3 client to handle our external API objects. I'm thinking this goes with any follow-up PRs though

apiserver/test/integration/clientset_test.go
caseydavenport: ah ok

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
caseydavenport: I think for the v3 API we have the opportunity to make all three consistent - it's only the v1 client that needs to delete-on-get-if-deleted. 

Sine v3 is read-only, we should just make watch / get / list all consistent with each other. Probably by not filtering out "Deleted=true" at all.

mgleung: I think the one downside of making the v3 client behavior differ from the v1 client behavior is that we will not be able to eventually move the IPAM logic to use the v3 client because of the behavior mismatch. Not a big deal but it does mean we may have to change API behavior on users if we decide to go that route later.

caseydavenport: Yeah, that is a downside, but I don't see how we get around that. We probably just need to keep the two clients separate and deal with the fact that that's the way it is.

Otherwise, we're intentionally baking in a defective API experience for affinities into the v3 API in order to satisfy cleaner code

apiserver/test/integration/clientset_test.go
```diff
@@ -1349,3 +1355,142 @@ func testIPAMConfigClient(client calicoclient.Interface, name string) error {
...
+	w, err := blockAffinityClient.Watch(ctx, v1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("error watching block affinities (%s)", err)
+	}
+	var events []watch.Event
+	done := sync.WaitGroup{}
+	done.Add(1)
+	timeout := time.After(500 * time.Millisecond)
+	var timeoutErr error
+	// watch for 2 events
```
caseydavenport, line 1467: Strictly speaking, we don't need to run this in a goroutine. 

The watch result channel will buffer the events and then we can assert on them synchronously below.

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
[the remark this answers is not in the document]
caseydavenport: > Since the call for Delete from the v3 client can be only a name

The v3 client allows specification of a revision. If it's not present, the semantic is "Don't perform a compare-and-delete" and if it is present, the semantic is "use the revision to perform a CAD"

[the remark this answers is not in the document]
caseydavenport: In other words, it doesn't need to be mandatory, but clients generally want to specify it.

api/pkg/apis/projectcalico/v3/blockaffinity.go
```diff
@@ -0,0 +1,86 @@
...
+	// Standard object's metadata.
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of the BlockAffinity.
+	Spec BlockAffinitySpec `json:"spec,omitempty"`
+}
+
+// BlockAffinitySpec contains the specification for a BlockAffinity resource.
+type BlockAffinitySpec struct {
+	// The state of the block affinity with regard to any referenced IPAM blocks.
+	State string `json:"state"`
```
caseydavenport, line 42: Oh right, doh this should be an enum type with proper k8s-style values.

caseydavenport (approved):
Minor comments but otherwise LGTM

PR #5454, opened 2022-01-17
Added dummy routetable for network policy only mode

juanfresia:
## Description

This PR adds a config flag for `felix` named `RouteSyncDisabled`:
* The default value is `false`.
* Setting this to `true` will prevent `felix` from doing any updates on the route table.
* This is applied to all instances of `RouteTable` (IPv4, IPv6, VXLAN and Wireguard).

In order to make sure no operations are done against the linux routing table, a new implementation of `RouteTable` called `DummyTable` was added. While using the `DummyTable` all operations are skipped, enabling the use of `felix` as a network policy controller only.

## Related issues/PRs

More information on the context for this change can be found in #5247.

## Release Note

```release-note
Added dummy routetable for network policy only mode
```

PR #6473, opened 2022-07-31
Refactor RouteTable interface

song-jiang:
## Description

This is the follow up PR for https://github.com/projectcalico/calico/pull/5454 
Conversations here https://github.com/projectcalico/calico/pull/5454#issuecomment-1099494329

- Move RouteTable interface from `init-dataplane` to `routetable` directory and rename it to `RouteTableInterface`. 
- Add two functions `RouteUpdate` and `RouteRemove` to the interface.
- Allow wireguard package to use the interface. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/routetable/interface.go
```diff
@@ -0,0 +1,37 @@
...
+// RouteTableSyncer is the interface used to manage data-sync of route table managers. This includes notification of
+// interface state changes, hooks to queue a full resync and apply routing updates.
+type RouteTableSyncer interface {
+	OnIfaceStateChanged(string, ifacemonitor.State)
+	QueueResync()
+	Apply() error
+}
+
+// RouteTable is the interface provided by the standard routetable module used to program the RIB.
+type RouteTableInterface interface {
```
caseydavenport, line 31: nit, but I think normally this would just be "RouteTable" rather than include the word "Interface" in the name.

caseydavenport (approved):
One nit but otherwise LGTM

felix/routetable/interface.go
song-jiang: Yeah, I tried to avoid adding `interface` as well but there's already a RouteTable structure declared in the package. We have similar cases (e.g. IPAMConfigInterface) so I decided to do the same.

PR #4043, opened 2020-10-01
Add RBAC for calico-node to watch secrets, for BGP password feature

neiljerram:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

PR #6475, opened 2022-07-31
Add get/secrets permissions to calico-node rbac template.

ruslanloman:
Hi,
We faced with issue that calico-node couldn't read bgp password from secrets since there is no appropriate permissions for this role [1]. This PR adds get/secrets permission to calico-node rbac template.  There is the same issue in tigera-operator [2]. I found similar PR [3] and it was merged long time ago, but now i don't see this changes in master. 

Thanks!
[1] https://projectcalico.docs.tigera.io/reference/resources/bgppeer#bgppassword
[2] https://github.com/tigera/operator/issues/2006
[3] https://github.com/projectcalico/calico/pull/4043

### DIFF
--- a/charts/calico/templates/calico-node-rbac.yaml
+++ b/charts/calico/templates/calico-node-rbac.yaml
@@ -45,6 +45,14 @@ rules:
       # Used to discover Typhas.
       - get
 {{- end }}
+  # This permission is needed to get bgp password for BGPPeer resource.
+  - apiGroups: [""]
+    resources:
+      - secrets
+    verbs:
+      - get
+      - list
+      - watch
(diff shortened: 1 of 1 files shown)


ruslanloman:
That's what guys replied to me from calico tigera-operator project. This is intentional - we don't want calico/node to have secrets access unless its really needed, and even then only to one named secret.

What you should do is create your own Role and RoleBinding to provide access to the specific secret you need. I'm not sure why this is missing from the Calico OSS docs, but this is what should be there: https://docs.tigera.io/v3.14/security/comms/secure-bgp#ensure-rbac-permissions

PR #6483, opened 2022-08-02
remove unnecessary defer function wrappers

lucming:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/app-policy/checker/match_test.go
+++ b/app-policy/checker/match_test.go
@@ -151,9 +151,7 @@ func TestMatchHTTPNil(t *testing.T) {
 func TestPanicHTTPPaths(t *testing.T) {
 	RegisterTestingT(t)
 
-	defer func() {
-		Expect(recover()).To(BeAssignableToTypeOf(&InvalidDataFromDataPlane{}))
-	}()
+	defer Expect(recover()).To(BeAssignableToTypeOf(&InvalidDataFromDataPlane{}))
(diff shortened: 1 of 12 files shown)


libcalico-go/lib/ipam/ipam.go
```diff
@@ -992,16 +992,13 @@ func (c ipamClient) ReleaseIPs(ctx context.Context, ips ...ReleaseOptions) ([]ne
 		switch ip.Version() {
 		case 4:
 			pool, err = c.blockReaderWriter.getPoolForIP(*ip, v4Pools)
-			if err != nil {
```
caseydavenport, line 995: This is the only diff here that I would prefer to revert - we like to keep error handling immediately after the call that generates the error. It's slightly more verbose but is less prone to unexpected bugs in the future as this code changes.

caseydavenport (commented):
Looks good, but there's one diff that I'd like to revert.

libcalico-go/lib/ipam/ipam.go
lucming: > This is the only diff here that I would prefer to revert - we like to keep error handling immediately after the call that generates the error. It's slightly more verbose but is less prone to unexpected bugs in the future as this code changes.

okay, I've recovered

caseydavenport (approved):

caseydavenport:
Very sorry this didn't get merged. Looks like it slipped through the cracks and is now conflicting, so will need a rebase.

Given how old it is, I'm going to close it but please shout if you'd like it reopened.

PR #6487, opened 2022-08-02
Add ability to set MTU of Calico OpenStack nets

tj90241:
## Description

Add an annotation to the etcd datamodel used to describe
WorkloadEndpoints for Calico OpenStack VMs that specifies
 the MTU of the OpenStack network.

This effectively adds non-default (>1500) MTU support to
 Calico OpenStack.

## Related issues/PRs

none

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
networking-calico and thus calico-dhcp-agent now respect the MTU property of OpenStack networks. Upon requesting a DHCP lease, an option for interface-mtu is provided that corresponds to that of the OpenStack network in which the port resides.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/agent/dhcp_agent.py
+++ b/networking-calico/networking_calico/agent/dhcp_agent.py
@@ -15,7 +15,6 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-from collections import defaultdict
 import logging
 import netaddr
 import os
@@ -29,10 +28,6 @@ from neutron.agent.dhcp.agent import DhcpAgent
 from neutron.agent.dhcp_agent import register_options
 from neutron.agent.linux import dhcp
 from neutron.common import config as common_config
-try:
-    from neutron.common import constants as neutron_constants
-except ImportError:
-    from neutron_lib import constants as neutron_constants
 try:
     from neutron.conf.agent import common as config
 except ImportError:
@@ -116,7 +111,7 @@ class FakePlugin(object):
 
 
 def empty_network(network_id=NETWORK_ID,
-                  network_mtu=neutron_constants.DEFAULT_NETWORK_MTU):
+                  network_mtu=constants.DEFAULT_NETWORK_MTU):
     """Construct and return an empty network model."""
     return make_net_model({"id": network_id,
                            "subnets": [],
@@ -189,8 +184,7 @@ class CalicoEtcdWatcher(etcdutils.EtcdWatcher):
         # Cache of the ports that we've asked Dnsmasq to handle, for each
         # network ID.
         self._last_dnsmasq_ports = {}
-        mtu_dict = defaultdict(lambda: neutron_constants.DEFAULT_NETWORK_MTU)
-        self._last_dnsmasq_mtu = mtu_dict
+        self._last_dnsmasq_mtu = {}
 
         # Cache of current local endpoint IDs.
         self.local_endpoint_ids = set()
@@ -290,7 +284,7 @@ class CalicoEtcdWatcher(etcdutils.EtcdWatcher):
         # spec existed. To ensure that we do not break users who have endpoints
         # in etcd without the MTU annotation, fallback to the previous default.
         network_mtu = int(annotations.get(datamodel_v3.ANN_KEY_NETWORK_MTU,
-                          neutron_constants.DEFAULT_NETWORK_MTU))
+                                          constants.DEFAULT_NETWORK_MTU))
         allowedIps = [e.split('/')[0] for e in endpoint.get('allowedIps', [])]
         for addrm in endpoint['ipNetworks']:
             ip_addr = addrm.split('/')[0]
@@ -472,7 +466,9 @@ class CalicoEtcdWatcher(etcdutils.EtcdWatcher):
 
         # Compare that against what we've last asked Dnsmasq to handle.
         last_ports = self._last_dnsmasq_ports.get(network_id)
-        mtu_changing = self._last_dnsmasq_mtu.get(network_id) != net.mtu
+        last_mtu = self._last_dnsmasq_mtu.get(network_id,
+                                              constants.DEFAULT_NETWORK_MTU)
+        mtu_changing = last_mtu != net.mtu
         if ports_needed != last_ports or mtu_changing:
             # Requirements have changed, so start, restart or stop Dnsmasq for
             # that network ID.
@@ -519,10 +515,9 @@ class CalicoEtcdWatcher(etcdutils.EtcdWatcher):
         # information.
         LOG.debug("Reset cache for new snapshot")
         for network_id in list(self.agent.cache.get_network_ids()):
-            network = self.agent.cache.get_network_by_id(network_id)
             self.dirty_networks.add(network_id)
             self._fix_network_cache_port_lookup(network_id)
-            self.agent.cache.put(empty_network(network_id, network.mtu))
+            self.agent.cache.put(empty_network(network_id))
 
         # Suppress Dnsmasq updates until we've processed the whole snapshot.
         self.suppress_dnsmasq_updates = True
--- a/networking-calico/networking_calico/agent/dhcp_agent.py
+++ b/networking-calico/networking_calico/agent/dhcp_agent.py
@@ -15,6 +15,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
+from collections import defaultdict
 import logging
 import netaddr
 import os
@@ -114,13 +115,14 @@ class FakePlugin(object):
     get_networks = None
 
 
-def empty_network(network_id=NETWORK_ID):
+def empty_network(network_id=NETWORK_ID,
+                  network_mtu=neutron_constants.DEFAULT_NETWORK_MTU):
     """Construct and return an empty network model."""
     return make_net_model({"id": network_id,
                            "subnets": [],
                            "ports": [],
                            "tenant_id": "calico",
-                           "mtu": neutron_constants.DEFAULT_NETWORK_MTU})
+                           "mtu": network_mtu})
 
 
 def copy_network(source_net):
@@ -187,6 +189,8 @@ class CalicoEtcdWatcher(etcdutils.EtcdWatcher):
         # Cache of the ports that we've asked Dnsmasq to handle, for each
         # network ID.
         self._last_dnsmasq_ports = {}
+        mtu_dict = defaultdict(lambda: neutron_constants.DEFAULT_NETWORK_MTU)
+        self._last_dnsmasq_mtu = mtu_dict
 
         # Cache of current local endpoint IDs.
         self.local_endpoint_ids = set()
@@ -282,6 +286,11 @@ class CalicoEtcdWatcher(etcdutils.EtcdWatcher):
         dns_assignments = []
         fqdn = annotations.get(datamodel_v3.ANN_KEY_FQDN)
         network_id = annotations.get(datamodel_v3.ANN_KEY_NETWORK_ID)
+        # The MTU annotation was retroactively added to the datamodel after the
+        # spec existed. To ensure that we do not break users who have endpoints
+        # in etcd without the MTU annotation, fallback to the previous default.
+        network_mtu = int(annotations.get(datamodel_v3.ANN_KEY_NETWORK_MTU,
+                          neutron_constants.DEFAULT_NETWORK_MTU))
         allowedIps = [e.split('/')[0] for e in endpoint.get('allowedIps', [])]
         for addrm in endpoint['ipNetworks']:
             ip_addr = addrm.split('/')[0]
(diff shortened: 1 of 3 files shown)


networking-calico/networking_calico/agent/dhcp_agent.py
```diff
@@ -457,10 +471,12 @@ def _update_dnsmasq(self, network_id):
         ports_needed.sort(key=lambda port: port.id)
 
         # Compare that against what we've last asked Dnsmasq to handle.
-        if ports_needed != self._last_dnsmasq_ports.get(network_id):
+        last_ports = self._last_dnsmasq_ports.get(network_id)
+        mtu_changing = self._last_dnsmasq_mtu.get(network_id) != net.mtu
```
tj90241, line 475: If the network is initially configured with 9k MTU, and then dropped, this may not trigger... because the `defaultdict` assumes that the standard Neutron MTU of 1500 is used? 🤔

tj90241:
There is a really nasty side effect with, at least, Cirros images here... the default route is not making it to the instance anymore if the instance's taps are at 1500 MTU and the OpenStack network is raised to 9000 MTU. You can reproduce this by e.g.:
* Creating a network with 1500 MTU
* Launching instances
* Changing the network to 9000 MTU
* Observing that `dnsmasq` restarts, the VMs are offered `interface-mtu` option of 9k
* Instances still get a lease but lack a default route

One must also use this option with care - it's not possible to live-migrate VMs once the network MTU changes wrt how they are presently configured... `libvirt/qemu` abort out at the end of migration and fail to move the VM. One must hard reboot all VMs in the network to get them to observe a change.
```
Live Migration failure: unsupported configuration: Target network card MTU 9000 does not match source 1500: libvirt.libvirtError: unsupported configuration: Target network card MTU 9000 does not match source 1500
```

networking-calico/networking_calico/agent/dhcp_agent.py
```diff
@@ -282,6 +286,11 @@ def on_endpoint_set(self, response, name):
         dns_assignments = []
         fqdn = annotations.get(datamodel_v3.ANN_KEY_FQDN)
         network_id = annotations.get(datamodel_v3.ANN_KEY_NETWORK_ID)
+        # The MTU annotation was retroactively added to the datamodel after the
+        # spec existed. To ensure that we do not break users who have endpoints
+        # in etcd without the MTU annotation, fallback to the previous default.
```
neiljerram, line 291: FWIW, this sounds fine to me.  I don't see a need for a v4 datamodel.

networking-calico/networking_calico/agent/dhcp_agent.py
neiljerram: Sorry, I don't understand the concern here.  Please explain further.

neiljerram: Separately, as this is the only place that accesses `self._last_dnsmasq_mtu`, I'd be inclined to make it a normal dict and use `self._last_dnsmasq_mtu.get(network_id, neutron_constants.DEFAULT_NETWORK_MTU)` here.

networking-calico/networking_calico/agent/dhcp_agent.py
```diff
@@ -502,9 +519,10 @@ def _pre_snapshot_hook(self):
         # information.
         LOG.debug("Reset cache for new snapshot")
         for network_id in list(self.agent.cache.get_network_ids()):
+            network = self.agent.cache.get_network_by_id(network_id)
             self.dirty_networks.add(network_id)
             self._fix_network_cache_port_lookup(network_id)
-            self.agent.cache.put(empty_network(network_id))
+            self.agent.cache.put(empty_network(network_id, network.mtu))
```
neiljerram, line 525: I think we don't need these changes, because it's OK for networks to have the default MTU at this point?  As endpoints are processed from the snapshot, the relevant networks' MTU will be updated as needed.

neiljerram (changes requested):
A few thoughts and suggested tweaks here - WDYT?

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
```diff
@@ -138,14 +138,16 @@ def neutron_to_etcd_write_data(self, port, context, reread=False):
             except n_exc.PortNotFound:
                 raise ResourceGone()
         port = self.add_extra_port_information(context, port)
+        network = self.db.get_network(context, port['network_id'])
```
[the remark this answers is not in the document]
tj90241: I ironically had thought about this while writing the code as it also prevents dedup of the need to `self.db.get_network`, but `endpoint_annotations` takes `port`... though now realizing that's really just a `dict` and it should be fine to slot MTU in... will make this change.

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
```diff
@@ -410,8 +412,11 @@ def endpoint_spec(port):
     return data
 
 
-def endpoint_annotations(port):
-    annotations = {datamodel_v3.ANN_KEY_NETWORK_ID: port['network_id']}
+def endpoint_annotations(network, port):
+    annotations = {
+        datamodel_v3.ANN_KEY_NETWORK_ID: port['network_id'],
+        datamodel_v3.ANN_KEY_NETWORK_MTU: str(network['mtu']),
```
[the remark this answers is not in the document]
tj90241: `self.db.get_network` always returns a dict containing an `mtu` key during testing, even for networks created without an MTU specified... though I can guard this with a `get` to be certain.

networking-calico/networking_calico/agent/dhcp_agent.py
tj90241: I've overthought this and had put in the `defaultdict` while working on a way to ensure that `dnsmasq` is restarted for any MTU change, but yes -- now realizing that none of it was probably needed. Let me do some quick testing and I'll make those changes.

networking-calico/networking_calico/agent/dhcp_agent.py
tj90241: Yes, good point, had also overthought this.

tj90241: Removed in latest patch, still tests good.

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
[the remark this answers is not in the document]
tj90241: Added a guard to fallback to `constants.DEFAULT_NETWORK_MTU` in latest patch.

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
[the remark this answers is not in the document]
tj90241: Call signatures are not changed in the latest patch.

tj90241:
I'm a bit baffled as I don't understand what changed, but the review comments addressed the issue with the instances no longer getting a default route in Cirros when changing the MTU on instances which were already built.

I also noticed that `dhcp_agent` was doing a try/except to import `neutron_constants` for different OpenStack releases, but this mechanism already existed in the `compat` module. So I removed the try/except and just imported `constants` from the `compat` module in the latest patch.

neiljerram (approved):
Many thanks @tj90241 , these code changes look good to me, but I just wanted to check on the broader issues that you mentioned.

Firstly, you talked about a network MTU change as triggering an update to all endpoints on that network, and that that might take hour(s) in a big cluster.  But I don't think a network MTU change on its own will actually trigger those updates, will it?  Rather, I think the result of the changes here is that when an endpoint is created or changed in some other way, it will then pick up the current MTU for its network.  Is that right?  I suppose it could be that Neutron sends an update for all ports to the ML2 driver, when a network property is changed, but I didn't think it did that, and it feels contrary to the idea of the ML2 driver also being able to hook network update events (even if Calico doesn't currently take advantage of that).

If it is right, then
- is that an acceptable semantic for you?
- I wonder if we should document it somewhere - WDYT?
- I think it removes the concern about following updates taking hour(s) - correct, or is there another mechanism coming into play here?

Secondly, the issue with live migrating when the network MTU has changed but the endpoint not yet updated.  Do you think this is a point that you can live with?  I guess if the workflow is to configure the network MTU first before creating endpoints on the network, that would avoid that problem.

Finally the issue with CirrOS - I suppose you'll monitor to check that it really has gone permanently away now.  In any case it feels like something we could dig into further if needed.

tj90241:
whoops, forgot to update the unit tests... will do that today.

> Rather, I think the result of the changes here is that when an endpoint is created or changed in some other way, it will then pick up the current MTU for its network. Is that right?
Certainly, *new* ports created immediately after an MTU change work exactly as you say. Added a more elaborate example of my concern below.

> I suppose it could be that Neutron sends an update for all ports to the ML2 driver, when a network property is changed, but I didn't think it did that,
It does not -- perhaps I didn't explain myself well, or there's some aspect of `etcd` that I'm missing here. The challenge is unique to Calico in this prospective:

Let's suppose we change the MTU of the OpenStack network that has a large number of pre-existing ports. As per above, any new ports created immediately after this change work just like we expect. However, the next time that the [ResourceSyncer](https://github.com/projectcalico/calico/blob/79b442a53adb7d7f1fd62927d9322daf87dce9de/networking-calico/networking_calico/plugins/ml2/drivers/calico/syncer.py#L25) runs, it will [observe that the WorkloadEndpoints for _existing_ ports all change](https://github.com/projectcalico/calico/blob/79b442a53adb7d7f1fd62927d9322daf87dce9de/networking-calico/networking_calico/plugins/ml2/drivers/calico/syncer.py#L132) as a result of the MTU annotation itself being added or changed, and begin rewriting each WorkloadEndpoint serially, one at a time.

When there are thousands or more of ports/WorkloadEndpoints in the cluster, this operation takes a very long time to complete. During this time, I expect that e.g., if a `calico-dhcp-agent` is bounced, it would observe that the MTU of some WorkloadEndpoints is set to a stale value, and for others it is the "current" value. As such, I don't think the MTU value that we update `dnsmasq` with would be deterministic -- it's the last port that gets processed by `calico-dhcp-agent` when `_update_dnsmasq` is called which determines the MTU used -- this may be a stale value?

It's quite likely that during this scenario, the ResourceSyncer will soon stumble on another port which is on the same host and again cause `dnsmasq` to be reloaded, I think, though -- so it's not a terribly huge concern. And, of course eventually, the ResourceSyncer will enumerate over all WorkloadEndpoints. However, an operator may not be aware of the above and can be caught by surprise.

For us, this is perfectly acceptable.

If my thinking above is right: Yes, I think this should be documented somewhere. Additionally, in the future, adding some kind of resource to the datamodel that represents an OpenStack network where MTU could be captured would make this slightly prickly in that it would allow the ResourceSyncer to make an update to a single resource, which all `calico-dhcp-agent`s could observe at the same time.

> Secondly, the issue with live migrating when the network MTU has changed but the endpoint not yet updated. Do you think this is a point that you can live with?
Yes -- on this front, I'm working on a Nova patch that retains the MTU. in the libvirt XML during migration instead of it adapting to whatever the network currently has. Hopefully upstream will accept it -- if not, we'll still make use of it, but others may stumble upon this.

> Finally the issue with CirrOS - I suppose you'll monitor to check that it really has gone permanently away now. In any case it feels like something we could dig into further if needed.
Yes, I've begun putting this change on our test clusters and we intend to begin phasing this feature in imminently as well. I do not see it being a problem now and will be prepared to deal with it if it comes up.

tj90241:
@neiljerram - could you please have a look at the CI tests? Looking closer now, a lot of the errors seem unrelated...

On another note, I have some further updates:
- I was able to reproduce the aforementioned Cirros issue again. It occurs under a unique set of circumstances that can only hold true if one makes code changes to `nova` (see next bullet)...
- Upstream does not support live-migration of a `libvirt`-hosted instance if any source VIF has an MTU which does not match that of the Neutron network (or, if the MTU of the VIF would change during live migration). They have actually turned down the patch I pushed which would have allowed for the instance to still migrate with a MTU that does not match the Neutron network. We still have a way to do this internally I think, but it goes outside of the scope of the code changes here.
- However, we probably need a release note that says one cannot live-migrate instances without hard rebooting them if and when the associated Neutron network MTU is changed. A future version of `nova` will hopefully have an explicit error on this, ref this WIP PS: https://review.opendev.org/c/openstack/nova/+/852367/1

neiljerram:
Thanks @tj90241 for the above notes.  Of course you are right about traversing all endpoints during a resync, and I'm sorry that I forgot about that!

It looks like there are two non-trivial next steps here:
1. I think we should look at passing MTU to the DHCP agent by implementing the network hooks in the ML2 driver and adding a network object to the datamodel.  That would allow an MTU change to be signaled immediately to the agent, and without having to update all of the ports on the network.
2. As you've noted, there appear to be unrelated problems in the CI - perhaps we haven't fully pinned the OpenStack code that we are testing against?

I'd like to dig into these more, but it's a matter of scheduling, so please let's have a (parallel) conversation about that...

Issue #6442, opened 2022-07-27
calico-node pods are failing after upgrade from 3.22 to 3.23: felix is not ready: readiness probe reporting 503


r0bj:
## Expected Behavior

Calico is working after upgrade to version 3.23.

## Current Behavior

`calico-node` pods are failing after upgrade from 3.22 to 3.23:
```
$ kubectl describe po calico-node-pg52g
Events:
  Type     Reason     Age                    From               Message
  ----     ------     ----                   ----               -------
  Normal   Scheduled  2m35s                  default-scheduler  Successfully assigned kube-system/calico-node-pg52g to dev-k8s-worker-r1
  Normal   Pulling    2m35s                  kubelet            Pulling image "docker.io/calico/cni:v3.23.3"
  Normal   Pulled     2m31s                  kubelet            Successfully pulled image "docker.io/calico/cni:v3.23.3" in 3.507665044s
  Normal   Created    2m31s                  kubelet            Created container install-cni
  Normal   Started    2m31s                  kubelet            Started container install-cni
  Normal   Pulling    2m28s                  kubelet            Pulling image "docker.io/calico/node:v3.23.3"
  Normal   Pulled     2m24s                  kubelet            Successfully pulled image "docker.io/calico/node:v3.23.3" in 3.680538035s
  Normal   Created    2m24s                  kubelet            Created container mount-bpffs
  Normal   Started    2m24s                  kubelet            Started container mount-bpffs
  Normal   Pulled     2m23s                  kubelet            Container image "docker.io/calico/node:v3.23.3" already present on machine
  Normal   Created    2m23s                  kubelet            Created container calico-node
  Normal   Started    2m23s                  kubelet            Started container calico-node
  Warning  Unhealthy  2m21s (x2 over 2m22s)  kubelet            Readiness probe failed: calico/node is not ready: BIRD is not ready: Error querying BIRD: unable to connect to BIRDv4 socket: dial unix /var/run/calico/bird.ctl: connect: connection refused
  Warning  Unhealthy  2m15s                  kubelet            Readiness probe failed: 2022-07-27 18:18:21.875 [INFO][438] confd/health.go 180: Number of node(s) with BGP peering established = 10
calico/node is not ready: felix is not ready: readiness probe reporting 503
  Warning  Unhealthy  2m5s  kubelet  Readiness probe failed: 2022-07-27 18:18:31.882 [INFO][665] confd/health.go 180: Number of node(s) with BGP peering established = 10
[...]

PR #6498, opened 2022-08-04
Set IPIPMode and VXLANMode to the default "Never" if they are empty strings

coutinhop:
Set IPIPMode and VXLANMode to the default "Never" if they are empty strings in convertIpPoolFromStorage() in clientv3/ippool.go (for when the IP pool was created by an older Calico version that did not have these fields).

Add tests for this to clientv3/ippool_e2e_test.go.

Roll back EncapsulationCalculator changes from https://github.com/projectcalico/calico/pull/6494 and add validation to EncapsulationCalculator.handleAPIPool in encapsulation_resolver.go. Also add tests for it in encapsulation_resolver_test.go.

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/6442
undoes https://github.com/projectcalico/calico/pull/6494

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Set IPIPMode and VXLANMode to the default "Never" if they are empty strings in IPPools.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/calc/encapsulation_resolver.go
+++ b/felix/calc/encapsulation_resolver.go
@@ -197,8 +197,26 @@ func (c *EncapsulationCalculator) handleAPIPool(p model.KVPair) error {
 		return fmt.Errorf("failed to convert %+v to *model.IPPool", p.Value)
 	}
 
+	// Validate pool's IPIPMode
+	switch pool.Spec.IPIPMode {
+	case apiv3.IPIPModeNever:
+	case apiv3.IPIPModeAlways:
+	case apiv3.IPIPModeCrossSubnet:
+	default:
+		return fmt.Errorf("invalid IPIPMode \"%v\" for %v", pool.Spec.IPIPMode, pool.Spec.CIDR)
+	}
+
+	// Validate pool's VXLANMode
+	switch pool.Spec.VXLANMode {
+	case apiv3.VXLANModeNever:
+	case apiv3.VXLANModeAlways:
+	case apiv3.VXLANModeCrossSubnet:
+	default:
+		return fmt.Errorf("invalid VXLANMode \"%v\" for %v", pool.Spec.VXLANMode, pool.Spec.CIDR)
+	}
+
 	poolKey := pool.Spec.CIDR
-	c.updatePool(poolKey, pool.Spec.IPIPMode != "" && pool.Spec.IPIPMode != apiv3.IPIPModeNever, pool.Spec.VXLANMode != "" && pool.Spec.VXLANMode != apiv3.VXLANModeNever)
+	c.updatePool(poolKey, pool.Spec.IPIPMode != apiv3.IPIPModeNever, pool.Spec.VXLANMode != apiv3.VXLANModeNever)
 
 	return nil
 }
--- a/libcalico-go/lib/clientv3/ippool.go
+++ b/libcalico-go/lib/clientv3/ippool.go
@@ -272,6 +272,17 @@ func convertIpPoolFromStorage(pool *apiv3.IPPool) error {
 	if pool.Spec.NodeSelector == "" {
 		pool.Spec.NodeSelector = "all()"
 	}
+
+	// Default IPIPMode to "Never" if not set.
+	if len(pool.Spec.IPIPMode) == 0 {
+		pool.Spec.IPIPMode = apiv3.IPIPModeNever
+	}
+
+	// Default VXLANMode to "Never" if not set.
+	if len(pool.Spec.VXLANMode) == 0 {
+		pool.Spec.VXLANMode = apiv3.VXLANModeNever
+	}
+
 	return nil
 }
 
--- a/libcalico-go/lib/clientv3/ippool_e2e_test.go
+++ b/libcalico-go/lib/clientv3/ippool_e2e_test.go
@@ -18,16 +18,20 @@ import (
 	"context"
 	"time"
 
+	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 
 	"github.com/projectcalico/calico/libcalico-go/lib/apiconfig"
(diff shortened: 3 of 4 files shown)


libcalico-go/lib/clientv3/ippool_e2e_test.go
```diff
@@ -318,10 +335,34 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests", testutils.DatastoreAll, f
 			_, outError = c.IPPools().Get(ctx, name2, options.GetOptions{})
 			Expect(outError).To(HaveOccurred())
 			Expect(outError.Error()).To(ContainSubstring("resource does not exist: IPPool(" + name2 + ") with error:"))
+
+			By("Adding an IPPool with empty string for IPIPMode and VXLANMode and expecting it to be defaulted to 'Never'")
+			res, outError = c.IPPools().Create(ctx, &apiv3.IPPool{
```
caseydavenport, line 340: I don't think this is testing the code branch that you added, because we do defaulting on `Create` here: https://github.com/projectcalico/calico/blob/e20bfee71d71f288386e097bd8d2041da329753d/libcalico-go/lib/clientv3/ippool.go#L431-L435

You need to bypass that logic somehow (maybe using the backend client directly) to ensure the data that gets stored is missing the VXLAN setting, and then confirm that Get() defaults the field.

felix/calc/encapsulation_resolver.go
```diff
@@ -197,8 +197,26 @@ func (c *EncapsulationCalculator) handleAPIPool(p model.KVPair) error {
 		return fmt.Errorf("failed to convert %+v to *model.IPPool", p.Value)
 	}
 
+	// Validate pool's IPIPMode
+	switch pool.Spec.IPIPMode {
+	case apiv3.IPIPModeNever:
```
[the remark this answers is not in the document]
coutinhop: I'm aware, my intention is to do nothing on the 'valid' values, and to error on default for an 'unknown'/'invalid' value, wouldn't that be doing this?

libcalico-go/lib/clientv3/ippool_e2e_test.go
coutinhop: Oh right, let me look into that. Thanks for catching this!

felix/calc/encapsulation_resolver.go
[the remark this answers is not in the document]
caseydavenport: doh yeah, I just read this way too fast. This is correct.

libcalico-go/lib/clientv3/ippool_e2e_test.go
coutinhop: I added a new `DebugUnsafeCreateNoDefaults()` method that skips setting defaults and now the test seems to be testing properly. Thanks!

libcalico-go/lib/clientv3/ippool.go
```diff
@@ -515,6 +527,25 @@ func (r ipPools) UnsafeCreate(ctx context.Context, res *apiv3.IPPool, opts optio
 	return nil, err
 }
 
+func (r ipPools) DebugUnsafeCreateNoDefaults(ctx context.Context, res *apiv3.IPPool, opts options.SetOptions) (*apiv3.IPPool, error) {
```
caseydavenport, line 530: Hm, I'm not sure about adding a new public method to the interface just to satisfy this test.

We have other ways of writing "raw" data into the datastore - e.g., using a backend client directly. 

It's easy to get a backend client from a client - we do it in some of our other tests as well, like here: https://github.com/projectcalico/calico/blob/bc619e7ef607229f40f473c8baf60096a2eb58ab/kube-controllers/pkg/controllers/node/kdd_ipam_gc_fv_test.go#L97-L101

caseydavenport: You can then use this function directly: https://github.com/projectcalico/calico/blob/bc619e7ef607229f40f473c8baf60096a2eb58ab/libcalico-go/lib/backend/api/api.go#L62-L65

e.g.,

```go
kvp := model.KVPair{
  Key: model.ResourceKey{},
  Value: &v3.IPPool{},
}
bc.Create(ctx, kvp)
```

coutinhop: I see, yes this is much better. Will do, thanks!

libcalico-go/lib/clientv3/ippool_e2e_test.go
```diff
@@ -318,10 +339,57 @@ var _ = testutils.E2eDatastoreDescribe("IPPool tests", testutils.DatastoreAll, f
...
+			kvp := &model.KVPair{
+				Key: model.ResourceKey{
+					Name: name3,
+					Kind: apiv3.KindIPPool,
+				},
+				Value: &apiv3.IPPool{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:              name3,
+						CreationTimestamp: metav1.Now(),
+						ResourceVersion:   "1",
```
caseydavenport, line 361: You shouldn't set `ResourceVersion` on Create() requests btw

caseydavenport: Strictly speaking, I don't think you need to specify CreationTimestamp or UID either - but you can if you want.

caseydavenport (approved):

libcalico-go/lib/clientv3/ippool_e2e_test.go
coutinhop: I know I shouldn't/shouldn't need to, but `MatchResource()` on this line gives me trouble unless I set all of those:
https://github.com/projectcalico/calico/blob/c389512aba2ecf3548ae94486c66c319058ac45c/libcalico-go/lib/clientv3/ippool_e2e_test.go#L370

I could just use reflect to compare the pool's Spec, but I figured it would be good to keep the test consistent (and thus appease `MatchResource()` by including all of those). Let me know if you think that could be a problem. Thanks!

caseydavenport: Ah I see, yeah I guess that's OK. MatchResource is probably doing the wrong thing here but I think it's fine to leave it as you have it.

rastislavs:
Hey, do you plan to cherry-pick the fix to previous releases as well? We just hit it on 3.23, it looks like the fix was merged only into the master branch.

caseydavenport:
This was cherry-picked to v3.24 and v3.23 - should be present in v3.23.4+

Issue #4509, opened 2021-04-05
DNS lookups from Nginx to CoreDNS in eBPF mode fail when the CoreDNS Pods move


diranged:
Slack Thread: https://calicousers.slack.com/archives/CPEPF833L/p1614095727031700

We discovered that when Nginx is doing DNS lookups, it uses the same source-port for all of its DNS queries. When operating in eBPF mode, Calico is doing the "service-to-ip" mapping the first time a connection is created, and then keeps that data for the duration of that particular connection. Because Nginx is using the same source port each time, Calico tracks this as the same connection. If your backing pods (coredns in this case) move, their IPs are never updated as part of this connection. This causes Nginx to fail to execute any future DNS lookups, breaking Nginx in many different ways.

## Expected Behavior

If the backing pod for an already-created connection session has been gone away, I think that any caching that calico does should get wiped out so that when a new session is created (even if its from the same source port), a new mapping is created.

## Steps to Reproduce (for bugs)

See the slack thread above for more details...

1. Create a cluster running with eBPF mode.
2. Create a service "guestbook"
3. Create a service "guestbook-proxy" that is an nginx reverse proxy pointing to "guestbook.<namespace>".
    (Make sure to set Nginx up to do DNS lookups against the coredns service)
4. `kubectl rollout restart deployment/coredns -n kube-system`
5. `kubectl rollout restart deployment/guestbook`

At this point, the `guestbook-proxy` pod should be unable to lookup the backend address for the `guestbook` service. It will report DNS lookup timeouts.

## Context
We make heavy use of reverse-proxies in our environment.. not only for our own apps, but even for Loki (the "loki-gateway" pod is an Nginx reverse proxy). We've had several outages in our staging environment due to our Coredns pods moving IPs and DNS resolution breaking.

## Your Environment
Kubernetes: EKS 1.19
Calico: 3.18.2
OS: Bottlerocket v1.0.7 (pre-release)

### tcpdump of the DNS lookups that work

```IP (tos 0x0, ttl 255, id 34455, offset 0, flags [DF], proto UDP (17), length 63)
    100.64.99.171.53869 > 100.64.147.127.53: [bad udp cksum 0xbfe7 -> 0xa0a3!] 44873+ A? foo.okta.com. (35)
IP (tos 0x0, ttl 254, id 34455, offset 0, flags [DF], proto UDP (17), length 63)
[...]

PR #6527, opened 2022-08-10
felix/bpf: CTLB workaround for host and nodeports

tomastigera:
## Description   
    Packets from the host are intercepted at the first device they hit. If
    it is towards the same host, it is the loopback device, otherwise it is
    the main device on the cluster network. At that point we do the DNAT -
    select the backed pod for the nodeport. We assume that there is a
    default route so it hits a device. Otherwise we cannot do much.
    
    We either use FIB to redirect the packet to an outgoing device (if on lo
    or if it should go to a different one than the default) or we redirect
    the packet to the bpfin.cali iface to turn the packet around into the
    host address namespace that can route the packet. If an overlay is used,
    we SNAT the packet so that it has the tunnel's source and we redirect it
    to the tunnel.

## Related issues/PRs

refs https://github.com/projectcalico/calico/issues/4509

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/fib.h
+++ b/felix/bpf-gpl/fib.h
@@ -21,6 +21,17 @@
 
 static CALI_BPF_INLINE bool fib_approve(struct cali_tc_ctx *ctx, __u32 ifindex)
 {
+#ifdef UNITTEST
+	/* Let's assume that unittest is setup so that WEP's are ready - for UT simplicity */
+	return true;
+#else
+	/* If we are turnign packets around on lo to a remote pod, approve the
+	 * fib as it does not concern a possibly not ready local WEP.
+	 */
+	if (CALI_F_TO_HEP && ctx->state->flags & CALI_ST_CT_NP_REMOTE) {
+		return true;
+	}
+
 	struct cali_tc_state *state = ctx->state;
 
 	/* To avoid forwarding packets to workloads that are not yet ready, i.e
@@ -44,6 +55,7 @@ static CALI_BPF_INLINE bool fib_approve(struct cali_tc_ctx *ctx, __u32 ifindex)
 	}
 
 	return true;
+#endif
 }
 
 static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
@@ -85,11 +97,12 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
 		CALI_DEBUG("Redirect to the same interface (%d) failed.\n", ctx->skb->ifindex);
 		goto deny;
 	} else if (rc == CALI_RES_REDIR_IFINDEX) {
-		struct arp_value *arpv;
 		__u32 iface = state->ct_result.ifindex_fwd;
 
+		struct arp_value *arpv;
+
 		struct arp_key arpk = {
-			.ip = state->ip_dst,
+			.ip = iface != NATIN_IFACE ? state->ip_dst : 0 /* 0.0.0.0 */,
 			.ifindex = iface,
 		};
 
@@ -125,6 +138,11 @@ skip_redir_ifindex:
 	}
 
 #if CALI_FIB_ENABLED
+	/* Only do FIB for packets to be turned around at a HEP on HEP egress. */
+	if (CALI_F_TO_HEP && !(ctx->state->flags & CALI_ST_CT_NP_LOOP)) {
+		goto skip_fib;
+	}
+
 	// Try a short-circuit FIB lookup.
 	if (fwd_fib(&ctx->fwd)) {
 		/* Revalidate the access to the packet */
@@ -147,7 +165,7 @@ skip_redir_ifindex:
 		struct bpf_fib_lookup fib_params = {
 			.family = 2, /* AF_INET */
 			.tot_len = 0,
-			.ifindex = ctx->skb->ingress_ifindex,
+			.ifindex = CALI_F_TO_HOST ? ctx->skb->ingress_ifindex : ctx->skb->ifindex,
 			.l4_protocol = state->ip_proto,
 			.sport = bpf_htons(state->sport),
 			.dport = bpf_htons(state->dport),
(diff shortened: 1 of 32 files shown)


felix/bpf-gpl/fib.h
```diff
@@ -21,6 +21,17 @@
 
 static CALI_BPF_INLINE bool fib_approve(struct cali_tc_ctx *ctx, __u32 ifindex)
 {
+#ifdef UNITTEST
+	/* Let's assume that unittest is setup so that WEP's are ready - for UT simplicity */
+	return true;
+#else
+	/* If we are turnign packets around on lo to a remote pod, approve the
```
fasaxc, line 28:
```suggestion
	/* If we are turning packets around on lo to a remote pod, approve the
```

felix/bpf-gpl/tc.c
```diff
@@ -64,11 +64,37 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
...
 	}
 
+	if (CALI_F_NAT_IF) {
+		switch (skb->mark) {
+		case CALI_SKB_MARK_BYPASS:
+			/* We are turning a packet around to a local WEP using bpfnat
+			 * iface, the WEP should do normal processing.
+			 */
+			{
+				__u32 mark = 0;
```
fasaxc, line 79: Why the extra variable?

felix/bpf-gpl/tc.c
```diff
@@ -467,7 +492,24 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 		}
 	}
 
-	if (rt_addr_is_local_host(ctx->state->post_nat_ip_dst)) {
+	if (rt_addr_is_local_host(ctx->state->ip_src)) {
```
fasaxc, line 495: Should this be limited to program types where the packet is leaving the host?  Don't want the outbound failsafe to kick in on a (spoofed) inbound packet.

felix/bpf-gpl/tc.c
```diff
@@ -476,16 +518,31 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
...
+			if (CALI_F_LO) {
+				CALI_DEBUG("NP redir remote on LO\n");
+				ctx->state->flags |= CALI_ST_CT_NP_LOOP;
+			} else if (CALI_F_MAIN && cali_rt_is_tunneled(dest_rt)) {
+				CALI_DEBUG("NP redir remote on HEP to tunnel\n");
+				ctx->state->flags |= CALI_ST_CT_NP_LOOP;
+			}
+			ctx->state->flags |= CALI_ST_CT_NP_REMOTE;
+			ctx->state->pol_rc = CALI_POL_ALLOW;
+			/* XXX We need to do HEP policy here, where we do NAT. */
```
fasaxc, line 541: Bit confused by setting the `pol_rc = CALI_POL_ALLOW` and then this comment.  But then you fall through to `do_policy`

felix/bpf-gpl/tc.c
```diff
@@ -476,16 +518,31 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
...
-		if (CALI_F_TO_HEP && is_failsafe_out(ctx->state->ip_proto, ctx->state->post_nat_dport, ctx->state->post_nat_ip_dst)) {
-			CALI_DEBUG("Outbound failsafe port: %d. Skip policy.\n", ctx->state->post_nat_dport);
-			COUNTER_INC(ctx, CALI_REASON_ACCEPTED_BY_FAILSAFE);
+
+	if (CALI_F_TO_HEP && ctx->nat_dest && !skb_seen(ctx->skb) && !(ctx->state->flags & CALI_ST_HOST_PSNAT)) {
+		CALI_DEBUG("Host accesses nodeport backend %x:%d state->flags 0x%x\n",
+			   bpf_htonl(ctx->state->post_nat_ip_dst), ctx->state->post_nat_dport,
+			   ctx->state->flags);
+		if (cali_rt_flags_local_workload(dest_rt->flags)) {
+			CALI_DEBUG("NP redir on HEP - skip policy\n");
```
fasaxc, line 527: Is this the first program to see this traffic?  I think "wildcard" HEP egress policy should apply from host to local workload even if going via a nodeport.

felix/bpf-gpl/tc.c
```diff
@@ -1226,6 +1292,33 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct cali_tc_ctx *ctx
 		fib = false; /* Disable FIB because we want to drop to iptables */
 	}
 
+	if (CALI_F_TO_HEP && !skb_seen(skb)) {
+		CALI_DEBUG("Host accesses nodeport backend %x:%d\n",
```
fasaxc, line 1296: Not sure I follow; looks like there are several codepaths that lead here (via `goto allow`) and not all are nodeports?

felix/fv/bpf_test.go
```diff
@@ -2493,43 +2494,116 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+										},
+									},
 								}
 								w00Slector := fmt.Sprintf("name=='%s'", w[0][0].Name)
 								pol.Spec.Selector = w00Slector
 
 								pol = createPolicy(pol)
 							})
 
-							if testOpts.connTimeEnabled {
```
fasaxc, line 2503: :tada:

fasaxc (changes requested):
Few initial questions, made me wonder if I was missing something so thought it best to discuss before I do more digging

felix/bpf/libbpf/libbpf.go
```diff
@@ -307,6 +307,7 @@ func TcSetGlobals(
 	hostTunnelIP uint32,
 	flags uint32,
 	wgPort uint16,
+	natin, natout uint32,
```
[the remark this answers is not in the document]
tomastigera: yeah, it is annoying to add the args :+1:

felix/bpf-gpl/tc.c
fasaxc: Ah OK, there's a CALI_F_TO_HEP condition inside this but it doesn't cover the `ctx->state->flags |= CALI_ST_SRC_IS_HOST;` line.

felix/bpf-gpl/tc.c
fasaxc: I was wrong, wildcard HEPs don't affect host to own pod (and we have FVs that assert that)

felix/bpf-gpl/tc.c
```diff
@@ -1226,6 +1291,39 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct cali_tc_ctx *ctx
...
+					state->ct_result.ifindex_fwd = NATIN_IFACE  ;
+					CALI_DEBUG("NP remote WL %x:%d on LO or main HEP\n",
+						bpf_htonl(state->post_nat_ip_dst), state->post_nat_dport);
+					ctx->state->flags |= CALI_ST_CT_NP_LOOP;
+				}
+				ctx->state->flags |= CALI_ST_CT_NP_REMOTE;
+				fib = true; /* Enforce FIB since we want to redirect */
+			}
+		} else {
+			/* Did not find a route for a service BE? Should be just
```
fasaxc, line 1315: What about out-of-cluster service backends such as the k8s API server?  I don't think they'll have routes

fasaxc (changes requested):
Spotted a couple of points on second read through.  I'm still wary of accept_local; could we get rid of that if we assume bpf_redirect_neigh?

tomastigera:
> Spotted a couple of points on second read through. I'm still wary of accept_local; could we get rid of that if we assume bpf_redirect_neigh?

Not sure as if a pod is responding to local host it needs to revers DNAT and the resulting packet may have the local host's source if it was accessing the pod via local nodeport :sweat_smile:  We may be able to deliver straight to a socket but I did not explore that concept yet.

fasaxc (approved):

PR #6566, opened 2022-08-16
Kube controller: Making initialization retry

tmjd:
It has been observed the kube-controller logs reporting
cmdwrapper.go 95: Received exit status 129, restarting
which is a restart due to a configuration change and then
the new kube-controller process failing to initialize (access) the
datastore causing a pod restart. This change extends the timeout to give
more time.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Retry kube-controllers initialization on failure
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/kube-controllers/cmd/kube-controllers/main.go
+++ b/kube-controllers/cmd/kube-controllers/main.go
@@ -121,13 +121,38 @@ func main() {
 	// Create the context.
 	ctx, cancel := context.WithCancel(context.Background())
 
+	// Create the status file. We will only update it if we have healthchecks enabled.
+	s := status.New(statusFile)
+
 	log.Info("Ensuring Calico datastore is initialized")
-	initCtx, cancelInit := context.WithTimeout(ctx, 10*time.Second)
-	defer cancelInit()
-	err = calicoClient.EnsureInitialized(initCtx, "", "k8s")
-	if err != nil {
-		log.WithError(err).Fatal("Failed to initialize Calico datastore")
+	s.SetReady("Startup", false, "initialized to false")
+	initCtx, cancelInit := context.WithTimeout(ctx, 60*time.Second)
+	// Loop until context expires or calicoClient is initialized.
+	for {
+		err := calicoClient.EnsureInitialized(initCtx, "", "k8s")
+		if err != nil {
+			log.WithError(err).Info("Failed to initialize datastore")
+			s.SetReady(
+				"Startup",
+				false,
+				fmt.Sprintf("Error initializing datastore: %v", err),
+			)
+		} else {
+			// Exit loop
+			break
+		}
+
+		select {
+		case <-initCtx.Done():
+			log.Fatal("Failed to initialize Calico datastore")
+			break
+		case <-time.After(5 * time.Second):
+			// Try to initialize again
+		}
 	}
+	log.Info("Calico datastore is initialized")
+	s.SetReady("Startup", true, "")
+	cancelInit()
 
 	controllerCtrl := &controllerControl{
 		ctx:         ctx,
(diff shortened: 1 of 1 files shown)


kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -121,14 +121,38 @@ func main() {
...
 	ctx, cancel := context.WithCancel(context.Background())
 
+	// Create the status file. We will only update it if we have healthchecks enabled.
+	s := status.New(statusFile)
+
 	log.Info("Ensuring Calico datastore is initialized")
-	initCtx, cancelInit := context.WithTimeout(ctx, 10*time.Second)
-	defer cancelInit()
-	err = calicoClient.EnsureInitialized(initCtx, "", "k8s")
 	if err != nil {
```
song-jiang, line 128: `if err != nil` should be removed as well?

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -121,14 +121,38 @@ func main() {
...
 		log.WithError(err).Fatal("Failed to initialize Calico datastore")
 	}
 
+	s.SetReady("StartupCalicoDatastoreInitialization", false, "initialized to false")
+	initCtx, cancelInit := context.WithTimeout(ctx, 60*time.Second)
+	// Loop until context expires or calicoClient is initialized.
```
song-jiang, line 134: We should call `defer cancelInit()` here rather than calling it directly later after the loop?

song-jiang (commented):
Just two comments but looks good.

kube-controllers/cmd/kube-controllers/main.go
tmjd: If i defer it then it won't happen until main returns. Calling it after the loop means the context will be canceled as soon as we're done with it. The only exit from the loop is to break (which means the cancelInit will be called) or a "fatal" exit which means the process exits in which case we don't care if we call the cancel.

song-jiang: Right. That's fine with me.

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -121,14 +121,38 @@ func main() {
...
+	for {
+		select {
+		case <-initCtx.Done():
+			log.Fatal("Failed to initialize Calico datastore")
+			break
+		default:
```
caseydavenport, line 140: Very minor, but instead of a `default` you could do something like this:

```go
select {
  case <- ctx.Done():
    break
  case <- time.After(5*time.Second):
    // Try to initialize, break if successful. 
}
```

I find that to be slightly more idiomatic / easy to read.

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -121,14 +121,38 @@ func main() {
...
+
 	log.Info("Ensuring Calico datastore is initialized")
-	initCtx, cancelInit := context.WithTimeout(ctx, 10*time.Second)
-	defer cancelInit()
-	err = calicoClient.EnsureInitialized(initCtx, "", "k8s")
 	if err != nil {
 		log.WithError(err).Fatal("Failed to initialize Calico datastore")
 	}
 
+	s.SetReady("StartupCalicoDatastoreInitialization", false, "initialized to false")
```
caseydavenport, line 132: can probably just call this "Startup" and then if there are future startup tasks we can clump them together.

caseydavenport (commented):
LGTM, will let @song-jiang have the final say though

song-jiang (approved):
LGTM

PR #6574, opened 2022-08-18
BPF Policy program for IPv6

mazdakn:
## Description
Add support for IPv6 to the policy program

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/jump.h
+++ b/felix/bpf-gpl/jump.h
@@ -8,7 +8,7 @@
 #include "conntrack.h"
 #include "policy.h"
 
-CALI_MAP(cali_v5_state, 5,
+CALI_MAP(cali_state, 2,
 		BPF_MAP_TYPE_PERCPU_ARRAY,
 		__u32, struct cali_tc_state,
 		1, 0, MAP_PIN_GLOBAL)
@@ -16,7 +16,7 @@ CALI_MAP(cali_v5_state, 5,
 static CALI_BPF_INLINE struct cali_tc_state *state_get(void)
 {
 	__u32 key = 0;
-	return cali_v5_state_lookup_elem(&key);
+	return cali_state_lookup_elem(&key);
 }
 
 struct bpf_map_def_extended __attribute__((section("maps"))) cali_jump2 = {
--- a/felix/bpf-gpl/jump.h
+++ b/felix/bpf-gpl/jump.h
@@ -8,7 +8,7 @@
 #include "conntrack.h"
 #include "policy.h"
 
-CALI_MAP(cali_v4_state, 4,
+CALI_MAP(cali_v5_state, 5,
 		BPF_MAP_TYPE_PERCPU_ARRAY,
 		__u32, struct cali_tc_state,
 		1, 0, MAP_PIN_GLOBAL)
@@ -16,7 +16,7 @@ CALI_MAP(cali_v4_state, 4,
 static CALI_BPF_INLINE struct cali_tc_state *state_get(void)
 {
 	__u32 key = 0;
-	return cali_v4_state_lookup_elem(&key);
+	return cali_v5_state_lookup_elem(&key);
 }
 
 struct bpf_map_def_extended __attribute__((section("maps"))) cali_jump2 = {
--- a/felix/bpf/polprog/pol_prog_builder.go
+++ b/felix/bpf/polprog/pol_prog_builder.go
@@ -45,6 +45,7 @@ type Builder struct {
 	stateMapFD         bpf.MapFD
 	jumpMapFD          bpf.MapFD
 	policyDebugEnabled bool
+	forIPv6            bool
 }
 
 type ipSetIDProvider interface {
@@ -58,6 +59,7 @@ func NewBuilder(ipSetIDProvider ipSetIDProvider, ipsetMapFD, stateMapFD, jumpMap
 		stateMapFD:         stateMapFD,
 		jumpMapFD:          jumpMapFD,
 		policyDebugEnabled: policyDebugEnabled,
+		forIPv6:            false,
 	}
 	return b
 }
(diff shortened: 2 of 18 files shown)


felix/bpf-gpl/jump.h
```diff
@@ -8,15 +8,15 @@
 #include "conntrack.h"
 #include "policy.h"
 
-CALI_MAP(cali_v4_state, 4,
+CALI_MAP(cali_v5_state, 5,
```
tomastigera, line 11: the v4 was for IPv4 ;-) If we are going to reuse it, we may just remove the v4 and start from version 2 (because 1 is special for historical reasons). We may reconsider in the future whether it was a good idea to merge these.

felix/bpf/polprog/pol_prog_builder.go
```diff
@@ -58,6 +59,7 @@ func NewBuilder(ipSetIDProvider ipSetIDProvider, ipsetMapFD, stateMapFD, jumpMap
 		stateMapFD:         stateMapFD,
 		jumpMapFD:          jumpMapFD,
 		policyDebugEnabled: policyDebugEnabled,
+		forIPv6:            false,
```
tomastigera, line 62: this is the default value, no need to be explicit :)

felix/bpf/polprog/pol_prog_builder.go
```diff
@@ -317,7 +312,11 @@ func (p *Builder) writeProgramFooter(forXDP bool) {
 	// Execute the tail call to drop program
 	p.b.Mov64(R1, R6)                      // First arg is the context.
 	p.b.LoadMapFD(R2, uint32(p.jumpMapFD)) // Second arg is the map.
-	p.b.MovImm32(R3, jumpIdxDrop)          // Third arg is the index (rather than a pointer to the index).
+	jumpDropIndex := bpf.ProgIndexDrop
+	if p.forIPv6 {
+		jumpDropIndex = bpf.ProgIndexV6Drop
+	}
+	p.b.MovImm32(R3, int32(jumpDropIndex)) // Third arg is the index (rather than a pointer to the index).
```
tomastigera, line 319: Pehaps hide these ifs in methods?

```suggestion
	p.b.MovImm32(R3, p.jumpDropIndex()) // Third arg is the index (rather than a pointer to the index).
```

felix/bpf/polprog/pol_prog_builder.go
```diff
@@ -535,7 +538,12 @@ func (p *Builder) writeRule(r Rule, actionLabel string, destLeg matchLeg) {
 		log.Panic("empty action label")
 	}
 
-	rule := rules.FilterRuleToIPVersion(4, r.Rule)
+	ipVersion := proto.IPVersion_IPV4
```
tomastigera, line 541: ditto
```suggestion
	ipVersion := p.ipVersion()
```

felix/bpf/polprog/pol_prog_builder.go
```diff
@@ -738,25 +746,78 @@ func (p *Builder) writeCIDRSMatch(negate bool, leg matchLeg, cidrs []string) {
...
+		} else { // IPv4
+			addrU32[0] = bits.ReverseBytes32(cidr.Addr().(ip.V4Addr).AsUint32())
+			maskU32[0] = bits.ReverseBytes32(math.MaxUint32 << (32 - cidr.Prefix()) & math.MaxUint32)
+		}
+
+		lastAddr := addrU32[0]
+		for section, addr := range addrU32 {
+			// Optimisation: If mask for this section, i.e. this match, is 0,
+			// then we can skip the match since the result of AND operation is
+			// irrelevent of packet address. However, we need to check at least one 32bit section.
```
tomastigera, line 796:
```suggestion
			// irrelevant of packet address. However, we need to check at least one 32bit section.
```

felix/bpf/state/map.go
```diff
@@ -98,13 +132,13 @@ func StateFromBytes(bytes []byte) State {
...
-	Filename:   "/sys/fs/bpf/tc/globals/cali_v4_state",
+	Filename:   "/sys/fs/bpf/tc/globals/cali_v5_state",
 	Type:       "percpu_array",
 	KeySize:    4,
 	ValueSize:  expectedSize,
 	MaxEntries: 1,
-	Name:       "cali_v4_state",
-	Version:    4,
+	Name:       "cali_v5_state",
+	Version:    5,
```
tomastigera, line 141: the point was that only the version would change ;-)

felix/bpf/xdp/attach.go
```diff
@@ -73,7 +73,7 @@ func (ap AttachPoint) ProgramName() string {
 	return "xdp_calico_entry"
 }
 
-func (ap *AttachPoint) Log() *log.Entry {
+func (ap AttachPoint) Log() *log.Entry {
```
tomastigera, line 76: Is it better to remove the '*' or to add it? If we remove it, we keep copying the structure on each call  where we pass it. I am not saying one way or another, just reminding you that it makes a difference.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1851,42 +1851,71 @@ func (m *bpfEndpointManager) writePolicyDebugInfo(insns asm.Insns, ifaceName, po
...
+	ipVersions := []proto.IPVersion{proto.IPVersion_IPV4}
+	if m.ipv6Enabled {
+		ipVersions = append(ipVersions, proto.IPVersion_IPV6)
 	}
-	defer func() {
-		// Once we've put the program in the map, we don't need its FD any more.
-		err := progFD.Close()
+
+	for _, ipFamily := range ipVersions {
+		pg := polprog.NewBuilder(m.ipSetIDAlloc, m.bpfMapContext.IpsetsMap.MapFD(), m.bpfMapContext.StateMap.MapFD(), jumpMapFD, m.bpfPolicyDebugEnabled)
```
tomastigera, line 1861: perhaps move the bodyinto a method that can be parametrized by the family?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1851,42 +1851,71 @@ func (m *bpfEndpointManager) writePolicyDebugInfo(insns asm.Insns, ifaceName, po
...
-	if err != nil {
-		return fmt.Errorf("failed to update jump map: %w", err)
+func (m *bpfEndpointManager) removePolicyProgram(jumpMapFD bpf.MapFD, ap attachPoint) error {
+	ipVersions := []proto.IPVersion{proto.IPVersion_IPV4}
+	if m.ipv6Enabled {
+		ipVersions = append(ipVersions, proto.IPVersion_IPV6)
+	}
+
+	for _, ipFamily := range ipVersions {
+		k := make([]byte, 4)
```
tomastigera, line 1913: ditto

felix/bpf/ut/pol_prog_test.go
```diff
@@ -225,10 +222,19 @@ func icmpPktWithTypeCode(src, dst string, icmpType, icmpCode int) packet {
 }
 
 func packetNoPorts(proto int, src, dst string) packet {
+	// Just using ResolveUDPAddr to parse a string to ip and port
+	srcAddr := net.ParseIP(src)
+	if srcAddr == nil {
+		panic(fmt.Errorf("failed to parse src addr %v", src))
```
tomastigera, line 228: Should we rather use like `Expected` in test rather then if then panic?

felix/bpf/ut/pol_prog_test.go
```diff
@@ -300,6 +338,37 @@ var polProgramTests = []polProgramTest{
 			tcpPkt("10.0.0.2:80", "10.0.0.1:31245"),
 			icmpPkt("10.0.0.1", "10.0.0.2")},
 	},
+	{
+		PolicyName: "unreachable tier-v6",
```
tomastigera, line 342: I was wondering if it was just possible to remove the hardcoded strings, replace them with variables and parametrize the tests with two sets of IPs so that we would have the same set for both 4/6 and when we add a test, we would automatically add that test for both :thinking:

felix/bpf/ut/pol_prog_test.go
```diff
@@ -613,9 +952,166 @@ var polProgramTests = []polProgramTest{
 			tcpPkt("10.0.0.2:80", "10.0.0.1:31245"),
 			udpPkt("10.0.0.2:80", "10.0.0.1:31245")},
 	},
+	{
```
tomastigera, line 955: The tests below are v6 specific tests or is it for mized 4 and 6?

tomastigera (commented):
This is great progress! I have some questions and suggestions, but nothing major.

felix/bpf/ut/pol_prog_test.go
mazdakn: Ideally, but that need changes which prevents using `packetNoPorts` in the list of test cases as it is now.

felix/bpf/ut/pol_prog_test.go
mazdakn: It's possible, but I think in the end we would need to specify both IPv4/6 rules and packets separately, which does not shorten the test cases. Also for IPv6 CIDR tests I had to add more tests to cover situation about mask length greater and shorter than 64 which is not valid for IPv4. Do you think I should try it?

felix/bpf/ut/pol_prog_test.go
mazdakn: mixed IPv4 and 6. I noticed there is no test about /0 in IPv4 so I added it.

tomastigera (approved):
LGTM

PR #6576, opened 2022-08-18
Fixing broken nested tab structure.

ctauchen:
This page has a nested tab structure that doesn't work. I've removed the page-wide tab and replaced that information with a note. Perhaps not the best way, but it allows the other information to display properly.

See what happens when you click the Operator tab at [Using host-local IPAM](https://projectcalico.docs.tigera.io/reference/cni-plugin/configuration#using-host-local-ipam)

Preview: https://deploy-preview-6576--calico-master.netlify.app/reference/cni-plugin/configuration

CC @alexeymagdich-tigera

### DIFF
--- a/calico/reference/cni-plugin/configuration.md
+++ b/calico/reference/cni-plugin/configuration.md
@@ -4,18 +4,9 @@ description: Details for configuring the Calico CNI plugins.
 canonical_url: '/reference/cni-plugin/configuration'
 ---
 
-{% tabs %}
-  <label:Operator,active:true>
-<%
-
-The {{site.prodname}} CNI plugins do not need to be configured directly when installed by the operator. For a complete operator 
-configuration reference, see [the installation API reference documentation][installation].
-
-%>
-
-  <label:Manifest>
-<%
-
+>**Note**: The {{site.prodname}} CNI plugins do not need to be configured directly when installed by the operator.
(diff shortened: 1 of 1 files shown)


ctauchen:
@caseydavenport Does this work for you?

calico/reference/cni-plugin/configuration.md
```diff
@@ -4,18 +4,9 @@ description: Details for configuring the Calico CNI plugins.
...
-configuration reference, see [the installation API reference documentation][installation].
-
-%>
-
-  <label:Manifest>
-<%
-
+>**Note**: The {{site.prodname}} CNI plugins do not need to be configured directly when installed by the operator.
+>For a complete operator configuration reference, see [the installation API reference documentation][installation].
+{: .alert .alert-info}
```
coutinhop, line 9: Should we consider using `.alert-danger` to make sure the users don't just scroll past this and inadvertently break things? Especially since operator is the default... @caseydavenport @mgleung what do you think?

ctauchen:
bump @caseydavenport

calico/reference/cni-plugin/configuration.md
caseydavenport: I think it would be better, but since we want to restructure soon anyway I'm not too bothered. Let's get this in and then I might just restructure.

caseydavenport (approved):

PR #6586, opened 2022-08-19
Fix felix RPM build

mgleung:
## Description

With the addition of https://github.com/projectcalico/calico/commit/283bb4b0378b84a45c896469e925698a6e445d55 in order to speed up our test builds, our RPM build script became unable to handle an unexpected `bin/bpf` directory. This specifies the correct felix binary for our RPM builds.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/rpm/felix.spec.in
+++ b/felix/rpm/felix.spec.in
@@ -90,7 +90,7 @@ fi
 %install
 rm -rf $RPM_BUILD_ROOT
 install -d $RPM_BUILD_ROOT/usr/bin/
-install -m 755 bin/* $RPM_BUILD_ROOT/usr/bin/
+install -m 755 bin/calico-felix $RPM_BUILD_ROOT/usr/bin/
 install -d $RPM_BUILD_ROOT/usr/lib/calico/bpf/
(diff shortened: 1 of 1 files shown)


neiljerram (approved):
Looks good, but is it possible to double check an older build, or an old RPM contents, to check that there wasn't anything else from `bin/*` other than `bin/calico-felix`?  (Or perhaps you have already done that?)

caseydavenport (commented):
LGTM but might be worth a look from @neiljerram

caseydavenport (approved):

mgleung:
@neiljerram , good point, I took a look at an older build and the only thing that I found was `calico-felix` so I think this should capture everything we need. My hunch is that we were using `*` in case the binary name changed, but I don't think it does so we should be fine.

neiljerram:
>  I took a look at an older build and the only thing that I found was calico-felix so I think this should capture everything we need. My hunch is that we were using * in case the binary name changed, but I don't think it does so we should be fine.

@mgleung Sounds good, thanks for checking that!

PR #6593, opened 2022-08-22
revise doc of calico-network-policy

Congrool:
## Description

documentation

According to the context, I think the `action` field of this example should be "Allow" instead of "Deny".

### DIFF
--- a/calico/security/calico-network-policy.md
+++ b/calico/security/calico-network-policy.md
@@ -212,7 +212,7 @@ spec:
   types:
     - Egress
   egress:    
-    - action: Deny
+    - action: Allow
       destination:
         nets:
         - 1.2.3.0/24


calico/security/calico-network-policy.md
```diff
@@ -212,7 +212,7 @@ spec:
   types:
     - Egress
   egress:    
-    - action: Deny
+    - action: Allow
```
caseydavenport, line 215: > In the following example, outgoing traffic is allowed from pods with the label **color: red** if it goes to an IP address in the **1.2.3.4/24** CIDR block. 

The existing Allow rule is actually correct - the example is meant to show how to allow traffic to a specific CIDR, not deny it.

caseydavenport (approved):

PR #6598, opened 2022-08-23
Include policy counters in policy dump

sridhartigera:
## Description

This PR has the changes to dump policy counters.
1. Write the rule matchID into the json
2. when reading the json, find the match ID, get the dump of the policyCounters bpfmap and get the count value of the match ID.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/cmd/calico-bpf/commands/policy_debug.go
+++ b/felix/cmd/calico-bpf/commands/policy_debug.go
@@ -58,8 +58,15 @@ var policyDumpCmd = &cobra.Command{
 			hooks = append(hooks, hook)
 		}
 
+		rmap := counters.PolicyMap(&bpf.MapContext{})
+		m, err := counters.LoadPolicyMap(rmap)
+		if err != nil {
+			log.WithError(err).Error("error loading rule counters map.")
+			return
+		}
+
 		for _, dir := range hooks {
-			err := dumpPolicyInfo(cmd, iface, dir)
+			err := dumpPolicyInfo(cmd, iface, dir, m)
 			if err != nil {
 				log.WithError(err).Error("Failed to dump policy info.")
 			}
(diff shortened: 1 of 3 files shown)


felix/cmd/calico-bpf/commands/policy_debug.go
```diff
@@ -87,6 +89,32 @@ func printInsn(cmd *cobra.Command, insn asm.Insn) {
...
+	return id, nil
+}
+
+func getRuleCounter(comment string) (uint64, error) {
+	matchId, err := getRuleMatchID(comment)
+	if err != nil {
+		return 0, err
+	}
+	rmap := counters.PolicyMap(&bpf.MapContext{})
+	m, err := counters.LoadPolicyMap(rmap)
```
tomastigera, line 108: Do we need to load the policy map every time we query a single ID? I wonder if it would be just good to read the map once and then just query it. I do not think we are really aiming at the most recent accuracy, we want to get a rough approximation. Otherwise we would need to stop the dp from processing packets anyway.

tomastigera: Or we can just always querry the map directly without loading it into memory, but that may be trickier as it is per-cpu map :thinking:

sridhartigera: Querying the map will be tricky.

felix/cmd/calico-bpf/commands/policy_debug.go
```diff
@@ -134,17 +126,22 @@ func dumpPolicyInfo(cmd *cobra.Command, iface, hook string) error {
...
 		for _, comment := range insn.Comments {
 			if strings.Contains(comment, "Rule MatchID") {
-				count, err := getRuleCounter(comment)
+				count = 0
+				matchId, err := getRuleMatchID(comment)
 				if err == nil {
-					cmd.Printf("// count = %d\n", count)
+					if val, ok := m[matchId]; ok {
+						count = val
+					}
```
tomastigera, line 142: what if it is not OK? Perhaps make a note if we found a counter that is not in the map. That could happen if the policy just changed and it got removed I think. Or not? Anyway, might be good not to be silent about this.

tomastigera: Or you can just do `count = m[matchID]` and it would default to zero ...

tomastigera (commented):
LGTM, just one nit/question

tomastigera (approved):

PR #6610, opened 2022-08-24
Adjust how kube-controllers timeout works

caseydavenport:
## Description

- Remove 2 second timeout for healthz call 
- Replace with exponential backoff timeout using a context
- Retry sooner if we encounter an error

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Use exponential backoff for kube-controllers health check timeout, retry sooner if failed.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/charts/tigera-operator/README.md
+++ b/charts/tigera-operator/README.md
@@ -88,7 +88,7 @@ The default values.yaml should be suitable for most basic deployments.
 
 ```
 # Image pull secrets to provision for pulling images from private registries.
-# If provided, references to the secrets must also be provided in the installation section.
+# This field is a map of desired Secret name to .dockerconfigjson formatted data to use for the secret.
 imagePullSecrets: {}
 
 # Configures general installation parameters for Calico. Schema is based
--- a/kube-controllers/cmd/kube-controllers/main.go
+++ b/kube-controllers/cmd/kube-controllers/main.go
@@ -252,6 +252,11 @@ func runHealthChecks(ctx context.Context, s *status.Status, k8sClientset *kubern
 	s.SetReady("CalicoDatastore", false, "initialized to false")
 	s.SetReady("KubeAPIServer", false, "initialized to false")
 
+	// Initialize a timeout for API calls. Start with a short timeout. We'll increase the timeout if we start seeing errors.
+	defaultTimeout := 4 * time.Second
+	maxTimeout := 16 * time.Second
+	timeout := defaultTimeout
+
 	// Loop until context expires and perform healthchecks.
 	for {
 		select {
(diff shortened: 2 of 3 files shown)


charts/tigera-operator/README.md
```diff
@@ -88,7 +88,7 @@ The default values.yaml should be suitable for most basic deployments.
 
 ```
 # Image pull secrets to provision for pulling images from private registries.
-# If provided, references to the secrets must also be provided in the installation section.
```
caseydavenport, line 91: Whoops, this is unrelated but still good to merge :sweat_smile:

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -311,6 +300,19 @@ func runHealthChecks(ctx context.Context, s *status.Status, k8sClientset *kubern
...
 		}
 
+		// If we encountered errors, retry again with a longer timeout.
+		if !s.GetReadiness() {
+			timeout = 2 * timeout
+			if timeout > maxTimeout {
+				timeout = maxTimeout
+			}
+			log.Info("Health check is not ready, retrying in 2 seconds with new timeout: %s", timeout)
+			time.Sleep(2 * time.Second)
```
Josh-Tigera, line 310: Should this be 
```suggestion
			time.Sleep(timeout * time.Second)
```
?

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -311,6 +300,19 @@ func runHealthChecks(ctx context.Context, s *status.Status, k8sClientset *kubern
...
+			continue
+		}
+
+		// Success! Reset the timeout.
+		timeout = defaultTimeout
 		time.Sleep(10 * time.Second)
```
Josh-Tigera, line 316: Is this 10s sleep still needed alongside the exponential backoff? Seems redundant.

caseydavenport: There are two time durations at play here:

- The timeout we use for our HTTP requests
- The period at which we perform the HTTP requests.

The exponential backoff is only for the first of those two. This 10s sleep is the time we wait in between successful checks (i.e., after 10s we double check that things are still OK).

The exponential backoff added in this PR is only used if we fail a health check, so that on the next iteration we give our HTTP requests a bit longer to execute.

kube-controllers/cmd/kube-controllers/main.go
caseydavenport: Nope, the timeout is for the HTTP calls (used in the context) whereas this is just to prevent tight looping.

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -261,9 +266,10 @@ func runHealthChecks(ctx context.Context, s *status.Status, k8sClientset *kubern
 			// carry on
 		}
 
-		// skip healthchecks if configured
-		// Datastore HealthCheck
-		healthCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
caseydavenport, line 266: Maybe we should have two different timeouts - EnsureInitialized can potentially make a few HTTP calls under the hood.

kube-controllers/cmd/kube-controllers/main.go
Josh-Tigera: Ah, the timeout-ception had me confused, thanks for clarification.

Josh-Tigera (approved):

mgleung (approved):

PR #6630, opened 2022-08-26
Parse ipv6 extensions

mazdakn:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/parsing.h
+++ b/felix/bpf-gpl/parsing.h
@@ -10,6 +10,40 @@
 #define PARSING_ALLOW_WITHOUT_ENFORCING_POLICY 2
 #define PARSING_ERROR -1
 
+#define MAX_EXTENSIONS 2
+
+static CALI_BPF_INLINE int parse_ipv6_extensions(struct cali_tc_ctx *ctx) {
+	__u8 next_header = ipv6_hdr(ctx)->nexthdr;
+	struct ipv6_opt_hdr ipv6_ext;
+
+	for (int i = 0; i < MAX_EXTENSIONS; i++) {
+		switch (next_header) {
+		case IPPROTO_HOPOPTS:
+		// Must be the first option, but at the moment, we only want to find
+		// transport layer, so we don't care about ordering.
+		case IPPROTO_ROUTING:
+		case IPPROTO_DSTOPTS:
+		case IPPROTO_AH:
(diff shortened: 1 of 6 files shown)


felix/bpf-gpl/parsing.h
```diff
@@ -129,11 +167,12 @@ static CALI_BPF_INLINE int tc_state_fill_from_nexthdr(struct cali_tc_ctx *ctx)
 	switch (ctx->state->ip_proto) {
 	case IPPROTO_TCP:
 		// Re-check buffer space for TCP (has larger headers than UDP).
-		if (skb_refresh_validate_ptrs(ctx, TCP_SIZE)) {
+		if (validate_ptrs(ctx, TCP_SIZE)) {
```
tomastigera, line 170: This does not pull extra data if the packet is long enough but there is not enough for direct access. Perhaps that is the problem :thinking:

felix/bpf-gpl/parsing.h
```diff
@@ -10,6 +10,44 @@
...
+			}
+
+			if (next_header == IPPROTO_AH)
+				header_length += (ipv6_ext.hdrlen * 4 + 8);
+			else
+				header_length += (ipv6_ext.hdrlen * 8 + 8);
+
+			next_header = ipv6_ext.nexthdr;
+			break;
+		default:
```
tomastigera, line 42: How is the end of the header signalled? There is some sort of sentinel, right? This default would cut off at the first header we do not know. Right?

felix/bpf-gpl/parsing.h
mazdakn: Exactly, that's the issue I am trying to solve. If I use `skb_refresh_validate_ptrs` the verifier again complains and reject the program. It seems the issue is around `skb_refresh_start_end(ctx);`. But I have not figured it out yet.

felix/bpf-gpl/parsing.h
mazdakn: The switch statement just checks for the IPv6 extensions, if the value is something other than the one we check, the code hits the default path, and exits. This function only calculated the length of IPv6 extension headers, and the rest will be handled in `tc_state_fill_from_nexthdr`. If the next header is one of `tcp`, `udp` or `icmp`, then we process it, otherwise we drop the protocols we don't recognise. Does it make sense?

PR #6633, opened 2022-08-28
fix some typos

cuishuang:


node/tests/k8st/infra/apiserver.yaml
```diff
@@ -1,4 +1,4 @@
-# This is a tech-preview manifest which installs the Calico API server. Note that this manfiest is liable to change
```
mgleung, line 1: I think this is supposed to be `manifest`

node/tests/k8st/infra/apiserver.yaml
```diff
@@ -1,4 +1,4 @@
...
+# This is a tech-preview manifest which installs the Calico API server. Note that this manliest is liable to change
```
mgleung, line 1: I think this is supposed to be `manifest`

lwr20 (approved):

Issue #5828, opened 2022-03-31
Pods can't be killed normally after Node restart


vitaliy-leschenko:
**Describe the bug**
Command `kubectl delete pods/%podname%` stucks on deleting pods.
When I try to describe the pods, I see the message:
```
Events:
  Type     Reason          Age               From               Message
  ----     ------          ----              ----               -------
  Normal   Scheduled       17m               default-scheduler  Successfully assigned default/demo-85d57c4449-f5wgk to k8s-windows-2
  Normal   Pulling         17m               kubelet            Pulling image "mcr.microsoft.com/windows/nanoserver:ltsc2022"
  Normal   Pulled          15m               kubelet            Successfully pulled image "mcr.microsoft.com/windows/nanoserver:ltsc2022" in 1m30.8318989s
  Normal   Created         15m               kubelet            Created container demo-server
  Normal   Started         15m               kubelet            Started container demo-server
  Normal   SandboxChanged  3m36s             kubelet            Pod sandbox changed, it will be killed and re-created.
  Normal   Pulled          3m32s             kubelet            Container image "mcr.microsoft.com/windows/nanoserver:ltsc2022" already present on machine
  Normal   Created         3m32s             kubelet            Created container demo-server
  Normal   Started         3m30s             kubelet            Started container demo-server
  Normal   Killing         68s               kubelet            Stopping container demo-server
  Warning  FailedKillPod   3s (x8 over 68s)  kubelet            error killing pod: failed to "KillPodSandbox" for "4723efd3-cb10-4c0c-8b45-c8f7ab981155" with KillPodSandboxError: "rpc error: code = Unknown desc = failed to remove network namespace for sandbox \"d8be8cd9b8acfb344ddd598358e8464fc9580121c576af2410235594d4210d89\": hcnDeleteNamespace failed in Win32: The specified request is unsupported. (0x803b0015) {\"Success\":false,\"Error\":\"The specified request is unsupported. \",\"ErrorCode\":2151350293}"
```

**To Reproduce**
1. Deploy new k8s cluster with CRI contaierD and CNI Calico
2. Deploy any pod on windows node
3. Restart windows node
4. Try to delete the pod on windows node

**Expected behavior**
Pods delete without any events.

**Kubernetes (please complete the following information):**
 - Windows Server version: 2022
 - Kubernetes Version: 1.22.2
 - CNI: Calico
 - CRI: containerD 1.6.0

[...]

PR #6656, opened 2022-09-02
fix: pods lost ip addreses after windows node reboot.

vitaliy-leschenko:
## Description

Currently pods (on Windows nodes) lost network adapters after node reboot. 
It happens because after node reboot kubelet starts when calico is not ready yet and restart pods using network adapters that will be deleted during CalicoNode startup.

Health checks don't solve the issue because network adapter doesn't recreated in this case. We can't delete this pods normally (without --force key).

This PR change startup of kubelet ensure that it starts after the calico network initialized.

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/5828

## Todos

- [x] Release note
- [x] Rebase
- [ ] Squash commits

## Release Note

```release-note
windows: ensure calico-managed kubelet starts after the calico network has been initialized
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/windows-packaging/CalicoWindows/kubernetes/kubelet-service.ps1
+++ b/node/windows-packaging/CalicoWindows/kubernetes/kubelet-service.ps1
@@ -22,6 +22,8 @@ $baseDir = "$PSScriptRoot\.."
 ipmo $baseDir\libs\calico\calico.psm1 -Force
 
 Write-Host "Running kubelet service."
+Wait-ForCalicoInit
+
 Write-Host "Using configured nodename: $env:NODENAME DNS: $env:DNS_NAME_SERVERS"
 
 Write-Host "Auto-detecting node IP, looking for interface named like '$InterfaceName' ..."
--- a/node/windows-packaging/CalicoWindows/kubernetes/kubelet-service.ps1
+++ b/node/windows-packaging/CalicoWindows/kubernetes/kubelet-service.ps1
@@ -22,6 +22,10 @@ $baseDir = "$PSScriptRoot\.."
 ipmo $baseDir\libs\calico\calico.psm1 -Force
 
 Write-Host "Running kubelet service."
+
+# After restart of node network adapter goes to the wrong state. So, CalicoNode service recreates it.
+# But kubelet can start some pods using broken adapter before CalicoNode complete initialization.
+# So, we need to wait for complete Calico initialization.
 Wait-ForCalicoInit
 
 Write-Host "Using configured nodename: $env:NODENAME DNS: $env:DNS_NAME_SERVERS"


song-jiang:
@vitaliy-leschenko Thanks for your PR and details.  There's many considerations of how user installs/starts kubelet/kube-proxy service. Calico provides some sample scripts but that's not the only way to do it, e.g. some cloud platform has their own node provision procedures for it. I don't think node-service should start kubelet/kube-proxy and this step should be left to users.

node/windows-packaging/CalicoWindows/node/node-service.ps1
```diff
@@ -118,6 +118,10 @@ if ($env:CALICO_NETWORKING_BACKEND -EQ "windows-bgp" -OR $env:CALICO_NETWORKING_
     }
 }
 
+# Start k8s services
+Start-Service kubelet
+Start-Service kube-proxy
+
```
song-jiang, line 124: I think we shouldn't use node-service to start kubelet/kube-proxy services.

song-jiang (changes requested):

node/windows-packaging/CalicoWindows/node/node-service.ps1
vitaliy-leschenko: I agree that this is not a cool solution, but if we start kubelet before primary network adapter has been prepared it might use wrong adapter to start pods.
It was a reason for the PR.

When we restart a node, we have hyper-v network adapter instead of native adapter but we don't know if it is new or old one. CalicoNode service will remove old network adapters. So, if kubelet starts a pod with it the pod will be broken when CalicoNode completes their job.

vitaliy-leschenko: @song-jiang I change the strategy. Now `kubelet` is awaiting for the calico network initialized.
What do you think about this way?

node/windows-packaging/CalicoWindows/kubernetes/kubelet-service.ps1
```diff
@@ -22,6 +22,8 @@ $baseDir = "$PSScriptRoot\.."
 ipmo $baseDir\libs\calico\calico.psm1 -Force
 
 Write-Host "Running kubelet service."
+Wait-ForCalicoInit
```
lmm, line 25: I think it would be helpful to have a comment explaining why we're waiting for CalicoNode to finish initialiation

lmm (commented):
Added a quick comment but otherwise I think this is looking good.

lmm:
Btw, thank you for raising this PR @vitaliy-leschenko! 🙌

node/windows-packaging/CalicoWindows/kubernetes/kubelet-service.ps1
vitaliy-leschenko: If I correctly understand you I did it.

lmm (approved):

Issue #6450, opened 2022-07-28
VM access was blocked when eBPF dataplane used


TrevorTaoARM:
When I enabled the Calico eBPF dataplane for a K8s cluster, the VMs(for which the NIC was bridged on the physical NIC of the server) on the node which had been configured with the eBPF dataplane can't be accessed with normal ssh access. 
When the kube-proxy was restored and eBPF DP disabled, the SSH access to VM was also restored.

## Steps to Reproduce (for bugs)

The following script was used to enable eBPF dataplane:
#!/bin/bash
set -x

WORKDIR=$(pwd)
TMP_DIR=$(mktemp -d)
MARCH=$(uname -m)
CALICO_VERSION=${1:-3.23.2}

if [ $MARCH == "aarch64" ]; then ARCH=arm64;
elif [ $MARCH == "x86_64" ]; then ARCH=amd64;
else ARCH="unknown";
fi
echo ARCH=$ARCH

k8s_ep=$(kubectl get endpoints kubernetes -o wide | grep kubernetes | cut -d " " -f 4)
k8s_host=$(echo $k8s_ep | cut -d ":" -f 1)
k8s_port=$(echo $k8s_ep | cut -d ":" -f 2)

cat <<EOF > ${WORKDIR}/k8s_service.yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: kubernetes-services-endpoint
  namespace: kube-system
data:
  KUBERNETES_SERVICE_HOST: "__KUBERNETES_SERVICE_HOST__"
  KUBERNETES_SERVICE_PORT: "__KUBERNETES_SERVICE_PORT__"
EOF
sed -i "s/__KUBERNETES_SERVICE_HOST__/${k8s_host}/" ${WORKDIR}/k8s_service.yaml
sed -i "s/__KUBERNETES_SERVICE_PORT__/${k8s_port}/" ${WORKDIR}/k8s_service.yaml
kubectl apply -f ${WORKDIR}/k8s_service.yaml

echo "Disable kube-proxy:"
kubectl patch ds -n kube-system kube-proxy -p '{"spec":{"template":{"spec":{"nodeSelector":{"non-calico": "true"}}}}}'

if [ ! -f /usr/local/bin/calicoctl ]; then
   echo "No calicoctl, install now:"
   curl -L https://github.com/projectcalico/calico/releases/download/v${CALICO_VERSION}/calicoctl-linux-${ARCH} -o ${WORKDIR}/calicoctl;
[...]

PR #6663, opened 2022-09-05
Let route unknown traffic skip the eBPF processing

TrevorTaoARM:
When we enable Calico eBPF dataplane, and a packet(e.g, a ping ICMP packet) destined
for a VM of the host(VMs are usually connected to host's physical interface through
the macvtap/macvlan interfaces in either Bridge, VEPA or passthrough mode)
from the physical interface, would be falsely bypassed
by the eBPF program here and can't reach the target VM from the virtual interface
(macvtap/macvlan).

When a packet comes into the eBPF program of traffic control, its
destination address(daddr) should be checked if it's for a known route
by checking the route map, and if it's for an unknown route, it should
be thought as it's not destined for this system, so we should just let
it go through(skip) all our eBPF programs processing here by setting the
action to TC_ACT_OK, which would skip for subsequent eBPF checkings
and processings.

So here we also should not check the unknown route traffic against FIB by bpf_fib_lookup
(in forward_or_drop()), since in some systems, the lookup result would be successful
like this:
  <idle>-0       [088] d.s. 1810775.267240: bpf_trace_printk: enp9s0---I: Traffic is towards the host namespace, doing Linux FIB lookup
  <idle>-0       [088] d.s. 1810775.267243: bpf_trace_printk: enp9s0---I: FIB lookup succeeded - with neigh
  <idle>-0       [088] d.s. 1810775.267244: bpf_trace_printk: enp9s0---I: Got Linux FIB hit, redirecting to iface 2.
  <idle>-0       [088] d.s. 1810775.267245: bpf_trace_printk: enp9s0---I: Traffic is towards host namespace, marking with 0x3000000.
  <idle>-0       [088] d.s. 1810775.267247: bpf_trace_printk: enp9s0---I: Final result=ALLOW (0). Program execution time: 31307ns
  <idle>-0       [088] d.s. 1810775.267249: bpf_trace_printk: enp9s0---E: New packet at ifindex=2; mark=3000000
  <idle>-0       [088] d.s. 1810775.267250: bpf_trace_printk: enp9s0---E: Final result=ALLOW (3). Bypass mark bit set.

and it's a wrong processing here since for the packet of a mark of 3000000 at
the egress direction would be discarded by the system.

On the other side, we also noticed in some systems, the issue of VM
access blocking seems to be disappeared, and the packet can go through the
eBPF program and finally reach the target VM. In this case, it does not mean the
original action is correct, but just because the FIB lookup just fails here(see the log
below), so the packet would be bypass by the eBPF program here with a mark 0x1000000:
  <idle>-0       [014] ..s. 17619198.981285: 0: eno1np0--I: Traffic is towards the host namespace, doing Linux FIB lookup
  <idle>-0       [014] ..s. 17619198.981287: 0: eno1np0--I: FIB lookup failed (FIB problem): 7.
  <idle>-0       [014] ..s. 17619198.981287: 0: eno1np0--I: Traffic is towards host namespace, marking with 0x1000000.
  <idle>-0       [014] ..s. 17619198.981288: 0: eno1np0--I: Final result=ALLOW (0). Program execution time: 16040ns
So it can correctly skip the wrong marking action above.

At the same time, we would like to say there is a similar processing for
the unrelevant traffic in Cilium eBPF implementation:
        ep = lookup_ip4_endpoint(ip4);
https://github.com/cilium/cilium/blob/master/bpf/bpf_host.c#L571

and
        if (!from_host)
                return CTX_ACT_OK;
https://github.com/cilium/cilium/blob/master/bpf/bpf_host.c#L586

Here the endpoint of Cilium eBPF is similar to the route of Calico eBPF.

This patch is also a fix for the issue of
"VM access was blocked when eBPF dataplane used"
https://github.com/projectcalico/calico/issues/6450

Signed-off-by: trevor tao <trevor.tao@arm.com>

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/fib.h
+++ b/felix/bpf-gpl/fib.h
@@ -56,6 +56,11 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
 		goto deny;
 	}
 
+	if (rc == TC_ACT_OK) {
+		CALI_DEBUG("See TC_ACT_OK\n");
+		goto ok;
+	}
+
 	if (rc == CALI_RES_REDIR_BACK) {
 		int redir_flags = 0;
 		if  (CALI_F_FROM_HOST) {
@@ -271,6 +276,15 @@ skip_fib:
 	}
 
 	return rc;
+
+ok:
+	if (CALI_LOG_LEVEL >= CALI_LOG_LEVEL_INFO) {
+		__u64 prog_end_time = bpf_ktime_get_ns();
+		CALI_INFO("Final result=OK (%x). Program execution time: %lluns\n",
+				reason, prog_end_time-state->prog_start_time);
+	}
+
+	return TC_ACT_OK;
 
 deny:
 	if (CALI_LOG_LEVEL >= CALI_LOG_LEVEL_INFO) {
--- a/felix/bpf-gpl/parsing.h
+++ b/felix/bpf-gpl/parsing.h
@@ -8,6 +8,7 @@
 #define PARSING_OK 0
 #define PARSING_OK_V6 1
 #define PARSING_ALLOW_WITHOUT_ENFORCING_POLICY 2
+#define PARSING_ALLOW_WITHOUT_ROUTE 3
 #define PARSING_ERROR -1
 
 static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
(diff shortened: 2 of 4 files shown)


fasaxc:
I'm not comfortable just allowing all traffic with no routes through; it feels like that could be exploited somehow.  It seems safest to block this traffic unless there's a way to distinguish "expected" unknown traffic from "unexpected".  Perhaps a config flag that accepts a list of CIDRs to consider "safe"?

TrevorTaoARM:
> 

A packet without known routes should be thought as irrelevant to our proccessing logic of Calico eBPF program, since the traffic that is needed to be forwarded to entities(e.g., pods, bridge, virtual IP(service IP) or virtual interfaces) managed by Calico itself should be route known to Calico.
The same processing to unknown traffic exists in another CNI--Cilium, which is mainly eBPF based. For Cilium, the "endpoint" is similar to Calico's "route", and Cilium just bypass(let it go) for traffic without matching endpoint.

PR #6669, opened 2022-09-07
Reinstate DevStack testing (for OpenStack)

neiljerram:
Yay, we have DevStack testing again.  Changes needed were as follows:

- Use e1-standard-4 instead of e1-standard-2.  With e1-standard-2 we get OOM killing during the test run.
- Tweaked method for passing in the location and ref of the networking-calico code to use as a DevStack plugin, now that networking-calico is part of the monorepo.
- Top-level `devstack` -> `networking-calico/devstack` symbolic link, needed for the monorepo as a whole to function as a DevStack plugin.
- Pin to Tempest code similar to what we were using (from Tempest master) when this test was last running and passing.


mgleung (approved):
LGTM

Issue #6670, opened 2022-09-07
Not possible to disable IPv6 VXLAN


muff1nman:
## Expected Behavior
When upgrading from Calico 3.22 to 3.23 (7dbb81af0c4292db09abf8a416f597e6487fb1b8), ipv6 vxlan should not be enabled by default, or if it is, there should be a way to disable it.

## Current Behavior
On upgrading to Calico 3.23 and above, Calico fails to setup IPv6 VXLAN with the following messages:
```
2022-09-06 17:17:54.882 [INFO][43] felix/route_table.go 1159: Failed to access interface because it doesn't exist. error=Link not found ifaceName="vxlan-v6.calico" ifaceRegex="^vxlan-v6.calico$" ipVersion=0x6 tableIndex=0
2022-09-06 17:17:54.882 [INFO][43] felix/route_table.go 1227: Failed to get interface; it's down/gone. error=Link not found ifaceName="vxlan-v6.calico" ifaceRegex="^vxlan-v6.calico$" ipVersion=0x6 tableIndex=0
2022-09-06 17:17:54.882 [ERROR][43] felix/route_table.go 994: Failed to get link attributes error=interface not present ifaceRegex="^vxlan-v6.calico$" ipVersion=0x6 tableIndex=0
2022-09-06 17:17:54.882 [INFO][43] felix/route_table.go 587: Interface missing, will retry if it appears. ifaceName="vxlan-v6.calico" ifaceRegex="^vxlan-v6.calico$" ipVersion=0x6 tableIndex=0
```
However, no IPv6 pools were configured with IPv6 VXLAN encapsulation. Furthermore it is not possible to disable IPv6 encapsulation, as removing `vxlanEnabled` from the `default` FelixConfiguration, leads to IPv4 VXLAN encapsulation being broken. This is contrary to the documentation which states this about `vxlanEnabled`:
```
Optional, you shouldn’t need to change this setting as Felix calculates if VXLAN should be enabled based on the existing IP Pools. When set, this overrides whether Felix should create the VXLAN tunnel device for VXLAN networking.
```

## Possible Solutions
- Decouple vxlanEnabled from vxlan6Enabled ( #6671 has been commandeered to go down this route)
- ~Only have `vxlanEnabled` be a field to override the ippools to Enabled. Setting it to Never would have no effect if the ippols have vxlanMode enabled. (#6671 explores this idea)~
- Instead focus on fixing why vxlan-v6.calico interface is having issues

[...]

PR #6671, opened 2022-09-07
felix/encap: ignore vxlanEnabled for ipv6 

muff1nman:
## Description
This decouples ipv6 vxlan from ipv4 allowing either to be enabled independently and further reduces the need to set vxlanEnabled at the global felix configuration layer.

## Related issues/PRs

https://github.com/projectcalico/calico/issues/6670
https://github.com/projectcalico/calico/pull/6498
https://github.com/projectcalico/calico/pull/6494
https://github.com/projectcalico/calico/issues/6442
https://github.com/projectcalico/calico/pull/5576

## Todos

- [X] Tests
- [ ] Documentation
- [X] Release note

## Release Note

```release-note
vxlanEnabled is now ignored for the IPv6 VXLAN pools allowing VXLAN to have IPv4 enabled independently from IPv6.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/calc/encapsulation_resolver.go
+++ b/felix/calc/encapsulation_resolver.go
@@ -267,9 +267,5 @@ func (c *EncapsulationCalculator) VXLANEnabled() bool {
 }
 
 func (c *EncapsulationCalculator) VXLANEnabledV6() bool {
-	if c.config != nil && c.config.VXLANEnabled != nil {
-		return *c.config.VXLANEnabled
-	}
-
 	return len(c.vxlanPoolsv6) > 0
 }
--- a/api/pkg/apis/projectcalico/v3/felixconfig.go
+++ b/api/pkg/apis/projectcalico/v3/felixconfig.go
@@ -211,7 +211,7 @@ type FelixConfigurationSpec struct {
 	// IPIPMTU is the MTU to set on the tunnel device. See Configuring MTU [Default: 1440]
 	IPIPMTU *int `json:"ipipMTU,omitempty" confignamev1:"IpInIpMtu"`
 
-	// VXLANEnabled overrides whether Felix should create the VXLAN tunnel device for VXLAN networking. Optional as Felix determines this based on the existing IP pools. [Default: nil (unset)]
+	// VXLANEnabled overrides whether Felix should create the VXLAN tunnel device for IPv4 VXLAN networking. Optional as Felix determines this based on the existing IP pools. [Default: nil (unset)]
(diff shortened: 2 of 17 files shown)


coutinhop:
@muff1nman thanks for your PR, though I'm not sure about these changes... The intention when making felix determing VXLAN encap by the existing IP pools was that the vxlanEnabled FelixConfig would no longer be necessary, but would be respected to allow for seamless upgrading. This is broken it seems, but I think the real issue is that the FelixConfig should not touch IPv6 (seeing as it was added after these changes).

What do you think @caseydavenport?

felix/calc/encapsulation_resolver.go
```diff
@@ -251,25 +251,33 @@ func (c *EncapsulationCalculator) removePool(cidr string) {
...
+	return false
 }
 
 func (c *EncapsulationCalculator) VXLANEnabledV6() bool {
+	if len(c.vxlanPoolsv6) > 0 {
+		return true
+	}
 	if c.config != nil && c.config.VXLANEnabled != nil {
 		return *c.config.VXLANEnabled
 	}
```
coutinhop, line 280: @muff1nman I'm thinking in order to fix https://github.com/projectcalico/calico/issues/6670 we could leave the rest of this file as it was (so revert all changes) + remove these lines. My intention would be to decouple the EncapsulationCalculator logic for IPv6 VXLAN from FelixConfig entirely, as it doesn't need an "override" like the IPv4 encapsulations.

We'd also need some test changes to conform to that.

muff1nman: Okay sounds good, let me try going down that route

muff1nman:
@coutinhop Testing this PR allows upgrading from 3.22 to 3.23 without IPv6 VXLAN getting in the way. However it does not solve the issue that `vxlanEnabled: true` cannot be dropped without removing all of the IPv4 vxlan interfaces. I imagine it makes sense to pursue that in a separate change? And maybe it makes sense to not yet update the `vxlanEnabled` description in this PR because I don't feel the change I've made is accurate. Considering just this change alone, the description should maybe say:

```
VXLANEnabled must be set to true for IPv4 VXLAN to be enabled. Ignored for IPv6. [Default: nil (unset)]
```
But maybe we just don't change that quite yet and wait for this hypothetical second change? Also would make backporting easier..

coutinhop (approved):
LGTM, thanks @muff1nman for these changes!

Issue #5146, opened 2021-12-04
ipv6 only deployement, Open Message Error: Bad BGP Identifier


abasitt:
In IPv6 only kubernetes deployment, router-id is derived from the hash of nodename. This is good and i saw some previous issues to make it this way to automate the process of router-id. But if you are unlucky like me and router-id ends up in the range of 224.xxxx reserved multicast range, the peering router start complaining and the BGP get stuck in the open message with an error of Open Message Error: Bad BGP Identifier. 

## Expected Behavior

The expected behavior should be to put a check on the router-id in a way that it is not from the reserved multicast range. I believe this can be achieved by not using specific bits in the first octet of router-id to avoid this range. https://kb.juniper.net/InfoCenter/index?page=content&id=KB26645&cat=FIREWALL_IPSEC_VPN&actp=LIST

## Current Behavior
The router-id is generated randomly including the reserved range of 224.xxx. 

2021/12/01 02:32:05 BGP: x:x:x:4::11 bad OPEN, wrong router identifier 249.183.186.85
2021/12/01 03:16:34 BGP: x:x:x:4::12 bad OPEN, wrong router identifier 241.228.95.65

## Possible Solution

The best and easy solution will be to fix the code and don't include the reserved ranges. This can be achieved by avoiding specific bits in the last octet of IPv4.
Another suggestion is to please include router-id parameter in the BGP node configuration yaml file. Currently there is no easy way to pass the manual router-id via daemon-set.

## Steps to Reproduce (for bugs)
1. I used a kubernete hostname of cl1-master1.rcp.local and this generated a router id of 249.183.186.85. 
2. 
3.
4.

## Context
I am using IPv6 only kubernetes cluster, my nodes have no IPV4 addresses and I am relying on node hash for BGP router-id. Everything works fine when BGP remains within BIRD which full-mesh. When i am trying to do the peering with external router using eBGP, the upstream router doesn't accept router-id in reserved ranges. With adoption of IPv6, I believe more people will face this issues in future.

## Your Environment

* Calico version
* Orchestrator version (e.g. kubernetes, mesos, rkt):
* Operating System and version:
* Link to your project (optional):
##
Client Version:    v3.21.1
Git commit:        183af6a8
Cluster Version:   v3.21.1
Cluster Type:      typha,kdd,k8s,operator,bgp,kubeadm

[...]

PR #6674, opened 2022-09-08
fix: update hashToIPV4 method

ramanujadasu:
## Description
Allow unicast ip addresses with existing logic hashToIPv4 method
For bug fix: https://github.com/projectcalico/calico/issues/5146

## Related issues/PRs
https://github.com/projectcalico/calico/issues/5146

## Todos

- [X] Tests
- [X] Documentation
- [X] Release note

## Release Note

```release-note
Fix bug in IPv6 router ID calculation on IPv6 single-stack clusters that resulted in invalid router IDs being calculated. Note that this change will result in new router IDs being used for some IPv6 single-stack nodes.
```

## Reminder for the reviewer
release-note-not-required
docs-not-required

Other optional labels:


confd/pkg/resource/template/template_funcs.go
```diff
@@ -56,7 +56,12 @@ func hashToIPv4(nodeName string) string {
 	}
 	hashBytes := hash.Sum(nil)
 	ip := hashBytes[:4]
-	routerId := strconv.Itoa(int(ip[0])) + "." +
+	//Allow unicast ip
+	ip0Value := int(ip[0])
+	if ip0Value > 223 {
+		ip0Value = ip0Value - 32
```
caseydavenport, line 62: I wonder if we can some up with an algorithm that provides better entropy than this approach. This approach condenses certain ranges.

e.g.,

- 224 always maps to 192
- 225 always maps to 193
- etc.

It would be nice if we could instead remap 224 across all of the available bit space - 0 through 223.

What if instead, we did something like this:

```go

hashString := nodename
bytes := hash(hashString )

while bytes[0] > 223 {
  // BGP doesn't allow router IDs in special IP ranges (e.g., 224.x.x.x). 
  // So long as the generated hash is invalid, keep adding characters and re-hashing.
  hashString += nodename
  bytes = hash(hashString)
}
```

I think that's still deterministic - a given nodename will always hash to the same value. It also should give us slightly better entropy because instead of directly mapping 224 -> 192, 225 -> 193, we have all of 0->223 to remap to. Sometimes 224 will remap to 192, but sometimes it will remap to something else. 

WDYT?

caseydavenport: The tradeoff here of course is that we'd use additional CPU in some cases hashing 2-3 extra times. But I think that's probably OK given the relative rarity of this function being called. 

Perhaps if we're worried about that we could put a cap on the number of times we retry the hash, or maybe cache the computed value so that future calls won't need to recalculate it.

abasitt: @caseydavenport we thought about that but the reason we put such a simple logic is because hash is already calculated once and IMO there is no need to recalculate again. The reason we come up with number 32 is because highest octet value 255-223=32. 
IMO if 224 always maps to 192 is fine, isn't it the same thing e.g. if we say the node name ABCD will map to 224, but BCDE always map to 225. Since the initial hash already randomize the first octet to a random value, if we minus a constant from it, the result will always be a random as well.
On top since we are not touching the 3 additional octet, the combined 4 octet will always be random anyway.

abasitt: The second option we thought of was to use bit-wise operations but minus a constant 32 is the most simple one. Please let me know your thoughts. 
Thank you so much for the feedback.

caseydavenport: > I am afraid we might end up having different router_id on restarts or upgrades 

It won't calculate a new router ID since the hash is deterministic, and the nodename will be the same across restarts. So, given nodename XYZ, the same router ID will always be generated. 

> IMO if 224 always maps to 192 is fine, isn't it the same thing e.g. if we say the node name ABCD will map to 224, but BCDE always map to 225. Since the initial hash already randomize the first octet to a random value, if we minus a constant from it, the result will always be a random as well.

Yep, it will still be random, but it will be a smaller space than if we were to use the suggested algorithm.

If we subtract 32 from the value, it means all hashes with a first octet of >223 will be remapped to somewhere between 192 and 223. 

If we use the suggested algorithm, hashes with a first octet of >223 will be remapped to somewhere between 0 and 223, giving us a bit more breathing room to prevent hashing conflicts (two nodes hashing to the same router ID). 

> because hash is already calculated once and IMO there is no need to recalculate again

The hash is actually calculated every time that the BGP configuration or BGP peerings change, although it will always return the same value. That's why I was suggesting perhaps caching the value so that future calls to the function don't need to re-hash, but that's just optimization and could be done in a separate PR.

abasitt: I actually thought of that conflict but IMO it's still very unlikely because even if we end up getting two router with first octet of 192, the other three octets will be different. Anyway I understand the concern of worst case scenario, but I am not in favor of re-calculating hash multiple times. We can make it random with below as well. WDYT ?

	//BGP doesn't allow router IDs in special IP ranges (e.g., 224.x.x.x)
	//Scan and pick the value < 223 in the hashBytes[]
	if ip0Value > 223 {
		for i := 5; i < len(hashBytes); i++ {
			ip0Value = int(hashBytes[i])
			if ip0Value < 223 {
				break
			}
			//if no value < 223 in hashBytes[] then force it by subtraction
			if ip0Value > 223 {
				ip0Value = ip0Value - 32
			}
		}
	}
I test this with different hostnames, it is extremely unlikely that subtraction by 32 will ever hit but just keeping it there for worst case scenario. https://go.dev/play/p/ZfBDDy6sgHF

caseydavenport (approved):
Let's just leave it as-is then. If we hit issues with conflicts we'll have to revisit and implement another algorithm.

caseydavenport:
@abasitt we'll probably need to update the expected test data in the confd FV tests

abasitt (approved):

abasitt:
@caseydavenport will work on the FV test cases. thank you.

caseydavenport:
@abasitt I'm not sure why I pinged you on this one, probably because you raised the issue originally!😅 

I believe that ping was originally meant for @ramanujadasu, although obviously you're welcome to help here if you want!

abasitt:
That's fine, we are in the same team ;)

caseydavenport (approved):

PR #6675, opened 2022-09-08
Split Typha metrics by syncer type

fasaxc:
## Description

This PR splits Typha's Prometheus metrics by syncer type. Many of Typha's statistics are only meaningful within the scope of a syncer but previously, all the syncers wrote to the same metrics resulting in metrics that would bounce around between the values associated with the different syncers.

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Many of Typha's Prometheus metrics are now split by syncer (client) type, represented by a label "syncer" on the metrics. This prevents cross-talk where the syncers would all share the same metrics and the last writer to the metric would "win".
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/typha/pkg/syncserver/sync_server.go
+++ b/typha/pkg/syncserver/sync_server.go
@@ -60,11 +60,15 @@ var (
 		Name: "typha_connections_dropped",
 		Help: "Total number of connections dropped due to rebalancing.",
 	})
+	gaugeNumConnections = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "typha_connections_active",
+		Help: "Number of open client connections, including connections that are still in the handshake.",
+	})
 
 	// Counters with per-syncer-type values.
-	gaugeVecNumConnections = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "typha_connections_active",
-		Help: "Number of open client connections.",
+	gaugeVecNumConnectionsStreaming = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "typha_connections_streaming",
+		Help: "Number of client connections that completed the handshake and are streaming data.",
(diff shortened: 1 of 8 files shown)


typha/pkg/syncserver/sync_server.go
```diff
@@ -566,6 +555,7 @@ func (h *connection) handle(finishedWG *sync.WaitGroup) (err error) {
 	if err = h.doHandshake(); err != nil {
 		return // Error already logged.
 	}
+	h.gaugeNumConnections.Inc()
```
caseydavenport, line 558: Do we need a nil check here as well? I see the Dec() call above is protected.

caseydavenport: I suppose moving this down below the handshake changes this gauge's meaning as well, but that's probably OK.

caseydavenport (approved):

typha/pkg/syncserver/sync_server.go
fasaxc: No, the gauge is set up by `doHandshake()` so no nil check needed but the defer block runs even if we don't get this far.  Suppose I could do `defer h.gaugeNumConnections.Dec()` here instead but previously we only decremented after all connection resources were freed.

marvin-tigera:

marvin-tigera:

typha/pkg/syncserver/sync_server.go
fasaxc: Hmm, some tests were broken by that change (and they were using it to test the interaction with handshake).  I'll reinstate the old stat and add a new one for each syncer type.

fasaxc:
Failures look to be from tests that don't use typha in any way so writing off as flakes.

PR #6676, opened 2022-09-08
Always program OpenStack floating IPs

neiljerram:
## Description

Felix's programming of NAT mappings was disabled by default in https://github.com/projectcalico/calico/pull/5861.  This was a security fix for Kubernetes, as in Kubernetes anyone could create a floating IP for themselves by adding a "floatingips" annotation to their own pods.

For OpenStack, pod annotations obviously do not apply, and NAT mappings come from the ML2 driver running as part of the Neutron server, which is already access-protected in an appropriate way.  And we don't want to have to document that OpenStack users need to add "FloatingIPs = Enabled" to their felix.cfg file.

Therefore, arrange for OpenStack-configured floating IPs to be programmed regardless of the FloatingIPs setting in FelixConfiguration.

## Related issues/PRs

qualifies https://github.com/projectcalico/calico/pull/5861

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/endpoint_mgr.go
```diff
@@ -658,13 +659,9 @@ func (m *endpointManager) resolveWorkloadEndpoints() {
...
-				if m.floatingIPsEnabled && len(natInfos) != 0 {
-					// Include any floating IP NATs if the feature is enabled.
-					old := ipStrings
-					ipStrings = make([]string, len(old)+len(natInfos))
-					copy(ipStrings, old)
-					for ii, natInfo := range natInfos {
-						ipStrings[len(old)+ii] = natInfo.ExtIp + addrSuffix
+				for _, natInfo := range natInfos {
+					if m.floatingIPsEnabled || natInfo.Source == apiv3.IPNATSourceOpenStack {
+						ipStrings = append(ipStrings, natInfo.ExtIp+addrSuffix)
```
fasaxc, line 664: Not sure if this is OK, appending to a slice can share the same backing array so if two bits of code decide to append different things to the same slice then they might see each other's updates.  Better to take your own copy?

fasaxc (changes requested):
How about using the orchestrator ID in the WEP key then we don't need a datamodel change?  We've carried that ID around for years maybe it should pay its way!

neiljerram:
Great idea about the orchestrator ID, and that should also make the change a lot smaller - I'll take a look at that.

felix/dataplane/linux/endpoint_mgr.go
neiljerram: Yes, I've restored logic like we had here before.

fasaxc (approved):

PR #6680, opened 2022-09-09
Add FV test for host endpoint policy using a label inherited from a profile

nelljerram:
Based on a failure in our OpenStack ST, I suspected that this scenario had stopped working, and so wrote this test to investigate.  In fact the test passes, but it's worth keeping.  There must be another explanation for the failure in the OpenStack ST.

Update: the OpenStack ST failure was caused by configuring a Profile with an invalid name, "test_profile".  Underscore is not allowed as a character in Calico resource names.  This used not to matter, but started mattering because of the change (also by me) to stream v3 Profile objects from the Felix syncer instead of the model/v1 ProfileLabels, ProfileRules and ProfileTags objects.

## Release Note

```release-note
None
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
@nelljerram seems like the CLA bot is confused about this PR for some reason

nelljerram:
@caseydavenport I'll close it.  IIRC I discovered after posting this that we already have another test covering this function.

Issue #6649, opened 2022-09-01
Built binaries are not stripped


croustiPoulet:
Built binaries are not stripped

## Expected Behavior
Built binaries are as small as possible to reduce docker images size.

## Current Behavior
As an example, calico-node is not stripped:
calico-node: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, Go BuildID=Zj-dNERc1XZs22xTIG4a/6iGR1YjI3SePsudHbMMA/BM-owAa7TDzFpy2od1aW/84jlD79A8qOMoOkeJqpA, BuildID[sha1]=519824671dc019b4983ac44e98740d4c76d2c9fa, not stripped

This is the same for binaries in kube-controllers, cni and node docker images.

## Possible Solution
Use ldflags "-w -s"

caseydavenport:
The one complexity here is that we use the symbols as part of our dev process to verify the correct TLS boringcrypto symbols show up.

What we should do is keep the symbols for pre-release / development images, and then strip them when we cut a release.

afshinpaydar-binary:
@caseydavenport Please help to review [this](https://github.com/projectcalico/calico/pull/6686) solution for adding `ldflags "-w -s"` exactly during the final release.

fasaxc:
Stripping symbols makes it very hard to debug core dumps, which don't happen often, but when they do... I feel like, if we're going to strip images, we should save off the symbols somewhere so that they can be dug out when needed.  Could also try a binary compressor instead.  I think UPX works with Go now.

PR #6686, opened 2022-09-11
Built stripped binaries when release

afshin-deriv:
Signed-off-by: Afshin Paydar <afshin.paydar@binary.com>

## Description
- Built binaries as small as possible to reduce docker images size, by stripping symbols from binaries only when release.
- type of fix: `kind/enhancement`

## Related issues/PRs
- https://github.com/projectcalico/calico/issues/6649

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/apiserver/Makefile
+++ b/apiserver/Makefile
@@ -45,7 +45,8 @@ APISERVER_GIT_DESCRIPTION?=$(shell git describe --tags)
 LDFLAGS = -X $(PACKAGE_NAME)/cmd/apiserver/server.VERSION=$(APISERVER_VERSION) \
 	-X $(PACKAGE_NAME)/cmd/apiserver/server.BUILD_DATE=$(APISERVER_BUILD_DATE) \
 	-X $(PACKAGE_NAME)/cmd/apiserver/server.GIT_DESCRIPTION=$(APISERVER_GIT_DESCRIPTION) \
-	-X $(PACKAGE_NAME)/cmd/apiserver/server.GIT_REVISION=$(APISERVER_GIT_REVISION)
+	-X $(PACKAGE_NAME)/cmd/apiserver/server.GIT_REVISION=$(APISERVER_GIT_REVISION) \
+	$(RELEASE_LDFLAGS)
 
 ##############################################################################
 # Download and include ../lib.Makefile before anything else
@@ -85,7 +86,7 @@ ifeq ($(ARCH),amd64)
 	$(call build_cgo_boring_binary, $(PACKAGE_NAME)/cmd/apiserver, $@)
 else
 	$(DOCKER_RUN) -e CGO_ENABLED=$(CGO_ENABLED) $(CALICO_BUILD) \
-		sh -c '$(GIT_CONFIG_SSH) go build -v -o $@ -ldflags "$(VERSION_FLAGS)" $(PACKAGE_NAME)/cmd/apiserver'
+		sh -c '$(GIT_CONFIG_SSH) go build -v -o $@ -ldflags "$(LDFLAGS)" $(PACKAGE_NAME)/cmd/apiserver'
 endif
 
 $(BINDIR)/filecheck-$(ARCH): $(K8SAPISERVER_GO_FILES)
--- a/kube-controllers/Makefile
+++ b/kube-controllers/Makefile
@@ -42,7 +42,7 @@ clean:
 ###############################################################################
 # Building the binary
 ###############################################################################
-LDFLAGS = -X main.VERSION=$(GIT_VERSION)
+LDFLAGS= -X main.VERSION=$(GIT_VERSION) $(RELEASE_LDFLAGS)
 
 build: bin/kube-controllers-linux-$(ARCH) bin/check-status-linux-$(ARCH)
 build-all: $(addprefix sub-build-,$(VALIDARCHES))
--- a/node/Makefile
+++ b/node/Makefile
@@ -153,7 +153,8 @@ DATE:=$(shell date -u +'%FT%T%z')
 LDFLAGS= -X $(PACKAGE_NAME)/pkg/lifecycle/startup.VERSION=$(GIT_VERSION) \
 	-X $(PACKAGE_NAME)/buildinfo.GitVersion=$(GIT_DESCRIPTION) \
 	-X $(PACKAGE_NAME)/buildinfo.BuildDate=$(DATE) \
-	-X $(PACKAGE_NAME)/buildinfo.GitRevision=$(GIT_COMMIT)
+	-X $(PACKAGE_NAME)/buildinfo.GitRevision=$(GIT_COMMIT) \
+	$(RELEASE_LDFLAGS)
 
 # Source golang files on which compiling the calico-node binary depends.
 SRC_FILES=$(shell find ./pkg -name '*.go') \
--- a/pod2daemon/Makefile
+++ b/pod2daemon/Makefile
@@ -15,6 +15,7 @@ BUILD_IMAGES    ?=$(FLEXVOL_IMAGE) $(CSI_IMAGE) $(REGISTRAR_IMAGE)
 
 include ../lib.Makefile
 
+LDFLAGS= $(RELEASE_LDFLAGS)
 ###############################################################################
 
 
@@ -58,14 +59,14 @@ bin/flexvol-armv7: ARCH=armv7
 bin/flexvol-ppc64le: ARCH=ppc64le
 bin/flexvol-s390x: ARCH=s390x
 bin/flexvol-%: $(SRC_FILES)
-	$(DOCKER_RUN) -e CGO_ENABLED=0 $(CALICO_BUILD) go build -buildvcs=false -v -o bin/flexvol-$(ARCH) flexvol/flexvoldriver.go
+	$(DOCKER_RUN) -e CGO_ENABLED=0 $(CALICO_BUILD) go build -buildvcs=false -v -o bin/flexvol-$(ARCH) -ldflags "$(LDFLAGS)" flexvol/flexvoldriver.go
 
 bin/csi-driver-arm64: ARCH=arm64
 bin/csi-driver-armv7: ARCH=armv7
 bin/csi-driver-ppc64le: ARCH=ppc64le
 bin/csi-driver-s390x: ARCH=s390x
 bin/csi-driver-%: $(SRC_FILES)
-	$(DOCKER_RUN) -e CGO_ENABLED=$(CGO_ENABLED) $(CALICO_BUILD) go build -buildvcs=false -v -o bin/csi-driver-$(ARCH) csidriver/main.go
+	$(DOCKER_RUN) -e CGO_ENABLED=$(CGO_ENABLED) $(CALICO_BUILD) go build -buildvcs=false -v -o bin/csi-driver-$(ARCH) -ldflags "$(LDFLAGS)" csidriver/main.go
 
 bin/csi-driver-amd64: $(SRC_FILES)
 	$(call build_static_cgo_boring_binary, csidriver/main.go, $@)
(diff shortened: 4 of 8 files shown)


.semaphore/semaphore-scheduled-builds.yml
```diff
@@ -152,8 +152,10 @@ blocks:
       matrix:
       - env_var: ARCH
         values: [ "arm64", "ppc64le" ]
+      - env_var: LDFLAGS
```
caseydavenport, line 155: I think this is actually the wrong spot to put this - these are our scheduled builds which are dev builds and we _do_ want the debug symbols present for these builds. 

We only want to strip the binaries for the released images, which are built through a different workflow at the moment.

Here's where our release logic exists: https://github.com/projectcalico/calico/blob/master/hack/release/pkg/builder/builder.go#L374-L388

It calls through to the `release-build` target in each directory, passing in a limited set of env vars.

.semaphore/semaphore-scheduled-builds.yml
```diff
@@ -253,10 +255,12 @@ blocks:
       matrix:
       - env_var: ARCH
         values: [ "arm64", "armv7", "ppc64le", "s390x" ]
+      - env_var: LDFLAGS
```
caseydavenport, line 258: Also looks like you're passing LDFLAGS but the Makefile changes expect RELEASE_LDFLAGS?

pod2daemon/Makefile
```diff
@@ -66,7 +67,7 @@ bin/csi-driver-armv7: ARCH=armv7
 bin/csi-driver-ppc64le: ARCH=ppc64le
 bin/csi-driver-s390x: ARCH=s390x
 bin/csi-driver-%: $(SRC_FILES)
-	$(DOCKER_RUN) -e CGO_ENABLED=$(CGO_ENABLED) $(CALICO_BUILD) go build -v -o bin/csi-driver-$(ARCH) csidriver/main.go
+	$(DOCKER_RUN) -e CGO_ENABLED=$(CGO_ENABLED) $(CALICO_BUILD) go build -v $(RELEASE_FLAGS) -o bin/csi-driver-$(ARCH) csidriver/main.go
```
caseydavenport, line 70: In some cases you're doing this:

```
go build -v $(RELEASE_FLAGS) 
```

And in others (see kube-controllers) you have:

```
go build -v -ldflags "$(RELEASE_LDFLAGS) "
```

I think we need to make these consistent or else one of them won't work.

I'm thinking the pattern we should establish is like the one in the node repo.

At the top of the Makefile, we establish;

```
LDFLAGS = -ldflags "... $(RELEASE_LDFLAGS)"
```

Then in the build command we just do:

```
go build -v $LDFLAGS
```

WDYT?

Makefile
```diff
@@ -61,14 +62,14 @@ bin/tigera-operator-$(GIT_VERSION).tgz: bin/helm $(shell find ./charts/tigera-op
 
 # Build all Calico images for the current architecture.
 image:
```
caseydavenport, line 64: This is actually just a dev target, and isn't used as part of the release process.

pod2daemon/Makefile
afshinpaydar-binary: @caseydavenport Thanks for reviewing and for pointing me in the right direction. I did some changes based on your suggestions, could you please take a look and let me know if there is any problem?

Makefile
afshinpaydar-binary: Reverted this change.

.semaphore/semaphore-scheduled-builds.yml
afshinpaydar-binary: Reverted this changes.

.semaphore/semaphore.yml
```diff
@@ -152,8 +152,10 @@ blocks:
       matrix:
       - env_var: ARCH
         values: [ "arm64", "ppc64le" ]
+      - env_var: RELEASE_LDFLAGS
+        values: [ "-w -s" ]
       commands:
-      - ../.semaphore/run-and-monitor image-$ARCH.log make image ARCH=$ARCH
+      - ../.semaphore/run-and-monitor image-$ARCH.log make image ARCH=$ARCH RELEASE_LDFLAGS=$RELEASE_LDFLAGS
```
caseydavenport, line 158: I still think we shouldn't be touching semaphore.yaml at all - we want the dev builds (the ones produced by this file) to include symbols.

pod2daemon/Makefile
```diff
@@ -15,6 +15,16 @@ BUILD_IMAGES    ?=$(FLEXVOL_IMAGE) $(CSI_IMAGE) $(REGISTRAR_IMAGE)
 
 include ../lib.Makefile
 
+# Calculate a timestamp for any build artefacts.
+DATE:=$(shell date -u +'%FT%T%z')
```
caseydavenport, line 19: Just adding the date here isn't going to get it added to the binary since we also need to add a code variable to store that inforamation. That said, we're internally considering removing build date from our artifacts since it makes the binaries not reproducible, so IMO we should just remove the build date from the new LDFLAGS you're adding in this PR.

pod2daemon/Makefile
```diff
@@ -15,6 +15,16 @@ BUILD_IMAGES    ?=$(FLEXVOL_IMAGE) $(CSI_IMAGE) $(REGISTRAR_IMAGE)
...
+
+LDFLAGS=-ldflags "\
+	-X $(PACKAGE_NAME)/pkg/lifecycle/startup.VERSION=$(GIT_VERSION) \
```
caseydavenport, line 22: all of these LDFLAGS look to be copied from the node Makefile, which means they aren't going to work here because the package structure isn't the same.

e.g., `pkg/lifecycle/startup.VERSION` doesn't exist in the pod2daemon repository.

I think all we need here is:

```
LDFLAGS=-ldflags "$(RELEASE_LDFLAGS)"
```

pod2daemon/Makefile
```diff
@@ -58,15 +68,15 @@ bin/flexvol-armv7: ARCH=armv7
 bin/flexvol-ppc64le: ARCH=ppc64le
 bin/flexvol-s390x: ARCH=s390x
 bin/flexvol-%: $(SRC_FILES)
-	$(DOCKER_RUN) -e CGO_ENABLED=0 $(CALICO_BUILD) go build -v -o bin/flexvol-$(ARCH) flexvol/flexvoldriver.go
+	$(DOCKER_RUN) -e CGO_ENABLED=0 $(CALICO_BUILD) go build -v -o bin/flexvol-$(ARCH) flexvol/flexvoldriver.go $(LDFLAGS)
```
caseydavenport, line 71: This isn't going to work - you'll need to put `LDFLAGS` prior to any golang source files in this command.

You can try it out by running `make build` inside the pod2daemon repository, and you'll see that the build fails.

caseydavenport:
@afshinpaydar-binary is this something you're still planning to work on?

afshin-deriv:
@caseydavenport Yes, I will address your comments soon. hopefully next week

kube-controllers/Makefile
```diff
@@ -31,6 +31,10 @@ else
 CGO_ENABLED=0
 endif
 
+LDFLAGS=-ldflags "\
```
caseydavenport, line 34: All these shorter ones don't need to be split across multiple lines - let's just keep it on one for simplicity.

caseydavenport (commented):
Sorry for the long delay here. It looks like this will need a rebase though.

Just one minor comment from me otherwise.

caseydavenport:
@afshin-deriv Also, I'm afraid we can't accept this PR unless you sign the CLA to make the bot happy.

pod2daemon/Makefile
afshin-deriv: @caseydavenport One question: Where does the value of the RELEASE_LDFLAGS parameter come from? as It's expected to contain "-w -s".

kube-controllers/Makefile
afshin-deriv: That seems reasonable. I have addressed it.

apiserver/Makefile
```diff
@@ -42,10 +42,12 @@ APISERVER_BUILD_DATE?=$(shell date -u +'%FT%T%z')
 APISERVER_GIT_REVISION?=$(shell git rev-parse --short HEAD)
 APISERVER_GIT_DESCRIPTION?=$(shell git describe --tags)
 
-LDFLAGS = -X $(PACKAGE_NAME)/cmd/apiserver/server.VERSION=$(APISERVER_VERSION) \
+LDFLAGS =-ldflags "\
```
hjiawei, line 45: @rene-dekker You moved `-ldflags` out of `LDFLAGS` into the `go build` command line in a previous PR (for example, this [one](https://github.com/projectcalico/calico/pull/7383/files#diff-72be7afda7c33dd9a23c823b2824820d758e68e8f467465dc58f26b52a942471R35)). This change moved them back. Any reasons why we are moving it back and forth?

afshin-deriv: @hjiawei Thank you for pointing this out. I have removed `-ldflags` again from `LDFLAGS`.

caseydavenport:
```
Starting with UID : 1001
CGO enabled, switching GOROOT to /usr/local/go-cgo.
sh: 1: Syntax error: Unterminated quoted string
Makefile:81: recipe for target 'bin/calico-typha-amd64' failed
```

^ Builds seem to be failing with this error

apiserver/Makefile
```diff
@@ -45,7 +45,8 @@ APISERVER_GIT_DESCRIPTION?=$(shell git describe --tags)
 LDFLAGS = -X $(PACKAGE_NAME)/cmd/apiserver/server.VERSION=$(APISERVER_VERSION) \
 	-X $(PACKAGE_NAME)/cmd/apiserver/server.BUILD_DATE=$(APISERVER_BUILD_DATE) \
 	-X $(PACKAGE_NAME)/cmd/apiserver/server.GIT_DESCRIPTION=$(APISERVER_GIT_DESCRIPTION) \
-	-X $(PACKAGE_NAME)/cmd/apiserver/server.GIT_REVISION=$(APISERVER_GIT_REVISION)
+	-X $(PACKAGE_NAME)/cmd/apiserver/server.GIT_REVISION=$(APISERVER_GIT_REVISION) \
+	$(RELEASE_LDFLAGS)"
```
rene-dekker, line 49: Please remove the quote from this line

typha/Makefile
```diff
@@ -29,7 +29,8 @@ include ../lib.Makefile
 LDFLAGS := -X $(PACKAGE_NAME)/pkg/buildinfo.GitVersion=$(GIT_DESCRIPTION) \
 			-X $(PACKAGE_NAME)/pkg/buildinfo.BuildDate=$(DATE) \
 			-X $(PACKAGE_NAME)/pkg/buildinfo.GitRevision=$(GIT_COMMIT) \
-			-B 0x$(BUILD_ID)
+			-B 0x$(BUILD_ID) \
+			$(RELEASE_LDFLAGS)"
```
rene-dekker, line 33: Please remove the quote from this line.

node/Makefile
```diff
@@ -153,7 +153,8 @@ DATE:=$(shell date -u +'%FT%T%z')
 LDFLAGS= -X $(PACKAGE_NAME)/pkg/lifecycle/startup.VERSION=$(GIT_VERSION) \
 	-X $(PACKAGE_NAME)/buildinfo.GitVersion=$(GIT_DESCRIPTION) \
 	-X $(PACKAGE_NAME)/buildinfo.BuildDate=$(DATE) \
-	-X $(PACKAGE_NAME)/buildinfo.GitRevision=$(GIT_COMMIT)
+	-X $(PACKAGE_NAME)/buildinfo.GitRevision=$(GIT_COMMIT) \
+	$(RELEASE_LDFLAGS)"
```
rene-dekker, line 157: Please remove the quote from this line too. To make our build targets consistent, we want to remove quotes in any of the LDFLAGS variable declarations, and then place the quotes around the LDFLAGS variable inside of the build command.

afshin-deriv: Done 👍

caseydavenport:
Is this still in progress?

PR #6703, opened 2022-09-14
Optimise KeyFromDefaultPath.

fasaxc:
## Description

The old version was linear, trying one key type after another. As the number of key types has grown this now causes noticable CPU usage when benchmarking. This commit reworks the algorithm to parse keys using a tree of switch statements and other conditions.

I considered making this a bit more generic but there are a lot of exceptions and special cases (for example, Felix config keys allow slashes in the name of the key), and, this should be a one-off since all new resources will go through the "resources/v3" path.

Add a Fuzz test that compares with the old implementation.
## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Reduce parsing overhead when parsing key/value pairs from Typha.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/model/block_affinity.go
+++ b/libcalico-go/lib/backend/model/block_affinity.go
@@ -100,6 +100,10 @@ func (options BlockAffinityListOptions) KeyFromDefaultPath(path string) Key {
 	}
 	cidrStr := strings.Replace(r[0][2], "-", "/", 1)
 	_, cidr, _ := net.ParseCIDR(cidrStr)
+	if cidr == nil {
+		log.Debugf("Failed to parse CIDR in block affinity path: %q", path)
+		return nil
+	}
 	host := r[0][1]
 
 	if options.Host != "" && options.Host != host {
--- a/libcalico-go/lib/backend/model/keys.go
+++ b/libcalico-go/lib/backend/model/keys.go
@@ -135,10 +135,6 @@ func KeyToDefaultDeletePath(key Key) (string, error) {
 	return key.defaultDeletePath()
 }
 
-func KeyToValueType(key Key) (reflect.Type, error) {
-	return key.valueType()
-}
-
 // KeyToDefaultDeleteParentPaths returns a slice of '/'-delimited
 // paths which are used to delete parent entries that may be auto-created
 // by directory-based KV stores (e.g. etcd v3).  These paths should also be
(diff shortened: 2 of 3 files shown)


libcalico-go/lib/backend/model/block_affinity.go
```diff
@@ -100,6 +100,10 @@ func (options BlockAffinityListOptions) KeyFromDefaultPath(path string) Key {
 	}
 	cidrStr := strings.Replace(r[0][2], "-", "/", 1)
 	_, cidr, _ := net.ParseCIDR(cidrStr)
+	if cidr == nil {
+		log.Debugf("Failed to parse CIDR in block affinity path: %q", path)
```
fasaxc, line 104: Fuzzer found a latent bug

libcalico-go/lib/backend/model/keys.go
```diff
@@ -193,18 +189,257 @@ func IsListOptionsLastSegmentPrefix(listOptions ListInterface) bool {
...
+// OldKeyFromDefaultPath is the old, (slower) implementation of KeyFromDefaultPath.  It is kept to allow
+// fuzzing the new version against it.  Parses the default path representation of a key into one
+// of our <Type>Key structs.  Returns nil if the string doesn't match one of
+// our key types.
+func OldKeyFromDefaultPath(path string) Key {
 	if m := matchWorkloadEndpoint.FindStringSubmatch(path); m != nil {
 		log.Debugf("Path is a workload endpoint: %v", path)
 		return WorkloadEndpointKey{
-			Hostname:       m[1],
+			Hostname:       unescapeName(m[1]),
```
fasaxc, line 434: I think the old code was bugged here: we apply encoding when converting a key to a path but not the other way around.  In practice, validation prevents encoding from being needed but, given a bad key added to etcd we'd be unable to round trip it with the old code.

libcalico-go/lib/backend/model/keys.go
```diff
@@ -213,7 +448,11 @@ func KeyFromDefaultPath(path string) Key {
 			Name: unescapeName(m[1]),
 		}
 	} else if m := matchGlobalResource.FindStringSubmatch(path); m != nil {
-		ri := resourceInfoByPlural[unescapeName(m[1])]
+		ri, ok := resourceInfoByPlural[unescapeName(m[1])]
+		if !ok {
+			log.Warnf("(BUG) unknown resource type: %v", path)
```
fasaxc, line 453: Fuzzer hit a panic here if the resource was unknown.

libcalico-go/lib/backend/model/keys_test.go
```diff
@@ -335,3 +456,53 @@ func mustParseCIDR(s string) net.IPNet {
...
+		EndpointID: "eth0",
+	},
+	ResourceKey{
+		Name:      "projectcalico-default-allow",
+		Namespace: "default",
+		Kind:      apiv3.KindNetworkPolicy,
+	},
+}
+
+func BenchmarkOldKeyFromDefaultPath(b *testing.B) {
```
fasaxc, line 481: Benchmark shows a 4-5x improvement in perf for the mix of keys above.

libcalico-go/lib/backend/model/keys_test.go
```diff
@@ -15,17 +15,139 @@
...
+		if newKey == nil {
+			t.Errorf("Old KeyFromDefaultPath(%q)=%v but new KeyFromDefaultPath() returned nil", path, oldKey)
+			return
+		}
+
+		func() {
+			if !safeKeysEqual(oldKey, newKey) {
+				t.Errorf("%q -> (new output) %v != (old output) %v", path, newKey, oldKey)
+			}
+		}()
```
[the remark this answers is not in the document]
fasaxc: Nothing! I had a defer in there that I used to debug a panic and forgot to remove the func after removing the defer.

neiljerram (commented):
Some queries

neiljerram:
Also, in order to actually execute the fuzz testing, I think we need to add `-fuzz` to a go test invocation somewhere.

libcalico-go/lib/backend/model/keys_test.go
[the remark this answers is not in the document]
fasaxc: Was using it for debugging the test (had a defer in there).  Can be removed now

libcalico-go/lib/backend/model/keys_test.go
```diff
@@ -15,17 +15,137 @@
...
+	"/calico/ipam/v2/host/foobar/ipv4/block/1.2.3.4-5",
+	"calico/ipam/v2/host/0/ipv0/block/0",
+	"/calico/felix/v1/host",
+	"/calico/felix/v1/host/foo/endpoint/bar",
+	"/calico/felix/v1/endpoint",
+	"/calico/felix/v2/foo/host",
+	"/calico/felix/v2//foo/host",
+	"/calico/felix/v2/region-Europe/host/h1/status",
+	"/calico/resources/v3/projectcalico.org/foo/bar/baz",
+	"/bar/v1/host/foobar/workload/open%2fstack/work%2fload/endpoint/end%2fpoint",
```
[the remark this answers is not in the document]
fasaxc: They cover URL encoding

libcalico-go/lib/backend/model/block_affinity.go
fasaxc: I think you could hit this by manually creating a bad entry in etcd (in an etcd system).  It is UTed by the fuzz entry (all the listed Fuzz seed entries are run as unit tests).

libcalico-go/lib/backend/model/keys.go
fasaxc: I changed it so the Fuzzer would stop finding this problem; couldn't think of a good way to avoid that issue in the fuzz test.

libcalico-go/lib/backend/model/keys_test.go
```diff
@@ -15,17 +15,137 @@
...
+			// Old version couldn't parse this, so we don't care if the new one can.
+			return
+		}
+		if newKey == nil {
+			t.Errorf("Old KeyFromDefaultPath(%q)=%v but new KeyFromDefaultPath() returned nil", path, oldKey)
+			return
+		}
+
+		if !safeKeysEqual(oldKey, newKey) {
+			t.Errorf("%q -> (new output) %v != (old output) %v", path, newKey, oldKey)
```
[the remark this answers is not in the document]
fasaxc: `Errorf` doesn't return on its own.  I'll swap it for `Fatalf`, which terminates the test.

(I was only returning in some places because those places were guarding further logic.)

libcalico-go/lib/backend/model/keys_test.go
```diff
@@ -15,17 +15,137 @@
...
+		}
+
+		serialised, err := KeyToDefaultPath(newKey)
+		if err != nil {
+			t.Errorf("Failed to reserialise %q -> %v -> %s", path, newKey, err)
+			return
+		}
+		if len(path) > 0 && path[0] != '/' {
+			serialised = serialised[1:]
+		}
```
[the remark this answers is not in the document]
fasaxc: This is checking that the two implementations behave the same with respect to leading slash.  either both rejecting it or both allowing it.

The confusion comes from the early days of `etcdv1`, where paths were "directory like" so `foo/bar` was equivalent to `/foo/bar`.  For v3 paths we require the leading slash.  

Do we need to maintain these corner cases?  Probably not but I'm not 100% sure that there's no stray path where it might matter.  The fuzz test tends to find them so I tried to be consistent to get the most bang for buck from the fuzz test.

fasaxc:
> Also, in order to actually execute the fuzz testing, I think we need to add -fuzz to a go test invocation somewhere.

To actually run the fuzzer we need some new infra but the test already runs, just as a static unit test that executes all the canned patterns (so the test can't rot).

Fuzzing shouldn't be done as part of normal CI because it may find new bugs in old code at any time so it's like adding a flake factory.  The best way to run it is as a side job that runs for hours at a time and raises tickets to be looked at later.

libcalico-go/lib/backend/model/keys_test.go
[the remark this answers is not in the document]
fasaxc: :face_with_spiral_eyes: got a bug in that condition, should be `==`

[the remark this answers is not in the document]
fasaxc: Have rewritten this bit and added comments.

libcalico-go/lib/backend/model/block_affinity.go
neiljerram: So just to check that I'm understanding: the fuzzer found a bug and told you the input needed to hit that, and then you added that input as one of the entries in `interestingPaths` ?

libcalico-go/lib/backend/model/keys.go
neiljerram: OK, agree that that makes sense.

libcalico-go/lib/backend/model/keys_test.go
```diff
@@ -15,17 +15,163 @@
...
+		newKey2 := KeyFromDefaultPath(serialised)
+		oldKey2 := OldKeyFromDefaultPath(serialised)
+		if !safeKeysEqual(newKey2, oldKey2) {
+			t.Fatalf("%q -> (new output) %v != (old output) %v", serialised, newKey2, oldKey2)
+		}
+		if newKey2 != nil && !safeKeysEqual(newKey, newKey2) {
+			t.Fatalf("%q -> %v but %q -> %v", path, newKey, serialised, newKey2)
+			return
```
neiljerram, line 155: nit: unnecessary return?

libcalico-go/lib/backend/model/keys_test.go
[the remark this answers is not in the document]
neiljerram: But now we have only `Fatalf`s, so all of the cases are terminating, IIUC.

neiljerram (approved):
LGTM.  I left a couple more comments to double check my understanding, but I think it's all good.  Also see the point now about not doing `-fuzz` in standard CI.

libcalico-go/lib/backend/model/block_affinity.go
fasaxc: Yes, exactly.  If we were running fuzz tests nightly, we'd get them to save their own corpus to somewhere persistent.  Then, any "previously failed" entries int he corpus are run first to make the tests deterministically fail.  Since we're not doing that, copying the bad entry back into the seed corpus seemed like a good stop gap.

PR #6705, opened 2022-09-14
Replace encoding/json with jsoniter.

fasaxc:
## Description

Using jsoniter's drop-in replacement mode to speed up our JSON encoding/decoding.  Seems to be a simple win; cuts the time to decode a workload endpoint in half, for example.  This has a significant impact on parsing the Typha protocol.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico now uses a faster JSON parsing library; this reduces CPU load and improves start-up latency.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/net/mac.go
+++ b/libcalico-go/lib/net/mac.go
@@ -15,8 +15,9 @@
 package net
 
 import (
-	"encoding/json"
 	"net"
+
+	"github.com/projectcalico/calico/libcalico-go/lib/json"
 )
 
 // Sub class net.HardwareAddr so that we can add JSON marshalling and unmarshalling.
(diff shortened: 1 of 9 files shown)


libcalico-go/lib/net/mac.go
```diff
@@ -26,12 +27,14 @@ type MAC struct {
 
 // MarshalJSON interface for a MAC
 func (m MAC) MarshalJSON() ([]byte, error) {
+	var json = jsoniter.ConfigCompatibleWithStandardLibrary
```
caseydavenport, line 30: Would it make sense to have a libcalico-go package to wrap this up?

```go
package json

jsoniter "github.com/json-iterator/go"


func Marshal(s string) []byte, error {
  return jsoniter.ConfigCompatibleWithStandardLibrary.Marshal(s)
}
```

Would turn this into an import change rather than an import + local variable change. Unless you think there's a future where we want to use different jsoniter implementations in different places?

marvin-tigera:

caseydavenport (approved):

PR #6741, opened 2022-09-19
Tweak TLS config wherever used and make it FIPS compliant.

rene-dekker:
- Apiserver will always use the recommended ciphers to make it fips compliant, the operator will add the tls-max-version flag.
- Replace tls.Config creation with our own convenience function that sets recommended settings.
- Go mod uses our fork of k8s.io/apiserver, which is identical to v0.24.0, with an added flag. We will submit a PR upstream to get this into Kubernetes.

### DIFF
--- a/crypto/pkg/tls/tls.go
+++ b/crypto/pkg/tls/tls.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2022 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tls
+
+import (
+	"crypto/tls"
+)
+
+// NewTLSConfig returns a tls.Config with the recommended default settings for Calico.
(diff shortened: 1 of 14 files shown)


crypto/pkg/tls/tls.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"crypto/tls"
+)
+
+// NewTLSConfig returns a tls.Config with the recommended default settings for Calico Enterprise components.
+// Read more recommendations here:
+// https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4095
+func NewTLSConfig(fipsMode bool) *tls.Config {
+	cfg := &tls.Config{
+		MinVersion: tls.VersionTLS12,
```
rene-dekker, line 26: Is there any reason to believe that we should still be supporting old (and known to be vulnerable) TLS versions?

crypto/pkg/tls/tls.go
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tls
+
+import (
+	"crypto/tls"
+)
+
+// NewTLSConfig returns a tls.Config with the recommended default settings for Calico Enterprise components.
```
mgleung, line 21: nit: Should this be `Calico Enterprise` or `Calico` here?

mgleung (approved):
Just one nit on a comment, otherwise LGTM

crypto/pkg/tls/tls.go
mgleung: I don't think so since this is just used by our components. It's possible you might need it for an upgrade, but I'm not exactly sure how the upgrade logic works.

marvin-tigera:

marvin-tigera:

marvin-tigera:

PR #6743, opened 2022-09-20
Deletes unused content in `getting-started/clis` and `networking`

ctauchen:
This PR removes some stale content.
* `getting-started/clis` isn't linked in the navigation. It's duplicated in `maintenance/clis`. Content is different, with updates seeming to go into `maintenance/clis` and not `getting-started/clis`.
* `networking/enabling-ipvs.md` isn't linked in the navigation. Its content is replicated in the `networking/configuring/use-ipvs.md` file.

But I may be missing something. 

@bmckercher123 @caseydavenport


bmckercher123:
My guess is IPVS is a result of merge from OS to CE.  Unless work was done that I did not know about, IPVS is OS only doc. So its appearance in CE is an artifact of a doc merge without a parallel TOC entry. So if Casey verifies this scenario, you can delete it.

bmckercher123:
Yes, it looks like the CLI content under /getting-started is duplicate of /maintenance. I believe it is safe to delete.

caseydavenport:
Awesome, thanks for spotting this and fixing it up!

caseydavenport (approved):

Issue #6522, opened 2022-08-10
HostPort of pod/container can't access from other nodes in eBPF mode 


gongzh:
After enabling eBPF mode of calico, then I created a pod which has a hostPort setting, what I expected is the clients from every node and outside of cluster can access hostIP:hostPort, but the result is: it can be accessed only from the host where the pod is located, but can't be accessed from other nodes or outside of cluster.

## Expected Behavior
the clients from every node and outside of cluster can access hostIP:hostPort

## Current Behavior
it can be accessed only from the host where the pod is located.
it can't be accessed from other nodes or outside of cluster. when I use "curl hostIP:hostPort", it just hang.

## Steps to Reproduce (for bugs)
1. stop kube-proxy on every nodes
2. restart linux
3. enable ebpf dataplane:
`calicoctl patch felixconfiguration default --patch='{"spec": {"bpfEnabled": true}}' `
4. recreate all calico pods and test pods.

## Context
"curl PodIP:containerPort" is always successful from other nodes.

## Your Environment
* Calico version: v3.23.3
* Orchestrator version: kubernetes 1.23.9
* Operating System and version: CentOS 8.2
* Link to your project (optional):

caseydavenport:
@tomastigera

seh:
See [today's discussion in the "networking" channel](https://calicousers.slack.com/archives/CPEPF833L/p1661868642674109) of the "Calico Users" Slack team for a similar report.

In my case, I'm also using Calico version 3.23.3, Kubernetes 1.24.4, Flatcar Container Linux 3277.1.1, provisioned by _kOps_, running in AWS EC2.

Using _tcpdump_, I can see the following packet flow, with the client designated as _C_, the target container in the pod as _T_, and the EC2 instance hosting the container _H_. The container's requested host port is _hp_; its corresponding container port is _cp_.
- _C_:* → _H_:_hp_ on _eth0_ [_SYN_]  
  This is the incoming packet from the client, trying to open the connection.
- _C_:* → _T_:_cp_ on _cali*_ [_SYN_]  
  This packet is forwarded via DNAT.
- _T_:_cp_ → _C_:* on _cali*_ [_SYN-ACK_]  
  This is the container responding to the connection request.

There are no more packets exchanged in that conversation. After between one half and a full second later, the client sends the opening packet again, and the same three-packet pattern repeats.

[...]

PR #6766, opened 2022-09-26
BPF support for 3rd party iptables DNAT rules that redirect host->pod

mazdakn:
## Description
Add support for 3rd party translation of host addresses to workloads using DNAT rules in iptables.
This should fix this issue: https://github.com/projectcalico/calico/issues/6522

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
```

## Reminder for the reviewer

Other optional labels:


felix/fv/bpf_notracking_test.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+
+		infra.Stop()
+	})
+
+	It("should allow 3rd party DNAT to workloads work", func() {
+		expectNormalConnectivity := func() {
+			cc.ResetExpectations()
+			for i := range felixes {
+				cc.ExpectSome(felixes[i], workloads[0].Port(80))
```
tomastigera, line 106: I think the problem was more like workload->host then redirected to workload by host.

So I think there should be first Expect(None, workloads[0], felixs[i]. port) then Expect(Some...) after the rules are added then Expect(None,...) again.

I wonder if it would be simpler to add this in the bpf_test.go as we can then run it in various overlay scenarios.

felix/fv/bpf_notracking_test.go
```diff
@@ -0,0 +1,156 @@
...
+			expectNormalConnectivity()
+		})
+
+		By("installing 3rd party rules", func() {
+			// Install a DNAT in first felix
+			felix := felixes[0]
+
+			felix.Exec(
+				"iptables", "-t", "nat", "-A", "PREROUTING", "-p", "tcp", "-m", "tcp",
+				"--dport", "8080", "-j", "DNAT", "--to-destination", "10.65.0.2:80")
```
tomastigera, line 127:
```suggestion
				"--dport", "8080", "-j", "DNAT", "--to-destination", workloadX+portX)
```

tomastigera (commented):
I think this is not testing and fixing the issues where we first got SYN from outside te host which we tracked and then on the way back we set NOTRACK. It is not just about setting the SKIP_FIB mark. We have to set it when we know we need to set it.

felix/fv/bpf_notracking_test.go
mazdakn: This line is just to test if normal connectivity is OK, and the change is not breaking something that should always work.
The test that is expected to change is 
```hostIP := connectivity.TargetIP(felixes[i].IP)
   cc.ExpectNone(externalClient, hostIP, 8080)```

Comment (tomastigera) [reply felix/fv/bpf_notracking_test.go:106]: > This line is just to test if normal connectivity is OK, and the change is not breaking something that should always work.

Don't the other tests take care of that. Perhaps it would make sense to add this _after_ the DNAT rule is injected?

I think we are really interested here only in pod->host or extclient->host cases. Those are the cases that do not work. And afaict, they are not supposed to work before the the DNAT rule is injected because there should be no host networked backend (ideally).

Comment (mazdakn) [reply felix/fv/bpf_notracking_test.go:106]: Ok, I removed those tests.

File: felix/fv/bpf_notracking_test.go [hunk 3]
```diff
@@ -0,0 +1,125 @@
...
+
+		for _, felix := range felixes {
+			felix.Stop()
+		}
+
+		infra.Stop()
+	})
+
+	It("should allow 3rd party DNAT to workloads work", func() {
+		expectNormalConnectivity := func() {
```
tomastigera, line 86: I am not sure whether this is needed. We expect that. There is no application listening on that port on the host. So it would be a total surprise if it worked. And we are not quite testing that DNAT in iptables work, we trust that iptables work and are tested :)

felix/fv/bpf_notracking_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+			felix.Exec(
+				"iptables", "-t", "nat", "-A", "PREROUTING", "-p", "tcp", "-m", "tcp",
+				"--dport", "8080", "-j", "DNAT", "--to-destination", target)
+
+			cc.ResetExpectations()
+			hostIP0 := connectivity.TargetIP(felixes[0].IP)
+			cc.ExpectSome(felixes[1], hostIP0, 8080)
+			cc.ExpectSome(externalClient, hostIP0, 8080)
+			cc.ExpectSome(workloads[1], hostIP0, 8080)
```
tomastigera, line 113: This is the key bit :+1:  This should not work without your fix. If you had this test as an independent commit followed by the fix commit. Running the test commit should fail and running the fix commit should pass. That is what we are testing

felix/fv/bpf_notracking_test.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		for _, felix := range felixes {
+			felix.Stop()
+		}
+
+		infra.Stop()
+	})
+
+	It("should allow 3rd party DNAT to workloads work", func() {
```
tomastigera, line 85:
```suggestion
	It("should have connectivity when DNAT redirects to-host traffic to a local pod.", func() {
```

felix/rules/static.go
```diff
@@ -1122,8 +1122,14 @@ func (r *DefaultRuleRenderer) StaticBPFModeRawChains(ipVersion uint8,
 
 	rawRules = append(rawRules,
 		Rule{
-			Match:  Match().NotDestAddrType(AddrTypeLocal),
-			Action: GotoAction{Target: ChainRawUntrackedFlows},
+			Match:   Match().DestAddrType(AddrTypeLocal),
+			Action:  SetMaskedMarkAction{Mark: tcdefs.MarkSeenSkipFIB, Mask: tcdefs.MarkSeenSkipFIB},
+			Comment: []string{"To support 3rd party iptables managers"},
```
tomastigera, line 1127:
```suggestion
			Comment: []string{"Mark traffic towards the host - it is TRACKed"},
```

felix/rules/static.go
```diff
@@ -1122,8 +1122,14 @@ func (r *DefaultRuleRenderer) StaticBPFModeRawChains(ipVersion uint8,
...
+		},
+		Rule{
+			Match:   Match().NotDestAddrType(AddrTypeLocal),
+			Action:  GotoAction{Target: ChainRawUntrackedFlows},
+			Comment: []string{"Jump to target for packets with not local destination address"},
```
tomastigera, line 1132:
```suggestion
			Comment: []string{"Check if forwarded traffic needs to be TRACKed"},
```

felix/bpf-gpl/tc.c
```diff
@@ -773,6 +773,10 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct cali_tc_ctx *ctx
 		if (CALI_F_FROM_WEP && state->flags & CALI_ST_SKIP_FIB) {
 			ct_ctx_nat.flags |= CALI_CT_FLAG_SKIP_FIB;
 		}
+		// To support iptables DNAT from 3rd party iptables managers
```
tomastigera, line 776:
```suggestion
		/* Packets received at WEP with CALI_CT_FLAG_SKIP_FIB mark signal
		 * that all traffic on this connection must flow via host namespace as it was
		 * originally meant for host, but got redirected to a WEP by a 3rd party DNAT rule.
		 */
```

felix/bpf/ut/icmp_test.go
```diff
@@ -79,7 +79,7 @@ func TestICMPCTPlain(t *testing.T) {
 			Expect(k.PortB()).To(Equal(uint16(0)))
 		}
 	})
-	expectMark(tcdefs.MarkSeen)
+	expectMark(tcdefs.MarkSeenSkipFIB)
```
tomastigera, line 82: I do not see how the change should modify this. We may have unclean conntrack at the start of this test or there is a bug.

tomastigera (commented):
LGTM, just some comment clarification and the UT change is suspicious, likely an info leak in the tests.

felix/bpf/ut/icmp_test.go
mazdakn: Fixed. It happened because `CALI_SKB_MARK_SEEN & CALI_SKB_MARK_SKIP_FIB` is equal to `CALI_SKB_MARK_SEEN` which made the whole condition True. It's fixed now.

felix/fv/bpf_notracking_test.go
mazdakn: I agree, but without this we cannot be sure if the actual test is passing by chance or because of the fix. Do you prefer to remove it?

felix/rules/static.go
```diff
@@ -1121,16 +1121,22 @@ func (r *DefaultRuleRenderer) StaticBPFModeRawChains(ipVersion uint8,
 	}
 
 	rawRules = append(rawRules,
-		Rule{
-			Match:  Match().NotDestAddrType(AddrTypeLocal),
-			Action: GotoAction{Target: ChainRawUntrackedFlows},
-		},
 		Rule{
 			// Return, i.e. no-op, if bypass mark is not set.
```
tomastigera, line 1125: whyis this the first rule now? Does it make any difference?

tomastigera (approved):
LGTM, just one question to double check.

felix/rules/static.go
mazdakn: Because this rule uses `0xffffffff` for mask and we change the mark the newly added rule. If we change the mark in with `SKipFIB` then this rule always fails and for that `doNotTrack` policies won't work. 
At the same time, the next two rules are relevant so I put them near each others.

PR #6771, opened 2022-09-27
Allow dnsmasq to start when there's a race with no ports

neiljerram:
## Description

https://github.com/projectcalico/calico/commit/9a8a6c9bc2bd083c95c8adf209a74c4ac2506fc8 introduced a
race by decoupling the threads ("A") that can update the network model from the thread ("B") that
restarts dnsmasq with configuration for the network model.

The result is that:
- thread A can say "update dnsmasq for network X, which definitely has at least one port"
- this means that a trigger is queued for thread B
- thread A handles an endpoint deletion which means that network X now has no ports
- thread B runs and generates invalid dnsmasq options because there are now no ports.

It would be fine if that happened just once, but this Neutron code causes the situation to persist
for 5 minutes:

    def enable(self):
        """Enables DHCP for this network by spawning a local process."""
        try:
            common_utils.wait_until_true(self._enable, timeout=300)
        except common_utils.WaitTimeout:
            LOG.error("Failed to start DHCP process for network %s",
                      self.network.id)

It would still be fine if another thread updated the network concerned during those 5 minutes, and
if the updated network object was the same object as the one that the `enable` thread has a
reference to.  However there are a couple of cases where our DHCP agent deliberately creates a new
network object for an existing network ID:

1. When resyncing against the data store, which happens every 10 minutes.

2. If it sees an endpoint with a new subnet.

In those cases the `enable` thread keeps referencing the old network object with no ports, and hence
it blocks actual dnsmasq updating for 5 minutes.

So, what to do about this?  Possibly the best fix would be to take a deep copy of the network model
when passing it to the dnsmasq update thread.  But unfortunately the model (aka NetworkCache) object
appears to have no support for that.  Or perhaps we could add more locking, but that feels complex
and risky.  Instead, for now, let's just ensure that the dnsmasq configuration is valid even when
there are unexpectedly no ports on the network.  Then dnsmasq should start up, and `enable` return
and unblock the update thread, and shortly after there will be a further trigger that either stops
dnsmasq or restarts it again with correct config.

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/agent/dhcp_agent.py
+++ b/networking-calico/networking_calico/agent/dhcp_agent.py
@@ -260,7 +260,7 @@ class DnsmasqUpdater(object):
 
     def start(self):
         while True:
-            LOG.info("DnsmasqUpdater: wait until updates needed")
+            LOG.debug("DnsmasqUpdater: wait until updates needed")
             dirty_network_ids = set()
             dirty_network_ids.add(self.updates_needed.get())
             try:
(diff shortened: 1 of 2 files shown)


fasaxc (approved):
Good explanation, thanks!

PR #6783, opened 2022-09-28
felix/bpf: allow using CTLB workaround just for UDP

tomastigera:
Issues described in https://github.com/projectcalico/calico/issues/4509 are only UDP related. We need to turn off CTLB for UDP. TCP would resolve broken connections as the original pod would not exist anymore and would make a new connection. Therefore it does not need to take the performance hit associated with disabling CTLB.

## Related issues/PRs

refs https://github.com/projectcalico/calico/issues/4509 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: we can apply the CTLB-turned-off workaround just to UDP
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -228,7 +236,7 @@ type bpfEndpointManager struct {
 	rpfStrictModeEnabled string
 
 	// Service routes
-	ctlbWorkaroundEnabled bool
+	ctlbWorkaroundEnabled ctlbWorkaroundState
```
fasaxc, line 239:
```suggestion
	ctlbWorkaroundMode ctlbWorkaroundMode
```

fasaxc: "Enabled" has connotations of being a bool so I think it's worth changing to say Mode.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -2067,10 +2080,14 @@ func (m *bpfEndpointManager) getInterfaceIP(ifaceName string) (*net.IP, error) {
...
 
 func (m *bpfEndpointManager) onServiceUpdate(update *proto.ServiceUpdate) {
-	if !m.ctlbWorkaroundEnabled {
+	if m.ctlbWorkaroundEnabled == ctlbWorkaroundDisabled {
 		return
 	}
 
+	// XXX It would be great to exclude other than UDP services if
+	// m.ctlbWorkaroundEnabled == ctlbWorkaroundUDPOnly but ServiceUpdate does
+	// not carry the protocol :(
```
fasaxc, line 2089: If it's in the k8s Service object you can add this trivially.  See dataplane_passthru.go.  It's literally just copying the fields we need from the k8s Service to our own protobuf struct (which you can add fields to by editing felixbackend.proto) :-)

felix/fv/test-connection/test-connection.go
```diff
@@ -148,6 +149,16 @@ func main() {
 		log.WithError(err).Fatal("Invalid --stdin")
 	}
 
+	timeout := 0
+
+	if toval := arguments["--timeout"]; toval != nil {
+		timeout, err = strconv.Atoi(toval.(string))
+		if err != nil {
+			// panic on error
+			log.WithField("timeout", timeout).Fatal("Invalid duration argument")
```
fasaxc, line 158:
```suggestion
			log.WithField("timeout", timeout).Fatal("Invalid --timeout argument")
```
It's explicitly not a time.Duration here so probably best not to use that word

felix/fv/test-connection/test-connection.go
```diff
@@ -420,18 +432,46 @@ func (tc *testConn) tryLoopFile(loopFile string, logPongs bool) error {
...
-		if logPongs {
-			fmt.Println("PONG")
+		var err error
+		respRaw, err = tc.protocol.Receive()
+		if err == nil {
+			if logPongs {
+				fmt.Println("PONG")
+			}
+			retryStart = zeroTime
+		} else if e, ok := err.(net.Error); ok && e.Timeout() {
```
fasaxc, line 463: Just FYI, I think there's `os.IsTimeout(err)` as a helper for this check

fasaxc (changes requested):
Just some minor feedback on naming etc.

Is/should the feature flag be documented somewhere?

fasaxc (approved):

Issue #5594, opened 2022-02-09
[CRD backend] Syncing Calico node labels cannot be disabled


invidian:
## Expected Behavior
When `kubecontrollersconfiguration.spec.controllers.node.syncLabels` is set to `Disabled`, Calico `Node` objects do not get updated with new labels.

## Current Behavior
Regardless of the value of `kubecontrollersconfiguration.spec.controllers.node.syncLabels`, labels for nodes are always synced to Calico `Node` objects.

## Steps to Reproduce (for bugs)

1. Install Calico.
2. `kubectl patch --type merge kubecontrollersconfiguration default --patch='{"spec": {"controllers": {"node": {"syncLabels": "Disabled"}}}}'`
3. `kubectl label node x foo=bar`
4. `calicoctl get node x | grep 'foo: bar'`

## Context

Findings around #5593 and #5592.

## Your Environment

* Calico version v3.22.0
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes 1.23

caseydavenport:
@invidian are you running with Calico in etcd mode or using Kubernetes CRDs?

invidian:
Using Kubernetes CRDs I think:
```yaml
  environmentVars:
    DATASTORE_TYPE: kubernetes
    ENABLED_CONTROLLERS: node
```

caseydavenport:
Yep, so I think that is the confusion here, and not sure how to resolve it really.

When using that mode, the Kubernetes Node API == the Calico node API. In other words, there's no controller syncing the two resources, they are one in the same, and so there's no way to disable syncing of the labels.

For etcd mode, there are two data stores in play, which means we run a controller to keep them in sync. The proper fix is most likely to validate against setting `syncLabels: Disabled` for CRD mode, since functionally that's the only behavior supported.

invidian:
Now that you asked and explained, that all make sense, thanks!

>For etcd mode, there are two data stores in play, which means we run a controller to keep them in sync. The proper fix is most likely to validate against setting syncLabels: Disabled for CRD mode, since functionally that's the only behavior supported.

The setting option could also be renamed with an `etcd` prefix or something.

Or even documentation could mention that this option is only applicable for etcd data store. Or perhaps controller could generate a warning log on this settings?

As I suppose the validation of that would not be trivial since it's about 2 separate keys in the CR, so it would require a validating webhook, which is non trivial to deploy right now with k8s from my experience.

caseydavenport:
[...]

PR #6786, opened 2022-09-28
Fix syncing calico node labels for Kubernetes CRD mode

huiyizzz:
## Description

For Kubernetes CRD mode, sync labels should always be set to enabled.
This PR adds a validation rule for `syncLabels`. When the datastore type is Kubernetes, it does not allow disabling sync labels.
## Related issues/PRs

https://github.com/projectcalico/calico/issues/5594
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fixed SyncLabels validation for Kubernetes datastore.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/k8s/resources/kubecontrollersconfig.go
+++ b/libcalico-go/lib/backend/k8s/resources/kubecontrollersconfig.go
@@ -15,13 +15,15 @@
 package resources
 
 import (
+	"errors"
 	"reflect"
 
-	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
-
+	log "github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+
+	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 )
 
 const (
@@ -43,5 +45,20 @@ func NewKubeControllersConfigClient(c *kubernetes.Clientset, r *rest.RESTClient)
 		},
 		k8sListType:  reflect.TypeOf(apiv3.KubeControllersConfigurationList{}),
 		resourceKind: apiv3.KindKubeControllersConfiguration,
+		validator:    kubeControllersConfigValidator{},
 	}
+}
+
+type kubeControllersConfigValidator struct{}
+
+func (v kubeControllersConfigValidator) Validate(res Resource) error {
+	config := res.(*apiv3.KubeControllersConfiguration)
+	if node := config.Spec.Controllers.Node; node != nil {
+		log.Debugf("Validate SyncLabels for Kubernetes datastore type: %s", node.SyncLabels)
+		if node.SyncLabels == apiv3.Disabled {
+			log.Debugf("SyncLabels value cannot be set to disabled with Kubernetes datastore driver.")
+			return errors.New("invalid SyncLabels value")
+		}
+	}
+	return nil
 }
--- a/libcalico-go/lib/backend/k8s/resources/customresource.go
+++ b/libcalico-go/lib/backend/k8s/resources/customresource.go
@@ -71,12 +71,20 @@ type customK8sResourceClient struct {
 	// versionConverter is an optional hook to convert the returned data from the CRD
 	// from one format to another before returning it to the caller.
 	versionconverter VersionConverter
+
+	// validator used to validate resources.
+	validator Validator
 }
 
 // VersionConverter converts v1 or v3 k8s resources into v3 resources.
 // For a v3 resource, the conversion should be a no-op.
 type VersionConverter interface {
 	ConvertFromK8s(Resource) (Resource, error)
+}
+
+// Validator validates a resource.
+type Validator interface {
+	Validate(Resource) error
 }
 
 // Create creates a new Custom K8s Resource instance in the k8s API from the supplied KVPair.
@@ -93,6 +101,14 @@ func (c *customK8sResourceClient) Create(ctx context.Context, kvp *model.KVPair)
 	if err != nil {
 		logContext.WithError(err).Debug("Error creating resource")
 		return nil, err
+	}
+
+	// Validate the resource if the Validator is defined.
+	if c.validator != nil {
+		if err = c.validator.Validate(resIn); err != nil {
+			logContext.WithError(err).Debug("Error creating resource")
+			return nil, err
+		}
 	}
 
 	// Send the update request using the REST interface.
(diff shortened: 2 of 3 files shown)


api/go.sum
```diff
@@ -166,6 +166,7 @@ github.com/google/pprof v0.0.0-20201203190320-1bf35d6f28c2/go.mod h1:kpwsk12EmLe
 github.com/google/pprof v0.0.0-20210122040257-d980be63207e/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/pprof v0.0.0-20210226084205-cbba55b83ad5/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/renameio v0.1.0/go.mod h1:KWCgfxg9yswjAJkECMjeO8J8rahYeXnNhOm40UhjYkI=
+github.com/google/uuid v1.1.2 h1:EVhdT+1Kseyi1/pUmXKaFxYsDNy9RQYkMWRH68J/W7Y=
```
coutinhop, line 169: The changes to this file don't seem related to your other changes, are they necessary for this PR? (or are they just from running some `go mod` command?)

Could you try reverting these and testing if everything works without this file's changes?

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -445,6 +447,17 @@ func validateLabels(fl validator.FieldLevel) bool {
 	return true
 }
 
+func validateSyncLabels(fl validator.FieldLevel) bool {
+	s := fl.Field().String()
+	log.Debugf("Validate SyncLabels: %s", s)
+	dst := os.Getenv("DATASTORE_TYPE")
+	if s == "Disabled" && dst == "kubernetes" {
```
coutinhop, line 454: I'm not sure if it would be too extreme (@mgleung let me know if it is), but could we maybe use `strings.ToLower()` in order to still catch cases where, for example, DATASTORE_TYPE is "Kubernetes"?
(if we do go ahead with this, please remember to add test cases for it in validator_test.go)

coutinhop (changes requested):
Overall LGTM, just a couple of questions/comments...

Thanks!

libcalico-go/lib/validator/v3/validator.go
mgleung: I just had a chat with @huiyizzz and we'll try a different approach. I don't think that we can guarantee that the env variable will be set each time for every component that calls this so we'll try something that doesn't depend upon this value.

mgleung: `strings.ToLower()` would be great to use here though normally to ensure we're matching properly though.

libcalico-go/lib/backend/k8s/resources/kubecontrollersconfig.go
```diff
@@ -30,7 +43,7 @@ const (
 )
 
 func NewKubeControllersConfigClient(c *kubernetes.Clientset, r *rest.RESTClient) K8sResourceClient {
-	return &customK8sResourceClient{
+	return &kubeControllersConfigClient{
```
caseydavenport, line 46: @huiyizzz rather than reimplement all of `customK8sResourceClient`, would it be possible to add a new option field to it?

e.g.,

```go
type customK8sResourceClient struct {
  validator Validator
}

// A Validator is a function that validates a resource and returns an error if it is not valid.
type Validator func(kvp KVPair) error
```

Then the Create / Update methods can have something like this in it:

```go
if c.Validator != nil {
  if err = c.Validator(kvp); err != nil {
    return nil, err
  }
}
```

WDYT?

coutinhop: +1 (I was just writing a comment on how we should try to avoid "copying" `customK8sResourceClient`, @caseydavenport's solution is better than what I had)

huiyizzz: Oh right! Sounds like a better solution!

libcalico-go/lib/backend/k8s/resources/kubecontrollersconfig.go
```diff
@@ -17,11 +17,13 @@ package resources
...
 
-	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+
+	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+
+	v3 "github.com/projectcalico/calico/libcalico-go/lib/validator/v3"
```
caseydavenport, line 26: I'd recommend calling this import "validator" - "v3" is too vague and normally refers to the api/projectcalico/v3 package in other locations, so this might be confusing.

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -159,6 +159,7 @@ func init() {
 	registerFieldValidator("containerID", validateContainerID)
 	registerFieldValidator("selector", validateSelector)
 	registerFieldValidator("labels", validateLabels)
+
```
coutinhop, line 162: nit: no need to add newline here

libcalico-go/lib/validator/v3/validator_test.go
```diff
@@ -2723,25 +2723,6 @@ func init() {
 			Node:    "node-1",
 		}, false),
 	)
-
-	DescribeTable("SyncLabels Validator",
```
coutinhop: I think we need to have an equivalent test now that the code was moved to libcalico-go/lib/backend/k8s/resources/kubecontrollersconfig.go, perhaps add a kubecontrollersconfig_test.go if we don't have one already?

coutinhop (approved):

libcalico-go/lib/backend/k8s/resources/kubecontrollersconfig.go
```diff
@@ -17,18 +17,23 @@ package resources
...
+
+	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 )
 
 const (
 	KubeControllersConfigResourceName = "KubeControllersConfigurations"
 	KubeControllersConfigCRDName      = "kubecontrollersconfigurations.crd.projectcalico.org"
 )
 
+var validate *validator.Validate
```
mgleung, line 35: Though I like the pattern of adding a validator object in order to make it extensible to other resources defined by the `customResourceClient`, I think that the validator we use here can be much simpler than the validator we use for the client in general. I think it might be easier (and a little cleaner) to define a separate validator here that will just have a `Validate` method that is overridden per resource. So in our `KubeControllersConfig` case, `Validate` would only need to validate the `k8sSyncLabels` field and we wouldn't need to use an `init` call or a separate call to `Var` in order to target the specific fields to validate.

mgleung (approved):
LGTM

PR #6788, opened 2022-09-29
DHCP agent: Fix check for linux interface being up

neiljerram:
We were looking for "state UP" in the output from `ip monitor link`, but that's not the right thing to look for, and in practice often shows as "state UNKNOWN" when the interface is actually up.

According to https://superuser.com/questions/1463162/unknown-state-of-network-interface, the correct thing is to look for "UP" in the set of flags between < and >, as for example here:

    3: wlp59s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DORMANT group default qlen 1000

I've adjusted the rx_up regexp to do that.

This could impact in practice when a VM is rebooted, because it appears that when that happens, the TAP interface is deleted and then recreated with "state UNKNOWN".  Then, if the VM is also supposed to have a non-default MTU, the DHCP agent would fail to spot that, and so the rebooting VM would internally set a default MTU of 1500 instead of what it ought to have.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/agent/dhcp_agent.py
+++ b/networking-calico/networking_calico/agent/dhcp_agent.py
@@ -166,7 +166,7 @@ class MTUWatcher(object):
         self.port_handler = port_handler
         self.mtu_by_if_name = {}
         self.port_id_by_if_name = {}
-        self.rx_up = re.compile('[0-9]+: (tap[a-z0-9-]+).*: <(?:UP|.+,UP)[,>].* mtu ([0-9]+)')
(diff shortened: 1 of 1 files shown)


fasaxc (approved):
One thought: do you actually need to check that it's UP at all?  It'll still have an mtu if it's down, right?

neiljerram:
Thanks, good thought, I've made that simplification.  Will merge after some more testing...

PR #6804, opened 2022-10-03
add info to routeUpdate msg in Felix for node updates

hedibouattour:
## Description
In the context of passing through typha in calicoVpp to avoid scale issues (by using Felix messages), this PR adds some info to routeUpdate messages in Felix to get nodes' (add/update/delete) updates. And it activates route resolver in the case of external dataplane (vpp dataplane).


felix/calc/l3_route_resolver.go
```diff
@@ -561,7 +574,7 @@ func (c *L3RouteResolver) onNodeUpdate(nodeName string, newNodeInfo *l3rrNodeInf
 	if newNodeInfo != nil {
 		c.nodeNameToNodeInfo[nodeName] = *newNodeInfo
 		for _, a := range newNodeInfo.AddressesAsCIDRs() {
-			c.trie.AddHost(a, nodeName)
+			c.trie.AddHost(a, nodeName, newNodeInfo)
```
caseydavenport, line 577: Sorry if this is obvious, but why the switch to using nodeinfo here instead of just node name?

hedibouattour: The goal of this patch is to add some information to the message in Felix, so using nodeInfo allows to pass in data like: `ASNumber`, `WireguardPublicKey`,  `Labels`.

hedibouattour:
just rebased on master, please review :)

felix/proto/felixbackend.proto
```diff
@@ -568,6 +568,9 @@ message RouteUpdate {
   bool nat_outgoing = 8;
   bool local_workload = 9;
   TunnelType tunnel_type = 10;
+  map<string, string> labels = 11;
```
caseydavenport, line 571: I'd like to discuss this one with @coutinhop and @fasaxc - it seems incongruous to be including things like node labels, node AS number, and wireguard public key on a RouteUpdate message since those aren't really attributes of a route itself.

I understand that this object already has a few somewhat questionable fields (e.g., ip_pool_type) and that we want to avoid unnecessary calculations in the data plane. But conceptually, tying these to a route update feels wrong and will result in sending substantially more information (repeating labels across routes rather than just once per node as needed).

We already send WireguardPublicKey to the data plane it seems:

- https://github.com/projectcalico/calico/blob/3f3d6b353d70695dc2d5ff6a64d40773a2b9847a/felix/proto/felixbackend.proto#L457

The node labels and AS number I would expect to be tied to a message representing a node object. For example, the HostMetadataUpdate object.
 https://github.com/projectcalico/calico/blob/3f3d6b353d70695dc2d5ff6a64d40773a2b9847a/felix/proto/felixbackend.proto#L465

coutinhop: I agree with @caseydavenport, it makes more sense to add AS number and labels to HostMetadataUpdate and subscribe to it + WireguardStatusUpdate wherever you need this information instead of adding everything to RouteUpdate...

hedibouattour: https://github.com/projectcalico/calico/pull/7042
Here's a new PR with the new logic.

hedibouattour:
#7042

PR #6816, opened 2022-10-05
Changing redirect paths for 404 links

ctauchen:
An SEO audit turned up organic traffic that's ending in 404s. This PR redirects those specific URLs to https://projectcalico.docs.tigera.io.

@penkeysuresh Can you check I've got this configuration right?


penkeysuresh:
Looks good to me. We are basically redirecting all those missing links to the docs root site. most of the URL's are pretty old so it makes sense. Would be nice if @bmckercher123 can take to look at this too, but won't block my 👍🏼

penkeysuresh (approved):

bmckercher123:
I agree with Suresh. The only link in the bunch that is of any importance, would be the one for eBPF. So I think it is good strategy to redirect to main site.

PR #6820, opened 2022-10-05
Install hping3 where needed in XDP FV tests

mazdakn:
## Description
Fix the flakiness of XDP FV tests. `hping3` is installed in the main `BeforeEach()` while only needed by one test. The installation of package fails randomly because the package repository is not available. The failure could be due to different reasons including remote server rate limiting policies. Installing it when needed reduces the number of fetches, and fixes the issue. 

The delay as the result of installation allows many tests to sync, and removing it causes other test failure. This PR also fixes that.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/xdp_test.go
+++ b/felix/fv/xdp_test.go
@@ -86,8 +86,6 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ XDP tests with initialized 
 				"8055,8056,22,68",
 				protos[ii])
 
-			felix.Exec("apt-get", "install", "-y", "hping3")
-
 			hostEp := api.NewHostEndpoint()
 			hostEp.Name = fmt.Sprintf("host-endpoint-%d", ii)
 			hostEp.Labels = map[string]string{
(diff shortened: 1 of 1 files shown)


felix/fv/xdp_test.go
```diff
@@ -395,6 +398,7 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ XDP tests with initialized
 			It("should block ICMP too", func() {
 				client, server := clientServerIndexes("tcp")
 
+				time.Sleep(applyPeriod)
```
sridhartigera, line 401: Do we need a sleep here?

mazdakn: It won't work without it. The policy program needs some time to get applied.

mazdakn:
Fixed in https://github.com/projectcalico/calico/pull/6822

PR #6835, opened 2022-10-11
Let calico-bpf decide whether to fail when some counters error - XDP counters may not be present

mazdakn:
## Description
Failing in dumping or flushing a counter map should not prevent the operation for the rest of maps. So instead of returning an error, we just need to log and continue with the rest of maps.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/counters/counters.go
+++ b/felix/bpf/counters/counters.go
@@ -164,7 +164,11 @@ func (c Counters) Read() ([][]uint64, error) {
 	for hook, name := range bpf.Hooks {
 		val, err := c.read(c.maps[hook])
 		if err != nil {
-			return values, fmt.Errorf("Failed to read bpf counters. hook=%s err=%v", name, err)
+			if name == bpf.HookXDP {
+				logrus.Infof("Failed to read XDP bpf counters. err=%v", err)
+				continue
+			}
+			return values, fmt.Errorf("Failed to read bpf counters. hook=%s err=%w", name, err)
 		}
 		if len(values[hook]) < MaxCounterNumber {
 			return values, fmt.Errorf("Failed to read enough data from bpf counters. hook=%s", name)
@@ -210,6 +214,10 @@ func (c *Counters) Flush() error {
 	for hook, name := range bpf.Hooks {
 		err := c.flush(c.maps[hook])
 		if err != nil {
+			if name == bpf.HookXDP {
+				logrus.Infof("Failed to flush XDP bpf counters. err=%v", err)
+				continue
+			}
 			return fmt.Errorf("Failed to flush bpf counters for interface=%s hook=%s. err=%w", c.iface, name, err)
 		}
 		logrus.Infof("Successfully flushed counters map for interface=%s hook=%s", c.iface, name)
--- a/felix/bpf/counters/counters.go
+++ b/felix/bpf/counters/counters.go
@@ -31,6 +31,12 @@ const (
 	counterMapValueSize int = 8
 )
 
+var (
+	// zeroKey is the key to the counters map, and it is set to 0 as it has only one entry
+	zeroKey = make([]byte, counterMapKeySize)
+	zeroVal = make([]byte, counterMapValueSize*MaxCounterNumber*bpf.NumPossibleCPUs())
+)
+
 // The following values are used as index to counters map, and should be kept in sync
 // with constants defined in bpf-gpl/reasons.h.
 const (
@@ -136,8 +142,6 @@ type Counters struct {
 	iface    string
 	numOfCpu int
 	maps     []bpf.Map
-	zeroKey  []byte
-	zeroVal  []byte
 }
 
 func NewCounters(iface string) *Counters {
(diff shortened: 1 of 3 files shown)


felix/bpf/counters/counters.go
```diff
@@ -164,7 +164,11 @@ func (c Counters) Read() ([][]uint64, error) {
...
 		val, err := c.read(c.maps[hook])
 		if err != nil {
-			return values, fmt.Errorf("Failed to read bpf counters. hook=%s err=%v", name, err)
+			if name == bpf.HookXDP {
+				logrus.Infof("Failed to read XDP bpf counters. err=%v", err)
+				continue
+			}
+			return values, fmt.Errorf("Failed to read bpf counters. hook=%s err=%w", name, err)
 		}
 		if len(values[hook]) < MaxCounterNumber {
```
tomastigera, line 173: Is this ever possible given that on line 161 you allocate exactly this much space?

I think it is meant to test `len(val)`  and not to allocate at 161

tomastigera (changes requested):
It feels like the `Read()` logic should be reserved to the user and not the lib/pkg and the pkg could just export the `read()` but with a few defined constants for the hooks. Let the user know that the XDP failed and the user can decide what to do about that.

I think that at line 182 `k := make([]byte, counterMapKeySize)` should not allocate but there should be a static pre-created key, that would also save the comment.

felix/cmd/calico-bpf/commands/counters.go
```diff
@@ -99,11 +99,25 @@ func dumpInterface(cmd *cobra.Command, iface string) error {
...
-
 	bpfCounters := counters.NewCounters(iface)
-	values, err := bpfCounters.Read()
-	if err != nil {
-		return fmt.Errorf("Failed to read bpf counters. iface=%s err=%v", iface, err)
+
+	values := make([][]uint64, len(bpf.Hooks))
+	for i := range values {
+		values[i] = make([]uint64, counters.MaxCounterNumber)
+	}
```
tomastigera, line 107: why this allocation?

tomastigera (commented):
LGTM, one nit :crossed_fingers: the CI passes :)

felix/cmd/calico-bpf/commands/counters.go
mazdakn: `make` only creates the `[][]byte' but each entry `[]byte` is still `nil` thus accessing it panics. It's seems mandatory to also init each of `[]byte` in the 2D `[][]byte`.

tomastigera: But it is then assigned `values[index] = val` sure you would need to check if it is set at `newRow = append(newRow, fmt.Sprintf("%v", values[index][c.Counter]))` but isn't that likely what you want anyway when say the XDP counters are missing?

felix/cmd/calico-bpf/commands/counters.go
```diff
@@ -99,11 +99,25 @@ func dumpInterface(cmd *cobra.Command, iface string) error {
...
+				log.Infof("Failed to read XDP bpf counters. err=%v", err)
+				continue
+			}
+			return fmt.Errorf("Failed to read bpf counters. iface=%v hook=%s err=%w", iface, hook, err)
+		}
+		if len(val) < counters.MaxCounterNumber {
+			return fmt.Errorf("Failed to read enough data from bpf counters. iface=%v hook=%s", iface, hook)
+		}
+		values[index] = make([]uint64, counters.MaxCounterNumber)
+		values[index] = val
```
tomastigera, line 120:
```suggestion
	values := make([][]uint64, len(bpf.Hooks))
	for index, hook := range bpf.Hooks {
		val, err := bpfCounters.Read(index)
		if err != nil {
			if hook == bpf.HookXDP {
				log.Infof("Failed to read XDP bpf counters. err=%v", err)
				continue
			}
			return fmt.Errorf("Failed to read bpf counters. iface=%v hook=%s err=%w", iface, hook, err)
		}
		if len(val) < counters.MaxCounterNumber {
			return fmt.Errorf("Failed to read enough data from bpf counters. iface=%v hook=%s", iface, hook)
		}
		values[index] = val
```

felix/cmd/calico-bpf/commands/counters.go
```diff
@@ -114,8 +128,12 @@ func dumpInterface(cmd *cobra.Command, iface string) error {
 	for _, c := range counters.Descriptions() {
 		newRow := []string{c.Category, c.Caption}
 		// Now add value related to each hook, i.e. ingress, egress and XDP
-		for index := range bpf.Hooks {
-			newRow = append(newRow, fmt.Sprintf("%v", values[index][c.Counter]))
+		for index, hook := range bpf.Hooks {
+			if hook == bpf.HookXDP && !xdpOK {
```
tomastigera, line 132:
```suggestion
			if values[index] == nil {
```

tomastigera (commented):
I think this is a little simpler :-)

```
		values[index] = make([]uint64, counters.MaxCounterNumber)
		values[index] = val
```

I wonder what static checks say about this assignment ;-)

tomastigera (approved):
pls squash

PR #6836, opened 2022-10-11
felix/bpf: fix switch from FeatureDetectOverrides to FeatureGates

tomastigera:
Also set BPFConnectTimeLoadBalancingWorkaround=udp when testing for udp
    tests.
    
    Test that BPFConnectTimeLoadBalancingWorkaround=udp only programs UDP
    related services, but also mixed services.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/bpf_test.go
+++ b/felix/fv/bpf_test.go
@@ -294,7 +294,11 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 			options.ExtraEnvVars["FELIX_BPFExtToServiceConnmark"] = "0x80"
 
 			if ctlbWorkaround {
-				options.ExtraEnvVars["FELIX_FeatureGates"] = "BPFConnectTimeLoadBalancingWorkaround=enabled"
+				if testOpts.protocol == "udp" {
+					options.ExtraEnvVars["FELIX_FeatureGates"] = "BPFConnectTimeLoadBalancingWorkaround=udp"
+				} else {
+					options.ExtraEnvVars["FELIX_FeatureGates"] = "BPFConnectTimeLoadBalancingWorkaround=enabled"
+				}
 			}
 		})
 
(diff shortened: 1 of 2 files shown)


felix/fv/bpf_test.go
```diff
@@ -945,6 +1046,30 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 
 					pol = createPolicy(pol)
 
+					By("allowing to self via MASQ", func() {
```
tomastigera, line 1049: I only wonder why we had to add this policy. I'd expect that stuff from the host is allowed. It is not needed when CTLB is used as it is always resolved at connect time and such connections never leave the pod.

felix/fv/bpf_test.go
```diff
@@ -945,6 +1046,30 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+
+						nets := []string{felixes[0].IP + "/32"}
+						switch testOpts.tunnel {
+						case "ipip":
+							nets = []string{felixes[0].ExpectedIPIPTunnelAddr + "/32"}
+						}
+
+						pol := api.NewGlobalNetworkPolicy()
+						pol.Namespace = "fv"
```
fasaxc, line 1058: GNPs aren't namespaced; surprised it doesn't reject it!

fasaxc: In iptables mode, I think traffic to self via a service does require a policy but the policy sees the Pod IP.  I don't think it's acceptable to require a policy that matches the post-SNAT IP since that's different (and the IP is hard to guess)

fasaxc (changes requested):
Unfortunately, I think that policy _is_ a deviation from the policy model so I think we'll need to fix that up

felix/fv/bpf_test.go
tomastigera: OK, this is not an issue created by this PR, so I'd prefer to address it separately. WDYT

fasaxc: Please drop a FIXME explaining that this policy shouldn't be needed.

fasaxc (approved):

PR #6839, opened 2022-10-12
networking_calico: Prime project cache at startup

tj90241:
## Description

The Calico mechanism driver maintains a cache of project data that it uses to annotate ports with. This cache is currently filled whenever an item is found missing at runtime.

This means, though, that the first port annotation performed after a cold restart of `neutron-server` will result in a fill penalty that could be avoided by priming the cache during driver startup.

This commit thus primes the cache during startup.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico for OpenStack: prime the project (aka tenant) data cache on Neutron server startup
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
+++ b/networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
@@ -78,6 +78,10 @@ class WorkloadEndpointSyncer(ResourceSyncer):
         self.region_string = calico_config.get_region_string()
         self.namespace = datamodel_v3.get_namespace(self.region_string)
 
+        # Prime the project data cache now so that we do not pay a fill
+        # penalty the first time we need to annotate a port on a cold start.
+        self.cache_port_project_data()
+
     def delete_legacy_etcd_data(self):
         if self.namespace != datamodel_v3.NO_REGION_NAMESPACE:
             datamodel_v3.delete_legacy(self.resource_kind, '')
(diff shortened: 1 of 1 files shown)


tj90241:
aside: The entire cache is flushed whenever it is too full.  This is probably not the best practice, especially since most operators are not going to notice this unless they go fishing for it. Something for a future PR.

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
```diff
@@ -295,25 +299,31 @@ def add_port_project_data(self, port, context):
             port[PORT_KEY_PROJ_DATA] = proj_data
             return
 
+        # Not cached, so look up the port's project in the Keystone DB.
+        self.cache_port_project_data()
```
neiljerram, line 303: I think this means we get the penalty of calling `self.keystone.projects.list()` every time, doesn't it?  Perhaps I am not understanding exactly where the penalty concern is.

neiljerram (commented):
Not sure I am understanding this yet.

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
tj90241: See the condition immediately above this line -- we (still) only do this when we miss in the cache.

This PR strictly avoids potentially having to fill the cache **during**, e.g., the first Neutron port update upon a wSGI webserver start by forcing the cache fill to be performed as part of mechanism driver initialisation. The cache fill logic was ripped out of this section and put into a function so that it could be called both here, where it previously resided, and in the constructor, for that purpose.

There still is the problem where a new project is created at runtime and we perform a port request associated with that project, then we have a fill operation.... that is probably best fixed separately.

neiljerram: Sorry, of course you are right, and thanks for explaining further.

neiljerram (approved):

PR #6841, opened 2022-10-12
networking_calico: Fix spurious endpoint delete

tj90241:
## Description
A port which is created and subsequently updated may race into code which was meant to delete the WorkloadEndpoint on the source hypervisor during a live-migration.

When the mechanism driver gets a port-update, first check that the binding existed to begin with before deleting the WorkloadEndpoint. If no such binding exists on the original port, then the port just came into existence and was updated just very shortly after.

When this happens, logs like the following are printed: "Migration, delete WorkloadEndpoint on old host"
(rather than a message with an actual hostname to which the original port was bound).

The fact that this was happening should have been harmless and just resulted in some confusing log messages, but fix it nonetheless.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico for OpenStack: suppress spurious attempted endpoint deletions (which aren't needed because the endpoint does not exist at that point) and associated logging, in some port update scenarios
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
+++ b/networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
@@ -790,8 +790,10 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
             endpoint_should_already_exist = port_bound(original)
 
             # Check for migration so that we can reliably delete the
-            # WorkloadEndpoint on the old host.
-            if original['binding:host_id'] != port['binding:host_id']:
+            # WorkloadEndpoint on the old host. That happens when the host
+            # formerly had a binding, and it doesn't match the current one.
+            if original['binding:host_id'] and \
+                    original['binding:host_id'] != port['binding:host_id']:
                 LOG.info("Migration, delete WorkloadEndpoint on old host %s",
                          original['binding:host_id'])
                 self.endpoint_syncer.delete_endpoint(original)


networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -790,8 +790,10 @@ def update_port_postcommit(self, context):
             endpoint_should_already_exist = port_bound(original)
 
             # Check for migration so that we can reliably delete the
-            # WorkloadEndpoint on the old host.
-            if original['binding:host_id'] != port['binding:host_id']:
+            # WorkloadEndpoint on the old host. That happens when the host
+            # formerly had a binding, and it doesn't match the current one.
+            if original['binding:host_id'] and \
+                    original['binding:host_id'] != port['binding:host_id']:
```
neiljerram, line 796: Just thinking through this...

The change only makes a difference if `original['binding:host_id']` is Python-false.

It looks very plausible, but we use a different condition in all of the places where we're deciding to call `self.endpoint_syncer.write_endpoint`, namely `port_bound(port)` or `port['binding:vif_type'] != 'unbound'`.  Hence I think the proposed change is only correct if Neutron updates `port['binding:host_id']` and `port['binding:vif_type']` in perfect sync.  It feels like it would be safer to say `if endpoint_should_already_exist and` instead of `if original['binding:host_id'] and` - that should be correct in all cases, including when there is a perfect sync with `binding:host_id`.  WDYT?

neiljerram (commented):
Thanks Tyler; a query on this.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
neiljerram: I agree with you as far as "And we try to delete something which does not exist."

My point, though, is that we only ever **create** an endpoint (in etcd) if `port_bound(port)`, so it follows that we only need to **delete** an endpoint if `port_bound(original)` is true for the previous port state `original`.  I think that should be 100% reliable so long as we don't miss any transitions between `port_bound(port)` and `!port_bound(port)`, and I've re-reviewed the code to check that it won't miss any such transitions.

caseydavenport:
What's the latest status on this PR?

tj90241:
Hello, @caseydavenport  - sorry for the delay. This is just a perf. enhancement/not a bug. I have to think about what Neil mentioned above and run it through regression testing. I'll try to do that soon, but in the meantime this doesn't need to land in a particular release or anything.

caseydavenport:
Closing as this seems inactive. Please shout if you pick this up again and I can reopen.

PR #6845, opened 2022-10-13
bpf pcap filter

tomastigera:
## Description

    felix/bpf: pcap filters
    
    This implements translation of cBPF generated by libpcap to eBPF and
    wraps it in code that allows us to test it and jump from it to another
    program based on match or not.

    felix/bpf: do not patch no_log binaries
    
    It is a no-op, there is no logging prefix to patch.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/cmd/calico-bpf/commands/policy_debug.go
```diff
@@ -55,7 +55,7 @@ var policyDumpCmd = &cobra.Command{
 		}
 		hooks := []string{}
 		if hook == "all" {
-			hooks = []string{"ingress", "egress"}
+			hooks = []string{"ingress", "egress", "xdp"}
```
mazdakn, line 58: This can be replaced by `bpf.Hooks`: https://github.com/projectcalico/calico/blob/57b84fa83985a288ea7c2bddb858e5014ab5705f/felix/bpf/bpf.go#L60

felix/bpf/tc/defs/defs.go
```diff
@@ -50,9 +52,13 @@ const (
 	ProgIndexV6Allowed
 	ProgIndexV6Icmp
 	ProgIndexV6Drop
+
+	ProgIndexStart = ProgIndexPolicy
```
mazdakn, line 56: Where this const is used? cannot find it.

felix/bpf/asm/asm.go
```diff
@@ -307,6 +319,10 @@ func MakeInsn(opcode OpCode, dst, src Reg, offset int16, imm int32) Insn {
 }
 
 func (n Insn) String() string {
+	if n.OpCode() == LoadImm64Pt2 && n.Dst() == R0 && n.Src() == R0 && n.Off() == 0 {
+		return fmt.Sprintf("%v dst=%v src=%v off=%v imm=%#08x/%d", "LoadImm64Pt2",
+			n.Dst(), n.Src(), n.Off(), uint32(n.Imm()), n.Imm())
+	}
```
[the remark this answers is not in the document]
tomastigera: The problem is that the opcode has the same value so the stringer prints the wrong generated value. So in this case, we have to do it unfortunately this way :shrug:

felix/bpf/tc/defs/defs.go
tomastigera: no used, will remove :+1:

felix/bpf/asm/asm.go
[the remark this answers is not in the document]
tomastigera: I will revert this. eBPF has only 64bit load imm, so I should remove the load 32 imm and there will be no more conflict

felix/bpf/filter/filter.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+func NewStandAlone(minLen int, expression string) (asm.Insns, error) {
+	return newFilter(minLen, expression, 0, true)
+}
+
+func newFilter(minLen int, expression string, jumpMapFD bpf.MapFD, standAlone bool) (asm.Insns, error) {
+	b := asm.NewBlock(true)
+
+	insns, err := pcap.CompileBPFFilter(layers.LinkTypeEthernet, minLen, expression)
```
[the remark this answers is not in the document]
tomastigera: OK, I added L3 for IPv4 we will need to revisit this when we have IPv6 and we will need to have 2 filters in the same program each for different L3 proto

felix/bpf/ut/filter_test.go
```diff
@@ -0,0 +1,114 @@
...
+	)
+
+	type testCase struct {
+		expression string
+		match      bool
+	}
+
+	tests := []testCase{
+		// L2
+		{"ip", true},
```
fasaxc, line 61: Worth checking a negated version of each of these?

fasaxc (approved):
Nice!

The assembly looks plausible and the tests look solid.  Do the tests get 100% coverage; think that's probably the best way to test that we can't be handed a "bad" PCAP string that'll blow up?

tomastigera:
> The assembly looks plausible and the tests look solid. Do the tests get 100% coverage; think that's probably the best way to test that we can't be handed a "bad" PCAP string that'll blow up?

The plan in the next PR that starts using this is that if the compilation fails, we either disable or enable all, havent decided yet which one is better. I am leaning towards enable all, but that may have perf impact

Issue #6180, opened 2022-06-03
Calico Node container issues when FIPS mode is enabled


gabrielericciardi:
In our system there is an InitContainer requiring a kubectl exec on the Calico node container. This worked until we decided to upgrade runc from runc-1.0.0-70.rc92.module+el8.4.0+10607+f4da7515.x86_64 to runc-1.1.0-1.rhaos4.10.el8.x86_64: after that, the kubectl exec fails reporting FIPS violations in the system log.
Contacted the Red Hat support, and their analysis points to the Calico Node container as the issue: they provided a code snippet to execute a FIPS self-test from inside the container, and they proved that, on a node with FIPS mode enabled, the self-test passes from within a RedHat 8.6 UBI container but it fails from within the Calico Node container.

## Expected Behavior

The kubectl exec command executed in the Calico Node container succeeds on nodes with FIPS enabled.
The provided FIPS self-test passes within the Calico Node container.

## Current Behavior

The kubectl exec command executed in the Calico Node container fails on nodes with FIPS enabled.
The provided FIPS self-test fails within the Calico Node container.

## Possible Solution

The analysis seems to point to the lack of the libssl library and both the .hmac files for libssl and libcrypto library, which are required by FIPS. RedHat proved that by injecting into the Calico Node container the missing parts the FIPS self-test passes (see Context). 
The UBI container which succeeds in the test has indeed indicated libraries and .hmac files.
Here's the code snippet provided by Red Hat:
```c
//check-fips.c
#include <stdio.h>
#include <openssl/ssl.h>
#include <openssl/fips.h>
#include <openssl/err.h>
int main(int argc, char *argv[])
{
   fprintf(stderr,"Startup\n");
   fprintf(stderr,"all algos added\n");
   if(FIPS_mode())
      fprintf(stderr,"FIPS mode already set.\n");
   else {
      fprintf(stderr,"Not to set FIPS mode...\n");
   }
   fprintf(stderr,"Attempt FIPS self tests...\n");
   if (FIPS_selftest()) {
      fprintf(stderr,"FIPS self tests succeeded.\n");
   }
   else {
      fprintf(stderr,"ERROR: FIPS self tests failed.\n");
      ERR_print_errors_fp(stderr);
   }
   return 0;
}
```

## Steps to Reproduce (for bugs)

1. Compile the provided code snippet on a Red Hat 8.6 machine with ```gcc -o check-fips check-fips.c -lssl -lcrypto ```
[...]

PR #6852, opened 2022-10-16
Retain OpenSSL FIPS dependent files in calico-node image

hjiawei:
## Description

The current calico-node cleanup script removes certain files from the base
image that FIPS mode relies on. When FIPS is enabled, OpenSSL gives internal
error `FATAL FIPS SELFTEST FAILURE`. It also causes calico-node readiness
probe to fail with the same error and become not ready. The checkup script
is modified to not remove libssl.so and hidden hmac files.

## Related issues/PRs

https://github.com/projectcalico/calico/issues/6180

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/clean-up-filesystem.sh
+++ b/node/clean-up-filesystem.sh
@@ -221,6 +221,16 @@ while read -r path; do
     libs_to_keep[$path]=true
     continue
   fi
+  # These libraries and hmac files under /usr/lib64 are needed when ubi container
+  # is running in FIPS mode. They are not directly linked by the allowed binaries.
+  # * /usr/lib64/.libcrypto.so.x.y.z.hmac
+  # * /usr/lib64/.libssl.so.x.y.z.hmac
+  # * /usr/lib64/libssl.so.x.y.z
+  if [[ "$path" =~ .libcrypto|.libssl|libssl ]]; then
+    echo "FIPS PLUGIN: $path"
+    libs_to_keep[$path]=true
+    continue
+  fi
 done < <(find /usr/lib64 \( -type f -or -type l \))
 
 # Now remove all but a keep-list of RPM packages.  Cleaning up packages with rpm itself updates the


fasaxc (changes requested):
Why not just change `clean-up-filesystem.sh` to not delete the libs that you need?  You don't need a whole new dummy binary to link in the libs unless that's the only way to figure out what the libs are?

node/Dockerfile.amd64
```diff
@@ -46,7 +46,9 @@ RUN dnf install -y 'dnf-command(config-manager)' && \
     # Install required packages for building rpms. yum-utils is not required but it gives us yum-builddep to easily install build deps.
     yum install --allowerasing -y rpm-build yum-utils make && \
     # Need these to build runit.
-    yum install --allowerasing  -y wget glibc-static gcc && \
+    yum install --allowerasing -y wget glibc-static gcc && \
```
fasaxc, line 49: gcc is a heavy dep to install only to delete it again, can we not build the binary in go-build?

hjiawei: This is a white space change. `gcc` is an existing dependency to build runit. I think the fips OpenSSL is a redhat additional to the vanilla OpenSSL project. We have to build the test problem in Redhat (and derived) OSs (in this case centos stage). Our go-build is Debian based and it is missing openssl dev pages and openssl/fips.h header.

hjiawei:
> Why not just change `clean-up-filesystem.sh` to not delete the libs that you need? You don't need a whole new dummy binary to link in the libs unless that's the only way to figure out what the libs are?

Only cleanup script is modified to exclude the libssl.so and hidden hmac files. The dummy check-fips binary is removed. Manually verified calico-node image after this change.

hjiawei:
This change is tested in a FIPS enabled Openshift cluster and calico-node is ready/running.

mgleung (approved):
LGTM

fasaxc (approved):

node/Dockerfile.amd64
fasaxc: Sorry, my bad!

PR #6857, opened 2022-10-17
networking_calico: Do not checksum DHCP packets

tj90241:
## Description

Very old versions of isc-dhcp-client broke when UDP packets had a checksum field not properly filled in as part of GSO/checksum offload support being introduced for virtio_net: https://lwn.net/Articles/373209/

The missing/incorrect checksum in the UDP header was problematic for isc-dhcp-client (and possibly other DHCP clients) because they use raw sockets, and were unaware of how to handle TP_STATUS_CSUMNOTREADY.

OpenStack has historically worked around this by adding an iptables mangle rule to always rewrite the UDP header for DHCP packets. But, isc-dhcp-client has also been patched to account for this since 2014, and likewise has qemu (for virtio_net) since 2009: https://git.qemu.org/?p=qemu.git;a=commit;h=1d41b0c

The consequence of patching the mangle rule in iptables is that the underlying OpenStack mechanisms tend to add the rules in an order that upsets Felix. When Felix sees this (often during live-migration, when the VM has no networking), it can spend a fair amount of time resyncing the rules... thus leaving a live-migrating instance with no networking for an appreciable amount of time.

As this should not be needed for any Linux distribution in the last decade and it can perturb Felix, simply do not fill in DHCP UDP checksums here with the assumption that it has been fixed/is being done elsewhere.

## Related issues/PRs

* https://lwn.net/Articles/373209/
* https://git.qemu.org/?p=qemu.git;a=commit;h=1d41b0c
* https://github.com/isc-projects/dhcp/commit/7ff6ae5aa85754119319def3c7f225a40da299c4

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico for OpenStack: remove iptables programming by the DHCP agent that is no longer needed, and that was increasing the need for Felix to resync Calico's iptables programming
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/agent/linux/dhcp.py
+++ b/networking-calico/networking_calico/agent/linux/dhcp.py
@@ -220,21 +220,31 @@ class CalicoDeviceManager(dhcp.DeviceManager):
         pass
 
     def fill_dhcp_udp_checksums(self, *args, **kwargs):
-        retries = 10
-        while retries > 0:
-            try:
-                super(CalicoDeviceManager, self).fill_dhcp_udp_checksums(
-                    *args, **kwargs)
-            except RuntimeError:
-                # fill_dhcp_udp_checksums() can fail transiently if iptables
-                # is modified concurrently, especially with an aggressive
-                # iptables writer such as Felix running.
-                LOG.exception("Failed to insert checksum rule, may retry...")
-                time.sleep(0.1)
-                retries -= 1
-            else:
-                LOG.debug("Inserted DHCP checksum rule.")
-                break
-        else:
-            LOG.error("Failed to insert DHCP checksum rule. Exiting...")
-            sys.exit(1)
+        # NOTE(tstachecki): Very old versions of isc-dhcp-client broke when
+        # UDP packets had a checksum field not properly filled in as part of
+        # GSO/checksum offload support being introduced for virtio_net:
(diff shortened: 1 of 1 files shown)


tj90241:
Thumbing over the qemu commit, it does explicitly check for both a) IPv4 and b) MTU <= 1500... so this may be impactful to IPv6 DHCP clients, or if jumbo frames are used to renew a lease (???) if said DHCP client is both using raw sockets and is also not aware of the checksum complications here.

If simply removing this code is deemed insufficient, I can look into seeing what can be done to patch the OpenStack code that's doing the iptables management closer.

beddari:
Great work @tj90241  , for us it would be perfectly fine to merge this as-is.

neiljerram (approved):
Thanks @tj90241 , this makes sense to me so long as it passes our testing...

neiljerram:
@tj90241 @beddari Do you happen to know which versions of CirrOS are fixed so as not to need this DHCP checksumming rule?  Our internal system tests are using CirrOS 0.3.2, and that version does not boot successfully with the change in this PR.

neiljerram:
Actually I probably don't need a detailed answer here.  I've done an initial test with CirrOS 0.6.0 and that seems to be working.

PR #6862, opened 2022-10-18
Add generalized TTL security mechanism (GTSM)

Josh-Tigera:
## Description
Introduce new functionality for the generalized TTL security mechanism (GTSM) by introducing a new BGPPeer resource API field, as well as corresponding BIRD template changes to ingest this new field and write the appropriate BIRD configuration lines.

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Add generalized TTL security mechanism (GTSM) via BGPPeer API
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/bgppeer.go
```diff
@@ -100,6 +100,10 @@ type BGPPeerSpec struct {
 	// This removes BGP loop prevention and should only be used if absolutely necesssary.
 	// +optional
 	NumAllowedLocalASNumbers *int32 `json:"numAllowedLocalASNumbers,omitempty"`
+	// TTLSecurity enables the generalized TTL security mechanism (GTSM). The provided value is the number of hops
+	// (edges) between the peers.
```
caseydavenport, line 104: Probably worth adding a bit more context to this.

e.g.,

"GTSM protects against spoofed packets by ignoring received packets with a smaller than expected TTL"

confd/tests/compiled_templates/explicit_peering/keepnexthop-global/bird.cfg
```diff
@@ -68,13 +67,21 @@ template bgp bgp_template {
...
 # For peer /global/peer_v4/172.19.4.87
 protocol bgp Global_172_19_4_87 from bgp_template {
+  ttl security off;
+  multihop;
   neighbor 172.19.4.87 as 64533;
   source address 10.192.0.2;  # The local address we use for the TCP connection
   next hop keep;
 }
 
+
```
caseydavenport, line 81: Wonder where these newlines are coming from. Would be nice to find out and get rid of them, if possible.

api/pkg/apis/projectcalico/v3/bgppeer.go
```diff
@@ -100,6 +100,10 @@ type BGPPeerSpec struct {
...
+	// +optional
+	TTLSecurity *uint32 `json:"ttlSecurity,omitempty"`
```
caseydavenport, line 106: Do we need any validation on this field? What is the maximum allowed value? Will BIRD be OK with a value of max(uint32)?

confd/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -170,6 +169,7 @@ template bgp bgp_template {
 {{if eq $onode_ip ($node_ip) }}# Skipping ourselves ({{$node_ip}})
 {{- else if ne "" ($onode_cluster_id)}}# Skipping {{$onode_ip}} ({{.}}) as it is a route reflector with cluster ID {{$onode_cluster_id}}
 {{else if ne "" $onode_ip}}protocol bgp Mesh_{{$id}} from bgp_template {
+  multihop;
```
mgleung, line 172: Do we need to explicitly set `ttl security off` for the node-to-node mesh?

api/pkg/apis/projectcalico/v3/bgppeer.go
coutinhop: Adding to this, both TTL in IPv4 and hop limit in IPv6 are 8 bit fields, so technically their maximum value is 255. Would it be possible and/or worth it to use `*uint8` here instead?

api/pkg/apis/projectcalico/v3/bgppeer.go
Josh-Tigera: Done.

confd/tests/compiled_templates/explicit_peering/keepnexthop-global/bird.cfg
Josh-Tigera: Looks like these probably came from something I was doing during debugging. I didn't realize we diffed with `--ignore-blank-lines` so was expecting the test to tell me about any formatting changes which never came. I've gone ahead and fixed up all the whitespace changes.

calico/reference/resources/bgppeer.md
```diff
@@ -34,18 +34,19 @@ spec:
...
+| ttlSecurity              | Enables the generalized TTL security mechanism (GTSM) which protects against spoofed packets by ignoring received packets with a smaller than expected TTL value. The provided value is the number of hops (edges) between the peers. Does not apply to mesh peers.                                                                                                                                | 0 - 255                                                                                     | 8-bt integer                      | `nil` (results in BIRD configuration `ttl security off`)          |
```
caseydavenport, line 49: I don't think you need the "does not apply to mesh peers" bit here - the fact that this is on a BGPPeer object means it could never apply to mesh peers anyway.

Josh-Tigera: Ah that's true. I've removed it.

caseydavenport (approved):
LGTM

PR #6864, opened 2022-10-18
FV test to check for bpf program leak

mazdakn:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/bpf_programs_test.go
+++ b/felix/fv/bpf_programs_test.go
@@ -96,7 +96,7 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ Felix bpf dataplane", []api
 	})
 
 	It("should not leak bpf programs", func() {
-		noOfBPFPRograms := func() int {
+		noOfBPFPrograms := func() int {
 			out, err := felixes[0].ExecCombinedOutput("bpftool", "prog", "list")
 			Expect(err).NotTo(HaveOccurred())
 			noOfProgs := strings.Count(out, "sched_cls")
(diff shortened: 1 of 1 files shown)


felix/fv/bpf_programs_test.go
```diff
@@ -92,4 +94,48 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ Felix bpf reattach object",
...
+
+		felixes[0].TriggerDelayedStart()
+		Eventually(noOfBPFPRograms, "10s", "100ms").Should(Equal(initVal + expectedHostPrograms))
+
+		By("creating workloads")
+		for i := 0; i < noOfWorkloads; i++ {
+			wl[i] = workload.New(felixes[0], fmt.Sprintf("w%d", i), "default", fmt.Sprintf("10.65.0.%d", i+2), "8085", "tcp")
+			wl[i].Start()
+			wl[i].ConfigureInInfra(infra)
+			Eventually(noOfBPFPRograms, "10s", "100ms").Should(Equal(initVal + (i+1)*expectedWorkloadPrograms + expectedHostPrograms))
```
sridhartigera, line 120: Can we do this once all the workloads are created?

felix/fv/bpf_programs_test.go
```diff
@@ -92,4 +94,48 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ Felix bpf reattach object",
 		Eventually(secondRunProg2, "10s", "100ms").Should(BeClosed())
 		Expect(secondRunBase).NotTo(BeClosed())
 	})
+
+	It("should not leak bpf programs", func() {
+		noOfBPFPRograms := func() int {
```
sridhartigera, line 99: Please change this to ```noOfBPFPrograms```

felix/fv/bpf_programs_test.go
```diff
@@ -92,4 +94,48 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ Felix bpf reattach object",
...
+			Eventually(noOfBPFPRograms, "10s", "100ms").Should(Equal(totalProgs - expectedWorkloadPrograms))
+			wl[i].Start()
+			Eventually(noOfBPFPRograms, "10s", "100ms").Should(Equal(totalProgs))
+		}
+
+		By("removing workloads")
+		for i := 0; i < noOfWorkloads; i++ {
+			wl[i].Stop()
+			wl[i].RemoveFromInfra(infra)
+			Eventually(noOfBPFPRograms, "10s", "100ms").Should(Equal(totalProgs - (i+1)*expectedWorkloadPrograms))
```
sridhartigera, line 136: Since we have tested the churn in the above block, I think it is ok to do a single check after all workloads are removed. WDYT?

felix/fv/bpf_programs_test.go
mazdakn: We can do that, but I think checking the change in number of program after each change (either addition or removal of a workload) is more exact.

felix/fv/bpf_programs_test.go
mazdakn: Fixed it.

sridhartigera (approved):

felix/fv/bpf_programs_test.go
```diff
@@ -92,4 +94,48 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ Felix bpf reattach object",
...
+		felixes[0].TriggerDelayedStart()
+		Eventually(noOfBPFPrograms, "10s", "100ms").Should(Equal(initVal + expectedHostPrograms))
+
+		By("creating workloads")
+		for i := 0; i < noOfWorkloads; i++ {
+			wl[i] = workload.New(felixes[0], fmt.Sprintf("w%d", i), "default", fmt.Sprintf("10.65.0.%d", i+2), "8085", "tcp")
+			err := wl[i].Start()
+			Expect(err).ToNot(HaveOccurred())
+			wl[i].ConfigureInInfra(infra)
+			Eventually(noOfBPFPrograms, "10s", "100ms").Should(Equal(initVal + (i+1)*expectedWorkloadPrograms + expectedHostPrograms))
```
tomastigera, line 121: I wonder if all these `Eventually` just make the tests run nicely. I think we would need more of a fuzzy test that randomly in parallel creates and tears down a bunch of workloads and then after a while checks that the numbers match. There can be some race somewhere and a nice sync test may not trigger that.

We may avoid running the test regularly or we can configure the test so that it normally runs for a short period only and then use it as a base for testing when we need it :thinking:

mazdakn: I ran the test more about 50 times, and it looks that it works fine. There should not be any flake.

But about fuzzing, do you think FV is the right place? golang 1.18 seems to have a fuzzer, maybe we can use it. Or do you any other tool?

Issue #6772, opened 2022-09-27
Wrong Felix docs for IptablesBackend


manuelbuil:
[Documentation](https://projectcalico.docs.tigera.io/reference/felix/configuration) says that it is possible to set the value "IptablesBackend: Auto". However, when this is done, felix fails to start because it does not understand "Auto". When checking the code, the option "Auto" is not part of the correct values:

https://github.com/projectcalico/calico/blob/master/libcalico-go/lib/validator/v3/validator.go#L380-L384

It can only be empty, Legacy or NFT. 

## Expected Behavior

Docs reflecting the correct values

## Current Behavior

Docs do not reflect the correct values

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Calico version
* Orchestrator version (e.g. kubernetes, mesos, rkt):
* Operating System and version:
* Link to your project (optional):

PR #6871, opened 2022-10-19
Add `Auto` option for `IptablesBackend` to FelixConfiguration

huiyizzz:
## Description

We support `auto` to be set in IptablesBackend as an env var in Felix. But for the FelixConfiguration API resource, there is no `Auto` option. In the API code we say unset means legacy, but actually it is auto.

This PR adds an "Auto" option for `IptablesBackend` and updates the comments to say "Auto" is the default.
## Related issues/PRs

https://github.com/projectcalico/calico/issues/6772
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add missing Auto option for IptablesBackend FelixConfiguration field
```

## Reminder for the reviewer

Other optional labels:


caseydavenport (changes requested):
I'm not sure yet if this is the right fix - we do support auto-detection of the iptables backend in Felix. Certainly the changes to config_params.go seem wrong to me.

I think this might just be a case of the documentation being confusing.

i.e.,

- We support a value of Auto being set as an env var in FELIX_IPTABLESBACKEND
- We do not support a value of Auto being set in the v3 FelixConfiguration API.

This is technically OK since they are different APIs, but our docs imply that the same values are appropriate on both the env var and v3 APIs.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -133,7 +134,7 @@ type FelixConfigurationSpec struct {
 	// disable iptables refresh. [Default: 90s]
 	IpsetsRefreshInterval *metav1.Duration `json:"ipsetsRefreshInterval,omitempty" configv1timescale:"seconds"`
 	MaxIpsetSize          *int             `json:"maxIpsetSize,omitempty"`
-	// IptablesBackend specifies which backend of iptables will be used. The default is legacy.
+	// IptablesBackend specifies which backend of iptables will be used. The default is auto.
```
caseydavenport, line 137: This should be "Auto" (capitalized, to match the enum value).

caseydavenport (changes requested):
I think we're missing a couple of things in this PR:

- Docs updates to this file: https://github.com/projectcalico/calico/blob/master/calico/reference/resources/felixconfig.md
- Unit tests to assert that Auto is accepted through validation, and that Felix config parsing handles it correctly.

I'd look at these two suites:

- https://github.com/projectcalico/calico/blob/master/felix/config/config_params_test.go#L136
- https://github.com/projectcalico/calico/blob/master/libcalico-go/lib/validator/v3/validator_test.go

libcalico-go/lib/validator/v3/validator_test.go
```diff
@@ -665,6 +667,7 @@ func init() {
 		Entry("should reject : in an interface", libapiv3.WorkloadEndpointSpec{InterfaceName: "Invalid:Intface"}, false),
 
 		// (API) FelixConfiguration.
+		Entry("should accept a valid IptablesBackend value 'Auto'", api.FelixConfigurationSpec{IptablesBackend: &iptablesBackendAuto}, true),
```
coutinhop, line 670: Could you add cases for accepting "NFT" and "Legacy" and rejecting an invalid value here as well?

mgleung (approved):

Issue #5549, opened 2022-01-31
Egress Address and Device based on service


teiclap:
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context

On an environment having more than one network connection, I'd need to assign the exact
egress and device to a service.
In the service definition I should be able to explicit the Egress IP address as well the device, for instance
as 
    - egressIP : XX.YY.ZZ.WW
    - egressDEV : eth1
resulting that IP traffic originated by the POD implementing the service and not directed towards
other PODs is masqued with egressIP as Source and statically routed through /dev/eth1

## Your Environment

* Calico version
* Orchestrator version (e.g. kubernetes, mesos, rkt):
* Operating System and version:
* Link to your project (optional):

teiclap:
Hi @mgleung

> Hi @teiclap, just wanted to clarify your use case a little. Are you trying to make it so that all traffic external to your cluster from pods that implement a service appear to be from the same network interface? If the IP that you're trying to specify is the node's IP address (primary network interface) then would [configuring outgoing NAT](https://projectcalico.docs.tigera.io/networking/workloads-outside-cluster) work for you? I know that it does not directly connect to a service, but you should be able to annotate the pods for your service such that they [reference specific IPPools](https://projectcalico.docs.tigera.io/networking/legacy-firewalls).

No, the requested behavior is different than what you can achieve with simple NAT.
Let's assume that the host operating system has 3 interfaces to the network, seen as /dev/net1, /dev/net2, /dev/net3
and an IP address is assigned on each of them, IP1, IP2 and IP3.
The routing table has a default gw as net1
You want your service S to be published at the address IP2 but also want that the EGRESS IP address is IP2 and that
that specific traffic ir routed via net2.
You'd need that once an instance of S is created bu K8s and has assigned internal IP address IPX, there's a rule in NAT
that if the destination is not in the scope of the internal K8s network, the packet is SNATTED and S-IP is replaced with IP2
and at the same time a new static rule is added to the routing table as SBR where packets with Source as IP2 are routed via net2.

Best regards,
Claudio.

caseydavenport:
> In the service definition I should be able to explicit the Egress IP address as well the device, for instance

This sounds like you're asking for a change to a Kubernetes API, which is out of Calico's control. If you want this to be implemented via `Service` you'll need to raise an enhancement request against Kubernetes. 

[...]

Issue #3275, opened 2020-02-24
FELIX_NATOUTGOINGADDRESS and IPv6


duylong:
Hi,

I have a problem when I'm trying to configure Calico in order to use FELIX_NATOUTGOINGADDRESS. 

I'm in a dualstack IPv4/IPv6 with Kubernetes.  When I use this configuration in my calico.yaml deployment:
```
            - name: FELIX_NATOUTGOINGADDRESS
              valueFrom:
                fieldRef:
                  fieldPath: status.hostIP
```
... I get this error:

```
2020-02-24 14:12:52.093 [WARNING][1558] table.go 1191: Failed to execute ip(6)tables-restore command error=exit status 2 errorOutput="ip6tables-restore v1.8.2 (legacy): Bad IP address \"192.168.10.100\"\n\nError occurred at line: 2\nTry `ip6tables-restore -h' or 'ip6tables-restore --help' for more information.\n" input="*nat\n-R cali-nat-outgoing 1 -m comment --comment \"cali:ltvx3bX8efbq2000\" -m set --match-set cali60masq-ipam-pools src -m set ! --match-set cali60all-ipam-pools dst --out-interface bond+ --jump SNAT --to-source 192.168.10.100\nCOMMIT\n" ipVersion=0x6 output="" table="nat"
```
How to configure for FELIX_NATOUTGOINGADDRESS in IPv6 mode ?

spikecurtis:
@duylong there is an issue with this config option in dual stack mode. We don't have a way to specify both an IPv4 address and an IPv6 address.  What's happening is that `status.hostIP` is being listed as an IPv4 address, which ip6tables rightly rejects.

As a workaround, you could remove this config option entirely and Calico should default to using the address of the interface the packet is forwarded out of.

duylong:
I have several interfaces on my server in different VLANs. Without filtering and with the MASQUERADE mode by default, we give containers full access to the different routes, I wanted to avoid that. The idea of a single IP output does not suit me very much. According to you, is it possible to have a MASQUERADE according to the IP of the service?

spikecurtis:
If I understand correctly, you are attempting to use the choice of source IP in NAT as a way to limit which networks pods can send traffic to.  A more robust solution would be to define a [GlobalNetworkSet](https://docs.projectcalico.org/reference/resources/globalnetworkset) that represents the networks pods are allowed to access and then use Calico policy on pod egress to limit them to just those networks.

[...]

PR #6876, opened 2022-10-20
Egress SNAT for egress traffic from workload based on an annotation on the pod

sriramy:
## Description

As I mentioned in the issue 3275, this proposal provides a consistent egress IP for outbound connections from the pod without the need for an edge router.

A proposal to apply SNAT on egress traffic from workload towards external network (IP address that doesn't fall in any of the calico IP pools). Some issues that this addresses:
https://github.com/projectcalico/calico/issues/3275
https://github.com/projectcalico/calico/issues/5549

I tested this with Felix FV with the following tests:
- ETCD datastore with linux dataplane
- ETCD datastore with eBPF dataplane
- Kubernetes datastore with linux dataplane (Fails, and not sure why)
- Kubernetes datastore with eBPF dataplane

## Related issues/PRs

https://github.com/projectcalico/calico/issues/3275
https://github.com/projectcalico/calico/issues/5549

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


libcalico-go/lib/apis/v3/workloadendpoint.go
```diff
@@ -65,6 +65,10 @@ type WorkloadEndpointSpec struct {
 	// to connect one of its own external IPs. Each internal IP must be associated with the same
 	// endpoint via the configured IPNetworks.
 	IPNATs []IPNAT `json:"ipNATs,omitempty" validate:"omitempty,dive"`
+	// Egress SNAT is a 1:1 NAT mapping to apply for oubtound connections from the endpoint.
```
neiljerram, line 68: sp, should be "outbound"

felix/dataplane/linux/masq_mgr.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2021 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2021 Tigera, Inc. All rights reserved.Address
```
neiljerram, line 1: accidental paste here

felix/dataplane/linux/egress_snat_mgr.go
```diff
@@ -0,0 +1,121 @@
...
+// "masquerade" feature.  The feature adds a boolean flag to each IPAM pool, which controls how
+// outgoing traffic to non-Calico destinations is handled.  If the "masquerade" flag is set,
+// outgoing traffic is source-NATted to appear to come from the host's IP address.
+//
+// The egressSNATManager maintains two CIDR IP sets: one contains the CIDRs for all Calico
+// IPAM pools, the other contains only the NAT-enabled pools.
+//
+// When NAT-enabled pools are present, the egressSNATManager inserts the iptables masquerade rule
+// to trigger NAT of outgoing packets from NAT-enabled pools.  Traffic to any Calico-owned
+// pool is excluded.
```
neiljerram, line 37: Comment mostly wrong IIUC, needs rewriting from scratch.

felix/Makefile
```diff
@@ -275,7 +275,7 @@ ut combined.coverprofile: $(SRC_FILES) $(LIBBPF_A) build-bpf
 fv/fv.test: $(LIBBPF_A) $(SRC_FILES) $(FV_SRC_FILES)
 	# We pre-build the FV test binaries so that we can run them
 	# outside a container and allow them to interact with docker.
-	$(DOCKER_GO_BUILD_CGO) go test $(BUILD_FLAGS) ./$(shell dirname $@) -c --tags fvtests -o $@
+	$(DOCKER_GO_BUILD_CGO) go test $(BUILD_FLAGS) ./$(shell dirname $@) -c --tags nattests -o $@
```
neiljerram, line 278: Do we still want fvtests as well?

sriramy: This was supposed to be reverted, done now.

felix/dataplane/linux/egress_snat_mgr.go
sriramy: Done.

felix/dataplane/linux/masq_mgr.go
sriramy: Removed.

libcalico-go/lib/apis/v3/workloadendpoint.go
sriramy: Done.

sriramy:
Just a note that this solution has outstanding issues to be fixed, more here: [3275](https://github.com/projectcalico/calico/issues/3275#issuecomment-1288922328)

caseydavenport:
@sriramy thanks for the PR! 

We're discussing this one a bit within the maintainers team at the moment. We need to work through a few scenarios to make sure this solution is comprehensive enough and is something we are able to merge. We'll report back here with the outcome of that discussion.

I also saw your comment here https://github.com/projectcalico/calico/issues/3275#issuecomment-1288922328 - we will think about that. The multi-node scenario is one of the key limitations that we need to understand and that we were discussing earlier as well.

felix/rules/nat.go
```diff
@@ -151,6 +151,62 @@ func (r *DefaultRuleRenderer) SNATsToIptablesChains(snats map[string]string) []*
...
+		extIp := snats[intIp]
+		rules = append(rules, iptables.Rule{
+			Match:  iptables.Match().NotDestIPSet(allIPsSetName).SourceNet(intIp),
+			Action: iptables.SNATAction{ToAddr: extIp},
+		})
+	}
+	return []*iptables.Chain{{
+		Name:  ChainNATEgress,
+		Rules: rules,
+	}}
```
tomastigera, line 207: This seems to be duplicated by egressSNATsToIptablesChainsBPF. What if we just have the if bpf enabled here?

sriramy: Done. I had copied what was done in MakeNatOutgoingRule, but I agree it is better to check BPFEnabled in the same function.

sriramy:
> @sriramy thanks for the PR!
> 
> We're discussing this one a bit within the maintainers team at the moment. We need to work through a few scenarios to make sure this solution is comprehensive enough and is something we are able to merge. We'll report back here with the outcome of that discussion.
> 
> I also saw your comment here [#3275 (comment)](https://github.com/projectcalico/calico/issues/3275#issuecomment-1288922328) - we will think about that. The multi-node scenario is one of the key limitations that we need to understand and that we were discussing earlier as well.

Hi, wondering if there was some conclusion on this topic? :)

caseydavenport:
@sriramy sorry for the long delay, we've been a bit slow over the holidays. I don't think there's been much traction on this one I'm afraid.

mgleung:
Hey @sriramy, sorry for the delay still, I think we are still missing some of the discussion needed for this so I'll re-ping the team to get the conversation rolling.

sriramy:
No worries, we are still interested in the feature though. I would like to contribute if some way forward is suggested. :) Do you think something like Cilium Egress Gateway[1] or Antrea Egress[2] can be a better idea than what I have suggested above? 
[1]: https://docs.cilium.io/en/v1.10/gettingstarted/egress-gateway/
[2]: https://antrea.io/docs/v1.3.0/docs/egress/

xagent003:
I would like to see this scenario as well. We use Calico BGP to advertise an external IP for a service. It works great for ingress, the Pod is reachable by the external IP. For  egress it gets SNAT'd using the nodes IP. Would like for it to go out with the same external IP used for BGP advertisements and ingress.

For our usecase these are actually kubevirt VMs so there is only 1 Pod endpoint

There are some egress gateway solutions(Calico Enterprise has one) but thats not what I was looking for. That means all pods go to one egress vrouter and get NATd out using the same IP.

In our case each Pod (VM) has its own external IP that is externally routaBle, It may not always be the case theres an edge gateway for the nodes (so no Internet access, at least via the Calico interface). Our nodes only have Internet via a separate mgmt NIC. This is why node SNAT may not work.

As an example w/ openstack neutron, when a VM has a floating IP, it ingresses and egresses with the floating IP. The router does NAT both directions. It isnt: `"ingress on floating, but egress using node or some other  egress gateway"` Only VMs without a floating IP share a common snat egress gateway

In this K8s case the "floating IP" is our external IP on the Service(or external Loadbalancer IP)

mgleung:
Hey @sriramy, sorry for the delay again, I'll definitely keep you updated as we progress on this. The members of the team that would have the best insight into this have been focused on some other feature work so I'll try to get their attention on this. Thanks for your patience and we're looking forward to working with you on this.

@xagent003 could I ask you to clarify this portion of your use case?
> In our case each Pod (VM) has its own external IP that is externally routaBle, It may not always be the case theres an external gateway for the nodes (so no Internet access, at least via the Calico interface). Our nodes only have Internet via a separate mgmt NIC. This is why node SNAT may not work for VMs/Pods needing Internet.
My understanding is that if everything has to go through the separate NIC on your nodes, traffic would leave with that IP which seems pretty similar to the case where we're using the node's IP. Is the issue that those IPs differ and you're looking to choose one over the other?

xagent003:
> My understanding is that if everything has to go through the separate NIC on your nodes, traffic would leave with that IP which seems pretty similar to the case where we're using the node's IP. Is the issue that those IPs differ and you're looking to choose one over the other?

So, this is a multi-tenant multi-cluster cloud so all the nodes are locked down in different VLANs and VRFs at the TORs to provide isolation between different clusters. I am told right now there's an edge gateway device per, but, they want to get rid of it as it won't scale. On the other hand, another "mgmt interface" does exist, but its connected via an entirely different means, no K8s traffic should use it. I'm not a data center expert, so hope that made sense... But for now ignore it, imagine it doesn't exist :)

Now, suppose bond1 has an IP of 10.240.0.34. This is the interface we use for Calico. Traffic outbound gets SNAT'd using 10.24.0.34. Except there is no public gateway for this (it does not route to the Internet). Now we do have public IP space. We set up Calico BGP to peer with the TORs then assign an explicit public IP to the service like such:

```
apiVersion: v1
kind: Service
metadata:
  name: netdebug4
  annotations:
    metallb.universe.tf/loadBalancerIPs: 1.2.3.4
spec:
  selector:
    ngpc.pf9.io/vmService: vmcalicobridge2
  ports:
    - protocol: TCP
      port: 22
  externalTrafficPolicy: Local
  type: LoadBalancer
  #externalIPs:
    #- 1.2.3.4
 ```

The BGP works great, using either the externalIP on the Service, or using MetalLB (we still use Calico for the BGP, metalLB is only so we can set type: LoadBalancer and externalTrafficPolicy: Local). Either way above Service gets a public IP of 1.2.3.4. Inbound traffic is routed down to bond1 for the public IP, 1.2.3.4 in above example. And then kube-proxy NATs it to the Pod (Kubevirt VM).

On egress I wanted to also use 1.2.3.4 as source IP. Not any node IP.

The containers CIDR is not routable. The Services CIDR is not routable. The node IPs potentially not either. Which means egress(public, to the Internet) won't work unless it goes out using the public IP

mgleung:
hey @sriramy and @xagent003 , I touched base with the team and while there was a lot of discussion, I believe that one of the major hold ups is that we're not exactly sure how the changes in this PR ensure that reply traffic gets routed back to the pod. For floating IPs, Felix programs a local device route for each floating IP and BIRD propagates them in the same way it propagates pod IPs. To help us understand how this is supposed to work for these changes, maybe we need to back up and ask:

1. What is the pattern of SNAT IPs that are going to be assigned to pods when using this feature?
2. How do we arrange that traffic to one of those SNAT IPs (coming from outside the cluster) will be routed to the correct node?

sriramy:
> hey @sriramy and @xagent003 , I touched base with the team and while there was a lot of discussion, I believe that one of the major hold ups is that we're not exactly sure how the changes in this PR ensure that reply traffic gets routed back to the pod. For floating IPs, Felix programs a local device route for each floating IP and BIRD propagates them in the same way it propagates pod IPs. To help us understand how this is supposed to work for these changes, maybe we need to back up and ask:
> 
> 1. What is the pattern of SNAT IPs that are going to be assigned to pods when using this feature?
> 2. How do we arrange that traffic to one of those SNAT IPs (coming from outside the cluster) will be routed to the correct node?

Hi @mgleung, This PR was just an RFC patch really.
For a full solution, I was hoping to create a IPPool object for IP addreses that can be used for SNAT so the routes are exported via BGP. We could add a new allowedUse type called "SNAT" or something similar.

xagent003:
Hi matt, not sure I understand your questions...

I dont understand #1 what you mean by pattern? As I mentioned, we want it to be SNAT'd using the same externalIP assigned to the service. So packets will ingress and egress using the same IP. We take a public IPv4 address and add it to a Service as the externalIP

For #2, doesn't BGP take care of it? We advertise the external IP associated with a Service. The packet ingress's to that node. thereafter, kube-proxy "loadbalances" it to the Pod. Which in our case is a singular Kubevirt VM. Very likely on the same node because we want to use externalTrafficPolicy: local.

On egress, it would take the exact same path outbound... except it gets SNAT'd using the Service's external IP rather than the node IP on the NIC.

xagent003:
@mgleung @sriramy OVN-Kubernetes seems ot have something like this: https://github.com/ovn-org/ovn-kubernetes/blob/master/docs/egress-service.md

`The Egress Service feature enables the egress traffic of pods backing a LoadBalancer service to exit the cluster using its ingress IP. This is useful for external systems that communicate with applications running on the Kubernetes cluster through a LoadBalancer service and expect that the source IP of egress traffic originating from the pods backing the service is identical to the destination IP they use to reach them - i.e the LoadBalancer's ingress IP.`

caseydavenport:
Thank you everyone for the interest and discussion above, and apologies this hasn't come sooner. I think for now I am going to have to close this PR. There are a few reasons at play here:

- Our enterprise product has committed to a different approach to solving this problem via the use of egress gateways. 
- We aren't fully sure this approach is a complete solution. Specifically the concerns around multi-node clusters and reliance on host IP addresses for SNAT require additional work to make viable.
- This type of feature is - for now - out of scope for the Calico opens-source project, so our ability to invest time in fully building out this feature set is too limited.

Again, thanks all for your input and contributions above.

PR #6882, opened 2022-10-20
mark CALI_ST_SKIP_FIB packets on ingress of heps

tomastigera:
## Description
    felix/bpf: mark CALI_ST_SKIP_FIB packets on ingress of heps
    
    Disable FIB, let the packet go through the host after it is
    policed. It is ingress into the system and we do not know what
    exactly is the packet's destination. It may be a local VM or
    something similar and we let the host to route it or dump it.
    
    Fixes https://github.com/projectcalico/calico/issues/6450

    felix/bpf: log packet properties when CALI_SKB_MARK_BYPASS
    
    We want to be able to confirm which packet it is when logging. We skip
    the packet parsing otherwise.

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/6450

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: Fixes https://github.com/projectcalico/calico/issues/6450
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -68,6 +68,26 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 	 * skip all processing. */
 	if (CALI_F_FROM_HOST && skb->mark == CALI_SKB_MARK_BYPASS) {
 		CALI_INFO("Final result=ALLOW (%d). Bypass mark set.\n", CALI_REASON_BYPASS);
+		if  (CALI_LOG_LEVEL >= CALI_LOG_LEVEL_DEBUG) {
+			/* This generates a bit more richer output for logging */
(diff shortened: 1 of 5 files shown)


mazdakn (approved):
LGTM. Can we probably have a FV test to mimic this scenario?

tomastigera:
> LGTM. Can we probably have a FV test to mimic this scenario?

I was thinking about that and I was not quite sure how would we mimic that scenario. Afaict all we could do is the have a second device in felix and not include that route in the bpf routes. And then we could perhaps verify that such packet has the mark set - i guess by seeing the connection in conntrack. But I cannot see how we could repro the connectivity issue.

And this part would be much easier in UT and given that we had to fix the UTs I though it might be enough.

PR #6949, opened 2022-11-04
felix/bpf: pod->svc->self does not need allow policy

tomastigera:
We allow this traffic to match iptables dp and not require to have a special policy that allows traffic based on its host IP which is hard to express when pods start and go.

The check is done after we parsed the packet and we do it before policy and before CT. So we only do it for the first packet of a connection with minimal overhead.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: When a pod connects via a service to self, ingress traffic is policed as if it's source is the pod and not the host after MASQ
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -445,7 +445,8 @@ syn_force_policy:
 		}
 	}
 
-	if (CALI_F_TO_WEP && !skb_seen(ctx->skb) &&
+	if (CALI_F_TO_WEP && (!skb_seen(ctx->skb) ||
+			skb_mark_equals(ctx->skb, CALI_SKB_MARK_FROM_NAT_IFACE_OUT, CALI_SKB_MARK_FROM_NAT_IFACE_OUT)) &&
 			cali_rt_flags_local_host(cali_rt_lookup_flags(&ctx->state->ip_src))) {
 		/* Host to workload traffic always allowed.  We discount traffic that was
 		 * seen by another program since it must have come in via another interface.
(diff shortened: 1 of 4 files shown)


fasaxc (commented):
I think in iptables mode we do apply policy but the policy sees `srcIP=dstIP=podIP`.  That "just works" because filter.FORWARD is after DNAT but before SNAT.

Auto-allowing this traffic is better than requiring a policy (matches the CTLB behaviour) but might be best to be consistent with iptables if we can do it easily.  One fewer gotcha for E2Es and whatnot to worry about.

tomastigera:
> I think in iptables mode we do apply policy but the policy sees `srcIP=dstIP=podIP`. That "just works" because filter.FORWARD is after DNAT but before SNAT.

That makes sense, unfortunately, in our case we do the MASQ before the ingress policy back into the pod. We do not know the original src IP by that time. We may try to redirect on the same interface using BPF_F_INGRESS and do the MASQ there ourselves. It would go through the policy then. But it is untested.

tomastigera:
I think we can just do the policy against our own source IP as we know that the packet is from ourselves. We cannot restore the port but lets assume that it is most likely a random source port so it does not matter that we use the port after MASQ if it was changed.

We could use kfunc to lookup the MASQ conntrack and lookup the source port too, but that would only work on kernels newer than 5.18. Something we can do in a future iteration if the IP alone is a problem.

tomastigera:
Seems like not creating policy that would allow MASQ access to self, also found an issue in how we handle host access to local pods https://github.com/projectcalico/calico/pull/6949/commits/fd732e2f5512dbbc6eaad84797a96bd9b08a1822

marvin-tigera:

fasaxc (changes requested):
Looks like `calico_tc_skb_accepted` accesses state->ct_result` _before_ jumping to `calico_tc_skb_new_flow_entrypoint`; I think that'll be a problem now that that storage is shared?

tomastigera:
> Looks like `calico_tc_skb_accepted` accesses state->ct_result`_before_ jumping to`calico_tc_skb_new_flow_entrypoint`; I think that'll be a problem now that that storage is shared?

Yes, but since we had CT miss and hence it is a new flow, there is no data in that area which we might loose. Perhaps warrants a comment.

marvin-tigera:

marvin-tigera:

fasaxc (approved):

PR #6960, opened 2022-11-07
felix/bpf: tc.AttachPoint is always a pointer

tomastigera:
Some methods have pointer and some value receivers. It is unified now.

Given that it is passed in places as an interface, we do not really save the allocations and the size of the object keeps growing so no need to copy it for simple method calls.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


mazdakn (approved):
LGTM, but can you please also check if the same unification is needed for xdp.AttachPoint?

tomastigera:
> LGTM, but can you please also check if the same unification is needed for xdp.AttachPoint?

I had the impression the XDP uses pointers, but yeah, it is also needed there! Will add that.

PR #6965, opened 2022-11-08
DOCS-700: Changing 'else if' to 'elsif'

ctauchen:
We found a bug while doing migration work for https://tigera.atlassian.net/browse/DOCS-700

Jekyll, and Liquid, use "elseif", not "else if"

See: https://shopify.github.io/liquid/tags/control-flow/

Because of this, this include isn't being rendered properly in the current Jekyll sites. Changing to elsif fixes this problem.

See current example: https://projectcalico.docs.tigera.io/master/getting-started/bare-metal/requirements
With 'elsif' working properly: https://deploy-preview-6965--calico-master.netlify.app/getting-started/bare-metal/requirements

CC: @alexeymagdich-tigera


alexeymagdich-tigera (approved):
Thank you!

Issue #6710, opened 2022-09-14
Missing libnss_files causes localhost lookup issues in calico/typha:v3.24.1


wdoekes:
In `calico/typha:v3.24.1` docker image, there are only a handful libs in /lib64.
In `calico/typha:v3.23.3` there are a bunch more.

The lack of libnss_files causes localhost to get searched, which unfortunately resulted in a * CNAME being matched, causing localhost not to be 127.0.0.1.

## Expected Behavior

`/etc/hosts` should be read by libc.

## Current Behavior

If `/etc/resolv.conf` does not contain any `options`, things are fine:
- `/etc/hosts` is read
- `libnss_dns` is called (if needed)

If `/etc/resolv.conf` does contain any `options`, things break:
- `/etc/hosts` is NOT read
- `libnss_files` is tried but is unavailable in v3.24.1
- `libnss_dns` is called

This means that we got totally the wrong answer for `localhost` because it got translated to `localhost.nodedomain.` and that was a CNAME which resolved to REMOTE_IP.

This resulted in the health check not being able to bind on that remote IP:
```
health.go 296: Health endpoint failed, trying to restart it... error=listen tcp REMOTE_IP:9098: bind: cannot assign requested address
```

## Possible Solution

Add more `libnss_*` here:
https://github.com/projectcalico/calico/blob/65368ae5cf3abd20297a8226b6d94f9f1caf4b34/typha/docker-image/Dockerfile.amd64#L46-L50

Especially everything that you've put here:
https://github.com/projectcalico/calico/blob/65368ae5cf3abd20297a8226b6d94f9f1caf4b34/typha/docker-image/nsswitch.conf
```
hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4
```

Available in `registry.access.redhat.com/ubi8/ubi-minimal:8.6` are:
```
/lib64/libnss_compat.so.2
/lib64/libnss_dns.so.2
/lib64/libnss_files.so.2
/lib64/libnss_myhostname.so.2
/lib64/libnss_resolve.so.2
/lib64/libnss_systemd.so.2
```
I don't see any mdns4 or mdns4_minimal there, so you can drop those from nsswitch.conf.

## Example with no options in resolv.conf
```
[pid   106] 14:42:42.686147 openat(AT_FDCWD, "/etc/nsswitch.conf", O_RDONLY|O_CLOEXEC) = 8
[...]

PR #6971, opened 2022-11-09
typha: Fix missing nsswitch files causing localhost lookup fails

wdoekes:
## Description

Between the calico/typha:v3.23.5 and calico/typha:v3.24.0 Docker images, the included C libraries were reduced from many to only a handful. However, most prominently, it lacked libnss_files, which is needed to do proper lookups from /etc/hosts.

In some environments, things still worked. In others, the 'localhost' lookup would turn into a 'localhost.the-domain-the-pod-is-in' lookup, which could become something unexpected. For instance when '*' CNAMEs are used.

The result of the 'localhost' lookup is used to decide on which address the health check responder should bind. For 127.0.0.1, it would bind just fine. For an external IP, it would fail:

  health.go 296: Health endpoint failed, trying to restart it...
    error=listen tcp REMOTE_IP:9098: bind: cannot assign requested address

This in turn caused the pod to never become Ready.

Interestingly, the 'options' parameter in /etc/resolv.conf also affected the behaviour. Only when *any* 'options' was set AND the 'search' parameter was set, would this result in such a faulty lookup.

This changeset:

- removes the duplicate setting of nsswitch.conf
- sets only nsswitch libraries that are actually available

An alternative fix would be to run typha with GODEBUG=netdns=go forcing the use of the golang DNS subsystem. Using CGO_ENABLED=0 is not an option as typha relies on boringssl C libraries.

Fixes: #6710

Cheers,
Walter Doekes
OSSO B.V.

## Related issues/PRs

#6710 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix missing nsswitch files in Typha causing localhost lookup fails
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/typha/docker-image/nsswitch.conf
+++ b/typha/docker-image/nsswitch.conf
@@ -1,1 +1,1 @@
-hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4
+hosts: files dns
--- a/typha/docker-image/Dockerfile.amd64
+++ b/typha/docker-image/Dockerfile.amd64
@@ -46,20 +46,21 @@ COPY --from=ubi /usr/include /usr/include
 COPY --from=ubi /lib64/libpthread.so.0 /lib64/libpthread.so.0
 COPY --from=ubi /lib64/libc.so.6 /lib64/libc.so.6
 COPY --from=ubi /lib64/ld-linux-x86-64.so.2 /lib64/ld-linux-x86-64.so.2
-COPY --from=ubi /lib64/libnss_dns.so.2 /lib64/libnss_dns.so.2
 COPY --from=ubi /lib64/libresolv.so.2 /lib64/libresolv.so.2
+# These two libnss_* files are referenced in our local nsswitch.conf
+# They depend on some of the previous libs
+COPY --from=ubi /lib64/libnss_dns.so.2 /lib64/libnss_dns.so.2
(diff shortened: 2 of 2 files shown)


typha/docker-image/nsswitch.conf
```diff
@@ -1 +1,3 @@
-hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4
+# This is so minimal. We should've built the typha binary with
+# CGO_ENABLED=0 instead. See github.com/projectcalico/calico/issues/6710
```
caseydavenport, line 2: We need to build with CGO enabled since we are relying on the boringssl C library to provide FIPS compatible encryption, which isn't available otherwise. So I think we should remove this comment.

Otherwise the PR looks good to me.

caseydavenport (commented):
One comment but otherwise LGTM

typha/docker-image/nsswitch.conf
wdoekes: Ack. Updated the comment. Updated the commit-msg. Updated the PR-text.

wdoekes:
Thanks for the review. Review comments addressed :ok_hand:

caseydavenport (approved):

PR #6973, opened 2022-11-09
Typha graceful shutdown

fasaxc:
## Description

Add graceful shutdown to Typha, shed load at a configured rate during shutdown rather than dying immediately.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Typha now supports graceful shut down, disconnecting calico-node pods at a configured rate instead of all at once.
```

## Reminder for the reviewer

Other optional labels:


typha/pkg/syncserver/sync_server.go
```diff
@@ -481,6 +504,80 @@ func (s *Server) governNumberOfConnections(cxt context.Context) {
...
+	return s.shuttingDown
+}
+
+func (s *Server) TerminateRandomConnection(logCtx *log.Entry, reason string) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	for connID, conn := range s.connIDToConn {
+		logCtx.WithField("connID", connID).Infof("Closing connection; reason: %s.", reason)
+		conn.cancelCxt()
+		counterNumConnectionsDropped.Inc()
```
caseydavenport, line 576: Does it make sense to treat graceful termination connection drops as rebalancing? Or should we have a separate counter for them?

I'm not sure either way, but I can imagine some users wanting to distinguish between them.

caseydavenport (approved):
Looks nice!

typha/pkg/syncserver/sync_server.go
fasaxc: Moved this up to the caller; I think there's a good way to track shutdown already (num connections metric) so I've only kept the rebalancing one.

PR #6975, opened 2022-11-09
felix/bpf: tcpdump-like debug expressions

tomastigera:
## Description
    If BPFLogFilters is set only specified packets on specified devices
    trigger log actions. In any other case packets go through without the
    overhead of logging.
    
    A filter program is attached to the qdisc and that program decides
    whether packets should be processed by a set of programs with or without
    logging. This effectively means that there are two paths that a packet
    can take when logging is enabled. When logging is turned off, there is
    only a single path as was always the case.
    
    If BPFLogFilters is not specified, every program logs for every packet
    like the current firehose.
    
    CLTB logging must be explicitly enabled if BPFLogFilters is specified by
    including 'ctlb=on'.
    
    This feature is currently gated and needs to be enabled by adding
    BPFLogFilters=enabled into FeatureGates. Without that, log filters are
    ignored and only one set of (debug) programs is installed as was always
    the case.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


tomastigera:
Closing in favor of https://github.com/projectcalico/calico/pull/7607 which is based on the new code structure and programs layout based on https://github.com/projectcalico/calico/pull/7219

PR #7008, opened 2022-11-18
[BPF] detach cali programs from devices no longer in regex

tomastigera:
## Description

    [BPF] fv test for prog cleanups

    [BPF] detach cali programs from devices no longer in regex
    
    It can happen that the device regexp changes and calico programs then
    remain attached to devices that we do not observe anymore. For instance,
    user has the default ethX in the regexp, but those devices are bonded
    and the user does not have bondX in the regexp. The user fixes the
    regexp to include bondX only, but then the programs stay attached to
    ethX. User has to remove the programs manually or restart the node.
    
    We figure out that the programs on devices not in regexp are ours when
    those programs have a jump map pinned inthe calico bpffs hierarchy.
    
    This is not really a problem for workload devices as they usually
    disapear and programs get detached then.

    [BPF] add ListCalicoAttached() to list calico programs
    
    We need that to list programs only attached by calico and not attached
    to other ifaces by someone else. We may want to clean up someof these if
    we are not interested in a device anymore.

    [BPF] bpfEndpointManager.ensureStarted should  exec once in tests too.

    [BPF] cleaning up maps is for both TC and XDP
    
    Move it from tc package to bpf package with some code reorganization for
    simplicity.
    
    Also lock the clean up lock whenupdating XDP. It wasn't synced till now.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: cleanup previously attached programs when BPFDataIfacePattern changes.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/attach.go
+++ b/felix/bpf/attach.go
@@ -201,3 +201,74 @@ func sha256OfFile(name string) (string, error) {
 	}
 	return hex.EncodeToString(hasher.Sum(nil)), nil
 }
+
+// EPAttachInfo tells what programs are attached to an endpoint.
+type EPAttachInfo struct {
+	TCId    int
+	XDPId   int
+	XDPMode string
+}
+
+// ListCalicoAttached list all programs that are calico attached to TC or XDP
+// and are rlated to Calico. That is, they have jumpmap pinned in our dir
+// hierarchy.
+func ListCalicoAttached() (map[string]EPAttachInfo, error) {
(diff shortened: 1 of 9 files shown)


mazdakn (commented):
LGTM, but I am not confident enough with `dataplane/bpf_ep_manager` so maybe someone else should review as well.
I left few comments fixes.

mazdakn (approved):
LGTM

PR #7009, opened 2022-11-18
Docs-openstack-unsupported

bmckercher123:
### Add OpenStack unsupported for VXLAN and IP-in-IP

- Approved by Neil (https://tigera.slack.com/archives/CC08CBB43/p1668797414425539)
- Moved other limitation to align with standards

### DIFF
--- a/calico/networking/vxlan-ipip.md
+++ b/calico/networking/vxlan-ipip.md
@@ -35,6 +35,16 @@ Calico supports two types of encapsulation: VXLAN and IP in IP.  VXLAN is suppor
 
 Encapsulation of workload traffic is typically required only when traffic crosses a router that is unable to route workload IP addresses on its own. Calico can perform encapsulation on: all traffic, no traffic, or only on traffic that crosses a subnet boundary.
 
+### Before you begin
+
+**Not supported**
+
+- OpenStack
+
+**Limitations** 
+
+- IP in IP supports only IPv4 addresses
+
 ### How to
 
 - [Configure default IP pools at install time](#configure-default-ip-pools-at-install-time)
@@ -42,10 +52,6 @@ Encapsulation of workload traffic is typically required only when traffic crosse
 - [Configure IP in IP encapsulation for all inter workload traffic](#configure-ip-in-ip-encapsulation-for-all-inter-workload-traffic)
 - [Configure VXLAN encapsulation for only cross-subnet traffic](#configure-vxlan-encapsulation-for-only-cross-subnet-traffic)
 - [Configure VXLAN encapsulation for all inter workload traffic](#configure-vxlan-encapsulation-for-all-inter-workload-traffic)
-
-#### IPv4/6 address support
-
-IP in IP supports only IPv4 addresses.
 
 #### Best practice
 
@@ -152,4 +158,4 @@ If you use only VXLAN pools, BGP networking is not required. You can disable BGP
 
 ### Above and beyond
 
-For details on IP pool resource options, see [IP pool]({{ site.baseurl }}/reference/resources/ippool).
+For details on IP pool resource options, see [IP pool]({{site.baseurl}}/reference/resources/ippool).


neiljerram (approved):
LGTM, thanks!

mnaser:
@bmckercher123 & @nelljerram -- can we have some more context about why this is marked as "unsupported"?  AFAIK, OpenStack can carry `ipip` traffic just fine and we use it to deploy Calico with the Cluster API and Magnum.

nelljerram:
Hi @mnaser .  I just reviewed our discussions from the time, and they went like this:

> me: I feel embarrassed for asking this, but can anyone remind me if we support VXLAN or IPIP encaps with OpenStack? (and if so, how does that work without IPPools?)
> 
> colleague1:
> I don't think so, we never tied our IPAM to OpenStack's.  Maybe IPIP would "just work" if you created an IPPool that overlapped the OpenStack network?  I don't remember either!
> 
> colleague2:
> I thought we added encap after we changed to k8s. I.e openstack doesn't support encap. The way to check is to look at calico-test - if we don't have a way to set it up in there, it's not an openstack feature.

In other words, the concerns were:

1. Primarily, that we don't system test Calico with OpenStack and IPIP (or VXLAN) ourselves, and hence can't immediately be sure that it works like that, and doesn't accidentally regress when we make a new Calico release.  (Whereas we absolutely do cover this for Kubernetes.)

2. Secondly, uncertainties about how the IPAM would work.

I'm not sure precisely what the (2) concerns were, but thinking it through a bit now...

- OpenStack allocates an IP for each VM.

- Suppose VM A communicates with VM B, on a different compute host.  On A's host something needs to set up a routing rule like `<VM B IP> via <VM B's host's tunnel IP> dev tunl0 onlink` (and vice versa).  Our [BIRD fork](https://github.com/projectcalico/bird) can do that, but in our Kubernetes product it involves BIRD configuration that is generated by another part of the calico-node pod (confd), and that in turn depends on an IPPool having been configured with `ipipMode: Always` or `ipipMode: CrossSubnet`.  That all just works with Kubernetes, but (as far as I recall) we've never documented how someone would need to put together an equivalent solution for OpenStack.

Does that help?  How does it mesh with your experience?

mnaser:
Thank you for your response. There would be two ways that this can work:

1. Without encap (not really recommended): you can disable port security on the network or ports assigned to the VM (if you own the network the VMs are connected to).  By doing this, it will allow traffic to egress/ingress over the server with any src/dst IP which would allow native IP routing.  There is also an allowed address pairs settings which allows you to configure pairs of MAC addresses and CIDRs allowed on the host (a stricter version of disabling port security pretty much). I don't recommend/like this too much because you are opening up for other potential issues. 

2. Encap: OpenStack can transport encap traffic (in this case I know for a fact ipip traffic works). With this setup, everything works out of the box if we just allow ipip traffic. I suspect vxlan allowing the UDP port for it would also work. 

We deploy as part of our CI: Kubernetes with CAPI on top of OpenStack, install Calico and run Sonobuoy and everything passes all the time.

lwr20:
Hang on, it it possible that you are talking at cross purposes?  @nelljerram is talking about Calico being an neutron plugin for Openstack itself, while (IIUC) you are talking about running a kubernetes cluster using Calico CNI on Openstack VMs.

Those are quite different scenarios.

mnaser:
@lwr20 I am talking about a Kuberentes cluster using Calico CNI for Kubernetes running inside OpenStack VMs.

https://docs.tigera.io/calico/latest/networking/configuring/vxlan-ipip

Under the "Before you begin" section, it lists OpenStack as not supported.  Could that maybe be revised to say "Calico for OpenStack" since that seems to imply that you cannot run Calico on top of OpenStack VMs running Kubernetes.

lwr20:
> Could that maybe be revised to say "Calico for OpenStack" since that seems to imply that you cannot run Calico on top of OpenStack VMs running Kubernetes.

That would seem sensible, if @nelljerram confirms that I've interpreted what they're talking about correctly.

nelljerram:
Sorry for the delay.  Yes, my comments above were only for OpenStack in its own right.  As things stand, we don't have any documentation that specifically covers Calico CNI for Kubernetes inside OpenStack VMs
- either where we are just treating the OpenStack layer as generic infrastructure
- or in the special case where Calico is also being used as the Neutron plugin.

@mnaser I'll post a PR for your suggested clarification.

PR #7039, opened 2022-11-29
[BPF] no need for src fixup for packets from wl to tunnel

tomastigera:
Previously, we had to preserve the pod's source on the already encapped vxlan packet when returning to the tunnel because when FIB failed, the packet was dropped to host namespace and strict RPF on the pod's iface would drop it.

We no longer need that because (1) RPF is now relaxed since BPF enforces strict on its own as there are other cases when a pod replies with service IP and (2) we forward 100% of the packets straight to the external interface (from where the traffic originally came) to avoid issues with source routing in EKS (which no longer is a problem with the non-pod's source)

## Related issues/PRs
refs https://github.com/projectcalico/calico/issues/7040

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: no src fixup on host iface for traffic returning from pod to the nodeport tunnel
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/ipip_test.go
+++ b/felix/fv/ipip_test.go
@@ -37,7 +37,6 @@ import (
 	"github.com/projectcalico/api/pkg/lib/numorstring"
 
 	"github.com/projectcalico/calico/libcalico-go/lib/apiconfig"
-	libapi "github.com/projectcalico/calico/libcalico-go/lib/apis/v3"
 	client "github.com/projectcalico/calico/libcalico-go/lib/clientv3"
 	cerrors "github.com/projectcalico/calico/libcalico-go/lib/errors"
 	"github.com/projectcalico/calico/libcalico-go/lib/options"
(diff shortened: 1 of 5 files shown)


felix/fv/ipip_test.go
```diff
@@ -422,31 +398,52 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ IPIP topology before adding
...
+				}, "5s", "200ms").Should(Equal(3))
+			}
+
+			Eventually(func() error {
+				err := externalClient.ExecMayFail("ip", "tunnel", "add", "tunl0", "mode", "ipip")
+				if err != nil && strings.Contains(err.Error(), "SIOCADDTUNNEL: File exists") {
+					return nil
 				}
+				return err
+			}).Should(Succeed())
```
fasaxc, line 422: `Succeed()` TIL you can do `Should(Succeed())` instead of `ShouldNot(HaveOccurred())`

felix/fv/ipip_test.go
```diff
@@ -422,31 +398,52 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ IPIP topology before adding
...
+		AfterEach(func() {
+			externalClient.Stop()
+		})
 
 		It("should have all-hosts-net ipset configured with the external hosts and workloads connect", func() {
-			f := felixes[0]
-			// Add the ip route via tunnel back on the Felix for which we nuked when we removed its BGP spec.
-			f.Exec("ip", "route", "add", w[1].IP, "via", felixes[1].IP, "dev", "tunl0", "onlink")
-			cc.ExpectSome(w[0], w[1])
+			cc.ExpectSome(externalClient, w[0])
```
fasaxc, line 446: The key thing that we were testing was that, without an entry in the IP set, IPIP should be blocked.  So, suggest you add a couple of lines to this test that check that removing the IP from the config breaks connectivity.

fasaxc (changes requested):
Yep, adding the ext client was what I expected, just think we need the negative aspect of the test too to really verify the IP set is used to block unexpected IPIP.

felix/fv/ipip_test.go
tomastigera: sure I can add that :+1:

felix/fv/ipip_test.go
```diff
@@ -378,35 +377,57 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ IPIP topology before adding
...
+			for _, f := range felixes {
+				Eventually(func() int {
+					return getNumIPSetMembers(f.Container, "cali40all-hosts-net")
+				}, "5s", "200ms").Should(Equal(2))
 			}
 
+			cc.ExpectNone(externalClient, w[0])
+			cc.CheckConnectivity()
+
+			By("chding configuration to include the external client")
```
fasaxc, line 429:
```suggestion
			By("changing configuration to include the external client")
```

fasaxc (approved):

marvin-tigera:

PR #7042, opened 2022-11-30
add info to hostmetadataupdate

hedibouattour:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/calc/event_sequencer.go
```diff
@@ -100,6 +103,13 @@ type EventSequencer struct {
 	Callback EventHandler
 }
 
+type hostInfo struct {
+	ip4Addr  *net.IP
+	ip6Addr  *net.IP
+	labels   map[string]string
+	asnumber string
```
mgleung, line 110: This feels a little bit strange to me since in the Calico model, we attach the AS number to a BGP peer (which could be a host but conceptually feels like it encompasses a bit more than that). That being said, I'm not entirely sure if keeping track of BGP peers just to match hosts to an AS number is really worth it either. @caseydavenport WDYT?

caseydavenport: AS is a property of both the local node as well as the remote peer that we're connecting to - I believe this represents the local node's AS number, in which case it makes sense to go here IMO

PR #7043, opened 2022-11-30
Add compression to the Typha protocol.

fasaxc:
## Description

- Allow the decoder to be restarted by the server.  (This is a key building block for #7035.)
- When restarting decoder, allow a snappy compression reader to be included.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
The Typha protocol now supports compression.  This is enabled automatically if client and server both support it.
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #7047, opened 2022-12-01
Typha: share snapshots between clients

fasaxc:
Builds on #7043 

- When first client connects, calculate an in-memory
  compressed snapshot.
- If another client connects while the snapshot is still
  being built it shares that snapshot.
- Avoid blocking during the snapshot calculation, send
  the ready portion of the snapshot as it becomes available.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Typha now shares snapshots between clients that connect at roughly the same time. This dramatically reduces load when many clients connect at once.
```

## Reminder for the reviewer

Other optional labels:


typha/pkg/syncserver/snap_precalc.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+			currentWriteDeadline = newDeadline
+		}
+
+		n, err := reader.WriteTo(w)
+		if err != nil {
+			if n > 0 && os.IsTimeout(err) {
+				// Managed to write _some_ bytes, loop again to reset the timeout.  If the snapshot was
+				// very large then we might have written a big chunk of it but simply not had enough time to
+				// complete.  Only give up if we see no progress at all.
```
caseydavenport, line 265: Would it be worth a log here in case we're hitting this a lot? 

Seems like a branch that would in normal operation be pretty rare, but would be interesting to know about in the cases we are hitting it.

caseydavenport (approved):
A couple trivial comments, but otherwise LGTM.

Some of this went a bit over my head but I think I largely understand what's going on :sweat_smile:

marvin-tigera:

PR #7061, opened 2022-12-05
Improvements to Felix liveness/readiness reporting

fasaxc:
## Description

* Tune the liveness/readiness timeouts to avoid false positives.
* Reset liveness from within calc graph's tight loops.
* Add capability to override any problematic timeout.
  * Refactor configuration handling to make it easier to handle changes without restart.
* Make names of health probes uniform `CamelCase`

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Felix now supports overriding the timeouts of its internal readiness/liveness watchdog.  This is useful for dealing with issues "in prod" without needing a new release.  The timeouts have also been tuned to reduce false positives.
```

## Reminder for the reviewer

Other optional labels:


typha/pkg/snapcache/cache.go
```diff
@@ -152,7 +153,6 @@ type Cache struct {
 }
 
 const (
-	nameDefault    = "cache"
 	healthInterval = 10 * time.Second
```
song-jiang, line 156: healthInterval hasn't been replace here.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -249,6 +251,11 @@ type FelixConfigurationSpec struct {
 	HealthEnabled *bool   `json:"healthEnabled,omitempty"`
 	HealthHost    *string `json:"healthHost,omitempty"`
 	HealthPort    *int    `json:"healthPort,omitempty"`
+	// HealthTimeoutOverrides allows the internal watchdog timeouts of individual subcomponents to be
+	// overriden; example: "internal-dataplane-main-loop=30s,calculation-graph=2m".  This is useful for
+	// working around "false positive" liveness timeouts that can occur in particularly stressful workloads
+	// or if CPU is constrained.  For a list of active subcomponents, see Felix's logs.
+	HealthTimeoutOverrides string `json:"healthTimeoutOverrides,omitempty" validate:"omitempty,keyDurationList"`
```
caseydavenport, line 258: How feasible is it to make this a structured field?

For example:

```yaml
- name: InternalDataplaneMainLoop
  value: 30s
```
or

```
timeoutOverrides:
  internalDataplaneMainLoop: 30s
```

Or similar? I think these comma-separated string lists are fairly awkward to use. Their main benefit IIRC is that they are easier to encode into an env var, but we're pushing away from that being the primary means of configuring Felix.

I understand this might be a challenge given Felix's config merging machinery, but figured I'd ask!

song-jiang, line 258: Shall we make the the field a type which contains timeout values for all active components? 
```
type HealthTimeoutOverrides struct {
    CalculationGraph *metav1.Duration
    ....
```

The benefit is that the API is much clearer to user, and it you get the default FelixConfiguration, it will show you all the default values for timeout.

song-jiang (commented):
Looking good. Just one question on if we should put timeout settings into an API structure and a few nits.

caseydavenport (commented):
Some minor questions - I haven't gotten through all of it but I think @song-jiang is taking a closer look.

felix/calc/async_calc_graph.go
```diff
@@ -89,8 +89,9 @@ type AsyncCalcGraph struct {
 }
 
 const (
-	healthName     = "async_calc_graph"
+	healthName     = "calculation-graph"
 	healthInterval = 10 * time.Second
+	healthTimeout  = 30 * time.Second
```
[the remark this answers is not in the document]
fasaxc: I didn't rename it, I added a new constant for the timeout.  That was due to a change in an earlier revision; I could back it out but maybe it's clearer this way anyway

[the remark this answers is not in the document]
song-jiang: I see. My mistake.

api/pkg/apis/projectcalico/v3/felixconfig.go
fasaxc: I should be able to make it a list or a map but I'm reluctant to pre-define all the health report names since they're currently allocated in a dispersed way.  I can't easily write a UT to ensure that none are missed either since the ones that are active depend on build and run-time configuration.  

I envisioned this as a safety valve that we'd tell a user to set in an emergency rather than something that users would necessarily discover on their own.

fasaxc: (The main reason I didn't do it was naming convention mismatch with k8s; seemed a bit odd to use `lowerCamel` for health report names. Could live with UpperCamel, as Casey suggested.)

caseydavenport: > pre-define all the health report names since they're currently allocated in a dispersed way

I think I'm begrudgingly OK with the argument that this is an escape valve that is meant more for developers to help customers than it is for users to set themselves. The problem is that we mix those sorts of fields with proper user-facing fields in the same resource so it's not super obvious which are which.

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -589,6 +593,23 @@ func validateKeyValueList(fl validator.FieldLevel) bool {
 	return true
 }
 
+// validateKeyDurationList validates the field is a comma separated list of key=duration pairs.
+func validateKeyDurationList(fl validator.FieldLevel) bool {
```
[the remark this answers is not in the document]
fasaxc: Removed this but added UT for the new validation

api/pkg/apis/projectcalico/v3/felixconfig.go
fasaxc: Done, made it a slice

fasaxc: I made this a slice of override objects and improved the documentation

caseydavenport (approved):

marvin-tigera:

PR #7078, opened 2022-12-12
[v3.24] Add necessary libs to kube controllers

rene-dekker:


caseydavenport (approved):
LGTM - we should cherry-pick to master as well

PR #7080, opened 2022-12-13
[BPF] no need for CALI_SKB_MARK_BYPASS_FWD_SRC_FIXUP mark

tomastigera:
We do not have a usecase for fixing up host IP when a packet is leaving the host. We always set the source when we do vxlan encap as expected. Therefore we do not need to ever set the mark and check for it, which makes the code simpler a slightly faster.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -167,36 +167,6 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 			CALI_DEBUG("Packet approved for forward.\n");
 			COUNTER_INC(&ctx, CALI_REASON_BYPASS);
 			goto allow;
-		case CALI_SKB_MARK_BYPASS_FWD_SRC_FIXUP:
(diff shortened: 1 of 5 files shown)


felix/bpf-gpl/tc.c
```diff
@@ -166,36 +166,6 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 		case CALI_SKB_MARK_BYPASS_FWD:
```
mazdakn, line 166: Maybe change to an `if` statement since there is no other branch in `switch`?

mazdakn (approved):
LGTM, just one comment.

felix/bpf-gpl/tc.c
tomastigera: I thought about that because (1) smaller diff (2) may be handy in the future and (3) compiler will figure it out :-)

PR #7081, opened 2022-12-13
DOCS-436: Minor docs updates to the "Encrypt in-cluster pod traffic" topic

girishshankaran:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


girishshankaran:
@ctauchen Can you review this?

calico/security/encrypt-cluster-pod-traffic.md
```diff
@@ -22,17 +28,21 @@ This how-to guide uses the following {{site.prodname}} features:
...
+- GKE
+- Using your own custom keys to encrypt traffic
+
+**Limitations**
 
-- Kubernetes, on-premises
-- EKS using Calico CNI
-- EKS using AWS CNI
-- AKS using Azure CNI
+- IPv4 only
```
coutinhop, line 38: This should be removed as we now support IPv6

calico/security/encrypt-cluster-pod-traffic.md
```diff
@@ -22,17 +28,21 @@ This how-to guide uses the following {{site.prodname}} features:
...
+- EKS, only with AWS CNI
```
coutinhop, line 39: Is there a reason we're no longer claiming to support EKS + Calico CNI?

calico/security/encrypt-cluster-pod-traffic.md
```diff
@@ -22,17 +28,21 @@ This how-to guide uses the following {{site.prodname}} features:
...
+- AKS, only with Azure CNI
 
 **Supported encryption**
 
-- Host-to-host encryption for pod traffic
-- Encryption for direct node-to-node communication - supported only on managed clusters deployed on EKS and AKS
+- Pod-to-pod traffic
+- Encryption for direct node-to-node communication is supported only on managed clusters deployed on EKS (AWS CNI) and AKS (Azure CNI)
```
coutinhop, line 45: We'd need to update this line too if Calico CNI is to be supported for EKS

calico/security/encrypt-cluster-pod-traffic.md
```diff
@@ -22,17 +28,21 @@ This how-to guide uses the following {{site.prodname}} features:
 
 ### Before you begin...
 
-The following platforms:
+**Unsupported**
+
+- GKE
```
coutinhop, line 33: IMHO, switching from including all supported platforms to excluding unsupported ones is complicated, I don't claim to know this for certain and would ask @mgleung, @caseydavenport and/or @lwr20, would you please chime in regarding platforms we support (Calico in general, in addition to wireguard)?

My point being we're going from claiming to support Kubernetes onprem, EKS and AKS, to only not supporting GKE (and that might include things we're not anticipating). Again, I'm not the most knowledgeable in this, but aren't there other non-kubernetes platforms, for example? (on a tangent, besides k8s onprem, don't we support k8s + clouds such as gcp, aws and azure? or are those considered as the same as onprem?)

Sorry about the braindump here, I feel like I should be able to answer at least some of these questions (and I'll investigate documentation+code to do so), but I wanted to start a discussion...

lwr20: I agree this is a concern.  The thing about OSS is that new platforms pop up all the time, and users need to understand what to expect with them.  How can we possibly list the infinity of things we do not support when they're being created all the time?  

Its better to state the platforms we test regularly, so that users can know which platforms will "just work".  
Also useful to state platforms where we know it cannot work so they don't waste time trying.

girishshankaran: @bmckercher123 
Including Barbara in the discussion.

bmckercher123: You raise a good point about OSS and what "support" means. The doc was written from the perspective of Enterprise, which has changed to documenting the handful of features that we know do not work, or have significant concerns. For Enterprise and OSS, we cannot use the measure of "what we test regularly" because we simply do not test everything every release. Right now, we are following what we've done since the beginning by aligning the doc sets. I've created a ticket to revisit this with PM: https://tigera.atlassian.net/jira/software/projects/DOCS/boards/69/backlog?selectedIssue=DOCS-1072.

calico/security/encrypt-cluster-pod-traffic.md
bmckercher123: I thought IPv6/ipv4 support was only for host-to-host traffic. And that pod-to-pod was only IPv4. Check with Mike Stephens to verify. If true, then suggest this to be more clear:

**Limitations**

- ipv4 for pod-to-pod traffic
- ipv4/ipv6 for host-to-host traffic
- For EKS, TBD...if Sujeet says we support both, then delete this entry
- For AKS, only Azure CNI

calico/security/encrypt-cluster-pod-traffic.md
caseydavenport: Agree with @coutinhop and @lwr20.

lwr20: > For Enterprise and OSS, we cannot use the measure of "what we test regularly" because we simply do not test everything every release. 

No, we don't test every possible combination, but I think that's _expected_ in OSS.  What I'm proposing here is that we state what we test regularly and what we _know_ doesn't work, so that users can know how easy/difficult/impossible it will be to use whatever platform they're thinking of using.

calico/security/encrypt-cluster-pod-traffic.md
bmckercher123: Sujeet says EKS with Calico CNI is not supported. Again, until we revisit "support" for OSS versus CE with PM, I suggest we stay aligned with CE.

calico/security/encrypt-cluster-pod-traffic.md
bmckercher123: According to Sujeet, we do not support EKS with Calico CNI.

girishshankaran:
Can someone review this PR please? @coutinhop @lwr20 
@bmckercher123 Can you please check if I missed anything?

calico/security/encrypt-cluster-pod-traffic.md
```diff
@@ -11,8 +11,14 @@ Enable WireGuard to secure on-the-wire, in-cluster pod traffic in a {{site.prodn
...
 
-{{ site.prodname }} supports WireGuard encryption for both IPv4 and IPv6 traffic. These can be independently enabled in the FelixConfiguration resource: `wireguardEnabled`
-enables encrypting IPv4 traffic over an IPv4 underlay network and `wireguardEnabledV6` enables encrypting IPv6 traffic over an IPv6 underlay network.
+## Concepts
+### About WireGuard
+
+WireGuard supports both host-to-host encryption for pod traffic and direct node-to-node communication. Because {{site.prodname}} is not implemented using a sidecar, traffic is not encrypted for the full journey from one pod to another; traffic is only encrypted on the host-to-host portion of the journey.
```
lwr20, line 17: Could probably do with explaining why this isn't an issue.  "To be able to intercept pod-pod traffic on a node would require root access to the host's network namespace - an attacker with root on a host can do anything".

calico/security/encrypt-cluster-pod-traffic.md
lwr20: > According to Sujeet, we do not support EKS with Calico CNI.
There's confusion here, I think.  I think there are 2 features here.
- One encrypts ALL host-host traffic (including pod traffic) and is only available on EKS and AKS with their respective CNIs.
- The other is where we encrypt only pod-pod traffic between hosts, and that's available with Calico CNI on all supported platforms where we can run wireguard.

calico/security/encrypt-cluster-pod-traffic.md
bmckercher123: I would go with this for now and get this merged for the migration. I will ask Sujeet to confirm. (Sujeet has already confirmed we do not support EKS with Calico CNI in 3.15). 

**Limitations**

- Encryption for pod-to-pod traffic, ipV4 only
- Encryption for direct node-to-node communication (ipv4/ipv6) is supported only on EKS and AKS-managed clusters:
   - EKS, AWS CNI only
   - AKS, Azure CNI only

bmckercher123 (approved):
Approve with the limitations I noted (until I hear otherwise from Sujeet).

calico/security/encrypt-cluster-pod-traffic.md
girishshankaran: Adding this text:
Though there is unencrypted traffic between host-to-pod portion of the journey, attackers cannot easily intercept this traffic. To intercept the unencrypted traffic, they would need root access to the node.

calico/security/encrypt-cluster-pod-traffic.md
sujeet-kr: Based on a very recent round of testing (thanks @aaaaaaaalex ) I think Wireguard can be considered as supported for EKS + CalicoCNI.

girishshankaran: Thanks @sujeet-kr 
I'm removing "EKS with Calico CNI" from the Unsupported list.

lwr20 (approved):

PR #7083, opened 2022-12-13
Add necessary libs to kube controllers

rene-dekker:


kube-controllers/Dockerfile.amd64
```diff
@@ -47,12 +47,13 @@ COPY --from=ubi /lib64/libc.so.6 /lib64/libc.so.6
 COPY --from=ubi /lib64/ld-linux-x86-64.so.2 /lib64/ld-linux-x86-64.so.2
 
 # The following dependencies do not show up when performing ldd on the binary, but are still necessary.
-COPY --from=ubi /lib64/libresolv.so.2 /lib64/libresolv.so.2
```
caseydavenport, line 50: Are you sure we don't need this file?

kube-controllers/Dockerfile.amd64
```diff
@@ -47,12 +47,13 @@ COPY --from=ubi /lib64/libc.so.6 /lib64/libc.so.6
...
 COPY --from=ubi /lib64/libnss_dns.so.2 /lib64/libnss_dns.so.2
+COPY --from=ubi /lib64/libnss_files.so.2 /lib64/libnss_files.so.2
 
 # Copy hostname configuration files from UBI so glibc hostname lookups work.
 COPY --from=ubi /etc/host.conf /etc/host.conf
 COPY --from=ubi /etc/nsswitch.conf /etc/nsswitch.conf
+COPY --from=ubi /etc/networks /etc/networks
```
caseydavenport, line 56: What is the relevant file from within this dir that we need?

kube-controllers/Dockerfile.amd64
hjiawei: It is unlikely for `libc.so` to require `libresolv.so` to do domain resolution as `libresolv.so` links directly to `libc.so`. If this is the case, system loader has to do cycling loading libc/libresolv shared libraries. I believe `libnss_dns.so` is the plugin who does the generic resolving domain to address work.

caseydavenport (approved):

PR #7088, opened 2022-12-14
Add enPxxx pattern support to MTUIfacePattern

TrevorTaoARM:
For the Linux network device naming rules talked in [0],[1]: prefix [Pdomain] pbus sdevice [ffunction] [nphys_port_name | ddev_port]

The domain is omitted if it's 0 and the function is only present if it's a multi-function device. But for some network devices and/or Linux system, the domain is non-zero, so the pattern would be "enPxxxxxx".

On the other side, from the real case showed in our server, the ethernet interface name like "enP1p1s0f1" is valid as a normal data interface.

Now we lack this pattern in MTUIfacePattern, so when findHostMTU, this kind of interfaces would be skipped. It will affect the final MTU calculation for this miss.

[0] https://utcc.utoronto.ca/~cks/space/blog/linux/UdevNetworkDeviceNaming [1] https://www.freedesktop.org/software/systemd/man/systemd.net-naming-scheme.html

Signed-off-by: trevor tao <trevor.tao@arm.com>

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


caseydavenport (approved):

caseydavenport:
@TrevorTaoARM this LGTM, but there's now a conflict (sorry for the delay in reviewing).

I think this is also going to conflict with some documentation changes we're making on Monday, so we may need to revisit this later next week once that effort is done.

ctauchen:
@TrevorTaoARM @caseydavenport I've made the corresponding changes in the new docs repo here: https://github.com/tigera/docs/pull/244

PR #7093, opened 2022-12-15
[BPF] separate objects for v4/6 programs

tomastigera:
We do not need to load programs that we are not going to use. When needed, the v6 object is loaded separately and the programs get included in the jump map.

FV tests no longer load v6 programs, but the programs are tested with ap.AttachProgram in UTs. FVs should be a little faster to update endpoints. Nothing changes for production except that the code is never loaded into kernel - so a little faster too.

UTs do not excercise TestIPv6Parsing as that is a little bit pointless and fixing how the programs would load in UTs for BPF_RUN is only worth exploring with further progress on v6.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: ipv4 and ipv6 code separated to different object files so the v6 code gets never loaded outside tests.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/tc/attach.go
+++ b/felix/bpf/tc/attach.go
@@ -96,22 +96,11 @@ func (ap *AttachPoint) AlreadyAttached(object string) (int, bool) {
 	return -1, false
 }
 
-// AttachProgram attaches a BPF program from a file to the TC attach point
-func (ap *AttachPoint) AttachProgram() (int, error) {
-	logCxt := log.WithField("attachPoint", ap)
-
-	filename := ap.FileName()
-	binaryToLoad := path.Join(bpf.ObjectDir, filename)
-
-	progsToClean, err := ap.listAttachedPrograms()
+func (ap *AttachPoint) loadObject(ipVer int, file string) (*libbpf.Obj, error) {
+	obj, err := libbpf.OpenObject(file)
 	if err != nil {
-		return -1, err
+		return nil, err
 	}
-	obj, err := libbpf.OpenObject(binaryToLoad)
-	if err != nil {
-		return -1, err
-	}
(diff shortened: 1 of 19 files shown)


felix/bpf/tc/attach.go
```diff
@@ -96,43 +96,56 @@ func (ap *AttachPoint) AlreadyAttached(object string) (int, bool) {
...
+	if err != nil {
+		return nil, fmt.Errorf("error updating jump map %v", err)
+	}
+
+	return obj, nil
+}
+
+// AttachProgram attaches a BPF program from a file to the TC attach point
+func (ap *AttachPoint) AttachProgram() (int, error) {
+	logCxt := log.WithField("attachPoint", ap)
```
mazdakn, line 139: I think it's probably better to use a common method with ip version argument to not repeat the same code here. 
For example, `RememberAttachedProg` and other related functions save the name of object file and its hash to prevent loading an object if not changed. With the new approach, it only save the name of v4 object file and reloads v6 object every time.

I think to not miss other logic, it's probably better to use a loop and call a common function to take care of everything.

mazdakn (changes requested):
It looks great, I just found one issue in `ap.AttachProgram`

felix/bpf/tc/attach.go
tomastigera: If I understand you concern right, it is true that reloading the v6 objects is broken. However, I left it so intentionally because (a) it is never used in production (and now neither in fv tests) and (b) this is all going to change so I did not want to spend much time on it. What will remain is the split which I needed to prototype :)

mazdakn (approved):
LGTM

PR #7102, opened 2022-12-16
Fix race in typha FV tests during shutdown of client.

fasaxc:
Was using the same context for client and the state recorder. If the state recorder stopped while the client was calling its OnUpdates method then it could block forever.

## Description

Was using the same context for client and the state recorder. If the state recorder stopped while the client was calling its OnUpdates method then it could block forever.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/typha/fv-tests/server_test.go
+++ b/typha/fv-tests/server_test.go
@@ -642,7 +642,9 @@ var _ = Describe("With an in-process Server with short ping timeout", func() {
 		)
 		err := client.Start(clientCxt)
 		Expect(err).NotTo(HaveOccurred())
-		go recorder.Loop(clientCxt)
+		recorderCtx, recorderCancel := context.WithCancel(context.Background())
+		defer recorderCancel()
+		go recorder.Loop(recorderCtx)
 		defer func() {
 			clientCancel()
 			client.Finished.Wait()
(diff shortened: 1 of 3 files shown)


typha/fv-tests/server_test.go
```diff
@@ -901,7 +905,9 @@ var _ = Describe("With an in-process Server with long ping interval", func() {
 			},
 		)
 		err := client.Start(clientCxt)
-		go recorder.Loop(clientCxt)
+		recorderCtx, recorderCancel := context.WithCancel(context.Background())
+		defer recorderCancel()
+		go recorder.Loop(recorderCtx)
```
tomastigera, line 910: wouldn't it make more sense to start it after checking the Start() error `Expect(err).NotTo(HaveOccurred())` may not make any difference with the defer though.

tomastigera (approved):
LGTM

typha/fv-tests/server_test.go
fasaxc: Oh yeah, that would make sense (was just doing a handle-turn find/replace)

fasaxc: Gonna leave this PR as is because I've already backported to enterprise; will put up a follow-on PR

PR #7106, opened 2022-12-19
Separate calico-node and calico-cni-plugin service accounts

MichalFupso:
## Description

Split of the calico-node account to calico-node and calico-cni-plugin service account.
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Separate calico-node and calico-cni-plugin service accounts
```

## Reminder for the reviewer

Other optional labels:


charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -176,6 +141,61 @@ rules:
 
 ---
 
+metadata:
+  name: calico-cni-plugin
+rules:
+  # Used for creating service account tokens to be used by the CNI plugin
+  - apiGroups: [""]
+    resources:
+      - serviceaccounts/token
```
caseydavenport, line 150: This permission should only be on the calico-node serviceaccount, not on the CNI plugin

charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -176,6 +141,61 @@ rules:
 
 ---
 
+metadata:
```
caseydavenport, line 144: You're missing kind / apiVersion fields on this guy

charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -138,26 +123,6 @@ rules:
       - create
       - update
 {{- if eq .Values.network "calico" }}
-  # These permissions are required for Calico CNI to perform IPAM allocations.
```
caseydavenport, line 141: I think with testing you'll find that calico/node also needs some of these permissions.

charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -17,14 +17,6 @@ rules:
 {{- end }}
     verbs:
       - create
-  # The CNI plugin needs to get pods, nodes, and namespaces.
```
caseydavenport, line 20: I think calico/node also needs some of these, unless they are already covered somewhere else.

charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -138,26 +123,6 @@ rules:
...
-      - blockaffinities
-      - ipamblocks
-      - ipamhandles
-    verbs:
-      - get
-      - list
-      - create
-      - update
-      - delete
-  # The CNI plugin and calico/node need to be able to create a default
```
caseydavenport, line 153: This comment suggests calico-node still needs the IPAMConfigs permission

node/pkg/cni/token_watch.go
```diff
@@ -217,7 +218,7 @@ func Run() {
 	if err != nil {
 		logrus.WithError(err).Fatal("Failed to create in cluster client set")
 	}
-	tr := NewTokenRefresher(clientset, NamespaceOfUsedServiceAccount(), CNIServiceAccountName())
+	tr := NewTokenRefresher(clientset, NamespaceOfUsedServiceAccount(), NodeServiceAccountName())
```
caseydavenport, line 221: The goal of this PR is so that calico/node is generating new tokens for the CNI plugin's service account, so I think we want to leave this line unchanged.

node/pkg/cni/token_watch.go
```diff
@@ -244,7 +245,15 @@ func CNIServiceAccountName() string {
 		logrus.WithField("name", sa).Debug("Using service account from CALICO_CNI_SERVICE_ACCOUNT")
 		return sa
 	}
-	return defaultServiceAccountName
+	return defaultCNIPluginAccountName
+}
+
+func NodeServiceAccountName() string {
```
caseydavenport, line 251: I don't think we need this function - we shouldn't ever be doing any actions on the calico-node serviceaccount.

caseydavenport (changes requested):
A few minor comments. I think you'll want to do a few things:

- Test this on a live cluster to make sure all the permissions are correct (check the calico-node and CNI plugin logs for unauthorized permissions errors)
- Run `make generate` to update the manifests - you changed the templates in the charts/ dir, but the artifacts need re-rendering afterwards.
- You will need to make the equivalent changes in the tigera/operator repository, since most users install Calico that way nowadays. https://github.com/tigera/operator/blob/master/pkg/render/node.go#L287

charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -12,8 +12,10 @@ rules:
     resourceNames:
 {{- if eq .Values.network "flannel" }}
       - canal
+      - canal-cni-plugin
```
caseydavenport, line 15: A bit of a nit, but I'd just call this `calico-cni-plugin` since under the covers it's still using the Calico CNI plugin, even for Canal.

charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -76,7 +78,7 @@ rules:
     verbs:
       - list
       - watch
-  # The CNI plugin patches pods/status.
+#  # The CNI plugin patches pods/status.
```
caseydavenport, line 81: Extra `#` here

charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -176,6 +178,47 @@ rules:
 
 ---
 
+#CNI cluster role
```
caseydavenport, line 181: nit, but could you keep the comment styling consistent with the others?

basically:

```
# CNI cluster role
```

instead of 

```
#CNI cluster role    
```

charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -227,6 +270,19 @@ subjects:
 - kind: ServiceAccount
   name: canal
   namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: canal-cni-plugin
```
caseydavenport, line 277: Same here - I think if we just keep the name the same (calico-cni-plugin) then we don't need two separate ClusterRoleBindings - one for Canal and one for Calico - we can just use a single one that works for both.

node/tests/k8st/infra/calico-kdd.yaml
```diff
@@ -29,6 +29,13 @@ metadata:
   name: calico-node
   namespace: kube-system
 ---
+# Source: calico/templates/calico-node.yaml
```
caseydavenport, line 32: I was expecting to see this change in `manifests/calico.yaml` and other files in the `manifests/` directory as well.

You probably need to update `charts/calico/calico-node.yaml` to include this serviceaccount, and then run `make generate`

node/pkg/cni/token_watch.go
```diff
@@ -238,7 +238,7 @@ func Run() {
 }
 
 // CNIServiceAccountName returns the name of the serviceaccount to use for the CNI plugin token request.
-// This can be set via the CALICO_CNI_SERVICE_ACCOUNT environment variable, and defaults to "calico-node" otherwise.
+// This can be set via the CALICO_CNI_SERVICE_ACCOUNT environment variable, and defaults to "calico-cni-plugin" otherwise.
```
caseydavenport, line 241: I think it would be good to add a unit test or a new assertion to `token_watch_test.go` to verify that the returned token has the correct serviceaccount information in it. 

You can take a ServiceAccount and base64 decode it. It's a JWT that will include information about the service account to which the token is bound.

caseydavenport (changes requested):

manifests/tigera-operator.yaml
```diff
@@ -5,6 +5,28 @@ metadata:
   labels:
     name: tigera-operator
 ---
+# Source: crds/calico/_.yaml
```
caseydavenport, line 8: I think this shouldn't be here - probably generated by mistake? looks like a stray `_.yaml` file somewhere

caseydavenport (approved):

caseydavenport (approved):

charts/calico/templates/calico-node-rbac.yaml
```diff
@@ -242,3 +284,15 @@ subjects:
   name: calico-node
   namespace: kube-system
 {{- end }}
+kind: ClusterRoleBinding
```
caseydavenport, line 287: I think this can be a `RoleBinding` since it's only for a single serviceaccount in one namespace.

caseydavenport: In your other PR, this is a RoleBinding: https://github.com/tigera/operator/pull/2393/files#diff-d4d31d15e69aa9a0bad7776e758dfbd50b8e0e1a969fd16719f56c470cac4e7bR316

caseydavenport (approved):

PR #7114, opened 2022-12-20
add post release tests for manifests archive contents

radTuti:
## Description

This adds tests to verify the manifests of the archive included in a release.

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/hack/postrelease/test_manifests_archives.py
+++ b/hack/postrelease/test_manifests_archives.py
@@ -1,0 +1,54 @@
+import requests
+import tarfile
+from nose.tools import with_setup
+
+from versions import RELEASE_VERSION
+
+url = "https://github.com/projectcalico/calico/releases/download/{v}/release-{v}.tgz".format(
+    v=RELEASE_VERSION
+)
+
+manifest_list = [
+    "calico.yaml",
+    "calico-etcd.yaml",
+    "calico-bpf.yaml",
+    "calico-typha.yaml",
+    "calico-vxlan.yaml",
+    "calico-windows-bgp.yaml",
+    "calico-windows-vxlan.yaml",
+    "calicoctl.yaml",
+    "calicoctl-etcd.yaml",
+    "canal.yaml",
+    "canal-etcd.yaml",
+    "tigera-operator.yaml",
+    "custom-resources.yaml",
+]
+
+
+def setup_archive():
+    response = requests.get(url, stream=True)
+    global file
+    file = tarfile.open(fileobj=response.raw, mode="r|gz")
+
+
+def teardown_archive():
+    file.close()
+
+
+@with_setup(setup=setup_archive, teardown=teardown_archive)
+def test_manifest_present():
+    for manifest in manifest_list:
+        yield check_manifest_present, manifest
+
+
+def check_manifest_present(manifest):
+    print("[INFO] checking {} is in archive".format(manifest))
+    try:
+        manifest_info = file.getmember(
+            "release-{v}/manifests/{m}".format(v=RELEASE_VERSION, m=manifest)
+        )
+        print(manifest_info.name, manifest_info.size)
+        assert manifest_info.isfile()
+        assert manifest_info.size > 100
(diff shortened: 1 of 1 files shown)


hack/postrelease/test_manifests_archives.py
```diff
@@ -0,0 +1,54 @@
...
+    "calicoctl.yaml",
+    "calicoctl-etcd.yaml",
+    "canal.yaml",
+    "canal-etcd.yaml",
+    "tigera-operator.yaml",
+    "custom-resources.yaml",
+]
+
+
+def test_manifest_archive_on_release_page():
```
lwr20, line 32: This works fine as-is, how hard would it be to make this a parameterised test (like in `test_tags_tars`).  That way the output is more explicit about exactly which files have been found/not found in the archive.

lwr20 (approved):
Approved, one suggestion for improving.

hack/postrelease/test_manifests_archives.py
```diff
@@ -0,0 +1,54 @@
+import requests
+import tarfile
+
+from versions import RELEASE_VERSION
+
+urls = {
+    "docs": "https://github.com/projectcalico/calico/releases/download/{v}/release-{v}.tgz".format(
```
caseydavenport, line 7: These are both the same URL - is there a reason we need both?

radTuti: oversight

hack/postrelease/test_manifests_archives.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def teardown_archive():
+    file.close()
+
+
+@with_setup(setup=setup_archive, teardown=teardown_archive)
+def test_manifest_present():
+    for manifest in manifest_list:
+        yield check_manifest_present, manifest
```
caseydavenport, line 41: I don't think a `yield` is the right execution method here - there's never a situation where we're wanting to pause execution of check_manifest_present and then start where we left off, right?

Probably better to just use a function call - `check_manifest_present(manifest)`

radTuti: i'm using nosetest[ test generators](https://nose.readthedocs.io/en/latest/writing_tests.html#test-generators) which creates individual tests for each manifest - more explict with failure

a function call gives this result summary

```sh
test_manifests_archives.test_manifest_present ... ok (23.9096s)
```

comapred to a yieded tests

```sh
test_manifests_archives.test_manifest_present('calico.yaml',) ... ok (23.4394s)
test_manifests_archives.test_manifest_present('calico-etcd.yaml',) ... ok (0.0001s)
test_manifests_archives.test_manifest_present('calico-bpf.yaml',) ... ok (0.0003s)
test_manifests_archives.test_manifest_present('calico-typha.yaml',) ... ok (0.0002s)
test_manifests_archives.test_manifest_present('calico-vxlan.yaml',) ... ok (0.0002s)
test_manifests_archives.test_manifest_present('calico-windows-bgp.yaml',) ... ok (0.0003s)
test_manifests_archives.test_manifest_present('calico-windows-vxlan.yaml',) ... ok (0.0002s)
test_manifests_archives.test_manifest_present('calicoctl.yaml',) ... ok (0.0002s)
test_manifests_archives.test_manifest_present('calicoctl-etcd.yaml',) ... ok (0.0002s)
test_manifests_archives.test_manifest_present('canal.yaml',) ... ok (0.0002s)
test_manifests_archives.test_manifest_present('canal-etcd.yaml',) ... ok (0.0002s)
test_manifests_archives.test_manifest_present('tigera-operator.yaml',) ... ok (0.0002s)
test_manifests_archives.test_manifest_present('custom-resources.yaml',) ... ok (0.0002s)
```

caseydavenport: Ahhh, gotcha. Unfortunate that we still have Python tests hanging around but maybe one day we'll switch these all over to go. Good for now.

PR #7116, opened 2022-12-21
Add note to help debug apiserver error

perithompson:
## Description
This PR adds a note to the documentation which explains an error that occurs when using kubectl with the calico api-resources.  If the API server cannot talk to calico on port 5443 a user will receive a kubectl error, it is worth pointing out where the error is likely to be as it isn't clear in any logs what is going on, especially when using a managed cluster and the Kube-apiserver logs are not available.

## Related issues/PRs
N/A

## Todos

- [ ] Tests
- [x] Documentation
- [ ] Release note

## Release Note

N/A

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/calico/maintenance/install-apiserver.md
+++ b/calico/maintenance/install-apiserver.md
@@ -132,6 +132,8 @@ After following the above steps, you should see the API server pod become ready,
    profiles                                                                          projectcalico.org              false        Profile
    ```
 
+If the above command returns `error: unable to retrieve the complete list of server APIs: projectcalico.org/v3: the server is currently unable to handle the request` you should check that the `kube-apiserver` is able to communicate with the `calico-apiserver` on TCP 5443.
+ 
 > **Note:** kubectl may continue to prefer the crd.projectcalico.org API group due to the way it caches APIs locally. You can force kubectl to update
 >           by removing its cache directory for your cluster. By default, the cache is located in `$(HOME)/.kube/cache`.
 {: .alert .alert-info}


calico/maintenance/install-apiserver.md
```diff
@@ -132,6 +132,8 @@ After following the above steps, you should see the API server pod become ready,
    profiles                                                                          projectcalico.org              false        Profile
    ```
 
+If the above command returns `error: unable to retrieve the complete list of server APIs: projectcalico.org/v3: the server is currently unable to handle the request` you should check that the `kube-apiserver` is able to communicate with the `calico-apiserver` on TCP 5443.
```
mgleung, line 135: nit: I think that since this is a troubleshooting step, it might be good to put in a `**Note:** {: .alert .alert-info}` section

caseydavenport: I'm not sure it is appropriate at all here to be honest, it might be better placed in our troubleshooting documentation rather than this section. 

The install page isn't really the right location for arbitrary debugging information. Rather, we should make sure any networking requirements are laid out in the requirements and prerequisites documentation.

caseydavenport:
We're moving our docs  to a new location in the imminent future, and I think that this content probably belongs elsewhere (in a prerequisite or troubleshooting page) so I'm going to close this PR for now.

Please feel free to re-open this against github.com/tigera/docs at the end of next week and we can revisit the discussion, though

PR #7131, opened 2022-12-26
cleanup: remove old apiversion for deployment

my-git9:
## Description
The apiversion for `extensions/v1beta1` has been deprecated for a long time, mybe we should remove it.

### DIFF
--- a/app-policy/config/demo/10-yaobank.yaml
+++ b/app-policy/config/demo/10-yaobank.yaml
@@ -32,11 +32,15 @@ metadata:
   labels:
     app: yaobank
 ---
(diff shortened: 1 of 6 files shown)


caseydavenport:
@my-git9 have you tried deploying these manifests? ISTR that there were other changes required when upgrading from v1beta1 to v1

my-git9:
> @my-git9 have you tried deploying these manifests? ISTR that there were other changes required when upgrading from v1beta1 to v1

Updated. Thanks

caseydavenport (approved):

PR #7148, opened 2023-01-05
Fix incorrect map cleanup in service policy index.

fasaxc:
## Description

There was a bug in this scenario:

- Ingress and egress rules both become active, using the same service.
- Either rule becomes inactive.
- Service index would clean up the endpoints to IP set mapping for both IP sets.
- Updates to endpoint slices were then ignored for the remaining IP set.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Fix incorrect cleanup in the service policy index after having both ingress and egress rules that reference the same service, resulting in missed IP set updates after one rule was deactivated.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/calc/calc_graph_fv_test.go
+++ b/felix/calc/calc_graph_fv_test.go
@@ -421,6 +421,15 @@ var baseTests = []StateList{
 	{
 		endpointSliceActiveSpecNoPorts,
 	},
+	{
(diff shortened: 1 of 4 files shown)


mgleung (approved):
LGTM

PR #7176, opened 2023-01-11
Reload Typha addresses between connection attempts.

fasaxc:
## Description

Rework Typha discovery to avoid a long timeout if we happen to race with a Typha upgrade.  Previously we could:

- Load the list of typhas, just before...
- ...all the typhas start shutting down to be upgraded
- Attempt to connect to all the dead typhas one by one.
- Only once we'd timed out trying to connect to each gone Typha, give up and restart.  This could take 10s * number of Typhas; not the end of the world but also much longer than desired.

This change makes sure that we reload the list of Typhas between calls:

- Extract out a Discoverer object to keep track of the discovery parameters.
- Have the discoverer call back to initiate wireguard bootstrap handling.
- Replace loop over slice of Typhas with an object that handles refreshing the list of Typhas and avoiding connecting to the same one twice.

## Todos

- [x] Tests
- [x] Release note

## Release Note

```release-note
Fix that, if a Typha client loads the list of Typha instances just before they all get upgraded, it takes 30s+ to time out.  Reload the list of Typha instances between each connection attempt.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/clientv3/interface.go
+++ b/libcalico-go/lib/clientv3/interface.go
@@ -21,45 +21,25 @@ import (
 )
 
 type Interface interface {
-	// Nodes returns an interface for managing node resources.
-	Nodes() NodeInterface
-	// GlobalNetworkPolicies returns an interface for managing global network policy resources.
-	GlobalNetworkPolicies() GlobalNetworkPolicyInterface
-	// NetworkPolicies returns an interface for managing namespaced network policy resources.
-	NetworkPolicies() NetworkPolicyInterface
-	// IPPools returns an interface for managing IP pool resources.
-	IPPools() IPPoolInterface
-	// IPReservations returns an interface for managing IP reservation resources.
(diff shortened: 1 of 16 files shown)


libcalico-go/lib/clientv3/interface.go
```diff
@@ -21,53 +21,128 @@ import (
 )
 
 type Interface interface {
+	NodesClient
```
fasaxc, line 24: I split out these smaller interfaces to make mocking easier.  Then I refactored the code and I didn't need to mock the NodeClient any more.  Nevertheless, it seemed like a good idea to make mocking easier in future...

electricjesus (approved):
Looks good to me! The integration with the existing bootstrap for WireGuard has been done quite nicely. There shouldn't be any concern if the post-filter discovery function ends up calling bootstrap (that will remove WireGuard config) more than once.

PR #7200, opened 2023-01-16
[BPF] sync ifstate upon startup

tomastigera:
Atm we delete all entries for devices that we have not seen updated by the first CompleteDeferredWork() execution. This should preserve the entries for devices that we have not seen the update yet working correctly - as they did while Felix was restarting.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -804,6 +804,25 @@ func (m *bpfEndpointManager) markExistingWEPDirty(wlID proto.WorkloadEndpointID,
 	}
 }
 
+func (m *bpfEndpointManager) syncIfStateMap() {
+	m.ifStateMap.IterDataplaneCache(func(k ifstate.Key, v ifstate.Value) {
+		ifindex := int(k.IfIndex())
+		_, err := net.InterfaceByIndex(ifindex)
+		if err != nil {
+			// "net" does not export the strings or err types :(
+			if strings.Contains(err.Error(), "no such network interface") {
+				m.ifStateMap.DeleteDesired(k)
+			} else {
+				// It will get deleted by the first CompleteDeferredWork() if we
+				// do not get any state update on that interface.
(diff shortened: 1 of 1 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -804,6 +804,18 @@ func (m *bpfEndpointManager) markExistingWEPDirty(wlID proto.WorkloadEndpointID,
 	}
 }
 
+func (m *bpfEndpointManager) syncIfStateMap() {
+	m.ifStateMap.IterDataplaneCache(func(k ifstate.Key, v ifstate.Value) {
+		ifindex := int(k.IfIndex())
+		_, err := net.InterfaceByIndex(ifindex)
+		if err != nil {
+			m.ifStateMap.DeleteDesired(k)
```
fasaxc, line 812: Since this is a one-off, I'd log this at Info.  Also, worth doing this so we'll have something to go on if we ever get an unexpected error here:
```
if err != nil {
  if os.IsNotExists() {  // Or whatever the right check is
    log.Info...
  } else {
    log.WithError(err).Error(...
  }
```

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1762,6 +1778,10 @@ func (m *bpfEndpointManager) ensureStarted() {
 	if err != nil {
 		log.WithError(err).Warn("Failed to list previously attached programs. We may not clean up some.")
 	}
+
+	if err := m.ifStateMap.LoadCacheFromDataplane(); err != nil {
+		log.WithError(err).Fatal("Cannot load interface state map - essential for consisten operation.")
```
fasaxc, line 1783:
```suggestion
		log.WithError(err).Fatal("Cannot load interface state map - essential for consistent operation.")
```

fasaxc (approved):
Couple of suggestions but don't think I need to re-review

tomastigera:
:red_circle: because of a flake

PR #7211, opened 2023-01-19
[BPF] DSR optout CIDRs

tomastigera:
Setting BPFDSROptoutCIDRs to a list of CIDRs allows clients from these CIDRs to opt out from DSR when DSR is enabled. That allows to use DSR in the part of your network where it works and keep using standard nodeports in that part of the network where it is not possible to use DSR.

BPFDSROptoutCIDRs can overlaps with other CIDRs

For instance 0.0.0.0/0 would optout all clients from DSR. Those CIDRs
must be either disjoint with other CIDRs in the cluster or their
super-sets. That means, for instance, if an IP pool overlaps only
partially with the opt out CIDR, the optout may or may not be applied as
it is applied to the route blocks on each node, but does not split those
blocks.

The BPFDSROptoutCIDRs is meant to be used with external entities to the 
cluser. It has no effect on in-cluster entities since DSR does not 
have much meaning fot them anyway.

A useful case may be to apply this to external nodes, provider specific
ip like AKS healt check IPs etc.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: Setting BPFDSROptoutCIDRs to a list of CIDRs allows clients from these CIDRs to opt out from DSR when DSR is enabled.
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -381,6 +381,10 @@ type FelixConfigurationSpec struct {
 	// is sent directly from the remote node.  In "DSR" mode, the remote node appears to use the IP of the ingress
 	// node; this requires a permissive L2 network.  [Default: Tunnel]
 	BPFExternalServiceMode string `json:"bpfExternalServiceMode,omitempty" validate:"omitempty,bpfServiceMode"`
+	// BPFDSROptoutCIDRs is a list of CIDRs which are excluded from DSR. That is, clients
+	// in those CIDRs will accesses nodeports as if BPFExternalServiceMode was set to
+	// Tunnel.
+	BPFDSROptoutCIDRs *[]string `json:"bpfDSROptoutCIDRs,omitempty" validate:"omitempty,cidrs"`
```
fasaxc, line 387: Was surprised to see the pointer to a slice (slices already have a nil value) but it seems like we do it elsewhere :thinking: . Maybe it allows for a node-specific felix config to override the global one to `nil`.  Not sure that will actually work with omitempty etc.

felix/bpf-gpl/globals.h
```diff
@@ -27,8 +27,9 @@ enum cali_globals_flags {
 	CALI_GLOBALS_RESERVED1			= 0x00000002,
 	CALI_GLOBALS_RESERVED2			= 0x00000004,
 	CALI_GLOBALS_RESERVED3			= 0x00000008,
-	CALI_GLOBALS_RPF_OPTION_ENABLED	= 0x00000010,
-	CALI_GLOBALS_RPF_OPTION_STRICT	= 0x00000020,
+	CALI_GLOBALS_RPF_OPTION_ENABLED		= 0x00000010,
+	CALI_GLOBALS_RPF_OPTION_STRICT		= 0x00000020,
+	CALI_GLOBALS_NO_DRS_CIDRS		= 0x00000040,
```
fasaxc, line 32:
```suggestion
	CALI_GLOBALS_NO_DSR_CIDRS		= 0x00000040,
```

felix/dataplane/linux/bpf_route_mgr.go
```diff
@@ -230,6 +246,11 @@ func (m *bpfRouteManager) calculateRoute(cidr ip.V4CIDR) *routes.Value {
 		flags |= routes.FlagHost
 	}
 
+	if m.dsrOptoutCIDRs.Contains(cidr) {
```
fasaxc, line 249: `For this feature I think you need to check if the route is inside any of the `dsrOptoutCIDRs` For (bad) example, if the user sets `DSROptout=0.0.0.0/0` then all traffic should be excluded, even if it's also a node IP, or something from an IP pool.  The logic will put 0.0.0.0/0 in the routes map but the IP pool routes will be excluded.

I think we've got a trie implementation in the ip package, should be trivial to use the trie instead of the set and then the contains check will just do the right thing.

fasaxc: Yep, `CIDRTrie.Covers` does the check

felix/bpf/ut/nat_test.go
```diff
@@ -1896,6 +1896,260 @@ func TestNATNodePortIngressDSR(t *testing.T) {
 	Expect(v.Flags()).To(Equal(conntrack3.FlagNATFwdDsr | conntrack3.FlagNATNPFwd))
 }
 
+func TestNATNodePortDSROptout(t *testing.T) {
```
fasaxc, line 1899: Wonder if you could refactor a bit to make the between-packet set-up common and maybe add more comments to really spell out what you're doing.  I think that'd aid readability, especially for folks who aren't familiar with these tests
```
// Pretend that we're on node 1, set up NAT maps with a node port.
maps := mustSetUpMaps(WithNodePort(...)) 

// Simulate a packet arriving at node 1.  Expect the BFP program to encap the packet and redirect to node 2.
...

// Now pretend that we're on node 2...  Configure the NAT maps again.
...
// Feed it the packet from node 1, expect it to spot that it's a no-DSR source
// and record that in conntrack before forwarding to a local pod.

// Now pretend we've got the response packet form the pod at egress from the pod...

```

fasaxc (changes requested):
Nice; just one potential issue

felix/dataplane/linux/bpf_route_mgr.go
tomastigera: OK, excellent, I was pondering that and then decided that it is meant for external cidrs only. This will make it robust :+1:

api/pkg/apis/projectcalico/v3/felixconfig.go
tomastigera: yup me also surprised :man_shrugging:

marvin-tigera:

fasaxc (approved):
LGTM!

PR #7215, opened 2023-01-19
DRAFT WIP BPF 2079 Topology Aware Hints task

StevenTigera:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/proxy/proxy.go
+++ b/felix/bpf/proxy/proxy.go
@@ -337,3 +337,20 @@ type loggerRecorder struct{}
 
 func (r *loggerRecorder) Eventf(regarding runtime.Object, related runtime.Object, eventtype, reason, action, note string, args ...interface{}) {
 }
+
+func InvokeCategorizeEndpoints(svcMap k8sp.ServiceMap, epsMap k8sp.EndpointsMap, nodeLabels map[string]string) (clusterEndpoints, localEndpoints, allReachableEndpoints []k8sp.Endpoint, hasAnyEndpoints bool) {
+
+	// TODO - correct this because if the map has more than one entry then only the last entry would be used!
+	var svcInfo k8sp.ServicePort
+	for k := range svcMap {
+		svcInfo = svcMap[k]
+	}
+
+	var endpoints []k8sp.Endpoint
+	for k := range epsMap {
+		endpoints = epsMap[k]
+	}
+
+	clusterEndpoints, localEndpoints, allReachableEndpoints, hasAnyEndpoints = k8sp.CategorizeEndpoints(endpoints, svcInfo, nodeLabels)
+	return clusterEndpoints, localEndpoints, allReachableEndpoints, hasAnyEndpoints
+}
--- a/felix/bpf/proxy/proxy_test.go
+++ b/felix/bpf/proxy/proxy_test.go
@@ -16,6 +16,7 @@ package proxy_test
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"net"
 	"runtime"
 	"time"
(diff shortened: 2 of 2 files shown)


felix/bpf/proxy/proxy_test.go
```diff
@@ -536,6 +537,133 @@ var _ = Describe("BPF Proxy", func() {
...
+					Expect(err).NotTo(HaveOccurred())
+				})
+			})
+
+			AfterEach(func() {
+				By("stopping the proxy", func() {
+					close(syncStop)
+					p.Stop()
+				})
+			})
```
tomastigera, line 645: isn't this already done?

felix/bpf/proxy/proxy.go
```diff
@@ -337,3 +337,20 @@ type loggerRecorder struct{}
...
+	// TODO - correct this because if the map has more than one entry then only the last entry would be used!
+	var svcInfo k8sp.ServicePort
+	for k := range svcMap {
+		svcInfo = svcMap[k]
+	}
+
+	var endpoints []k8sp.Endpoint
+	for k := range epsMap {
+		endpoints = epsMap[k]
+	}
```
tomastigera, line 352: We do need to do this rewrite ourselves? Does not k8s provide it? It might be that k8s proxy switched to these structures and it would be better for us to change our code as well :man_shrugging:

felix/bpf/proxy/proxy_test.go
```diff
@@ -536,6 +537,133 @@ var _ = Describe("BPF Proxy", func() {
...
+			It("should invoke k8s CategorizeEndpoints", func() {
+
+				By("successfully syncing the service and endpoint maps", func() {
+					dp.checkState(func(s proxy.DPSyncerState) {
+						Expect(len(s.SvcMap)).To(Equal(1))
+						Expect(len(s.EpsMap)).To(Equal(1))
+
+						// Test Topology Aware Hints function.
+						nodeLabels := map[string]string{v1.LabelTopologyZone: "us-west-2a"}
+						clusterEndpoints, localEndpoints, allReachableEndpoints, hasAnyEndpoints := proxy.InvokeCategorizeEndpoints(s.SvcMap, s.EpsMap, nodeLabels)
```
tomastigera, line 656: this function should not be called here but somewhere in proxy.go This dp.checkState is only supposed to check that the state we are passing to syncer it what we expect. So maybe that rewrite above is not necessary is you call it from the proxy.go code.

felix/bpf/proxy/proxy_test.go
StevenTigera: I was following the pattern - the other two Describe blocks have k8s | BeforeEach | AfterEach - thanks

felix/bpf/proxy/proxy.go
StevenTigera: At a higher level the K8s code does work with the svcMap and epsMap objects but the [proxier.go](https://github.com/kubernetes/kubernetes/blob/b1d98bd7b34da70bd37187aa88b662e0d047d0ff/pkg/proxy/iptables/proxier.go#L997) code iterates thru the map key/val pairs invokes the `CategorizeEndpoints` passing in each svc and eps value for the corresponding key but I was trying to prototype invoking `CategorizeEndpoints` once in isolation to be unit tested

felix/bpf/proxy/proxy_test.go
StevenTigera: The `InvokeCategorizeEndpoints` above is just a helper function - the real call to K8s `CategorizeEndpoints` function is invoked in proxy.go  
PS: ideally I would have liked to have created the svcMap and epsMap objects directly and invoke in isolation i.e. unit tested but I see that the hintsAnnotation, which must be set on BaseServiceInfo object in order for Topology Aware Hints to work, is private.  I did see that if I create a service and endpoint the syncer would "generate" the corresponding svcMap and epsMap resources that could be used to test Topology Aware Hints but the current code here does feel wrong

StevenTigera:
As discussed with @tomastigera closing this draft PR - there will be replacement e.g. PR [7241](https://github.com/projectcalico/calico/pull/7241) - thanks

PR #7237, opened 2023-01-26
Change Mkdir to MkdirAll

MiladYarmohammadi:
Change Mkdir to MkdirAll to prevent existence error

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

### DIFF
--- a/calicoctl/calicoctl/commands/node/diags.go
+++ b/calicoctl/calicoctl/commands/node/diags.go
@@ -114,7 +114,7 @@ func runDiags(logDir string) error {
 		return fmt.Errorf("Error changing directory to temp directory to dump logs: %v", err)
 	}
 
-	err = os.Mkdir("diagnostics", os.ModeDir)
+	err = os.MkdirAll("diagnostics", os.ModeDir)
 	if err != nil {
 		return fmt.Errorf("Error creating diagnostics directory: %v\n", err)
 	}
@@ -179,7 +179,7 @@ func runDiags(logDir string) error {
 
 // getNodeContainerLogs will attempt to grab logs for any "calico" named containers for hosted installs.
 func getNodeContainerLogs(logDir string) {
-	err := os.Mkdir(logDir, os.ModeDir)
+	err := os.MkdirAll(logDir, os.ModeDir)
 	if err != nil {
 		fmt.Printf("Error creating log directory: %v\n", err)
 		return


caseydavenport:
@MiladYarmohammadi were you seeing an error produced somehow? 

This call to `Mkdir` is just for a local directory relative to the current path without any depth, so I don't think MkdirAll will change the behavior at all, but maybe I'm misunderstanding.

MiladYarmohammadi:
> @MiladYarmohammadi were you seeing an error produced somehow?
> 
> This call to `Mkdir` is just for a local directory relative to the current path without any depth, so I don't think MkdirAll will change the behavior at all, but maybe I'm misunderstanding.

By repeating this command, you will get the following error which may confuse the reader.
`Error:
Copying Calico logs
Error creating log directory: mkdir /tmp/calico1740305604/diagnostics/logs: file exists`
So I decided to propose this change.

caseydavenport:
I think that's expected though - we don't want to put multiple diags bundles in the same place, IIUC. 

Maybe we need to have a better error message? i.e., "Diags bundle already found - please remove it or rename it before continuing"?

MiladYarmohammadi:
> I think that's expected though - we don't want to put multiple diags bundles in the same place, IIUC.
> 
> Maybe we need to have a better error message? i.e., "Diags bundle already found - please remove it or rename it before continuing"?

No, that error belongs to the existence of `/tmp/calico2759398904/diagnostics/logs` directory, not to the existence of the log file. So we need to change the `Mkdir` to `MkdirAll`, just like using 'mkdir -p` instead of `mkdir`, to prevent this error.

calicoctl/calicoctl/commands/node/diags.go
```diff
@@ -114,7 +114,7 @@ func runDiags(logDir string) error {
 		return fmt.Errorf("Error changing directory to temp directory to dump logs: %v", err)
 	}
 
-	err = os.Mkdir("diagnostics", os.ModeDir)
+	err = os.MkdirAll("diagnostics", os.ModeDir)
```
mgleung, line 117: I think that this works, but it is a little awkward since we seem to be only creating one level of directory at a time with `MkdirAll`. Maybe we should just call `MkdirAll("diagnostics/logs", os.ModeDir)` here instead so we don't have to call `MkdirAll` twice?

MiladYarmohammadi: Our goal is to prevent throwing that exception, not to create all the directories at once, and it's easier to read the codes, so I think it's better to use it twice, but it's up to you.

caseydavenport:
@mgleung are we good to merge this?

mgleung (approved):
Sorry for the delay. I think I'm fine with leaving it as is @MiladYarmohammadi . Thanks for the contribution!

PR #7248, opened 2023-01-30
update pod2daemon nodeagent's DaemonSet apiVersion.

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

/kind cleanup

The pod2daemon/README.md says
```
The code here was tested with Kubernetes version v1.8.
```
But i found the pod2daemon/flexvol's Deployment is the 'apps/v1'.
So we can update the pod2daemon/nodeagent's DaemonSet to the match version 'apps/v1'.

Is that ok?


mgleung:
@yanggangtony thanks for updating this. The nodeagent hasn't been updated in a while which is why the manifests are using the older API version for the Daemonset. It should be fine to update.

yanggangtony:
> @yanggangtony thanks for updating this. The nodeagent hasn't been updated in a while which is why the manifests are using the older API version for the Daemonset. It should be fine to update. We have tests that have been running this on newer versions of K8s (looks like 1.22 at the time or writing). K8s v1.8 is just from when this was originally created.

thanks for review..
So should i can update the doc , replaced the v1.8 with 1.22?
is that ok?

mgleung:
@yanggangtony , sorry for the delay. I think it should be fine to update the K8s version to 1.22 though I don't particulary think it matters too much since we are running tests on all versions of K8s.

yanggangtony:
@mgleung 
thanks for reply.
Now the pr is k8s 1.22.😀😀

mgleung (approved):

PR #7259, opened 2023-01-31
fix words for more accurate.

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

/kind cleanup

fix words for more accurate.


caseydavenport:
Same here - docs are now at github.com/tigera/docs, so we'll need to re-open this there.

CC @ctauchen

PR #7263, opened 2023-02-01
Avoid logging WARNINGs about readiness during initial startup

nelljerram:
With Skimble we sometimes see WARNINGs about non-readiness during the first 10 seconds or so of
operation.  This is just because Skimble has not yet reached its inital ready state.

Those WARNINGs are benign, so we'd like to suppress them for Skimble - and more generally it makes
sense that what really matters is regression from a previously ready state.  So tweak the
coding so as only to Warn after previously being ready.

Also, it was confusing how part of the status gathering was inside the HasXYZProblem() functions, and part in the code calling those; I've refactored a little to (I hope) make that easier to understand.  And these methods didn't need to be public.


libcalico-go/lib/health/health.go
```diff
@@ -249,29 +255,23 @@ func (aggregator *HealthAggregator) Summary() *HealthReport {
...
-			summary.Live = false
-			if reporter.TimedOut() {
-				livenessStr = "timed out"
+		live, livenessStr := reporter.liveness()
+		if !live {
+			if aggregator.previouslyLive {
+				log.WithField("name", reporter.name).Warnf("Reporter is not live: %v.", livenessStr)
 			} else {
-				livenessStr = "reporting non-live"
+				log.WithField("name", reporter.name).Infof("Reporter is not live: %v.", livenessStr)
```
fasaxc, line 263: For liveness, would it make more sense to assume that a component is live at start of day (until first timeout)?  Letting liveness time out is really the only sensible way to go non-live (in any other case, why set non-live instead of panicing or doign a graceful shutdown?)

libcalico-go/lib/health/health.go
```diff
@@ -134,6 +138,10 @@ type HealthAggregator struct {
 
 	// HTTP server.  Non-nil when there should be a server running.
 	httpServer *http.Server
+
+	// Track whether we have previously reported live and ready overall.
```
fasaxc, line 142: `s/previously/ever/` ? (Often think of previous as "the immediately previous".)

fasaxc (approved):
Couple of minor thoughts but I think it'd be an improvement as-is.  I noticed Typha raising the same scary warnings at startup and was just thinking it'd be good to squash those at some point :-)

libcalico-go/lib/health/health.go
nelljerram: We do have that start of day assumption for liveness here: https://github.com/projectcalico/calico/blob/master/libcalico-go/lib/health/health.go#L150

I suppose that means we could remove the `previouslyLive` field, because it should always be true.  I'll do that.

You might also be suggesting: why would we support a reporter later being able to say `Live: false` ?  I don't think there are any reporters (in our repos) that can ever do that, so I agree that we could rip that out too.

libcalico-go/lib/health/health.go
nelljerram: Good idea, will do.

marvin-tigera:

PR #7287, opened 2023-02-06
fix:extension/v1beta1 to apps/v1 for DaemonSet

yulng:
Signed-off-by: yulng <wei.yang@daocloud.io>
Fix extension/v1beta1 to apps/v1 for DaemonSet
According to the following document:
https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/
image


yulng:
@caseydavenport 
:)

PR #7289, opened 2023-02-06
Update windows scripts for new docs

MichalFupso:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


node/windows-packaging/install-calico-windows.ps1
```diff
@@ -21,8 +21,9 @@
 Param(
     # Note: This URL only works for releases. To test development code, build calico-windows.zip from source.
     # VERSION is replaced by our build tooling.
-    [parameter(Mandatory = $false)] $ReleaseBaseURL="https://github.com/projectcalico/calico/releases/download/VERSION/",
-    [parameter(Mandatory = $false)] $ReleaseFile="calico-windows-VERSION.zip",
+    [parameter(Mandatory = $false)] $CalicoVersion="v3.25.0",
```
mgleung, line 24: What's the rationale we're using for replacing `VERSION` with `$CalicoVersion`? Is it just that we can support hardcoding `$CalicoVersion` separately?

I think if we decide to go this route, we could do something like https://github.com/projectcalico/calico/blob/master/manifests/generate.sh#L119-L128 which replaces the versions for our "static" manifests. This way the release process will pick up these changes automatically. This could be used to either replace `VERSION` or `CalicoVersion` from my understanding.

coutinhop (changes requested):
Looking good, thanks @MichalFupso 

It looks like we need to update the release builder for the new location of the install-calico-windows.ps1 script:
https://github.com/projectcalico/calico/blob/29ea74d9227d98886b98712746b824fa95e54b20/hack/release/pkg/builder/builder.go#L329

And also make it copy Install-Containerd.ps1 and PrepareNode.ps1 to this new location.

hack/postrelease/test_windows_artifacts.py
```diff
@@ -11,7 +11,7 @@ def test_calico_release_has_windows_zip():
     assert req.status_code == 302
 
 def test_calico_windows_script_uses_expected_install_zip():
-    resp = requests.get('https://projectcalico.docs.tigera.io/archive/%s/scripts/install-calico-windows.ps1' % RELEASE_STREAM)
+    resp = requests.get('https://github.com/projectcalico/calico/releases/download/%s/install-calico-windows.ps1' % RELEASE_STREAM)
```
coutinhop, line 14: I think we need to use `RELEASE_VERSION` (vX.Y.Z) here, instead of `RELEASE_STREAM` (just vX.Y)

coutinhop (approved):
LGTM

PR #7320, opened 2023-02-09
Fix manifests URL used by Windows FV; also pull in Windows FV testing code from tigera/process

nelljerram:
#### Pull Windows FV coding into this repo

Windows FV coding has lived in the tigera/process repo - I think dating from when tigera/process was effectively acting as our monorepo substitute.  Now that we have a real monorepo, we can pull that code in here.

#### https://docs.projectcalico.org/manifests no longer exist, use hash release instead

Windows FV was using https://docs.projectcalico.org/manifests/tigera-operator.yaml - meaning the latest Calico release - to install Calico.  There is no equivalent URL available now.  Let's try using the latest master hash release instead.


song-jiang (approved):
LGTM. Thanks @nelljerram . The original code in process repo needs to be tidy up a bit but that's something I can work on later.

PR #7291, opened 2023-02-07
Fix UID for v3 resources in KDD mode

fasaxc:
## Description

Picking up the fix for #5715.  Looks like we adopted this fix in Calico Enterprise already so let's be consistent.

## Related issues/PRs

fixes #5715

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that Calico API server would reuse UUIDs from the underlying CRD objects that underpin the datamodel (thus confusing Kubernetes ownership tracking and ArgoCD).  This will result in the apparent UUIDs of calico "v3" resources changing over upgrade.  This was unavoidable in order to split them from the underlying CRD UUIDs.
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

Issue #5715, opened 2022-03-07
Calico Apiserver not delete Resource with OwnerReference 


dtrouillet:
I have an operator which create a networkpolicies.projectcalico.org/v3 with a OwnerRefernce and when I delete my Custom Resource, the networkpolicies.projectcalico.org/v3 is not deleted.

## Expected Behavior
I expect that the deletion of my Custom Resources delete at the same time the networkpolicies.projectcalico.org/v3 (with the OwnerReference)

## Current Behavior
When I delete my Custom Resources, the networkpolicies.projectcalico.org/v3 is not deleted.

## Steps to Reproduce (for bugs)
1. Create a resource 
2. Create networkpolicies.projectcalico.org/v3 with OwnerReference of last created resource
3. Delete with kubectl the step 1' resource

## Context
The Calico Api Server is deployed inside the cluster

My Custom Resource :
```
apiVersion: mycrd.fr/v1alpha1
kind: FluxNP
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"mycrd.fr/v1alpha1","kind":"FluxNP","metadata":{"annotations":{},"name":"ext","namespace":"test"},"spec":{"allow":true,"interval":3600}}
  creationTimestamp: "2022-03-07T16:03:18Z"
  generation: 1
  name: ext
  namespace: test
  resourceVersion: "16336"
  uid: 0e5e3214-5bda-429e-8d53-2f1b77363286
spec:
  allow: true
  interval: 3600
status:
  lastUpdated: "2022-03-07T16:03:18Z"
  revision: 1
```

The networkpolicies.projectcalico.org/v3 : 
```
apiVersion: projectcalico.org/v3
kind: NetworkPolicy
metadata:
  creationTimestamp: "2022-03-07T16:03:18Z"
  name: ext
  namespace: test
  ownerReferences:
  - apiVersion: mycrd.fr/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: FluxNP
    name: ext
    uid: 0e5e3214-5bda-429e-8d53-2f1b77363286
  resourceVersion: "16335"
  uid: f14af896-94a3-453c-a9b1-4a29aaa9e20f
spec:
  egress:
  - action: Allow
    destination: {}
    source: {}
  types:
  - Egress
```

## Your Environment
* Calico version : 3.22.1
* Orchestrator version (e.g. kubernetes, mesos, rkt): rk2 1.21.5
[...]

PR #7356, opened 2023-02-16
Let UID be stored and retrieved by annotation instead of cloning UIDs.

geezyx:
## Description

When converting from calico to k8s objects and back, there is already logic storing the ObjectMeta in an annotation for the conversion. There should be no need to set the UUID of the converted object to match the source object, and it's a bad practice because kubernetes expectes UUIDs to be unique from one object to the next for operations like garbage collection.

If there are other parts of calico which rely on these UUIDs matching, I recommend treating those occurrences as bugs to be remediated.

## Related issues/PRs

fixes (https://github.com/projectcalico/calico/issues/5715)

Alternative approach to https://github.com/projectcalico/calico/pull/7291

My thought here is that kubernetes can manager UUIDs, there shouldn't be a need for the UUID of the calico object to be related to the UUID of the k8s object. The relationship between the two can be tracked via the storage of the k8s object in the calico metadata annotation.

@fasaxc stated there was a regression for the other PR, and I suspect this change might cause the same issues or similar issues. I'd love to help track down those regressions and get them fixed, I don't think calico should rely on matching UUIDs.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that Calico API server would reuse UUIDs from the underlying CRD objects that underpin the datamodel (thus confusing Kubernetes ownership tracking and ArgoCD).  This will result in the apparent UUIDs of calico "v3" resources changing over upgrade.  This was unavoidable in order to split them from the underlying CRD UUIDs.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/k8s/resources/resources.go
+++ b/libcalico-go/lib/backend/k8s/resources/resources.go
@@ -158,7 +158,6 @@ func ConvertCalicoResourceToK8sResource(resIn Resource) (Resource, error) {
 	meta.Namespace = rom.GetNamespace()
 	meta.ResourceVersion = rom.GetResourceVersion()
 	meta.Labels = rom.GetLabels()
-	meta.UID = rom.GetUID()
 
 	resOut := resIn.DeepCopyObject().(Resource)
 	romOut := resOut.GetObjectMeta()
@@ -201,7 +200,6 @@ func ConvertK8sResourceToCalicoResource(res Resource) error {
 	meta.ResourceVersion = rom.GetResourceVersion()
 	meta.Labels = rom.GetLabels()
 	meta.Annotations = annotations
-	meta.UID = rom.GetUID()
 
 	// If no creation timestamp was stored in the metadata annotation, use the one from the CR.
 	// The timestamp is normally set in the clientv3 code. However, for objects that bypass


fasaxc:
> I don't think calico should rely on matching UUIDs

Totally agree!  It was a mistake in our initial implementation.  However, I think there are some downsides to storing UID in an annotation that were mentioned on the other ticket.  Something about not being able to handle preconditions on a patch or deletion correctly (because the client might just say "delete the resource with UID _calico UID_" and we'd have no way to translate that into a request to the API server if the UIDs are unrelated.

geezyx:
> because the client might just say "delete the resource with UID _calico UID_" and we'd have no way to translate that into a request to the API server if the UIDs are unrelated.

Thanks for the reply! Just thinking through that example, would the annotation containing the related k8s object's metadata not be present in that scenario? I'm sure I'm missing something about the relationship between the calico objects and k8s objects, is there some examples in the code you could point me to that would help me grok that?

fasaxc:
> Oh! I think I understand what you are saying now. Would we be able to handle that case by getting the objectmeta for the thing being patched/deleted and extract the related k8s object UUID from the annotation in that scenario?

Yes, I think the Calico API server would need to do a read in order to find the real UID and to check the annotation UID matches.

caseydavenport:
@fasaxc @geezyx this one has been inactive for a while. Do we think it's something we can revive? Or do we need more design discussion?

fasaxc:
I think the approach of storing a second UUID in an annotation might be good but we need to review all the ways that you can interact with a UUID through our API server and make sure they all work.  The one that I'm aware of is that you can send a PATCH request with a pre-condition on the UIDUUID  You can probably do a GET by UUID too.  The problem is: how does the Calico API server handle these?  How do we get-by-UUID if the UUID of the CRD isn't in its UUID field (can we translate that to get-by-annotation) or do we need to list all or something else?

The other main approach that we were looking at was deriving the v3 UUID from the CRD one.  Then if you ask to patch or get by UUID we can derive the CRD UUID from the one presented in the request.  A little bit naughty to mess with UUIDs because they have some structure but, for example, we could encrypt the non-metadata bits of the UUID with a static key to get a random-looking UUID with low likelyhood of collision.

caseydavenport:
@fasaxc my understanding is that "Get-by-UID" is not something that is currently possible in the Kubernetes API, although there is this issue tracking both "get" and "delete" by UID so it theoretically might be added some day: https://github.com/kubernetes/kubernetes/issues/20572

However, the last comments of any substance are from 2016! I would normally trend towards being prepared for the unlikely situation that they implement "get by UID" anyway, but given this bug is pretty horrible and has been open for a while, I wonder if banking the annotation based storage and crossing that bridge when we get there is good enough? 

UID preconditions in Delete are definitely a real thing, though. We implement them explicitly in our API server code here: https://github.com/caseydavenport/calico/blob/6c471a9b26f1c0fc2a8e5de2fafeef90206d7782/apiserver/pkg/storage/calico/resource.go#L157-L168

IIUC, that logic would work as expected without any additional changes, since it's operating on the libcalico-go "frontend" representation of the objects (i.e., after the translation performed by the backend code being changed here). I'll give it a spin to see how it works to confirm, though.

caseydavenport:
One corner case I just encountered with this annotation based approach is that CRDs created with calicoctl or via CRDs directly don't have the `projectcalico.org/metadata` annotation on them, so reads of those resources via the API server will not have a UID.

caseydavenport:
Closing this in favor of https://github.com/projectcalico/calico/pull/8586/

PR #7358, opened 2023-02-16
Using the netlink.NewLinkAttrs to bring default values from kernel

radixo:
## Description

This PR solves a miscommunication between calico-cni and kernel, where a misconfiguration of the structure netlink.LinkAttrs was sending to kernel the information of the queue discipline for the veth pairs generated by calico-cni to be of value 0.
The problem was caused by creating the structure without the function netlink.NewLinkAttrs which initializes the TxQLen to -1, for not sending this info to kernel and getting the kernel default value instead.
For good practices I changed in everywhere the code.

## Todos

- [ ] Tests
- [ ] Release note

## Release Note

```release-note
Ensure that veths are created with the proper default values from the kernel.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
+++ b/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
@@ -76,14 +76,14 @@ func (d *linuxDataplane) DoNetworking(
 	}
 
 	err = ns.WithNetNSPath(args.Netns, func(hostNS ns.NetNS) error {
+		la := netlink.NewLinkAttrs()
+		la.Name = contVethName
+		la.MTU = d.mtu
+		la.NumTxQueues = d.queues
+		la.NumRxQueues = d.queues
 		veth := &netlink.Veth{
-			LinkAttrs: netlink.LinkAttrs{
-				Name:        contVethName,
-				MTU:         d.mtu,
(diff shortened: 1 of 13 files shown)


cni-plugin/pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -76,14 +76,14 @@ func (d *linuxDataplane) DoNetworking(
 	}
 
 	err = ns.WithNetNSPath(args.Netns, func(hostNS ns.NetNS) error {
+		la := netlink.NewLinkAttrs()
```
mgleung, line 79: I'm worried that using the constructor masks the issue and that we may not realize what values we are intending to set here for `TxQLen` and `NetNsID`. Maybe we should add a comment here stating that this constructor should also be setting those values so that we will use the kernel defaults?

tomastigera: Hmmm idk, the library specifically says to use the constructor or set all the values properly. We would have to comment every usecase. I wish there was no accesses to the individual items and then there would be no such an issue but to use the constructor :man_shrugging: I think this is a great catch! Thanks @radixo !

mgleung (approved):

PR #7368, opened 2023-02-17
Increase MTU tests timeout to remove flakes

coutinhop:
## Description

Increase timeouts on tests that verify MTU to 60s, due to the fact that Felix monitors host MTU every 30s. This should get rid of some flakiness in these tests.

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/mtu_test.go
+++ b/felix/fv/mtu_test.go
@@ -104,7 +104,7 @@ var _ = infrastructure.DatastoreDescribe("VXLAN topology before adding host IPs 
 						Eventually(func() string {
 							out, _ := felix.ExecOutput("cat", "/var/lib/calico/mtu")
 							return out
-						}, "30s", "100ms").Should(ContainSubstring(vxlanEnabledMtu))
+						}, "60s", "100ms").Should(ContainSubstring(vxlanEnabledMtu))
 					}
 
 					// Disable VXLAN. We should expect the MTU to change, but still based on the default 1460.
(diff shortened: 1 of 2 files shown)


felix/fv/mtu_test.go
```diff
@@ -104,7 +104,7 @@ var _ = infrastructure.DatastoreDescribe("VXLAN topology before adding host IPs
 						Eventually(func() string {
 							out, _ := felix.ExecOutput("cat", "/var/lib/calico/mtu")
 							return out
-						}, "30s", "100ms").Should(ContainSubstring(vxlanEnabledMtu))
+						}, "60s", "100ms").Should(ContainSubstring(vxlanEnabledMtu))
```
tomastigera, line 107: I wonder if it would make sense to only check the condition every say 500ms or 1s given the huge timeout

tomastigera (approved):
with a nit that does not need a re-review if addressed

felix/fv/mtu_test.go
coutinhop: that's probably a good idea, will do! thanks!

mgleung (approved):
LGTM, though we may want to keep note of these tests since increasing timeouts won't fix the issue if we end up running the tests on slower instances in the future.

PR #7383, opened 2023-02-22
Go1.19 upgrade

rene-dekker:
A number of things have been done in this PR.

Modify all VERSION_FLAGS to LDFLAGS
We want to be consistent in specifying go build -ldflags "$(LDFLAGS)" in our functions to build binaries. Therefore the usage of these FLAGS have been made consistent throughout the repo. Otherwise it leads to random or double double-quotes in the build instruction.

Move everything to go-build:v0.80 and build as much as we can with standardized build functions.
go-build:v0.80 is based on go1.19.5, while go-build:v0.78 was based on go1.18.9. Since go1.19 boringssl is added to the main tree of the project. These changes build all the binaries and checks that go-boring functions are present in the symbols using go tool nm.


hjiawei (approved):
lgtm

mgleung (approved):

Issue #7110, opened 2022-12-19
Calico in eBPF mode doesn't seem to honor terminationGracePeriodSeconds and blocks outbound traffic from pods being terminated


igcherkaev:
## Expected Behavior
Once pod is being deleted and marked as being Terminated by Kubernetes, calico should still allow traffic (only for already established connections)  from the pod back to the client and vice versa, until it's gone for good.

## Current Behavior
Traffic from pods being terminated is dropped. While tcpdump'ing it, I noticed TCP retransmits on the cali interface and the client didn't receive anything. Connection was kept open even when the pod had been long gone.

## Possible Solution

Ideally, memorize the value of terminationGracePeriodSeconds set for the pod, once endpoint is removed, calculate the moment in time when it's safe to clean up and clean up after that moment is reached. Until it's not reached, keep connection entry/nat entry and all related policies in place.

## Steps to Reproduce (for bugs)

1. Deploy a pod
2. Make a long-running request to it
3. Delete the pod
4. Even though the app handles SIGTERM properly and attempts to complete the request, generated response never gets to the client.

I can upload my calico-test docker image to reproduce this, if need be.

## Context

To achieve zero downtime deployments it is required to allow instances of the app complete in-flight requests while a new version is being rolled out and old pods are gracefully terminated. We've been receiving complaints from clients during deployments and HPA scale down events that some of their requests time out.

## Your Environment

* Calico version: `3.24.5`
* Orchestrator version (e.g. kubernetes, mesos, rkt): `Vanilla Kubernetes 1.22.14`
* Operating System and version: `Linux rndkbw503 5.15.63-flatcar #1 SMP Mon Aug 29 18:27:27 -00 2022 x86_64 Intel(R) Xeon(R) CPU E5-2660 v2 @ 2.20GHz GenuineIntel GNU/Linux`

caseydavenport:
@tomastigera want to look at this one? 

I wouldn't have thought we would need to do anything special to handle termination grace period properly, since we key off the deleted event and I wouldn't expect us to change our programming until we receive it. But perhaps BPF mode does something I'm not aware of?

Or maybe it's something in our eBPF kube-proxy implementation?

igcherkaev:
[...]

PR #7386, opened 2023-02-23
[BPF] Extend syncer to include endpoint slices at Terminating status

StevenTigera:
Extend syncer to include endpoint slices at Terminating status. The endpoint are excluded from the NAT backend table, but are not removed from conntrack as long as they are terminating to to honor the shut down grace period and to allow pod to shut down gracefully.

fixes https://github.com/projectcalico/calico/issues/7110

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: Connections to terminating endpoints of services not removed from conntrack in order to honor their grace shutdown period.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/proxy/syncer.go
+++ b/felix/bpf/proxy/syncer.go
@@ -552,7 +552,7 @@ func (s *Syncer) apply(state DPSyncerState) error {
 		eps := make([]k8sp.Endpoint, 0, len(state.EpsMap[sname]))
 		for _, ep := range state.EpsMap[sname] {
 			zoneHints := ep.GetZoneHints()
-			if ep.IsReady() {
+			if ep.IsReady() || ep.IsTerminating() {
 				if ShouldAppendTopologyAwareEndpoint(nodeZone, hintsAnnotation, zoneHints) {
 					eps = append(eps, ep)
 				} else {
@@ -707,8 +707,12 @@ func (s *Syncer) updateService(skey svcKey, sinfo k8sp.ServicePort, id uint32, e
 		if !ep.GetIsLocal() {
 			continue
 		}
-		if err := s.writeSvcBackend(id, uint32(cnt), ep); err != nil {
-			return 0, 0, err
+
+		// eps could contain Ready and Terminating pods but only write Ready pods to backend.
+		if ep.IsReady() {
+			if err := s.writeSvcBackend(id, uint32(cnt), ep); err != nil {
+				return 0, 0, err
+			}
 		}
 
 		cpEps = append(cpEps, ep)
@@ -720,8 +724,12 @@ func (s *Syncer) updateService(skey svcKey, sinfo k8sp.ServicePort, id uint32, e
 		if ep.GetIsLocal() {
 			continue
 		}
-		if err := s.writeSvcBackend(id, uint32(cnt), ep); err != nil {
-			return 0, 0, err
+
+		// eps could contain Ready and Terminating pods but only write Ready pods to backend.
+		if ep.IsReady() {
+			if err := s.writeSvcBackend(id, uint32(cnt), ep); err != nil {
+				return 0, 0, err
+			}
 		}
 
 		cpEps = append(cpEps, ep)
--- a/felix/bpf/proxy/syncer_test.go
+++ b/felix/bpf/proxy/syncer_test.go
@@ -127,6 +127,13 @@ var _ = Describe("BPF Syncer", func() {
 			},
 		}
 
+		svcKey4 := k8sp.ServicePortName{
+			NamespacedName: types.NamespacedName{
+				Namespace: "default",
+				Name:      "fourth-service",
+			},
+		}
+
 		By("inserting another service with multiple endpoints", makestep(func() {
 			state.SvcMap[svcKey2] = proxy.NewK8sServicePort(
 				net.IPv4(10, 0, 0, 2),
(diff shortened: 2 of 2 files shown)


felix/bpf/proxy/syncer.go
```diff
@@ -552,7 +552,7 @@ func (s *Syncer) apply(state DPSyncerState) error {
 		eps := make([]k8sp.Endpoint, 0, len(state.EpsMap[sname]))
 		for _, ep := range state.EpsMap[sname] {
 			zoneHints := ep.GetZoneHints()
-			if ep.IsReady() {
+			if ep.IsReady() || ep.IsTerminating() {
```
fasaxc, line 555: Does applySvc filter out Terminating endpoints later?  Terminating endpoints shouldn't be in the BPF maps.

emptywee: Also, placing `IsTerminating()` next to `IsReady()` implies that if endpoint became not ready, in-flight requests would also be dropped? Or I am just being crazy here?

StevenTigera: Thanks @fasaxc - [applySvc](https://github.com/projectcalico/calico/blob/master/felix/bpf/proxy/syncer.go#L349)() does not seem to filter out Terminating endpoints.  My understanding of the current syncer code in master is that all code paths seem to be channelled thru the [apply](https://github.com/projectcalico/calico/blob/master/felix/bpf/proxy/syncer.go#L526)() function which initially filters out only endpoints at status [Ready](https://github.com/projectcalico/calico/blob/master/felix/bpf/proxy/syncer.go#L555) and adds them to the eps collection for further processing.  

I have tried to grok other parts of the code and it seems that if an endpoint is at status other than Ready during the apply() function then it would not be included here thus not have an opportunity to be processed later on in the code - hence the modification in the PR but pls correct me if I am wrong about this - thx.

tomastigera: I think, checking the terminating flag is the simplest (nice!) way to tackle this problem. However,this will need to be a little bit more complicated.

(1) yes, we need to let the terminating through here so that we know later on that they are terminating.
(2) we must remove the terminating from the bpf maps because we should not allow new connections to that endpoint via the service as `At this point, the pod stops getting new traffic. Containers running in the pod will not be affected.`
(3) we need to remember that the pod is terminating to answer `true` when [this callback](https://github.com/projectcalico/calico/blob/master/felix/bpf/proxy/syncer.go#L1153) is called.

Perhaps we need to keep the state in [this struct](https://github.com/projectcalico/calico/blob/master/felix/bpf/proxy/syncer.go#L43) so that when we (a) are writing out the bpf backend map we skip the terminating and (b) when answering the question whether that backend exists, we know about it.

StevenTigera: Update code to prevent terminating pods from being added to the bpf maps as outlined above @tomastigera

felix/bpf/proxy/syncer.go
```diff
@@ -707,8 +707,12 @@ func (s *Syncer) updateService(skey svcKey, sinfo k8sp.ServicePort, id uint32, e
 		if !ep.GetIsLocal() {
 			continue
 		}
-		if err := s.writeSvcBackend(id, uint32(cnt), ep); err != nil {
```
StevenTigera, line 710: Update code to prevent terminating pods from being added to the bpf maps as outlined above @tomastigera

felix/bpf/proxy/syncer.go
```diff
@@ -707,8 +707,12 @@ func (s *Syncer) updateService(skey svcKey, sinfo k8sp.ServicePort, id uint32, e
...
-			return 0, 0, err
+
+		// eps could contain Ready and Terminating pods but only write Ready pods to backend.
+		if ep.IsReady() {
```
StevenTigera, line 712: Update code to prevent terminating pods from being added to the bpf maps as outlined above @tomastigera

felix/bpf/proxy/syncer.go
```diff
@@ -720,8 +724,12 @@ func (s *Syncer) updateService(skey svcKey, sinfo k8sp.ServicePort, id uint32, e
 		if ep.GetIsLocal() {
 			continue
 		}
-		if err := s.writeSvcBackend(id, uint32(cnt), ep); err != nil {
-			return 0, 0, err
+
+		// eps could contain Ready and Terminating pods but only write Ready pods to backend.
+		if ep.IsReady() {
```
StevenTigera, line 729: Update code to prevent terminating pods from being added to the bpf maps as outlined above @tomastigera

felix/bpf/proxy/syncer_test.go
```diff
@@ -1096,6 +1096,44 @@ var _ = Describe("BPF Syncer", func() {
 			Expect(eps.m).To(HaveLen(2))
```
StevenTigera, line 1096: Update unit test as logic has changed above

felix/bpf/proxy/syncer_test.go
```diff
@@ -1096,6 +1096,44 @@ var _ = Describe("BPF Syncer", func() {
...
 		}))
 
+		By("checking endpointslice terminating status should be included in endpointslice collection for processing", makestep(func() {
```
StevenTigera, line 1099: Additional unit test as discussed with @tomastigera

felix/bpf/proxy/syncer_test.go
```diff
@@ -1096,6 +1096,44 @@ var _ = Describe("BPF Syncer", func() {
...
+			}
+
+			err := s.Apply(state)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(eps.m).To(HaveLen(2))
+		}))
+
+		By("after checking endpointslice terminating status ensure conntrack scan processes correct number of maps", makestep(func() {
+			svc := state.SvcMap[svcKey]
+			ep := state.EpsMap[svcKey][0]
```
tomastigera, line 1118: We need to make sure that the selected ep is the terminating one. Imo there is no guarantee it is the [0] one.

tomastigera (changes requested):
looks good, simpler than I thought :raised_hands:

felix/bpf/proxy/syncer_test.go
StevenTigera: Thanks @tomastigera - updated unit test to better represent real world scenario and modified accordingly

felix/bpf/proxy/syncer_test.go
```diff
@@ -1096,6 +1096,44 @@ var _ = Describe("BPF Syncer", func() {
...
+				&k8sp.BaseEndpointInfo{Ready: true, Endpoint: "10.2.0.2:5555"},
+				&k8sp.BaseEndpointInfo{Terminating: true, Endpoint: "10.1.0.3:5555"},
+			}
+
+			err := s.Apply(state)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(eps.m).To(HaveLen(2))
+		}))
+
+		By("after checking endpointslice terminating status ensure conntrack scan processes correct number of maps", makestep(func() {
```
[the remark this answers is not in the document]
StevenTigera: Thanks @tomastigera - updated accordingly

felix/bpf/proxy/syncer.go
StevenTigera: Implemented changes as per thread

felix/bpf/proxy/syncer_test.go
```diff
@@ -1096,6 +1096,46 @@ var _ = Describe("BPF Syncer", func() {
...
+			err := s.Apply(state)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(eps.m).To(HaveLen(1))
+		}))
+
+		By("checking that conntrack scan does not remove the terminating endpoint connection", makestep(func() {
+			svc := state.SvcMap[svcKey]
+			eps := state.EpsMap[svcKey]
+			for _, ep := range eps {
+				if ep.IsReady() {
```
tomastigera, line 1119: We should either include the Terminating only or Ready and Terminating. Perhaps the latter is better to make sure that the change does not effect the Ready ones. But the **the key is to include the Teminating**, because that is what we want to make sure that does not get removed from the ct table!

So you could either find the terminating and create the entry for that one. Or create entry for all 3 and then make sure that you also find 3 after the clean up.

If you wrote the test before the rest of the changes, it would fail and now it should succeed after the change to the syncer code, which proves the change.

felix/bpf/proxy/syncer_test.go
```diff
@@ -1096,6 +1096,46 @@ var _ = Describe("BPF Syncer", func() {
...
+				cnt++
+				key := conntrack.KeyFromBytes(k)
+				val := conntrack.ValueFromBytes(v)
+				log("key = %s\n", key)
+				log("val = %s\n", val)
+				return maps.IterNone
+			})
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(cnt).To(Equal(2))
```
tomastigera, line 1136: The test passes because this was true already before the change!

felix/bpf/proxy/syncer_test.go
StevenTigera: Thanks @tomastigera - updated unit test but thought best actually to "start fresh" i.e. clean up all prior state and then follow suggestion i.e. create entry for all 3x and also followed the convention [here](https://github.com/projectcalico/calico/blob/master/felix/bpf/proxy/syncer_test.go#L174) i.e. manually unroll loop to add multiple conntrack entries - thanks

felix/bpf/proxy/syncer_test.go
StevenTigera: Updated as per previous entry

tomastigera (approved):
LGTM, thanks!

PR #7395, opened 2023-02-25
Add imageRegistry and imagePath support in Helm Chart

dthomson-sysdig:
## Description

Add in support for third-party (private) image registry and image path in the helm chart.

Testing has been done manually on a private k8s cluster with a private ECR.

Docs probably want to be updated for Helm installation so that users are aware of the feature flags.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/charts/tigera-operator/templates/crs/custom-resources.yaml
+++ b/charts/tigera-operator/templates/crs/custom-resources.yaml
@@ -16,6 +16,13 @@ spec:
 
 {{ end }}
 
+{{- if .Values.imageRegistry }}
+  registry: {{.Values.imageRegistry}}
+{{- end }}
+{{- if .Values.imagePath }}
+  imagePath: {{.Values.imagePath}}
+{{- end }}
+
 {{ if .Values.apiServer.enabled }}
 {{ $apiServerSpec := omit .Values.apiServer "enabled" }}
 ---
--- a/charts/tigera-operator/README.md
+++ b/charts/tigera-operator/README.md
@@ -92,6 +92,12 @@ The default values.yaml should be suitable for most basic deployments.
 # Populates the `imagePullSecrets` property for all Pods controlled by the `Installation` resource.
 imagePullSecrets: {}
 
+# Install from a third-party (private) container repository
+imageRegistry: ""
+
+# Directory in the third-party (private) registry that contains images required
+imagePath: ""
+
 # Configures general installation parameters for Calico. Schema is based
 # on the operator.tigera.io/Installation API documented
 # here: https://projectcalico.docs.tigera.io/reference/installation/api#operator.tigera.io/v1.InstallationSpec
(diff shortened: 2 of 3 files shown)


caseydavenport:
https://github.com/projectcalico/calico/pull/7395/files#diff-1ef25acab082b9c39db773ccde81f063b32519075a43897fddc882256bda701dR15

The helm chart values already allow you to put an arbitrary json spec in for the installation in order to configure these two fields, so I don't think we should have explicit call outs for them.

The design of this chart is that the `values.installation` field is free-form, and is directly passed as the spec to the Installation that's created on the cluster, so that we don't need to constantly add new fields to the chart as the API evolves.

In other words, the following yaml should already work:

```yaml
installation:
  imagePath: ""
  registry: ""
```

charts/tigera-operator/templates/crs/custom-resources.yaml
```diff
@@ -16,6 +16,13 @@ spec:
 
 {{ end }}
 
+{{- if .Values.imageRegistry }}
```
caseydavenport, line 19: Per my other comment, I don't think we should add these fields. But as an aside, if we were to add them they would need to be above the `{{ end }}` clause on line 17. Otherwise, setting `values.installation.enabled` to false will result in the generation of an invalid custom resources file.

dthomson-sysdig:
> https://github.com/projectcalico/calico/pull/7395/files#diff-1ef25acab082b9c39db773ccde81f063b32519075a43897fddc882256bda701dR15
> 
> The helm chart values already allow you to put an arbitrary json spec in for the installation in order to configure these two fields, so I don't think we should have explicit call outs for them.
> 
> The design of this chart is that the `values.installation` field is free-form, and is directly passed as the spec to the Installation that's created on the cluster, so that we don't need to constantly add new fields to the chart as the API evolves.
> 
> In other words, the following yaml should already work:
> 
> ```yaml
> installation:
>   imagePath: ""
>   registry: ""
> ```

ah sure, that's fair, @caseydavenport

charts/tigera-operator/templates/crs/custom-resources.yaml
dthomson-sysdig: Right, makes sense.

caseydavenport:
@dthomson-sysdig I'm going to close this one for now, but appreciate you putting in the time to submit this! Please let us know if there's some reason what I suggested above is insufficient and we can discuss more.

lwr20:
If the above wasn't obvious, perhaps we need a PR in the docs to make it more obvious for the next person?

dthomson-sysdig:
> If the above wasn't obvious, perhaps we need a PR in the docs to make it more obvious for the next person?

Hi @lwr20 , yes, I didn't catch that `installation` is free-form.

A mention for private registry/path on [https://docs.tigera.io/calico/3.25/getting-started/kubernetes/helm](https://docs.tigera.io/calico/3.25/getting-started/kubernetes/helm) would go a long way I reckon.

lwr20:
@ctauchen FYI

PR #7408, opened 2023-03-01
Removes the remaining docs files

ctauchen:
Some remnants from the general docs clear-out.


ctauchen:
Warning: I'm removing calico/scripts/PrepareNode.ps1. I don't think this is being used, but perhaps someone who knows better ought to confirm.

Currently we're hosting a copy of this file on the new docs site: https://docs.tigera.io/calico/next/scripts/preparenode.ps1.

ctauchen:
@caseydavenport This PR removes those files that somehow got back in.

caseydavenport (approved):

Issue #5250, opened 2020-12-07
Felix clears XDP programs it doesn't own


sjh5205:
I'm using XDP for modify packet payload directly on datapath.
But also Calico use XDP for failsafe ports and CIDR map.

When each Calico's XDP resync period, my XDP program has been wiped out, our system is not working properly.

## Context
Felix wipe out attached XDP program on each XDP resync period.
Can I shutdown XDP functionalities in Calico?

## Your Environment
I'm using Kubernetes from Kubespray.
Calico version is v3.15.2

## Problematic Code

Here is code sinppet from Felix.
https://github.com/projectcalico/felix/blob/bbbd53863935730011636e6fa9652def9dc48586/dataplane/linux/int_dataplane.go#L449-L476

xdpState is initialize, XDP program has been wiped out.

https://github.com/projectcalico/felix/blob/bbbd53863935730011636e6fa9652def9dc48586/dataplane/linux/int_dataplane.go#L1569-L1595

Again, on resync.

alexanderstephan:
Any updates on this?
As it seems like the previous PR attempt by @sjh5205 has become stale, I would love to take this over and create a new PR directly in Calico (the Felix fork that was used is deprecated as Felix was moved to Calico itself).

PR #7423, opened 2023-03-03
Verify XDP program name before removal

alexanderstephan:
## Description
This PR verifies that the name of the BPF program is `prefilter` which is of the XDP program that Felix uses
when running, the `RemoveXDP()` function that is run during XDP re-sync.

I also implemented a unit test to verify the fix.
The idea of the unit test is to create an XDP BPF file with a different program name and load it to a network interface.
My first idea was to use the existing `prefilter` program and modify the name by altering the function name in the symbol table (the name is derived from the function). By altering the name, we can simulate a program that does not match during the name verification and therefore should not get deleted.

This can be achieved with `objcopy`. The version of `objcopy` in the Debian 10 docker image that is used for tests does not yet support the `elf64-bpfle` file format. A solution would be to upgrade to Debian 11.
However, to keep it simple, I added the modified BPF program in a `testdata` package for now, as it seems like the most convenient solution.

## Disclamer

Currently, this only works for the prefilter program.  It is not clear to me whether this needs to be further generified.

## Related issues/PRs
https://github.com/projectcalico/calico/issues/5250

Based on existing commits from user @sjh5205. Thanks! I added you as co-author.

https://github.com/sjh5205/felix/commit/6f4a643e988cb61ebeb4bf80b749f737619a04a1
https://github.com/sjh5205/felix/commit/b6bc626a48f3d27726f8fff1dd69e4f572c5a007

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


tomastigera:
Hey @alexanderstephan thank for the contribution. This part of the code is going though some significant refactoring atm (my wip  https://github.com/projectcalico/calico/pull/7219) and this issue will be also addressed, so I am a little hesitant to merge this now just to create a conflict and delete this code soon after. Lets keep this open and revisit that when the refactoring merges, wdyt?

alexanderstephan:
Hey, @tomastigera, thanks for the update.

> Lets keep this open and revisit that when the refactoring merges, wdyt?

Sounds like a great idea!

Especially if your PR it almost finished, it doesn't really make sense.
I assume your PR is also included in the next release, right? This is the only case where merging this could make sense.
(This fix is very urgent for me and also others, I think.)

Other than that, let's keep this open for now. Maybe we can still make use of the unit test.

Besides, could you also link the respective issue in your PR for better visibility?

tomastigera:
@alexanderstephan #7219 merge, could you review whether/how would your change apply? :pray:

alexanderstephan:
Yes, I just saw it. I am trying out if my tests still works, but I am having a bit trouble with the setup. If the test works we can safely close https://github.com/projectcalico/calico/issues/5250, I think.

alexanderstephan:
@tomastigera OK, my fix still seems relevant. Tests now passing again locally with the additional logic. Feel free to take another look, but looks good from my side.

Garfield96:
@tomastigera Can you please have a look at this PR? This feature/bug fix would be really useful.

felix/bpf/bpf_test.go
```diff
@@ -357,6 +360,30 @@ func TestXDP(t *testing.T) {
 		t.Fatalf("cannot create veth pair: %v\n%s", err, output)
 	}
 
+	// Load dummy BPF program.
+	progPath := filepath.Join(bpfDP.GetBPFCalicoDir(), "test")
+	err = bpfDP.loadBPF(filepath.Join("../testdata", "test.o"), progPath, "xdpgeneric", nil)
```
tomastigera, line 365: These days, you can use libbpf and set the name you want so you do not need to include the test.o https://github.com/projectcalico/calico/blob/master/felix/bpf/xdp/attach.go#L180

tomastigera (commented):
I think it in general LGTM except that there is no need to add the test.o

Pls note that this part of the bpf codebase is legacy, tbh, I am not even sure if we ever use these functions. This BPFLib code precedes the ebpf dataplane and common libbpf which the dataplane uses these days instead of shelling out to linux tools.

We should remobe big chunk of this code (some is still used). That would be a higly appreciated contribution :sweat_smile: 

This said, did you come across a usecase there you ran into calico removing other/your program from xdp?

Note that both ebpf and iptables dataplane use xdp for untracked policies, however, it is different code, see the link above. We never tested cohabiting with other xdp users, so if you have such a usecase, we are happy to work with you once fixing it.

PR #7424, opened 2023-03-03
add calicovpp fv tests

hedibouattour:
## Description
This PR is a first draft for fv tests in calico vpp.
A number of tests related to vpp are added to the fv test suite.
These tests require to have a running vpp instance and a running calicovpp agent server to receive Felix messages and configure vpp.
The general idea is to have these tests enabled by an env var for calicovpp testing, and disabled by default in calico fv test suite.


caseydavenport:
This one is pretty old, and I think we opted for another direction.

Issue #5218, opened 2021-09-07
Memory leak in calico-kube-controllers - v3.20


kppullin:
## Expected Behavior
Reasonable memory usage with no apparent memory leaks.

## Current Behavior
`calico-kube-controllers` v3.20 leaks memory (e.g: 9GB on a 16GB host) running on EKS 1.19.  Eventually this leads to pod evictions and alerts firing on the hosting k8s node until the controller pod is recycled.

After dumping all the strings in the process' memory space we see what appears to be the pod specs for all current **and** formerly running pods are in memory.  Why this _might_ impact us especially hard is that we utilize `Airflow` to run and schedule many frequent jobs in our cluster.

For example, from the `calico-kube-controllers` memory dump we obtained, we see `304052` copies of the following pod spec snippet:

```304052 {"f:metadata":{"f:labels":{".":{},"f:airflow-worker":{},"f:airflow_version":{},"f:app.kubernetes.io/name":{},"f:dag_id":{},"f:execution_date":{},"f:kubernetes_executor":{},"f:task_id":{},"f:try_number":{}}},"f:spec":{"f:affinity":{".":{},"f:nodeAffinity":{".":{},"f:requiredDuringSchedulingIgnoredDuringExecution":{".":{},"f:nodeSelectorTerms":{}}}},"f:containers":{"k:{\"name\":\"base\"}":{".":{},"f:args":{},"f:command":{},"f:env":{".":{},"k:{\"name\":\"AIRFLOW_HOME\"}"....```

Note that while `Airflow` managed pods appear most frequently, we see stale pod specs in the memory dump for non-airflow pods as well.

## Steps to Reproduce (for bugs)
Unknown.  Per above, _possibly_ by running creating and deleting pods frequently, particularly doing so via `Airflow` .

## Your Environment
* Kubernetes version (`kubectl version`): v1.19.13
* Operating System and version: EKS Amazon Linux - v1.19.13-eks-8df270 
* Manifest:
```
apiVersion: operator.tigera.io/v1
kind: Installation
metadata:
  name: default
spec:
  calicoNetwork:
    bgp: Enabled
    hostPorts: Enabled
    ipPools:
    - blockSize: 26
      cidr: 100.64.0.0/10
      encapsulation: IPIP
      natOutgoing: Enabled
      nodeSelector: all()
    linuxDataplane: Iptables
    mtu: 1440
    multiInterfaceMode: None
    nodeAddressAutodetectionV4:
[...]

Issue #5037, opened 2021-10-28
kube-controllers: Increased memory usage after update


fbsb:
After updating calico from version v3.19.1 to v3.20.2 we noticed the calico-kube-controller memory usage increased 
quite dramatically in our kubernetes clusters. 

small clusters (red, ~10 nodes): ~24MiB -> ~73MiB
medium clusters (blue, ~60 nodes): ~32MiB -> ~132MiB (peak 163MiB)
big clusters (yellow, ~300 nodes): ~64MiB -> ~800MiB (peak >1.2GiB)

![Screenshot from 2021-10-28 15-15-01](https://user-images.githubusercontent.com/8526438/139264481-4e6bacb0-d654-4e66-aea1-584f46079e72.png)
3

Also calico-node pods seem to have increased memory usage aswell but regardless of the cluster size ~200MiB -> ~600MiB

![Screenshot from 2021-10-28 15-26-08](https://user-images.githubusercontent.com/8526438/139264935-0152ff3e-1bbf-462e-af70-cb79c98a860a.png)

## Your Environment

* Calico version v3.20.2
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes v1.20.11
* Operating System and version: Flatcar Container Linux by Kinvolk 2905.2.

lmm:
A possibly related issue that was reported recently: https://github.com/projectcalico/kube-controllers/issues/850

Could you please provide your kube-controllers logs if possible? (cc @caseydavenport since you were looking at that other issue)

DockToFuture:
Any updates on the issue? This looks exactly like the same issue we experience after updating our calico version from v3.19.1 to v3.20.2.

caseydavenport:
> Also calico-node pods seem to have increased memory usage aswell but regardless of the cluster size ~200MiB -> ~600MiB

This is likely the same leak that was recently fixed in v3.21, and will be fixed in the next v3.20 patch release. Reference: https://github.com/projectcalico/calico/issues/5018


There is a known leak in kube-controllers in v3.20, tracked here: https://github.com/projectcalico/kube-controllers/issues/850

@DockToFuture @fbsb can you confirm whether or not you see similar logs in your kube-controllers pod? Something like this would indicate it is likely the same issue as the one linked above:

```
[...]

PR #7433, opened 2023-03-07
kube-controllers/node: Use pod/node indexes

dilyevsky:
## Description

In large clusters (>20K pods) with a lot of pod churn we found that calico-kube-controllers Node controller will leak memory in its internal pod cache within IPAM controller loop https://github.com/projectcalico/calico/blob/54ade211b5b81f71fa265a50d85afd134bcd51c3/kube-controllers/pkg/controllers/node/ipam.go#L191

We were able to track it down to the combination of https://github.com/kubernetes/kubernetes/pull/115620 (will file separate issue to bump this dependency) and extremely inefficient cache refresh loop that is issuing separate GET requests on every pod in the cluster on startup which causes Informer's internal FIFO queue to fall far behind and start dropping deletes.

This change addresses the above with:

* Re-use Informer's index to query the Pod/Node state - this is more efficient than paying the roundtrip to APIServer as the state it already cached.
* Enable re-sync on the SharedIndexInformer to guard against index getting out of sync
* Use `cache.DeletionHandlingMetaNamespaceKeyFunc` in the Informer's delete callback

## Related issues/PRs

Possibly related to #5218 and #5037

## Todos

- [x] Tests
- [ ] Documentation (n/a)
- [x] Release note

## Release Note

```release-note
Prevents Node kube-controller's internal pod cache from getting out-of-sync thus leaking memory.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/kube-controllers/cmd/kube-controllers/main.go
+++ b/kube-controllers/cmd/kube-controllers/main.go
@@ -450,7 +450,7 @@ type controllerControl struct {
 func (cc *controllerControl) InitControllers(ctx context.Context, cfg config.RunConfig, k8sClientset *kubernetes.Clientset, calicoClient client.Interface) {
 	// Create a shared informer factory to allow cache sharing between controllers monitoring the
 	// same resource.
-	factory := informers.NewSharedInformerFactory(k8sClientset, 0)
+	factory := informers.NewSharedInformerFactory(k8sClientset, 1*time.Minute)
 	podInformer := factory.Core().V1().Pods().Informer()
 	nodeInformer := factory.Core().V1().Nodes().Informer()
 
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -30,8 +30,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
+	v1lister "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/util/workqueue"
 
 	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
@@ -113,20 +112,20 @@ func init() {
 	prometheus.MustRegister(legacyBlocksGauge)
 }
 
-func NewIPAMController(cfg config.NodeControllerConfig, c client.Interface, cs kubernetes.Interface, ni cache.Indexer) *ipamController {
+func NewIPAMController(cfg config.NodeControllerConfig, c client.Interface, pi, ni cache.Indexer) *ipamController {
 	var leakGracePeriod *time.Duration
 	if cfg.LeakGracePeriod != nil {
 		leakGracePeriod = &cfg.LeakGracePeriod.Duration
 	}
 	return &ipamController{
-		client:    c,
-		clientset: cs,
-		config:    cfg,
-		rl:        workqueue.DefaultControllerRateLimiter(),
+		client: c,
+		config: cfg,
+		rl:     workqueue.DefaultControllerRateLimiter(),
 
 		syncChan: make(chan interface{}, 1),
 
-		nodeIndexer: ni,
+		podLister:  v1lister.NewPodLister(pi),
+		nodeLister: v1lister.NewNodeLister(ni),
 
 		// Buffered channels for potentially bursty channels.
 		syncerUpdates: make(chan interface{}, batchUpdateSize),
@@ -154,11 +153,11 @@ func NewIPAMController(cfg config.NodeControllerConfig, c client.Interface, cs k
 }
 
 type ipamController struct {
-	rl          workqueue.RateLimiter
-	client      client.Interface
-	clientset   kubernetes.Interface
-	nodeIndexer cache.Indexer
-	config      config.NodeControllerConfig
+	rl         workqueue.RateLimiter
+	client     client.Interface
+	podLister  v1lister.PodLister
+	nodeLister v1lister.NodeLister
+	config     config.NodeControllerConfig
 
 	syncStatus bapi.SyncStatus
 
@@ -918,7 +917,7 @@ func (c *ipamController) allocationIsValid(a *allocation, preferCache bool) bool
 	}
 	if p == nil {
 		logc.Debug("Querying Kubernetes API for pod")
-		p, err = c.clientset.CoreV1().Pods(ns).Get(context.Background(), pod, metav1.GetOptions{})
+		p, err = c.podLister.Pods(ns).Get(pod)
 		if err != nil {
 			if !errors.IsNotFound(err) {
 				log.WithError(err).Warn("Failed to query pod, assume it exists and allocation is valid")
@@ -1110,7 +1109,7 @@ func (c *ipamController) cleanupNode(cnode string) error {
 
 // nodeExists returns true if the given node still exists in the Kubernetes API.
 func (c *ipamController) nodeExists(knode string) bool {
-	_, err := c.clientset.CoreV1().Nodes().Get(context.Background(), knode, metav1.GetOptions{})
+	_, err := c.nodeLister.Get(knode)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return false
@@ -1129,17 +1128,9 @@ func (c *ipamController) nodeIsBeingMigrated(name string) (bool, error) {
 		return false, err
 	}
 	// Get node to inspect labels
-	obj, ok, err := c.nodeIndexer.GetByKey(kname)
-	if !ok {
-		// Node doesn't exist, so isn't being migrated.
-		return false, nil
-	}
+	node, err := c.nodeLister.Get(kname)
 	if err != nil {
 		return false, fmt.Errorf("failed to check node for migration status: %w", err)
-	}
-	node, ok := obj.(*v1.Node)
-	if !ok {
-		return false, fmt.Errorf("failed to check node for migration status: unexpected error: object is not a node")
 	}
 
 	for labelName, labelVal := range node.ObjectMeta.Labels {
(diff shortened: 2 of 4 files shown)


kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -450,7 +450,7 @@ type controllerControl struct {
 func (cc *controllerControl) InitControllers(ctx context.Context, cfg config.RunConfig, k8sClientset *kubernetes.Clientset, calicoClient client.Interface) {
 	// Create a shared informer factory to allow cache sharing between controllers monitoring the
 	// same resource.
-	factory := informers.NewSharedInformerFactory(k8sClientset, 0)
+	factory := informers.NewSharedInformerFactory(k8sClientset, 1*time.Minute)
```
odinuge, line 453: This is configuring the resync period right? If so, this will probably cause the queue to have more pressure, and you will most likely make the problem described in https://github.com/kubernetes/kubernetes/issues/115658 worse.

Since this will not cause a new list call to k8s, it will only put all the objects in the cache, including the leaked ones, on the queue once a minute.

odinuge: I initially thought setting a resync period would force a relist, but turns out that isn't the case; and it makes sense as well, since it shouldn't be necessary.

lwr20:
Looks like `make ci` hit errors, you should probably be able to see them if you run it locally, shout if not

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1129,18 +1128,10 @@ func (c *ipamController) nodeIsBeingMigrated(name string) (bool, error) {
 		return false, err
 	}
 	// Get node to inspect labels
-	obj, ok, err := c.nodeIndexer.GetByKey(kname)
-	if !ok {
-		// Node doesn't exist, so isn't being migrated.
-		return false, nil
-	}
+	node, err := c.nodeLister.Get(kname)
```
aojea, line 1131: if you change the indexer by the lister, you should replace the !ok by err == ErrNotFound

aojea: or you will fail now if the object is not found, that is an important behavior change

caseydavenport:
Thanks for raising this - I am out sick right now but will take a detailed look at this once I'm back.

kube-controllers/cmd/kube-controllers/main.go
odinuge: > it doesn't force a relist but IIRC it generates a "fake" update for each object in the cache, that is an important performance change ... I only see it used to "force" a reconcilation, typically when your controller has a dependency on an external system that doesn't have watches

Yee. I was mostly talking about it in the context of this PR and https://github.com/kubernetes/kubernetes/issues/115658, and from how I read the logic changed in this PR, there are no parts of the modified code that would benefit from those extra "fake" resync events, right?

There might be other things in calicio that could benefit from it tho., but I don't know the project well enough to know. So there could be a chance that adding a resync period is a good idea

dilyevsky: > there are no parts of the modified code that would benefit from those extra "fake" resync events, right?

Yeah it doesn't do a relist. My thinking was that since we now rely heavily on indexes being consistent as opposed to re-querying apiserver which was pretty slow on large clusters, this adds extra "insurance" that everything will be there in a good state. In our mitigation testing on the clusters where upstream was causing problems we've seen much faster processing and resync seems to have no noticeable negative effects ¯\_(ツ)_/¯

kube-controllers/pkg/controllers/node/ipam.go
caseydavenport: Agreed, we need to handle all three cases - node exists, node doesn't exist, and an error when determining existence.

kube-controllers/cmd/kube-controllers/main.go
aojea: at scale you'll have performance problems for sure

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -918,7 +917,7 @@ func (c *ipamController) allocationIsValid(a *allocation, preferCache bool) bool
 	}
 	if p == nil {
 		logc.Debug("Querying Kubernetes API for pod")
-		p, err = c.clientset.CoreV1().Pods(ns).Get(context.Background(), pod, metav1.GetOptions{})
+		p, err = c.podLister.Pods(ns).Get(pod)
```
caseydavenport, line 920: So, the way this code was originally intended to work is a bit confusing. Here's what I remember:

The controller has its own `c.podCache`, which is populated with pods as we get events from the pod informer. It's inefficient, because we're keeping two copies of each pod (one within the informer's cache and another in this cache) but it allowed us to prime this cache on misses ourselves (as is done on line 932).

The reason we have the clientset call to get the pod here instead of always using the pod lister as you have done here is because we were worried about scenarios where the informer might be behind actual cluster state, and we wanted a way to be 100% confident that the state of the pod was what we thought it was.

The intention was that this line would only be hit in rare situations, since most of the time podCache would be populated by the informer and we would have a cache hit, skipping this branch. But it seems that's not what you're experiencing, so I wonder why that is.

The only other time we would hit this is if `preferCache` is `false`, which is set [here](https://github.com/projectcalico/calico/blob/f413cf361aaceb84fa12696468d7cc08e72ca2b4/kube-controllers/pkg/controllers/node/ipam.go#L1056) only after we have made a decision to garbage collect an IP address, so we can do a final check that we're not acting on obsolete information about the pod. The assumption here again was that this would be called infrequently enough that a direct pod GET wouldn't have much cost.

All of this clunky code was a trade-off to increase our confidence, since the action of garbage collecting an IP address can be pretty problematic if the IP address itself is actually still in use somehow!

All of that said, I am not necessarily against the changes in this PR, but I need to think through them a little more and I think there are probably some other consequences. For example, I don't think `podCache` is serving any purpose any more with your changes so can probably be removed entirely. The other thing I need to consider is whether or not we still want that final escape hatch for `preferCache=false` from line 1056.

kube-controllers/cmd/kube-controllers/main.go
caseydavenport: > there are no parts of the modified code that would benefit from those extra "fake" resync events, right?

I don't believe so. In the context of this suite of controllers in particular, with the changes you've made I think we could ignore the output of informer events entirely and just use its cache. The only thing that the events are used for right now is... populating another cache :shrug: 

I'd probably go the other way - don't send the fake resync events, remove `c.podCache` and the code that populates it from those events. I mentioned this a bit in my other comment.

kube-controllers/pkg/controllers/node/ipam.go
dilyevsky: >  I'm thinking how we can solve the issue I've outlined above while maintaining consistent state. Open to any suggestions.

How about go to informer indexer if `perferCache=true` and do standalone GET otherwise (on GC)? I need to test this but I think this should do it.

kube-controllers/cmd/kube-controllers/main.go
dilyevsky: > I'd probably go the other way - don't send the fake resync events, remove c.podCache and the code that populates it from those events. I mentioned this a bit in my other comment.

Did exactly this. PTAL

dilyevsky:
Huh weird, you can only re-request one person. Anyways @odinuge @odinuge @caseydavenport thanks for review, PTAL

kube-controllers/pkg/controllers/node/ipam.go
caseydavenport: > This causes thousands of GETs (one for each pod) in large clusters and informer falls far behind. I suspect the clientset also self-throttles itself and that affects the informer as well.

Ick! Yep, that definitely wasn't the original intention at least. We might be missing a signal somewhere to indicate that the informer has finished synchronizing? 

> How about go to informer indexer if perferCache=true and do standalone GET otherwise (on GC)? I need to test this but I think this should do it.

This sounds like a good approach to me :+1:

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -252,23 +240,10 @@ func (c *ipamController) acceptScheduleRequests(stopCh <-chan struct{}) {
...
-			for i = 1; i < batchUpdateSize; i++ {
-				select {
-				case pu = <-c.podUpdate:
-					c.handlePodUpdate(pu)
-				default:
-					break
-				}
-			}
 		case upd := <-c.syncerUpdates:
+			log.Infof("Handling syncer updatges")
```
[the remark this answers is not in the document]
dilyevsky: Yep, removed it, thanks

dilyevsky:
Fixed tests. Should be ready to merge after CI run

dilyevsky:
Hm the Semaphore error since not related to changes in this PR:
```
* Pulling artifacts failed: exit status 1 00:01
125[Mar 14 13:29:08.334] Error pulling artifact: failed to generate signed URLs - hub returned 404 status code 00:01
126[Mar 14 13:29:08.334] Please check if the artifact you are trying to pull exists. 00:01
127Error: exit status 1
```

Is the pipeline broken or I'm looking at the wrong thing?

caseydavenport (approved):

caseydavenport:
I think something flaked, re-running and hoping they pass this time :+1:

dilyevsky:
@caseydavenport looks like a timeout now? I can tell what is going on there 😂 Mind taking a look please

caseydavenport:
Yeah, they all seem to be timing out, so guessing this is a real issue. It's not obvious to me from the logs what's happening, but it seems like one of the tests might be deadlocked / hanging?

dilyevsky:
@caseydavenport I messed up couple of tests in ipam_test.go so they were hanging. Re-ran ` make -C kube-controllers ci` without failures. PTAL

dilyevsky:
Is there automation for cherry-picks or should I create one manually for v3.25? Any other releases?

caseydavenport:
No automation for cherry-picks, so will need to submit that manually I'm afraid.

@mgleung is the release czar and can help with which versions.

Thanks so much for working on this!

dilyevsky:
@caseydavenport @mgleung followed instructions at https://github.com/projectcalico/calico/blob/f3a8ef66832a99b07cc39fd20f4fc2f44abec109/CONTRIBUTING_CODE.md#patching-an-older-release and created https://github.com/projectcalico/calico/pull/7503 but I can't edit this PR labels to add `cherry-pick-candidate`...

PR #7453, opened 2023-03-10
Validate RouteReflector IP

sridhartigera:
## Description

Route reflectors can be configured using calicoctl or my annotating the node resource directly.
While the calicoctl configuration goes through validations, the annotation value is not validated if the node resource is directly annotated. This PR adds validation to this IP.

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/k8s/resources/node.go
+++ b/libcalico-go/lib/backend/k8s/resources/node.go
@@ -38,6 +38,7 @@ import (
 	cerrors "github.com/projectcalico/calico/libcalico-go/lib/errors"
 	"github.com/projectcalico/calico/libcalico-go/lib/json"
 	"github.com/projectcalico/calico/libcalico-go/lib/net"
+	validatorv3 "github.com/projectcalico/calico/libcalico-go/lib/validator/v3"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


libcalico-go/lib/backend/k8s/resources/node.go
```diff
@@ -230,7 +230,20 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node, usePodCIDR bool) (*model.KVPair, erro
 	annotations := k8sNode.ObjectMeta.Annotations
 	bgpSpec.IPv4Address = annotations[nodeBgpIpv4AddrAnnotation]
 	bgpSpec.IPv6Address = annotations[nodeBgpIpv6AddrAnnotation]
-	bgpSpec.RouteReflectorClusterID = annotations[nodeBgpCIDAnnotation]
+	// Validate the IP address specified for route reflector.
+	// When route reflector is configured using calicoctl, the address
+	// is already validated. When the node is annotated directly without
+	// using calicoctl, the IP address specified must be validated here.
+	routeReflectorClusterID := annotations[nodeBgpCIDAnnotation]
```
caseydavenport, line 237: We're going to need this logic for all of the annotations eventually. I might do it like this for re-use:

```go
type validatorFunc func(string) error

func getAnnotation(key string, validator validatorFunc) string
```

and then

```go
bgpSpec.RouteReflectorClusterID = getAnnotation(nodeBgpCIDAnnotation, validateIPv4Address)
```

libcalico-go/lib/backend/k8s/resources/node.go
```diff
@@ -230,7 +232,11 @@ func K8sNodeToCalico(k8sNode *kapiv1.Node, usePodCIDR bool) (*model.KVPair, erro
 	annotations := k8sNode.ObjectMeta.Annotations
 	bgpSpec.IPv4Address = annotations[nodeBgpIpv4AddrAnnotation]
```
caseydavenport, line 233: I'd personally be in favor of using the logic we just wrote to also validate the other IP annotations - v4 / v6 addresses. They have the same behavior. 

There are a few others below as well - Wireguard IPs, tunnel IPs

libcalico-go/lib/backend/k8s/resources/node_test.go
```diff
@@ -77,6 +77,84 @@ var _ = Describe("Test Node conversion", func() {
 		Expect(asn.String()).To(Equal("2546"))
 	})
 
+	It("should ignore RR cluster ID if its an invalid IPv4 address", func() {
+		l := map[string]string{"net.beta.kubernetes.io/role": "master"}
```
caseydavenport, line 81: what is this label? is it needed for this test?

caseydavenport: At a minimum this should be "control-plane" instead of master, but I don't actually think we need any labels for this test so could just be removed.

caseydavenport (approved):

PR #7468, opened 2023-03-15
Felix: int dataplane main loop cleanups

fasaxc:
## Description

Upstreaming some changes form Enterprise.  We needed an "in sync" message from the iface monitor in Enterprise and the code was very messy so it needed a refactor.

- Add an in-sync message to the interface monitor. (Required combining its channels into one for synchronisation).
- Move 'peek' loops out of main loop method and introduce common drainChan() helper.
- Block dataplane updates until interface monitor is in sync.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/int_dataplane.go
```diff
@@ -1871,6 +1780,127 @@ func (d *InternalDataplane) loopUpdatingDataplane() {
...
+		d.processIfaceInSync()
+	}
+}
+
+func (d *InternalDataplane) processIfaceInSync() {
+	if d.ifaceMonitorInSync {
+		return
+	}
+	log.Info("Interface monitor now in sync.")
+	d.ifaceMonitorInSync = true
```
tomastigera, line 1860: does this ever become false? Seems like the iface monitor may set it many times to true

tomastigera (commented):
LGTM

felix/dataplane/linux/int_dataplane.go
fasaxc: No, it doesn't; left over from previous iteration.

fasaxc: No, it's a one-way flag; just to let the main loop know that we're in sync at one point.

PR #7518, opened 2023-03-30
[BPF] BPFEnforceRPF is Loose by default

tomastigera:
In environments like bonded main device or AKS's main slave device, the default BPFDataIfacePattern setting would attach tc programs to a cascade of devices. That leads to the Strict (previous default) option to drop packets on the devices that are not in route tables. Although attaching to all these devices is unnecessary and makes a packet being processed multiple times at a host endpoint, it is not technically incorrect.

If Strict option is required,it has to be set explicitly and the BPFDataIfacePattern changed accordingly in such environments, that is, for instance, exclude ethX  and add bond0 where bonded devices are used.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: BPFEnforceRPF is Loose by default to avoid issues in some environments. If Strict option is required,it has to be set explicitly and the BPFDataIfacePattern may need to be changed accordingly to avoid attaching to "slave" devices.
```

## Reminder for the reviewer

Other optional labels:


sridhartigera (approved):

tomastigera:
All test were :green_circle:

Issue #7396, opened 2023-02-27
feature request: Hardware addresses that support pods have an same prefix


cyclinder:
## Expected Behavior

Support custom mac address to the pod, For example, add a field to the CNI configuration file: `mac_prefix`:

```shell
root@controller:~# cat /etc/cni/net.d/10-calico.conflist
{
  "name": "calico",
  "cniVersion":"0.3.1",
  "plugins":[
    {
      "datastore_type": "kubernetes",
      "nodename": "controller",
      "type": "calico",
      "log_level": "info",
      "log_file_path": "/var/log/calico/cni/cni.log",
      "mac_prefix": "1a:3b",    # <======
      "ipam": {
        "type": "calico-ipam",
        "assign_ipv6": "true",
        "ipv6_pools": ["fd85:ee78:d8a6:8607::1:0000/112"],
        "ipv4_pools": ["10.233.64.0/18"],
        "assign_ipv4": "true"
      },
      "policy": {
        "type": "k8s"
      },
      "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/calico-kubeconfig"
      }
    },
    {
      "type":"portmap",
      "capabilities": {
        "portMappings": true
      }
    },
    {
      "type":"bandwidth",
      "capabilities": {
        "bandwidth": true
      }
    }
  ]
}
```

This is a global configuration, When a Pod is created, the mac addresses of all Pods will have the same prefix：`0a:1b`. The complete mac address is stitched together from two parts: `mac_prefix` + `hexcode(ip)`. For example:

```
root@controller:~# kubectl get po -o wide
NAME                                        READY   STATUS      RESTARTS      AGE    IP               NODE         NOMINATED NODE   READINESS GATES
test-794b96cd5b-d92pp                       1/1     Running     0             132m   10.233.119.208   controller   <none>           <none>
root@controller:~# kubectl exec -it test-794b96cd5b-d92pp -- ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: tunl0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN qlen 1000
[...]

PR #7524, opened 2023-04-03
Support configuring pods with the same mac address prefix

cyclinder:
## Description

Support configuring pods with the same mac address prefix via annotations: `cni.projectcalico.org/hwAddr`. and added unit-tests for it

## Related issues/PRs

https://github.com/projectcalico/calico/issues/7396

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
+++ b/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
@@ -15,11 +15,16 @@
 package linux
 
 import (
+	"bytes"
 	"context"
+	"encoding/hex"
 	"fmt"
 	"io"
+	"math/big"
 	"net"
+	"net/netip"
 	"os"
+	"regexp"
 	"syscall"
 	"time"
 
@@ -170,21 +175,12 @@ func (d *linuxDataplane) DoNetworking(
 		// Check if there is an annotation requesting a specific fixed MAC address for the container Veth, otherwise
 		// use kernel-assigned MAC.
 		if requestedContVethMac, found := annotations["cni.projectcalico.org/hwAddr"]; found {
-			tmpContVethMAC, err := net.ParseMAC(requestedContVethMac)
+			contVethMAC, err = resetHardwareAddr(contVeth, requestedContVethMac, result.IPs[0].Address.IP.String())
 			if err != nil {
-				return fmt.Errorf("failed to parse MAC address %v provided via cni.projectcalico.org/hwAddr: %v",
-					requestedContVethMac, err)
+				return err
 			}
-
-			err = netlink.LinkSetHardwareAddr(contVeth, tmpContVethMAC)
-			if err != nil {
-				return fmt.Errorf("failed to set container veth MAC to %v as requested via cni.projectcalico.org/hwAddr: %v",
-					requestedContVethMac, err)
-			}
-
-			contVethMAC = tmpContVethMAC.String()
-			d.logger.Infof("successfully configured container veth MAC to %v as requested via cni.projectcalico.org/hwAddr",
-				contVethMAC)
+			d.logger.Infof("successfully reset container veth's mac to %v with cni.projectcalico.org/hwAddr: %v",
+				contVethMAC, requestedContVethMac)
 		} else {
 			contVethMAC = contVeth.Attrs().HardwareAddr.String()
 		}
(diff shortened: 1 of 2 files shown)


cyclinder:
/cc @mgleung 

could you please review this? Thanks :)

mgleung:
@cyclinder sorry for the late reply. I've been a bit preoccupied but I think @mazdakn should be able to take a look.

cyclinder:
Hi @mazdakn Can you please take a look if you have free time?

mazdakn:
@cyclinder I am going to review it today.

mazdakn:
@cyclinder have you tested the functionality in a real cluster?

cyclinder:
Hi @mazdakn I'm sorry for the delay.  I've tested this feature on a real cluster and it works fine, I think this PR is ready to merge, Can you take a look? thanks.

cyclinder:
```shell
root@controller:~# kubectl get deploy dao-2048 -o yaml | grep calico
        cni.projectcalico.org/hwAddr: 0a:2b
root@controller:~# kubectl get po -o wide | grep dao-2048
dao-2048-bccd8459f-jh7lj                                    1/1     Running       0          17s    10.233.105.190   worker1      <none>           <none>
root@controller:~# kubectl exec -it dao-2048-bccd8459f-jh7lj -- ip a show eth0
4: eth0@if547069: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1430 qdisc noqueue state UP qlen 1000
    link/ether 0a:2b:0a:e9:69:be brd ff:ff:ff:ff:ff:ff
    inet 10.233.105.190/32 scope global eth0
       valid_lft forever preferred_lft forever
```

mazdakn:
@cyclinder thanks for testing. But did you check if a switch can see the pod's mac address? Because in Calico design this is impossible, so I would be confused if a switch can see pod's mac address.

cyclinder:
@mazdakn I've checked it, and I found that a switch can't see the pod's mac address indeed, I'm a little confused about our customers' needs, but they do classify deployments by Mac address prefixes. So I think this feature is still make a sense, do we still have a chance to merge it?

mazdakn:
@cyclinder we expand the feature set based on the use case. If the use case is not valid, then it does not make sense to merge a code as it would make it harder in log run to maintain and keep the code quality. I believe we should close this PR and GH issue. Does it make sense?

cyclinder:
@mazdakn Sorry, I just got confirmation from our users and PM: In our user scenario, the pod's IP is fixed, and part of the pod's Mac address is translated through a fixed IP address, so its Mac address is also fixed.

We have a server in our cluster that makes licenses, which will make licenses based on information such as the IP and MAC address of the application clients so that each license is unique.

I'm sorry for my previous words about the needs, That requirement is to deploy some underlay pods in our cluster, such as macvlan, etc., the IP of these pods belongs to the real underlay subnet, their gateway is located at the switch, and does not forward network traffic through the host network stack, so we will do access control in the switch based on the MAC address of the pod. yeah, which is a different scenario.

mazdakn:
@cyclinder thanks for the update. Can you please close this PR and also the related GH issue?

cyclinder:
@mazdakn Oh sorry, I think this feature is an enhancement from the original, It was just possible to customize the pod's MAC address, now we can fix the pod's MAC address, If the reason for not being able to merge is just because of maintaining the code quality, I'm a bit hard to accept 😅, and if you can suggest some code comments I would be very appreciated.

This PR maintains backward compatibility and does not affect the previous way of using it, and our users really need this feature. So please consider it again, Thanks.

mazdakn:
@cyclinder we cannot merge a code when there is no use case for it. Just because we can enhance a config does not justify having it when the actual use case is invalid.

lwr20:
> and our users really need this feature.

@cyclinder this is the part @mazdakn needs to be convinced of.  You haven't shown _why_ this is useful to your users yet.

cyclinder:
@mazdakn @lwr20  Let me introduce the user case. We have a server in our cluster that makes licenses, which will make licenses based on information such as the IP and MAC address of the application clients so that each license is unique. So we want the _IP and MAC address of the pod to be fixed._

Refer to https://github.com/projectcalico/calico/pull/6249, Calico currently supports assigning specified MAC addresses to pods, So _why_ not use it?

* This can only be for the pod, not a deployment. Otherwise, each replica of the deployment will have the same MAC address.
* Our client pod will go to the license server to apply for a license when it is created, and the license server will issue a license based on the IP and Mac address of the client pod. So we need to annotate the following for every pod:

```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    "cni.projectcalico.org/ipAddrs": "[\"192.168.0.1\"]"
    "cni.projectcalico.org/hwAddr": "1c:0c:0a:c0:ff:ee"
```

We need to make sure that the license is still valid after the pod is restarted. But we need to annotate these for each one, and when there are many such pods, it becomes difficult for administrators to maintain...

We want to reduce the work of administrators in the following ways:

```shell
apiVersion: apps/v1
kind: Deployment
metadata:
  name: client-2048
spec:
  replicas: 10
  selector:
    matchLabels:
      app: dao-2048
  template:
    metadata:
      annotations:
        cni.projectcalico.org/ipv4pools: '["pool-1"]'
        cni.projectcalico.org/hwAddr: 0a:2b
```

Each replica of this deployment is assigned an IP address from pool1, and the mac address of each replica is fixed as: `0a:1b + convert(IP)`,  Even if they are restared, the license is still valid, so this greatly reduces the administrator's work.

This is our user case, I don't know if you can understand my description, thank you for your comment.

lwr20:
> We have a server in our cluster that makes licenses, which will make licenses based on information such as the IP and MAC address of the application clients so that each license is unique. So we want the IP and MAC address of the pod to be fixed.

So this has nothing to do with networking, which is the point of Calico, and is entirely about a workaround for a legacy app which was written in a time before Kubernetes, when applications ran on real hardware and therefore IPs and MACs didn't change often and could be used to create a unique identifier for a machine.

The "correct" way to fix this would be to use some other unique identifier for the license server.  Presumably the license server or client is not something you are able to change the code in?

If we "fix" this in Calico, we add complexity and code that is only useful to a small subset of Calico's users, with all the subsequent maintenance and potential bugs that implies, which is why there is hesitation about accepting this.  

That said, it is true that https://github.com/projectcalico/calico/pull/6249 added function for something very very similar, and this is just improving that.

@caseydavenport  can you make a call to accept/reject this enhancement please?

cyclinder:
> Presumably the license server or client is not something you are able to change the code in?

Yeah, We can't change their code.

> could you solve this another way?

I once wrote some codes for this user case (now stored [here](https://github.com/spidernet-io/spiderpool/blob/d57a543aa841cc4d5dbb97e87b04f1b6335cb97f/pkg/networking/networking/mac.go#L22), This also is the changes in this PR). We build it into a CNI binary and worked in a CNI chain after calico was called to update the pod's Mac address at the cost of one more CNI call, which is just a temporary solution, so I want to donate this part of the code to calico, that's great.

I understand that you are concerned about the quality of this part of the code, and which we have used in users environments, For now, no issues have been reported, and we have full E2E for it Test (https://github.com/spidernet-io/spiderpool/blob/d57a543aa841cc4d5dbb97e87b04f1b6335cb97f/test/e2e/coordinator/macvlan-overlay-one/macvlan_overlay_one_test.go#L251).

I think we just improved https://github.com/projectcalico/calico/pull/6249.

caseydavenport:
> Otherwise, each replica of the deployment will have the same MAC address

@cyclinder could you maybe explain why duplicate MAC addresses for pods in the same deployment is a problem in your scenario? In a Calico network, there's no functional problem with two pods having the same MAC address, since routing is point-to-point and there's no switching on MAC involved. I'm not sure how the license server uses this information, but it sounds like it's using both IP and MAC, so can produce a unique license based on the combination?

I tend to agree with @fasaxc here that a mutating webhook seems to be the most elegant solution here. Calico provides a way to set a specific MAC address per-pod already, and a mutating webhook allows you to leverage that existing code. I'd really like to avoid having two very similar config options when one would suffice.

cyclinder:
@caseydavenport We just want a unique IP and MAC address of the pod, Since the license server needs this infos.

> In a Calico network, there's no functional problem with two pods having the same MAC address,

OK, it looks like there are no functional problem with two pods having the same MAC address indeed, but every pod should have a unique IP and Mac address from common sense, right?

> I tend to agree with @fasaxc here that a mutating webhook seems to be the most elegant solution here.

It's a choice, but sad I think we don't want this, It's a bit expensive for us... We need to develop additional components and maintain them. Also, I once wrote these codes to a CNI meta-plugin, It works well in my production. It is still used through Calico's annotations.

> I'd really like to avoid having two very similar config options when one would suffice.

Yeah, I understand this, I can accept this if you guys don't really want to merge the PR.

PR #7525, opened 2023-04-03
Fix static-check with golanglint-ci v1.52.2

hjiawei:
## Description

This change fixes static-check with golanglint-ci v1.52.2.

## Related issues/PRs

This change is needed when we include the new golanglint-ci to our go-build.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/internal/pkg/testutils/utils_linux.go
+++ b/cni-plugin/internal/pkg/testutils/utils_linux.go
@@ -265,7 +265,10 @@ func RunCNIPluginWithId(
 	if version.Compare(nc.CNIVersion, "0.3.0", "<") {
 		// Special case for older CNI versions.
 		var out []byte
-		out, err = json.Marshal(r)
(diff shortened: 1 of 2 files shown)


cni-plugin/internal/pkg/testutils/utils_linux.go
```diff
@@ -265,7 +265,7 @@ func RunCNIPluginWithId(
 	if version.Compare(nc.CNIVersion, "0.3.0", "<") {
 		// Special case for older CNI versions.
 		var out []byte
-		out, err = json.Marshal(r)
+		out, _ = json.Marshal(r)
```
mgleung, line 268: I understand that we wouldn't normally expect `Marshal` to error, but shouldn't we include a case here to check just in case?

mgleung (approved):

PR #7526, opened 2023-04-04
[BPF] install rules to mark pre-established flows asap

tomastigera:
To avoid disruption to the flows, mark install the rules asap so that the tc programs get the information about existing conntrack immediately.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/int_dataplane.go
+++ b/felix/dataplane/linux/int_dataplane.go
@@ -1367,6 +1367,7 @@ func (d *InternalDataplane) doStaticDataplaneConfig() {
 	d.configureKernel()
 
 	if d.config.BPFEnabled {
+		d.setUpIptablesBPFEarly()
 		d.setUpIptablesBPF()
 	} else {
 		d.setUpIptablesNormal()
@@ -1381,6 +1382,18 @@ func (d *InternalDataplane) doStaticDataplaneConfig() {
 	} else {
 		log.Info("IPIP disabled. Not starting tunnel update thread.")
 	}
+}
+
+func bpfMarkPreestablishedFlowsRules() []iptables.Rule {
+	return []iptables.Rule{{
+		Match: iptables.Match().
+			ConntrackState("ESTABLISHED,RELATED"),
+		Comment: []string{"Mark pre-established flows."},
+		Action: iptables.SetMaskedMarkAction{
+			Mark: tcdefs.MarkLinuxConntrackEstablished,
+			Mask: tcdefs.MarkLinuxConntrackEstablishedMask,
+		},
+	}}
 }
 
 func (d *InternalDataplane) setUpIptablesBPF() {
@@ -1416,17 +1429,7 @@ func (d *InternalDataplane) setUpIptablesBPF() {
 		)
 
 		// Mark traffic leaving the host that already has an established linux conntrack entry.
-		outputRules = append(outputRules,
-			iptables.Rule{
-				Match: iptables.Match().
-					ConntrackState("ESTABLISHED,RELATED"),
-				Comment: []string{"Mark pre-established host flows."},
-				Action: iptables.SetMaskedMarkAction{
-					Mark: tcdefs.MarkLinuxConntrackEstablished,
-					Mask: tcdefs.MarkLinuxConntrackEstablishedMask,
-				},
-			},
-		)
+		outputRules = append(outputRules, bpfMarkPreestablishedFlowsRules()...)
 
 		for _, prefix := range rulesConfig.WorkloadIfacePrefixes {
 			fwdRules = append(fwdRules,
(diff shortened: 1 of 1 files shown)


felix/dataplane/linux/int_dataplane.go
```diff
@@ -1563,6 +1556,26 @@ func (d *InternalDataplane) setUpIptablesBPF() {
...
+	cp := []iptables.Rule{}
+
+	for _, r := range rules {
+		r.Comment = append(r.Comment, "cali:BOOTSTRAP")
+		cp = append(cp, r)
+	}
+
+	for _, t := range d.iptablesFilterTables {
+		if err := t.InsertRulesNow("FORWARD", cp); err != nil {
+			log.Warn("Failed inserting some early rules to filter FORWARD, some flows may get temporarily disrupted.")
```
fasaxc, line 1571:
```suggestion
			log.WithError(err).Warn("Failed inserting some early rules to filter FORWARD, some flows may get temporarily disrupted.")
```

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1563,6 +1556,26 @@ func (d *InternalDataplane) setUpIptablesBPF() {
...
+		}
+		if err := t.InsertRulesNow("OUTPUT", cp); err != nil {
+			log.Warn("Failed inserting some early rules to filter OUTPUT, some flows may get temporarily disrupted.")
```
fasaxc, line 1574:
```suggestion
			log.WithError(err).Warn("Failed inserting some early rules to filter OUTPUT, some flows may get temporarily disrupted.")
```

felix/iptables/table.go
```diff
@@ -1346,6 +1293,88 @@ func (t *Table) applyUpdates() error {
 	return nil
 }
 
+func (t *Table) writeOutBuffer(buf *RestoreInputBuilder, features *environment.Features) error {
```
fasaxc, line 1296:
```suggestion
func (t *Table) execIptablesRestore(buf *RestoreInputBuilder) error {
```
Think I'd call this `execIptablesRestore` and could you do `features := t.featureDetector.GetFeatures()` inside the function to avoid passing something that's easy to get?  (I think the features are cached so GetFeatures should be fast.)

felix/iptables/table.go
```diff
@@ -1346,6 +1293,88 @@ func (t *Table) applyUpdates() error {
...
+// InsertRulesNow insets the given rules immediately without removing or syncing
+// other rules. This is primarily useful when bootstrapping and we cannot wait
+// until we have the full state.
+func (t *Table) InsertRulesNow(chain string, rules []Rule) error {
+	features := t.featureDetector.GetFeatures()
+
+	buf := new(RestoreInputBuilder)
+	buf.StartTransaction(t.Name)
+	for i, r := range rules {
+		buf.WriteLine(r.RenderInsertAtRuleNumber(chain, i+1, "", features))
```
fasaxc, line 1367: Instead of adding the BOOTSTRAP comment outside, how about making this method set it?  Feels like that'd be a nice encapsulation; so the table code would be responsible for making sure bootstrap rules get cleaned up.

The table code has a nice UT harness that simulates iptables, would be good to have UT of the new function.

I don't think it's too important, but you could go the whole hog and calculate the proper rule hashes here.  Then if they match correctly later, we'll avoid churning the rules.

fasaxc (changes requested):
Can we check if the rules are present before activating the workaround? Two reasons I think that's worth doing:

* On a system with a lot of iptables rules (e.g. because kube-proxy is still running), the iptables transaction could take a few seconds, which will slow down start-up.  Suggestion: add a distinctive comment to the rules (both bootstrap and non-bootstrap versions) that you can grep for in `iptables-save`
* If felix ever goes into a cyclic restart where bootstrap rules go in but not the real ones, we'll build up rules.

felix/iptables/table.go
tomastigera: yeah, right, I has this funtion in util before, not in table :+1:

felix/iptables/table.go
tomastigera: > The table code has a nice UT harness that simulates iptables, would be good to have UT of the new function.

The harness needed some fixing ;-)

tomastigera:
> Can we check if the rules are present before activating the workaround? Two reasons I think that's worth doing:
> 
> * On a system with a lot of iptables rules (e.g. because kube-proxy is still running), the iptables transaction could take a few seconds, which will slow down start-up.  Suggestion: add a distinctive comment to the rules (both bootstrap and non-bootstrap versions) that you can grep for in `iptables-save`
> * If felix ever goes into a cyclic restart where bootstrap rules go in but not the real ones, we'll build up rules.

Done :heavy_check_mark:

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1563,6 +1556,42 @@ func (d *InternalDataplane) setUpIptablesBPF() {
 	}
 }
 
+// setUpIptablesBPFEarly that need to be written asap
+func (d *InternalDataplane) setUpIptablesBPFEarly() {
+	rules := bpfMarkPreestablishedFlowsRules()
+
+	for _, t := range d.iptablesFilterTables {
+		// We want to prevent inserting the rules over and over again if something later
+		// crashed. We do not expect that we would inset just a part of the batch as that
```
fasaxc, line 1565:
```suggestion
		// crashed. We do not expect that we would insert just a part of the batch as that
```

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1563,6 +1556,42 @@ func (d *InternalDataplane) setUpIptablesBPF() {
...
+		// should be handled by the iptables-restore transaction.  Never the less if we
+		// see that unexpected case, perhaps due to an upgrade, we skip over updating the
+		// iptables now and will wait for the full resync. That could be temporarily
+		// disrupting.
+		if present := t.CheckRulesPresent("FORWARD", rules); present != nil {
+			if len(present) != len(rules) {
+				log.WithField("present rules", present).
```
fasaxc, line 1572:
```suggestion
				log.WithField("presentRules", present).
```
I generally treat these like variable names

fasaxc (approved):
LGTM, just a couple of typos

marvin-tigera:

PR #7570, opened 2023-04-19
Fix links in custom-resources.yaml manifest

krnowak:
These were 404.


fasaxc:
:wave: and thanks for the PR.  I think those files are autogenerated from the `.go` files in the operator repo so they need to be updated there first.

fasaxc:
Ah OK, right you are; I thought it was the CRDs you were editing bu it's the CRs.

caseydavenport:
Sorry for the delay!

PR #7574, opened 2023-04-20
fix the err check

winstonfly:
## Description
as original err check block make me confused. so i changed it for easy to read.

### DIFF
--- a/cni-plugin/pkg/install/install.go
+++ b/cni-plugin/pkg/install/install.go
@@ -526,20 +526,16 @@ func setSuidBit(file string) error {
 func destinationUptoDate(src, dst string) (bool, error) {
 	// Stat the src file.
 	f1info, err := os.Stat(src)
-	if os.IsNotExist(err) {
+	if os.IsNotExist(err) || err != nil {
 		// If the source file doesn't exist, that's an unrecoverable problem.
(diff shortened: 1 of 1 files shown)


cni-plugin/pkg/install/install.go
```diff
@@ -526,20 +526,16 @@ func setSuidBit(file string) error {
...
 		// If the source file doesn't exist, that's an unrecoverable problem.
 		return false, err
-	} else if err != nil {
-		return false, err
 	}
 
 	// Stat the dst file.
 	f2info, err := os.Stat(dst)
-	if os.IsNotExist(err) {
+	if os.IsNotExist(err) || err != nil {
```
mazdakn, line 536: @winstonfly in these two branches, different `err` are returned. You cannot combine them like this.

mazdakn (changes requested):

cni-plugin/pkg/install/install.go
winstonfly: so, it should not be changed? @mazdakn mazdakn

mazdakn: This one should not change, since the calling function acts differently when the return error is either `nil` or not.

Issue #7420, opened 2023-03-03
the node local docker network is not wrok when the calico opened ebpf


xxoxx:
# the node local docker network is not wrok when the calico opened ebpf
## Current Behavior

nodes in the network cannot connect to containers port in the local docker

## Steps to Reproduce (for bugs)

1.install k8s （version:1.26.2）（1 master-node, 2 worker-node）by kubeadm
2. install calico：   kubectl apply -f https://projectcalico.docs.tigera.io/archive/v3.25.0/manifests/calico.yaml
3. open ebpf: calicoctl patch felixconfiguration default --patch='{"spec": {"bpfEnabled": true}}' && calicoctl patch felixconfiguration default --patch='{"spec": {"bpfExternalServiceMode": "DSR"}}'
4. master node run a docker : docker run -itd --name nginx1 -p 18080:80 nginx
5. other node:curl http://masterip:18080   (not connect)
6. when i close ebpf: calicoctl patch felixconfiguration default --patch='{"spec": {"bpfEnabled": false}}'
7. other node:curl http://masterip:18080   (works good)

I want to be able to connect the docker network normally in ebpf mode !
Please confirm that this is a bug or my configuration is not configured correctly

## Your Environment

* Calico version：3.25.0
* k8s:1.26.2
* Operating System and version: debian11 Linux debian-242 5.10.0-21-amd64

tomastigera:
How is the networking for docker done on the master node? Could be an issue with a policy on host endpoint. Could you provide some more info? Like devices, routing etc. on the master node?

xxoxx:
my docker is installed by default: apt-get install docker-ce docker-ce-cli containerd.io
docker version: 23.0.1

route:
"""
root@debian-241:~# route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.3.210   0.0.0.0         UG    0      0        0 enp1s0
10.244.17.0     0.0.0.0         255.255.255.192 U     0      0        0 *
10.244.99.0     192.168.3.242   255.255.255.192 UG    0      0        0 tunl0
10.244.160.192  192.168.3.243   255.255.255.192 UG    0      0        0 tunl0
172.17.0.0      0.0.0.0         255.255.0.0     U     0      0        0 docker0
192.168.3.0     0.0.0.0         255.255.255.0   U     0      0        0 enp1s0
"""

[...]

PR #7595, opened 2023-04-26
[BPF] Fix tracking packets that host forwards to devices not managed by calico

StevenTigera:
## Description

Adds an option to list devices from which we always let conntrack to track traffic so that if there are interfaces like docker0 and docker injects rules to forward traffic that goes to the host to docker, we do not discrupt NAT on returning traffic.

fixes https://github.com/projectcalico/calico/issues/7420

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
BPF mode: add felix configuration parameter to allow NOTRACK to be disabled for certain external interfaces interfaces.  Useful for workaround conflicts with non-Calico networked containers etc.
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -442,6 +442,10 @@ type FelixConfigurationSpec struct {
 	// BPFPolicyDebugEnabled when true, Felix records detailed information
 	// about the BPF policy programs, which can be examined with the calico-bpf command-line tool.
 	BPFPolicyDebugEnabled *bool `json:"bpfPolicyDebugEnabled,omitempty"`
+	// BPFForceTrackPacketsFromIfaces forces BPF to track packets from a list of specified interfaces.
+	// Essentially this excludes the specified list of interfaces from being NOTRACKed in IP Tables.
+	BPFForceTrackPacketsFromIfaces *[]string `json:"bpfForceTrackPacketsFromIfaces,omitempty"`
```
fasaxc, line 447: Think this field needs validation in libcalico-go to make sure that it is a good list of interface patterns.  That validation probably needs to be duplicated in felix config parsing.  I think we already have a regex for validating interface patterns.

felix/config/config_params.go
```diff
@@ -962,6 +963,8 @@ func loadParams() {
 			}
 		case "cidr-list":
 			param = &CIDRListParam{}
+		case "string-slice":
+			param = &StringSliceParam{}
```
fasaxc, line 967: As noted above, I think you need validation that the string values are valid interface patterns.  Perhaps add a validation regex to your new  `StringSliceParam` ?

felix/rules/static.go
```diff
@@ -1144,40 +1144,53 @@ func (r *DefaultRuleRenderer) StaticBPFModeRawChains(ipVersion uint8,
 		Rules: rawRules,
 	}
 
-	bpfUntrackedFlowChain := &Chain{
-		Name:  ChainRawUntrackedFlows,
-		Rules: []Rule{},
-	}
+	// BPF Untracked Flow Rules:
+	var bpfUntrackedFlowRules []Rule
+	var bpfUntrackedFlowChain *Chain
```
fasaxc, line 1149: Minor: as a general rule, it's best to minimise scope of variables and I don't think bpfUntrackedFlowChain is used until line 1191 so may as well declare it there.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -442,6 +442,10 @@ type FelixConfigurationSpec struct {
 	// BPFPolicyDebugEnabled when true, Felix records detailed information
 	// about the BPF policy programs, which can be examined with the calico-bpf command-line tool.
 	BPFPolicyDebugEnabled *bool `json:"bpfPolicyDebugEnabled,omitempty"`
+	// BPFForceTrackPacketsFromIfaces forces BPF to track packets from a list of specified interfaces.
+	// Essentially this excludes the specified list of interfaces from being NOTRACKed in IP Tables.
```
fasaxc, line 446:
```suggestion
	// BPFForceTrackPacketsFromIfaces in BPF mode, forces traffic from these interfaces 
	// to skip Calico's iptables NOTRACK rule, allowing traffic from those interfaces to be
	// tracked by Linux conntrack.  Should only be used for interfaces that are not used for
	// the Calico fabric.  For example, a docker bridge device for non-Calico-networked 
	// containers.
```
Suggest rewording to avoid implying that BPF is doing the tracking

felix/config/config_params.go
```diff
@@ -196,6 +196,7 @@ type Config struct {
 	BPFHostConntrackBypass             bool             `config:"bool;true"`
 	BPFEnforceRPF                      string           `config:"oneof(Disabled,Strict,Loose);Loose;non-zero"`
 	BPFPolicyDebugEnabled              bool             `config:"bool;true"`
+	BPFForceTrackPacketsFromIfaces     []string         `config:"string-slice;;die-on-fail"`
```
fasaxc, line 199: Should we make the default useful, perhaps `docker+`?  Do we need `die-on-fail`; think the default behaviour might be good enough for this param.

felix/rules/static.go
```diff
@@ -1144,40 +1144,53 @@ func (r *DefaultRuleRenderer) StaticBPFModeRawChains(ipVersion uint8,
...
-					Action:  ReturnAction{},
-					Comment: []string{"MarkSeenNATOutgoing Mark"},
-				},
-				Rule{
-					Action: NoTrackAction{},
-				},
-			},
+		// Iterate all BPF interfaces forced to track packets and append rule.
+		for _, interfaceName := range r.BPFForceTrackPacketsFromIfaces {
+			if len(interfaceName) > 0 {
```
fasaxc, line 1154: Should add a UT for the new feature

api/pkg/apis/projectcalico/v3/felixconfig.go
StevenTigera: Implemented as recommended; found ifaceFilterRegex which is preferred over interfaceRegex as this accommodates wildcards e.g. "docker+"

felix/config/config_params.go
StevenTigera: Implemented as suggested: make `docker+` the default and remove `die-on-fail`

felix/config/config_params.go
StevenTigera: Could I pls choose to not implement this suggestion because `StringSliceParam` was named added as a generic helper function to parse the raw input string and return as string slice i.e. no affinity to any input like valid interface(s) thus not subject to validation - the validation has been implemented above in the Felix config however which, if input there failed then, would not be need to be re-validated when set as config_params - thanks

api/pkg/apis/projectcalico/v3/felixconfig.go
tomastigera: Also it seems like we allow iptables style prefix like `docker+` so that needs to be documented here, saying, that you can use a specific device like `docker0` or a prefix like `docker+`. Note that we use regexps in most places. It should also state what the default is like `[Default: docker+]`, see the other options for examples.

felix/rules/static.go
StevenTigera: Updated as recommended

felix/rules/static.go
StevenTigera: Additional unit tests for BPF mode raw chains

libcalico-go/config/crd/crd.projectcalico.org_felixconfigurations.yaml
```diff
@@ -108,6 +108,13 @@ spec:
                   node appears to use the IP of the ingress node; this requires a
                   permissive L2 network.  [Default: Tunnel]'
                 type: string
+              bpfForceTrackPacketsFromIfaces:
+                description: BPFForceTrackPacketsFromIfaces forces BPF to track packets
```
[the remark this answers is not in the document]
StevenTigera: Thanks @tomastigera - you checked this too soon :)  I wanted to address all comments and regression test before git commit / push

api/pkg/apis/projectcalico/v3/felixconfig.go
StevenTigera: Thanks @tomastigera - I realized that `[Default: docker+]` was omitted - added now and re did `make generate` and confirm visible upon `kubectl explain felixconfig.spec | grep -7 bpfForceTrackPacketsFromIfaces`

felix/config/config_params.go
fasaxc: I do think you need validation here because Felix also supports env-var based configuration in manifest mode (and a config file, which is used in OpenStack).

I think the change here would be something like this:
```
case "interface-name-slice":
	param = &StringSliceParam{ValidationRegex: ...}
```
So your StringSliceParam could stay generic, but you'd have a particular sub-type with particular validation rules.

StevenTigera: Implemented string slice parameter validation as recommended

felix/config/config_params_test.go
```diff
@@ -504,6 +504,11 @@ var _ = DescribeTable("Config parsing",
 		map[string]time.Duration{"a": 0}, false),
 	// Not a required parameter so a bad value is translated to nil:
 	Entry("HealthTimeoutOverrides non-duration", "HealthTimeoutOverrides", "foo=bar", map[string]time.Duration(nil), false),
+
+	Entry("BPFForceTrackPacketsFromIfaces", "BPFForceTrackPacketsFromIfaces", "", []string{"docker+"}),
+	Entry("BPFForceTrackPacketsFromIfaces", "BPFForceTrackPacketsFromIfaces", "docker0", []string{"docker0"}),
+	Entry("BPFForceTrackPacketsFromIfaces", "BPFForceTrackPacketsFromIfaces", "docker+", []string{"docker+"}),
+	Entry("BPFForceTrackPacketsFromIfaces", "BPFForceTrackPacketsFromIfaces", "docker0,docker1", []string{"docker0", "docker1"}),
```
fasaxc, line 511: Please add an entry that fails validation.

fasaxc (changes requested):
One tiny request on the test side, otherwise good to go

felix/config/config_params_test.go
StevenTigera: Thanks @fasaxc I as per request I am trying to add an entry that fails validation something like
```Entry("BPFForceTrackPacketsFromIfaces", "BPFForceTrackPacketsFromIfaces", "cali@123", []string{"docker+"}, true),```
But in its current form the test will not pass because the config params `die-on-failure` was removed from `BPFForceTrackPacketsFromIfaces` as suggested prior.  I see when this is the case the expected Error will never be true as the err is reverted to nil and the updated parameter will revert to its default value i.e. `docker+` in this case.  Therefore, I can add the following test instead
```Entry("BPFForceTrackPacketsFromIfaces", "BPFForceTrackPacketsFromIfaces", "cali@123", []string{"docker+"}),```
 i.e. expected Error false but pls let me know if you would like the config params `die-on-failure` validation to be re-instated - thanks

fasaxc: SGTM

StevenTigera: Align config params test with params type test and include valid and invalid use cases

fasaxc (approved):
LGTM, probably worth squashing on merge?

StevenTigera:
> LGTM, probably worth squashing on merge?

Done

PR #7606, opened 2023-04-27
Stop interfering with non-Calico VXLAN packets

### LINKED ISSUE #6752: Calico unconditionally drops VXLAN packets that are not related to Calico
I set up a Kubernetes cluster using Calico with the following configuration.

* VXLAN mode: Always
* VXLAN port: 4789 (default)
* VXLAN ID (VNI): 4096 (default)

I also configured a VXLAN tunnel with the same port number but a different VXLAN ID.

Then, Calico drops all packets for the second VXLAN tunnel, even though I am using a different VXLAN ID for it.

## Expected Behavior

Calico does not interfere packets for the second VXLAN tunnel.

## Current Behavior

Calico drops all packets for the second VXLAN tunnel, since Calico adds the following iptables rule.

```
-A cali-INPUT -p udp -m comment --comment "cali:c3JpWekSTKWcdGH0" -m comment --comment "Drop IPv4 VXLAN packets from non-whitelisted hosts" -m multiport --dports 4789 -m addrtype --dst-type LOCAL -j DROP
```

This rule is generated here.
https://github.com/projectcalico/calico/blob/c041eec3ea3714c3a92b9f992f55af1e42a4842e/felix/rules/static.go#L243-L249

## Possible Solution

Revert https://github.com/projectcalico/calico/commit/cce544613169bd84a8a06f5536abd65cdc3c7ab8 if possible. The commit was introduced to fix errors that occur when the `u32` extension is used with `iptables-nft`.

According to this comment https://github.com/moby/moby/issues/43382#issuecomment-1143240624, `iptables-nft` also supports the `u32` extension, when  the `xt_u32` kernel module is loaded. This kernel module is available by Default in Ubuntu, and available in the `kernel-modules-extra` package of RHEL.

If revering the commit is not possible, providing it as an optional feature is another possible solution.

1. Set up Kubernetes with Calico as described above
2. Establish a VXLAN tunnel using the `ip` command at a worker node
3. Send PING packets from the other end of the VXLAN tunnel

## Your Environment

* Calico version: v3.24.1
* Orchestrator version: Kubernetes v1.25.1
* Operating System and version: Ubuntu 20.04.3 LTS

corhere:
## Description

Bug fix: reintroduce the iptables match clauses which restrict matches on VXLAN packets to those with a VNI related to Calico, reverting cce544613169bd84a8a06f5536abd65cdc3c7ab8. And add the same match clauses to the ip6tables versions of the rules for the first time. Calico will no longer interfere with other users of VXLAN on the same host.

Switch the implementation of the `VXLANVNI` match rule to use the `bpf` match which, unlike the `u32` match, is supported and installed by default on RHEL/CentOS 7, 8 and 9. And unlike the `u32` match expression which was only compatible with iptables, the cBPF socket filter program used with the `bpf` match rule is compatible with both iptables and ip6tables. The cBPF filter program is a straight port from https://github.com/moby/moby/pull/45308.

I have tested the filter program in isolation with both IPv4 and IPv6. I can port [the IPv4 unit test from moby/moby](https://github.com/moby/moby/blob/f117aef2ea63ee008c05a7506c8c9c50a1fa0c7f/libnetwork/drivers/overlay/bpf_linux_test.go) to Ginkgo if there is interest, with the caveat that it depends on `CAP_NET_RAW`. (Due to the use of cBPF extensions, the filter program cannot be tested with [user-space cBPF VMs.](https://pkg.go.dev/golang.org/x/net/bpf#VM))

## Related issues/PRs

- Fixes #6752

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/iptables/match_builder.go
+++ b/felix/iptables/match_builder.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/net/bpf"
 
 	"github.com/projectcalico/calico/felix/proto"
 )
@@ -310,16 +311,19 @@ func (m MatchCriteria) NotICMPV6TypeAndCode(t, c uint8) MatchCriteria {
 
 // VXLANVNI matches on the VNI contained within the VXLAN header.  It assumes that this is indeed a VXLAN
 // packet; i.e. it should be used with a protocol==UDP and port==VXLAN port match.
-//
-// Note: the -m u32 option is not supported on iptables in NFT mode.
-// https://wiki.nftables.org/wiki-nftables/index.php/Supported_features_compared_to_xtables#u32
 func (m MatchCriteria) VXLANVNI(vni uint32) MatchCriteria {
-	// This uses the U32 module, a simple VM for extracting bytes from a packet.  See
-	// http://www.stearns.org/doc/iptables-u32.current.html
-	return append(m, fmt.Sprintf(`-m u32 --u32 "`+
-		`0>>22&0x3C@` /* jump over the IP header */ +
-		`12>>8=0x%x` /* skip over 8 bytes of UDP header and 4 of VXLAN and compare 3 bytes with the expected VNI */ +
-		`"`, vni))
+	// This uses the bpf module, which matches with a cBPF (and on more recent kernels, eBPF)
+	// socket filter program. See https://ipset.netfilter.org/iptables-extensions.man.html#lbAH
+	prog := matchVXLANVNI(vni)
+
+	var b strings.Builder
+	b.WriteString(`-m bpf --bytecode "`)
+	fmt.Fprintf(&b, "%d", len(prog))
+	for _, ins := range prog {
+		fmt.Fprintf(&b, ",%d %d %d %d", ins.Op, ins.Jt, ins.Jf, ins.K)
+	}
+	b.WriteByte('"')
+	return append(m, b.String())
 }
 
 func PortsToMultiport(ports []uint16) string {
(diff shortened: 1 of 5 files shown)


felix/iptables/match_builder.go
```diff
@@ -343,3 +347,29 @@ func PortRangessToMultiport(ports []*proto.PortRange) string {
...
+		bpf.RetConstant{Val: 0},                                // ret #0
+		bpf.RetConstant{Val: ^uint32(0)},                       // match: ret #-1
+	})
+	// bpf.Assemble() only errors if an instruction is invalid. As the only variable
+	// part of the program is an instruction value for which the entire range is
+	// valid, whether the program can be successfully assembled is independent of
+	// the input. Given that the only recourse is to fix this function and
+	// recompile, there's little value in bubbling the error up to the caller.
+	if err != nil {
+		panic(err)
```
fasaxc, line 372:
```suggestion
		log.WithError(err).Panic("Failed to assemble static BPF VNI match program.")
```
Agree a panic is reasonable here.  Please panic via the logging library; it flushes the log on the way down.

felix/iptables/match_builder.go
```diff
@@ -343,3 +347,29 @@ func PortRangessToMultiport(ports []*proto.PortRange) string {
...
 }
+
+// vniMatchBPF returns a cBPF socket filter program which matches on the VXLAN
+// Network ID of encapsulated packets. The program assumes that it will only be
+// applied to VXLAN UDP datagrams. It is compatible with both IPv4 and IPv6.
+func matchVXLANVNI(vni uint32) []bpf.RawInstruction {
+	asm, err := bpf.Assemble([]bpf.Instruction{
+		// Load offset of UDP payload into X.
+		bpf.LoadExtension{Num: bpf.ExtPayloadOffset}, // ld poff
+		bpf.TAX{}, // tax
```
fasaxc, line 358:
```suggestion
		bpf.TAX{}, // Copy X into A
```
Wasn't familiar with this op so worth spelling it out.

corhere:
> Are you sure that the bpf match is supported by `iptables-nft`, it is still in the [list](https://wiki.nftables.org/wiki-nftables/index.php/Supported_features_compared_to_xtables#bpf) of "unsupported extensions"?

[`iptables-nft` supports all `xtables` extensions because it uses the xtables extensions directly.](https://developers.redhat.com/blog/2020/08/18/iptables-the-two-variants-and-their-relationship-with-nftables#using_iptables_nft) As I understand it, the list of unsupported extensions is a list of xtables extensions with no equivalent nftables extension. It is only relevant when writing rules using the native `nft` syntax. Not that any extension would be needed with native nftables: the `nft` [raw payload expression](https://www.mankier.com/8/nft#Payload_Expressions-Raw_Payload_Expression) `@th,96,24` (or, [since kernel 5.16](https://lore.kernel.org/netdev/YZZV42ERgpDbk%2FzL@salvia/t/), `@ih,32,24`) can be used to match the VNI field of a VXLAN datagram.

I suspect that what happened with https://github.com/projectcalico/felix/pull/2086 is that the failure to program the rule was misattributed to "NFT mode" when really the culprit was that the `xt_u32` kernel module was not installed on the test system. Red Hat moved that module into the `kernel-modules-extra` package in RHEL 8 which is not installed by default. RHEL 8 is also when the iptables backend was switched to `iptables-nft`, and combined with the confusing netfilter documentation it would have been easy to come to the wrong conclusion.

The only feature which would need to be detected is whether the `xt_bpf` kernel module is installed. It gets tricky to detect if it is built into the kernel rather than as a loadable module, though. The `xt_bpf` module has been available since Linux 3.9 and is installed by default on every distro I've checked. Are there any existing examples of Calico feature-detecting on the availability of kernel modules which I could take inspiration from?

caseydavenport:
How do we progress with this one?

tomastigera:
This would unfortunately not fly with nftables and ebpf dataplanre addresses this issue already.

PR #7607, opened 2023-04-27
[BPF] tcpdump-like debug expressions for tc programs

tomastigera:
## Description

    If BPFLogFilters is set only specified packets on specified devices
    trigger log actions. In any other case packets go through without the
    overhead of logging.
    
    A filter program is called by the preamble and that program decides
    whether packets should be processed by a set of programs with or without
    logging. This effectively means that there are two paths that a packet
    can take when logging is enabled.
    
    If BPFLogFilters is not specified, that filter program matches all
    packets and every program logs for every packet like the current
    firehose.
    
    As CTLB logging also goes to the same trace pipe, it mixes with tc logs
    and thus CLTB logging must be explicitly enabled if BPFLogFilters is
    specified by including 'ctlb=on' to avoid logging more than desired.
    CTLB is still a firehose.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
BPFLogFilters allow specifying fine grained log filters that reduce the amount of logs emitted when debugging.
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -700,32 +707,49 @@ func (m *bpfEndpointManager) updateIfaceStateMap(name string, iface *bpfInterfac
...
 			log.WithError(err).Warn("Policy program may leak.")
 		}
-		if err := m.policyMapAlloc.Put(iface.dpState.policyIdx[hook.Ingress]); err != nil {
+		if err := m.jumpMapAlloc.Put(iface.dpState.policyIdx[hook.Egress]); err != nil {
 			log.WithError(err).Error("Ingress")
 		}
 
-		if err := m.policyMapDelete(hook.Egress, iface.dpState.policyIdx[hook.Egress]); err != nil {
+		if err := m.jumpMapDelete(hook.Egress, iface.dpState.filterIdx[hook.Egress]); err != nil {
 			log.WithError(err).Warn("Policy program may leak.")
```
fasaxc, line 738:
```suggestion
			log.WithError(err).Warn("Filter program may leak.")
```

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -700,32 +707,49 @@ func (m *bpfEndpointManager) updateIfaceStateMap(name string, iface *bpfInterfac
...
 		}
-		if err := m.policyMapAlloc.Put(iface.dpState.policyIdx[hook.Egress]); err != nil {
+		if err := m.jumpMapAlloc.Put(iface.dpState.filterIdx[hook.Egress]); err != nil {
+			log.WithError(err).Error("Ingress")
+		}
+
+		if err := m.jumpMapDelete(hook.Ingress, iface.dpState.filterIdx[hook.Ingress]); err != nil {
+			log.WithError(err).Warn("Policy program may leak.")
```
fasaxc, line 745:
```suggestion
			log.WithError(err).Warn("Filter program may leak.")
```

fasaxc (commented):
LGTM at high level:
* Follows pattern for handling maps form previous PR.
* Couldn't see any map/program leaks.
* Activated in tests.

One thought on the datamodel.  I got pushback from Casey on using comma-delimited strings in FelixConfig and I think he was right so would be good to avoid adding any more

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -364,6 +364,15 @@ type FelixConfigurationSpec struct {
...
 	BPFLogLevel string `json:"bpfLogLevel" validate:"omitempty,bpfLogLevel"`
+	// BPFLogFilters is a comma-separated list of key=values where the value is
+	// a pcap filter expression and the key is an interface name with 'all'
+	// denoting all interfaces, 'weps' all workload endpoints and 'heps' all host
+	// endpoints. 'ctlb' can take a value of 'on' or 'off' to set whether
+	// connect time load balance debug output is enabled when BPFLogLevel is
+	// debug. It is off by default.
+	// [Default: unset - means all debug logs are emitted]
+	// +optional
+	BPFLogFilters string `json:"bpfLogFilters,omitempty" validate:"omitempty,keyValueList"`
```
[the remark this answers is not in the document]
tomastigera: @caseydavenport can I have :eyes: on the last two commits that make it a map[string]string :pray:

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -364,6 +364,15 @@ type FelixConfigurationSpec struct {
 	// [Default: Off].
 	// +optional
 	BPFLogLevel string `json:"bpfLogLevel" validate:"omitempty,bpfLogLevel"`
+	// BPFLogFilters is a comma-separated list of key=values where the value is
```
caseydavenport, line 367: The comment here needs to be updated to match the new `map[string]string` format.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -364,6 +364,15 @@ type FelixConfigurationSpec struct {
...
+	// a pcap filter expression and the key is an interface name with 'all'
+	// denoting all interfaces, 'weps' all workload endpoints and 'heps' all host
+	// endpoints. 'ctlb' can take a value of 'on' or 'off' to set whether
+	// connect time load balance debug output is enabled when BPFLogLevel is
+	// debug. It is off by default.
+	// [Default: unset - means all debug logs are emitted]
+	// +optional
+	BPFLogFilters *map[string]string `json:"bpfLogFilters,omitempty" validate:"omitempty,bpfLogFilters"`
```
caseydavenport, line 375: I don't think this needs to be a pointer - a map can natively be `nil`.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -364,6 +364,15 @@ type FelixConfigurationSpec struct {
 	// [Default: Off].
 	// +optional
 	BPFLogLevel string `json:"bpfLogLevel" validate:"omitempty,bpfLogLevel"`
+	// BPFLogFilters is a comma-separated list of key=values where the value is
+	// a pcap filter expression and the key is an interface name with 'all'
+	// denoting all interfaces, 'weps' all workload endpoints and 'heps' all host
+	// endpoints. 'ctlb' can take a value of 'on' or 'off' to set whether
```
caseydavenport, line 370: It's not clear to me how `ctlb` fits in here - the filters map is a map of interface name to filter string, right? `ctlb` isn't an interface name, so it's a bit out of place to wedge that in here (at least by my naive understanding)

tomastigera: ok, will add another option as we would like to expand it later from on/off to a true filter, which is not going to be a pcap filter

caseydavenport (approved):

tomastigera:
errors in CI are unrelated

PR #7622, opened 2023-05-03
OpenStack packaging updates & Jammy support

danudey:
## Description

A revamp of the existing packaging support for OpenStack

* Move from sequential docker builds to `docker buildx bake` which can build multiple images in parallel
* Add support for Ubuntu Jammy
* Some shell cleanups
* Remove unnecessary extra steps for building CentOS 7 ppc64le build image
* Improved output for easier debugging

## Release Note

```release-note
Add support for OpenStack on Ubuntu Jammy
```

### DIFF
--- a/hack/release/packaging/docker-build-images/docker-bake.hcl
+++ b/hack/release/packaging/docker-build-images/docker-bake.hcl
@@ -1,0 +1,66 @@
+# docker-bake.hcl
+
+# To use a different arch, define it in an environment variable;
+# for example, `ARCH=ppc64le docker buildx bake`. 
+variable "ARCH" {
+    default = "amd64"
+}
+
+variable "UID" {
+    default = 1000
+}
+
+variable "GID" {
+    default = 1000
+}
+
+# Define groups for the builds we want to be able to do
+
+# This is the default rule if you don't specify one. It'll build
+# everything.
+
+group "default" {
+    targets = ["ubuntu", "centos"]
+}
+
+# All ubuntu images
+group "ubuntu" {
+  targets = ["trusty", "xenial", "bionic", "focal", "jammy"]
+}
+
+# All centos images
+group "centos" {
+    targets = ["centos7"]
+}
+
+# Ubuntu builds
+target "trusty" {
+  dockerfile = "ubuntu-trusty-build.Dockerfile.${ARCH}"
+  tags = ["calico-build/trusty"]
+}
+target "xenial" {
+  dockerfile = "ubuntu-xenial-build.Dockerfile.${ARCH}"
+  tags = ["calico-build/xenial"]
+}
+target "bionic" {
+  dockerfile = "ubuntu-bionic-build.Dockerfile.${ARCH}"
+  tags = ["calico-build/bionic"]
+}
+target "focal" {
+  dockerfile = "ubuntu-focal-build.Dockerfile.${ARCH}"
+  tags = ["calico-build/focal"]
+}
+target "jammy" {
+  dockerfile = "ubuntu-jammy-build.Dockerfile.${ARCH}"
+  tags = ["calico-build/jammy"]
+}
+
+# CentOS builds
+target "centos7" {
+  dockerfile = "centos7-build.Dockerfile.${ARCH}"
+  args = {
+    UID = UID
+    GID = GID
+  }
+  tags = ["calico-build/centos7"]
+}
--- a/hack/release/packaging/docker-build-images/ubuntu-jammy-build.Dockerfile.amd64
+++ b/hack/release/packaging/docker-build-images/ubuntu-jammy-build.Dockerfile.amd64
@@ -1,6 +1,6 @@
-FROM ppc64le/ubuntu:bionic
+FROM ubuntu:jammy
 MAINTAINER Daniel Fox <dan.fox@tigera.io>
-ENV STREAM bionic
+ENV STREAM focal
 
 ADD install-ubuntu-build-deps install-ubuntu-build-deps
 RUN ./install-ubuntu-build-deps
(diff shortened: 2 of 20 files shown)


hack/release/packaging/docker-build-images/docker-bake.hcl
nelljerram, line 1: This file does not seem to be explicitly referenced anywhere else in this PR.  Is it automatically picked up by `docker buildx bake`?

hack/release/packaging/utils/create-update-packages.sh
```diff
@@ -82,22 +95,28 @@ function require_repo_name {
 function require_rpm_host_vars {
     # HOST and GCLOUD_ARGS must be set to indicate the RPM host, and a
     # gcloud identity that permits logging into that host.
-    test -n "$GCLOUD_ARGS"
-    echo GCLOUD_ARGS is "$GCLOUD_ARGS"
-    test -n "$HOST"
-    echo HOST is $HOST
+    if [[ -z  "$GCLOUD_ARGS" ]]; then
```
nelljerram, line 98: nit: I believe you could just use single brackets here.  (+ same just below)

hack/release/packaging/utils/create-update-packages.sh
```diff
@@ -160,26 +180,14 @@ function precheck_pub_rpms {
...
 
 function do_bld_images {
     # Build the docker images that we use for building for each target platform.
     pushd ${rootdir}/hack/release/packaging/docker-build-images
-    docker build -f ubuntu-trusty-build.Dockerfile.${ARCH} -t calico-build/trusty .
-    docker build -f ubuntu-xenial-build.Dockerfile.${ARCH} -t calico-build/xenial .
-    docker build -f ubuntu-bionic-build.Dockerfile.${ARCH} -t calico-build/bionic .
-    docker build -f ubuntu-focal-build.Dockerfile.${ARCH} -t calico-build/focal .
-    docker build --build-arg=UID=`id -u` --build-arg=GID=`id -g` -f centos7-build.Dockerfile.${ARCH} -t calico-build/centos7 .
+    docker buildx bake
```
nelljerram, line 189: I think we've lost the passing in of `id -u` and `-id -g` in the centos7 case.  Might now be passing by accident because the current user will often be UID 1000.

hack/release/packaging/utils/make-packages.sh
```diff
@@ -49,32 +49,25 @@ for package_type in "$@"; do
...
+  ${changelog_message}
 
+ -- Daniel Fox <dan.fox@tigera.io>  ${timestamp}
 EOF
 		} > debian/changelog
 
 		excludes="${DPKG_EXCL:--I}"
 
-		${DOCKER_RUN_RM} calico-build/${series} dpkg-buildpackage ${excludes} -S -d
+		${DOCKER_RUN_RM} calico-build/${series} dpkg-buildpackage ${excludes} -S -d | ts "[build $series]"
```
nelljerram, line 70: If I'm understanding this and the manpage correctly, this won't actually add a timestamp to each line of output.  Is that what you intended?

hack/release/packaging/utils/publish-debs.sh
```diff
@@ -38,6 +38,6 @@ for series in trusty xenial bionic focal; do
                 break
             fi
         done
-        ${already_exists} || docker run --rm -v ${rootdir}:/code -w /code/hack/release/packaging/output calico-build/${series} dput -u ppa:project-calico/${REPO_NAME} ${changes_file}
+        ${already_exists} || docker run --rm -v ${rootdir}:/code -w /code/hack/release/packaging/output calico-build/${series} dput -u ppa:project-calico/${REPO_NAME} ${changes_file} | ts "[upload $series]"
```
nelljerram, line 41: Ditto.

nelljerram (changes requested):
Nice work; I've mentioned some minor tweaks and queries.

hack/release/packaging/docker-build-images/docker-bake.hcl
danudey: Yep! Per the docs at https://docs.docker.com/build/bake/file-definition/ `docker buildx bake` will search for a list of files, with hcl files being the preferred file format

hack/release/packaging/docker-build-images/ubuntu-jammy-build.Dockerfile.amd64
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:jammy
+MAINTAINER Daniel Fox <dan.fox@tigera.io>
+ENV STREAM focal
```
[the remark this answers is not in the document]
danudey: Good catch, thanks! Fixing.

hack/release/packaging/utils/create-update-packages.sh
danudey: That's true in this case, but I'm in the habit of using `[[ .. ]]` instead of `[ .. ]` because of the various syntax improvements it provides and the fewer gotchas that can happen later on.

(http://mywiki.wooledge.org/BashFAQ/031 for reference if anyone is curious)

hack/release/packaging/utils/create-update-packages.sh
danudey: This is true, and I'll resolve this.

The centos7 case is defined in the `docker-bake.hcl` file. In that file we define a default UID/GID of 1000, which can be overridden by environment variables, but we don't override it here. I've pushed a fix for this.

hack/release/packaging/utils/lib.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
danudey: The purpose for adding the shebang line is to be able to parse the file correctly with `shellcheck`. I haven't done a full linting pass on these files, but I intend to do so in the future, so that's why this line was added. If you think it's a better idea to leave it unchanged though, I'll revert this.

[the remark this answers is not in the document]
nelljerram: I don't want to impose my view here, but I do think it's wrong for tools to drive an incorrect agenda.  Perhaps there's another way to tell shellcheck that this is a file that it should check?

nelljerram (approved):
LGTM.  I'm happy to leave the shellcheck point to your judgement.

PR #7625, opened 2023-05-03
Update OpenStack dependencies to Yoga

nelljerram:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/.semaphore/semaphore-scheduled-builds.yml
+++ b/.semaphore/semaphore-scheduled-builds.yml
@@ -682,19 +682,14 @@ blocks:
       - name: 'Unit and FV tests (tox)'
         commands:
           - ../.semaphore/run-and-monitor tox.log make tox
-      - name: 'Mainline ST (DevStack + Tempest) on Ussuri'
+      - name: 'Mainline ST (DevStack + Tempest) on Yoga'
         commands:
           - git checkout -b devstack-test
           - export LIBVIRT_TYPE=qemu
-          - export UPPER_CONSTRAINTS_FILE=https://releases.openstack.org/constraints/upper/ussuri
-          # Use proposed fix at
-          # https://review.opendev.org/c/openstack/requirements/+/810859.  See commit
-          # message for more context.
-          - export REQUIREMENTS_REPO=https://review.opendev.org/openstack/requirements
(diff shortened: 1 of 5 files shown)


.semaphore/semaphore-scheduled-builds.yml
```diff
@@ -702,6 +702,40 @@ blocks:
...
+    prologue:
+      commands:
+      - cd networking-calico
+    jobs:
+      - name: 'Unit and FV tests (tox) on Yoga'
+        commands:
+          - ../.semaphore/run-and-monitor tox.log make tox-yoga
+      - name: 'Mainline ST (DevStack + Tempest) on Yoga'
+        commands:
+          - dpkg -l "python*" || true
```
Josh-Tigera, line 723: What's the purpose of this command? I can't see a `-l` flag specified in the dpkg man page.

.semaphore/semaphore-scheduled-builds.yml
```diff
@@ -702,6 +702,40 @@ blocks:
...
+          - sudo find / -name "*wrapt*" || true
```
Josh-Tigera, line 724: Is this just for debugging? Looks like we remove the python3-wrapt package unconditionally.

.semaphore/semaphore-scheduled-builds.yml
nelljerram: It means to list the installed packages matching "python*".  I am surprised it's not in the man page!

I can take this line out now.  It was debugging for a problem with the "python3-wrapt" package.  It was pre-installed on the Semaphore node with a different version from what OpenStack needs, and that was causing the DevStack setup to fail.  The fix (2 lines below) is to uninstall the existing version first.

.semaphore/semaphore-scheduled-builds.yml
nelljerram: This, it was just for debugging, in case the previous line didn't find any packages, but wrapt was pre-installed in some different way.

Josh-Tigera (approved):

nelljerram:
Docs PR is at https://github.com/tigera/docs/pull/633

PR #7629, opened 2023-05-04
Fips changes

sridhartigera:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


lib.Makefile
```diff
@@ -203,6 +203,48 @@ define build_cgo_boring_binary
             && go tool nm $(2) | grep '_Cfunc__goboringcrypto_' 1> /dev/null'
 endef
 
+# Build a static binary with boring crypto support.
```
hjiawei, line 206: If we don't want FIPS, we probably don't want boringssl/cgo either. We can use the pure-go TLS implementation. I believe the function `build_binary` defined a few lines below is doing so. Why do we want the two new definitions with boring/cgo but no FIPS?

apiserver/Makefile
```diff
@@ -82,7 +93,11 @@ cd: image-all cd-common
 #########################################################################
 $(BINDIR)/apiserver-$(ARCH): $(K8SAPISERVER_GO_FILES)
 ifeq ($(ARCH),amd64)
+ifeq ($(FIPS), true)
 	$(call build_cgo_boring_binary, $(PACKAGE_NAME)/cmd/apiserver, $@)
+else
+	$(call build_cgo_boring_binary_no_fips, $(PACKAGE_NAME)/cmd/apiserver, $@)
```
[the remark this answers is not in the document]
sridhartigera: I too thought the same. If we are to use ```build_binary```, should wrap ```check_boring_ssl``` under ```ifeq ($FIPS, true)```. WDYT @rene-dekker ?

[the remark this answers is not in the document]
rene-dekker: I was thinking that we could always build using GOEXPERIMENT=boring, but that we set the fipsstrict tag only when fips is enabled. This minimizes the difference between our fips and non-fips builds, but allows for tls1.3 in non-fips builds.
Maybe the `ifeq ($(FIPS), true)` statement can add the tag to a FIPS_STRICT_TAG variable and we can have the `build_cgo_boring_binary` and `build_static_cgo_boring_binary` add that tag to the build command.

node/Makefile
```diff
@@ -231,12 +230,27 @@ CGO_LDFLAGS=""
 CGO_CFLAGS=""
 endif
 
+ifeq ($(FIPS), true)
```
mgleung, line 233: Is this clause a little low down in the file for this to work? I figured that it would need to be set before it is first called in the Makefile in order for it to guarantee that there's a definition?

pod2daemon/Makefile
```diff
@@ -30,6 +30,7 @@ clean:
 	rm -rf bin/flexvol-$(ARCH)
 	rm -rf bin/$(CSI_IMAGE)-$(ARCH)
 	rm -rf bin/$(REGISTRAR_IMAGE)-$(ARCH)
+	rm -rf bin
```
mgleung, line 33: Do we need to do the equivalent to easily clean out all of the FIPS images in the other Makefiles too?

pod2daemon/Makefile
```diff
@@ -42,50 +43,72 @@ endif
 ###############################################################################
 # Building the binary
 ###############################################################################
+FIPS ?= false
+
+FLEXVOL_CONTAINER_MARKER=$(FLEXVOL_CONTAINER_CREATED)
```
mgleung, line 48: Do we need a separate FIPS flag for the flexvol image creation too?

pod2daemon/csidriver/Dockerfile.amd64
```diff
@@ -38,6 +38,7 @@ COPY --from=ubi /lib64/libc.so.6 /lib64/libc.so.6
 COPY --from=ubi /etc/host.conf /etc/host.conf
 COPY --from=ubi /etc/nsswitch.conf /etc/nsswitch.conf
 
-ADD bin/csi-driver-amd64 /usr/local/bin/csi-driver
+ARG BIN_DIR
+ADD ${BIN_DIR}/csi-driver-amd64 /usr/local/bin/csi-driver
```
mgleung, line 42: do we need to repeat this step for the flexvol image too?

mgleung: flexvol does not use SSL libs so making FIPS versions of the image is unnecessary

mgleung (approved):
LGTM

PR #7635, opened 2023-05-09
Move UID/GID assignment to command-line

danudey:
Specifying a read-only variable in front of a command works in some shells' interactive sessions, but not in bash. Move the variable assignment to command-line parameters for docker buildx build instead.


mgleung (approved):
LGTM

PR #7640, opened 2023-05-11
Create a symlink for libpcap.so.0.8

sridhartigera:
## Description

BPF log filters use gopacket/pcap which requires libpcap.so.0.8. Hence create a symlink for libpcap.so.0.8 pointing to libpcap.so.1

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/Dockerfile.amd64
+++ b/node/Dockerfile.amd64
@@ -223,8 +223,6 @@ RUN chmod u+s /bin/mountns
 ADD clean-up-filesystem.sh /
 RUN /clean-up-filesystem.sh
 
-RUN ln -s /usr/lib64/libpcap.so.1 /usr/lib64/libpcap.so.0.8
-
 # Copy everything into a fresh scratch image so that naive CVE scanners don't pick up binaries and libraries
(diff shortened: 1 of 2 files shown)


hjiawei:
Do we know if `libpcap.so.0.8` is binary compatible with `libpcap.so.1`? It is fine if they are binary compatible but can cause hard to debug issues when they are not.

sridhartigera:
> Do we know if features we use in `libpcap.so.0.8` is binary compatible with `libpcap.so.1`? It is fine if they are but this symlink might cause hard to debug issues when they are not.

I am not sure how to check if its binary compatible. Any idea?

hjiawei:
From the `.so` file version info (0.8 vs 1), they are likely not compatible. I think a better fix for this issue is to build/link calico node using a same or compatible base as we ship (UBI 8). It is very likely that the go-build base (debian buster, oldstable) is outdated.

PR #7657, opened 2023-05-15
allow reference to existing secret as imagePullSecret

ozdanborne:
## Description

new feature - allows reference to existing secrets for imagePullSecrets without passing the secret itself. this enables management of secrets by an external system like sealedsecrets and prevents the secret data from being stored in helm.

It works by allowing use of the installation's imagePullSecret field directly, whereas previously only the toplevel imagePullSecrets field was used.

also added basic unit tests for the feature via helm terratest which makes it easier to render the chart given certain values and assert based off the resulting k8s resources.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
allows reference to existing secrets for imagePullSecrets without passing the secret itself to helm. this enables management of secrets by an external system like sealedsecrets and prevents the secret data from being stored in helm.

It works by allowing use of the installation's imagePullSecret field directly, whereas previously only the toplevel imagePullSecrets field was used.
```

## Reminder for the reviewer

Other optional labels:


manifests/ocp/02-serviceaccount-tigera-operator.yaml
```diff
@@ -3,5 +3,5 @@ kind: ServiceAccount
 metadata:
   name: tigera-operator
   namespace: tigera-operator
-imagePullSecrets:
+imagePullSecrets:
```
caseydavenport, line 6: We should avoid the extra space at the end of this line as it fails yaml linting, which some users run on our manifests.

charts/tigera-operator/values.yaml
```diff
@@ -1,8 +1,12 @@
+# imagePullSecrets specified here will result in helm attempting to create a secret resource.
```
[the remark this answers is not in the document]
ozdanborne: Updated to better explain the fields and their format. But to be clear - you say "they now differ pretty substantially" but this PR has not changed their format. Previously, you needed to use the toplevel field with helm and avoid using the field directly nested in the installation resource field. Any specification of the nested field would be discarded. Now, you can use both, and behavior for the nested field is now defined. I'll update the PR description to cover this more clearly as well.

charts/tigera-operator/templates/_helpers.tpl
```diff
@@ -5,3 +5,16 @@
...
 {{- .image -}}:{{- .version -}}
 {{- end -}}
+
+{{/*
+generate imagePullSecrets for installation and deployments
+by combining installation.imagePullSecrets with toplevel imagePullSecrets.
+*/}}
+
+{{- define "tigera-operator.imagePullSecrets" -}}
+{{- $secrets := default list .Values.installation.imagePullSecrets -}}
```
caseydavenport, line 15: I suppose if a secret is defined in both places we'll get an error back from k8s? 

This is probably OK, although I can definitely imagine users trying to put a secret in both places.

charts/tigera-operator/values.yaml
```diff
@@ -1,8 +1,20 @@
+# imagePullSecrets are a special helm field which, when specified, creates a secret
```
caseydavenport, line 1: "imagePullSecrets _are_" doesn't sound right to me (despite it being plural).

charts/tigera-operator/values.yaml
```diff
@@ -1,8 +1,20 @@
...
+# containing the pull secret and configures operator's serviceaccount to use it to pull the operator image
```
caseydavenport, line 2: Maybe just "all images deployed by this helm chart and the resulting operator" is a more concise way of saying this?

charts/test/tigera_operator_chart_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Tigera Operator Helm Chart", func() {
```
caseydavenport, line 17: You've already written it, so I won't make you change it, but I think we've said all new test suites should be written in standard `go test` without ginkgo / gomega, in case you're interested in tweaking these.

caseydavenport (approved):

PR #7668, opened 2023-05-17
Call Router() to enact side-effects necessary to work on Yoga

Josh-Tigera:
Yoga changes to some functions we inherit from the `l3_db.L3_NAT_db_mixin` class we use in our `CalicoPlugin` class require some extended attributes that are not present by default and cause tracebacks in the neutron-server log eg. `AttributeError: type object 'Port' has no attribute 'port_forwardings'`. Calling `Router()` triggers the side-effects necessary to pull in those extended attributes and eliminate the tracebacks.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/plugins/calico/plugin.py
+++ b/networking-calico/networking_calico/plugins/calico/plugin.py
@@ -13,6 +13,7 @@
 #    under the License.
 
 from neutron.db import l3_db
+from neutron.db.models import l3
 from neutron.plugins.ml2.plugin import Ml2Plugin
 
 from networking_calico.compat import cfg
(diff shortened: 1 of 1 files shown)


nelljerram (approved):
LGTM.  If you still have one around, can you just paste in an example of the tracebacks that we get in the neutron-server.log, and add a couple more notes to clarify that:

1. If those tracebacks occur, they do impact subsequent operation, e.g. following tests fail, so we presume they cause the Neutron DB to be left in an invalid state.
2. We haven't fully understood the chain of detail that causes those tracebacks, or of why the `l3.Router` addition fixes them.  (But empirically, it does.)

I'm thinking it's probably enough to note those things in a PR comment or the commit message, rather than in a code comment - but defer to whatever you think will work best; it just needs to be somehow discoverable.

Josh-Tigera:
Thanks @nelljerram I've added more detail to the comment within the code itself. I'm a fan of co-locating details like this with the code it directly relates to so that there's less of an onus on the next person to perform forensic analysis via PRs and git blame to understand what's going on.

nelljerram:
@Josh-Tigera FYI I closed and reopened this to get round an apparent glitch in the license/cla checking.

PR #7669, opened 2023-05-17
Fix Felix packaging for Ubuntu

danudey:
## Description

Fix openstack packaging by adding `libpcap-dev` as a build dependency, and update the `debian/control` file to include the dpkg-calculated shared library dependencies.


nelljerram (approved):
LGTM, thanks @danudey !

PR #7687, opened 2023-05-23
Update issue templates

caseydavenport:
Moving and editing the issue template so that it's recognized by the GitHub UI


caseydavenport:
@mgleung for review!

We might want to play with auto assigning labels / assignees as well - GH now lets you have multiple issue templates to choose from when creating an issue, so users can self-triage as bug / enhancement / support

Issue #4736, opened 2021-07-11
Add eBPF support for IPv6


telmich:
## Expected Behavior

* calico support eBPF mode for IPv6 only clusters

## Current Behavior

* it is not supported

## Context

* Cannot use kube-proxy free environments in IPv6 only custers

## Your Environment
* Calico version: v3.19.0
* Orchestrator version (e.g. kubernetes, mesos, rkt): kuberentes
* Operating System and version: Alpine Linux 3.14
* Link to your project (optional): www.datacenterlight.ch

lwr20:
@fasaxc FYI

mazdakn:
@svallala we have started the development, but cannot provide a timeline at the moment.

MTRNord:
Hi any updates for this in 2023? :) (A "no update" is still an update I would appreciate. :))

tomastigera:
We did not get much further due to other business priorities, and also other dev priorites as we needed to restructure the dp to be able to add some features and also ipv6 in a simpler way. We had to take a step back. That should be ready for 3.27 and my goal is to resume ipv6 as the very next thing :crossed_fingers:

MTRNord:
> We did not get much further due to other business priorities, and also other dev priorites as we needed to restructure the dp to be able to add some features and also ipv6 in a simpler way. We had to take a step back. That should be ready for 3.27 and my goal is to resume ipv6 as the very next thing crossed_fingers

thanks a lot for this update :)

telmich:
@tomastigera in case there is something to support you with such as an ipv6 only k8s cluster, let me know. 

Being able to use eBPF and thus allow source IP preservation would be highly appreciated in the ipv6 context, as services can be directly accessed from the Internet. And for many services, being able receive the original IP address is crucial.

davidwadddell:
hi guys appreciate the work on calico.  wondered if support for this was any closer. source IP preservation is a desirable feature for us but is blocked by lack of ip6  support with eBPF; support is present in Cilium for this at the moment, but we'd prefer not to switch,

tomastigera:
We are actually making a good progress and we started passing fv tests with IPv6 code. With the summer vacations in full swing, my personal estimate is that we may have a beta of IPv6 released around end of September :crossed_fingers:

davidwadddell:
Thanks for the update @tomastigera

tomastigera:
[...]

PR #7701, opened 2023-05-25
[BPF] support for ipv4 options

tomastigera:
## Description

Although IPv4 options are not widely used, parsing packets with flexible offsets is a prerequisite for IPv6 processing. The code restructuring needed for IPv4 options will be used for IPv6 processing.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: handles IPv4 packets with IP options. XDP only does so in kernels 5.18+ it drops them in older kernels.
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/tc.c
```diff
@@ -557,622 +569,843 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
...
+
+deny:
+	{
+		struct fwd fwd = {
+			.res = TC_ACT_SHOT,
+			.reason = ctx->fwd.reason,
+		};
+		return fwd;
+	}
+}
```
tomastigera, line 1058: This essentially splits the original code in do_nat() and post_nat() as we want to reuse the same code in two programs.

felix/bpf-gpl/list-ut-objs
```diff
@@ -11,7 +11,7 @@
 
 emit_filename() {
   echo "bin/test_${from_or_to}_${ep_type}_fib_${log_level}${dsr}.o"
-  echo "bin/test_${from_or_to}_${ep_type}_fib_${log_level}${dsr}_v6.o"
+#  echo "bin/test_${from_or_to}_${ep_type}_fib_${log_level}${dsr}_v6.o"
```
fasaxc, line 14: WIP?

felix/bpf-gpl/xdp.c
```diff
@@ -61,6 +61,7 @@ int calico_xdp_main(struct xdp_md *xdp)
 		return XDP_DROP;
 	}
 	__builtin_memset(ctx->state, 0, sizeof(*ctx->state));
+	ctx->scratch = (void *)(ctx->xdp_globals + 1);
```
fasaxc, line 64: What's going on here?

fasaxc (commented):
Looked at everything apart from the extensive changes to tc.c.  Could do with some pointers on what the high-level changes are there.

felix/bpf-gpl/xdp.c
tomastigera: although not used in XDP we need to have the pointer set, so pointing behind the globals as there is enough space.

tomastigera:
Rebased and pushed to make sure we use the latest code even though is would still merge cleanly.

fasaxc (approved):

PR #7707, opened 2023-05-26
[BPF] clasiffy correctly forwarded traffic

tomastigera:
Previously, anything that had source the same as the host was considered as "from host" by the policy engine and global policies with applyOnForward=false would apply to such traffic. However, for instance SNATed traffic due to nat-outgoing would be also considered "from host" eventhough it originated in a workload.

This fixes it by first looking at a SEEN mark which indicates that the packet was seen by another endpoint and thus is forwarded with the exception of packets being routed via the bpf nat iface which is used by the host processes.

fixes https://github.com/projectcalico/calico/issues/7717

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: fix applyOnforward=false in global policies
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -239,6 +239,19 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 {
 	CALI_DEBUG("conntrack entry flags 0x%x\n", ctx->state->ct_result.flags);
 
+	/* We are forwarding a packet if it has a seen mark (that is another
+	 * program has seen it already) and is either not routed through the
+	 * bpfnat iface (which may be true for host traffic) or has the specific
+	 * reasons set.
+	 */
+	bool forwarding = CALI_F_EGRESS &&
+		skb_mark_equals(ctx->skb, CALI_SKB_MARK_SEEN_MASK, CALI_SKB_MARK_SEEN) &&
+		(!skb_mark_equals(ctx->skb, CALI_SKB_MARK_FROM_NAT_IFACE_OUT, CALI_SKB_MARK_FROM_NAT_IFACE_OUT) ||
+		 (skb_mark_equals(ctx->skb, CALI_SKB_MARK_BYPASS_MASK, CALI_SKB_MARK_FALLTHROUGH) ||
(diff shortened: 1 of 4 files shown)


felix/bpf-gpl/tc.c
```diff
@@ -239,6 +239,19 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 {
 	CALI_DEBUG("conntrack entry flags 0x%x\n", ctx->state->ct_result.flags);
 
+	/* We are forwarding a packet if it has a seen mark (that is another
+	 * program has seen it already) and is either not routed through the
+	 * bpfnat iface (which may be true for host traffic) or has the specific
+	 * reasons set.
+	 */
+	bool forwarding = CALI_F_EGRESS &&
```
mazdakn, line 247: Just wondering why not defining this when it's used. Does the packet mark changes from here to the place it's used?

felix/fv/bpf_test.go
```diff
@@ -1124,30 +1124,90 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+							pol.Spec.Selector = "node=='" + felixes[0].Name + "'"
+							pol.Spec.ApplyOnForward = true
+
+							pol = createPolicy(pol)
+						})
+
+						bpfWaitForPolicy(felixes[0], "eth0", "egress", "default.host-0-1")
+
+					})
+					It("should handle NAT outgoing", func() {
```
mazdakn, line 1186: We probably need another test with only `applyOnForward` with drop action, to make sure it's actually hit and dropped. WDYT?

mazdakn (commented):
LGTM. Just left some comments.

felix/bpf-gpl/tc.c
tomastigera: my reasoning was that we do it before we start mangling the marks ourselves, so the sooner the better, but yeah sounds like a lame reason :sweat_smile: :grimacing:

felix/fv/bpf_test.go
tomastigera: The fact that we allow some traffic implicitly carries a default deny rule. I would let the all the other case to be checked by the thorough e2e test, which were not enabled for bpf

mazdakn (approved):
LGTM.

PR #7550, opened 2023-04-12
Skip setting node IP annotation if CALICO_NETWORKING_BACKEND=none

skmatti:
## Description

`projectcalico.org/IPv4Address` and `projectcalico.org/IPv4IPIPTunnelAddr` annotations are used for BGP so disabling these in `policy` only mode

If we remove this dependency, calico-node in policy only mode does not need k8s node update/ patch permissions which addresses a vulnerability associated with node agent's over provisioned permissions.

## Release Note

```release-note
When running Calico in policy-only mode, do not write the IP annotations to the node.
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #7714, opened 2023-05-29
Remove remaining node update calls in calico-node startup

skmatti:
## Description

Remove node update on startup if CALICO_NETWORKING_BACKEND=none. Followup to https://github.com/projectcalico/calico/pull/7550

## Release Note

```release-note
None
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/pkg/lifecycle/startup/startup.go
+++ b/node/pkg/lifecycle/startup/startup.go
@@ -182,24 +182,21 @@ func Run() {
 		}
 	}
 
-	// If Calico is running in policy only mode we don't need to write BGP related details to the Node.
-	if os.Getenv("CALICO_NETWORKING_BACKEND") != "none" {
-		configureAndCheckIPAddressSubnets(ctx, cli, node, k8sNode)
-		// Configure the node AS number.
-		configureASNumber(node)
-	}
-
+	needsNodeUpdate := configureAndCheckIPAddressSubnets(ctx, cli, node, k8sNode)
+	// Configure the node AS number.
+	needsNodeUpdate = configureASNumber(node) || needsNodeUpdate
 	// Populate a reference to the node based on orchestrator node identifiers.
-	configureNodeRef(node)
+	needsNodeUpdate = configureNodeRef(node) || needsNodeUpdate
+	if needsNodeUpdate {
+		// Apply the updated node resource.
+		if _, err := CreateOrUpdate(ctx, cli, node); err != nil {
+			log.WithError(err).Errorf("Unable to set node resource configuration")
+			utils.Terminate()
+		}
+	}
 
 	// Check expected filesystem
 	ensureFilesystemAsExpected()
-
-	// Apply the updated node resource.
-	if _, err := CreateOrUpdate(ctx, cli, node); err != nil {
-		log.WithError(err).Errorf("Unable to set node resource configuration")
-		utils.Terminate()
-	}
 
 	// Configure IP Pool configuration.
 	configureIPPools(ctx, cli, kubeadmConfig)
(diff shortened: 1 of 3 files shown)


skmatti:
@mgleung Can you take a look?. Thanks

node/pkg/lifecycle/startup/startup.go
```diff
@@ -187,20 +187,18 @@ func Run() {
...
+		configureNodeRef(node)
+		// Apply the updated node resource.
+		if _, err := CreateOrUpdate(ctx, cli, node); err != nil {
+			log.WithError(err).Errorf("Unable to set node resource configuration")
+			utils.Terminate()
+		}
 	}
 
-	// Populate a reference to the node based on orchestrator node identifiers.
-	configureNodeRef(node)
```
caseydavenport, line 193: So I think the original PR worked because the IP address itself wasn't used by Calico in CALICO_NETWORKING_BACKEND=none mode, so simply not setting it in that mode was OK.

However, the node reference _is_ used in such a mode - namely, it's used when running Calico against an etcd data store (i.e., not via the k8s API) and using another routing solution (e.g., flannel). So I think this will break a small number of clusters if we move this inside that branch.

I think instead we'll need to conditionally call `CreateOrUpdate` based on whether we _need_ to do so.

e.g., track output from each of the `configureX` methods and then only update the Node if one of them signaled that the node object needs updating. 

WDYT?

skmatti: Thanks @caseydavenport for the quick review. I have updated the change based on your recommendation. Does this LGTY?.

node/pkg/lifecycle/startup/startup.go
```diff
@@ -182,25 +182,22 @@ func Run() {
...
-	// If Calico is running in policy only mode we don't need to write BGP related details to the Node.
-	if os.Getenv("CALICO_NETWORKING_BACKEND") != "none" {
-		configureAndCheckIPAddressSubnets(ctx, cli, node, k8sNode)
-		// Configure the node AS number.
-		configureASNumber(node)
-	}
-
+	needsNodeUpdate := configureAndCheckIPAddressSubnets(ctx, cli, node, k8sNode)
+	// Configure the node AS number.
+	needsNodeUpdate = needsNodeUpdate || configureASNumber(node)
```
caseydavenport, line 187: I think this line will skip execution of `configureASNumber` if `needsNodeUpdate` is true.

We probably want to do it like this instead.

```go
needsNodeUpdate = configureASNumber(node) || needsNodeUpdate
```

So that `configureASNumber` is always called

caseydavenport:
Just need to swap the ordering of the `||` clauses but otherwise this LGTM, will trigger tests once that is updated

node/pkg/lifecycle/startup/startup.go
skmatti: Thanks for catching this, revered the order.

caseydavenport:
Not sure if you've looked into this, but it would be nice to have some sort of simply unit test verifying that this behavior doesn't get regressed.

e.g., a UT that queries the node, makes note of its revision, runs this code, and verifies the revision is unchanged?

skmatti:
> Not sure if you've looked into this, but it would be nice to have some sort of simply unit test verifying that this behavior doesn't get regressed.
> 
> e.g., a UT that queries the node, makes note of its revision, runs this code, and verifies the revision is unchanged?

Do you mean a FV test like https://github.com/projectcalico/calico/blob/31bef1eaa73c2234fa8377f940c06b382a5fa0dd/node/pkg/lifecycle/startup/startup_test.go#L988 ?

caseydavenport (approved):

caseydavenport:
@skmatti yep, that file is probably the right place for it to go. 

I'm going to merge this for now so that it doesn't rot, but if you can I think it would be great to do a follow-on that adds such a test. Feel free to ping me here or on another PR and we work through it!

Issue #7717, opened 2023-05-30
GlobalNetworkPolicy does not distinguish between locally originated and forwarded/natted traffic in ebpf mode


sfudeus:
## Expected Behavior

It should be possible to generally restrict access to certain IP ranges (e.g. "the internet") for a cluster in general, i.e. for workloads and the worker nodes themselves, but to open up the connectivity for certain explicitly defined workloads. 

## Current Behavior

After enabling the ebpf dataplane, a previously working configuration of K8s network policies and Calico GlobalNetworkPolicy does not work as expected anymore. The GlobalNetworkPolicy (see below) is expected to disallow traffic to the internet while a namespaces-based k8s network policy allows traffic for a subset of pods. 
With the ebpf dataplane, even the labelled pods cannot reach the internet anymore, blocked via the GlobalNetworkPolicy. 

## Possible Solution

Possibly the `applyOnForward` is not properly working anymore, applying the implicit Deny restriction from the GlobalNetworkPolicy even to forwarded/natted traffic. 
Might by fixed by !7707 according to Slack discussion https://calicousers.slack.com/archives/CUKP5S64R/p1684923320556429.

## Steps to Reproduce (for bugs)

1. Apply a global networkpolicy for host endpoints, disallowing traffic to the internet
```
apiVersion: crd.projectcalico.org/v1
kind: GlobalNetworkPolicy
metadata:
  name: deny-host-internet
spec:
  applyOnForward: false
  egress:
  - action: Allow
    destination:
      nets:
      - 100.64.0.0/13
      - 100.72.0.0/16
      - 127.0.0.0/8
      - 169.254.0.0/16
      - 10.0.0.0/8
      - 172.16.0.0/12
      - 192.168.0.0/16
    ipVersion: 4
  selector: has(node-role.kubernetes.io/node)
  types:
  - Egress
```
2. Apply a k8s network policy to a namespace (here: `ams-testing-dev`)
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: internet-egress
  namespace: ams-testing-dev
spec:
  egress:
  - to:
    - ipBlock:
        cidr: 0.0.0.0/0
        except:
        - 10.0.0.0/8
        - 100.64.0.0/10
        - 172.16.0.0/12
        - 192.168.0.0/16
        - 224.0.0.0/4
        - 240.0.0.0/4
  podSelector:
    matchLabels:
      allow-egress-to-internet: "true"
  policyTypes:
[...]

PR #7724, opened 2023-05-31
[release-v3.26] Auto pick #7707: clasiffy correctly forwarded traffic

tomastigera:
Cherry pick of #7707 on release-v3.26.

#7707: clasiffy correctly forwarded traffic

# Original PR Body below

Previously, anything that had source the same as the host was considered as "from host" by the policy engine and global policies with applyOnForward=false would apply to such traffic. However, for instance SNATed traffic due to nat-outgoing would be also considered "from host" eventhough it originated in a workload.

This fixes it by first looking at a SEEN mark which indicates that the packet was seen by another endpoint and thus is forwarded with the exception of packets being routed via the bpf nat iface which is used by the host processes.

fixes https://github.com/projectcalico/calico/issues/7717

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: fix applyOnforward=false in global policies
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/skb.h
+++ b/felix/bpf-gpl/skb.h
@@ -172,4 +172,6 @@ static CALI_BPF_INLINE void skb_set_mark(struct __sk_buff *skb, __u32 mark)
 	);
 }
 
+#define skb_mark_equals(skb, mask, val) (((skb)->mark & (mask)) == (val))
+
 #endif /* __SKB_H__ */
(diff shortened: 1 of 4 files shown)


marvin-tigera:

mazdakn (approved):
LGTM.

PR #7753, opened 2023-06-06
Run yoga devstack tests

nelljerram:
Since doing the work for a DevStack setup with OpenStack Yoga - #7625 - we've had a successful DevStack setup in our CI, but haven't actually been running any tests!  A quick way of seeing this is that the CI job only takes 15 minutes when it should take about 25.  Obviously it can also be seen in the logs.

This change brings back the tests by splitting a single `sudo -u stack -H -E bash -x <<'EOF'` invocation, in networking-calico/devstack/bootstrap.sh, into two: one for the setup and one for the tests.  The presumed explanation for the problem is that something in the setup was closing stdin, which meant that bash would not read further commands after the exit of the `./stack.sh` line.

### DIFF
--- a/networking-calico/devstack/bootstrap.sh
+++ b/networking-calico/devstack/bootstrap.sh
@@ -163,11 +163,12 @@ ls -la /opt/stack
 
 # Stack!
 sudo -u stack -H -E bash -x <<'EOF'
-
-set
 cd /opt/stack/devstack
 ./stack.sh
+EOF
 
+sudo -u stack -H -E bash -x <<'EOF'
+cd /opt/stack/devstack
 if ! ${TEMPEST:-false}; then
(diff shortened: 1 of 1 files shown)


networking-calico/devstack/bootstrap.sh
```diff
@@ -163,11 +163,12 @@ ls -la /opt/stack
 
 # Stack!
 sudo -u stack -H -E bash -x <<'EOF'
-
-set
 cd /opt/stack/devstack
 ./stack.sh
+EOF
 
+sudo -u stack -H -E bash -x <<'EOF'
```
lwr20, line 170: Is it worth a little comment to explain why this is here twice?  It does look odd, but I understand from the explanation in the PR why its here :)

lwr20 (approved):
One minor comment, LGTM

marvin-tigera:

PR #7783, opened 2023-06-13
Fix checks for integer conversion overflows

mgleung:
## Description

Fixes a boundary comparison to Uint32 which overflows an Int32 in the comparison without type casting. Since the mask value is supposed to be 0-32 anyways, we can compare this to the bounds of an Int32.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/bpf.go
+++ b/felix/bpf/bpf.go
@@ -1311,8 +1311,9 @@ func CidrToHex(cidr string) ([]string, error) {
 		return nil, fmt.Errorf("IP %q is not IPv4", ip)
 	}
 
-	if mask > math.MaxUint32 || mask < 0 {
-		return nil, fmt.Errorf("mask %d should be between 0 and 4,294,967,295", mask)
+	// Check bounds on the mask since the mask will be in CIDR notation and should range between 0 and 32
+	if mask > 32 || mask < 0 {
+		return nil, fmt.Errorf("mask %d should be between 0 and 32", mask)
 	}
 
 	maskBytes := make([]byte, 4)


danudey (approved):
lgtm

felix/bpf/bpf.go
```diff
@@ -1311,8 +1311,10 @@ func CidrToHex(cidr string) ([]string, error) {
 		return nil, fmt.Errorf("IP %q is not IPv4", ip)
 	}
 
-	if mask > math.MaxUint32 || mask < 0 {
-		return nil, fmt.Errorf("mask %d should be between 0 and 4,294,967,295", mask)
+	// Check bounds on the mask against an int32 instead of uint32 since the mask value
+	// will be in subnet mask notation and should range between 0 and 32
+	if mask > math.MaxInt32 || mask < 0 {
```
mgleung, line 1316: I could change this check to check against `32` instead of `math.MaxInt32` but figured that since this is an integer overflow check that I would make it more obvious as opposed to making this a validation of proper CIDR input.

coutinhop: I don't fully follow, if this is subnet mask notation, the max (255.255.255.255) would be `MaxUint32`, not `MaxInt32` (4294967295)... Was there an issue with using the unsigned max?

coutinhop (approved):
LGTM, thanks @mgleung for the changes

PR #7812, opened 2023-06-22
Update timeout to be per-job and not per-promotion

danudey:
The current promotion job has a tendency to time out, due to the fact that the timeout is per-promotion. Currently that means that the whole promotion, including waiting for a Semaphore executor, is timed.

Moving the timeout into the job itself will only start the timer once an executor is available and the job has started, reducing the likelihood of a timeout cancelling a job.

Testing on the test-release-build promotion first.

### DIFF
--- a/.semaphore/release/test-release-build.yml
+++ b/.semaphore/release/test-release-build.yml
@@ -5,9 +5,6 @@ agent:
     type: e1-standard-4
     os_image: ubuntu2004
 
-execution_time_limit:
-  minutes: 180
-
 blocks:
   - name: "Test build official release"
     skip:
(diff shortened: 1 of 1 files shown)


.semaphore/release/test-release-build.yml
```diff
@@ -5,9 +5,6 @@ agent:
     type: e1-standard-4
     os_image: ubuntu2004
 
-execution_time_limit:
```
lwr20, line 8: You can't remove this - removal means that semphore will default to 1 hour.  And then it will complain that the job timeout is longer than the pipeline timeout.

You need to leave this in, but increase the pipeline timeout to be much longer than the job timeout.

danudey: Thanks for the catch!

PR #7818, opened 2023-06-26
Add Makefile utility to pull operator CRD updates

tmjd:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/charts/tigera-operator/Makefile
+++ b/charts/tigera-operator/Makefile
@@ -1,0 +1,14 @@
+include ../../metadata.mk
+include ../../lib.Makefile
+
+OPERATOR_REPO ?= tigera/operator
+OPERATOR_VERSION ?= $(shell grep '^ *version:' values.yaml | sed -e 's/.*version: //')
+
+
+fetch-manifests:
+	$(eval tmp_dir := $(shell mktemp -d))
+	git clone --depth=1 -b $(OPERATOR_VERSION) https://github.com/$(OPERATOR_REPO) $(tmp_dir)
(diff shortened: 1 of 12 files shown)


charts/tigera-operator/Makefile
```diff
@@ -0,0 +1,23 @@
+include ../../metadata.mk
+include ../../lib.Makefile
+
+OPERATOR_REPO ?= tigera/operator
+OPERATOR_VERSION ?= $(shell grep '^ *version:' values.yaml | sed -e 's/.*version: //')
+
+TMP_DIR = .tmp
```
caseydavenport, line 7: You might want to use `mktemp` for this

caseydavenport (approved):
LGTM - might be nice to use a proper temp directory though rather than a checkout in the local filesystem. Less worry about stale data getting left behind for some reason in-tree.

charts/tigera-operator/Makefile
tmjd: I think it should be consistent. If there were more targets that used that the directory then it will be better for that to be a specific directory. Would you prefer if I called it "build" or something like that?

caseydavenport: Consistent with what? `mktemp` makes a temporary directory in `/tmp` - e.g., `/tmp/tmp.Awllr9y9VR`

I'm just suggesting that the temporary dir doesn't need to be within the git tree

Issue #7735, opened 2023-06-02
An option to configure the felix via operator


resworld:
Hello there,

basically I want to be able to configure the felix software via helm installation so I can change some configs defined here: https://docs.tigera.io/archive/v3.7/reference/felix/configuration
I can find that my calico configuration is having felix cfg file, but it is hardcoded in the container: /etc/calico/felix.cfg

## Expected Behavior

I am expecting to have ability to configure felix via values.yaml.

## Current Behavior

Currently I am seeing it is not possible: https://docs.tigera.io/calico/latest/reference/felix/configuration
```
If you have installed Calico using the operator, you cannot modify the environment provided to felix directly.
```

## Possible Solution

Maybe would be great if we can add different options in the values.yaml, for example:
```
installation:
  kubernetesProvider: EKS
  felix:
    LogFilePath: /tmp/logfile.log
    UsageReportingEnabled: false

```

## Steps to Reproduce (for bugs)

Follow instruction for helm install calico

## Context
I want to change some configurations in felix

## Your Environment
* Calico version: docker.io/calico/node:v3.25.0
* Orchestrator version (e.g. kubernetes, mesos, rkt): EKS

caseydavenport:
Right now, the way to do this is to create a FelixConfiguration API object. However, this has to happen post-helm, because the API won't exist until after Calico is installed.

resworld:
Hi @caseydavenport and thank you for the reply,

Is there any document how to do that? Sorry, I am not very familiar with the Felix and not sure if I know the proper steps I need to follow. Thank you.

PauloGDPeixoto:
I created a PR to address this issue: https://github.com/projectcalico/calico/pull/7836

WoodyWoodsta:
Is there a way to configure the `default` `FelixConfiguration` that is created by the operator?

I want to enable wireguard on all nodes by default. The documentation for felix configuration is this:

> Configuration for Felix is read from one of four possible locations, in order, as follows.
> - Environment variables.
> - The Felix configuration file.
> - Host-specific FelixConfiguration resources (node.<nodename>).
> - The global FelixConfiguration resource (default).

Based on my understanding of this and a bit of experimentation, the only way to set a global default is to update the `default` resource. However, since this is something that is created by the operator, it cannot be IaC managed, unless the operator exposes configuration via helm.

caseydavenport:
@WoodyWoodsta I believe there is some work to expose FelixConfiguration via the helm chart in the works. 

[...]

PR #7836, opened 2023-07-05
enable direct customization of the FelixConfiguration

PauloGDPeixoto:
## Description

New feature: Include an option to directly customize the FelixConfiguration.

## Related issues/PRs

#7735

### DIFF
--- a/charts/tigera-operator/templates/crs/custom-resources.yaml
+++ b/charts/tigera-operator/templates/crs/custom-resources.yaml
@@ -23,3 +23,17 @@ metadata:
 spec:
 {{ $apiServerSpec | toYaml | indent 2 }}
 {{ end }}
+
+
+{{ if .Values.felixconfiguration.enabled }}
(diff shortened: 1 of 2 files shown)


lwr20:
Have you tried this out?  I can think of some reasons why this won't work, but would love to be proved wrong.

PauloGDPeixoto:
> Have you tried this out? I can think of some reasons why this won't work, but would love to be proved wrong.

Yes I have, and it works. This is basically duplicating what is done for the Installation, why wouldn't it work?

lwr20:
FelixConfiguration is a `projectcalico.org/v3` object.  That will only be applyable once the calico-apiserver is running - while that's down (e.g. in the case of installing calico on a fresh cluster), applying that object will fail.

PauloGDPeixoto:
You're absolutely correct, I was under the impression that I had created a cluster with this code in place, but in fact, I didn't.

PauloGDPeixoto:
Awesome, thanks for the feedback and good luck working that out!

PR #7852, opened 2023-07-10
Enhance kubebuilder validation in felixconfig crd

ti-afra:
### What
adding kubebuilder validation for felixconfiguration CRD where-ever possible, mostly adding regex to have kubernetes api validate user input to prevent user errors.

### Why
customers have asked for option to apply felix configurations via helm install for calico enterprise. Since apiserver is not up at that time, apiversion crd.projectcalico.org/v1 should be used for that workflow. This means that, creating custom resources is not going through the apiserver and missing the validations that comes with it.

## Related issues/PRs
related to https://github.com/tigera/calico-private/pull/6386
jira: EV-3846
userstory: PMREQ-466

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -116,34 +124,48 @@ type FelixConfigurationSpec struct {
...
 	// attempts to acquire the iptables lock if it is not available. Lower values make Felix more
 	// responsive when the lock is contended, but use more CPU. [Default: 50ms]
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ms|s|m|h))+$"
 	IptablesLockProbeInterval *metav1.Duration `json:"iptablesLockProbeInterval,omitempty" configv1timescale:"milliseconds" confignamev1:"IptablesLockProbeIntervalMillis"`
 	// FeatureDetectOverride is used to override feature detection based on auto-detected platform
 	// capabilities.  Values are specified in a comma separated list with no spaces, example;
 	// "SNATFullyRandom=true,MASQFullyRandom=false,RestoreSupportsLock=".  "true" or "false" will
 	// force the feature, empty or omitted values are auto-detected.
+	// +kubebuilder:validation:Pattern="^([a-zA-Z]+=(true|false|),)*([a-zA-Z]+=(true|false|))$"
```
tomastigera, line 140: strings dont need `// +kubebuilder:validation:Type=string` because kubebuilder can derive that?

tomastigera: is it possible that the feauteru name can include digits and - or _ ? Like say `myipv6feature` ?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -116,34 +124,48 @@ type FelixConfigurationSpec struct {
...
 	FeatureDetectOverride string `json:"featureDetectOverride,omitempty" validate:"omitempty,keyValueList"`
 	// FeatureGates is used to enable or disable tech-preview Calico features.
 	// Values are specified in a comma separated list with no spaces, example;
 	// "BPFConnectTimeLoadBalancingWorkaround=enabled,XyZ=false". This is
 	// used to enable features that are not fully production ready.
+	// +kubebuilder:validation:Pattern="^([a-zA-Z]+=(enabled|disabled|true|false),)*([a-zA-Z]+=(enabled|disabled|true|false))$"
```
tomastigera, line 146: the list of values here is tricky, we have a gate atm with `udp` as a value. Perhaps anything that does not include `=` ? idk :thinking:

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -180,6 +202,7 @@ type FelixConfigurationSpec struct {
 	// Calico's rules from being bypassed. If you switch to append mode, be sure that the other rules in the chains
 	// signal acceptance by falling through to the Calico rules, otherwise the Calico policy will be bypassed.
 	// [Default: insert]
+	// +kubebuilder:validation:Enum=insert;append
```
tomastigera, line 205: could it be that we also allow `Insert` or `Append` ? :thinking:

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -303,8 +339,12 @@ type FelixConfigurationSpec struct {
 	// server. For example, if a significant security vulnerability has been discovered in the version of Calico being used. [Default: true]
 	UsageReportingEnabled *bool `json:"usageReportingEnabled,omitempty"`
```
tomastigera, line 340: bool does not need the kubebuilder annotation?

tomastigera (commented):
I am not sure about case sensitivity. If our validations of values are not case sensitive (which ikd :sweat_smile: ) this could break existing setups. That is my only worry.

api/pkg/apis/projectcalico/v3/felixconfig.go
ti-afra: I remember testing this and it only accepted "true" and "false" for bool. I think kubebuilder already handles that based on the type.

api/pkg/apis/projectcalico/v3/felixconfig.go
ti-afra: replacing the regex with ^([a-zA-Z]+=(enabled|disabled|true|false),)*([a-zA-Z]+=([^=]+))$ to allow any character other than "=" for values.

api/pkg/apis/projectcalico/v3/felixconfig.go
ti-afra: your brought up a good point - I was thinking everything is case-insensitive. However looks like it's not. Just checked the config_params.go and looks like the values for this one is reflecting what has been defined in the code:

`ChainInsertMode             string `config:"oneof(insert,append);insert;non-zero,die-on-fail"`

api/pkg/apis/projectcalico/v3/felixconfig.go
ti-afra: updating the regex to `^([a-zA-Z0-9-_]+=(true|false|),)*([a-zA-Z0-9-_]+=(true|false|))$`

ti-afra:
changing enum validations to regex to accept values independent of the case sensitivity. need to add `(?i)` in front of the regex.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -116,34 +124,48 @@ type FelixConfigurationSpec struct {
...
 	// IpsetsRefreshInterval is the period at which Felix re-checks all iptables
 	// state to ensure that no other process has accidentally broken Calico's rules. Set to 0 to
 	// disable iptables refresh. [Default: 90s]
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern=`^([0-9]+(\\.[0-9]+)?(ms|s|m|h))*$`
 	IpsetsRefreshInterval *metav1.Duration `json:"ipsetsRefreshInterval,omitempty" configv1timescale:"seconds"`
 	MaxIpsetSize          *int             `json:"maxIpsetSize,omitempty"`
 	// IptablesBackend specifies which backend of iptables will be used. The default is Auto.
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern=`^(?i)(Auto|FelixConfiguration|FelixConfigurationList|Legacy|NFT)?$`
```
caseydavenport, line 157: For enum type validation, you can actually use a kubebuilder tag specifically for enums - here's an example from our operator codebase: https://github.com/tigera/operator/blob/master/api/v1/installation_types.go#L642-L643

I believe you can also put the validation tag on the `type IptablesBackend struct` line so that it is inherited everywhere this is used (although I think it's just in once place)

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -180,6 +202,7 @@ type FelixConfigurationSpec struct {
 	// Calico's rules from being bypassed. If you switch to append mode, be sure that the other rules in the chains
 	// signal acceptance by falling through to the Calico rules, otherwise the Calico policy will be bypassed.
 	// [Default: insert]
+	// +kubebuilder:validation:Pattern=`^(?i)(insert|append)?$`
```
caseydavenport, line 205: Do we support `Insert` and `Append` (uppercase) as well?

caseydavenport (commented):
@ti-afra thank you! This is great. A couple of minor comments but I'm excited to roll this out!

api/pkg/apis/projectcalico/v3/felixconfig.go
caseydavenport: > changing enum validations to regex to accept values independent of the case sensitivity. need to add (?i) in front of the regex.


haha, I see now that you already moved away from this in order to do case insensitivity. Nice!

tomastigera (approved):
lgtm

PR #7856, opened 2023-07-11
Issue 7057: Print out the whole failed iptables command in the log for helping debugging

lubronzhan:
## Description

Print out the whole failed iptables command so user could manually execute it

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Print out the whole failed iptables command if it fails in debug mode
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/iptables/table.go
+++ b/felix/iptables/table.go
@@ -880,7 +880,8 @@ func (t *Table) attemptToGetHashesAndRulesFromDataplane() (hashes map[string][]s
 	}
 	waitErr := cmd.Wait()
 	if waitErr != nil {
-		log.WithError(waitErr).Warnf("iptables save command '%s' failed", cmd.String())
+		log.WithError(waitErr).Warn("iptables save failed")
+		log.Debugf("failing iptables save command is '%s", cmd.String())
 		if err == nil {
 			err = waitErr
 		}
--- a/felix/iptables/table.go
+++ b/felix/iptables/table.go
@@ -880,8 +880,11 @@ func (t *Table) attemptToGetHashesAndRulesFromDataplane() (hashes map[string][]s
 	}
 	waitErr := cmd.Wait()
 	if waitErr != nil {
-		log.WithError(waitErr).Warn("iptables save failed")
-		log.WithFields(log.Fields{"table": t.Name, "cmd": t.iptablesSaveCmd}).Warn("iptables save failed")
+		if log.IsLevelEnabled(log.DebugLevel) {
+			t.logCxt.WithFields(log.Fields{"cmd": t.iptablesSaveCmd}).Warn("iptables save failed")
+		} else {
+			log.WithError(waitErr).Warn("iptables save failed")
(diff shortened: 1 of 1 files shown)


caseydavenport:
I think this could get quite spammy. I'd rather us have a debug log to print this out so that it's only emitted if specifically in debug /developer mode.

lubronzhan:
Ok I changed it to print it out in a different line with debug

felix/iptables/table.go
```diff
@@ -869,6 +869,7 @@ func (t *Table) attemptToGetHashesAndRulesFromDataplane() (hashes map[string][]s
 	waitErr := cmd.Wait()
 	if waitErr != nil {
 		log.WithError(waitErr).Warn("iptables save failed")
+		log.Debugf("failing iptables save command is '%s", cmd.String())
```
caseydavenport, line 872: I think for optimization purposes, we should structure this like so:

```go
if log.IsLevelEnabled(log.DebugLevel) {
  log.Debugf(...)
}
```

This ensures we don't do any extra work (e.g., rendering the String() function) when not actually in debug mode.

caseydavenport: Looking at this a little bit closer, I am not sure there's a whole lot of value to be gained from this log.

Namely, because we always run the following:

```
cmd := t.newCmd(t.iptablesSaveCmd, "-t", t.Name)
```

which turns into

```
iptables-save -t TABLE
```

The only thing that really varies is the table name. I think it would be better to modify the existing warning log to look like this:

```go
log.WithFields(log.Fields{"table": t.Name, "cmd": t.iptablesSaveCmd}).Warn("iptables save failed")

**Comment by lubronzhan** [comment]
Hi @caseydavenport any chance to get this merged? Thank you

File: felix/iptables/table.go [hunk 2]
```diff
@@ -869,6 +869,7 @@ func (t *Table) attemptToGetHashesAndRulesFromDataplane() (hashes map[string][]s
 	waitErr := cmd.Wait()
 	if waitErr != nil {
 		log.WithError(waitErr).Warn("iptables save failed")
+		log.WithFields(log.Fields{"table": t.Name, "cmd": t.iptablesSaveCmd}).Warn("iptables save failed")
```
caseydavenport, line 872:
```go
if log.IsLevelEnabled(log.DebugLevel) {
  t.logCxt.WithFields(log.Fields{"cmd": t.iptablesSaveCmd}).Warn("iptables save failed")
}
```

I think I suggested only doing this at debug level in a previous comment. It should look something like this ideally (since log.WithFields will otherwise be executed even if not running in debug).

We can also use `t.logCxt` which already includes the table and ipVersion fields.

lubronzhan: Got it. Sorry misunderstood the previous comment.
I just fixed it

caseydavenport (approved):

PR #7857, opened 2023-07-11
Add GA support for Windows hostprocess containers

coutinhop:
Build on the changes from https://github.com/projectcalico/calico/pull/7260:
- Add a Windows cni-plugin hostprocess image, which will install the cni binaries and config on the host.
- Add a Windows node hostprocess image, which no longer needs to match the hosts' Windows version

Manually tested with an operator image from https://github.com/tigera/operator/pull/2732 and a VXLAN setup.

Many thanks to @jsturtevant and @fabi200123

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Add support for Windows host process containers (HPC), with Calico for Windows daemonsets managed by kubernetes.
```

## Reminder for the reviewer

Other optional labels:


node/Dockerfile-windows
```diff
@@ -0,0 +1,44 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG WINDOWS_VERSION
+
+FROM mcr.microsoft.com/windows/nanoserver:${WINDOWS_VERSION}
```
jsturtevant, line 17: should be able to use `mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0` since it is HPC image.  This will make it OS agnostic and faster download: https://github.com/microsoft/windows-host-process-containers-base-image

cni-plugin/pkg/install/install.go
```diff
@@ -118,6 +120,20 @@ func loadConfig() config {
 	return c
 }
 
+// getServiceAccountFilePaths returns the mount paths for a container
+// In the case of Windows HostProcess containers this prepends the CONTAINER_SANDBOX_MOUNT_POINT env variable
+// for other operating systems or if the sandbox env variable is not set it returns the standard mount points
+// see https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/#volume-mounts
+func getHostPath(path string) string {
```
jsturtevant, line 127: if you only support Containerd  1.7+ this shouldn't be needed and would clean up this section of code

node/Dockerfile-windows
coutinhop: @jsturtevant I didn't know about that image, this is awesome! Thanks, will use it!

cni-plugin/pkg/install/install.go
coutinhop: I see, so then just referring to the paths as `/host/foo` would work on Windows without any need for change? Thanks, I hadn't tried that

jsturtevant: yes, if they are volumemouns to the `/host/` location that should work.  This is why the In-Cluster config now works 🚀

coutinhop: @jsturtevant I got rid of this `getHostPath()` function on my latest commit, though I still needed to use the `$CONTAINER_SANDBOX_MOUNT_POINT` env var when referring to the binaries inside the container, which from what I read on https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/#containerd-v1-7-and-greater shouldn't be needed, right? Will tag you on the relevant chunk of code...

cni-plugin/pkg/install/install.go
```diff
@@ -196,14 +197,20 @@ func Install() error {
 			continue
 		}
 
+		containerBinDir := "/opt/cni/bin/"
+		// The binaries dir in the container needs to be prepended by the CONTAINER_SANDBOX_MOUNT_POINT env var on Windows Host Process Containers
+		// see https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/#containerd-v1-7-and-greater
+		if runtime.GOOS == "windows" {
+			containerBinDir = os.Getenv("CONTAINER_SANDBOX_MOUNT_POINT") + "/" + containerBinDir
```
coutinhop, line 204: @jsturtevant I still needed to use the `$CONTAINER_SANDBOX_MOUNT_POINT` env var when referring to the binaries inside the container here, which from what I read on https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/#containerd-v1-7-and-greater shouldn't be needed, right? Could you please take a look an let me know if I missed something?

jsturtevant: what does the yaml look like for your deployment? I don't see it in the PR here.  Could it be that the Working directory is set to something?

coutinhop: I've mostly been working with the tigera operator with the changes from this PR: https://github.com/tigera/operator/pull/2732, but the deployment also works with this manifest: https://gist.github.com/coutinhop/7528503be5d0aeb22001d69d82fa61ed

No working dir on the install-cni initContainer, but I did include `$env:CONTAINER_SANDBOX_MOUNT_POINT` in the path for `command` in it, could that cause issues?

jsturtevant: > No working dir on the install-cni initContainer, but I did include `$env:CONTAINER_SANDBOX_MOUNT_POINT` in the path for `command` in it, could that cause issues?

yes I think that would do it.  could you try with something like `command: ["opt/cni/bin/install.exe"]` or `/opt/cni/bin/install.exe` depending on where those files live?

coutinhop: Thanks! I've tried it (removing mentions to `$env:CONTAINER_SANDBOX_MOUNT_POINT`), but I got
```
'/opt/cni/bin/install.exe' is not recognized as an internal or external command
```

I see that the yaml for kube-proxy also still uses `$env:CONTAINER_SANDBOX_MOUNT_POINT`
https://github.com/kubernetes-sigs/sig-windows-tools/blob/3ed28295f8694dffd03e144af4423137b2b8c30b/hostprocess/calico/kube-proxy/kube-proxy.yml#L25

Apparently the binaries (and scripts, I couldn't manage to make the uninstall-calico initContainer work either) are not in these absolute paths, I'm guessing some path in `c:/hpc/...` would work, but then why not just use `$env:CONTAINER_SANDBOX_MOUNT_POINT`, right?

This is by far not a deal breaker, stuff was working so I wouldn't block anything by it (unless you see `$env:CONTAINER_SANDBOX_MOUNT_POINT` being dropped at some point and these containers as they are becoming unsupported).

One other quick question: is the presence of `$env:CONTAINER_SANDBOX_MOUNT_POINT` a reliable indicator that we're inside a host process container? We've used it in a couple of places, I wonder if we should look into changing that...

Once again, thanks so much!

cni-plugin/pkg/install/install.go
coutinhop: Just a quick update: I ended up implementing support for containerd v1.6, so we're using `$CONTAINER_SANDBOX_MOUNT_POINT` just about everywhere that's necessary. Everything's working on my tests so far, and that way we can support a containerd version that's widely adopted right now...

cni-plugin/pkg/install/install.go
```diff
@@ -447,6 +439,11 @@ func copyFileAndPermissions(src, dst string) (err error) {
 		return fmt.Errorf("failed to rename file: %s", err)
 	}
 
+	if runtime.GOOS == "windows" {
+		// chmod doesn't work on windows
```
mgleung, line 443: Should we include a log of some sort here so that we have an idea why it quit early?

cni-plugin/pkg/install/install.go
```diff
@@ -502,12 +499,16 @@ current-context: calico-context`
...
 
-	if err := os.WriteFile("/host/etc/cni/net.d/calico-kubeconfig", []byte(data), 0600); err != nil {
+	if err := os.WriteFile(winutils.GetHostPath("/host/etc/cni/net.d/calico-kubeconfig"), []byte(data), 0600); err != nil {
 		logrus.Fatal(err)
 	}
 }
 
 func setSuidBit(file string) error {
+	if runtime.GOOS == "windows" {
+		// chmod doesn't work on windows
```
mgleung, line 509: Should we include a log of some sort here so that we have an idea why it quit early?

apiserver/cmd/apiserver/server/watch.go
```diff
@@ -30,10 +29,16 @@ func WatchExtensionAuth(stopChan chan struct{}) (bool, error) {
...
+	// be trusted in this case
+	// FIXME: this will no longer be needed when containerd v1.6 is EOL'd
+	if winutils.InHostProcessContainer() {
+		cfgFile = ""
+	}
+	cfg, err := winutils.BuildConfigFromFlags("", cfgFile)
 	if err != nil {
 		// attempt 2: in cluster config
-		if cfg, err = rest.InClusterConfig(); err != nil {
+		if cfg, err = winutils.GetInClusterConfig(); err != nil {
```
mgleung, line 41: I've noticed this takes the place of our old function for retrieving the in cluster config. Is this general-purpose enough where it can replace our old call for Linux envs too?

cni-plugin/pkg/install/install.go
coutinhop: Good point, will do!

cni-plugin/pkg/install/install.go
coutinhop: Will do, thanks!

apiserver/cmd/apiserver/server/watch.go
coutinhop: Yes it is, `winutils.GetInClusterConfig()` is an exact copy of `rest.InClusterConfig()` except for it uses `winutils.GetHostPath()` on the `tokenFile` and `rootCAFile`. `winutils.GetHostPath()` in turn, is a no-op on Linux envs (and on non-HPC windows envs too). I've intentionally left `rest.InClusterConfig()` in tests so that we're able to catch it if the functions drift apart. Hopefully containerd v1.7 will be widely adopted soon and we can drop these workarounds... Thanks @mgleung!

mgleung (approved):
LGTM

PR #7863, opened 2023-07-13
v3.26 postrelease test revamp

danudey:
## Description

A (largely structural) revamp of the postrelease tests:

* Update the tests to use Python 3
* Update the tests for pytest
* Refactor some multiple-test tests into separate tests

Several tests were also updated with more checks, e.g. the test for Github release artifacts was updated to include all current release artifacts instead of only some.

Some test sets were also combined to reduce redundancy.


lwr20 (commented):
Looks good. 
I recommend using `black` to format the python.  Its a quick, cheap way of getting consistent python formatting between users and PRs.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+# NOTE: this pre-commit file is currently only configured to run on
+#       files in hack/postrelease. If you want to add additional
+#       paths to test, update the value for files:, which is a regex.
+#
+#       See: https://pre-commit.com/index.html#regular-expressions
+
+
+default_language_version:
+  python: python3.11
```
mgleung, line 12: I'm assuming that these pre commit checks are just for our release/post-release tests since they are the only things written in python right? Will this conflict with our go code?

danudey: There won't be any interference, no.

`pre-commit` (if you've set up the hooks) will run a series of linters/formatters before commits (to catch various issues before you commit them). The `default_language_version` setting just tells `pre-commit` to tell those linters to assume they're linting Python 3.11 code.

On top of that, the `files:` key at the bottom of the file limits `pre-commit` to only process files matching the supplied regex (in this case, `hack/postrelease/.*`) and ignore anything else.

Developers are welcome to extend the pre-commit configuration to include other checks on other code, but for the moment doing so will likely cause significant whitespace changes that no one wants so that's not enabled.

mgleung: ahhhh sorry, missed the lines just below :stuck_out_tongue:

hack/postrelease/tests/test_manifests_archives.py
```diff
@@ -24,31 +24,37 @@
...
-
-def check_manifest_present(manifest):
-    print("[INFO] checking {} is in archive".format(manifest))
+OVERRIDE_FILE = os.getenv("OVERRIDE_FILE")
+
+
+@pytest.fixture(scope="session")
+def image_file_members(tmpdir_factory):
+    if OVERRIDE_FILE:
+        tf = tarfile.open(OVERRIDE_FILE, "r|gz")
```
mgleung, line 33: I think you'll want to use a context manager to keep track of the opened file so that it will be closed automatically when the context ends. That or call `close()` manually.

danudey: Done!

mgleung (approved):
LGTM

PR #7864, opened 2023-07-14
Add deduplicating syncer callbacks buffer.

fasaxc:
## Description

The deduplicating buffer implements the syncer callbacks API on its input, and calls another syncer callback on its output. In-between it maintains an in-order queue of KV updates and a tracking map to record what is in the queue.

If an update comes in for a key that already has a value in the queue then the value in the queue is replaced with the updated KV.

This can cause reordering (which is allowed by the syncer API) but it ensures that the amount of KVs in flight is bounded to the total size of the datastore even under substantial overload. The effect is that the client will periodically "skip ahead" to the more recent state of the datatstore without seeing intermediate states.
## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Add a decoupling buffer between the Typha client library and downstream code.  This prevents the Typha client from being blocked if the downstream code blocks for too long.  This prevents timeouts at very high scale.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/daemon/daemon.go
+++ b/felix/daemon/daemon.go
@@ -33,6 +33,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
+	"github.com/projectcalico/calico/libcalico-go/lib/backend/syncersv1/dedupebuffer"
+
 	"github.com/projectcalico/calico/libcalico-go/lib/seedrng"
 
 	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
@@ -485,7 +487,7 @@ configRetry:
 	// which will feed the calculation graph with updates, bringing Felix into sync.
 	var syncer Startable
 	var typhaConnection *syncclient.SyncerClient
-	syncerToValidator := calc.NewSyncerCallbacksDecoupler()
+	syncerToValidator := dedupebuffer.New()
 
 	if typhaDiscoverer.TyphaEnabled() {
 		// Use a remote Syncer, via the Typha server.
@@ -629,7 +631,7 @@ configRetry:
 	// calculation graph.
 	validator := calc.NewValidationFilter(asyncCalcGraph, configParams)
 
-	go syncerToValidator.SendTo(validator)
+	go syncerToValidator.SendToSinkForever(validator)
 	asyncCalcGraph.Start()
 	log.Infof("Started the processing graph")
 	var stopSignalChans []chan<- *sync.WaitGroup
--- a/libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
+++ b/libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
@@ -16,6 +16,7 @@ package dedupebuffer
 
 import (
 	"container/list"
+	"fmt"
 	"sync"
 
 	log "github.com/sirupsen/logrus"
@@ -33,11 +34,11 @@ import (
 // If an update comes in for a key that already has a value in the queue then
 // the value in the queue is replaced with the updated KV.
 //
-// This can cause reordering (which is allowed by the syncer API) but it
-// ensures that the amount of KVs in flight is bounded to the total size of
-// the datastore even under substantial overload. The effect is that the
-// client will periodically "skip ahead" to the more recent state of the
-// datatstore without seeing intermediate states.
+// This can cause reordering between different resources (which is allowed
+// by the syncer API) but it ensures that the amount of KVs in flight is
+// bounded to the total size of the datastore even under substantial overload.
+// The effect is that the client will periodically "skip ahead" to the more
+// recent state of the datastore without seeing intermediate states.
 type DedupeBuffer struct {
 	lock sync.Mutex
 	cond *sync.Cond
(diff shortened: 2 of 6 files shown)


libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
```diff
@@ -0,0 +1,283 @@
...
+		first := d.pendingUpdates.Front()
+		buf = append(buf, first.Value)
+		d.pendingUpdates.Remove(first)
+		if u, ok := first.Value.(updateWithStringKey); ok {
+			key := u.key
+			delete(d.keyToPendingUpdate, key)
+			// Update sentKeys now, before we drop the lock.  Once we drop
+			// the lock we're committed to sending these keys.
+			if u.update.Value == nil {
+				d.sentKeys.Discard(key)
```
nelljerram, line 238: Perhaps `liveResourceKeys` would be a better name than `sentKeys`?  With `sentKeys` my intuition was that we would never discard from the set.  The coding this way is certainly better, but then the naming is a little misleading.

libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
```diff
@@ -0,0 +1,283 @@
...
+			} else {
+				d.sentKeys.Add(key)
+			}
+		}
+	}
+	return buf
+}
+
+func (d *DedupeBuffer) dropLockAndSendBatch(sink api.SyncerCallbacks, buf []any) {
+	// RELEASE(!) lock while we send updates downstream.
```
nelljerram, line 248: I assume because we can get back pressure here?  Probably worth saying that explicitly.

libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
```diff
@@ -0,0 +1,283 @@
...
+		switch msg := msg.(type) {
+		case updateWithStringKey:
+			updates = append(updates, msg.update)
+		case api.SyncStatus:
+			if len(updates) > 0 {
+				if debug {
+					log.WithField("updates", updates).Debug("Sending updates (pre status update)")
+				}
+				sink.OnUpdates(updates)
+				updates = updates[len(updates):] // Re-slice to end so we don't share storage.
```
nelljerram, line 264: Does this still preserve capacity?  I guess it probably does, so just checking.

libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
```diff
@@ -0,0 +1,283 @@
...
+			log.Debug("Queue transitioned to non-empty; signalling.")
+		}
+		d.cond.Signal()
+	}
+}
+
+func (d *DedupeBuffer) SendToSinkForever(sink api.SyncerCallbacks) {
+	for !d.stopped {
+		d.sendNextBatchToSink(sink)
+	}
```
nelljerram, line 189: Is there subtlety hiding here in the fact that the code for sendNextBatchToSink is not inlined?

- Is it OK to query `d.stopped` outside the lock?
- Does it matter that we drop and reacquire the lock between iterations of the above loop?  Is that needed to allow upstream to add stuff to our queue, or would the dropping in dropLockAndSendBatch already suffice for that?

Depending on the answers, I'd slightly prefer inlining here, with one overall
```
d.lock.Lock()
defer d.lock.Unlock()
```
at the start of this function.

felix/daemon/daemon.go
```diff
@@ -485,7 +487,7 @@ configRetry:
 	// which will feed the calculation graph with updates, bringing Felix into sync.
 	var syncer Startable
 	var typhaConnection *syncclient.SyncerClient
-	syncerToValidator := calc.NewSyncerCallbacksDecoupler()
```
nelljerram, line 488: Is NewSyncerCallbacksDecoupler not needed now?  If so, does that mean there's code that can be deleted?

libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer_test.go
```diff
@@ -0,0 +1,250 @@
...
+				"foo3=bar3",
+				string(api.InSync),
+			}))
+		})
+	}
+}
+
+func applyNextBatchSync(d *DedupeBuffer, r *Receiver) {
+	Expect(d.pendingUpdates.Len()).NotTo(BeZero(), "Nothing on queue, sendNextBatchToSink would block")
+	d.sendNextBatchToSink(r)
```
nelljerram, line 149: Oh, maybe this is why this code wasn't inlined...

nelljerram (commented):
Looks pretty nice to me.  I've raised a few points, but I think they're all small or debatable.

libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
fasaxc: Yes, I was pretty sure but I checked the playground: https://go.dev/play/p/7niEE3qk93h

libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer_test.go
fasaxc: Yep, I had it inlined and broke it out for ease of testing.

libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+
+		if element, ok := d.keyToPendingUpdate[key]; ok {
+			// Already got an in-flight update for this key.
+			if u.Value == nil && !d.sentKeys.Contains(key) {
+				// This is a deletion, but the key in question never made it
+				// off the queue, remove it entirely.
+				if debug {
+					log.WithField("key", key).Debug("Key deleted before being sent.")
+				}
```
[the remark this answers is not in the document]
fasaxc: Yes, I think so, the problem is that WithFields creates a lot of garbage and this logic is on the critical path, with garbage being generated for every KV at start of day.

felix/daemon/daemon.go
fasaxc: SyncerCallbacksDecoupler is still used elsewhere; downstream of the new dedupe buffer (so back pressure there will push back on the dedupe buffer anyway, indirectly).  I think we only want one dedupe buffer at the updates source so that we dedupe as early as possible.  If we did it later, I think we'd get undesirable interactions between them.

nelljerram (approved):

fasaxc:
I think I'm just hitting existing flakes/failures that are being investigated already...

PR #7868, opened 2023-07-18
Update update_versions.py

danudey:
Update and make use of the `update_versions.py` script:

Workflow changes:

* Add make targets to run the version update script in Docker
* Add make targets to `hack/release` to run the version update on the release VM (for testing)
* Move everything into `hack/versions`
* Add Makefile for `hack/versions`

Code changes:

* Clean up output from update_versions.py
* Reformat code with `black` and `isort`


hack/versions/update_versions.py
```diff
@@ -0,0 +1,123 @@
...
+    return [int(ver) for ver in version.strip("v").split(".")]
+
+
+def unparse_version(version):
+    return f"v{version[0]}.{version[1]}.{version[2]}"
+
+
+def bump_version_string(version_string):
+    version = parse_version(version_string)
+    version[2] += 1
```
mgleung, line 52: I guess this means that this is only used for patch releases? Should we have an option to pass in that will allow us to use this for minor releases as well?

danudey: It should be easy enough to detect when we're on a new minor release based on the branch name (e.g. `release-v3.29` not matching the major/minor from the latest tag name, which would currently be v3.19.0).

I'll definitely add this as a feature, but it's outside of the scope of implementation right now.

mgleung (approved):
LGTM

PR #7871, opened 2023-07-19
Strip descriptions from openshift manifests

rene-dekker:
Openshift installations can currently fail, due to the large size of our CRDs. This PR strips the descriptions, making them considerably smaller (from 1.5MB to 0.5MB). We used to have a script like this in the past, but it got overlooked and removed when we moved to the new docs website, making this limitation of openshift reappear.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PACKAGE_NAME = github.com/projectcalico/calico
 
-include metadata.mk 
+include metadata.mk
 include lib.Makefile
 
 DOCKER_RUN := mkdir -p ./.go-pkg-cache bin $(GOMOD_CACHE) && \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -126,9 +126,15 @@ helm-index:
...
 # Creates the tar file used for installing Calico on OpenShift.
-bin/ocp.tgz: manifests/ocp/
-	mkdir -p bin
-	tar czvf $@ -C manifests/ ocp
+bin/ocp.tgz: manifests/ocp/ bin/yq
+	mkdir -p bin/tmp
+	cp -r manifests/ocp bin/tmp/
+	$(DOCKER_RUN) $(CALICO_BUILD) /bin/bash -c "                                        \
+		for file in bin/tmp/ocp/*crd* ;                                                 \
+        	do bin/yq -i 'del(.. | select(has(\"description\")).description)' \$$file ; \
```
rene-dekker, line 134: This may create a problem if we start to rely on fields that are legitimately called "description" . There are two in the repo: OpsGenieConfig and PagerDutyConfig, neither of which we use today.

PR #7882, opened 2023-07-26
Removing weak ciphers

### LINKED ISSUE #7881: Weak cipher suites supported on calico-typha
When the TLS endpoint that calico-typha exposes, the following ciphers are advertised as supported which are considered weak as they do not offer PFS (Perfect Forward Secrecy).

- TLS_RSA_WITH_AES_128_GCM_SHA256
- TLS_RSA_WITH_AES_256_GCM_SHA384

## Expected Behavior
Only Ephemeral suites should be used
- TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
- TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384

## Current Behavior
These suites are supported by the server:
- TLS_RSA_WITH_AES_128_GCM_SHA256
- TLS_RSA_WITH_AES_256_GCM_SHA384

## Possible Solution
PR to remove the 2x weak ciphers.

1. Deploy the calico-typha
2. Test the TCP TLS endpoint with a utility such as drwetter's testssl.sh
3. Observe that the two weak ciphers are advertised along with the two recommended / strong ECDHE ciphers.

huornlmj:
## Description

- Bug fix.
- 2x weak ciphers (TLS_RSA_WITH_AES_256_GCM_SHA384 and TLS_RSA_WITH_AES_128_GCM_SHA256) being removed as they do not provide perfect forward secrecy (PFS).
- Endpoint tested with testssl.sh and ciphers checked against SSLlabs and ciphersuite.info.
- calico-typha.

## Related issues/PRs
Fixes #7881

### DIFF
--- a/apiserver/cmd/apiserver/server/options.go
+++ b/apiserver/cmd/apiserver/server/options.go
@@ -120,8 +120,6 @@ func (o *CalicoServerOptions) Config() (*apiserver.Config, error) {
 		tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
 		tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
 		tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
-		tls.TLS_RSA_WITH_AES_128_GCM_SHA256,
-		tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
 	}
 	serverConfig.SecureServing.CipherSuites = cipherSuites
 	serverConfig.SecureServing.MinTLSVersion = tls.VersionTLS12


huornlmj:
Having these weak ciphers last in the server cipher order list would still result in a downgrade style attack, if a client were to only accept the weak ciphers.  Is there a way at least for the ciphers to be configured by end users wishing to harden their installations?

fasaxc:
I thought TLS prevent downgrade attacks by signing the handshake and rechecking it at the end.

Isn't the "client" here the Kubernetes API server? This isn't a publicly accessible component; there are no browsers accessing the API server.

huornlmj:
It doesn't really matter whether it's a browser or any other type of client, that's moot.  It's the fact that the server is willing to offer weak ciphers to any client, internal or external.

caseydavenport:
Sounds like we have an issue with compatibility on this one. How do we progress it? 

Do we just need a configuration option here? If so, what would it look like and how might a switch to new default ciphers look like? Or do we just want to allow direct configuration of this?

Behnam-Shobiri:
My understanding is based on the [RFC in Certificate Verify section](https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.3),  TLS should prevent downgrade attacks (at least in TLS 1.3).
As I mentioned, the order of ciphers is strongest first, and 2 ciphers that do not provide PFS are only for customers who do not use EC-based algorithms for compliance or other reasons. I checked golang [ciphers](https://go.dev/src/crypto/tls/cipher_suites.go) and these ciphers are not considered insecure (the same goes for CodeQL). 
I understand that using ciphers that provide PFS is preferred over the ciphers that do not, but I did not see any reference that calls them insecure.

Behnam-Shobiri:
@caseydavenport 
"Do we just need a configuration option here? If so, what would it look like and how might a switch to new default ciphers look like?"
I think one choice is to let users choose between 2 lists, the default (what we already have) or only PFS-enabled ciphers (the current list without the 2 ciphers). Allowing users to configure TLS directly can lead to configs that our FIPS certificate does not support.

huornlmj:
Hi again.  Is there something that can be done to either allow the PR or propose what can be done to control the weak ciphers?  Thanks.

caseydavenport:
Sounds like we need a new config option added as part of this PR to allow selection.

e.g., a new env var - `CIPHER_SUITE` with options `DefaultCiphers` or `PFSCiphers` - that configures whether or not to use these two.

caseydavenport:
Closing this due to inactivity, but happy to see it reopened. The decision here was that a new configuration option was needed to allow switching between different pre-defined sets of ciphers.

huornlmj:
Adding a new weak ciphers to the list that Calico typha supports in its default configuration:
TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA
TLS_RSA_WITH_AES_128_GCM_SHA256
TLS_RSA_WITH_AES_256_GCM_SHA384

huornlmj:
> Closing this due to inactivity, but happy to see it reopened. The decision here was that a new configuration option was needed to allow switching between different pre-defined sets of ciphers.

New weak ciphers were added this issue in May - can we reopen this please as the issue is not improving and companies performing security enumeration of their endpoints will start finding these as potential blockers where they have specific minimal security requirement policies.

lwr20:
I'll reopen assuming you're going to continue work on this PR?

huornlmj:
Still a thing 2 years later. (anti-idle comment)

lwr20:
FWIW, we are hearing that different organisations have different requirements for TLS ciphers - there is no "one size fits all", so a hard-coded solution isn't suitable.  

With that in mind, some work is currently going on to make the ciphers list configurable.

tagging @lucastigera

huornlmj:
Ping.  Those weak ciphers aren't getting any stronger! :)

lucastigera:
Closing this PR as it's been addressed in PR #10380 and will be available in Calico v3.31.

PR #7888, opened 2023-07-27
Add version/manifest committing

danudey:
Add the ability to commit the version changes from `make update-versions` and `make generate`:

* Add `hack/versions/commit_update_versions.sh` script with some simple commands to commit the changes we make. Down the road, we will add in some more build metadata to these commits (e.g. the link to the semaphore pipeline/job)
* Add `commit-versions-manifests` top-level `Makefile` target to run the above script. It also calls `check-dirty` and fails if any filers are still dirty
* Add `commit-versions-remote` target to `hack/release/Makefile` to call `make commit-versions-manifests` on the remote server during releases

Also update `manifests/generate.sh` logic

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ generate:
 	$(MAKE) -C app-policy protobuf
 	$(MAKE) gen-manifests
 
-gen-manifests: bin/helm
+gen-manifests: bin/helm bin/yq
 	cd ./manifests && \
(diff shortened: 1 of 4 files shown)


mgleung (approved):
LGTM

We may want to make improvements in the future where we parse the new version based off of an incrementation of the git tags though as well (though we would need to figure out how to specify minor/patch releases)

PR #7909, opened 2023-08-08
fix: adding timeout to the typha server TLS handshake

### LINKED ISSUE #7908: calico-typha hangs during unclean client TLS handshake
On certain special conditions, a client TLS handshake can block the typha server indefinitely, as you can see [here](https://github.com/projectcalico/calico/blob/60f8a9e884567ca690487849ff9df991091eced0/typha/pkg/syncserver/sync_server.go#L427), the `Handshake()` call is performed inside the main server handle `for` loop, without any timeout, so a miss behaving client, with an unclean TLS handshake, can take the server down and block the main loop indefinitely while other connections will be setting idle and waiting for that handshake to eventually finish.

## Expected Behavior

A misbehaving client should not impact the availability of the typha server, the TLS handshake must be performed with a safe upper timeout and eventually moved out of the main server handle `for` loop.

## Current Behavior

The client TLS handshake is performed in the main server handle `for` loop without a safe timeout, misbehaving clients can impact the availability of the typha server if the connection is not clearly closed/aborted at client side.

## Possible Solution

1. Implement a timeout in the TLS handshake to the duration of one Ping interval.
2. Eventually move the TLS handshake to an async go routine and unblock the main server handle `for` loop, so multiple client connections could be handled at the same time.

1. Perform a few cycles of Kubernetes cluster growth and shrink, eventually one of the nodes running the `calico-node` client will be abruptly terminated during client TLS handshake, leaving the connection towards the `calico-typha` hanging in an unclean state during TLS handshake, the `calico-typha` will never, under a reasonable time, recover from that, the server will be blocked for days and eventually all `calico-typha` replicas will get to the same state, bringing the node network down.
2. Another easy way to simulate this is to perform a `port-forward` to a `calico-typha` pod at port `5473` and then open a `telnet` connection to that port, the typha server will hang with a message like the following and other clients won't be able to connect to that typha server anymore:
```
2023-08-07 15:12:57.284 [INFO][7] sync_server.go 421: Accepted from xxx.xxx.xxx.xxx:60824 port=5473
```

## Context

This issue can eventually bring a Kubernetes production cluster running Calico with calico-typha completely down, please keep in mind the error is only reproduced on certain special conditions, during race conditions in the client termination in the exact moment of a TLS handshake with the typha server, over time, that can affect all replicas of the typha deployment, bringing the availability of the cluster network down.

## Your Environment

* Calico version - `v3.26.1`
* Orchestrator version (e.g. kubernetes, mesos, rkt): `kubernetes` using the `tigera-operator`
* Operating System and version: AWS EKS using Amazon EKS optimized Amazon Linux AMIs

rodrigorfk:
## Description

This PR adds a timeout to the TLS handshake in the typha server equals to one Ping duration, the intention with this is to avoid misbehaving clients to take the typha server down when connections are left in an unclean state during the handshake.

## Related issues/PRs

fixes #7908

## Todos

- [x] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
Add a timeout to Typha's TLS handshake.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/typha/fv-tests/server_test.go
+++ b/typha/fv-tests/server_test.go
@@ -20,6 +20,7 @@ import (
 	"encoding/gob"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"path/filepath"
@@ -1611,6 +1612,19 @@ var _ = Describe("with server requiring TLS", func() {
 		}
 	}
 
+	testTcpHalfOpen := func() {
+		deadline := time.Now().Add(9 * time.Second)
+		serverAddr := fmt.Sprintf("127.0.0.1:%d", server.Port())
+		tcpConn, err := net.Dial("tcp", serverAddr)
+		Expect(err).NotTo(HaveOccurred())
+		err = tcpConn.SetDeadline(time.Now().Add(30 * time.Second))
+		Expect(err).NotTo(HaveOccurred())
+		received := make([]byte, 1024)
+		_, err = tcpConn.Read(received)
+		Expect(err).Should(Equal(io.EOF))
+		Expect(time.Now().Unix()).Should(BeNumerically(">=", deadline.Unix()))
+	}
+
 	Describe("and CN or URI SAN", func() {
 		BeforeEach(func() {
 			requiredClientCN = clientCN
@@ -1685,5 +1699,15 @@ var _ = Describe("with server requiring TLS", func() {
 		It("TLS connection with good URI should fail", testTLSGoodURI(false))
 
 		It("TLS connection with good CN and URI should fail", testTLSGoodCNURI(false))
+	})
+
+	Describe("handshake", func() {
+		BeforeEach(func() {
+			requiredClientCN = clientCN
+			requiredClientURISAN = clientURISAN
+			serverCertName = "server"
+		})
+
+		It("should timeout after 10 seconds for TCP half open connections", testTcpHalfOpen)
 	})
 })
--- a/typha/pkg/syncserver/sync_server.go
+++ b/typha/pkg/syncserver/sync_server.go
@@ -101,6 +101,7 @@ const (
 	defaultBatchingAgeThreshold           = 100 * time.Millisecond
 	defaultPingInterval                   = 10 * time.Second
 	defaultWriteTimeout                   = 120 * time.Second
+	defaultHandshakeTimeout               = 10 * time.Second
 	defaultDropInterval                   = 1 * time.Second
 	defaultShutdownTimeout                = 300 * time.Second
 	defaultMaxConns                       = math.MaxInt32
@@ -140,6 +141,7 @@ type Config struct {
 	MinBatchingAgeThreshold        time.Duration
 	PingInterval                   time.Duration
 	PongTimeout                    time.Duration
+	HandshakeTimeout               time.Duration
 	WriteTimeout                   time.Duration
 	DropInterval                   time.Duration
 	ShutdownTimeout                time.Duration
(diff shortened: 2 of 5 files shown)


typha/pkg/syncserver/sync_server.go
```diff
@@ -424,7 +424,9 @@ func (s *Server) serve(cxt context.Context) {
 			// Doing TLS, we must do the handshake...
 			tlsConn := conn.(*tls.Conn)
 			logCxt.Debug("TLS connection")
-			err = tlsConn.Handshake()
+			handshakeCxt, handshakeCancel := context.WithTimeout(cxt, s.config.PingInterval)
```
fasaxc, line 427: Good catch, but I don't think PingInterval is the right one to use (it can be quite short and it's not a timeout).  We should really have a new timeout for this, but perhaps the WriteTimeout would be acceptable for a quick fix.

rodrigorfk: Hi @fasaxc , thanks for your comment, I am more than happy to change it by the `WriteTimeout` if you believe this is the right path for a quick fix, my assumption when deciding by the `PingInterval` was based on the default value of it, which is `10s`, that initially sounded reasonable to me, because a TLS handshake is supposed to be pretty fast under normal operation, like under milliseconds, isn't it? The `WriteTimeout` has a default value of 120s, which seems quite a lot, meaning that a misbehaving client can still block the server up to 2 minutes, affecting the scalability of the Calico stack. 
I would appreciate your guidance on how we should proceed. Thanks in advance.

fasaxc: I was about to say "don't worry, this is on its own goroutine" but now I look at it, that's not the case, looks like we're doing the TLS handshake synchronously after the `Accept()`, which is a mistake.

Let's just add  a new timeout, that's not too hard to do if you're up for it.  Jus tneed to add it in config_params.go and copy it "down the stack" in a couple of places.  Here's a commit that does it: https://github.com/projectcalico/calico/commit/f145093a8469ff7ac5250bfe80c43adaf95aef73

rodrigorfk: @fasaxc , I have addressed your comment and added a new timeout to the config, I have also provided a PR to the documentation repository, please take a look.

typha/fv-tests/server_test.go
```diff
@@ -1611,6 +1612,19 @@ var _ = Describe("with server requiring TLS", func() {
 		}
 	}
 
+	testTcpHalfOpen := func() {
+		serverAddr := fmt.Sprintf("127.0.0.1:%d", server.Port())
+		tcpConn, err := net.Dial("tcp", serverAddr)
+		Expect(err).NotTo(HaveOccurred())
+		err = tcpConn.SetDeadline(time.Now().Add(30 * time.Second))
+		Expect(err).NotTo(HaveOccurred())
+		deadline := time.Now().Add(10 * time.Second)
```
fasaxc, line 1621: This calculation happens in parallel with the server's calculation so it races with that; suggest moving this line to before the `net.Dial()` and maybe reducing to 9s, just to allow some leeway.

fasaxc (changes requested):
Nice, just one comment from me

typha/pkg/syncserver/sync_server.go
fasaxc: Sorry for not getting back sooner, yes, 10s seems fine to me.  I'd like it to be 2s but it's not easy to configure these values when using the operator to install Calico so let's compromise on 10s for now.

typha/fv-tests/server_test.go
rodrigorfk: sure thing, thanks for the suggestion, it has been applied, could you please have another look @fasaxc ?

fasaxc (approved):

PR #7705, opened 2023-05-26
Improve pod startup time

alex-matei:
## Description

Moving an interface from one namespace to another is time consuming (especially in CANAL when trying to create multiple pods) and requires the interface to be set back to UP again.  We can improve the current implementation by creating the host veth directly in the host namespace.

We measured the time it takes in calico cni plugin to set up the networking stack, more precisely the time it takes to execute DoNetworking. We ran benchmarks measuring the time spent when creating 16 pods simultaneously on a cluster with 1 control-plane node and 1 worker node. 

<details>
<summary>System specs</summary>
Red Hat Enterprise Linux 9.1
Kernel: 6.3.0-rc3
</details>

Below are the benchmarks for various Calico setups:
<details>
<summary>CALICO IPV4</summary>

```
before:
containerd[5219]: 2023-05-26 07:58:13.018 [INFO][46450] dataplane_linux.go 319: Finished DoNetworking. ContainerID="276c70d6b1f9aaa0b012399b5c99c97ec5abf6f47ef9b7655855ed04ad515e90" Namespace="kbench-pod-namespace" Pod="job-kbench-pod-oid-0-tid-5" WorkloadEndpoint="robot--all--0a65ff80--7adb--4852--9247--5272deaa4bb7--vmss000000-k8s-job--kbench--pod--oid--0--tid--5-eth0" after=15.357635ms
containerd[5219]: 2023-05-26 07:58:13.043 [INFO][46448] dataplane_linux.go 319: Finished DoNetworking. ContainerID="6b3c632b3562577523a5acf4389f930be989aef01e1238a3ef0b29818f88a9fd" Namespace="kbench-pod-namespace" Pod="job-kbench-pod-oid-0-tid-7" WorkloadEndpoint="robot--all--0a65ff80--7adb--4852--9247--5272deaa4bb7--vmss000000-k8s-job--kbench--pod--oid--0--tid--7-eth0" after=22.190691ms
containerd[5219]: 2023-05-26 07:58:13.071 [INFO][46442] dataplane_linux.go 319: Finished DoNetworking. ContainerID="718ea8a154349f0339c349f3e8ae6cded7bd8deaf171d864ffb8e50ce87434d9" Namespace="kbench-pod-namespace" Pod="job-kbench-pod-oid-0-tid-2" WorkloadEndpoint="robot--all--0a65ff80--7adb--4852--9247--5272deaa4bb7--vmss000000-k8s-job--kbench--pod--oid--0--tid--2-eth0" after=28.392989ms
[...]

PR #7911, opened 2023-08-09
Remove outdated comments from the code that created veth

iiiceoo:
In #7705, the host veth is created directly in the host namespace, rather than being created first in the container namespace.

Signed-off-by: iiiceoo <iiiceoo@foxmail.com>


caseydavenport (approved):
Long time waiting, but LGTM

PR #7913, opened 2023-08-09
Add postrelease test promotion

danudey:
This PR will add the postrelease tests (in `hack/postrelease`) as a promotion, initially, to be run after a successful publish. Eventually it will be an automatic step.


mgleung (approved):
LGTM

PR #7922, opened 2023-08-13
Upgrade libbpf 1.2.2

sridhartigera:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/bpf/tc/attach.go
```diff
@@ -84,16 +84,20 @@ func (ap *AttachPoint) loadObject(ipVer int, file string) (*libbpf.Obj, error) {
 		// In case of global variables, libbpf creates an internal map <prog_name>.rodata
 		// The values are read only for the BPF programs, but can be set to a value from
 		// userspace before the program is loaded.
+		mapName := m.Name()
 		if m.IsMapInternal() {
+			if strings.HasPrefix(mapName, ".rodata") {
```
tomastigera, line 89: What other maps are internal now? Would this help with movind strings from code to .rodata?

felix/bpf-gpl/tc6.c
```diff
@@ -24,7 +24,7 @@
 const volatile struct cali_tc_globals __globals;
 
 
-SEC("classifier/tc/prologue")
+SEC("classifier")
```
tomastigera, line 27: Why did we need to place everything into a single section now?

sridhartigera: The way the section name is parsed has changed. In earlier versions, ```classifier/x/y``` was allowed but in the latest it is just ```classifier```.

sridhartigera: I have also changed the section name from ```classifier``` to ```tc``` as ```classifier``` is deprecated from v1.0.0

felix/bpf/tc/attach.go
sridhartigera: If you remember, we used to get unrecognized section .rodata warning. Now they have defined these as separate maps (BPF_MAP_TYPE_ARRAY). These maps along with the global data is marked as RODATA only difference being global data maps are named as <map_name>.<rodata> whereas these maps are named .rodata.1.1. Hence we need to skip these maps. I am sure about string handling. Have to check how that works.

PR #7939, opened 2023-08-17
Defer flush of the policy resolver

fasaxc:
## Description

Flushing after every match update causes high CPU usage for endpoints that are being added/removed with many matching policies.  Defer the sort and flush until we're ready to send data downstream.

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Fix slow performance in Felix calculation graph when adding/removing an endpoint with many matching policies. Policy sort order was rechecked once for each added/removed policy match rather than once only.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/calc/policy_resolver.go
+++ b/felix/calc/policy_resolver.go
@@ -111,7 +111,6 @@ func (pr *PolicyResolver) OnUpdate(update api.Update) (filterOut bool) {
 			pr.markEndpointsMatchingPolicyDirty(key)
 		}
 	}
-	pr.maybeFlush()
 	gaugeNumActivePolicies.Set(float64(pr.policyIDToEndpointIDs.Len()))
 	return
 }
@@ -119,7 +118,6 @@ func (pr *PolicyResolver) OnUpdate(update api.Update) (filterOut bool) {
 func (pr *PolicyResolver) OnDatamodelStatus(status api.SyncStatus) {
 	if status == api.InSync {
 		pr.InSync = true
-		pr.maybeFlush()
 	}
 }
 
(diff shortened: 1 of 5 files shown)


Josh-Tigera:
Was the inefficiency of the redundant flushes flagged by any tests? Would it be worthwhile either adding a benchmark test or an expectation of a canary-in-the-coal-mine function being called less than a threshold to protect against this kind of thing in the future?

fasaxc:
I have a benchmark test in a follow-on PR.  It can't go in yet because it runs too slowly without the _other_ enhancements that I have lined up!

fasaxc:
Added a basic kick-the-tires in the UT.

felix/calc/policy_resolver.go
```diff
@@ -111,15 +111,13 @@ func (pr *PolicyResolver) OnUpdate(update api.Update) (filterOut bool) {
 			pr.markEndpointsMatchingPolicyDirty(key)
 		}
 	}
-	pr.maybeFlush()
```
aaaaaaaalex, line 114: IIUC, the policyresolver now only flushes when the calcgraph does - are there any types of updates which should still be flushed more aggressively / decided explicitly by the pol-resolver / how much risk is there of the CG's flush strategy having a blind spot? More of a question than feedback :smile:

fasaxc: It's also the case that the policy resolver is at the bottom of the calc graph already so it's the last thing in its call path that emits events (so there's no chance of reordering events here, I don't think)

Josh-Tigera (approved):

PR #7954, opened 2023-08-23
BPF Dump policies without assembly replaced with more condensed rule counters output

StevenTigera:
Currently the Calico BPF utility policy dump is very verbose and includes assembly which is not relevant to most users i.e. output from the command:
```kubectl exec calico-node-xyz -n calico-system -- calico-node -bpf policy dump cali2c5f0ef34cf ingress```

Therefore, this simple requirement is to omit the irrelevant assembly and instead present more condense output including the rule counters only e.g.
```
IfaceName: cali2c5f0ef34cf
Hook: tc egress
Error: 
Policy Info:
// Start of policy default.policy-tcp
// Start of rule action:"allow" protocol:<name:"tcp" > src_net:"11.0.0.8/32" src_net:"10.0.0.8/32" src_ports:<first:8055 last:8055 > src_ports:<first:100 last:105 > dst_net:"12.0.0.8/32" dst_net:"13.0.0.8/32" dst_ports:<first:9055 last:9055 > dst_ports:<first:200 last:205 > rule_id:"STaSnQYmIymqngjx" 
// count = 0
```
However, there is an option to append either `-a` or `--asm` to the command above to include the original verbose output as before i.e.
```kubectl exec calico-node-xyz -n calico-system -- calico-node -bpf policy dump cali2c5f0ef34cf ingress -a```
OR
```kubectl exec calico-node-xyz -n calico-system -- calico-node -bpf policy dump cali2c5f0ef34cf ingress --asm```

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
calico-node -bpf dump policy will now suppress the previous verbose output by default.  
If you would like to see the verbose output then please add the --verbose flag [or -v] 
for example calico-node -bpf dump policy --verbose
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/cmd/calico-bpf/commands/policy_debug.go
+++ b/felix/cmd/calico-bpf/commands/policy_debug.go
@@ -22,14 +22,14 @@ import (
 	"strconv"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
 	"github.com/projectcalico/calico/felix/bpf"
 	"github.com/projectcalico/calico/felix/bpf/asm"
 	"github.com/projectcalico/calico/felix/bpf/counters"
 	"github.com/projectcalico/calico/felix/bpf/hook"
 	"github.com/projectcalico/calico/felix/proto"
-
-	log "github.com/sirupsen/logrus"
-	"github.com/spf13/cobra"
 )
 
 // policyCmd represents the counters command
@@ -40,8 +40,8 @@ var policyCmd = &cobra.Command{
 
 func init() {
 	policyCmd.AddCommand(policyDumpCmd)
+	policyDumpCmd.Flags().BoolP("asm", "a", false, "Includes eBPF assembler code of the policy program")
 	rootCmd.AddCommand(policyCmd)
-
 }
 
 var policyDumpCmd = &cobra.Command{
@@ -83,11 +83,13 @@ var policyDumpCmd = &cobra.Command{
 }
 
 func parseArgs(args []string) (string, string, error) {
-	if len(args) != 2 {
-		return "", "", fmt.Errorf("Insufficient arguments")
+	lenArgs := len(args)
+	if lenArgs != 2 {
+		return "", "", fmt.Errorf("Invalid number of arguments: %d", lenArgs)
 	}
-	if hook.StringToHook(args[1]) == hook.Bad && args[1] != "all" {
-		return "", "", fmt.Errorf("Invalid argument")
+	hookArg := args[1]
+	if hook.StringToHook(hookArg) == hook.Bad && hookArg != "all" {
+		return "", "", fmt.Errorf("Invalid argument: '%s'", hookArg)
 	}
 	return args[0], args[1], nil
 }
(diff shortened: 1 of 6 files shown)


felix/cmd/calico-bpf/commands/policy_debug.go
```diff
@@ -138,19 +151,34 @@ func dumpPolicyInfo(cmd *cobra.Command, iface string, h hook.Hook, m counters.Po
...
+			if strings.Contains(comment, "rule_id") {
+				rule_id = getRuleIdentfier(comment)
+			}
 			if strings.Contains(comment, "Rule MatchID") {
 				matchId := getRuleMatchID(comment)
-				cmd.Printf("// count = %d\n", m[matchId])
+				if verboseFlagSet {
+					cmd.Printf("// count = %d\n", m[matchId])
+				} else {
+					cmd.Printf("RuleID['%s'] => count = %d\n", rule_id, m[matchId])
```
tomastigera, line 166: The rule id is a very opaque internal identificator that does not convey much information. In iptables the rule IDs are somewhat readable, but unfortunately not in eBPF. This is how a sample verbose output looks like:
```
// Start of policy default.self-snat

// Start of rule action:"allow" src_net:"172.18.0.6/32" rule_id:"5vusbnpUZ-y8FtmR" 

// count = 2

// If source not in {172.18.0.6/32}, skip to next rule

rule_0_no_match:

  6191000000000000 LoadReg32 dst=R1 src=R9 off=0 imm=0x00000000/0  R1 = *(u32 *)(R9 + 0) /* state->ip_src */

  b4020000ffffffff MovImm32 dst=R2 src=R0 off=0 imm=0xffffffff/-1                                  

  5c12000000000000 And32 dst=R2 src=R1 off=0 imm=0x00000000/0                                      

  16020100ac120006 JumpEqImm32 dst=R2 src=R0 off=1 imm=0x060012ac/100668076  goto rule_1_part_0

  05000c0000000000 JumpA dst=R0 src=R0 off=12 imm=0x00000000/0               goto rule_1_no_match

rule_1_part_0:
```

I think a user would like to see the whole `Start of rule action:"allow" src_net:"172.18.0.6/32" rule_id:"5vusbnpUZ-y8FtmR" ` perhaps without the `relu_id` but with the `policy default.self-snat` so he/she can match it with the policy in a yaml.

tomastigera (commented):
I like the --verbose flag! :+1: see my comment on the output.

felix/cmd/calico-bpf/commands/policy_debug.go
tomastigera: yeah, that looks good, thanks

StevenTigera: Awesome thanks - will implement accordingly

tomastigera (commented):
LGTM, just a second thought nit ... i wonder whether --verbose would not collide with a possible global --verbose flag and whether something like `--asm | Includes eBPF assembler code of the policy program` would be better. Or perhaps `--code` :thinking:

StevenTigera:
> LGTM, just a second thought nit ... i wonder whether --verbose would not collide with a possible global --verbose flag and whether something like `--asm | Includes eBPF assembler code of the policy program` would be better. Or perhaps `--code` 🤔

Thanks @tomastigera - actually I did think about `-v` shorthand flag for verbose here may collide with global `-v` for version but I see that this was OK because for verbose as one would add the -v flag after the `policy dump` command i.e. like this 
`policy dump cali2c5f0ef34cf ingress -v` so all seemed fine - I wasn't aware of the --verbose global flag though

However, I like this suggestion to use something else - I think `--asm` is the best option because it is the most specific option for this particular command `policy dump` so I would like to change to `--asm` accordingly - thanks for the suggestion / recommendation!

StevenTigera:
Update `--verbose` flag to `--asm` flag as discussed above

tomastigera (approved):
Thanks!

Issue #7972, opened 2023-08-29
calico-kube-controllers panic (calico 3.26.1)


mchtech:
## Current Behavior

crash: concurrent map iteration and map write

## Steps to Reproduce (for bugs)

1. rollout restart deployment

## Context

```
fatal error: concurrent map iteration and map write

goroutine 371 [running]:
reflect.mapiternext(0x4cff2f?)
	/usr/local/go/src/runtime/map.go:1380 +0x19
reflect.(*MapIter).Next(0xc001a708e8?)
	/usr/local/go/src/reflect/value.go:1924 +0x7e
internal/fmtsort.Sort({0x1c44800?, 0xc01610cbe0?, 0xc0005a42e8?})
	/usr/local/go/src/internal/fmtsort/sort.go:62 +0x1f0
fmt.(*pp).printValue(0xc0073bc340, {0x1c44800?, 0xc01610cbe0?, 0x0?}, 0x76, 0x1)
	/usr/local/go/src/fmt/print.go:816 +0x986
fmt.(*pp).printValue(0xc0073bc340, {0x1db3800?, 0xc01610cbc0?, 0x30?}, 0x76, 0x0)
	/usr/local/go/src/fmt/print.go:853 +0x120a
fmt.(*pp).printArg(0xc0073bc340, {0x1db3800?, 0xc01610cbc0}, 0x76)
	/usr/local/go/src/fmt/print.go:759 +0x756
fmt.(*pp).doPrintf(0xc0073bc340, {0x1facf6b, 0x3d}, {0xc001a71108?, 0x2, 0x2})
	/usr/local/go/src/fmt/print.go:1077 +0x387
fmt.Sprintf({0x1facf6b, 0x3d}, {0xc001a71108, 0x2, 0x2})
	/usr/local/go/src/fmt/print.go:239 +0x59
github.com/sirupsen/logrus.(*Entry).Logf(0xc01660e000, 0x4, {0x1facf6b?, 0x1db3800?}, {0xc001a71108?, 0xc001a71188?, 0xc01610cbc0?})
	/go/pkg/mod/github.com/sirupsen/logrus@v1.9.0/entry.go:349 +0x49
github.com/sirupsen/logrus.(*Logger).Logf(0xc000062080, 0x4, {0x1facf6b, 0x3d}, {0xc001a71108, 0x2, 0x2})
	/go/pkg/mod/github.com/sirupsen/logrus@v1.9.0/logger.go:154 +0x85
[...]

PR #7975, opened 2023-08-30
calico-kube-controller: fix panic caused by concurrent map iteration and map write

cyclinder:
## Description

```
fatal error: concurrent map iteration and map write

goroutine 371 [running]:
reflect.mapiternext(0x4cff2f?)
	/usr/local/go/src/runtime/map.go:1380 +0x19
reflect.(*MapIter).Next(0xc001a708e8?)
	/usr/local/go/src/reflect/value.go:1924 +0x7e
internal/fmtsort.Sort({0x1c44800?, 0xc01610cbe0?, 0xc0005a42e8?})
	/usr/local/go/src/internal/fmtsort/sort.go:62 +0x1f0
fmt.(*pp).printValue(0xc0073bc340, {0x1c44800?, 0xc01610cbe0?, 0x0?}, 0x76, 0x1)
	/usr/local/go/src/fmt/print.go:816 +0x986
fmt.(*pp).printValue(0xc0073bc340, {0x1db3800?, 0xc01610cbc0?, 0x30?}, 0x76, 0x0)
	/usr/local/go/src/fmt/print.go:853 +0x120a
fmt.(*pp).printArg(0xc0073bc340, {0x1db3800?, 0xc01610cbc0}, 0x76)
	/usr/local/go/src/fmt/print.go:759 +0x756
fmt.(*pp).doPrintf(0xc0073bc340, {0x1facf6b, 0x3d}, {0xc001a71108?, 0x2, 0x2})
	/usr/local/go/src/fmt/print.go:1077 +0x387
fmt.Sprintf({0x1facf6b, 0x3d}, {0xc001a71108, 0x2, 0x2})
	/usr/local/go/src/fmt/print.go:239 +0x59
github.com/sirupsen/logrus.(*Entry).Logf(0xc01660e000, 0x4, {0x1facf6b?, 0x1db3800?}, {0xc001a71108?, 0xc001a71188?, 0xc01610cbc0?})
	/go/pkg/mod/github.com/sirupsen/logrus@v1.9.0/entry.go:349 +0x49
github.com/sirupsen/logrus.(*Logger).Logf(0xc000062080, 0x4, {0x1facf6b, 0x3d}, {0xc001a71108, 0x2, 0x2})
	/go/pkg/mod/github.com/sirupsen/logrus@v1.9.0/logger.go:154 +0x85
github.com/sirupsen/logrus.(*Logger).Infof(...)
	/go/pkg/mod/github.com/sirupsen/logrus@v1.9.0/logger.go:168
github.com/sirupsen/logrus.Infof(...)
	/go/pkg/mod/github.com/sirupsen/logrus@v1.9.0/exported.go:199
github.com/projectcalico/calico/kube-controllers/pkg/controllers/pod.(*podController).syncToCalico(0xc00060e840, {0xc00ec454d0, 0x2f})
	/go/src/github.com/projectcalico/calico/kube-controllers/pkg/controllers/pod/pod_controller.go:324 +0x90f
github.com/projectcalico/calico/kube-controllers/pkg/controllers/pod.(*podController).processNextItem(0xc00060e840)
	/go/src/github.com/projectcalico/calico/kube-controllers/pkg/controllers/pod/pod_controller.go:275 +0x70
github.com/projectcalico/calico/kube-controllers/pkg/controllers/pod.(*podController).runWorker(0x4?)
	/go/src/github.com/projectcalico/calico/kube-controllers/pkg/controllers/pod/pod_controller.go:260 +0x25
created by github.com/projectcalico/calico/kube-controllers/pkg/controllers/pod.(*podController).Run
	/go/src/github.com/projectcalico/calico/kube-controllers/pkg/controllers/pod/pod_controller.go:251 +0x3c6
```

## Related issues/PRs

https://github.com/projectcalico/calico/issues/7972

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/kube-controllers/pkg/controllers/pod/pod_controller.go
+++ b/kube-controllers/pkg/controllers/pod/pod_controller.go
@@ -321,7 +321,9 @@ func (c *podController) syncToCalico(key string) error {
 		new := wepData.(converter.WorkloadEndpointData)
 		if !reflect.DeepEqual(old, new) {
 			// The relevant wep data has changed - update the wep and write it to the datastore.
+			c.workloadEndpointCache.RLock()
 			log.Infof("Writing endpoint %s with updated data %#v to Calico datastore", key, new)
+			c.workloadEndpointCache.RUnlock()
 			converter.MergeWorkloadEndpointData(&wep, new)
 			_, err := c.calicoClient.WorkloadEndpoints().Update(c.ctx, &wep, options.SetOptions{})
 			if err != nil {


kube-controllers/pkg/controllers/pod/pod_controller.go
```diff
@@ -321,7 +321,9 @@ func (c *podController) syncToCalico(key string) error {
 		new := wepData.(converter.WorkloadEndpointData)
 		if !reflect.DeepEqual(old, new) {
 			// The relevant wep data has changed - update the wep and write it to the datastore.
+			c.workloadEndpointCache.RLock()
 			log.Infof("Writing endpoint %s with updated data %#v to Calico datastore", key, new)
```
fasaxc, line 325: If we need a lock to print out the `new` variable then the `reflect.DeepEqual()` on line 324 is also suspicious.  Looks like this whole are of code needs review for concurrency issues.

cyclinder: I think we should lock the `key` variable, and I noticed that elsewhere in this function, whenever this variable is read, the `RLock` function is called.

caseydavenport: I don't think the `key` variable is the problem here

The other lock instances aren't protecting `key` - they're protecting the `workloadEndpointCache` which is accessed from multiple routines.

caseydavenport: @cyclinder @fasaxc I suspect that the "concurrent map access" issue we're seeing here is because "new" is an instance of `WorkloadEndpointData` which has a map as a field: https://github.com/projectcalico/calico/blob/77ecdcda90c0a8a38ae247067169523484c6c8e3/kube-controllers/pkg/converter/pod_converter.go#L37

Likely we're hitting this line at the same time that we are updating that Label map elsewhere? 

If so, I think we definitely need a lock around the `DeepEqual` above as well.

This gets a bit tricky, because we need to make sure we're locking all of the relevant places that this map may be updated. The `workloadEndpointCache` lock is probably one of them, but there is also the `resourceCache` to think about.

It might make sense to make a copy of the labels when building the `WorkloadEndpointData` so that updates to the pod aren't acting on the same map as the wepData here? Would that allow us to simplify this? https://github.com/projectcalico/calico/blob/77ecdcda90c0a8a38ae247067169523484c6c8e3/kube-controllers/pkg/converter/pod_converter.go#L52-L64

fasaxc: I've lost the thread a bit, but  any chance this fix would help: https://github.com/projectcalico/calico/pull/9039.  If we got the WEPs above from the converter we'll have copied the labels already after my fix.

The bug I was fixing was that we were mutating the map that we got from the caching informer, but it also mutates the map when refreshing the cache.

caseydavenport: Yeah, think there's a good chance that fixed this already :+1:

PR #7979, opened 2023-08-31
Typha tolerate node.cloudprovider.kubernetes.io/uninitialized taint

ffilippopoulos:
## Description

Allow typha deployment to tolerate the taint set via kubelet when used with --cloud-provider=external flag. We need typha to be able to schedule as soon as possible on new nodes since calico-node relies on it.

## Related issues/PRs

Related Slack thread:
https://calicousers.slack.com/archives/C0BCA117T/p1693469918340899

## Todos

Someone should run the respective make target to generate manifests picking up this change.
## Release Note

```release-note
Add Typha Deployment tolerations to helm charts so that it can be scheduled on any nodes.
```

### DIFF
--- a/charts/calico/templates/calico-typha.yaml
+++ b/charts/calico/templates/calico-typha.yaml
@@ -71,11 +71,11 @@ spec:
         # Mark the pod as a critical add-on for rescheduling.
         - key: CriticalAddonsOnly
           operator: Exists
-        # this taint is set by all kubelets running `--cloud-provider=external`
-        # so we should tolerate it to schedule typha pods
-        - key: node.cloudprovider.kubernetes.io/uninitialized
-          value: "true"
-          effect: NoSchedule
+        # Make sure Typha can get scheduled on any nodes. 
+        - effect: NoSchedule 
+          operator: Exists 
+        - effect: NoExecute 
+          operator: Exists           
       # Since Calico can't network a pod until Typha is up, we need to run Typha itself
       # as a host-networked pod.
       serviceAccountName: calico-node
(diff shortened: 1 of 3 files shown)


lwr20:
I think Typha and calico-node should have the exact _same_ Tolerations since calico-node requires Typha to be up to come up itself.

Can a core-maintainer confirm that position please?

charts/calico/templates/calico-typha.yaml
```diff
@@ -71,6 +71,11 @@ spec:
         # Mark the pod as a critical add-on for rescheduling.
         - key: CriticalAddonsOnly
           operator: Exists
+        # this taint is set by all kubelets running `--cloud-provider=external`
+        # so we should tolerate it to schedule typha pods
+        - key: node.cloudprovider.kubernetes.io/uninitialized
+          value: "true"
+          effect: NoSchedule
```
lwr20, line 78: Perhaps just copy the Tolerations from the calico-node pod (since calico-node cannot go ready until typha comes up): https://github.com/projectcalico/calico/blob/ddbc8e8141c811422cb4bcf8af177b99ac5fb82e/charts/calico/templates/calico-node.yaml#L31-L39

lwr20:
```suggestion
         # Make sure Typha can get scheduled on any nodes. 
         - effect: NoSchedule 
           operator: Exists 
         - effect: NoExecute 
           operator: Exists           
```

coutinhop: +1 on this, it seems like we already do that for typha when using operator: https://github.com/tigera/operator/blob/master/pkg/render/typha.go#L400

caseydavenport:
@lwr20 @coutinhop anything else we need from this prior to merge?

The one thing I can see is that tests are failing - @ffilippopoulos you need to run `make generate` to update generated files.

coutinhop:
> @lwr20 @coutinhop anything else we need from this prior to merge?

I don't think so, @lwr20 please correct me if I'm wrong.

PR #7980, opened 2023-08-31
BPF Remove old versions of maps after upgrade

StevenTigera:
UPDATE: closing PR and creating new PR [8012](https://github.com/projectcalico/calico/pull/8012)

_Expose a `deletePreviousVersion()` function to maps package which allows client code to explicitly delete previous version of a map after upgrade.  Proposal to expose this as a separate function to allow for more control in the code flow when the deletion can take place as there are many examples of map manipulation in [map_upgrade_test.go](https://github.com/projectcalico/calico/blob/master/felix/bpf/ut/map_upgrade_test.go) such that if the deletion was implicitly invoked in the private `upgrade()` function then this could break existing code and tests._

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: Remove old versions of maps after upgrade
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/maps/maps.go
+++ b/felix/bpf/maps/maps.go
@@ -132,6 +132,7 @@ type Map interface {
 	Update(k, v []byte) error
 	Get(k []byte) ([]byte, error)
 	Delete(k []byte) error
+	DeletePreviousVersion() error
 }
 
 type MapWithExistsCheck interface {
@@ -419,11 +420,11 @@ func (b *PinnedMap) Delete(k []byte) error {
 	return DeleteMapEntry(b.fd, k)
 }
 
-func (b *PinnedMap) deletePreviousVersion() error {
+func (b *PinnedMap) DeletePreviousVersion() error {
 	log.WithField("name", b.Name).Debug("delete previous version")
 	oldVersion, err := b.getOldMapVersion()
 	log.WithError(err).Debugf("Upgrading from %d", oldVersion)
-	if err != nil {
+	if err != nil && !IsNotExists(err) {
 		return err
 	}
 	// fresh install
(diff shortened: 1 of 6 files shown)


felix/bpf/mock/map.go
```diff
@@ -165,6 +165,10 @@ func (m *Map) CopyDeltaFromOldMap() error {
 	return nil
 }
 
+func (m *Map) DeletePreviousVersion() error {
```
tomastigera, line 168: did you need to export it if you do not use it outside the maps package?

felix/bpf/maps/maps.go
```diff
@@ -757,6 +785,12 @@ func (b *PinnedMap) CopyDeltaFromOldMap() error {
 	if err != nil {
 		return fmt.Errorf("error upgrading data from old map %s, err=%w", b.GetName(), err)
 	}
+
+	err = b.DeletePreviousVersion()
+	if err != nil {
+		return fmt.Errorf("failed to delete previous %s map, err=%w", b.Name, err)
```
tomastigera, line 791: I think we should check whether the failure was because the map did not exist essentially getting ENOENT. In that case, that is not quite an error as we are in the state where we want to be and thus we can continue. 

Note that this function is called on for CT map https://github.com/projectcalico/calico/blob/2ebc1f92ecc39332cf1d55ba676d9101af24982f/felix/dataplane/linux/bpf_ep_mgr.go#L1341

But we need to make sure that any old map is deleted. For instance, state maps are not being upgraded as they are internal to the version of the programs that use them, but they are pinned so when we upgrade, they may be left over from the previous versions.

CT is sort of special, because we need to move the state from one version to the next verion so that we do not break connections. That is not true for most of the other maps.

tomastigera (changes requested):

felix/bpf/mock/map.go
StevenTigera: Oh great spot thanks @tomastigera - No, made deletePreviousVersion() private accordingly

felix/bpf/maps/maps.go
StevenTigera: Thanks @tomastigera - actually I see a neat helper function `IsNotExists()` which wraps this check for ENOENT so I have implemented as per comment following an existing example for consistency

tomastigera: but as I said :arrow_up: this only fixes the issue for CT map. There are many more. So we need to tke it out of the copy-deltas and add a test that also mkes sure that other aps are removed. The per-cpu state map ay be a good example.

StevenTigera: All of the above address in PR 8012 - therefore will close this PR now - thanks

PR #7997, opened 2023-09-07
fix: update calico-cni-plugin clusterrole rbac to include node/status patch and update

apsega:
update calico-cni-plugin clusterrole rbac to include node/status patch and update. Currently there's issue with calicoctl when running command `calicoctl patch node` to patch ASN in order to enable bgp peering.
```
[execute] Hit error: updating existing resource: connection is unauthorized: nodes "<node-name>" is forbidden: User "system:serviceaccount:kube-system:calico-cni-plugin" cannot update resource "nodes/status" in API group "" at the cluster scope
```

we're using `calico-typha.yaml` manifest, but fixing in other manifests, too.

### DIFF
--- a/charts/calico/templates/calico-node-rbac.yaml
+++ b/charts/calico/templates/calico-node-rbac.yaml
@@ -191,6 +191,14 @@ rules:
       - namespaces
     verbs:
       - get
+  - apiGroups: [""]
+    resources:
+      - nodes/status
+    verbs:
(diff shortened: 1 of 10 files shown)


caseydavenport:
The CNI plugin itself shouldn't need these permissions, so we shouldn't add them to the `calico-cni-plugin` role. 

Is `calicoctl` using the CNI plugin serviceaccount for some reason? If so, a proper fix is instead to have a separate set of permissions for `calicoctl`.

PR #8006, opened 2023-09-08
Fix docker image folder permissions

Tolsto:
Under some circumstances, the apiserver can and
should be run as non-root. Previously, the docker
image did not permit this because folders /code
and /tmp were solely writable by root.

In the existing Dockerfile, it's indicated that
both directories should be writable for user 1001. However, this logic is flawed due to:

1) The UBI running as root, resulting in permissions
   after mkdir at drwxr-xr-x owned by root:root
2) The COPY function not honoring folder permissions,
   even with --chmod, which again leads to
   drwxr-xr-x owned by root:root

To resolve this, we employed the fact that ADD
extracts .tar files while preserving the permissions. Now, the file structure is:

/code drwxrwxrwx
|-- apiserver -rwxr-xr-x
|-- filecheck -rwxr-xr-x
/tmp drwxrwxrwx

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
@Tolsto thanks for the PR. One thing I'll say is that I don't think we should be checking in a .tar file to the repository. 

I am a little bit surprised this can't be done as part of the Dockerfile itself, but only a little.

Tolsto:
@caseydavenport What are your objections with a .tar file? I've copied this approach from [Google's Distroless image](https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md).

caseydavenport:
I don't like the idea of checking in binary files that can't be reviewed as part of our standard PR process. It's an opaque blob of data that I don't trust.

lwr20:
I guess creating the tarfile could be done as commands in part of a make target instead.  The commands are then reviewable, and we don't have a binary blob in the repo.

caseydavenport:
I believe this commit

PR #8012, opened 2023-09-13
[DO NOT MERGE] BPF Remove old versions of maps after upgrade

StevenTigera:
Expose a `DeletePreviousVersion()` function to maps package which allows client code to explicitly delete previous version of a map after upgrade.  Proposal to expose this as a separate function to allow for more control in the code flow when the deletion can take place as there are many examples of map manipulation in [map_upgrade_test.go](https://github.com/projectcalico/calico/blob/master/felix/bpf/ut/map_upgrade_test.go) such that if the deletion was implicitly invoked in the private `upgrade()` function then this could break existing code and tests.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: Remove old versions of maps after upgrade
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/bpfmap/bpf_maps.go
+++ b/felix/bpf/bpfmap/bpf_maps.go
@@ -55,25 +55,20 @@ type Maps struct {
 	XDPJumpMap      maps.MapWithDeleteIfExists
 }
 
-func (m *Maps) Destroy() {
-	mps := []maps.Map{
-		m.IpsetsMap,
-		m.StateMap,
-		m.ArpMap,
-		m.FailsafesMap,
-		m.FrontendMap,
-		m.BackendMap,
-		m.AffinityMap,
-		m.RouteMap,
-		m.CtMap,
-		m.SrMsgMap,
-		m.CtNatsMap,
-		m.ProgramsMap,
-		m.JumpMap,
-		m.XDPProgramsMap,
-		m.XDPJumpMap,
+func (m *Maps) DeletePreviousVersion() error {
+	mps := m.extractMapSlice()
+	for _, m := range mps {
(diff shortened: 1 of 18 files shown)


felix/bpf/bpfmap/bpf_maps.go
```diff
@@ -55,25 +55,20 @@ type Maps struct {
...
-		m.JumpMap,
-		m.XDPProgramsMap,
-		m.XDPJumpMap,
+func (m *Maps) DeletePreviousVersion() error {
+	mps := m.extractMapSlice()
+	for _, m := range mps {
+		err := m.(pinnedMap).DeletePreviousVersion()
+		if err != nil {
+			return err
+		}
```
tomastigera, line 64: I was kinds hoping for something simpler that would also remove pins if we skipped a version, which would make the cleanup a bit more robust without adding this much boilerplate code to the maps. Unfortunately I think we also mix some program pins in this dir. But perhaps not and that would make the code simpler.

```suggestion
expectedPins[m.GetName()] = true
filepath.WalkDir(tcdefs.GlobalPinDir, func(f) {
    if f not in expectedPins and f not prefix of anything in expectedPins {
       delete f
    }
})
```

StevenTigera: Update: actually further grok'ing in the private helper method [getOldMapVersion()](https://github.com/projectcalico/calico/blob/master/felix/bpf/maps/maps.go#L778)  seems to assume that any previous version of a pinned map will be located in the same  `pinDir()` of the current version from this LOC:
```
files, err := os.ReadDir(b.pinDir())
```
which may also not be reliable / robust.

PR #8026, opened 2023-09-18
Correct policy for OpenStack sec group with no remote_ip_prefix

### LINKED ISSUE #7968: Not specifying remote_ip_prefix creates a policy that unexpectedly opens all ports
Related to the openstack networking_calico ml2 driver @fasaxc @nelljerram 

There is a bug that if remote_ip_prefix is not specified when a sec group rule is created using openstack API, the resulting policy in etcd will have no port, and no network prefix specified, thus suprisingly opens all ports(!). Which is not what you'd want to happen.

Relevant API call https://docs.openstack.org/api-ref/network/v2/index.html?expanded=create-security-group-rule-detail#create-security-group-rule

debug line of working rule generated /with/ remote_ip_prefix specified to the openstack API call:
`
2023-07-14 10:10:43.908 neutron 6838 DEBUG networking_calico.plugins.ml2.drivers.calico.policy [req-52944517-c7b1-4f61-a992-58d0510da207 9956b53cf1ca4967a7a945b4e6657cf6 74cf3e20e55345d29935625c7b3e5618 - 2926c4adca764d1a9107f5eba40681d7 b796718245bb4172a8acd73ebe510e21] => ingress Calico rule {'action': 'Allow', 'ipVersion': 4, 'destination': {'ports': ['25:34']}, 'protocol': 6, 'source': {'nets': [u'0.0.0.0/0']}} _neutron_rule_to_etcd_rule networking_calico/plugins/ml2/drivers/calico/policy.py:223
`

debug line of rule generated /without/ remote_ip_prefix specified, other data is the exact same openstack API call - the port range and nets information is now missing, resulting in a rule that _opens all ports_.
`
2023-07-14T10:28:51.216000Z debug xxx neutron[31686]networking_calico.plugins.ml2.drivers.calico.policy [req-e179ed2f-0263-4a96-9b32-4e0a4d10df53 9956b53cf1ca4967a7a945b4e6657cf6 74cf3e20e55345d29935625c7b3e5618 - 2926c4adca764d1a9107f5eba40681d7 b796718245bb4172a8acd73ebe510e21] => ingress Calico rule {'action': 'Allow', 'ipVersion': 4, 'protocol': 6} _neutron_rule_to_etcd_rule networking_calico/plugins/ml2/drivers/calico/policy.py:223
`

@fasaxc wrote:

https://github.com/projectcalico/calico/blob/master/networking-calico/networking_calico/plugins/ml2/drivers/calico/policy.py#L213

We check if the entity_rule is empty before adding the ports to it. That looks like a mistake, @nelljerram any idea if it was intentional?

nelljerram:
Fixes #7968

When an OpenStack security group does not specify a remote_ip_prefix - in other words, it applies to traffic from all possible sources - but it does specify ports - in other words, it only applies to those ports, the resulting Calico policy is _missing_ the restriction to the intended ports, and so is accidentally an allow policy for _all_ ports.

## Release Note

```release-note
Correct policy for OpenStack sec group with no remote_ip_prefix
```

### DIFF
--- a/networking-calico/networking_calico/plugins/ml2/drivers/calico/policy.py
+++ b/networking-calico/networking_calico/plugins/ml2/drivers/calico/policy.py
@@ -209,15 +209,17 @@ def _neutron_rule_to_etcd_rule(rule):
         entity_rule['nets'] = [rule['remote_ip_prefix']]
     LOG.debug("=> Entity rule %s" % entity_rule)
 
+    if port_spec is not None:
+        if rule['direction'] == 'ingress':
+            etcd_rule['destination'] = {'ports': port_spec}
+        else:
+            entity_rule['ports'] = port_spec
+
     # Store in source or destination field of the overall rule.
     if entity_rule:
         if rule['direction'] == 'ingress':
             etcd_rule['source'] = entity_rule
-            if port_spec is not None:
-                etcd_rule['destination'] = {'ports': port_spec}
         else:
-            if port_spec is not None:
-                entity_rule['ports'] = port_spec
             etcd_rule['destination'] = entity_rule
 
     LOG.debug("=> %s Calico rule %s" % (rule['direction'], etcd_rule))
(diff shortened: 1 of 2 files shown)


nelljerram:
The remaining CI issue is tracked at https://tigera.atlassian.net/browse/CORE-9844

PR #8035, opened 2023-09-19
OpenStack port spec issue backport to v3.25

nelljerram:
## Description

Pick #8026 to the release-v3.25 branch.
Plus copy a lot of other recent tweaks and fixes from master that are needed for UT CI to run correctly and successfully.

## Release Note

```release-note
Correct policy for OpenStack sec group with no remote_ip_prefix
```


nelljerram:
The devstack test now fails because of https://tigera.atlassian.net/browse/CORE-9844

PR #8053, opened 2023-09-26
Make iptables recheck timer dynamic

fasaxc:
## Description

Apply dynamic minimum to iptables recheck timer.
- Track the peak value of iptables save/restore with an exponential decay estimator.
- Set the minimum post-retry backoff to twice the combined save/restore time.

Increase default timeouts.  These were set aggressively low in the past to deal with other apps clobbering iptables state.  that problem no longer occurs since the iptables lock was widely adopted.  

## Related issues/PRs

CORE-9786
CORE-9780

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Felix now adjusts the iptables post-write recheck interval dynamically: if iptables operations are taking longer, it will increase the time between iptables-save calls to compensate.  The default recheck time and resync intervals have also been increased.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/iptables/table.go
+++ b/felix/iptables/table.go
@@ -252,6 +252,12 @@ type Table struct {
 	postWriteInterval        time.Duration
 	refreshInterval          time.Duration
 
+	// Estimates for the time taken to do an iptables save and restore.
+	// When a save/restore exceeds the one of these we update them immediately.
+	// When a save/restore takes less time we decay them exponentially.
+	peakIptablesSaveTime    time.Duration
+	peakIptablesRestoreTime time.Duration
+
 	// calicoXtablesLock, if enabled, our implementation of the xtables lock.
 	calicoXtablesLock sync.Locker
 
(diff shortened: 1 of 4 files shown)


felix/iptables/table.go
```diff
@@ -780,6 +786,17 @@ func (t *Table) getHashesAndRulesFromDataplane() (hashes map[string][]string, ru
 // attemptToGetHashesAndRulesFromDataplane starts an iptables-save subprocess and feeds its output to
 // readHashesAndRulesFrom() via a pipe.  It handles the various error cases.
 func (t *Table) attemptToGetHashesAndRulesFromDataplane() (hashes map[string][]string, rules map[string][]string, err error) {
+	startTime := t.timeNow()
+	defer func() {
+		saveDuration := t.timeNow().Sub(startTime)
+		t.peakIptablesSaveTime = t.peakIptablesSaveTime * 99 / 100
+		if saveDuration > t.peakIptablesSaveTime {
```
aaaaaaaalex, line 793: Not that I think it matters too much, but is there a reason to decay `t.peakIptablesSaveTime` _before_ checking if the `saveDuration` was greater or lesser?

felix/iptables/table.go
```diff
@@ -780,6 +786,17 @@ func (t *Table) getHashesAndRulesFromDataplane() (hashes map[string][]string, ru
 // attemptToGetHashesAndRulesFromDataplane starts an iptables-save subprocess and feeds its output to
 // readHashesAndRulesFrom() via a pipe.  It handles the various error cases.
 func (t *Table) attemptToGetHashesAndRulesFromDataplane() (hashes map[string][]string, rules map[string][]string, err error) {
+	startTime := t.timeNow()
+	defer func() {
+		saveDuration := t.timeNow().Sub(startTime)
+		t.peakIptablesSaveTime = t.peakIptablesSaveTime * 99 / 100
```
aaaaaaaalex, line 792: Is there any lower-bound to this value, or will it (slowly) tend towards lower and lower values forever?

fasaxc: It'll get lower and lower, until we see a higher value.  The idea is that, if the time is the same every time, the peak will reflect that, but if it's spiky, then the peak will capture the spike and then decay slowly (but it will eventually decay if there are no more spikes).

mazdakn (approved):
LGTM. I am not sure I am the right person for reviewing this PR as I have not worked much with iptables code base. But at least I understand the change.

fasaxc:
Thanks @mazdakn ; for a small change like this, I think it's OK to have a non-expert review, just looking for general coding mistakes (and thanks to @aaaaaaaalex for taking a look too).

Issue #7892, opened 2023-07-31
Calico eBPF fails to init on Talos Linux


monoxane:
I'm provisioning a cluster using the Talos Linux + Kube distro and am finding that the `calico-node` `mount-ebpffs` container fails to mount the cgroup2 file system as called from [calico/node/pkg/nodeinit/calico-init_linux.go](https://github.com/projectcalico/calico/blob/661171e2062b3bbb9c6ae56a5ecba9146ccbb027/node/pkg/nodeinit/calico-init_linux.go#L128).

```
W0731 07:12:21.604403       1 feature_gate.go:241] Setting GA feature gate ServiceInternalTrafficPolicy=true. It will be removed in a future release.
2023-07-31 07:12:21.607 [INFO][1] init/startup.go 432: Early log level set to info
2023-07-31 07:12:21.607 [INFO][1] init/calico-init_linux.go 57: Checking if BPF filesystem is mounted.
2023-07-31 07:12:21.607 [INFO][1] init/calico-init_linux.go 69: BPF filesystem is mounted.
2023-07-31 07:12:21.607 [INFO][1] init/calico-init_linux.go 92: Checking if cgroup2 filesystem is mounted.
2023-07-31 07:12:21.609 [INFO][1] init/calico-init_linux.go 120: Cgroup2 filesystem is not mounted. Trying to mount it...
2023-07-31 07:12:21.609 [INFO][1] init/calico-init_linux.go 126: Mount point /run/calico/cgroup is ready for mounting root cgroup2 fs.
2023-07-31 07:12:21.613 [ERROR][1] init/calico-init_linux.go 48: Failed to mount cgroup2 filesystem. error=failed to mount cgroup2 filesystem: exit status 1
```

## Expected Behavior
Calico with eBPF dataplane works on Talos

## Current Behavior
Calico with eBPF dataplane does not work on Talos due to an FS mount failure in the eBPF mount init container

## Possible Solution
I am currently under the impression this is because `bpfdefs.CgroupV2Path` is `/run/calico/cgroup` which seems to be a non-writable directory under Talos (the vast majority of rootfs is readonly with the exception of specific files and the entirety of `/var`), but mounting an `emptyDir` at that location in both the init and the main pod does not provide any improvement. 

[...]

PR #8085, opened 2023-10-05
Allow configuring cgroup2 path

amrut-asm:
## Description
Currently, the mount-bpffs init container tries to create the folder /run/calico/cgroup for mounting the cgroupv2 fs

This causes issues on distros like Talos Linux as described in the issue https://github.com/projectcalico/calico/issues/7892

This can be a configurable path instead

## Related issues/PRs
https://github.com/projectcalico/calico/issues/7892

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: alternative cgroup2 mount path can be specified by setting CALICO_CGROUP_PATH evn var for node.
```

## Reminder for the reviewer

Other optional labels:


tomastigera:
@amrut-asm the change lgtm. Would you also consider on [operator](https://github.com/tigera/operator) change so that Talos users are not limited to using manifest installation?

amrut-asm:
> @amrut-asm the change lgtm. Would you also consider an [operator](https://github.com/tigera/operator) change so that Talos users are not limited to using manifest installation?

Sure

amrut-asm:
@tomastigera  The relevant change in operator https://github.com/tigera/operator/pull/2919

felix/bpf/bpfdefs/defs.go
```diff
@@ -14,10 +14,20 @@
...
-	DefaultBPFfsPath = "/sys/fs/bpf"
-	CgroupV2Path     = "/run/calico/cgroup"
+	DefaultBPFfsPath    = "/sys/fs/bpf"
+	DefaultCgroupV2Path = "/run/calico/cgroup"
 
 	GlobalPinDir = DefaultBPFfsPath + "/tc/globals/"
 	ObjectDir    = "/usr/lib/calico/bpf"
 )
+
+func GetCgroupV2Path() string {
```
caseydavenport, line 27: @tomastigera I think we will want to expose this using a proper FelixConfiguration parameter rather than an environment variable, no? Is there precedence for this in the BPF data plane code?

tomastigera: I think it would be tricky as this needs to get into an init container.

caseydavenport: Ah ok, this code isn't executed in felix proper?

It could in theory still be done via FelixConfiguration (in addition to this env var on the init container) - the operator already reads FelixConfiguration and uses it to configure various components. The flow would look like this:

- Set this field in FelixConfiguration
- tigera/operator reads FelixConfiguration, sets and env var for this container based on the value it finds

That would mean if we ever do need this value inside felix proper, it would already be available and we wouldn't need two separate configuration options. But I am not well positioned to say how likely that is.

tomastigera: That sounds like a great suggestion to me :+1:

amrut-asm: Thanks, I will give it a shot

tomastigera (approved):
lgtm, thank you for the contribution. Lets merge this one and let figure out the operator thing separately.

tomastigera:
Only one unrelated failure/flake in BPF tests

PR #8091, opened 2023-10-06
Add debug pprof server to Felix/Typha

fasaxc:
## Description

- Add the stdlib pprof sserver to felix, for easy collection of CPU/memory/etc profiles.
- Since the pprof server auto-registers with the default mux, rework other uses of ListenAndServe to use a unique mux.
- Expose via felix config, disabled by default.

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Felix and Typha now support enabling the Go standard library's debug server via the DebugHost/DebugPort configuration options. This allows process profiling data to be collected more easily.
```

## Reminder for the reviewer

Other optional labels:


aaaaaaaalex:
Does our policysync/dikastes/protobuf code clash muxers?

felix/daemon/daemon.go
```diff
@@ -408,6 +406,10 @@ configRetry:
 		simulateDataRace()
 	}
 
+	if configParams.DebugPort != 0 {
+		debugserver.StartDebugPprofServer(configParams.DebugHost, configParams.DebugPort)
```
tomastigera, line 410: what if debughost is something bogus? Should it log an error here like "we did not start the debug server because ..." ?

libcalico-go/lib/debugserver/debug_server.go
```diff
@@ -0,0 +1,41 @@
...
+	log.Infof("Insecure debug port is enabled on %s:%d.", host, port)
+	_ = httppprof.Profile // Make sure we don't accidentally lose the import.
+	go func() {
+		addr := net.JoinHostPort(host, strconv.Itoa(port))
+		for {
+			log.Infof("Attempting to open debug port %s:%d", host, port)
+			err := http.ListenAndServe(addr, nil)
+			if err != nil {
+				log.WithError(err).Error("Debug port HTTP server failed.  Will retry...")
+				time.Sleep(time.Second)
```
tomastigera, line 37: Isn't it better to fail? Is it likely that listenandserve would fail during operation and it would make sense to restart it? Or is it more likely that it will  keep on failing with the same error? Or is it to work around reusing a port?

tomastigera (commented):
looks good, some trivial nit, but given how pprof starts, could be ignored

libcalico-go/lib/debugserver/debug_server.go
fasaxc: Think I copied this from prometheus, where we mainly do it to deal with port reuse.   Old Felix shutting down, we're starting up, seems best to retry.  If user configured a bogus hostname then at least they'll get errors in the logs!

felix/daemon/daemon.go
fasaxc: I don't think we know until we call ListenAndServe, which is on the background goroutine

fasaxc:
> Does our policysync/dikastes/protobuf code clash muxers?

@aaaaaaaalex Good thought.  No, I don't think so, it explicitly creates a GRPC server just for that socket.

tomastigera (approved):

PR #8101, opened 2023-10-10
BPF: fix that no IP set filter was set at start-of-day.

fasaxc:
## Description

In BPF mode, we don't want to program linux IP sets unless we're using "untracked" policy, which is partially implemented in iptables.  The filter to prevent IP set programming was only set after a policy churn.

- Defer calculation of the filter until CompleteDeferredWork.
- Make sure the "dirty" flag is set at start of day so that the programming triggers before the first dataplane apply().

## Related issues/PRs

CORE-9898

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/policy_mgr.go
+++ b/felix/dataplane/linux/policy_mgr.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2021 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2023 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -29,14 +29,15 @@ import (
 // policyManager simply renders policy/profile updates into iptables.Chain objects and sends
 // them to the dataplane layer.
 type policyManager struct {
-	rawTable       IptablesTable
-	mangleTable    IptablesTable
-	filterTable    IptablesTable
-	ruleRenderer   policyRenderer
-	ipVersion      uint8
-	rawEgressOnly  bool
-	neededIPSets   map[proto.PolicyID]set.Set[string]
-	ipSetsCallback func(neededIPSets set.Set[string])
+	rawTable         IptablesTable
+	mangleTable      IptablesTable
+	filterTable      IptablesTable
+	ruleRenderer     policyRenderer
+	ipVersion        uint8
+	rawEgressOnly    bool
+	ipSetFilterDirty bool // Only used in "raw only" mode.
+	neededIPSets     map[proto.PolicyID]set.Set[string]
+	ipSetsCallback   func(neededIPSets set.Set[string])
 }
 
 type policyRenderer interface {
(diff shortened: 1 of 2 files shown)


felix/dataplane/linux/policy_mgr.go
```diff
@@ -144,7 +122,46 @@ func (m *policyManager) OnUpdate(msg interface{}) {
...
+	m.ipSetFilterDirty = false
+
+	merged := set.New[string]()
+	for _, ipSets := range m.neededIPSets {
+		ipSets.Iter(func(item string) error {
+			merged.Add(item)
+			return nil
+		})
+	}
+	m.ipSetsCallback(merged)
```
tomastigera, line 165: so in bpf mode, we get here in the first call to CompleteDeferredWork because m.ipSetFilterDirty is true from the start, but after all the merging, there is nothing to program so we do not endup programming anything? While previously things got programmed in OnUpdate rightaway so we endup with say empty sets??

> In BPF mode, we don't want to program linux IP sets unless we're using "untracked" policy, which is partially implemented in iptables. The filter to prevent IP set programming was only set after a policy churn.

I am not sure how it used to be set "after a policy churn"

fasaxc: > but after all the merging, there is nothing to program so we do not endup programming anything?

Yes, if there are no do-not-track policies then `neededIPSets` will be empty, so `merged` will be empty and then we'll call ipSetsCallback with that empty set.  That tells the IP sets driver 
not to program any IP sets.

> I am not sure how it used to be set "after a policy churn"

Before, we never called ipSetsCallback unless:

- There was a do-not-track policy (which is rare).
- We received an ActivePolicyRemove message (i.e. _some_ policy was removed from the dataplane).  Then we'd do the same merge process and set an empty filter too.

fasaxc: Moving it to the CompleteDeferredWork is just good practice (to avoid recalculating the filter over and over) and it lets me make sure that it always gets done before the first apply.

marvin-tigera:

felix/dataplane/linux/policy_mgr.go
tomastigera: So IPset are added to the `IPSets` in ipsets.go but the policy manager neveer got a chance to say "we actually need only those"? Right? So now we always call it in defered because we start with dirty filter.

felix/dataplane/linux/policy_mgr.go
```diff
@@ -144,7 +122,46 @@ func (m *policyManager) OnUpdate(msg interface{}) {
...
+		m.neededIPSets[*id] = neededIPSets
+	} else {
+		delete(m.neededIPSets, *id)
+	}
+	m.ipSetFilterDirty = true
+}
+
 func (m *policyManager) CompleteDeferredWork() error {
-	// Nothing to do, we don't defer any work.
+	if !m.rawEgressOnly {
```
tomastigera, line 150: why do we have the `rawEgressOnly` if it is ever only set to true at the time of creation?

felix/dataplane/linux/policy_mgr.go
fasaxc: Yes, exactly.

felix/dataplane/linux/policy_mgr.go
fasaxc: That's the one that controls what mode the policy manager is in.  Either it's in BPF mode (filtering IP sets) or its in iptables mode.

We don't want to call the callback if we're in iptables mode.  We always want all IP sets in that case.

marvin-tigera:

felix/dataplane/linux/policy_mgr.go
tomastigera: Ohhh, there are 2 new() got it!

tomastigera (approved):

Issue #8051, opened 2023-09-25
Include RPM metadata in images built from UBI


wedaly:
## Expected Behavior

Some security scanners like trivy rely on deb/rpm package metadata. Images built from registry.access.redhat.com/ubi8/ubi-minimal should copy the RPM package metadata so security scanners report accurate results.

## Current Behavior

The Calico images do not copy RPM metadata from the base image, so some security scanners do not accurately report CVEs.

## Possible Solution

This is how distroless-iptables copies the deb pkg metadata:
https://github.com/kubernetes/release/blob/dbe1d9b0f7828726d69faacb2da4a832b66803c4/images/build/distroless-iptables/distroless/package-utils.sh#L45-L54

Suggest that Calico does something similar for RPM packages.

## Steps to Reproduce (for bugs)
```
docker run aquasec/trivy image registry.access.redhat.com/ubi8/ubi-minimal:8.7
docker run aquasec/trivy image calico/typha:v3.24.6
```
The second command should report the vulnerabilities from the base image, but it currently doesn't.

## Context
AKS has gotten reports from customers about CVEs in Calico images that do not appear in trivy results.

I believe the same issue occurs for the tigera/operator image.

## Your Environment
* Calico version v3.24.6
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes
* Operating System and version: Linux
* Link to your project (optional): AKS

wedaly:
related: https://github.com/projectcalico/calico/issues/8050

Behnam-Shobiri:
@wedaly Thanks for reporting this.
I noticed that the Calico version is old (3.24), as mentioned in https://github.com/projectcalico/calico/issues/8050#issuecomment-1745350468 both the latest patches of 3.25 and 3.26 are using UBI 8.8. We strongly encourage everyone to use the latest release to get the security updates.

Behnam-Shobiri:
@wedaly for 3.24, the PR that updates the UBI in this branch was also just merged: https://github.com/projectcalico/calico/pull/7663 
So the next release should include the fix.

PR #8119, opened 2023-10-13
Include OS release and package metadata in Typha images

duffney:
## Description

Without operating system and package details in the Docker image after starting with a FROM scratch statement in the Dockerfile, security scanners like Trivy won't be able to find vulnerabilities in the image. This PR fixes that by adding the necessary COPY commands to include this info in the final image. This way, Trivy and other scanners can identify vulnerabilities in the Typha images.

## Related issues/PRs

fixes [ISSUE 8051](https://github.com/projectcalico/calico/issues/8051)

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/typha/docker-image/Dockerfile.amd64
+++ b/typha/docker-image/Dockerfile.amd64
@@ -37,6 +37,10 @@ LABEL name="Calico Typha" \
       description="Calico Typha is a fan-out datastore proxy" \
       maintainer="Shaun Crampton <shaun@tigera.io>"
 
+# Copy OS release and package metadata
+COPY --from=ubi /var/lib/rpm/* var/lib/rpm
+COPY --from=ubi /etc/redhat-release /etc/redhat-release
+
 COPY --from=ubi /sbin/tini /sbin/tini
 COPY --from=ubi /licenses /licenses
 
--- a/typha/docker-image/Dockerfile.arm64
+++ b/typha/docker-image/Dockerfile.arm64
@@ -33,6 +33,12 @@ RUN chmod +x /sbin/tini
 FROM scratch
 COPY --from=base /sbin/tini /sbin/tini
 
+# Copy OS release and package metadata
+COPY --from=base /var/lib/dpkg/status /var/lib/dpkg/status
+COPY --from=base /var/lib/dpkg/available /var/lib/dpkg/available
+COPY --from=base /var/lib/apt/lists /var/lib/apt/lists
+COPY --from=base /etc/debian_version /etc/debian_version
+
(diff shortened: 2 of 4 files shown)


caseydavenport:
Looks sensible to me, asking @Behnam-Shobiri to take a look

caseydavenport:
@duffney looking at this again, I'm not entirely sure this is the right change.

Generally, this change would have us adding additional code to our images that isn't required for it to function. Generally I would call this a limitation of the scanning tool rather than a problem with the image itself.

Will still wait on @Behnam-Shobiri to comment as well.

Behnam-Shobiri:
Thank you for the PR @duffney. 
I also don't think this would be the right change. Our images have been specifically built to include only libraries that they need for successfully running the code by copying only what's needed to scratch. This will significantly reduce the attack surface. If we include metadata from the building phase, it will create false positives and unnecessary CVE in our image (not all the libraries in the building phase are needed in the final image). 
I think as @caseydavenport  mentioned this is more a limitation of the scanning tool and it would be a good feature for the scanner that you are using to add.

duffney:
> Thank you for the PR @duffney. I also don't think this would be the right change. Our images have been specifically built to include only libraries that they need for successfully running the code by copying only what's needed to scratch. This will significantly reduce the attack surface. If we include metadata from the building phase, it will create false positives and unnecessary CVE in our image (not all the libraries in the building phase are needed in the final image). I think as @caseydavenport mentioned this is more a limitation of the scanning tool and it would be a good feature for the scanner that you are using to add.

Totally understand. Pulling in all the package metadata is overkill, will bloat the image, and create false positives. I hadn't thought about all the other implications. However, if you'll humor me :) 

Would you be open to including the package metadata for the dependencies like `lib64` that are pulled into the image?

Behnam-Shobiri:
@duffney I think if we can only include the metadata for libraries that we are using, it should be fine. But I'm not sure how feasible that is (and how easy it is to maintain).

duffney:
> @duffney I think if we can only include the metadata for libraries that we are using, it should be fine. But I'm not sure how feasible that is (and how easy it is to maintain).

Looking more into it, it's not. lol For example, with Trivy it looks at the package library and there isn't anyway I could figure out to decouple say lib64 from the rest of the packages that existing in the base image. 

I'll go ahead and close the PR. Since the only way to address it is to rebuild the image from a different base without the packages or update the scanner to be more targeted in its search.

PR #8124, opened 2023-10-16
Use nginx:1.25.2 in node k8st tests

wedaly:
## Description

The k8s system tests were using nginx v1.7.9, which was published in 2015 and segfaults on Azure Linux. This commit replaces all hard-coded references to the version with utils.NGINX_IMAGE (overridable with an env var if necessary) and updates the default version to 1.25.2.


caseydavenport (approved):

caseydavenport:
Gave the tests a re-run, as it looked like some had failed. Hopefully a flake and not a consequence.

PR #8127, opened 2023-10-17
Create Dependabot Configuation to support Docker and go module updates

zhuxiaow0:
Enable Dependabot for Dockerfile and go.mod updates.
This will update images in Dockerfile `FROM` statements, and go modules.

This PR is created by a script. Please let me know if we should modify any values.


zhuxiaow0:
Please ignore this PR

PR #8137, opened 2023-10-18
Add ipam_last_sync_timestamp_seconds metric

ashi009:
## Description

IPAM Controller runs a single loop, and the metrics is dumped at the end of each sync operation. In case the sync is halting, the metrics will not be updated, and there is no way for monitoring system to know that the controller has stopped working.

This change adds `ipam_last_sync_timestamp_seconds` metric to export timestamp after each sync with status as label. So the monitoring system can easily create alerts based on that.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add ipam_last_sync_timestamp_seconds metric to kube-controllers
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -50,6 +50,8 @@ import (
 )
 
 var (
+	lastSyncTimestampGauges *prometheus.GaugeVec
+
 	// Multidimensional metrics, with a vector for each pool to allow resets by pool when handling pool deletion and
 	// refreshing metrics. See https://github.com/prometheus/client_golang/issues/834, option 3.
 	inUseAllocationGauges    map[string]*prometheus.GaugeVec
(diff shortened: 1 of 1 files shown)


kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -278,6 +286,9 @@ func (c *ipamController) acceptScheduleRequests(stopCh <-chan struct{}) {
 				// built in to the cleanup code.
 				log.WithError(err).Warn("error syncing IPAM data")
 				kick(c.syncChan)
+				lastSyncTimestampGauges.WithLabelValues("failure").SetToCurrentTime()
```
caseydavenport, line 289: I think we probably want to also catch the syncs that occur a few lines above, in the Periodic IPAM sync branch no?

PR #8139, opened 2023-10-18
Replace CTLB workaround with a config option

sridhartigera:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -437,8 +454,16 @@ type FelixConfigurationSpec struct {
 	// BPFConnectTimeLoadBalancingEnabled when in BPF mode, controls whether Felix installs the connection-time load
 	// balancer.  The connect-time load balancer is required for the host to be able to reach Kubernetes services
 	// and it improves the performance of pod-to-service connections.  The only reason to disable it is for debugging
-	// purposes.  [Default: true]
+	// purposes. This will be deprecated. Use BPFConnectTimeLoadBalancing [Default: true]
 	BPFConnectTimeLoadBalancingEnabled *bool `json:"bpfConnectTimeLoadBalancingEnabled,omitempty" validate:"omitempty"`
+	// BPFConnectTimeLoadBalancing when in BPF mode, controls whether Felix installs the connect-time load
+	// balancer. The connect-time load balancer is required for the host to be able to reach Kubernetes services
+	// and it improves the performance of pod-to-service connections.When set to TCP, connect time load balancing
+	// is available only for services with TCP ports. [Default: Enabled]
```
tomastigera, line 462: Imo our default should be TCP

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -437,8 +454,16 @@ type FelixConfigurationSpec struct {
...
+	BPFConnectTimeLoadBalancing *BPFConnectTimeLBType `json:"bpfConnectTimeLoadBalancing,omitempty" validate:"omitempty,oneof=TCP Enabled Disabled"`
+	// BPFHostNetworkedNATWithoutCTLB when in BPF mode, controls whether Felix does a NAT without CTLB. This along with BPFConnectTimeLoadBalancing
+	// determines the CTLB behavior. [Default: Disabled]
```
tomastigera, line 465: Default should be enabled

The end result is that DNS which is UDP does not break when backend dies, while TCP does not pay the perf hit.

felix/fv/infrastructure/topology.go
```diff
@@ -289,6 +289,8 @@ func StartNNodeTopology(n int, opts TopologyOptions, infra DatastoreInfra) (tc T
 		// host.  So, disable CTLB handling for subsequent Felixes.
 		if i > 0 {
 			optsPerFelix[i].ExtraEnvVars["FELIX_BPFConnectTimeLoadBalancingEnabled"] = "false"
+			optsPerFelix[i].ExtraEnvVars["FELIX_BPFConnectTimeLoadBalancing"] = string(api.BPFConnectTimeLBDisabled)
+			optsPerFelix[i].ExtraEnvVars["FELIX_BPFHostNetworkedNATWithoutCTLB"] = string(api.BPFHostNetworkedNATEnabled)
```
tomastigera, line 293: I think this should be also disabled. We do not want to install the workaround. CTLB is installed test-wide. We do not want to install it multiple times. That is why the option `FELIX_BPFConnectTimeLoadBalancing=false` is overriden above. But this one should be kept as set by the test.

libcalico-go/lib/clientv3/felixconfig.go
```diff
@@ -113,4 +113,12 @@ func setDefaults(fc *apiv3.FelixConfiguration) {
 		disabled := apiv3.FloatingIPsDisabled
 		fc.Spec.FloatingIPs = &disabled
 	}
+	if fc.Spec.BPFConnectTimeLoadBalancing == nil {
```
tomastigera, line 116: This should be the place where we also need to decide if the legacy `BPFConnectTimeLoadBalancingEnabled` is set, right? If it is is and none of the new ones are set, just set `BPFConnectTimeLoadBalancing=true` and `BPFHostNetworkedNATWithoutCTLB=disabled` and then you do not need to worry about `BPFConnectTimeLoadBalancingEnabled` in Felix. If the new one and the old one are set then we need to either raise a warning or fail validation or something, right? Once we deprecate the legacy option we would remove the code here.

tomastigera (changes requested):

felix/config/config_params.go
```diff
@@ -182,6 +182,8 @@ type Config struct {
 	BPFDataIfacePattern                *regexp.Regexp    `config:"regexp;^((en|wl|ww|sl|ib)[Popsx].*|(eth|wlan|wwan).*|tunl0$|vxlan.calico$|wireguard.cali$|wg-v6.cali$)"`
 	BPFL3IfacePattern                  *regexp.Regexp    `config:"regexp;"`
 	BPFConnectTimeLoadBalancingEnabled bool              `config:"bool;true"`
```
tomastigera, line 184: Does it make sense to leave this one without default? In fact, can we remove it here? (me not :100: sure)

felix/dataplane/linux/int_dataplane.go
```diff
@@ -750,13 +753,31 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 			log.Info("BPF enabled but no Kubernetes client available, unable to run kube-proxy module.")
 		}
 
-		if config.BPFConnTimeLBEnabled {
+		if config.BPFConnTimeLB == string(apiv3.BPFConnectTimeLBDisabled) &&
+			config.BPFHostNetworkedNAT == string(apiv3.BPFHostNetworkedNATDisabled) {
+			log.Warn("Access to services from host networked process wont work, forcing hostnetworked NAT to Enabled")
+			config.BPFHostNetworkedNAT = string(apiv3.BPFHostNetworkedNATEnabled)
```
tomastigera, line 759: would it make sense to just give a warning instead of forcing it?

felix/dataplane/linux/int_dataplane.go
```diff
@@ -750,13 +753,31 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
+		}
+		// HostNetworkedNAT is Enabled and CTLB enabled.
+		// HostNetworkedNAT is Disabled and CTLB is either disabled/TCP.
+		// The above cases are invalid configuration. Revert to CTLB enabled.
+		if config.BPFHostNetworkedNAT == string(apiv3.BPFHostNetworkedNATEnabled) {
+			if config.BPFConnTimeLB == string(apiv3.BPFConnectTimeLBEnabled) {
+				log.Warn("Access to services may not work properly, reverting to default CTLB configuration")
```
tomastigera, line 766: `config.BPFHostNetworkedNAT == string(apiv3.BPFHostNetworkedNATEnabled` doesnot make sense, but also would not do much harm. I would revert here to Disabled instead of flipping the other to TCP.

felix/fv/infrastructure/topology.go
```diff
@@ -289,6 +289,8 @@ func StartNNodeTopology(n int, opts TopologyOptions, infra DatastoreInfra) (tc T
 		// host.  So, disable CTLB handling for subsequent Felixes.
 		if i > 0 {
 			optsPerFelix[i].ExtraEnvVars["FELIX_BPFConnectTimeLoadBalancingEnabled"] = "false"
```
tomastigera, line 291: we can remove this one, right?

felix/dataplane/linux/int_dataplane.go
sridhartigera: I think flipping the ```ConnTimeLB``` to ```TCP``` keeps it line with the defaults.

tomastigera (approved):

PR #8153, opened 2023-10-23
[BPF] ipv6 bootstrap

tomastigera:
## Description

    [BPF] icmp errors must be always treated as related
    
    There is no reason why error would be generated if there was not already
    a related data stream. We should not treat icmp as related if there is
    no CT match yet. In ICMPv6 there is a high probability (it is pretty much
    certain that there is alredy a CT entry for ICMPv6 stream because of
    usage for ipv6 bootstrap, neighbor and router lookup etc. Related
    traffic should not match on such CT entries and passed unmodified in
    case of NAT etc.

    [BPF] conntrack - remove some explicit variable - use STATE

    [BPF] Do not allow forwarding icmpv6 bootstrap messages
    
    BPF program on tc hooks allow the messages in and out, but they must be
    allow only between endpoints and the host. No forwarding must be
    allowed. We always force the messages into iptables on any ingress to
    the host and we make iptable rules to make sure that they are dropped if
    forwarded.

    [BPF] always zero s/dports for icmp(v6)
    
    We only use state to communicate the type/code to the program that
    generates icmp responses. So we can safely zero the ports in the state
    to make conntrack correct and the code compilable also for v6.

    iptables ipv6 test suite - test required icmpv6 allowed

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: icmp errors must be always treated as related, not only after a CT miss.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/parsing.h
+++ b/felix/bpf-gpl/parsing.h
@@ -175,14 +175,13 @@ static CALI_BPF_INLINE int tc_state_fill_from_nexthdr(struct cali_tc_ctx *ctx, b
 		break;
 #ifdef IPVER6
 	case IPPROTO_ICMPV6:
-		/* XXX for now, just allow */
-		CALI_DEBUG("ICMPv6: allow");
-		goto allow;
+		ctx->state->sport = ctx->state->dport = 0;
+		CALI_DEBUG("ICMP; type=%d code=%d\n",
+				icmp_hdr(ctx)->type, icmp_hdr(ctx)->code);
+		break;
 #else
 	case IPPROTO_ICMP:
-		ctx->state->icmp_type = icmp_hdr(ctx)->type;
-		ctx->state->icmp_code = icmp_hdr(ctx)->code;
-
+		ctx->state->sport = ctx->state->dport = 0;
 		CALI_DEBUG("ICMP; type=%d code=%d\n",
 				icmp_hdr(ctx)->type, icmp_hdr(ctx)->code);
 		break;
--- a/felix/bpf-gpl/types.h
+++ b/felix/bpf-gpl/types.h
@@ -73,6 +73,9 @@ struct cali_tc_state {
 		__u16 dport;
 		struct
 		{
+			/* Only used to pass type/code to the program that generates and
+			 * send and ICMP error respose.
+			 */
 			__u8 icmp_type;
 			__u8 icmp_code;
 		};
--- a/felix/bpf-gpl/types.h
+++ b/felix/bpf-gpl/types.h
@@ -206,7 +206,9 @@ struct cali_tc_ctx {
 			}							\
 										\
 			x;							\
-	})									\
+	})
+
+#define STATE (ctx->state)
 
 #define fib_params(x) ((struct bpf_fib_lookup *)((x)->scratch))
 
--- a/felix/dataplane/linux/int_dataplane.go
+++ b/felix/dataplane/linux/int_dataplane.go
@@ -582,7 +582,17 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		// bpffs so there's nothing to clean up
 	}
 
-	ipsetsManager := common.NewIPSetsManager(ipSetsV4, config.MaxIPSetSize)
+	ipsetsManager := common.NewIPSetsManager("ipv4", ipSetsV4, config.MaxIPSetSize)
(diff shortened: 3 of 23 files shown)


felix/bpf-gpl/types.h
```diff
@@ -73,6 +73,9 @@ struct cali_tc_state {
 		__u16 dport;
 		struct
 		{
+			/* Only used to pass type/code to the program that generates and
+			 * send and ICMP error respose.
```
fasaxc, line 77:
```suggestion
			 * sends an ICMP error respose.
```

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1544,14 +1544,21 @@ func (d *InternalDataplane) setUpIptablesBPF() {
...
-				// In BPF mode, we don't support IPv6 yet.  Drop it.
-				fwdRules = append(fwdRules, iptables.Rule{
-					Match:   iptables.Match().OutInterface(prefix + "+"),
-					Action:  d.ruleRenderer.IptablesFilterDenyAction(),
-					Comment: []string{"To workload, drop IPv6."},
-				})
+		if t.IPVersion == 6 {
+			if !d.config.BPFIpv6Enabled {
+				for _, prefix := range rulesConfig.WorkloadIfacePrefixes {
+					// In BPF mode, we don't support IPv6 yet.  Drop it.
```
fasaxc, line 1550: Out of date comment.

Given we only have single stack right now, do we have code to block IPv4 packets when in IPv6 mode?  Think we should do that since we're not able to police them.

felix/fv/bpf_test.go
```diff
@@ -3688,6 +3688,9 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 
 					BeforeEach(func() {
 						icmpProto := numorstring.ProtocolFromString("icmp")
+						if testOpts.ipv6 {
```
fasaxc, line 3691: Looks like this code isn't reached due to early return on line 3680?

felix/rules/static.go
```diff
@@ -359,7 +360,7 @@ func (r *DefaultRuleRenderer) filterInputChain(ipVersion uint8) *Chain {
 	}
 }
 
-func (r *DefaultRuleRenderer) filterWorkloadToHostChain(ipVersion uint8) *Chain {
+func Icmpv6Filter(action iptables.Action) []Rule {
```
fasaxc, line 363:
```suggestion
func ICMPv6Filter(action iptables.Action) []Rule {
```

felix/bpf-gpl/parsing.h
```diff
@@ -175,14 +175,13 @@ static CALI_BPF_INLINE int tc_state_fill_from_nexthdr(struct cali_tc_ctx *ctx, b
...
-		/* XXX for now, just allow */
-		CALI_DEBUG("ICMPv6: allow");
-		goto allow;
+		ctx->state->sport = ctx->state->dport = 0;
+		CALI_DEBUG("ICMP; type=%d code=%d\n",
+				icmp_hdr(ctx)->type, icmp_hdr(ctx)->code);
+		break;
 #else
 	case IPPROTO_ICMP:
-		ctx->state->icmp_type = icmp_hdr(ctx)->type;
```
fasaxc, line 183: Do we not use this for policy?  I thought our policy model supported matching on it (so it would need to be in the state for the policy program to see).

fasaxc (changes requested):

felix/dataplane/linux/int_dataplane.go
tomastigera: yes. we only have ipv6 programs loaded and these will fail to parse ipv4 packets and thus they deny them.

felix/bpf-gpl/parsing.h
tomastigera: hmmm right , policy ... :thinking:

felix/fv/bpf_test.go
tomastigera: Will be once we enable this test above

felix/bpf-gpl/parsing.h
tomastigera: I made this change because verifier was exploding and stack was overflowing. I think I found the right permutation where everybody is happy.

PR #8156, opened 2023-10-24
better support of ipv6-only clusters: prevent creation of default-ipv4-pool during calico-node startup

kruftik:
## Description

Current calico-node startup logic supports creation of default ip-pools both for ipv4 and ipv6 stacks if the ones do not exist. However, it does not take into account accordingly the ipv6-only clusters (without ipv4 addresses on any interfaces at all): as soon as default-ipv4 pool being created, every calico-node instance begin log flooding with something like: 

```
felix/route_table.go 1016: Failed to get link attributes error=interface not present ifaceRegex="^vxlan.calico$" ipVersion=0x4 tableIndex=254
````

This PR mitigate such an issue: in addition to `NO_DEFAULT_POOLS` environment variable, `CALICO_IPV4POOL_CIDR`=`none` and `CALICO_IPV6POOL_CIDR`=`none` special values are added.

## Release Note

```release-note
In order to prevent default IP-pools creation, `CALICO_IPV4POOL_CIDR`=`none` and `CALICO_IPV6POOL_CIDR`=`none` environment variable special values are now supported.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/pkg/lifecycle/startup/startup.go
+++ b/node/pkg/lifecycle/startup/startup.go
@@ -749,16 +749,33 @@ func configureIPPools(ctx context.Context, client client.Interface, kubeadmConfi
 		return
 	}
 
+	var (
+		ipv4PoolEnabled = true
+		ipv6PoolEnabled = true
+	)
+
+	if ipv4Pool == "none" {
+		log.Info("Skipping IPv4 pool configuration")
+
+		ipv4PoolEnabled = false
+	}
+
+	if ipv6Pool == "none" {
+		log.Info("Skipping IPv6 pool configuration")
+
+		ipv6PoolEnabled = false
+	}
+
 	// If CIDRs weren't specified through the environment variables, check if they're present in kubeadm's
 	// config map.
-	if (len(ipv4Pool) == 0 || len(ipv6Pool) == 0) && kubeadmConfig != nil {
+	if ((ipv4PoolEnabled && len(ipv4Pool) == 0) || (ipv6PoolEnabled && len(ipv6Pool) == 0)) && kubeadmConfig != nil {
 		v4, v6, err := extractKubeadmCIDRs(kubeadmConfig)
 		if err == nil {
-			if len(ipv4Pool) == 0 {
+			if ipv4PoolEnabled && len(ipv4Pool) == 0 {
 				ipv4Pool = v4
 				log.Infof("found v4=%s in the kubeadm config map", ipv4Pool)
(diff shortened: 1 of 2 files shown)


caseydavenport:
I wonder if instead of introducing new env vars for this, we should follow the pattern we used for `CALICO_NETWORKING_BACKEND` and support a `none` option for the existing `IPV4POOL_CIDR` variable.

That said, I am curious why using `NO_DEFAULT_POOLS` and then creating an IPv6 pool out-of-band isn't viable here? That would be my preferred solution here, since in general I would like us to move away from using calico/node to create IP pools - configuring APIs from a collection of env vars makes it hard to support all of the necessary use cases. For now though, calico/node is where default IP pools are created, so I'm not completely against merging something like this as a stopgap.

kruftik:
@caseydavenport ,

> we should follow the pattern we used for `CALICO_NETWORKING_BACKEND` and support a `none` option for the existing `IPV4POOL_CIDR` variable.

yeah, it's a better option. I will refactor the PR into such an approach.

kruftik:
@caseydavenport , done. Could you look into?

node/pkg/lifecycle/startup/startup.go
```diff
@@ -762,16 +762,33 @@ func configureIPPools(ctx context.Context, client client.Interface, kubeadmConfi
...
+	)
+
+	if ipv4Pool == "none" {
+		log.Info("Skipping IPv4 pool configuration")
+
+		ipv4PoolEnabled = false
+	}
+
+	if ipv6Pool == "none" {
+		log.Info("Skipping IPv4 pool configuration")
```
com30n, line 777: It looks like it should be `Skipping IPv6 pool configuration`

caseydavenport (changes requested):
Apologies for the delay here - this looks good to me!

I do think we need a pair of unit tests to verify this behavior works - here's an example of a similar test you can base it on: https://github.com/projectcalico/calico/blob/8d31b2669e8cde1558b128fc7db3015970921e90/node/pkg/lifecycle/startup/startup_test.go#L455-L477

caseydavenport:
@kruftik is this still on your radar? The only thing remaining is a unit test (see my last comment)

kruftik:
> @kruftik is this still on your radar? The only thing remaining is a unit test (see my last comment)

seems like done

caseydavenport (approved):

PR #8170, opened 2023-10-30
Make operator.tigera.io cluster-role user settable

titanlien:
### Make operator.tigera.io cluster-role user settable. 

In order to achieve [Rbac - Least privilege](https://kubernetes.io/docs/concepts/security/rbac-good-practices/#least-privilege), we should allow user to setup customer resource scope.

Otherwise we should list down the necessary reasons we need a wildcard here.

### DIFF
--- a/charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
+++ b/charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
@@ -96,8 +96,7 @@ rules:
       - update
   - apiGroups:
       - operator.tigera.io
-    resources:
-      - '*'
+    resources: {{ toYaml .Values.clusterrole.operatorResources | nindent 4 }}
     verbs:
       - create
       - get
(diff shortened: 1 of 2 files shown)


titanlien:
> @titanlien thanks for the PR - could you explain your use-case a little bit?

Updated it on description

caseydavenport:
> Otherwise we should list down the necessary reasons we need a wildcard here.

I think this is the route we should go. Namely, we should minimize our RBAC and list out exactly why each permission is needed.

I 100% agree that it is too permissive at the moment. I am not yet convinced that making this user-configurable is the right move. It's a monumental task to attempt to quantify precisely what permissions are needed, and I think that this is a big "foot gun" and a headache to say "Figure it out yourself" to users.

We should be able to remove all of those `*` instances from our RBAC without this.

caseydavenport:
I've raised this PR which is a start in terms of removing the wildcard instance: https://github.com/projectcalico/calico/pull/8206

Further PRs should also add matching based on namespace and resource name.

PR #8177, opened 2023-11-01
Align types in API server's resource package.

fasaxc:
## Description

There was an awkward mix of Atoi -> int, int64 and uint64 with casting between them.  int is int64 on most platforms that we run on and resource IDs are unlikely to wrap an int64 but let's just use uint64 throughout so the code behaves the same on all systems.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/apiserver/pkg/storage/calico/resource.go
+++ b/apiserver/pkg/storage/calico/resource.go
@@ -5,7 +5,6 @@ package calico
 import (
 	"bytes"
 	"fmt"
-	"math"
 	"os"
 	"reflect"
 	"strconv"
@@ -291,7 +290,7 @@ func (rs *resourceStore) List(ctx context.Context, key string, optsK8s storage.L
 type objState struct {
 	obj  runtime.Object
 	meta *storage.ResponseMeta
-	rev  int64
+	rev  uint64
 	data []byte
 }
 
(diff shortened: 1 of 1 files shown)


apiserver/pkg/storage/calico/resource.go
```diff
@@ -428,15 +427,16 @@ func (rs *resourceStore) GuaranteedUpdate(
 		if err != nil {
 			return err
 		}
-		revInt, _ := strconv.Atoi(accessor.GetResourceVersion())
+		versionStr := accessor.GetResourceVersion()
+		var revInt uint64
+		if versionStr != "" {
+			revInt, err = strconv.ParseUint(versionStr, 10, 64)
+			return fmt.Errorf("failed to parse non-empty resource version %q key=%q: %w", versionStr, key, err)
```
fasaxc, line 434: This error check feels like the most contentious change here.  I'm going by the comment above: "Check for Revision no. If not set or less than the current version" implies that it should either be empty or a number we can parse.  Anything else feels like a problem.

apiserver/pkg/storage/calico/resource.go
```diff
@@ -428,15 +427,16 @@ func (rs *resourceStore) GuaranteedUpdate(
 		if err != nil {
 			return err
 		}
-		revInt, _ := strconv.Atoi(accessor.GetResourceVersion())
+		versionStr := accessor.GetResourceVersion()
+		var revInt uint64
+		if versionStr != "" {
+			revInt, err = strconv.ParseUint(versionStr, 10, 64)
```
caseydavenport, line 433: I think we need to check if the err is non-nil here

caseydavenport (approved):
One missing err check but otherwise this LGTM

apiserver/pkg/storage/calico/resource.go
fasaxc: Oops, yep, reworked the code and left it half baked

Issue #7832, opened 2023-07-03
Default IP Autodetection method should select real interfaces


nadenf:
The default IP Autodetection method as described [here](https://docs.tigera.io/calico/latest/networking/ipam/ip-autodetection#manually-configure-ip-address-and-subnet-for-a-node) is to select the first valid IP address. 

However if you have Tailscale or similar virtual interfaces they can be selected even if under almost all conditions they would not be a sensible choice.

## Expected Behavior
Rather than go through a [painful](https://github.com/rancher/rke2/issues/4320#) triage adventure it would be better if Calico could either (a) select a real interface i.e. network card or (b) select the interface associated with the InternalIP address that Kubernetes has selected.

## Current Behavior
Pods will run and have connectivity with each other but if say Tailscale is selected will not be able to access outside world.

## Possible Solution
If you look at ethtool for Tailscale it should be obvious it's not real:
```
Settings for tailscale0:
	Supported ports: [  ]
	Supported link modes:   Not reported
	Supported pause frame use: No
	Supports auto-negotiation: No
	Supported FEC modes: Not reported
	Advertised link modes:  Not reported
	Advertised pause frame use: No
	Advertised auto-negotiation: No
	Advertised FEC modes: Not reported
	Speed: Unknown!
	Duplex: Full
	Auto-negotiation: off
	Port: Twisted Pair
	PHYAD: 0
	Transceiver: internal
	MDI-X: Unknown
        Current message level: 0x00000000 (0)
	Link detected: yes 
```

## Steps to Reproduce (for bugs)
1. Install Tailscale and/or multiple network cards (one that has weird routing and should not be used)
2. Install Calico using VXLAN mode on RKE2/similar and have it select the Tailscale interface (you may need to rename others)
3. Add two nodes.
4. Launch pod that is pinned to second node.
5. Pod will be able to access other pods on first node but not access internet.

## Your Environment
* Calico version: v3.25.0
* Orchestrator version (e.g. kubernetes, mesos, rkt): RKE2 v1.25.11+rke2r1
* Operating System and version: Linux Ubuntu 22.04
* Link to your project (optional):

mazdakn:
@nadenf It is possible to configure [auto detection](https://docs.tigera.io/calico/latest/networking/ipam/ip-autodetection#autodetection-methods) functionality. You can use a regex to include or exclude an interface from auto detection.

nadenf:
@mazdakn .. Fully aware. But just close if it's not an issue. 

[...]

PR #8185, opened 2023-11-04
Add Tailscale to interface exclusion list

nadenf:
Fixes https://github.com/projectcalico/calico/issues/7832.

It excludes Tailscale interfaces from being used for Calico.


caseydavenport:
@nadenf I am not sure this reaches the bar required to be added to the code.

Calico already allows for a configurable way to auto detect interfaces - have you tried using the node address auto detection options in your environment and if so, could you add some context as to why they aren't suitable for your case?

https://docs.tigera.io/calico/latest/networking/ipam/ip-autodetection

caseydavenport:
I see @coutinhop suggested you raise this PR, so apologies for going against his recommendation! 

I'll chat with him and reach an agreement. In general, I am trying to set a high bar for this - we unfortunately can't accommodate every possible environment with our default values and so attempting to do so is both a lost cause and increases the complexity of the product.

caseydavenport:
I spoke with @coutinhop and I think we're in agreement that this should be done via configuration rather than via code change. Ultimately it's a losing battle to attempt to make the default value work for every possible configuration. Apologies for the confusion and thank you for your willingness to contribute.

PR #8192, opened 2023-11-08
Use upstream hcsshim instead of fork

coutinhop:
Use github.com/Microsoft/hcsshim v0.11.4 instead of the fork github.com/projectcalico/hcsshim v0.8.9-calico.

The fork was necessary at the time of v0.8.9 in order to access the endpoint's containers (https://github.com/projectcalico/hcsshim/blob/v0.8.9-calico/internal/hns/hnsendpoint.go#L119-L138), but it is no longer needed since now that information is accessible via the SharedContainers field in the HNSEndpoint struct in the upstream hcsshim package (https://github.com/microsoft/hcsshim/blob/main/internal/hns/hnsendpoint.go#L36).

## Todos

- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Use upstream hcsshim package instead of fork and update it to latest stable version (v0.11.4).
```

## Reminder for the reviewer

Other optional labels:


Behnam-Shobiri (approved):

hjiawei (approved):

felix/dataplane/windows/endpoint_mgr.go
```diff
@@ -192,15 +191,7 @@ func (m *endpointManager) RefreshHnsEndpointCache(forceRefresh bool) error {
...
-		containers, err := m.hns.GetAttachedContainerIDs(&endpoint)
-		if err != nil {
-			log.WithFields(log.Fields{
-				"id":   endpoint.Id,
-				"name": endpoint.Name,
-			}).Warn("Failed to get attached containers")
-			continue
-		}
-		if len(containers) == 0 {
+		if len(endpoint.SharedContainers) == 0 {
```
hjiawei, line 194: I am sorry I guess this is caught late. Do we need to make a `hnsCall("GET", "/endpoints/"+endpoint.Id, "", &attachInfo)` call before we check `SharedContainers`?

coutinhop: Calling `IsAttached()` wouldn't work because it requires a specific container ID as an arg (so we'd need to somehow list all containers and then loop through them calling `IsAttached()` for each of them, per endpoint). But I think that's not necessary.

The endpoints are retrieved via an `hnsCall()` which happens inside `HNSListEndpointRequest()`, which in turn is called here: https://github.com/projectcalico/calico/blob/59b28041cacd4819dcea4224e08686482cabc882/felix/dataplane/windows/endpoint_mgr.go#L162

I couldn't find explicit documentation on `hnsCall()`, but I spun up a windows cluster and used the `hnsdiag` tool on powershell and verified that it does populate `SharedContainers`:
```
PS C:\C\e815cc9c19ec519569d4eeabe514771c61f1531e2c443299a420f27e8bcc0dd8\CalicoWindows> hnsdiag list endpoints -d
(...)
    "SharedContainers":[
        "1c809ef2e5d94f56600fe3c217e28838ba7641d95f2a3cd566126912ec1dee2c"
    ]
}
```

The `SharedContainers` must be ultimately being unmarshalled from the response here: https://github.com/projectcalico/hcsshim/blob/master/internal/hns/hnsfuncs.go#L43

PR #8209, opened 2023-11-14
Get node's IPv6 address with non-calico CNI

tomastigera:
Without calico cni, Node.Spec.BGP is not populated. In v4 mode, libcalico falls back to getting the node's address from Node.Spec.Addresses. IPv6 detection takes a different path and always depended on Node.Spec.BPG being present. Do the same fallback in DataplanePassthru where the v6 updates are created.

This guarantees that we always know the node's IP in k8s environment. eBPF mode needs to know that to load its hep programs.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Node learns about it's ipv6 address in kubernetes even if BGP is turned off and CNI is not calico.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/calc/dataplane_passthru.go
+++ b/felix/calc/dataplane_passthru.go
@@ -21,6 +21,7 @@ import (
 	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 
 	libv3 "github.com/projectcalico/calico/libcalico-go/lib/apis/v3"
+	cresources "github.com/projectcalico/calico/libcalico-go/lib/resources"
 
 	"github.com/projectcalico/calico/felix/dispatcher"
 	"github.com/projectcalico/calico/felix/proto"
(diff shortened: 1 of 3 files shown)


felix/calc/dataplane_passthru.go
```diff
@@ -140,20 +143,47 @@ func (h *DataplanePassthru) OnUpdate(update api.Update) (filterOut bool) {
...
+					if ip != nil {
+						oldIP := h.hostIPv6s[hostname]
+						if oldIP != nil && ip.IP.Equal(oldIP.IP) {
+							log.WithField("update", update).Debug("Ignoring duplicate Node IPv6 address update")
+							return
+						}
+						log.WithField("update", update).Debug("Passing-through Node IPv6 address update")
+						h.hostIPv6s[hostname] = ip
+						h.callbacks.OnHostIPv6Update(hostname, ip)
+					} else {
```
fasaxc, line 182: Same here; could check if IP is in h.hostIPv6

fasaxc (approved):
Looks like it should do the trick but I wonder if we should move it to its own calc graph node.  The logic is starting to get a bit complex for the "pass through", which was mean to be a simple stateless component for passing through trivial objects.

marvin-tigera:

tomastigera:
only node-ci failed

PR #8212, opened 2023-11-14
Update flannel to v0.23.0

laibe:
## Description

Updates flannel to v0.23.0

Tested with:
* Talos v1.5.4
* Kernel 6.1.58
* Kubernetes v1.28.1

### DIFF
--- a/charts/calico/values.yaml
+++ b/charts/calico/values.yaml
@@ -14,7 +14,7 @@ kubeControllers:
   image: docker.io/calico/kube-controllers
 flannel:
   image: docker.io/flannel/flannel
-  tag: v0.21.3
+  tag: v0.23.0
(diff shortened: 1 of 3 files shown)


mazdakn:
@laibe is .0 release is stable? normally it's the .1 or .2 that's the stable release.

mazdakn:
@laibe, generally speaking x.x.0 release is the one with the new features which is expected to have some bugs. Those bugs are fixed in the patch releases like x.x.1, so in general x.x.0 is not a good candidate to run in production. This is a general pattern, and I am not sure about flannel release cadence.

mazdakn:
@laibe it seems flannel `v0.24.2` is released. Do you move to that one?

PR #8250, opened 2023-11-21
[BPF] fix ct cleaning for host processes w/o ctlb using service

tomastigera:
If there is an overlay, forwarding entry may not have a matching client IP as it may start with host's main IP, but then it is turned into host's tunnel IP. We cannot determine what is the client/service/backend from that entry only. We need to lookup the reverse entry. If the reverse entry does not exist, we delete the forarding entry anyway as there is no use for it without the reverse half.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/conntrack/cleanup.go
+++ b/felix/bpf/conntrack/cleanup.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 
 	"github.com/projectcalico/calico/felix/bpf/maps"
 	"github.com/projectcalico/calico/felix/timeshim"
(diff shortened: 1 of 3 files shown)


felix/bpf/conntrack/cleanup.go
```diff
@@ -288,8 +291,16 @@ func (sns *StaleNATScanner) Check(k KeyInterface, v ValueInterface, _ EntryGet)
...
+				rv, err := get(revKey)
+				if err != nil {
+					// There is no match for the reverse key, delete it, its useless - we
+					// can get here due to host networked program accessing service
+					// without ctlb when the backed is accessible via tunnel.
+					return ScanVerdictDelete
+				}
+				k = revKey
+				v = rv
+				goto reverse // handle it based on the reverse entry
```
fasaxc, line 303: Think I'd prefer recursion to a goto (at least that would die with a stack overflow if there was an infinite loop).

felix/bpf/conntrack/cleanup.go
```diff
@@ -288,8 +291,16 @@ func (sns *StaleNATScanner) Check(k KeyInterface, v ValueInterface, _ EntryGet)
 				svcIP = kA
 				svcPort = kAport
 			} else {
-				log.WithFields(log.Fields{"key": k, "value": v}).Error("Mismatch between key and rev key")
-				return ScanVerdictOK // don't touch, will get deleted when expired
+				rv, err := get(revKey)
+				if err != nil {
```
fasaxc, line 295: Can we check that it was a "not exists" error?

felix/bpf/conntrack/cleanup.go
```diff
@@ -288,8 +291,16 @@ func (sns *StaleNATScanner) Check(k KeyInterface, v ValueInterface, _ EntryGet)
...
+					// There is no match for the reverse key, delete it, its useless - we
+					// can get here due to host networked program accessing service
+					// without ctlb when the backed is accessible via tunnel.
+					return ScanVerdictDelete
```
fasaxc, line 299: Suggest at least a debug log here; if we ever have a bug in cleanup, good to have a log for every deletion so we can trace it back

fasaxc (changes requested):
Couple of suggestions.

How does the BPF program behave with these mismatched entries?  Is it truly broken in that case?  could we avoid _creating_ a broken entry in if we know that'll be the case

felix/bpf/conntrack/cleanup.go
tomastigera: `reverse` checks what the type is to avoid infinite looping. Will never jump back to this.

fasaxc (approved):
Don't think I need to re-review now you explained the fix.

PR #8279, opened 2023-11-30
Fix non-encap routes on parent device change

fasaxc:
## Description

Previously, there were two problems:

- The routing table was not updated to use the new parent.
- Even after restarting felix, the routing table was unable to clean up the old routes.

This PR makes a partial fix:

- VXLAN manager now recreates the RouteTable when the parent device changes.
- To make that safe, move the creation to the main goroutine, with a channel to send the new parent name.
- When making a new RouteTable, include previous parent interface names in the regex so that the old routes get cleaned up.  This only works if Felix is not restarted at the wrong moment.
- As belt-and-braces, add a disabled-by-default feature flag that tells the RouteTable to use RouteReplace to clobber conflicting routes.  (This was my first attempt at ta fix but it has its own problems.)

Also noticed that KeepVXLANDeviceInSync could take a long time to respond to changes, add a kick channel.

Move feature gates to the feature detector. Avoid needing to pass yet another object to 30 places.

## Related issues/PRs

CORE-9902

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Fix that cross-subnet routes were not moved when the VXLAN parent device was changed.
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/vxlan_mgr.go
```diff
@@ -525,54 +514,99 @@ func (m *vxlanManager) CompleteDeferredWork() error {
 	return nil
 }
 
+func (m *vxlanManager) OnParentNameUpdate(name string) {
+	if m.parentIfaceName == "" {
+		m.logCtx.Warn("Empty parent interface name? Ignoring.")
```
caseydavenport, line 519: Think we need a "return" here in this branch

felix/dataplane/linux/vxlan_mgr.go
```diff
@@ -597,7 +631,7 @@ func (m *vxlanManager) getParentInterface(localVTEP *proto.VXLANTunnelEndpointUp
 		}
 		for _, addr := range addrs {
 			if addr.IPNet.IP.String() == parentDeviceIP {
-				m.logCtx.Debugf("Found parent interface: %s", link)
+				m.logCtx.Debugf("Found parent interface: %v", link)
```
caseydavenport, line 634: Might be worth `%+v`?

felix/routetable/route_table.go
```diff
@@ -716,7 +719,64 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string, fullSync bool, firstTry
...
+					logCxt.WithFields(log.Fields{
+						"calicoRoute":          target,
+						"attemptedKernelRoute": route,
+						"conflictingRoutes":    conflictingRoutes,
+					}).Warn("Route conflicts with another, attempting to delete conflicting route.")
+				}
+				delErr := nl.RouteDel(routeToDel)
+				triedDeletion = true
+				if delErr != nil && !os.IsNotExist(delErr) {
+					logCxt.WithError(err).WithField("route", route).Warn(
```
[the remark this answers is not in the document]
fasaxc: It was to avoid clobbering the RouteAdd error, which gets logged below.  I'll introduce a new function, as you note, it's getting bug-prone.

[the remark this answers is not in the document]
fasaxc: (I had half an eye to backporting so I was holding off on doing a big refactor, but a small one should be fine :+1: )

felix/routetable/route_table.go
```diff
@@ -220,6 +221,7 @@ type RouteTable struct {
 
 	// The route deletion grace period.
 	routeCleanupGracePeriod time.Duration
+	featureDetector         environment.FeatureDetectorIface
```
coutinhop, line 224: CI's failing with:
```
routetable/route_table_test.go:74:4: cannot use dataplane (variable of type *mocknetlink.MockNetlinkDataplane) as environment.FeatureDetectorIface value in argument to NewWithShims: *mocknetlink.MockNetlinkDataplane does not implement environment.FeatureDetectorIface (missing method FeatureGate)
```
I think it should be added to `MockNetlinkDataplane` in https://github.com/projectcalico/calico/blob/master/felix/netlinkshim/mocknetlink/netlink.go

felix/routetable/route_table.go
```diff
@@ -748,6 +751,78 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string, fullSync bool, firstTry
...
+	}
+	allRoutes, err := nl.RouteListFiltered(r.netlinkFamily, &netlink.Route{
+		Table: route.Table,
+	}, filterFlags)
+	if err != nil {
+		return nil, err
+	} else {
+		var conflictingRoutes []netlink.Route
+		for _, r := range allRoutes {
+			if r.Dst.String() == route.Dst.String() {
```
mazdakn, line 818: `Dst` is of type `*net.IPNet`, so we should check for `nil` here, before calling `String()`. I recall this causing panic in some other places.

felix/dataplane/linux/vxlan_mgr.go
```diff
@@ -525,54 +514,100 @@ func (m *vxlanManager) CompleteDeferredWork() error {
...
+	parentName := ""
+
+	sleepMonitoringChans := func(maxDuration time.Duration) {
+		timer := time.NewTimer(maxDuration)
+		defer timer.Stop()
+		select {
+		case <-timer.C:
+			return
+		case <-ctx.Done():
+			logrus.Debug("Sleep returning early: context finished.")
```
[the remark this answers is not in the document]
fasaxc: No, but it's a bit odd that I put one in one case and not the other.  There's no default fallthrough in Go so all of these cases break out of the select and then the function returns.

caseydavenport (approved):
A couple minor comments but I think this looks good!

fasaxc:
Hmm, just investigating why the code is firing a lot more than it should.  Spotted it in CI and just managed to repro.  Seems to happen on many pods.  Perhaps a race with the CNI plugin creating the routes or something.

fasaxc:
Keep finding more issues; I've switched to using RouteReplace, which is a lot cleaner, but I think there are still some cases where we'll get into an inconsistent state if we're removing a route here and adding it over there.

fasaxc:
The problem I spotted with the previous iteration was that, when routes move between different RouteTables the previous iteration of the fix could fail to converge until the next full route resync. I've added a test for that; previous iteration of the fix failed the test but this now works.  Why did it fail to converge?  When two RouteTables use RouteReplace, both RouteTables succeed in writing their route and they mark it as "in sync".  However, what actually happens is that one clobbers the other's route.  Then, when the datastore conflict is resolved, there's no guarantee that the kernel gets updated correctly.

I've made the following changes:
- Default RouteReplace to off, behind a feature flag.  (So RouteTable behaves as before by default.)
- Adjust the VXLAN manager so that it keeps track of recently used parent device names and includes them all in the interface regex of the RouteTable that it uses.  This makes sure that the old routes get cleaned up when the parent device is changed.

The limitation to the new approach is that it can't handle Felix being restarted while changing the parent device because we lose track of the old parent devices.

My further refactoring work will solve that problem but it'd be good to get the fixes to the VXLAN manager banked so that they don't add to the refactoring PRs.

caseydavenport (approved):
:1st_place_medal:

felix/fv/routes_test.go
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) 2023 Tigera, Inc. All rights reserved.
```
mazdakn, line 1: 2024

felix/fv/infrastructure/felix.go
```diff
@@ -126,6 +126,7 @@ func RunFelix(infra DatastoreInfra, id int, options TopologyOptions) *Felix {
 		"SET_CORE_PATTERN": "true",
 
 		"FELIX_LOGSEVERITYSCREEN":        options.FelixLogSeverity,
+		"FELIX_LogDebugFilenameRegex":    options.FelixDebugFilenameRegex,
```
mazdakn, line 129: I didn't know about this flag! this is so useful.

mazdakn (approved):
LGTM. just one small thing.

PR #8280, opened 2023-11-30
fix dropped ipam error

alrs:
This fixes a dropped `err` variable in `libcalico-go/lib/ipam/`. No release note is required.


caseydavenport (approved):
LGTM - good catch!

PR #8289, opened 2023-12-06
Use COPY instead of ADD in Dockerfiles

fasaxc:
## Description

- Recommended by CIS benchmark.  Add linter to check it.
- Remove `tini` init daemon from Typha.  AFAICT, we now handle the signals we need to and Typha doesn't spawn sub-processes so we don't need a reaper.
  
## Related issues/PRs

CORE-10036

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Docker images now use COPY instead of ADD as recommended by CIS. Typha no longer relies on the tini init daemon, it handles the common signals internally (and it does not spawn any subprocesses so there is no need for a reaper).
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/typha/docker-image/Dockerfile.amd64
+++ b/typha/docker-image/Dockerfile.amd64
@@ -19,12 +19,12 @@ FROM ${UBI_IMAGE} as ubi
 
 # Since our binary isn't designed to run as PID 1, run it via the tini init daemon.
 ENV TINI_VERSION v0.18.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-amd64 /sbin/tini
+COPY https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-amd64 /sbin/tini
 RUN chmod +x /sbin/tini
 
 # Add in top-level license file
 RUN mkdir /licenses
-ADD licenses /licenses
+COPY licenses /licenses
 COPY LICENSE /licenses
 
 FROM scratch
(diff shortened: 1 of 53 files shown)


typha/docker-image/Dockerfile.s390x
```diff
@@ -27,16 +27,16 @@ COPY --from=qemu /usr/bin/qemu-*-static /usr/bin/
 
 # Since our binary isn't designed to run as PID 1, run it via the tini init daemon.
 ENV TINI_VERSION v0.18.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-s390x /sbin/tini
+COPY https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-s390x /sbin/tini
```
hjiawei, line 30: I think `ADD` is needed here. `COPY` won't download from URLs.

typha/docker-image/Dockerfile.amd64
```diff
@@ -19,12 +19,12 @@ FROM ${UBI_IMAGE} as ubi
 
 # Since our binary isn't designed to run as PID 1, run it via the tini init daemon.
 ENV TINI_VERSION v0.18.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-amd64 /sbin/tini
+COPY https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-amd64 /sbin/tini
```
hjiawei, line 22: I think `ADD` is needed here. `COPY` won't download from URLs.

hjiawei (approved):

Issue #8293, opened 2023-12-07
bpf_ep_mgr.go fails to list existing bpf programs when one has a numeric type


sfudeus:
Felix logs a warning if there are existing bpf programs (e.g. not created by Calico) which have a numeric type instead of a string type. This likely is caused by the potentially wrong assumption that the type must always be a string. 

## Expected Behavior

Regardless if "foreign" bpf programs have been applied, Calico should be able to cope and not stumble upon this. 

## Current Behavior

Example output:
```
2023-12-05 11:27:55.854 [WARNING][9529] felix/bpf_ep_mgr.go 1846: Failed to list previously attached programs. We may not clean up some. error=cannot parse json output: json: cannot unmarshal number into Go struct field ProgInfo.type of type string
[{
        "id": 8,
        "type": 26,
        "name": "tcp_connect",
        "tag": "1593e302b3115137",
        "gpl_compatible": true,
        "loaded_at": 1701773685,
        "uid": 0,
        "bytes_xlated": 2752,
        "jited": true,
        "bytes_jited": 1830,
        "bytes_memlock": 4096,
        "map_ids": [1
        ]
    },{
        "id": 39,
        "type": "sched_cls",
        "name": "calico_tc_skb_a",
        "tag": "a97a0a4e8b4470e5",
        "gpl_compatible": true,
        "loaded_at": 1701773693,
        "uid": 0,
        "bytes_xlated": 15440,
        "jited": true,
        "bytes_jited": 10242,
        "bytes_memlock": 16384,
        "map_ids": [3,15,14,27,28,4,10,9,13
        ],
...
```
The message suggests that Calico assumes that the type field always is a string, but for this program it is a numeric. 
In this example, it is a program created with bpftrace. 26 likely is `BPF_PROG_TYPE_RAW_TRACEPOINT_WRITABLE` but I am not 100% sure about that. Maybe the fact that a numeric type is returned is caused by an older Kernel or an outdated libbpf. 

## Possible Solution

If it is assured that Calico's own programs always have a string type, maybe always `sched_cls`, it would be fair for Calico to only handle those, but it shouldn't fail for other programs. 

## Steps to Reproduce (for bugs)

1. Add a bpf program using bpftrace, I could attach the one we are adding, tracing tcp-connects
2. restart felix/calico-node to let it process all registered bpf programs

## Context

We did not really observe an issue with this, we just noticed the warning in the logs. From the warning, it is not fully clear if this has a negative impact on calico or not. 

[...]

PR #8295, opened 2023-12-08
[BPF] don't stumble on unknown prog types passed as int in json

tomastigera:
bpftool may be older than the kernel and some program types may not be know and so only the number instead of the string is emitted in the json.

fixes https://github.com/projectcalico/calico/issues/8293

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: don't stumble on unknown prog types passed as int in json
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/bpf.go
+++ b/felix/bpf/bpf.go
@@ -396,12 +396,36 @@ type perCpuMapEntry []struct {
 }
 
 type ProgInfo struct {
-	Name   string `json:"name"`
-	Id     int    `json:"id"`
-	Type   string `json:"type"`
-	Tag    string `json:"tag"`
-	MapIds []int  `json:"map_ids"`
-	Err    string `json:"error"`
+	Name   string      `json:"name"`
+	Id     int         `json:"id"`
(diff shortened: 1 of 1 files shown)


felix/bpf/bpf.go
```diff
@@ -396,12 +396,32 @@ type perCpuMapEntry []struct {
...
+// fail on parsing them.
+type BPFProgType string
+
+func (t *BPFProgType) UnmarshalJSON(data []byte) error {
+	if string(data) == "null" || string(data) == `""` || len(data) < 1 {
+		return nil
+	}
+
+	if data[0] == '"' {
+		s := string(data)[1 : len(data)-1]
```
fasaxc, line 418: Probably OK here but this won't handle escape codes and the like.  I'd be temped to hand off to the main JSON decoder just so I don't have to worry.

tomastigera: could do that, although we know what the input is,don't we? We only parse bpftool.

fasaxc (approved):

PR #8307, opened 2023-12-11
Update upstream CNI plugins and Flannel downloads to latest golang patches

matthewdupre:
Also parametrizes the Flannel artifacts URL and points it to our patched fork: projectcalico/flannel-cni-plugin


Behnam-Shobiri (approved):
lgtm

PR #8324, opened 2023-12-12
Update release tooling and configuration

danudey:
Porting forward the release tooling fixes and updates from v3.26.

### DIFF
--- a/hack/release/Makefile
+++ b/hack/release/Makefile
@@ -10,6 +10,21 @@ release: .vm.created
 	./remote-execute.sh "cd calico && git fetch --tags && git checkout $(GIT_BRANCH)"
 	./remote-execute.sh "cd calico && make release"
 
+git-fetch-remote: .vm.created
+	./remote-execute.sh "cd calico && git fetch --tags && git checkout $(GIT_BRANCH)"
+
+update-versions-remote: .vm.created
+	./remote-execute.sh "cd calico && make update-versions"
+
+generate-manifests-remote:
+	./remote-execute.sh "cd calico && make generate"
+
+commit-versions-remote: .vm.created
+	./remote-execute.sh "cd calico && make commit-versions-manifests"
+
+build-release-remote: .vm.created
+	./remote-execute.sh "cd calico && make release"
+
 # Expects that a release has already been created on the remote VM via the "release" target.
 release-publish: .vm.created
 	./remote-execute.sh "cd calico && make release-publish"
--- a/hack/release/Makefile
+++ b/hack/release/Makefile
@@ -16,7 +16,7 @@ git-fetch-remote: .vm.created
 update-versions-remote: .vm.created
 	./remote-execute.sh "cd calico && make update-versions"
 
-generate-manifests-remote:
+generate-manifests-remote: .vm.created
 	./remote-execute.sh "cd calico && make generate"
 
 commit-versions-remote: .vm.created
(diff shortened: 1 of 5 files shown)


hack/release/Makefile
```diff
@@ -10,6 +10,21 @@ release: .vm.created
 	./remote-execute.sh "cd calico && git fetch --tags && git checkout $(GIT_BRANCH)"
 	./remote-execute.sh "cd calico && make release"
 
+git-fetch-remote: .vm.created
+	./remote-execute.sh "cd calico && git fetch --tags && git checkout $(GIT_BRANCH)"
+
+update-versions-remote: .vm.created
+	./remote-execute.sh "cd calico && make update-versions"
+
+generate-manifests-remote:
```
matthewdupre, line 19: Should this depend on .vm.created?

matthewdupre: (Might not matter much if this isn't the first step to be ran)

matthewdupre (approved):

hack/release/Makefile
danudey: Yes, it should. Thanks!

hack/release/pkg/builder/builder.go
```diff
@@ -288,7 +288,7 @@ func (r *ReleaseBuilder) publishPrereqs(ver string) error {
 //
 // - release-vX.Y.Z.tgz: contains images, manifests, and binaries.
 // - tigera-operator-vX.Y.Z.tgz: contains the helm v3 chart.
-// - calico-windows-vX.Y.Z.zip: Calico for Windows zip archive for non-HPC installation.
+// - calico-windows-vX.Y.Z.zip: Calico for Windows.
```
coutinhop, line 291: I think we need to revert this for v3.27 (which is when HPC support begins)

hack/release/pkg/builder/builder.go
```diff
@@ -445,33 +444,17 @@ func (r *ReleaseBuilder) buildContainerImages(ver string) error {
...
-		out, err := r.makeInDirectoryWithOutput(dir, "release-build", env...)
-		if err != nil {
-			logrus.Error(out)
-			return fmt.Errorf("Failed to build %s: %s", dir, err)
-		}
-		logrus.Info(out)
-	}
-
-	for _, dir := range windowsReleaseDirs {
-		out, err := r.makeInDirectoryWithOutput(dir, "image-windows", env...)
```
coutinhop, line 469: I think we need to revert this for v3.27

hack/release/pkg/builder/builder.go
```diff
@@ -533,11 +516,6 @@ func (r *ReleaseBuilder) publishContainerImages(ver string) error {
 		"node",
 	}
 
-	windowsReleaseDirs := []string{
-		"node",
-		"cni-plugin",
-	}
-
```
coutinhop, line 540: I think we need to revert this for v3.27

hack/release/pkg/builder/builder.go
```diff
@@ -568,25 +546,6 @@ func (r *ReleaseBuilder) publishContainerImages(ver string) error {
...
-				}
-				logrus.Error(out)
-				return fmt.Errorf("Failed to publish %s: %s", dir, err)
-			}
-
-			// Success - move on to the next directory.
-			logrus.Info(out)
-			break
-		}
-	}
```
coutinhop, line 589: I think we need to revert this for v3.27

coutinhop (commented):
We'll start supporting HPC and will push the `calico/node-windows` and `calico/cni-windows` images on v3.27, made a few comments on the code I don't think should have been deleted

Issue #4689, opened 2021-06-21
NodeLocal DNS with Calico/BPF


lluczaj:
We have an EKS cluster with Calico/BPF enabled (no Kube-proxy). Before BPF we used to run NodeLocal DNSCache in our clusters and it worked fine:

https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/

However, we have issues running it in a cluster with Calico/BPF mode enabled. Traffic is not forwarded to node-local-dns at all in this case. In docs there is only Kube-proxy/IPVS mode described. Cilium guys provide a solution with LocalRedirectPolicy in their docs for BPF mode but I haven't found anything for Calico. Is there any suggested solution for this for Calico BPF ?

## Your Environment

* Calico version: 3.19.1
* Orchestrator version (e.g. kubernetes, mesos, rkt): Kubernetes 1.20 / EKS
* Operating System and version: Amazon Linux 2 (5.4.117-58.216.amzn2.x86_64)

song-jiang:
@fasaxc Could you help?

fasaxc:
No workaround known right now. node-local DNS uses iptables hacks to bypass kube-proxy.  However, BPF bypasses iptables so we partially bypass the node-local DNS hacks too.

lluczaj:
Thanks for your answer @song-jiang @fasaxc 
We are thinking about workaround with injecting node-local DNS local IP directly in Kubelet's config so it populates it in resolv.conf files for pods. This way they will communicate directly with node-local DNS without any iptables hacks. Is there any potential issue with this approach from Calico-BPF perspective ?

One thing that we noticed is that tigera-operator cannot communicate with EKS cluster-endpoint in BPF mode when we inject node-local DNS ip into Kubelet and node-local DNS is not yet deployed (timeouts). Then when it is deployed into cluster tigera-operator starts to work fine but cluster DNS overall is not yet working properly as there are no calico-node pods and CoreDNS is still in Pending state therefore cannot handle any requests. So why does it matter ?

fasaxc:
> We are thinking about workaround with injecting node-local DNS local IP directly in Kubelet's config so it populates it in resolv.conf files for pods. This way they will communicate directly with node-local DNS without any iptables hacks. Is there any potential issue with this approach from Calico-BPF perspective ?

If you can get it to talk directly to the host on one of the host's IPs then I'd expect that to work.

zbkannan:
[...]

PR #8338, opened 2023-12-14
[BPF] BPFExcludeCIDRsFromNAT allows excluding service IPs from being NATed

tomastigera:
The main use case is node-local dns cache. In iptables mode with k8s provided kube-proxy, the cache installs such iptables rules that snatch the traffic before the kube-proxy rules, which would do the NAT.

However, in ebpf NAT is done way before packets may hit iptables. Often they do not hit iptables at all. Therefore, we need to tell our kube-proxy, not to translate a service IP that we want to let dns cache to process.

Unfortunatelly, we cannot annotate the service as the annotations are not propagate to the kube-proxy code by the k8s front-end.

To setup a node-local dns cache, set BPFExcludeCIDRsFromNAT as follows:

kubedns=`kubectl get svc kube-dns -n kube-system -o jsonpath={.spec.clusterIP}`

BPFExcludeCIDRsFromNAT=$kubedns

fixes https://github.com/projectcalico/calico/issues/4689

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: setting BPFExcludeIPsFromNAT allows node-local dns cache to work
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -505,6 +506,22 @@ func newBPFEndpointManager(
...
+
+		var excludeIPsMatch = 1
+
+		for _, c := range config.BPFExcludeIPsFromNAT {
+			cidr, err := ip.CIDRFromString(c)
+			if err != nil {
+				log.WithError(err).Warnf("Bad %s CIDR to exclude from NAT", c)
+			}
+
+			if int(cidr.Version()) == 6 && !m.ipv6Enabled {
```
fasaxc, line 519: What about the reverse, IPv4 CIDR and IPv6 support enabled?  
```suggestion
			if (cidr.Version() == 6) != m.ipv6Enabled {
```
do you need the cast?

felix/bpf-gpl/tc.c
```diff
@@ -407,6 +407,12 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 	} else {
 		ctx->state->post_nat_ip_dst = ctx->state->ip_dst;
 		ctx->state->post_nat_dport = ctx->state->dport;
+		if (nat_res == NAT_EXCLUDE) {
+			/* We want such packets to go through the host namespace. The main
+			 * usecase of this is node-local-dns.
+			 */
+			ctx->state->flags |= CALI_ST_SKIP_FIB;
```
fasaxc, line 414: Do we need to mark the conntrack entry too?  Otherwise only the first packet will skip the FIB.

fasaxc (changes requested):

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -549,6 +549,10 @@ type FelixConfigurationSpec struct {
 	// BPFDisableGROForIfaces is a regular expression that controls which interfaces Felix should disable the
 	// Generic Receive Offload [GRO] option.  It should not match the workload interfaces (usually named cali...).
 	BPFDisableGROForIfaces string `json:"bpfDisableGROForIfaces,omitempty" validate:"omitempty,regexp"`
+	// BPFExcludeIPsFromNAT is a list of CIDRs that are to be excluded from NAT
+	// resolution so that host can handle them. A typicall usecase is node local
+	// DNS cache.
+	BPFExcludeIPsFromNAT *[]string `json:"bpfExcludeIPsFromNAT,omitempty" validate:"omitempty,cidrs"`
```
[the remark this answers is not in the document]
tomastigera: yeah agree, noticed it myself too :sweat_smile:

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: Atm if `m.ipv6Enabled` then ipv4 should not be considered. As part of the dual stack effort, this will need to be addressed, but idk how that is going to pan out yet.

felix/bpf-gpl/tc.c
tomastigera: Correct, that is done where we create the ct entry here. That is an already established process. https://github.com/tomastigera/project-calico-calico/blob/tomas-bpf-node-local-dns/felix/bpf-gpl/tc.c#L1231-L1232

marvin-tigera:

fasaxc (approved):

Issue #5311, opened 2021-12-18
Don't rely on BGP in public clouds


caseydavenport:
## Expected Behavior

When running in a public cloud, we shouldn't need BGP topologies at all. BGP should be an optional integration point for on-prem clusters.

Felix should program routes for IPIP and unencapsualted mode, as it does VXLAN today. BIRD can be selectively enabled for integration with third-party devices. 

## Current Behavior

BGP is required for route distribution, even within Calico itself, unless using VXLAN mode. This makes scaling the networking layer a bit trickier than needed, due to things like route reflector topologies: https://github.com/projectcalico/calico/issues/2382

## Possible Solution

Have felix program these routes instead of BIRD.

mazdakn:
@caseydavenport great to mention unecapsulated traffic. While working on making Felix program IPIP routes, I was think exactly that we should do the same for unencap one as well.

mazdakn:
Felix can now program all in cluster routes:
No encap: https://github.com/projectcalico/calico/pull/10472
IPIP: https://github.com/projectcalico/calico/pull/10291

PR #8357, opened 2023-12-19
Felix programs IPIP tunnels

mazdakn:
## Description

Currently, it's BIRD that programs routes for IPIP encapsulation, which is unlike VXLAN|Wireguard where Felix programs the routes. This PR enables Felix to programs the IPIP routes as well, and not rely on BIRD anymore. The benefits are:
- Not needing an extra component (i.e. BIRD) for IPIP encapsulation. Enabling Felix to program IPIP routes and not running BIRD saves some memory/CPU.
- Allowing to get rid of Calico fork of BIRD (v1.6) which is way old. Instead we can use upstream BIRD.
- In future, we can also make Felix program no encap routes as well. and get rid of internal BGP mesh.

This PR includes:
- Adding support to Felix to program IPIP routes.
- Updating BIRD templates to not export IPIP routes, i.e routes with `tunl0` interface.
- Updating BIRD templates to not program IPIP routes to kernel
- Introducing a new value of `ipip` for `calico_backend` variable to enable IPIP encapsulation without running BIRD. `bird` value still enables IPIP encapsulation and also start BIRD service.

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
Whooaaaa wasn't expecting this one. Sweet!

felix/dataplane/linux/int_dataplane.go
```diff
@@ -807,10 +809,33 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
+		dp.ipipManager = newIPIPManager(
+			ipSetsV4,
+			routeTableIPIP,
+			IPIPIfaceName,
+			config,
+			dp.loopSummarizer,
+			4,
+			featureDetector,
+		)
+		dp.ipipParentC = make(chan string, 1)
```
fasaxc, line 837: Does IPIP have the concept of a parent device?  I thought that was a quirk of VXLAN.  I'd expect that aspect not to be needed.  Of course, if we do the IPIP device set-up in the background then we still need a kick channel but I don't think you need to send the parent device?

fasaxc (commented):
Had a very brief skim of the felix bits.  Nice to see that it's not a crazy amount of code in Felix; as we'd hoped.  Before looking in more detail/at the confd stuff it'd be good to understand the design a bit better:

* If I create a new cluster in non-bird mode it seems simple: I get felix routes and no BIRD.
* But what if I then decide to advertise routes later?  Do we disable Felix routes in favour of BIRD?  How do BIRD and Felix interact?
* Is there a way to get the exact old behaviour (i.e. disable Felix new function and let BIRD do what it always did)?  I'd quite like to have that as an option in case folks are doing odd things that we don't know about!  In simple cases, felix will program the same routes, but I bet some folks are peering with external nodes and doing IPIP with those external nodes!
* How is upgrade/downgrade handled? BIRD and felix will use different protocols on their routes, will Felix overwrite a BIRD route?  Will Felix clean up a BIRD route that's now no longer needed?  (Doing the latter would be easier once my big RouteTable refactor PR lands since there's a way to handle special cases like that.)

felix/dataplane/linux/int_dataplane.go
mazdakn: AFAIK, the parent device is needed to add routes to no encap routes. And for IPIP `CrossSubnet` we need to add no encap routes, right?

fasaxc:
Think you said this wasn't ready for review yet (and I still need to review your doc).  Sorry for being slow!

frozenprocess:
looks related https://github.com/projectcalico/calico/issues/5311

mazdakn:
Closing in favour of https://github.com/projectcalico/calico/pull/10291

mazdakn:
Closing in favour of https://github.com/projectcalico/calico/pull/10291

Issue #8162, opened 2023-10-25
BGP not announcing LoadBalancer services with externalTrafficPolicy Local for prefixes /32 and using route-reflectors


AMacedoP:
When announcing LoadBalancer services using BGP and route-reflectors, services using `externalTrafficPolicy: Local` are not announced if the pods are not running in the same nodes as the route-reflectors.

This looks like a edge case not covered in #6074 because we are also using /32 prefixes in BGPConfiguration.

## Expected Behavior

LoadBalancer service IPs with `externalTrafficPolicy: Local` are announced by the route-reflectors to external network devices

## Current Behavior

LoadBalancer service IPs are not announced by the route-reflectors, unless the route-reflector node has a pod that matches the service selectors

## Possible Solution
Route reflectors nodes should announce all service IPs with `externalTrafficPolicy: Local` regardless of whether a pod is scheduled there

## Steps to Reproduce (for bugs)

1. Setup a cluster with a node as a route reflector and disable to node-to-node mesh according to the [documentation](https://docs.tigera.io/calico/latest/networking/configuring/bgp#configure-a-node-to-act-as-a-route-reflector)
1. Configure the default BGPConfiguration to announce LoadBalancer IPs with /32 prefixes:
   ```yaml
   apiVersion: projectcalico.org/v3
   kind: BGPConfiguration
   metadata:
     name: default
   spec:
     asNumber: 65501
	 listenPort: 179
	 logSeverityScreen: Info
	 nodeToNodeMeshEnabled: false
	 serviceLoadBalancerIPs:
	 - cidr: "10.100.17.223/32"
	 - cidr: "10.100.17.72/32"
	 - cidr: "10.100.18.30/32"
   ```
1. Add an external network device as a BGPPeer only for the route-reflectors
   ```yaml
   apiVersion: projectcalico.org/v3
   kind: BGPPeer
   metadata:
     name: external-device
   spec:
     peerIP: 20.0.0.6
     asNumber: 65501
     nodeSelector: has(route-reflector)
   ```
1. Create a deployment and service with `externalTrafficPolicy: Local` and check that the pod was scheduled to a node not configured as a route reflector
1. Check that the new service is not present in the routing table of the external network device

## Your Environment

* Calico version: v.3.25.1
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes
* Operating System and version: Ubuntu 20.04

louygan:
[...]

PR #8358, opened 2023-12-20
Fix announcement of Single LB IPs with Route Reflectors.

AMacedoP:
## Description
This commit fixes a bug where services of type LoadBalancer, externalTrafficPolicy Local and IPs /32 or /128 where not announced correctly via the RouteReflector when the service doesn't have and endpoint in the RouteReflector node.

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/8162

## Todos

- [X] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Route reflector nodes now properly advertise Service LoadBalancer IP addresses even if there is no local endpoint on the node.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/confd/etc/calico/confd/templates/bird6_ipam.cfg.template
+++ b/confd/etc/calico/confd/templates/bird6_ipam.cfg.template
@@ -38,6 +38,20 @@ function calico_export_to_bgp_peers(bool internal_peer) {
   if ( net ~ {{$cidr}} ) then { accept; }
   {{- end}}
 {{- end}}
+{{- $rr_cluster_id_key := printf "/bgp/v1/host/%s/rr_cluster_id" (getenv "NODENAME")}}
+{{- $lb_ips := "/bgp/v1/global/svc_loadbalancer_ips"}}
+{{- if exists $rr_cluster_id_key}}{{$rr_cluster_id := getv $rr_cluster_id_key}}
+{{- if and (not (eq $rr_cluster_id "")) (exists $lb_ips)}}
+
+  # Configured as a RR - accept any routes within configured LB service IP ranges
+  {{- range split (getv $lb_ips) ","}}
+    {{- $cidr := .}}
+    {{- if contains $cidr ":"}}
+  if ( net ~ {{$cidr}} ) then { accept; }
+    {{- end}}
+  {{- end}}
+{{- end}}
+{{- end}}
 {{range ls "/v1/ipam/v6/pool"}}{{$data := json (getv (printf "/v1/ipam/v6/pool/%s" .))}}
 {{- if $data.disableBGPExport}}
   # Skip {{$data.cidr}} as BGP export is disabled for it
--- a/node/tests/k8st/tests/test_bgp_advert.py
+++ b/node/tests/k8st/tests/test_bgp_advert.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 Tigera, Inc. All rights reserved.
+# Copyright (c) 2018-2024 Tigera, Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


caseydavenport:
@AMacedoP thanks much for this PR! I am in the middle of reviewing it now. So far it is looking pretty good, although I have some simplifications I am mulling on.

I did have a thought while digging into this a bit more that perhaps this case is already solvable using existing configuration expressible in the Calico API. For example, you should be able to enable export by creating a disabled IP pool that encompasses your Service LB IP range. This is probably tedious in your particular case though, as you have many `/32` ranges instead of allocating a single encompassing range. So I think something like what's in this PR is still in order!

caseydavenport:
The main simplification I am wondering (and this is only half thought through at the moment) is whether we need to populate the `/allowlist` key dynamically based on routes, or if we could simply set it based on the `BGPConfiguration` itself.

For example, if we had a BGPConfiguration that looked like this:

```yaml
spec:
  serviceLoadBalancerIPs:
  - cidr: "10.0.0.0/24"
```

We wouldn't need an export filter for each service IP - we could simply allow export for the entire `/24` block.

Similarly for your case, if we had:

```yaml
spec:
  serviceLoadBalancerIPs:
  - cidr: "10.100.17.223/32"
  - cidr: "10.100.17.72/32"
  - cidr: "10.100.18.30/32"
```

We could configure our export filters to allow all three of those addresses without waiting to learn about the actual Service objects in the cluster. I think this would cut back on a lot of the bookkeeping complexity needed / reference counting / etc. and simplify the code, and is pretty safe because the code already expects that any addresses included here are reserved for use as service LB IPs.

The code could likely be implemented entirely within `updateBGPConfigCache` instead of plumbing through the route generator, or even [getServiceExternalIPsKVPair](https://github.com/projectcalico/calico/blob/93c577af291bd9ab28daa479d994211256a71abf/confd/pkg/backends/calico/client.go#L1263-L1278) WDYT?

AMacedoP:
Yeah, that sounds like a good idea. Do you think there'll be any problems if the same IP shows up in the bird filter multiple times? Because a service IP will be added by the route generator if the service is announced in that node and then it'll be added by the allowlist if the node is a RR. I think there shouldn't be any problems but I'm not an expert in BIRD so I'm not sure.

I'll change to PR to use this new method and ping you again when it's done.

Thanks for your review!

caseydavenport:
@AMacedoP a fair question - I don't forsee that being a problem though. I believe BIRD needs to be able to handle that scenario. If not then I think we can still do this based on BGPConfiguration, although it might require a little bit more reference counting to ensure we only add it once. That's probably a small improvement we could make in the future if needed.

AMacedoP:
Hi @caseydavenport, sorry for the delayed response to this PR.

I implemented your suggested change but I found out that the `updateBGPConfigCache` function isn't called when the calico-node pod starts up and it needs that someone changes the default `BGPConfiguration` via calicoctl before it's called. This is a problem because the `allowlist` configuration will be lost when the calico-node pod restarts. Is it intended behaviour?

caseydavenport:
> isn't called when the calico-node pod starts up and it needs that someone changes the default BGPConfiguration

That doesn't seem expected to me. I would expect this to be called whenever a change is made to BGPConfig, but also on startup (assuming there is a BGPConfig in the cluster))

AMacedoP:
@caseydavenport for some reason I'm not seeing that weird behavior when I tested it again, maybe it was a mistake on my end. I'm pushing the simpler implmentation we discussed, can you review it? I'm not sure how to test it so can you give some pointers on it?

Thanks for your time

confd/pkg/backends/calico/client.go
```diff
@@ -1279,15 +1289,24 @@ func (c *client) getServiceExternalIPsKVPair(v3res *apiv3.BGPConfiguration, key
...
 		}
 		c.updateCache(api.UpdateTypeKVUpdated, getKVPair(svcLoadBalancerIPKey, strings.Join(ipCidrs, ",")))
+
+		if c.IsRouteReflector(c.rg.nodeName) {
+			c.updateCache(api.UpdateTypeKVUpdated, getKVPair(allowListKey, strings.Join(ipCidrs, ",")))
+		}
 	} else {
 		c.updateCache(api.UpdateTypeKVDeleted, getKVPair(svcLoadBalancerIPKey))
+
+		if c.IsRouteReflector(c.rg.nodeName) {
```
caseydavenport, line 1307: This might be a non-issue in practice, but I think we should handle it regardless.

The scenario I'm thinking of:
- Start as a RR
- Update the cache with allowlist
- Disable the node as a RR
- Cache allowlist is no longer updated.

I think we might need similar checks to update the cache with the `allowlist` key when we switch from node -> RR and vice-versa.

Alternatively, we could always unconditionally update the allowlist key, and then in the template only render this section if we're a RR.

caseydavenport:
As far as testing goes, there are probably two places to look.

1. We have a suite of tests that runs a small local cluster and verifies service advertisement. For example: https://github.com/projectcalico/calico/blob/086c2754a422f0631622493a9ecf9a63aeb0c9c8/node/tests/k8st/tests/test_bgp_advert.py#L780-L795
2. There are tests within `confd/tests` that verify properly rendered BIRD configs. Here's an example of where they are called from: https://github.com/projectcalico/calico/blob/086c2754a422f0631622493a9ecf9a63aeb0c9c8/confd/tests/test_suite_common.sh#L580-L581

Adding the confd test is a little bit tedious, but you can model it off the existing `mesh/static-routes` tests. You'll need to:
- Add the test to `test_suite_common.sh`
- Add a new directory under `tests/mock_data/calicoctl/mesh/` which includes the yaml files to apply / delete for the test.
- Run the tests with UPDATE_EXPECTED_DATA=true, which will produce BIRD config output.

confd/pkg/backends/calico/client.go
AMacedoP: What do you think about adding an else statement to the `isRouteReflector` check and deleting the `allowlist` cache entry if it's not.

Something like

```go
		if c.IsRouteReflector(c.rg.nodeName) {
			c.updateCache(api.UpdateTypeKVUpdated, getKVPair(allowListKey, strings.Join(ipCidrs, ",")))
		} else {
			c.updateCache(api.UpdateTypeKVDeleted, getKVPair(allowListKey))
		}

```

I'm pretty sure this will handle all the cases you mentioned

caseydavenport: I think the only problem with that is that the RouteReflector state is carried by the `Node` update, whereas the Service IPs are carried by BGPConfiguration. This means that any node can switch between being a RR and a normal node without this function even being executed (since it's only executed for BGPConfig updates)

AMacedoP:
Hi @caseydavenport, at the end it was way easier than I thought. I only needed to modify the templates because there were 2 cache entries that I could use to solve original issue and all the scenarios you mentioned:

1. `bgp/v1/host/${node}/rr_cluster_id` to check if the node is a RR, and it's updated by the RouteReflector annotation
2. `bgp/v1/global/svc_loadbalancer_ips` to get the list of CIDRs configured to be advertised from the BGPConfiguration CR

I also added tests for IPv4 and IPv6 advertisement to the `node` module.

caseydavenport:
@AMacedoP Just getting back from three weeks of vacation, looking at this now! Thanks for your patience :heart:

confd/etc/calico/confd/templates/bird6_ipam.cfg.template
```diff
@@ -38,6 +38,19 @@ function calico_export_to_bgp_peers(bool internal_peer) {
   if ( net ~ {{$cidr}} ) then { accept; }
   {{- end}}
 {{- end}}
+{{- $is_rr_key := printf "/bgp/v1/host/%s/rr_cluster_id" (getenv "NODENAME")}}
+{{- $lb_ips := "/bgp/v1/global/svc_loadbalancer_ips"}}
+{{- if exists $is_rr_key}}{{$is_rr := getv $is_rr_key}}
```
caseydavenport, line 43: nit, but I think `is_rr` should be named `rr_cluster_id` since `is_rr` implies a boolean value, but actually the value here is the RR cluster ID (if set, an empty otherwise)

confd/etc/calico/confd/templates/bird6_ipam.cfg.template
```diff
@@ -38,6 +38,19 @@ function calico_export_to_bgp_peers(bool internal_peer) {
...
+{{- if and (not (eq $is_rr "")) (exists $lb_ips)}}
+
+  {{- range split (getv $lb_ips) ","}}
```
caseydavenport, line 46: I think we should have a comment here that gets rendered in the resulting config files to make them easier to read / debug. Similar to the comment on line 34 ("Export static routes")

e.g.,

```
# Configured as a RR - accept any routes within configured LB service IP ranges
```

node/tests/k8st/tests/test_bgp_advert.py
```diff
@@ -1,4 +1,3 @@
-# Copyright (c) 2018 Tigera, Inc. All rights reserved.
```
caseydavenport, line 1: I think this should stay, but be updated to say "2018-2024"

confd/etc/calico/confd/templates/bird6_ipam.cfg.template
AMacedoP: Fair enough, will change

confd/etc/calico/confd/templates/bird6_ipam.cfg.template
AMacedoP: In the first iterations I added something like that, but noticed that this message will be printed in the IPv6/IPv4 file even if there are no IPV6/IPv4 routes in the `lb_ips` variables as there's no distinct variables for each protocol. I can add it back but would it be a problem?

node/tests/k8st/tests/test_bgp_advert.py
AMacedoP: Deleted this line by mistake, I'll add it back

AMacedoP:
Hopefully the last set of changes!

caseydavenport (approved):

caseydavenport:
I think this is a candidate for back-porting, yeah - would you mind opening a cherry-pick PR against release-v3.27 and tagging me on it?

PR #8369, opened 2023-12-21
Move certificates permissions out of the else-block.

rene-dekker:
Some RBAC rules were mistakenly inside of the else block of the if statement, rather than outside of the statement, causing them not to be present in openshift clusters. As a result, the watch for CSRs cannot be setup in the operator.


tomastigera (approved):
LGTM

Issue #7252, opened 2023-01-30
KubeSpray - Cannot access local node services when using eBPF


Aslan-Liu:
I have one cluster and also install Prometheus in it. So, each node has a service (HostIP:9100) to export node information. However, if I run one Pod on Node1(Node1 Host IP: 172.21.149.119), I cannot access 172.21.149.119:9100 in the Pod.  But I can access other services on the other nodes, such as 172.21.149.xx:9100 .

## Expected Behavior
All local serivces on every node can be accessed in the Pod.

## Current Behavior
Now, only local services run on different nodes with my Pod can be accessed in my Pod.

## Possible Solution
https://github.com/projectcalico/calico/issues/6065

## Steps to Reproduce (for bugs)
1. Install Calico v3.23.1 or v3.22.3
2. Install Prometheus
3. Run a Pod on any node
4. In the Pod, use curl to access HostIP:9100 (HostIP is the node IP where the Pod runs on.)

## Context
I also can see the following log messages in calico-node Pod.
```
libbpf: prog 'calico_connect_v4': failed to attach to cgroup: Invalid argument
2023-01-30 12:01:50.409 [INFO][129] felix/connecttime.go 146: Loaded cgroup program cgroup="/run/calico/cgroup" program="calico_connect_v4"
libbpf: prog 'calico_sendmsg_v4': failed to attach to cgroup: Invalid argument
2023-01-30 12:01:50.499 [INFO][129] felix/connecttime.go 146: Loaded cgroup program cgroup="/run/calico/cgroup" program="calico_sendmsg_v4"
libbpf: prog 'calico_recvmsg_v4': failed to attach to cgroup: Invalid argument
2023-01-30 12:01:50.501 [INFO][129] felix/connecttime.go 146: Loaded cgroup program cgroup="/run/calico/cgroup" program="calico_recvmsg_v4"
libbpf: prog 'calico_sendmsg_v6': failed to attach to cgroup: Invalid argument
2023-01-30 12:01:50.502 [INFO][129] felix/connecttime.go 146: Loaded cgroup program cgroup="/run/calico/cgroup" program="calico_sendmsg_v6"
libbpf: prog 'calico_recvmsg_v6': failed to attach to cgroup: Invalid argument
```
I think this problem is very similar as #6065. However, the problem still exists after upgrading Calico to v3.23.1.

## Your Environment
* Calico version: v3.25.1
[...]

PR #8388, opened 2024-01-04
[BPF] Do not apply normal *-hep policy to wep - align behaviour with iptables

tomastigera:
## Description

If there is no wildcard HEP, there is no policy that should be applied. But without skipping, empty list of profiles would create a default deny rule if none of the non-existent profiles matches. That is obviously always hit and traffic toward the host is dropped if defaultEndpointToHostAction is set to RETURN.

In fact, to be compatible with iptables behaviour, we should never apply normal *-hep policy, only the preDNAT one. This PR leaves the normal policy code there for enabling it in the future.

Also includes FV tests for defaultEndpointToHostAction by @fasaxc that codify behaviour and unify it across dataplanes.

Removed blanket ACCEPT in INPUT chain, only wg traffic is explicitly accepted now, the rest is dropped if INPUT policy is DROP and defaultEndpointToHostAction==RETURN.
## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/7252
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: align defaultEndpointToHostAction with iptables - do not apply *-hep policy to wep
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -1965,8 +1965,9 @@ func (m *bpfEndpointManager) wepApplyPolicy(ap *tc.AttachPoint,
 	}
 
 	// If workload egress and DefaultEndpointToHostAction is ACCEPT or DROP, suppress the normal
-	// host-* endpoint policy.
-	if polDirection == PolDirnEgress && m.epToHostAction != "RETURN" {
+	// host-* endpoint policy. If it does not exist, suppress it as well, not to
+	// create deny due to the fact that there are not profiles or tiers etc.
+	if polDirection == PolDirnEgress && (m.epToHostAction != "RETURN" || !m.wildcardExists) {
 		rules.SuppressNormalHostPolicy = true
 	}
 
@@ -2966,6 +2967,7 @@ func (m *bpfEndpointManager) loadPolicyProgram(
 	log.WithFields(log.Fields{
 		"progName": progName,
 		"ipFamily": ipFamily,
+		"rules":    rules,
 	}).Debug("Generating policy program...")
 
 	if ipFamily == proto.IPVersion_IPV6 {
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -1964,10 +1964,16 @@ func (m *bpfEndpointManager) wepApplyPolicy(ap *tc.AttachPoint,
 		m.addHostPolicy(&rules, &m.wildcardHostEndpoint, polDirection.Inverse())
 	}
 
-	// If workload egress and DefaultEndpointToHostAction is ACCEPT or DROP, suppress the normal
-	// host-* endpoint policy. If it does not exist, suppress it as well, not to
-	// create deny due to the fact that there are not profiles or tiers etc.
-	if polDirection == PolDirnEgress && (m.epToHostAction != "RETURN" || !m.wildcardExists) {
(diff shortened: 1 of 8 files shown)


marvin-tigera:

marvin-tigera:

marvin-tigera:

felix/fv/bpf_test.go
```diff
@@ -631,6 +631,18 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 				})
 			})
 
+			Describe("with DefaultEndpointToHostAction=RETURN", func() {
```
fasaxc, line 634: This is only right if the iptables INPUT chain policy is ACCEPT.  To be sure it's doing the right thing, we need to test both cases of the iptables policy.

tomastigera: yes, that is another step in fixing a broader issue

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1965,8 +1965,9 @@ func (m *bpfEndpointManager) wepApplyPolicy(ap *tc.AttachPoint,
 	}
 
 	// If workload egress and DefaultEndpointToHostAction is ACCEPT or DROP, suppress the normal
-	// host-* endpoint policy.
-	if polDirection == PolDirnEgress && m.epToHostAction != "RETURN" {
+	// host-* endpoint policy. If it does not exist, suppress it as well, not to
+	// create deny due to the fact that there are not profiles or tiers etc.
+	if polDirection == PolDirnEgress && (m.epToHostAction != "RETURN" || !m.wildcardExists) {
 		rules.SuppressNormalHostPolicy = true
```
fasaxc, line 1971: Given what I discovered about iptables mode, shouldn't we _always_ suppress the normal HEP ingress policy?  We should render the HEP pre-DNAT policy only?

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1564,15 +1564,13 @@ func (d *InternalDataplane) setUpIptablesBPF() {
...
-		// Accept any SEEN packets that go to the host. What remains here should be from
-		// host interfaces. This should accept packets in default-DROP iptable
-		// environments.  Such packets go through BPF on host iface but must go through
-		// the INPUT of iptables too. default-DROP would block IPIP/VXLAN/WG/etc. traffic
-		// without explicit ACCEPT.
-		inputRules = append(inputRules, iptables.Rule{
-			Match:  iptables.Match().MarkMatchesWithMask(tcdefs.MarkSeen, tcdefs.MarkSeenMask),
-			Action: iptables.AcceptAction{},
-		})
+		if rulesConfig.EndpointToHostAction != "ACCEPT" {
```
fasaxc, line 1567: What about traffic arriving on eth0, i.e. traffic to the host but not from a workload?  Will that get caught by this rule change?

tomastigera: In that case, if INPUT is DROP it will get dropped unless it is wireguard. afaict it is what would happen in iptables. It is covered by the wg tests. If default hepToHost is ACCEPT, it will get accepted by other rules sooner.

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: Is that case covered by the new fv test?

tomastigera: wasn't covered, added those test cases

fasaxc (approved):

PR #8390, opened 2024-01-04
cni-plugin: some code optimization

lucming:
## Description
just some code optimization

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
@lucming appreciate the PR! I'm going to give this a run through CI. From a code read it looks functionally equivalent to me.

caseydavenport:
@lucming I do think the way you have factored this code is nicer than what we have right now, and if I were writing this code from scratch this is how I would probably do it.

However, because this isn't functionally different from the existing code I'd rather not merge it because it's not worth the risk of introducing a bug.

I appreciate the contribution but for now I'm going to close this one.

PR #8410, opened 2024-01-15
Truncate log values to 1024 bytes.

fasaxc:
## Description

Should prevent 1MB log lines when we accidentally log a protobuf object, for example.

* Special-case protobuf objects so that we can get the string marshaller to return early.  Avoid using the existing custom marshaller for protobuf types it doesn't support.
* Add truncating writer wrapper around byte buffer and rearrange formatting logic to Fprintf through that.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
The logging infrastructure now truncates very long values at 1024 bytes to avoid producing very long log lines.
```

## Reminder for the reviewer

Other optional labels:


fasaxc:
Hmm, thought of one or two logs that are intentionally long, for example, logging out iptables state before a panic...  Will need to tackle those.

fasaxc:
Closing this, think it's going to run into problems once we move to the main protoc compiler

PR #8411, opened 2024-01-15
add NAME COLUMN to the output of "calicoctl get wep" command

subaroon:
## Description

`calicoctl get wep` command should show NAME column.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
"calicoctl get wep" command now shows NAME column
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/calicoctl/calicoctl/resourcemgr/workloadendpoint.go
+++ b/calicoctl/calicoctl/resourcemgr/workloadendpoint.go
@@ -28,7 +28,7 @@ func init() {
 		api.NewWorkloadEndpointList(),
 		true,
 		[]string{"workloadendpoint", "workloadendpoints", "wep", "weps"},
-		[]string{"WORKLOAD", "NODE", "NETWORKS", "INTERFACE"},
+		[]string{"NAME", "WORKLOAD", "NODE", "NETWORKS", "INTERFACE"},
 		[]string{"NAME", "WORKLOAD", "NODE", "NETWORKS", "INTERFACE", "PROFILES", "NATS"},
 		// NAMESPACE may be prepended in GrabTableTemplate so needs to remain in the map below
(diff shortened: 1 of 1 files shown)


caseydavenport:
@subaroon thanks for the PR! However, the NAME heading is already available if you do `calicoctl get wep -o wide` - is there a reason you need it to be in the default output?

subaroon:
@caseydavenport 
thank you for your reply.

> @subaroon thanks for the PR! However, the NAME heading is already available if you do `calicoctl get wep -o wide` - is there a reason you need it to be in the default output?

- Resource types other than workloadEndpoint show NAME column without `-o wide`.
- NAME is the most important identifier for specifying a specific wep resource, as in `calicoctl get wep NAME`. Therefore, I think it's better to show NAME without `-o wide`.

caseydavenport:
Yep, that's a fair point. However, my concern is that changing this will break anyone using `calicoctl get wep` as part of their own scripts or tooling upon upgrade, as the column ordering will have changed.

subaroon:
> Yep, that's a fair point. However, my concern is that changing this will break anyone using `calicoctl get wep` as part of their own scripts or tooling upon upgrade, as the column ordering will have changed.

In general, scripts tend to prefer structured formats such as JSON or YAML over table formats. Therefore, I believe that prioritizing convenience and consistency should outweigh the potential impact on scripts.

caseydavenport:
That may be true, but I do know of users who are using this output in scripts regardless.

I agree with you that the NAME column should have been shown from the start, but given that it wasn't I don't know that the need is high enough to warrant a potentially breaking change at the moment. I appreciate you taking the time to open this PR, but am going to close it for now.

PR #8415, opened 2024-01-18
[BPF] test cleanup after switching back to iptables mode

tomastigera:
Also some fixes to remove all the maps.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: fixed cleaning of programs and map when switching from ebpf to iptables mode.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/cleanup.go
+++ b/felix/bpf/cleanup.go
@@ -28,7 +28,8 @@ func CleanUpCalicoPins(dir string) {
 		if err != nil {
 			return err
 		}
-		if strings.HasPrefix(info.Name(), "cali_") || strings.HasPrefix(info.Name(), "calico_") {
+		if strings.HasPrefix(info.Name(), "cali_") || strings.HasPrefix(info.Name(), "calico_") ||
+			strings.HasPrefix(info.Name(), "xdp_cali_") {
 			log.WithField("path", path).Debug("Deleting pinned BPF resource")
 			err = os.Remove(path)
 			if err != nil {
--- a/felix/bpf/cleanup.go
+++ b/felix/bpf/cleanup.go
@@ -28,6 +28,9 @@ func CleanUpCalicoPins(dir string) {
 		if err != nil {
 			return err
 		}
+		if path == "/sys/fs/bpf/calico/sockmap" {
+			return filepath.SkipDir
+		}
 		if strings.HasPrefix(info.Name(), "cali_") || strings.HasPrefix(info.Name(), "calico_") ||
 			strings.HasPrefix(info.Name(), "xdp_cali_") {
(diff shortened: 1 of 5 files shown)


felix/bpf/cleanup.go
```diff
@@ -28,7 +28,8 @@ func CleanUpCalicoPins(dir string) {
 		if err != nil {
 			return err
 		}
-		if strings.HasPrefix(info.Name(), "cali_") || strings.HasPrefix(info.Name(), "calico_") {
+		if strings.HasPrefix(info.Name(), "cali_") || strings.HasPrefix(info.Name(), "calico_") ||
```
fasaxc, line 31: Worth a utility function to share the match between these two files?

felix/fv/bpf_test.go
```diff
@@ -941,6 +941,33 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
 				})
+
+				It("should cleanup after we disable eBPF", func() {
+					By("Waiting for dp to get setup up")
+
+					ensureAllNodesBPFProgramsAttached(tc.Felixes)
+
+					By("Changing env and restarting felix")
+
+					tc.Felixes[0].SetEvn(map[string]string{"FELIX_BPFENABLED": "false"})
```
fasaxc, line 952:
```suggestion
					tc.Felixes[0].SetEnv(map[string]string{"FELIX_BPFENABLED": "false"})
```
typo in the utility func?

fasaxc: Didn't know we had that feature; FWIW, I added the `TestManagesBPF` flag to topology options for the "enable BPF" tests, you could use that (but this is fine too).

fasaxc (changes requested):

felix/fv/bpf_test.go
tomastigera: I am aware of that one, but I am happy to start in ebpf, but then i need to flip it to iptables.

felix/bpf/cleanup.go
tomastigera: Ack, but the parsing is done differently in the test. and atm it is imo just fine.

tomastigera:
@fasaxc the fv part of the tests passed

fasaxc:
The test that's failing is an old BPF sidecar acceleration in iptables mode test, we must be cleaning up its map?

tomastigera:
> The test that's failing is an old BPF sidecar acceleration in iptables mode test, we must be cleaning up its map?

yeah, that could be, there were some other collisions like that, let me check that.

fasaxc (approved):

PR #8417, opened 2024-01-18
Dual stack bpf - BPF endpoint manager refactor

sridhartigera:
## Description
This PR contains changes to BPF endpoint manager to make it viable to support dual stack in the near future. The changes are listed below
1. **Refactor BPF program loading** - We now load BPF wep, hep programs, attach preamble program, load policy. Have changed this to the following order. loadPrograms->loadPolicy->attach Preamble. This will help me load v4 program, v4 policy, v6 program, v6 policy and then load the preamble for both ingress, egress hooks.
2. **configure attach point** - Since we need attach point to load the preamble, have calculated the attachPoint which is common to both the dataplanes, passed them to v4 or v6 program loading path. 
3. **Add dataplane object to bpf ep mgr** - Introduced dataplane object which is part of bpf endpoint manager and moved the program loading, policy loading functions to the dataplane object.
4. Added v4/v6 policy, filter indices to the interface state.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -343,6 +341,45 @@ type bpfEndpointManager struct {
 	natOutIdx int
 
 	arpMap maps.Map
+
+	v4 *bpfEndpointManagerDataplane
+	v6 *bpfEndpointManagerDataplane
+}
+
+type bpfEndpointManagerDataplane struct {
```
tomastigera, line 349: Perhaps better to think of this like `bpfEpIPFamilyMgr` which manages what is specific for the ip family and it uses the `bpfDataplane` to achieve it?

tomastigera (commented):
I would expect that the manager does the top level management and then calls into the per-ip-family manager somehow. I am not sure we need to break the lowest level bpfDataplane interface, but it may need some changes.

Since the manager is atm for a single version, i'd go with whatever it does and extract out anything common rather than adding bunch of `if ipv6Enabled {` That i'd reserve for handling updates and the rest i'd handle with `if m.v6 exist also do the same for v6`

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1795,38 +1935,61 @@ func (m *bpfEndpointManager) doApplyPolicy(ifaceName string) (bpfInterfaceState,
...
+
+	//Attach preamble TC program
+	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		readiness := state.readiness
-		if readiness == ifaceIsReady {
-			if _, err := m.dp.queryClassifier(ifindex, state.qdisc.handle, state.qdisc.prio, false); err != nil {
-				readiness = ifaceNotReady
+		if readiness != ifaceIsReady {
```
tomastigera, line 1962: you can have this condition outside, for both in/egress

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1856,6 +2033,26 @@ func (m *bpfEndpointManager) applyPolicy(ifaceName string) error {
 	return err
 }
 
+func mergeAttachPoints(ap, ap6 attachPoint) attachPoint {
```
tomastigera, line 2036:
```suggestion
func mergeAttachPoints(ap4, ap6 attachPoint) attachPoint {
```

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1856,6 +2033,26 @@ func (m *bpfEndpointManager) applyPolicy(ifaceName string) error {
...
+	if aptcV4, v4ok := ap.(*tc.AttachPoint); v4ok {
+		if aptcV6, v6ok := ap6.(*tc.AttachPoint); v6ok {
+			if aptcV4 != nil && aptcV6 == nil {
+				return aptcV4
+			} else if aptcV6 != nil && aptcV4 == nil {
+				return aptcV6
+			}
```
tomastigera, line 2043: what if none of them is nil?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1531,29 +1548,82 @@ func (m *bpfEndpointManager) applyProgramsToDirtyDataInterfaces() {
...
+
+			if m.v6 != nil {
+				parallelWG.Add(1)
+				go func() {
+					defer parallelWG.Done()
+					ingressAP6, egressAP6, xdpAP6, err6 = m.v6.applyPolicyToDataIface(iface, hepPtr, &state,
+						tcAttachPoint, xdpAttachPoint)
+				}()
+			} else if m.v4 != nil {
```
tomastigera, line 1573:
```suggestion
			}
			if m.v4 != nil {
```

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1531,29 +1548,82 @@ func (m *bpfEndpointManager) applyProgramsToDirtyDataInterfaces() {
...
+				ingressAP4, egressAP4, xdpAP4, err4 = m.v4.applyPolicyToDataIface(iface, hepPtr, &state,
+					tcAttachPoint, xdpAttachPoint)
+			}
+
+			parallelWG.Wait()
```
tomastigera, line 1578: do you need to wait here? If you do not spin up v4 in goroutine,  it syncs and you can run the whole thin in parallel, no

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -2801,7 +3085,11 @@ func (m *bpfEndpointManager) ensureNoProgram(ap attachPoint) error {
 	}
 
 	// Forget the policy debug info
-	m.removePolicyDebugInfo(ap.IfaceName(), 4, ap.HookName())
+	if m.ipv6Enabled {
+		m.removePolicyDebugInfo(ap.IfaceName(), 6, ap.HookName())
+	} else {
```
tomastigera, line 3090: again, it is not "either or" can be both

tomastigera (changes requested):
I think it is getting where we want to get it, I'd suggest not to only fix the tests so that it compiles, but actually add dual stack tests.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -802,49 +882,19 @@ func (m *bpfEndpointManager) updateIfaceStateMap(name string, iface *bpfInterfac
 			flags |= ifstate.FlgReady
```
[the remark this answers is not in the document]
sridhartigera: I will do the readiness part as part of the subsequent PRs.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1691,39 +1761,106 @@ func (m *bpfEndpointManager) updateWEPsInDataplane() {
...
+	return nil
+}
 
+func (m *bpfEndpointManager) wepStateFillJumps(ifaceName string, state *bpfInterfaceState) error {
+	if m.ipv6Enabled {
+		err := m.allocJumpIndicesForWEP(ifaceName, &state.v6)
+		if err != nil {
+			return err
+		}
+	} else {
```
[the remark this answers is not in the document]
sridhartigera: This PR does not turn on dual stack yet. Will change all this when enabling dual stack in a subsequent PR

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1691,39 +1761,106 @@ func (m *bpfEndpointManager) updateWEPsInDataplane() {
...
+	return nil
+}
+
+func (m *bpfEndpointManager) dataIfaceStateFillJumps(ifaceName string, state *bpfInterfaceState) error {
+	if m.ipv6Enabled {
+		err := m.allocJumpIndicesForDataIface(ifaceName, &state.v6)
+		if err != nil {
+			return err
+		}
+	} else {
```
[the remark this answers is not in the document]
sridhartigera: Same as above.

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: Will handle this case, when turning on dual stack.

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: Same as above.

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: Same as above.

tomastigera (approved):
LGTM lets bank this intermediate step.

PR #8418, opened 2024-01-18
Unify all the RouteTables that point at the main kernel table

fasaxc:
## Description

Replace the plethora of RouteTable instances that we used to program routes into the main kernel routing table with a single unified RouteTable.

- Instead of handling conflicts by trying to program both routes into the kernel, detect conflicts in user space and program only the winning route in the kernel.  The previous behaviour resulted in a messy retry loop, spewing errors until the conflict was resolved. 
- Given each source of routes its own `RouteClass` with an associated priority to allow clean conflict resolution.  for example, a local workload route has a higher priority than an IPAM block "blackhole" route.  Bridge the gap from old API to new by introducing a ClassView object that exposes the old API to the "manager" layer objects.  They each act as if they have their own RouteTable but the ClassView adds the RouteClass in order to disambiguate routes.
- Align route programming with the indexing approach in the kernel.  In the kernel, routes are indexed on table, CIDR, ToS and metric.  Previously, we would read back routes and only take account of the CIDR.   This resulted in the possibility of aliasing: we would read back two routes and only see one CIDR. 
- Identifying which routes belong to Calico in the main routing table is very messy and requires heuristics that were spread across several components.  Unify those heuristics into a single "ownership policy" abstraction.  The main routing table gets its own ownership policy.
- Fix that conntrack cleanup wasn't properly handled when routes moved between RouteTables.  conntrack cleanup would only be sequenced correctly when routes moved within a RouteTable.  For example, if a route moved from one workload to another, we'd correctly sequence removal of the route, conntrack cleanup and adding the new route.  However, if a route moved from local workload to remote VXLAN via the tunnel, we'd do the cleanup in parallel with programming the new route resulting in potential glitches.
- Move conntrack cleanup to a new object  to encapsulate the fiddly calculation.

## Related issues/PRs

CORE-9902
CORE-10126

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Felix's route programming now handles routing conflicts deterministically, prioritising routes based on their type.  Conntrack cleanup has been improved; cleanup is now correctly sequenced with route programming when IP addresses move from local to remote workloads. 
```

## Reminder for the reviewer

Other optional labels:


fasaxc:
Still filling in the last bits of coverage but I think this is ready for a first pass now...

caseydavenport (approved):
Was easier to review than I feared, looks good to me!

marvin-tigera:

marvin-tigera:

PR #8420, opened 2024-01-19
Add pre-flight check to node STs to verify cgroup v1 is available.

fasaxc:
## Description

The ancient version of dind we use requires cgroupv1 so there's no point in running if it's not mounted.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/Makefile
+++ b/node/Makefile
@@ -365,12 +365,18 @@ load-container-images: $(K8ST_IMAGE_TARS) $(KUBECTL)
 
 .PHONY: st-checks
 st-checks:
-	# Check that we're running as root.
-	test `id -u` -eq '0' || { echo "STs must be run as root to allow writes to /proc"; false; }
-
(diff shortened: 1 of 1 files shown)


node/Makefile
```diff
@@ -365,12 +365,18 @@ load-container-images: $(K8ST_IMAGE_TARS) $(KUBECTL)
 
 .PHONY: st-checks
 st-checks:
```
fasaxc, line 367: This target wasn't being used and I think it was out of date so I repurposed it

tomastigera: it was used on line 412 below, right? Shouldn't we keep those checks and preps? Idk

fasaxc: It's been commented out for 7 years; think it's from before we containerised the build/test process!

tomastigera (approved):

PR #8424, opened 2024-01-22
workloadEndpoint.spec.mac is empty

subaroon:
## Description

The current code attempts to assign the current MAC address to the `workloadEndpoint.spec.mac` field, but this operation is ineffective.

https://github.com/projectcalico/calico/blob/086c2754a422f0631622493a9ecf9a63aeb0c9c8/cni-plugin/pkg/k8s/k8s.go#L404-L410

**change**
This change ensures that the current MAC address is assigned to both the `workloadEndpoint.spec.mac` field and the `cni.projectcalico.org/hwAddr` annotation.  If the user hasn't specified the `cni.projectcalico.org/hwAddr` annotation,  calico CNI will still set the current MAC address in this annotation.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/k8s/conversion/constants.go
+++ b/libcalico-go/lib/backend/k8s/conversion/constants.go
@@ -41,6 +41,12 @@ const (
 	// on older Pods.
 	AnnotationContainerID = "cni.projectcalico.org/containerID"
 
+	// AnnotationHwAddr is used to store the desired MAC address that will be assigned to a network interface
+	// on a pod. If a desired MAC address is not specified by the user, the Calico CNI plugin will still set
+	// the current MAC address in this annotation. This allows for both the specification of a custom MAC
+	// address and the automatic handling of MAC addresses by the plugin.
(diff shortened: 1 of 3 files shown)


caseydavenport:
@subaroon thanks for the PR - is there and issue that this fixes? Could you share a little bit more about your use-case and desired behavior?

subaroon:
**use-case**

Some user application relies on MAC address and users want to check the MAC address the user's pod uses for debug and observability.

**desired behavior**

`workloadEndpoint.spec.mac` annotation stores current MAC address assigned to pod.

caseydavenport:
> for debug and observability.

The MAC address selected by Calico never shows up on the wire - it's only used on the point-to-point veth interface between the container and its own host. So I think knowing this value has very limited uses.

> Some user application relies on MAC address

In what way? We already allow configuration of the MAC address for applications that, for example, tie licensing to a specific address.

PR #8451, opened 2024-01-26
Create V4 and V6 maps for BPF dual stack

sridhartigera:
## Description

This PR creates V4 and V6 maps when ipv6 mode is enabled. Also increased the size of program array map to 400 to accommodate both v4 and v6 programs.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


tomastigera (commented):
lgtm in general, but I wonder if it wasn't easier to handle two sets of maps with the same names rather than introducing v6 names. Wouldn't it lead to a bunch of code duplicity and snow-ball later on?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -451,10 +453,12 @@ func newBPFEndpointManager(
 		dsrOptoutCidrs:          len(config.BPFDSROptoutCIDRs) > 0,
 		bpfExtToServiceConnmark: config.BPFExtToServiceConnmark,
 		psnatPorts:              config.BPFPSNATPorts,
-		bpfmaps:                 bpfmaps,
+		v4BPFMaps:               v4Maps,
+		v6BPFMaps:               v6Maps,
+		commonBPFMaps:           commonMaps,
```
tomastigera, line 458: nit - feels like we do not have any other than bpf maps

```suggestion
		v4Maps:               v4Maps,
		v6Maps:               v6Maps,
		maps:                   commonMaps,
```

And I wonder if the v4/6 maps can't be just part of the `bpfEndpointManagerDataplane` and them you would access them like `m.v6.maps.ArpMap...`

In fact, if you embed the maps in the dataplane like 
```
type bpfEndpointManagerDataplane struct {
        *bpfmap.Maps                                                                <---
        
	ipFamily proto.IPVersion
	hostIP   net.IP
	mgr      *bpfEndpointManager

	ifaceToIpMap map[string]net.IP

	// IP of the tunnel / overlay device
	tunnelIP net.IP
}
```

You could just access them as `m.v6.ArpMap`, can't you?

felix/bpf/bpfmap/bpf_maps.go
```diff
@@ -34,18 +34,25 @@ import (
...
+	ArpMap       maps.Map
+	FailsafesMap maps.Map
+	FrontendMap  maps.Map
+	BackendMap   maps.Map
+	AffinityMap  maps.Map
+	RouteMap     maps.Map
+	CtMap        maps.Map
+	SrMsgMap     maps.Map
+	CtNatsMap    maps.Map
+
```
tomastigera, line 54: Since we have so many nil pointers in this struct, why not to split it further

```suggestion
type IPMaps {
	IpsetsMap    maps.Map
	ArpMap       maps.Map
	FailsafesMap maps.Map
	FrontendMap  maps.Map
	BackendMap   maps.Map
	AffinityMap  maps.Map
	RouteMap     maps.Map
	CtMap        maps.Map
	SrMsgMap     maps.Map
	CtNatsMap    maps.Map
}

type Generic/CommonMaps {
     <all the other maps>
}

type Maps interface {
    AsSlice() []maps.Map
}
```

As a benefit, then the functions like `getCommonBPFMaps(mapsPtr *Maps) []maps.Map` would not need to return the slice, `AsSlice90` would give it to you.

In any case `CreateBPFMaps(ipv6Enabled bool)` could return `CreateBPFMaps(ipv6Enabled bool) (*IPMaps, *IPMaps, *CommonMaps, error) or ([3]Maps, error).

The whole point of introducing the Maps struct was to be able to pass all the sorts of maps in some condensed way. One struct instead of 15. It might also make sense instead of defining `Maps interface`  to define 

```
type Maps struct {
    Common *CommonMaps
    V4             *IPMaps
    V6             *IPMaps
}
```

That would allow you not to change signature of func like `newBPFEndpointManager()` where you could still pass `*bpfmap.Maps`

Also you would not need to define as you would not have an array :)

```
const (
	V4MapIndex = iota
	V6MapIndex
	CommonMapIndex
)
```

felix/dataplane/linux/int_dataplane.go
```diff
@@ -747,12 +758,14 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 
 		kfb := conntrack.KeyFromBytes
 		vfb := conntrack.ValueFromBytes
+		ctMap := bpfMaps[0].CtMap
 		if config.BPFIpv6Enabled {
 			kfb = conntrack.KeyV6FromBytes
 			vfb = conntrack.ValueV6FromBytes
+			ctMap = bpfMaps[1].CtMap
```
tomastigera, line 765: i think this is also what you should do for the proxy maps and then instantiate once. You can then group all of the stuff that needs to be instantiated twice for v4 and v6. You will need 2 ct scanners too. Btw it might be good to run the ct scanners specifically in parallel.

tomastigera: And you should definitely use the name for the index instead of 0 or 1

tomastigera: Also the `bpfproxyOpts` is updated somewhere for v6 already

tomastigera (changes requested):

Issue #7987, opened 2023-09-05
calico does not release free blocks assigned to node


mchtech:
according to https://github.com/projectcalico/calico/blob/v3.26.1/kube-controllers/pkg/controllers/node/ipam.go#L648-L653 , 
calico will not release free blocks when `numAddressesAvailableOnNode < 3`

## Expected Behavior

user/calico can release free blocks even if `numAddressesAvailableOnNode < 3`

## Current Behavior

calico will not release free blocks when `numAddressesAvailableOnNode < 3`

## Possible Solution

1. make `numAddressesAvailableOnNode < 3` configurable.
2. provide steps to release free blocks manually.

## Context

We run 1~2 apps on a node, so we set ip pool `blockSize = 31` to mininize bgp route count that exported to idc. If calico do not release unused block, some nodes will be assigned two /31 blocks "forever". It siginificantly reduced the utilization of ippool.

## Your Environment

* Calico version: 3.26.1 with etcd datastore
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes v1.26.7
* Operating System and version: oracle linux 8.8
* Link to your project (optional):

mazdakn:
@caseydavenport is it a good idea to make this number configurable?

mchtech:
> > If calico do not release unused block, some nodes will be assigned two /31 blocks "forever". It siginificantly reduced the utilization of ippool.
> 
> Only if you have IPAM_STRICT_AFFINITY set. Otherwise the ip-borrowing behaviour will kick in when addresses are almost exhausted and borrow addresses from blocks allocated to other nodes.
> 
> Is this _actually_ causing you a problem? Can you describe that problem please? I think that even with blockSize = 31, Calico should be able to utilise ALL the IPs in the pool (because of "ip borrowing").

## problem: 
`calicoctl ipam show` output shows that there are lots of free addresses, but cni reports:
`
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "d41dc562be6437177983bf4dbb5c73bbbe4f86b1bc61c56dcff97a7b65bf8fc3": plugin type="calico" failed (add): failed to request IPv4 addresses: Assigned 0 out of 1 requested IPv4 addresses; No more free affine blocks and strict affinity enabled
`

I must add more ippools to solve this problem, but the utilization of ip is low. (max 50%, if calico does not reclaim /31 free blocks)

[...]

PR #8454, opened 2024-01-29
Release IPAM blocks if they have been empty for a certain time

caseydavenport:
## Description

Remove the check for total free addresses on a node, and instead just
rely on the time-based heuristic for whether an IPAM block is OK to
delete.

By default, blocks will now be deleted if they are consistently empty
for 15m or more (the default leak grace period). But this can be
configured using KubeControllersConfiguration. 

The previous check was meant to ensure we didn't allocate and delete
blocks rapidly, but the time-based grace period should do that just as
well.

## Related issues/PRs

Fixes https://github.com/projectcalico/calico/issues/7987

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Improve IPAM block garbage collection behavior for IP pools with small blocks.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -472,11 +472,18 @@ func (c *ipamController) onBlockUpdated(kvp model.KVPair) {
 		log.WithFields(alloc.fields()).Debug("New IP allocation")
 	}
 
-	// If the block is empty, mark it as such. We'll check if it needs to be
-	// cleaned up in the sync loop.
+	// Check if the block is empty and schedule it for GC if needed.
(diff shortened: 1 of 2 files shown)


fasaxc (changes requested):
+1 for simpler heuristic.  One thing I'm not quite sure about is how a block gets marked as in-use again if we later spot that it starts being used again.  I went looking but the only markInUse calls are in this code, which is only called for empty blocks?

caseydavenport:
@fasaxc good catch - just pushed a commit that adds a `markInUse` call to the block update logic if it spots that there are allocations.

fasaxc:
Would be good to have a test for the case that you fixed but don't want to block this going in today

PR #8469, opened 2024-02-01
Add optional pod startup delay to CNI - wait for Felix to program dataplane

aaaaaaaalex:
## Description
Currently there is a small interval between when the CNI programs basic pod networking, and when Felix programs the dataplane for the corresponding workload. This is normally unnoticeable, but in high-load clusters with slow dataplane programming, it can lead to pods making requests that fail, or pods being added to services before their networking is fully functional.

This PR adds an option enabling the CNI to delay a pod going ready until Felix has programmed the dataplane for it.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Adds options to Felix and the CNI to delay pods going ready until their dataplane programming is complete.
```

## Reminder for the reviewer

Other optional labels:


felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,202 @@
...
+
+// SyncForever blocks until ctx is cancelled.
+// Continuously pulls from endpoint-updates C,
+// and writes files to the status directory.
+func (fr *EndpointStatusFileReporter) SyncForever(ctx context.Context) {
+	recreatePolicyDir := true
+	for {
+		if recreatePolicyDir {
+			logrus.Debug("Creating/truncating policy status directory")
+			err := ensurePolicyStatusDir(fr.endpointStatusDirPrefix, true)
```
fasaxc, line 116: I'd prefer if you synced the directory contents instead of deleting everything and starting over.  I'd use a DeltaTracker[string,bool], where the string is filename and the bool is true if the file should exist (on the Desired() side)/does exist(on the Dataplane() side).

Tracking it like that will give you a way to handle corner cases such as disk full since the DeltaTracker will keep track of the desired state so you can retry creating the file later.  It's also trivial to load the directory contents back up in to the Dataplane side with ReplaceAllIter.

libcalico-go/lib/dispatcher/blocking.go
```diff
@@ -0,0 +1,100 @@
...
+			for _, o := range outputs {
+				selectCases = append(selectCases, reflect.SelectCase{
+					Dir:  reflect.SelectSend,
+					Chan: reflect.ValueOf(o),
+					Send: reflect.ValueOf(msg),
+				})
+			}
+
+			for range outputs {
+				i, _, _ := reflect.Select(selectCases)
```
fasaxc, line 89: This doesn't look right to me, you need to remove the "completed" cases from the select after each loop.  Otherwise you're doing this:
```
msg := <- input
for range outputs {
  select {
  case <-done:
  case out1 <- msg:
  case out2 <- msg:
  }
}
```
so, you could send the same message to the same output multiple times.

libcalico-go/lib/dispatcher/blocking.go
```diff
@@ -0,0 +1,100 @@
...
+	"reflect"
+
+	"github.com/sirupsen/logrus"
+)
+
+// BlockingDispatcher blocks on an input channel until stop,
+// and broadcasts each input message to all output channels.
+// Blocks until all output channels have consumed a message,
+// before pulling in the next input message.
+type BlockingDispatcher[T any] struct {
```
fasaxc, line 29: Couple of thoughts: 
- I don't see any problem with using reflect.Select to implement this kind of logic.  "One clever trick" behind a nice, minimal API is often a good way to go.
- Note that `reflect.Select` has an inefficient slice copy inside it to protect the runtime (nbd here but can be an issue in a tight loop).
- Overall, you might be better off giving all the output channels a buffer (say length 10) and then you could do:

```
msg <- input
for _, outC := range outputs {
  switch {
  case <-ctx.Done():
    return
  case outC <- msg:
  }
}
```

That has the advantage(?) that one consumer can be 10 ahead of another before they start to get paced by the slowest consumer.

felix/config/config_params.go
```diff
@@ -331,6 +331,14 @@ type Config struct {
 	EndpointReportingEnabled   bool          `config:"bool;false"`
 	EndpointReportingDelaySecs time.Duration `config:"seconds;1"`
 
+	// EndpointStatusPathPrefix is the path to the directory
+	// where endpoint status will be written. Endpoint status
+	// file reporting is disabled if field is left empty.
+	//
+	// Chosen directory should match the directory used by the CNI for PodStartupDelay.
+	// [Default: empty]
+	EndpointStatusPathPrefix *string `config:"file;;"`
```
fasaxc, line 340:
```suggestion
	EndpointStatusPathPrefix string `config:"file;;"`
```
Normally just use string instead of `*string`. Only need a `*string` if nil and "" have distinct meanings, which we should avoid unless its forced.

felix/daemon/daemon.go
```diff
@@ -642,16 +643,32 @@ configRetry:
...
-		dpConnector.statusReporter.Start()
+		statusReporter.Start()
+	}
+
+	if configParams.EndpointStatusPathPrefix != nil && *configParams.EndpointStatusPathPrefix != "" {
+		fromDataplaneC := make(chan interface{})
+		dpConnector.StatusUpdatesFromDataplaneConsumers = append(dpConnector.StatusUpdatesFromDataplaneConsumers, fromDataplaneC)
+		statusFileReporter := statusrep.NewEndpointStatusFileReporter(fromDataplaneC, *configParams.EndpointStatusPathPrefix)
+
+		// TODO: investigate if it's a requirement that
```
fasaxc, line 666: These components should never stop.  Felix will restart if the config changed.  We only use contexts here for easier testing, I think.

felix/statusrep/status_file_reporter.go
fasaxc: Of course, you'll need an "in sync" message to know when to delete old files.

fasaxc (changes requested):

libcalico-go/lib/dispatcher/blocking.go
aaaaaaaalex: Yikes - good catch

felix/statusrep/status_file_reporter.go
aaaaaaaalex: Would you be able to check what I've done with a SetDeltaTracker here? I may be misunderstanding the reason you recommended the base DeltaTracker, as opposed to SetDeltaTracker, or exactly what SetDeltaTracker is doing under the covers, in-which-case I'll adapt this to use the base DeltaTracker.

felix/config/config_params.go
aaaaaaaalex: Done

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,202 @@
...
+			return
+		case e, ok := <-fr.endpointUpdatesC:
+			logrus.WithField("endpoint", e).Debug("Handling endpoint update")
+			if !ok {
+				logrus.Panic("Input channel closed unexpectedly")
+			}
+			switch m := e.(type) {
+			case *proto.WorkloadEndpointStatusUpdate:
+				// Write file to dir.
+				filename := filepath.Join(fr.endpointStatusDirPrefix, m.Id.WorkloadId)
```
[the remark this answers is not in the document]
aaaaaaaalex: Added some libcalico-go functions for converting back-and-forth. Also made a function for converting the ID type used by the CNI "WorkloadIdentiiers". Need to check if those ultimately result in the same string being generated.

felix/daemon/daemon.go
```diff
@@ -642,16 +643,33 @@ configRetry:
 		delay := configParams.EndpointReportingDelaySecs
 		log.WithField("delay", delay).Info(
 			"Endpoint status reporting enabled, starting status reporter")
-		dpConnector.statusReporter = statusrep.NewEndpointStatusReporter(
+
+		fromDataplaneC := make(chan interface{}, 10)
+		inSyncC := make(chan bool, 1)
```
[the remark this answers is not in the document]
fasaxc: Hmm, I think the old code is wrong on this front but it papers over that by having a 180s delay before it does its first resync, so by that point it's very likely to have converged!  The new code doesn't have such a buffer so maybe we need to tackle this properly.

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,305 @@
...
+		select {
+		case <-ctx.Done():
+			logrus.Warn("Context cancelled, stopping...")
+			return
+		case b, ok := <-fr.inSyncC:
+			if !ok {
+				logrus.Panic("InSync channel closed unexpectedly.")
+			}
+
+			if b == true {
```
fasaxc, line 139: Quite a lot of duplication between the cases?   

It can often work well to have the select cases just update flags/caches and then fall through to common processing that actually does the sync or resync or whatever
```
select ...
case b, ok := <-inSync
  ...
   inSync = b
case msg := <- ...
   fr.handleMessage(msg)
case <-resyncC:
  resyncNeeded = true
}

if inSync  && resyncNeeded {
  ...
}
...
if <error occurred> {
  // update backoff and retryC
} else {
  retryC = nil
}
```

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,305 @@
...
+				logrus.Warn("InSync received from calc graph.")
+				inSyncWithUpstream = true
+				err := fr.syncForeverReconcilePolicyFiles(true)
+				if err != nil {
+					retryC = time.After(fr.bom.Step())
```
fasaxc, line 144: It's a little bit leaky to use `time.After`.  When you nil out or overwrite the retryC, the old timer is leaked until it pops.  They can build up if you do that in a tight loop.

I prefer to use a timer that is stopped/reset as needed but you need to make sure to use the correct Stop sequence from the docs to avoid races.

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,305 @@
...
+// Can only return an error after a failed commit, so a returned error should
+// always result in SyncForever queueing a retry.
+func (fr *EndpointStatusFileReporter) syncForeverHandleEndpointUpdate(e interface{}, commitToKernel bool) error {
+	switch m := e.(type) {
+	case *proto.WorkloadEndpointStatusUpdate:
+		fr.policyDirDeltaTracker.Desired().Add(names.WorkloadEndpointIDToStatusFilename(m.Id))
+	case *proto.WorkloadEndpointStatusRemove:
+		fr.policyDirDeltaTracker.Desired().Delete(names.WorkloadEndpointIDToStatusFilename(m.Id))
+	default:
+		logrus.WithField("update", e).Warn("Skipping unrecognized endpoint update")
```
fasaxc, line 208: You'll get HEP updates, which you shouldn't warn about.

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,305 @@
...
+
+		return deltatracker.IterActionUpdateDataplane
+	})
+
+	fr.policyDirDeltaTracker.PendingDeletions().Iter(func(name string) deltatracker.IterAction {
+		err := fr.deletePolicyFile(name)
+		if err != nil {
+			logrus.WithError(err).Warn("Failed to delete file in policy-status-dir")
+			// Carry on as normal (with a warning) if the file is somehow already deleted.
+			if !errors.Is(err, fs.ErrNotExist) {
```
fasaxc, line 255: I think there's a standard helper `os.IsNotExist(err)`  not sure if this is equivalent

libcalico-go/lib/dispatcher/blocking.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// NewBlockingDispatcher returns a BlockingDispatcher which
+// will consume messages from the input channel.
+// Dispatcher must be started with Start(ctx). Stopped by cancelling ctx.
+func NewBlockingDispatcher[T any](input <-chan T) (*BlockingDispatcher[T], error) {
+	log := logrus.WithField("component", "BlockingDispatcher")
+
+	if input == nil {
+		return nil, errors.New("No input channel provided for dispatcher to consume")
```
fasaxc, line 42: I'd usually panic rather than returning an error in this sort of case.  It's a programming bug to forget to pass a valid argument so the caller likely won't be able to handle it (since they caused the problem!)

Handling the extra error is just a bit tedious.

libcalico-go/lib/dispatcher/blocking.go
```diff
@@ -0,0 +1,85 @@
...
+	log.WithField("outputs", len(outputs)).Warn("Dispatcher running...")
+	for {
+		// Wait for input or ctx cancellation.
+		select {
+		case <-ctx.Done():
+			log.Debug("Stopping...")
+			return
+		case msg, ok := <-d.input:
+			if !ok {
+				log.Panic("Input channel closed unexpectedly")
```
fasaxc, line 71: We rarely close channels in felix but I think some Go users close channels to mean "I'm done sending the data" so the natural thing to do in that case might be to close the downstream channels.

I wouldn't change this, just a thought about what users may or may not expect!

libcalico-go/lib/names/workloadendpointstatus.go
```diff
@@ -0,0 +1,74 @@
...
+
+// WorkloadEndpointIDToStatusFilename accepts a workload endpoint ID
+// and converts it to a filename for use in WEP-policy status syncing
+// between Felix and the CNI.
+func WorkloadEndpointIDToStatusFilename(id *proto.WorkloadEndpointID) string {
+	parts := make([]string, len(expectedFields))
+	parts[fieldOrchestratorID] = id.OrchestratorId
+	parts[fieldWorkloadID] = strings.ReplaceAll(id.WorkloadId, "/", separator)
+	parts[fieldEndpointID] = id.EndpointId
+	logrus.WithField("parts", parts).Warn("gnerating filename from workload endpoint ID")
```
fasaxc, line 47: Typo: gnerating

libcalico-go/lib/names/workloadendpointstatus.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	fieldOrchestratorID = iota
+	fieldWorkloadID
+	fieldEndpointID
+)
+const (
+	separator = "-"
```
fasaxc, line 32: I think `-` is a valid char in the various parts of the ID so it's no good as a separator.  You could use `/` and URL-encode it.  We know `/` is OK because we use it in our etcd names.  Otherwise you could inspect the validation regexes to find a good separator that's not already in use.

fasaxc (changes requested):

felix/daemon/daemon.go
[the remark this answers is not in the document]
aaaaaaaalex: I noticed alright that the insync might be out-of-sync with the WEP updates, but Im not sure whether the impact of that is significant? I can see the case where a WEP doesn't come in, and so we incorrectly delete a file, but then subsequently re-add it once the late update comes in.

felix/statusrep/status_file_reporter.go
aaaaaaaalex: Yep it's equivalent I believe, I went with this one after checking the docstring of `IsNotExist`:

   >  This function predates errors.Is. It only supports errors returned by the os
    package. New code should use errors.Is(err, fs.ErrNotExist).

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,305 @@
...
+)
+
+const (
+	dirPolicyStatus = "policy"
+)
+
+// EndpointStatusFileReporter writes a file to the FS
+// any time it sees an Endpoint go up in the dataplane.
+//
+//   - Currently only writes to a directory "policy", creating
```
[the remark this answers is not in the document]
aaaaaaaalex: Sure, can just make it "status". My outlook was to name the directory as if it might have other forms of statuses, besides policy status in the future. If we're happy conflating policy-status with "status" then I'll consolidate that change across all the config params

cni-plugin/pkg/k8s/k8s.go
```diff
@@ -970,3 +983,95 @@ func getIPsByFamily(cidrs []string) (string, string, error) {
...
+			err = watcher.Add(directory)
+			if err != nil {
+				logrus.WithError(err).Warn("Couldn't add directory to watcher")
+				continue
+			}
+		}
+
+		// We should check once that we didn't miss the creation
+		// of the file, while setting up the watcher.
+		entries, err := os.ReadDir(directory)
```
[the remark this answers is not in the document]
aaaaaaaalex: Done :+1: Watcher error triggers a full retry including a Stat call.

cni-plugin/pkg/k8s/k8s.go
```diff
@@ -970,3 +983,95 @@ func getIPsByFamily(cidrs []string) (string, string, error) {
...
+}
+
+// notifyWhenFileExists closes notif when the specified file is seen on disk.
+// Returns without closing the channel if ctx is cancelled.
+func notifyWhenFileExists(ctx context.Context, notif chan struct{}, directory, filename string) {
+	var watcher *fsnotify.Watcher
+	var err error
+	// Retry loop for any errors are encountered.
+	for {
+		select {
```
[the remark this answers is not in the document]
aaaaaaaalex: While I didn't factor-out the body of the function, I've moved to a flag-based approach where each code path just sets the relevant flags and doesn't worry about resources - cleanup is centralised at the top of the iteration and is controlled with the flags.

cni-plugin/pkg/k8s/k8s.go
```diff
@@ -970,3 +983,95 @@ func getIPsByFamily(cidrs []string) (string, string, error) {
...
+
+		for {
+			select {
+			case e := <-watcher.Events:
+				logrus.WithField("name", e.Name).Info("Received watch event")
+				switch e.Op {
+				case fsnotify.Create:
+					if e.Name == filename {
+						logrus.WithField("file", e.Name).Debug("FS 'Create' event seen for file. Firing notification and stopping watch.")
+						close(notif)
```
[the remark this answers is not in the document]
aaaaaaaalex: Thanks, I've revised [how the loop manages resources / code paths](https://github.com/projectcalico/calico/pull/8469#discussion_r1494507494) now.

cni-plugin/pkg/k8s/k8s.go
```diff
@@ -970,3 +983,95 @@ func getIPsByFamily(cidrs []string) (string, string, error) {
...
+	}
+}
+
+// notifyWhenFileExists closes notif when the specified file is seen on disk.
+// Returns without closing the channel if ctx is cancelled.
+func notifyWhenFileExists(ctx context.Context, notif chan struct{}, directory, filename string) {
+	var watcher *fsnotify.Watcher
+	var err error
+	// Retry loop for any errors are encountered.
+	for {
```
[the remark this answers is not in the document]
aaaaaaaalex: Errors result in a watcher cleanup and stat() call now - retry interval is set at 500ms.

cni-plugin/pkg/types/types.go
```diff
@@ -125,6 +125,17 @@ type NetConf struct {
 	// The CNI plugin waits until all the endpoints specified in ReadinessGates are ready
 	ReadinessGates []string `json:"readiness_gates"`
 
+	// The maximum duration to delay pod startup while waiting
+	// for Felix to program policy for the endpoint.
+	//
+	// Default: 0
+	PodStartupDelaySeconds int `json:"pod_startup_delay_seconds,omitempty"`
```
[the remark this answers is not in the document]
aaaaaaaalex: Renamed.

cni-plugin/pkg/types/types.go
```diff
@@ -125,6 +125,17 @@ type NetConf struct {
...
+
+	// The directory to watch for policy updates for workloads.
+	//
+	// Default: /var/run/calico/policy
+	EndpointPolicyStatusDir string `json:"endpoint_status_dir,omitempty"`
```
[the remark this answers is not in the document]
aaaaaaaalex: Done

felix/daemon/daemon.go
```diff
@@ -1243,6 +1279,23 @@ func (fc *DataplaneConnector) Start() {
 
 	// Start a background thread to handle Wireguard update to Node.
 	go fc.handleWireguardStatUpdateFromDataplane()
+
+	log.WithFields(log.Fields{
+		"statusUpdatesFromDataplaneConsumers": len(fc.StatusUpdatesFromDataplaneConsumers),
+		"inSyncConsumers":                     len(fc.InSyncConsumers),
+	}).Warn("DataplaneConnector starting.")
```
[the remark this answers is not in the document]
aaaaaaaalex: Did this for log visibility during dev - will scan all changes for Warns before re-requesting review :ok_hand:

felix/statusrep/status_file_reporter.go
aaaaaaaalex: is this the case regardless of if I assign the timer to a variable? e.g. would it still happen when selecting directly with
```
case <-time.After:
```

Presumably yes, if another case short-circuits the select and causes a tight retry loop?

felix/statusrep/status_file_reporter.go
aaaaaaaalex: Updated. The loop is quite long now, so if you're happy with the pattern Im using now, I'll go ahead and try to factor bits out.

felix/statusrep/status_file_reporter.go
aaaaaaaalex: :point_up:  (Yes)

libcalico-go/lib/names/workloadendpointstatus.go
aaaaaaaalex: Now using spaces as separators which the FS is happy about. Im also percent-encoding/decoding slashes using PathEncode std function

felix/statusrep/status_file_reporter.go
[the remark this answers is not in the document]
fasaxc: Yep, that's my take, this is the done with endpoint high-level status.  Can add more statuses later if needed.  I think I'd probably add content to the file at that point.  Make it a JSON blob with more detail in it.

libcalico-go/lib/dispatcher/blocking_test.go
```diff
@@ -0,0 +1,89 @@
...
+				case o, ok = <-output2:
+				case <-time.After(1 * time.Second):
+					Fail("Timed out waiting for dispatcher to send to outputs.")
+				}
+				Expect(ok).To(BeTrue(), "A channel closed unexpectedly.")
+				dummyOutput, ok := o.(int)
+				Expect(ok).To(BeTrue(), "Output should be the same datatype as input.")
+				Expect(dummyOutput).To(Equal(dummyInput))
+			}
+		})
```
fasaxc, line 79: Gomega has some nice matchers for channels
```suggestion
			output1, output2 := make(chan interface{}, 1), make(chan interface{}, 1)
			dummyInput := 666

			runDispatcherInBackgroundWithOutputChans(output1, output2)

			// Dispatcher itself implicitly acts as a buffer of size 1.
			input <- dummyInput
			
			Eventually(output1).Should(Receive(Equals(dummyInput)))
			Eventually(output2).Should(Receive(Equals(dummyInput)))
		})
```

libcalico-go/lib/names/workloadendpointstatus.go
```diff
@@ -0,0 +1,111 @@
...
+// WorkloadEndpointToStatusFilename generates a string to be used as a status-file's name.
+// Operation should be reversible, i.e., the stringification shouldn't be lossy.
+// Returns "" if passed endpoint is nil.
+func WorkloadEndpointToStatusFilename(ep *libapi.WorkloadEndpoint) string {
+	if ep == nil {
+		return ""
+	}
+
+	parts := make([]string, len(expectedFields))
+	parts[fieldOrchestratorID] = escape(ep.Spec.Orchestrator)
```
fasaxc, line 105: As discussed, would be good to split conversion to a canonical ID type from the actual name construction.  Easier to get those two halves correct independently.

fasaxc (changes requested):

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,344 @@
...
+				return deltatracker.IterActionNoOp
+			}
+		}
+		return deltatracker.IterActionUpdateDataplane
+	})
+
+	fr.statusDirDeltaTracker.PendingDeletions().Iter(func(name string) deltatracker.IterAction {
+		err := fr.deleteStatusFile(name)
+		if err != nil {
+			// Carry on as normal (with a warning) if the file is somehow already deleted.
```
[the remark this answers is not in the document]
aaaaaaaalex: Hmm I might be missing something here - I should still continue if a deletion ended up already not-existing?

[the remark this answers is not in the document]
aaaaaaaalex: Ah: "with a warning" is not correct. :D fixing

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,344 @@
...
+
+// SyncForever blocks until ctx is cancelled.
+// Continuously pulls status-updates from updates C,
+// and reconciles the filesystem with internal state.
+func (fr *EndpointStatusFileReporter) SyncForever(ctx context.Context) {
+	// State flags influencing the loop.
+	// Each flag triggers one behaviour within the loop.
+	// Once the behaviour succeeds, the flag is switched off.
+	// If the behaviour fails, the flag is left on, and a retry is queued.
+	var retryTimerResetNeeded, resyncWithKernelNeeded, applyToKernelNeeded, exit bool
```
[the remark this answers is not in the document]
aaaaaaaalex: Good point. Simplified now.

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,344 @@
...
+
+// Overwrites our user-space representation of the kernel with a fresh snapshot.
+func (fr *EndpointStatusFileReporter) resyncDataplaneWithKernel() error {
+	// Load any pre-existing committed dataplane entries.
+	entries, err := ensureStatusDir(fr.endpointStatusDirPrefix)
+	if err != nil {
+		return err
+	}
+
+	fr.statusDirDeltaTracker.Dataplane().ReplaceFromIter(func(f func(k string)) error {
```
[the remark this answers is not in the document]
aaaaaaaalex: Returning the value of the replacement now - so any error will fall in naturally with the existing error handling.

felix/daemon/daemon.go
```diff
@@ -978,17 +991,27 @@ func newConnector(configParams *config.Config,
...
-		datastorev3:                      datastorev3,
-		ToDataplane:                      make(chan interface{}),
-		StatusUpdatesFromDataplane:       make(chan interface{}),
+		config:                              configParams,
+		configUpdChan:                       configUpdChan,
+		datastore:                           datastore,
+		datastorev3:                         datastorev3,
+		ToDataplane:                         make(chan interface{}),
+		StatusUpdatesFromDataplane:          make(chan interface{}),
+		StatusUpdatesFromDataplaneConsumers: make([]chan interface{}, 0),
```
fasaxc, line 1000: I'd leave the slice nil, making a 0-length slice is a (trivial) waste of time.

felix/statusrep/status_file_reporter.go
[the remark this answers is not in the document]
fasaxc: Still referencing policy here?

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,347 @@
...
+
+// A sub-call of SyncForever, not intended to be called outside the main loop.
+// Updates delta tracker state to match the received update.
+// Logs and discards errors generated from converting endpoint updates to endpoint keys.
+func (fr *EndpointStatusFileReporter) handleEndpointUpdate(e interface{}) {
+	switch m := e.(type) {
+	case *proto.WorkloadEndpointStatusUpdate:
+		key := names.WorkloadEndpointIDToWorkloadEndpointKey(m.Id, fr.hostname)
+		if key == nil {
+			logrus.WithField("update", e).Warn("Couldn't generate a WorkloadEndpointKey from update")
```
fasaxc, line 248: return in that case?

felix/statusrep/status_file_reporter.go
```diff
@@ -0,0 +1,347 @@
...
+		}
+		fr.statusDirDeltaTracker.Desired().Add(names.WorkloadEndpointKeyToStatusFilename(key))
+	case *proto.WorkloadEndpointStatusRemove:
+		key := names.WorkloadEndpointIDToWorkloadEndpointKey(m.Id, fr.hostname)
+		if key == nil {
+			logrus.WithField("update", e).Warn("Couldn't generate a WorkloadEndpointKey from update")
```
fasaxc, line 254: return?

libcalico-go/lib/names/workloadendpointstatus.go
```diff
@@ -0,0 +1,98 @@
...
+// Returns nil if passed endpoint is nil.
+func APIWorkloadEndpointToWorkloadEndpointKey(ep *libapi.WorkloadEndpoint) *model.WorkloadEndpointKey {
+	if ep == nil {
+		return nil
+	}
+
+	key := &model.WorkloadEndpointKey{
+		Hostname:       ep.Spec.Node,
+		OrchestratorID: ep.Spec.Orchestrator,
+		WorkloadID:     ep.Namespace + "/" + ep.Spec.Pod,
```
fasaxc, line 92: How does this handle OpenStack?  Presumably OpenStack VM WEPs don't have a namespace (or maybe they do!)

fasaxc (changes requested):

libcalico-go/lib/names/workloadendpointstatus.go
aaaaaaaalex: I have no clue, to be honest. I guess if namespace is empty, or if orchestrator is openstack, I can assign the same value that similar types do

cni-plugin/pkg/wait/wait_test.go
```diff
@@ -0,0 +1,100 @@
...
+			Expect(epFilename).NotTo(HaveLen(0), "Couldn't generate endpoint status filename from endpoint key.")
+
+			fullFilePath := filepath.Join("/tmp", epFilename)
+			_, err := os.Create(fullFilePath)
+			defer func() {
+				_ = os.Remove(fullFilePath)
+			}()
+			Expect(err).NotTo(HaveOccurred(), "Test failed to create a file in /tmp")
+
+			Eventually(exit, "15s").Should(BeClosed(), "Expected return with no error when file created.")
```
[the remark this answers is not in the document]
aaaaaaaalex: Yeah, I think this may not work since it needs to read to check closed, and may read/discard an Error. Will rewrite.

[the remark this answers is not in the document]
aaaaaaaalex: As-goes the timeout, Im trying to be very lenient for the CI, but I'll reduce them a bit.

cni-plugin/pkg/wait/wait.go
```diff
@@ -0,0 +1,197 @@
...
+
+		// Always call a stat before continuing with watcher processing.
+		// If watches are boken, the loop will then degrade into a poll.
+		// If the watch is healthy, we should still stat incase we missed
+		// the create event.
+		found, err := statFile(directory, filename)
+		if err != nil {
+			log.WithError(err).Warn("Encountered an error when polling filesystem for file")
+		} else if found {
+			break
```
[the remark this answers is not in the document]
aaaaaaaalex: factored-out the body of the retry loop so I can now just defer cleanup

felix/fv/pod_setup_wait_test.go
```diff
@@ -0,0 +1,173 @@
...
+		dummyWorkloads           [2]*workload.Workload
+		dataplaneInSyncReceivedC chan struct{}
+	)
+
+	BeforeEach(func() {
+		infra = getInfra()
+		topologyOptions = infrastructure.DefaultTopologyOptions()
+		topologyOptions.DelayFelixStart = true
+		topologyOptions.EnableIPv6 = false
+		// /tmp should be automatically mounted by an internal call to RunFelix
```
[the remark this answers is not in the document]
aaaaaaaalex: Would you still prefer a temp dir? Happy to throw it in, no hassle either way

felix/fv/pod_setup_wait_test.go
```diff
@@ -0,0 +1,173 @@
...
+		// by the infrastructure pkg when Topology is started.
+		topologyOptions.ExtraEnvVars["FELIX_ENDPOINTSTATUSPATHPREFIX"] = "/tmp"
+		topologyOptions.ExtraEnvVars["FELIX_LOGSEVERITYSCREEN"] = "DEBUG"
+		topologyOptions.ExtraEnvVars["FELIX_LOGSEVERITYFILE"] = "DEBUG"
+		topologyOptions.ExtraEnvVars["FELIX_LOGSEVERITYSYS"] = "DEBUG"
+		topologyOptions.ExtraEnvVars["FELIX_LOGDEBUGFILENAMEREGEX"] = "status_file_reporter"
```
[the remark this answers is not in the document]
aaaaaaaalex: One note on the regex here is that, in the case of a failure, all debug logs still seem to be outputted. I also noticed a potential bug in the FVs possibly pertaining to stack offset, where all logging modules' filenames are printed as "felix/<file>", i.e. `[DEBUG][71] felix/status_file_reporter.go 140: Endpoint status file reporter running.`

felix/fv/pod_setup_wait_test.go
```diff
@@ -0,0 +1,173 @@
...
+			tc.Felixes[0].Exec("touch", expectedFilename)
+			// This stat call returns the time since epoch when the file was last accessed.
+			lastAccessed, err := tc.Felixes[0].ExecOutput("stat", "--format='%X'", expectedFilename)
+			Expect(err).NotTo(HaveOccurred(), "stat call failed while trying to create a file")
+
+			By("waiting for Felix's status file reporter to become in-sync")
+			tc.Felixes[0].TriggerDelayedStart()
+			Eventually(dataplaneInSyncReceivedC, "10s").Should(BeClosed(), "receipt of DataplaneInSync message not seen in logs")
+
+			By("checking if the pre-existing file's access time changed")
```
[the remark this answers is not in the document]
aaaaaaaalex: was playing it safe here - the stat docs listed `%y` as the "time of last data modification" , and I wasn't sure if that meant the contents had to change (which we don't do). Will change if it's the semantically correct option, though

libcalico-go/lib/names/workloadendpointstatus.go
```diff
@@ -0,0 +1,96 @@
...
+		Hostname:       hostname,
+		OrchestratorID: id.OrchestratorId,
+		WorkloadID:     id.WorkloadId,
+		EndpointID:     id.EndpointId,
+	}
+}
+
+// APIWorkloadEndpointToWorkloadEndpointKey generates a WorkloadEndpointKey from the given WorkloadEndpoint.
+// Returns nil if passed endpoint is nil.
+func APIWorkloadEndpointToWorkloadEndpointKey(ep *v3.WorkloadEndpoint) *model.WorkloadEndpointKey {
```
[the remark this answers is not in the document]
aaaaaaaalex: If another area exists where this happens, it's been too obscured for me to find :/ 
There is a v1 converter which takes an `unversioned.ResourceMetadata` and does something similar, but I need a v3 endpoint due to what the CNI builds during the ADD call.

felix/fv/pod_setup_wait_test.go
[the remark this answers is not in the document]
fasaxc: Your approach has the advantage of not leaking dirs so happy to stick with that.

felix/fv/pod_setup_wait_test.go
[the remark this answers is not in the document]
fasaxc: I think you should check `mtime` , which is the %y one.  That'd be updated if we recreate the file or modify its contents.

Access could be anything, even the `stat` call.  It's often disabled at FS level due to having a high cost to update.

felix/statusrep/status_file_reporter.go
```diff
@@ -239,18 +240,12 @@ func (fr *EndpointStatusFileReporter) handleEndpointUpdate(e interface{}) {
 	switch m := e.(type) {
 	case *proto.WorkloadEndpointStatusUpdate:
 		key := names.WorkloadEndpointIDToWorkloadEndpointKey(m.Id, fr.hostname)
-		if key == nil {
-			logrus.WithField("update", e).Warn("Couldn't generate a WorkloadEndpointKey from update")
-			return
-		}
-		fr.statusDirDeltaTracker.Desired().Add(names.WorkloadEndpointKeyToStatusFilename(key))
+		fn := names.WorkloadEndpointKeyToStatusFilename(key)
+		fr.statusDirDeltaTracker.Desired().Add(fn)
```
fasaxc, line 244: Removed the nil check here?  Is that no longer possible?

libcalico-go/lib/backend/k8s/conversion/workload_endpoint_default.go
```diff
@@ -238,6 +238,7 @@ func (wc defaultWorkloadEndpointConverter) podToDefaultWorkloadEndpoint(pod *kap
 		Orchestrator:               "k8s",
 		Node:                       pod.Spec.NodeName,
 		Pod:                        pod.Name,
+		Workload:                   pod.Name,
```
fasaxc, line 241: Weird that this was needed, just reading the thread about it...

felix/statusrep/status_file_reporter.go
aaaaaaaalex: Will replace with a nil check on the original update's ID since that's the only case where the util func will return nil

fasaxc (approved):

gojoy:
This feature looks great. Is there a detailed document that explains how to use it? Thanks.
@aaaaaaaalex @fasaxc

PR #8473, opened 2024-02-01
[release-v3.27] Auto pick #8338: BPFExcludeCIDRsFromNAT allows excluding service CIDRs

tomastigera:
Cherry pick of #8338 on release-v3.27.

#8338: BPFExcludeCIDRsFromNAT allows excluding service CIDRs

# Original PR Body below

The main use case is node-local dns cache. In iptables mode with k8s provided kube-proxy, the cache installs such iptables rules that snatch the traffic before the kube-proxy rules, which would do the NAT.

However, in ebpf NAT is done way before packets may hit iptables. Often they do not hit iptables at all. Therefore, we need to tell our kube-proxy, not to translate a service IP that we want to let dns cache to process.

Unfortunatelly, we cannot annotate the service as the annotations are not propagate to the kube-proxy code by the k8s front-end.

To setup a node-local dns cache, set BPFExcludeCIDRsFromNAT as follows:

kubedns=`kubectl get svc kube-dns -n kube-system -o jsonpath={.spec.clusterIP}`

BPFExcludeCIDRsFromNAT=$kubedns

fixes https://github.com/projectcalico/calico/issues/4689

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: setting BPFExcludeIPsFromNAT allows node-local dns cache to work
```

## Reminder for the reviewer

Other optional labels:


tomastigera:
unrelated errors in CI

PR #8476, opened 2024-02-02
CTLB changes - v46 program will handle both v4 and v6.

sridhartigera:
## Description

This PR addresses the below scenarios related to CTLB for dual stack.

 1. V6 programs are loaded into the jump map when in dual stack mode.
 2. v46 program handles both IPv4 masqueraded as ipv6 and real IPv6. when receiving an IPv6 packet, v46 program jumps to the v6 program.
 
Also the CTLB object files are loaded once.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/nat/connecttime.go
+++ b/felix/bpf/nat/connecttime.go
@@ -30,6 +30,7 @@ import (
 	"github.com/projectcalico/calico/felix/bpf"
 	"github.com/projectcalico/calico/felix/bpf/bpfdefs"
 	"github.com/projectcalico/calico/felix/bpf/bpfutils"
+	"github.com/projectcalico/calico/felix/bpf/jump"
 	"github.com/projectcalico/calico/felix/bpf/libbpf"
 	"github.com/projectcalico/calico/felix/bpf/maps"
 	"github.com/projectcalico/calico/felix/bpf/utils"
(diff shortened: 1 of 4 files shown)


felix/bpf/nat/connecttime.go
```diff
@@ -161,60 +204,93 @@ func InstallConnectTimeLoadBalancer(ipFamily int, cgroupv2 string, logLevel stri
...
-			if err != nil {
+		// If dual-stack, populate the jump maps with v6 ctlb programs.
+		if v6Obj != nil {
+			if err := ctlbProgsMap.EnsureExists(); err != nil {
+				log.WithError(err).Error("Failed to create CTLB programs maps")
 				return err
 			}
-
-			err = installProgram("recvmsg", "6", bpfMount, cgroupPath, logLevel, udpNotSeen, false)
+			err = updateCTLBJumpMap(ctlbProgsMap.GetName(), v6Obj)
```
tomastigera, line 268: My expectation would be that you update the map in v46Obj. It just happens that you have the map in all of the 3 objects and pinned to the same one, so this update is "correct", but if you had it in v46 only, this would not work imo. v6 is not going to jump anywhere.

felix/bpf/nat/connecttime.go
```diff
@@ -161,60 +204,93 @@ func InstallConnectTimeLoadBalancer(ipFamily int, cgroupv2 string, logLevel stri
...
 			}
+		} else {
+			//delete the jump map
+			os.Remove(ctlbProgsMap.Path())
 		}
-	default:
-		return fmt.Errorf("unrecognized ip family %d", ipFamily)
+		return nil
+	}
+	err = attachProgram("connect", "6", bpfMount, cgroupPath, udpNotSeen, excludeUDP, v6Obj)
```
tomastigera, line 278: This is not covered by tests, but wouldn't this be attached "over" the v46 program?

tomastigera (commented):
Feel like merging this should be deferred until the dual-stack switch is made in the ebpf ep mgr so that it can have proper testing.

felix/bpf/nat/connecttime.go
sridhartigera: I have a return within the IPv4Enabled case. So this should not happen.

Issue #8423, opened 2024-01-19
Terminating pod still receiving service traffic


nascd:
Hi Calico team,
We recently noticed an issue where a CoreDNS pod that was stuck in Terminating status seemed to still be getting traffic sent to it.  We use the eBPF kube-proxy replacement in Calico, and we enable the use of endpoint slices for Felix via https://github.com/projectcalico/calico/blob/master/api/pkg/apis/projectcalico/v3/felixconfig.go#L501 (though I can't actually find any code that references that param).

When a Pod is in terminating state, Endpoint slices will still list the endpoint, but will include some additional context via endpoint conditions that applications can use to evaluate whether an endpoint should be treated as healthy and accepting of traffic.  My understanding is that terminating pods will have the ready condition set to false, while having the serving and terminating conditions set to true (ref: https://kubernetes.io/docs/tutorials/services/pods-and-endpoint-termination-flow/).

I tried looking at how Calico evaluates endpoint slices and I found the membersFromEndpointSlice function in Felix' service index (ref: https://github.com/projectcalico/calico/blob/master/felix/serviceindex/serviceindex.go#L195) -- it doesn't appear as though that particular function evaluates endpoint conditions, but I'm not familiar with the calico codebase and it's not clear to me that I'm looking at the relevant method.  Can you please help me understand how Felix behaves with respect to endpointslice conditions, and whether it will continue to route traffic to terminating endpoints, as my empirical experience seems to indicate?

If that is the case, please accept this as a feature request to optionally consider the status of endpoint conditions and to exclude endpoints that are in the process of terminating.

Thank you!

## Expected Behavior
I expect pods in Terminating status to receive no service traffic.

## Current Behavior
A pod that is stuck in Terminating status appears to still receive service traffic.

## Possible Solution
Ensure that endpointslice conditions are evaluated, and if an endpointslice is stated as terminating, discontinue sending traffic to it.

## Steps to Reproduce (for bugs)
Create a service with two healthy backing pods.
Force one pod into a terminating status, for example by causing a network interruption on the node hosting it.
Evaluate that some traffic does not make it to your remaining healthy pod.

## Context
In the case of CoreDNS, this issue presents with clients having intermittent DNS resolution failures, and has the potential to be quite damaging to a cluster's health.

## Your Environment
calico 3.26, kubernetes 1.24, eBPF kube-proxy replacement with endpointslices enabled.

tomastigera:
[...]

PR #8480, opened 2024-02-02
[BPF] use a label to clean up conntrack to terminating udp backends

tomastigera:
When a backend pod of a service is in terminating state, we keep existing connections as that pod is likely in a terminating grace period. However, a pod may get stuck in terminating state. New connections are not directed to that pod and if there is no connectivity to the pod TCP will fail eventually and new connections will select another backend. However UDP follows the established conntrack. So labeling a service with

projectcalico.org/udpConntrackCleanup = terminatingImmediately

would ensure that we clean up conntrack entries towards this backend once it becomes terminating. This make the dataplane pick another backend once there is a conntrack miss, which unblocks the client.

Fixes https://github.com/projectcalico/calico/issues/8423

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/proxy/proxy.go
+++ b/felix/bpf/proxy/proxy.go
@@ -173,7 +173,7 @@ func New(k8s kubernetes.Interface, dp DPSyncer, hostname string, opts ...Option)
 		p.recorder,
 		nil,
 	)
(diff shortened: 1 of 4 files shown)


sridhartigera (approved):
LGTM

PR #8489, opened 2024-02-07
Run 2 copies of BPF DP components, one for v4 and one for v6

sridhartigera:
## Description

Run 2 copies of the below BPF dataplane components, one for IPv4 and one for IPv6
1. BPF kube-proxy
2. BPF route manager
3. Failsafe manager

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -269,7 +269,8 @@ type bpfEndpointManager struct {
 	dataIfaceRegex          *regexp.Regexp
 	l3IfaceRegex            *regexp.Regexp
 	workloadIfaceRegex      *regexp.Regexp
-	ipSetIDAlloc            *idalloc.IDAllocator
+	ipSetIDAllocV4          *idalloc.IDAllocator
+	ipSetIDAllocV6          *idalloc.IDAllocator
 	epToHostAction          string
 	vxlanMTU                int
 	vxlanPort               uint16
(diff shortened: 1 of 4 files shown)


felix/dataplane/linux/bpf_route_mgr.go
```diff
@@ -610,12 +629,15 @@ func (m *bpfRouteManager) onRouteUpdate(update *proto.RouteUpdate) {
...
 func (m *bpfRouteManager) onRouteRemove(update *proto.RouteRemove) {
 	cidr := ip.MustParseCIDROrIP(update.Dst)
-	if !m.ipv6Enabled {
+	if m.ipFamily == proto.IPVersion_IPV4 {
 		if _, ok := cidr.(ip.V4CIDR); !ok {
 			return
 		}
+	} else {
+		if _, ok := cidr.(ip.V6CIDR); !ok {
+			return
```
tomastigera, line 638: It feels like it almost asks for making `bpfRouteManager` a generic parametrized with `ip.V4CIDR` or `ip.CIDR has `Version() uint8` method so we would not even need to know the type (my bad :sweat_smile: )

felix/dataplane/linux/bpf_route_mgr.go
```diff
@@ -679,7 +701,7 @@ func (m *bpfRouteManager) getWorkloadCIDRs(wep *proto.WorkloadEndpoint) (cidrs [
 	if wep == nil {
 		return
 	}
-	if m.ipv6Enabled {
+	if m.ipFamily == proto.IPVersion_IPV6 {
```
tomastigera, line 704: this would not work with the generic

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -269,7 +269,8 @@ type bpfEndpointManager struct {
 	dataIfaceRegex          *regexp.Regexp
 	l3IfaceRegex            *regexp.Regexp
 	workloadIfaceRegex      *regexp.Regexp
-	ipSetIDAlloc            *idalloc.IDAllocator
+	ipSetIDAllocV4          *idalloc.IDAllocator
+	ipSetIDAllocV6          *idalloc.IDAllocator
```
tomastigera, line 273: do we strictly need two allocators? (idk)

tomastigera (commented):
Some minor stuff. I think it would be better to include it with the change that enabled dual stack. This change should come at least with some testing, like did we program both v4 and v6 maps if we saw both etc. Likely something for FV when we have the dual stack.

PR #8521, opened 2024-02-16
Reduce number of docker exec calls when killing FV workloads.

fasaxc:
## Description

Felix FV infrastructure performance tweaks:

* When starting workload, output PID to stdout and read it immediately (saves `docker exec ... cat <pidfile>` later on).
* Combine workload teardown commands into one `sh -c "..."` invocation, saves a few more `docker exec` transactions, which cost ~100ms each.

Benchmarking locally, I see 1-200ms saving per teardown.  Seems to add up to ~1-2min across a run.  Smallish win for a small change.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/workload/workload.go
+++ b/felix/fv/workload/workload.go
@@ -61,6 +61,7 @@ type Workload struct {
 	isRunning             bool
 	isSpoofing            bool
 	listenAnyIP           bool
+	pid                   string
 
 	cleanupLock sync.Mutex
 }
@@ -94,13 +95,8 @@ func (w *Workload) Stop() {
 		log.Info("Stop no-op because nil workload")
 	} else {
 		log.WithField("workload", w).Info("Stop")
-		output, err := w.C.ExecOutput("cat", fmt.Sprintf("/tmp/%v", w.Name))
-		Expect(err).NotTo(HaveOccurred(), "failed to run docker exec command to get workload pid")
-		pid := strings.TrimSpace(output)
-		w.C.Exec("kill", pid)
-		_ = w.C.ExecMayFail("ip", "link", "del", w.InterfaceName)
-		_ = w.C.ExecMayFail("ip", "netns", "del", w.NamespaceID())
-		_, err = w.runCmd.Process.Wait()
+		_ = w.C.ExecMayFail("sh", "-c", fmt.Sprintf("kill -9 %s & ip link del %s & ip netns del %s & wait", w.pid, w.InterfaceName, w.NamespaceID()))
+		_, err := w.runCmd.Process.Wait()
 		if err != nil {
 			log.WithField("workload", w).Error("failed to wait for process")
 		}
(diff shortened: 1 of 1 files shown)


felix/fv/workload/workload.go
```diff
@@ -253,12 +248,22 @@ func (w *Workload) Start() error {
 		}
 	}()
 
+	pid, err := stdoutReader.ReadString('\n')
+	if err != nil {
+		// (Only) if we fail here, wait for the stderr to be output before returning.
+		defer errDone.Wait()
+		if err != nil {
```
nelljerram, line 255: Superfluous to repeat `if err != nil` here.

nelljerram: Oh, I see it's a copy from below.  Maybe fix all of them?

felix/fv/workload/workload.go
```diff
@@ -94,13 +95,8 @@ func (w *Workload) Stop() {
...
 	} else {
 		log.WithField("workload", w).Info("Stop")
-		output, err := w.C.ExecOutput("cat", fmt.Sprintf("/tmp/%v", w.Name))
-		Expect(err).NotTo(HaveOccurred(), "failed to run docker exec command to get workload pid")
-		pid := strings.TrimSpace(output)
-		w.C.Exec("kill", pid)
-		_ = w.C.ExecMayFail("ip", "link", "del", w.InterfaceName)
-		_ = w.C.ExecMayFail("ip", "netns", "del", w.NamespaceID())
-		_, err = w.runCmd.Process.Wait()
+		_ = w.C.ExecMayFail("sh", "-c", fmt.Sprintf("kill -9 %s & ip link del %s & ip netns del %s & wait", w.pid, w.InterfaceName, w.NamespaceID()))
```
nelljerram, line 98: Hmm, there are a few details changing here, are you sure they are all intended and reliable?
- If `kill` failed, we'd previously fail the test at that point.
- Added `-9` to the kill.  Are you sure we didn't want to give the process a chance to shut down gracefully?
- `kill` and `ip` commands happening in parallel, whereas previously they were sequenced.  Will the namespace deletion fail if the kernel sees it before link deletion?

fasaxc: Now that I've switched to `kill -9`, I think it can only fail is the process isn't running, which is probably OK given this is only called during AfterEach shut-down.  I changed to that to reduce the output logged on shutdown.  IIRC it was emitting a panic trace before, which was misleading/wasteful.

The other two commands seem OK to run in either order.  Deleting the netns actually deletes the veth too if the workload side of the veth was correctly moved into the netns but I thought it best to assume that moving the veth failed.

nelljerram: Fair enough; make sense, and I'm happy if you think this will all be fine.

nelljerram (approved):

felix/fv/workload/workload.go
fasaxc: I decided that I wasn't happy that there was no timeout on the Wait() call, especially if I increase a little extra risk here.  Added one along with the other markups.

PR #8538, opened 2024-02-21
EV-4698 Remove unnecessary FIPS code

rene-dekker:
The FIPS_ENABLED flags precede building with fips_strict mode. With the introduction of that flag, there is no point any longer of passing a bool into the `NewTLSConfig()` func.

Also, did some cleanup of the Makefile for the crypto suite.

### DIFF
--- a/crypto/Makefile
+++ b/crypto/Makefile
@@ -9,25 +9,21 @@ ci: fv
 .PHONY: fv
 fv: setup-fv bin/fips-test-build
 	# Run a server using the FIPS strict mode settings and wait 1s to let it come up.
-	./bin/fips-test-build &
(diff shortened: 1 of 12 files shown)


crypto/Makefile
```diff
@@ -9,24 +9,21 @@ ci: fv
 .PHONY: fv
 fv: setup-fv bin/fips-test-build
 	# Run a server using the FIPS strict mode settings and wait 1s to let it come up.
-	./bin/fips-test-build &
+	$(DOCKER_RUN) --name fips-test-build --rm $(CALICO_BUILD) ./bin/fips-test-build &
```
hjiawei, line 12: Can we use `-d` to run as a [background](https://docs.docker.com/engine/reference/run/#foreground-and-background) process?

hjiawei (approved):

PR #8552, opened 2024-02-22
Update calico postrelease tests [master]

danudey:
Add some fixes to the postrelease tests and bring them forward to the master branch


hack/postrelease/Makefile
```diff
@@ -0,0 +1,68 @@
...
+	@touch .docker_container.stamp
+
+## Execution
+
+run_pytests_docker: create_docker_container require-vars-QUAY_TOKEN
+	$(info +++ Running postrelease tests)
+	@docker run -it --rm                     \
+		-v .:/code                           \
+		-v /run/docker.sock:/run/docker.sock \
+		-e QUAY_TOKEN                        \
```
lwr20, line 49: Why is a quay token needed? These images need to be publicly available, so I think we should be testing without a quay-token.

hack/postrelease/tests/test_images.py
```diff
@@ -0,0 +1,208 @@
...
+        tag_suffixed = f"{variables.RELEASE_VERSION}-{tag_suffix}"
+        ALL_IMAGES.append((image, tag_suffixed))
+
+print("Found {} separate image tags to test".format(len(ALL_IMAGES)))
+
+
+def request_quay_image(image_name, image_version):
+    headers = {
+        "Content-Type": "application/json",
+        "Authorization": "Bearer {}".format(variables.QUAY_TOKEN),
```
lwr20, line 72: I think we should be testing without a quay-token.

hack/postrelease/tests/variables.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# Get flannel version.
+FLANNEL_VERSION = get_var_or_error("FLANNEL_VERSION")
+
+# Get expected operator version.
+OPERATOR_VERSION = get_var_or_error("OPERATOR_VERSION")
+
+# Quay.io API token
+QUAY_TOKEN = get_var_or_error("QUAY_TOKEN")
```
lwr20, line 35: I think we should be testing without a quay-token.

hack/postrelease/Dockerfile
```diff
@@ -1,13 +1,32 @@
-FROM docker:18.09
+# hadolint global ignore=DL3018
+FROM docker:24
```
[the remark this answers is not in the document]
danudey: The short answer is that we want a docker image which already has docker configured correctly. Installing and configuring docker is likely to be more complex/error-prone as compared to installing Python 3, which requires no configuration.

hack/postrelease/Makefile
danudey: Agree. I'll test and ensure we can perform these tests without `QUAY_TOKEN`, and then remove it.

hack/postrelease/tests/test_images.py
danudey: Agree (as above)

hack/postrelease/tests/variables.py
danudey: Agree (as above)

hack/postrelease/tests/test_images.py
```diff
@@ -0,0 +1,208 @@
...
+    "calico/pod2daemon-flexvol",
+    "calico/csi",
+]
+
+TAG_SUFFIXES = [
+    "amd64",
+    "arm64",
+    "armv7",
+    "fips-amd64",
+    "fips",
```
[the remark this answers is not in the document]
danudey: They still appear to be publishing to quay.io for master branch, but I'll check and see if that's something we do actually need going forward (and for which versions).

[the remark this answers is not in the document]
danudey: @hjiawei confirms that we should still be publishing/testing these images.

hack/postrelease/tests/test_images.py
```diff
@@ -0,0 +1,208 @@
...
+# a registry assigned.
+EXPECTED_IMAGES = [
+    "calico/node",
+    "calico/ctl",
+    "calico/apiserver",
+    "calico/typha",
+    "calico/cni",
+    "calico/kube-controllers",
+    "calico/upgrade",
+    "calico/windows",
```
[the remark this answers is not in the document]
coutinhop: That's right @lwr20, `windows` and `windows-upgrade` should be removed starting v3.27.0, and also `node-windows` and `cni-windows` should be added

caseydavenport:
@danudey is this PR still in progress or should it be closed?

PR #8553, opened 2024-02-22
Update calico postrelease tests [v3.27]

danudey:
Cherry-pick of #8552

PR #8554, opened 2024-02-22
Cherry-pick calico postrelease test fixes [v3.26]

danudey:
Cherry-pick fixes from #8553

### DIFF
--- a/hack/postrelease/Makefile
+++ b/hack/postrelease/Makefile
@@ -1,5 +1,32 @@
 include variables.mk
 
+### Binaries we need
+
+bin/yq:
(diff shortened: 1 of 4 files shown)


hack/postrelease/Makefile
```diff
@@ -22,9 +41,9 @@ create_docker_container: .docker_container.stamp
 
 ## Execution
 
-run_pytests_docker: create_docker_container
+run_pytests_docker: create_docker_container require-vars-QUAY_TOKEN
```
lwr20, line 44: Not sure we should be using a quay token at all.

PR #8586, opened 2024-03-04
Use distinct UIDs for Profiles generated for Namespaces and ServiceAccounts

caseydavenport:
## Description

As discussed in
https://github.com/projectcalico/calico/issues/5715#issuecomment-1414047534,
we currently have a problem with how we handle these automatically
generated Profile resources.

Namely, they are created with the same UID as the Namespace or
ServiceAccount from which they are generated. This was originally OK
since they never showed up in the API, but with the introduction of the
Calico API server they now appear in the Kubernetes API, thus confusing
things like Kubernetes cascading deletion via OwnerReferences.

This can cause confusing and subtle errors whereby deleting a parent
resource does not clean up the child resources in the cluster, and can
appear intermittently depending on event ordering and the garbage
collector's view of the world.

This PR attempts to generate a unique-but-still-deterministic UID for
use in this case, so that Profiles generated for Namespaces and
ServiceAccounts do not appear to be the same exact object as the
Namespace / ServiceAccount that it represents.

Note that the linked issue discusses two problems with our use of UID,
and that this PR only fixes the issue as it relates to these read-only
Profile objects. Another solution is required for CRDs that support
Write operations.

## Related issues/PRs

https://github.com/projectcalico/calico/issues/5715

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix bug that inhibited garbage collection of Namespaces and ServiceAccounts with OwnerReferences.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/k8s/conversion/conversion.go
+++ b/libcalico-go/lib/backend/k8s/conversion/conversion.go
@@ -19,9 +19,11 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 	kapiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	discovery "k8s.io/api/discovery/v1"
 	networkingv1 "k8s.io/api/networking/v1"
@@ -36,9 +38,7 @@ import (
 	cnet "github.com/projectcalico/calico/libcalico-go/lib/net"
 )
 
-var (
-	protoTCP = kapiv1.ProtocolTCP
-)
+var protoTCP = kapiv1.ProtocolTCP
 
 type selectorType int8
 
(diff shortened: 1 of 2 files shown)


libcalico-go/lib/backend/k8s/conversion/conversion.go
```diff
@@ -809,3 +823,15 @@ func stringsToIPNets(ipStrings []string) ([]*cnet.IPNet, error) {
 	}
 	return podIPNets, nil
 }
+
+// ConvertUID converts a UID to a new UID in a deterministic way. This is useful when we want to generate a new UID
+// for a resource that is derived from another resource, but we don't want to use the same UID in order to
+// ensure that the new resource is treated as unique. This is important, as two objects with the same UID causes
+// confusion in the Kubernetes garbage collection logic.
+func ConvertUID(uid types.UID) (types.UID, error) {
+	newUID, err := uuid.NewRandomFromReader(strings.NewReader(string(uid)))
```
caseydavenport, line 832: My big concern with this is that NewRandomFromReader loses randomness, as it's simply reading the input bytes into a new UUID and forcing the Version+Variant bits to the correct values. IIUC, it's really not too different from Rob's original approach of reordering the using the bits from the original UUID except that it loses more randomness because the non-random `-` characters are included in the resulting UUID.

We could also go with the original approach that Rob used by reversing the values and then asserting the Version+Variant bits in the same way that NewRandomFromReader does.

Either way, this function should only end up generating UIDs for:
- Profiles generated from Namespaces and ServiceAccounts
- v3 resources backed by CRDs NOT created via the apiserver or calicoctl (since those two tools generate a truly random UUID which is stored in the annotation)

libcalico-go/lib/clientv3/ippool_kdd_conversion_test.go
```diff
@@ -254,7 +253,8 @@ var _ = testutils.E2eDatastoreDescribe("IPPool KDD v1 to v3 migration tests", te
 
 			By("Updating the IPPool from the API client with the non-writable v1 IPIP field")
 			_, outError = c.IPPools().Update(ctx, &apiv3.IPPool{
-				ObjectMeta: metav1.ObjectMeta{Name: name1, ResourceVersion: "555", CreationTimestamp: metav1.Now(), UID: "a-rabbit-ate-my-carrot"},
+				// R.I.P: 'a-rabbit-ate-my-carrot' is no longer a valid UID according to Calico.
```
gunjan5, line 256: 🐰 🥕 😭

prateekgogia:
Can someone please share what happens in case kube-controller-manager is not restarted?

caseydavenport:
@prateekgogia that behavior isn't defined / deterministic. The kube-controller-manager will potentially be running on outdated information.

prateekgogia:
Is it safe to say, without restarting the kube-controller-manager no functionality is impacted only objects are not garbage collected which should be?

fasaxc:
I don't think the exact "blast radius" is known, and garbage collection not working can make your cluster grind to a halt, so it's not a problem to ignore.

PR #8608, opened 2024-03-13
handle proto.DataplaneInSync msg in status_reporter

aaaaaaaalex:
## Description

A panic is occurring on OpenStack Mainline ST due to an unhandled protobuf type recently added 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/status_report_test.go
+++ b/felix/fv/status_report_test.go
@@ -1,0 +1,104 @@
+// Copyright (c) 2024 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build fvtests
+
+package fv_test
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	api "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+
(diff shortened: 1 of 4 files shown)


felix/statusrep/status_reporter.go
```diff
@@ -182,6 +182,8 @@ loop:
 					Hostname:   esr.hostname,
 					EndpointID: msg.Id.EndpointId,
 				}
+			case *proto.DataplaneInSync:
```
fasaxc, line 185: Could we remove the old in-sync channel completely now?  This should be better.

felix/fv/status_report_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+			return statuses.KVPairs
+		}
+
+		It("should write endpoint to etcd", func() {
+			Eventually(getStatuses, "10s").Should(HaveLen(1))
+			kvs := getStatuses()
+			wepStatusKey, ok := kvs[0].Key.(model.WorkloadEndpointStatusKey)
+			Expect(ok).To(BeTrue(), "Unexpected key type when listing workload endpoint statuses")
+			Expect(wepStatusKey.Hostname).To(BeEquivalentTo(tc.Felixes[0].Hostname))
```
fasaxc, line 96: What does BeEquivalentTo() do vs Equal() (I tend to use the latter)

fasaxc (commented):
Looking good, I think I'd remove the old in-sync channel too, WDYT?

felix/statusrep/status_reporter.go
aaaaaaaalex: I guess so, but avoided out of nerves of breaking something else :laughing: WIll give it a shot

felix/fv/status_report_test.go
aaaaaaaalex: BeEquivalentTo neglects to check types are the same, as I understand it. Still checks values align, not sure of the specifics beyond that. Equal() would work here too

felix/statusrep/status_reporter.go
aaaaaaaalex: Done - will monitor CI in-case I broke anything else :laughing:

fasaxc (approved):
LGTM

PR #8615, opened 2024-03-14
autodetect ipv6 as well

nathanael-nsone:
## Description
Fixes an issue where a configured v6 BGPPeer is silently ignored.
This adds automatic detection for IPv6 as well.
Without this bird will treat a dual stack node as though IPv6 is disabled on it.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/charts/calico/templates/calico-node.yaml
+++ b/charts/calico/templates/calico-node.yaml
@@ -311,6 +311,8 @@ spec:
             # Auto-detect the BGP IP address.
             - name: IP
               value: "autodetect"
+            - name: IP6
+              value: "autodetect"
             # {{- if .Values.bpf }} For best performance, disable {{- else }} Enable {{- end }} IPIP
             - name: CALICO_IPV4POOL_IPIP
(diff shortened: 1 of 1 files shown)


matthewdupre (approved):

caseydavenport:
I think it is intentional that we don't autodetect IPv6 by default in order to make this work out of the box on most clusters. Note that Felix IPv6 support is also disabled: https://github.com/projectcalico/calico/blob/faaf362533b672135404c8ce62b73c018ebf6550/charts/calico/templates/calico-node.yaml#L384-L386

Our docs describe how to enable dual-stack support here: https://docs.tigera.io/calico/latest/networking/ipam/ipv6#enable-dual-stack

If we make this change (which I am not convinced we should, yet):
- We also need to update the FELIX_IPV6SUPPORT env var.
- We also need to update the docs to explain that dual-stack / v6 is enabled by default.

All of that said, I'd recommend using the tigera/operator to install Calico since that handles all of this complexity without the need to keep these env vars in-sync with each other.

PR #8618, opened 2024-03-14
[BPF] When lo has IP override it when CTLB is disabled

tomastigera:
When CTLB is disabled, we route traffic for services via bpfnatin/out device. Since the final destination isn't resolved yet, Linux picks up an address set on loopback device (if there is any) as source. This may not be (likely is not) an address that can be used by the destination to return traffic. Therefore we need to replace it by the host's IP that is routable within the cluster.

We use the same mechanism as for replacing main host device IP with a tunnel IP when we need to reach a remote workload via an overlay.

If the IP set on loopback is the host IP (say du to dual ToR connectivity), that would work, because the override of the lo IP kicks in only if that IP differs.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: fixed source IP used by host when CTLB is disabled and loopback device has non-local IP set.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -760,6 +760,12 @@ func (m *bpfEndpointManager) updateHostIP(ip net.IP, ipFamily int) {
 			m.dirtyIfaceNames.Add(ifaceName)
 		}
 		m.ifacesLock.Unlock()
+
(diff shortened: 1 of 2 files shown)


fasaxc (approved):
LGTM

marvin-tigera:

PR #8642, opened 2024-03-21
Prevent upgrade-ipam for host-local IPAM

robertvolkmann:
## Description

Otherwise, the init container upgrade-ipam would clear the state of the host-local plugin, potentially causing it to reassign IPs that are still in use.

Example:
```
 Warning  FailedCreatePodSandBox  102s  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "ff47431fb9a1758c7a14688e6e6339de5bd475ade3e68f73139cc0a4565f31c9": plugin type="calico" failed (add): error adding host side routes for interface: calic2c3994f071, error: route (Ifindex: 44, Dst: 10.233.64.2/32, Scope: link) already exists for an interface other than 'calic2c3994f071': route (Ifindex: 25, Dst: 10.233.64.2/32, Scope: link, Iface: calib8e5d9c71dc)
```

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

### DIFF
--- a/charts/calico/templates/calico-node.yaml
+++ b/charts/calico/templates/calico-node.yaml
@@ -43,7 +43,7 @@ spec:
       terminationGracePeriodSeconds: 0
       priorityClassName: system-node-critical
       initContainers:
-{{- if and (eq .Values.network "calico") (eq .Values.datastore "kubernetes") }}
+{{- if and (eq .Values.network "calico") (eq .Values.datastore "kubernetes") (eq .Values.ipam "calico-ipam") }}
         # This container performs upgrade from host-local IPAM to calico-ipam.
         # It can be deleted if this is a fresh installation, or if you have already
         # upgraded to use calico-ipam.
--- a/charts/calico/templates/calico-node.yaml
+++ b/charts/calico/templates/calico-node.yaml
@@ -637,7 +637,7 @@ spec:
           secret:
             secretName: calico-etcd-secrets
             defaultMode: 0400
-{{- else if and (eq .Values.network "calico") (eq .Values.datastore "kubernetes") }}
+{{- else if and (eq .Values.network "calico") (eq .Values.datastore "kubernetes") (eq .Values.ipam "calico-ipam") }}
         # Mount in the directory for host-local IPAM allocations. This is
         # used when upgrading from host-local to calico-ipam, and can be removed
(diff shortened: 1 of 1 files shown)


caseydavenport:
@robertvolkmann I think this change looks OK, however from what I can tell this will have no effect - all of the manifests that use Calico networking and Kubernetes storage already use calico-ipam. 

Namely:

```
$ grep -r "This container performs" .
```
```
./calico-vxlan.yaml:        # This container performs upgrade from host-local IPAM to calico-ipam.
./calico-bpf.yaml:        # This container performs upgrade from host-local IPAM to calico-ipam.
./calico.yaml:        # This container performs upgrade from host-local IPAM to calico-ipam.
./calico-typha.yaml:        # This container performs upgrade from host-local IPAM to calico-ipam.
./flannel-migration/calico.yaml:        # This container performs upgrade from host-local IPAM to calico-ipam.
```

Would you mind giving some more detail about how this is impacting you? How have you installed / upgraded Calico?

robertvolkmann:
Is the combination of Calico networking and host-local IPAM unsupported?

caseydavenport:
I'm trying to catch back up to speed on this, so apologies as I figure it out! It seems that the only manifest that we ship as part of this repository that uses host-local IPAM appears to be `calico-policy-only.yaml` which uses host-local IPAM without any Calico networking backend configured, relying on external configuration of cloud routes.

This document talks about it as well: https://docs.tigera.io/calico/latest/reference/configure-cni-plugins#using-host-local-ipam

> Host local IPAM is generally only used on clusters where integration with the Kubernetes [route controller](https://kubernetes.io/docs/concepts/architecture/cloud-controller/#route-controller) is necessary.

I do believe it is possible to configure Calico to run using the host-local IPAM plugin and BGP + IPIP encapsualtion though. We just don't have any off-the-shelf manifests that do that.

It seems like kubespray might, which is fine, but they vendor their own manifests and so I don't think this change will affect kubespray at all.

robertvolkmann:
Thank you for your clarifications. I though it may be helpful to prevent an not working setup with the calico Helm Chart.

I will fix the kubespray manifests.

PR #8655, opened 2024-03-26
Properly translate owner refs

caseydavenport:
## Description

Make sure that we are performing the same translation on OwnerReference UIDs for projectcalico.org/v3 resources as we are performing on the actual UIDs for those resources.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


go.mod
```diff
@@ -87,19 +87,20 @@ require (
 	gopkg.in/go-playground/validator.v9 v9.30.2
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.27.10
-	k8s.io/apiextensions-apiserver v0.27.10
-	k8s.io/apimachinery v0.27.10
-	k8s.io/apiserver v0.27.10
-	k8s.io/client-go v0.27.10
+	k8s.io/api v0.29.0
```
hjiawei, line 90: Is this intentional to bump to k8s v1.29 release train? They are replace in the `replace` section below back to v1.27 release. Shall we update both?

caseydavenport: @hjiawei nope. pushed that before I was ready to. That was a consequence of `go mod tidy` doing something I didn't want it to :laughing: 

I had to pin the controller-runtime import to make it behave.

PR #8656, opened 2024-03-27
CNI wait file waits for status up

aaaaaaaalex:
## Description

Bugfix: status_file_reporter currently writes files too early. It should wait for a WorkloadEndpointStatusUpdate to have status "up"

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/statusrep/status_file_reporter.go
+++ b/felix/statusrep/status_file_reporter.go
@@ -32,7 +32,9 @@ import (
 )
 
 const (
-	dirStatus = "endpoint-status"
+	dirStatus  = "endpoint-status"
+	statusUp   = "up"
+	statusDown = "down"
 )
 
 // EndpointStatusFileReporter writes a file to the FS
@@ -55,6 +57,8 @@ type EndpointStatusFileReporter struct {
 	reapplyInterval time.Duration
 
 	hostname string
+
+	filesys
 }
 
 // Backoff wraps a timer-based-retry type which can be stepped.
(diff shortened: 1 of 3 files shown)


felix/statusrep/status_file_reporter.go
```diff
@@ -55,6 +58,8 @@ type EndpointStatusFileReporter struct {
 	reapplyInterval time.Duration
 
 	hostname string
+
+	*sync.Mutex
```
fasaxc, line 62:
```suggestion
	lock sync.Mutex
```
The zero value of `sync.Mutex` is 100% valid so you don't need to have a pointer and use `new`  You can just leave it initialised and start using it.

I generally prefer not to embed the mutex in the object (and thuse expose public Lock/Unlock methods).  The mutex is an implementation detail; you should be able to swap it for channels, atomics or something else as needed down the line without worrying that someone external might call Lock/Unlock.

felix/statusrep/status_file_reporter.go
```diff
@@ -322,6 +351,13 @@ func (fr *EndpointStatusFileReporter) deleteStatusFile(name string) error {
 	return os.Remove(filename)
 }
 
+// GetSetDeltaTracker is thread-safe.
+func (fr *EndpointStatusFileReporter) GetSetDeltaTracker() *deltatracker.SetDeltaTracker[string] {
+	fr.Lock()
+	defer fr.Unlock()
```
fasaxc, line 357: The delta tracker itself is not thread safe so this lock is protecting access to the field, but not the data inside the tracker.  You could return the tracker, and then have it be mutated while reading from it, which may panic.

To make this pattern work you need to do the read from the delta tracker inside the lock.

felix/statusrep/status_file_reporter.go
```diff
@@ -322,6 +350,13 @@ func (fr *EndpointStatusFileReporter) deleteStatusFile(name string) error {
 	return os.Remove(filename)
 }
 
+// GetSetDeltaTracker is thread-safe.
+func (fr *EndpointStatusFileReporter) GetSetDeltaTracker() *deltatracker.SetDeltaTracker[string] {
```
fasaxc, line 355:
```suggestion
func (fr *EndpointStatusFileReporter) desiredFilenamesContains(filename string) bool {
```
* You could make this method private since it's only for testing.
* You need to do the read of the deltatracker with lock held so it needs to go in this method.

fasaxc (changes requested):
I think the code has a concurrency issue as it stands.  I can't say I'm too keen on adding locks just to be able to poke inside an object from a test; I'd prefer looking at what files get written if the test is already writing to the FS, or, mocking out file access and putting the lock in the test code.

Adding extra locks that aren't needed at runtime just adds some risk and overhead whereas putting them in the test harness means that their blast radius is the tests only.

fasaxc (approved):

PR #8659, opened 2024-03-27
Clean up openstack release promotions [cp 8657]

danudey:


.semaphore/release/release.yml
```diff
@@ -2,32 +2,44 @@ version: v1.0
...
       # Only run on branches, not PRs.
       when: "branch !~ '.+'"
     task:
       secrets:
-      - name: quay-robot-calico+semaphoreci
-      - name: docker
-      - name: oss-release-secrets
-      - name: google-service-account-for-gce
-      - name: openstack-signing-publishing
+        - name: quay-robot-calico+semaphoreci
```
caseydavenport, line 19: All of the other pre-existing lists don't use indentation like this - I think we should keep consistent with that.

caseydavenport (approved):

Issue #8705, opened 2024-04-09
calico-kube-controller crash with concurrent map read and map write


zamog:
## Expected Behavior
Not crash

## Current Behavior
calico-kube-controller exits every few hours with fatal error: concurrent map read and map write.
running the controller with debug reduces the race conduction to once a ~day

## Possible Solution
Add mutex lock to the set function
PR [8706](https://github.com/projectcalico/calico/pull/8706)

## Steps to Reproduce (for bugs)
Running Calico on production environments with high pod/nodes state change.
(We fail to reproduce this behavior on non production cluster)

## Context
```
2024-03-21 16:16:16.233 [DEBUG][1] cache.go 133: converter.WorkloadEndpointData{PodName:"daemonset-daemonset-extended-1711037756-1711037762-vgqms", Namespace:"kuberhealthy", Labels:map[string]string{"app":"daemonset-daemonset-extended-1711037756-1711037762", "checkRunTime":"1711037762", "controller-revision-hash":"7f95f86fd5", "creatingInstance":"daemonset-extended-1711037756", "khcheck":"daemonset", "pod-template-generation":"1", "projectcalico.org/namespace":"kuberhealthy", "projectcalico.org/orchestrator":"k8s", "projectcalico.org/serviceaccount":"default", "source":"kuberhealthy"}, ServiceAccount:"default"} already exists in cache - comparing. type=converter.WorkloadEndpointData
2024-03-21 16:16:16.233 [DEBUG][1] workload_endpoint_default.go 56: Using prefix to create a WorkloadEndpoint veth name prefix="cali"
fatal error: concurrent map read and map write

goroutine 230 [running]:
reflect.mapaccess_faststr(0x1999240?, 0x18?, {0xc0165e0f20?, 0x1999240?})
        /usr/local/go/src/runtime/map.go:1343 +0x1e
reflect.Value.MapIndex({0x1a5bca0?, 0xc006cb6620?, 0x4b078a?}, {0x1999240, 0xc00c0abc90, 0x98})
        /usr/local/go/src/reflect/value.go:1664 +0xc5
reflect.deepValueEqual({0x1a5bca0?, 0xc010371ee0?, 0xc0017a1820?}, {0x1a5bca0?, 0xc006cb6620?, 0xc0017a1860?}, 0x80000000000?)
        /usr/local/go/src/reflect/deepequal.go:147 +0x149e
[...]

PR #8706, opened 2024-04-09
calico-kube-controllers fix concurrent map writes issue

zamog:
## Description
Add mutex lock to the Set function to prevent a race condition that causes the process to panic with fatal error: concurrent map read and map write

## Related issues/PRs
fixes [8705](https://github.com/projectcalico/calico/issues/8705)

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/kube-controllers/pkg/cache/cache.go
+++ b/kube-controllers/pkg/cache/cache.go
@@ -241,6 +241,10 @@ func (c *calicoCache) reconcile(reconcilerPeriod string) {
 }
 
 func (c *calicoCache) performDatastoreSync() error {
+	// lock the cache
+	c.mut.Lock()
+	defer c.mut.Unlock()
+
 	// Get all the objects we care about from the datastore using ListFunc.
 	objMap, err := c.ListFunc()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


aaaaaaaalex:
Thanks for the PR.

Based on your issue, I'm not 100% convinced this will fix the race. If the concurrent write is also occurring in the same `Set` method, then maybe, but, It seems from your logs of the issue that:

 - `reflect.DeepEqual()` being called on the cache **item** is happening concurrently to a *write* on that item.

- The item being mutated is a `WorkloadEndpointData`, particularly, it seems like a mutation on it's `labels` map may be the culprit. Therefor I'm guessing the panic arises from a write to this struct's `labels` while the reflect pkg reads them.

Do we know what component is *writing* to that `labels` map while `reflect` is reading it? And can we confidently say that synchronising calls of the `Set` method removes that concurrency?



On a side-note, our process would be to commit this fix to master branch, rather than directly to the release branch, to ensure we don't regress in future releases. We can then backport the master patch to 3.26.

kube-controllers/pkg/cache/cache.go
```diff
@@ -126,7 +126,9 @@ func (c *calicoCache) Set(key string, newObj interface{}) {
 	if reflect.TypeOf(newObj) != c.ObjectType {
 		c.log.Fatalf("Wrong object type received to store in cache. Expected: %s, Found: %s", c.ObjectType, reflect.TypeOf(newObj))
 	}
-
+	// lock the cache
```
caseydavenport, line 129: Something isn't quite adding up to me here - the underlying map is within the `threadSafeCache` instance, which IIUC should already handle this. 

Do you know why we need this second mutex?

matthewdupre:
@zamog I think we'd be open to adding any necessary locking here, but we need to understand how the concurrent writes happen: adding locks without understanding why is not generally safe.  If you can come up with a UT or description of how the concurrent writes happen then we can make a fix.

PR #8707, opened 2024-04-09
[BPF] dual stack for v6 only

tomastigera:
## Description

    [BPF] When IPv6 is enabled, either v4 or v6 connectivity is enough
    
    If IPv6 is enabled (which enables dual stack), it is ok, if only ipv6
    addresses are avilable. Felix reports ready and connectivity exists.
    When IPv4 are assigned later, IPv4 connectivity will work. The same hold
    in reverse order.
    
    worded differently, Felix reports ready if the path for whatever ip
    version of its host IPs is ready. If there is a vX host IP and version
    X path is not loaded, Felix is not ready as there is an error.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/tc_preamble.c
+++ b/felix/bpf-gpl/tc_preamble.c
@@ -25,6 +25,7 @@ SEC("tc")
 int  cali_tc_preamble(struct __sk_buff *skb)
 {
 	volatile struct cali_tc_globals *globals = state_get_globals_tc();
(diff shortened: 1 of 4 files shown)


sridhartigera (approved):
LGTM.

Issue #8709, opened 2024-04-10
calico cni-plugin error: "failed to create host netlink handle: protocol not supported" on calico v3.27+


carloslima:
## Expected Behavior
Pods are created and have working network

## Current Behavior
After an upgrade from calico v3.26.3 to v3.27.0, all new pods are stuck in ContainerCreating state and kubelet reports `network: plugin type="calico" failed (add): failed to create host netlink handle: protocol not supported`

```
> kubectl get pods
NAME                                      READY   STATUS              RESTARTS   AGE
foo                               0/1     ContainerCreating   0          14h


> kubectl describe pod/foo
(...)
Events:
  Type     Reason                  Age                   From     Message
  ----     ------                  ----                  ----     -------
  Warning  FailedCreatePodSandBox  10m (x5431 over 14h)  kubelet  (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "8253479312ec2d1e370a012259d82b5fb216674bf2662a9d1e4bb3b50b9f371a" network for pod "foo": networkPlugin cni failed to set up pod "foo_bar" network: plugin type="calico" failed (add): failed to create host netlink handle: protocol not supported
```

Here's a partial reproduction of stracing the cni-plugin call:
```
(..)
write(9, "2024-04-09 19:50:33.173 [INFO][2115587] dataplane_linux.go 68: Setting the host side veth name to cali0c38c1a29b3 ContainerID=\"aad78641..\" Namespace=\"..\" Pod=\"..\" WorkloadEndpoint=\"..\"\n", 412) = 412
socket(AF_NETLINK, SOCK_RAW|SOCK_CLOEXEC, NETLINK_ROUTE) = 8
bind(8, {sa_family=AF_NETLINK, nl_pid=0, nl_groups=00000000}, 12) = 0
socket(AF_NETLINK, SOCK_RAW|SOCK_CLOEXEC, NETLINK_XFRM) = -1 EPROTONOSUPPORT (Protocol not supported)
futex(0xc000100148, FUTEX_WAKE_PRIVATE, 1) = 1
--- SIGURG {si_signo=SIGURG, si_code=SI_TKILL, si_pid=2115587, si_uid=0} ---
rt_sigreturn({mask=[]})                 = 1
getpid()                                = 2115587
[...]

PR #8710, opened 2024-04-10
Avoid failure to add interface when XFRM is not supported by the kernel

carloslima:
## Description
Calico cni-plugin errors out on kernels that don't support XFRM leaving all new pods stuck in ContainerCreating state and kubelet reporting `plugin type="calico" failed (add): failed to create host netlink handle: protocol not supported`.

This issue was introduced in v3.27.0 by
> commit 71d6f8385 Improve pod startup time

By default, netlink.NewHandle() will return a handle with all netlink families supported by the library (currently, ROUTE, XFRM, NETFILTER). 

This fix limits the requested handle to the ROUTE netlink family only.

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/8709

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
+++ b/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
@@ -29,6 +29,7 @@ import (
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 
 	"github.com/projectcalico/calico/cni-plugin/pkg/types"
 	api "github.com/projectcalico/calico/libcalico-go/lib/apis/v3"
@@ -67,7 +68,7 @@ func (d *linuxDataplane) DoNetworking(
 
 	d.logger.Infof("Setting the host side veth name to %s", hostVethName)
 
-	hostNlHandle, err := netlink.NewHandle()
+	hostNlHandle, err := netlink.NewHandle(unix.NETLINK_ROUTE)
 	if err != nil {
 		return "", "", fmt.Errorf("failed to create host netlink handle: %v", err)
 	}
--- a/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
+++ b/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
@@ -29,7 +29,6 @@ import (
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
-	"golang.org/x/sys/unix"
 
 	"github.com/projectcalico/calico/cni-plugin/pkg/types"
 	api "github.com/projectcalico/calico/libcalico-go/lib/apis/v3"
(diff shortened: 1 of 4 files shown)


cni-plugin/pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -67,7 +68,7 @@ func (d *linuxDataplane) DoNetworking(
 
 	d.logger.Infof("Setting the host side veth name to %s", hostVethName)
 
-	hostNlHandle, err := netlink.NewHandle()
+	hostNlHandle, err := netlink.NewHandle(unix.NETLINK_ROUTE)
```
cyclinder, line 71: Do `unix.NETLINK_ROUTE` is only related to route property?

carloslima: No, not only routing.

From the [netlink man page](https://www.man7.org/linux/man-pages/man7/netlink.7.html):
> NETLINK_ROUTE
>               Receives routing and link updates and may be used to
>               modify the routing tables (both IPv4 and IPv6), IP
>               addresses, link parameters, neighbor setups, queueing
>               disciplines, traffic classes, and packet classifiers (see
>               [rtnetlink(7)](https://www.man7.org/linux/man-pages/man7/rtnetlink.7.html)).

carloslima:
For the record, this change is working as expected in one of our clusters

cni-plugin/pkg/dataplane/linux/dataplane_linux.go
cyclinder: I got it! thanks for your details.

fasaxc:
Thanks for the fix, looks like there's also a bare NewHandle call in felix's vxlan_mgr.go.  Please can you update that too; I think it also needs to be a route handle to avoid this problem (when VXLAN is enabled)?

carloslima:
> Thanks for the fix, looks like there's also a bare NewHandle call in felix's vxlan_mgr.go. Please can you update that too; I think it also needs to be a route handle to avoid this problem (when VXLAN is enabled)?

Sure, will take a look at it shortly :)

carloslima:
@fasaxc I've made the changes but `make test` locally is taking forever to run :sweat_smile:

carloslima:
@fasaxc can I get another test run from semaphoreci? I don't see it showing up here and I'm having troubles running tests locally :)

carloslima:
Outch. Sorry about that. Fixed. Can you run it again? :)

fasaxc (approved):
LGTM, thanks for the contribution!  Would normally ask for regression tests but  I can't think of a good bang-for-buck way to test this in our CI system.  We don't have easy access to a custom kernel with XFRM compiled out.

carloslima:
I thought about testing but I couldn't come up with anything reasonable :sweat_smile: 

Thanks

fasaxc:
@carloslima just for my education, why bother compiling out XFRM?

carloslima:
I don't know the reasoning behind it because the nodes are managed by a different team. My guess is that they compile out a lot of things they don't use nor expect to be used :-)

PR #8654, opened 2024-03-26
Add the command to determine whether the system has netstat

opencmit2:
## Description

Add the command to determine whether the system has netstat，Can automatically select commands on the system for log prompt output.

## Related issues/PRs
fixes  #8653 

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #8727, opened 2024-04-18
Simplify system compatibility check logic and improve error reporting.

opencmit2:
## Description

Simplify system compatibility check logic and improve error reporting.

In the CI validation step, confirm whether spaces and tabs #8654

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/calicoctl/calicoctl/commands/node/checksystem.go
+++ b/calicoctl/calicoctl/commands/node/checksystem.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


opencmit2:
@caseydavenport 
I think this is reasonable. If the kernel version is not suitable, there is no need to check the following kernel modules. If the kernel version is suitable, the following kernel modules will also be checked in sequence.

caseydavenport:
I generally don't want to make changes in behavior unless we can point to a specific problem it is solving. In this case, I am not sure there is one.

PR #8740, opened 2024-04-19
Don't store Labels directly on v1 backing CRDs

caseydavenport:
## Description

This change stops storing projectcalico.org/v3 Labels directly in the
metadata of the underlying crd.projectcalico.org/v1 backing CRDs. 

This is specifically to fix an interaction with ArgoCI where a label
that is used to track which resources are owned by Argo shows up on both
the v1 and v3 objects with different UIDs, causing Argo to continuously
believe that it is out of sync.

Instead, we store the v3 Labels within a special Calico annotation which
will be populate in the v3 resource at read time.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
@fasaxc mind giving this one a look over?

PR #8746, opened 2024-04-22
Revert "Disable VXLAN checksum offload by default."

tomastigera:
The issue should be fixed in recent kernel and we did not manage to reproduce it. If the issue still persists, it is easier to disable it for the user who hits the issue then to reenable it on a live cluster where this is not a problem.

This reverts commit 53f3e1b62f9475ad8a5509923c1bcf98aa1a1bd1.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Important note: Re-enable VXLAN checksum offload for kernels > 5.7 (takes effect when nodes are restarted).  The offload was disabled in the past due to kernel bugs but has substantial performance impact.  We've been unable to find an environment where this needs to be disabled, so are re-enabling the offload with the intent of proving it works or finding a setup where it doesn't.

If you encounter problems after upgrading and restarting nodes, please let us know and set "ChecksumOffloadBroken=true" in the FelixConfiguration's featureDetectOverride field to restore the previous behavior.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/environment/feature_detect_linux.go
+++ b/felix/environment/feature_detect_linux.go
@@ -50,6 +50,8 @@ var (
 	v3Dot10Dot0 = MustParseVersion("3.10.0")
 	// v3Dot14Dot0 added the random-fully feature on the iptables interface.
 	v3Dot14Dot0 = MustParseVersion("3.14.0")
+	// v5Dot7Dot0 contains a fix for checksum offloading.
+	v5Dot7Dot0 = MustParseVersion("5.7.0")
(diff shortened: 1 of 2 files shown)


matthewdupre (approved):

matthewdupre:
We've found a kernel problem here, so will undo this reversion until we put in place a proper fix.

Specifically the checksum offload doesn't work in a certain case involving repeated SNAT.

PR #8780, opened 2024-05-02
Native nftables dataplane

caseydavenport:
## Description

This is still a WIP, but putting it up in order to get initial feedback
and to get some CI runs started.

## Todos

- [ ] UTs for new nftables code
- [ ] FVs for nftables
- [ ] Existing tests passing
- [x] Add FelixConfiguration option to enable nftables
- [ ] Release note

## Release Note

```release-note
Tech-preview support for nftables dataplane and nftables kube-proxy compatibility
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/driver.go
```diff
@@ -207,6 +207,7 @@ func StartDataplaneDriver(configParams *config.Config,
 				NetlinkTimeout:    configParams.NetlinkTimeoutSecs,
 			},
 			RulesConfig: rules.Config{
+				NFTables:              true, // TODO
```
caseydavenport, line 210: Still need to add a FelixConfiguration option and plumb this through.

felix/dataplane/linux/endpoint_mgr.go
```diff
@@ -110,7 +112,7 @@ func (c *endpointManagerCallbacks) InvokeRemoveWorkload(old *proto.WorkloadEndpo
 // endpointManager manages the dataplane resources that belong to each endpoint as well as
 // the "dispatch chains" that fan out packets to the right per-endpoint chain.
 //
-// It programs the relevant iptables chains (via the iptables.Table objects) along with
+// It programs the relevant iptables chains (via the generictables.Table objects) along with
```
caseydavenport, line 115: Need to do a sweep through of comments and variables that refer to iptables explicitly and make them generic. Low priority.

felix/dataplane/linux/endpoint_mgr.go
```diff
@@ -125,14 +127,16 @@ type endpointManager struct {
...
-	filterTable  IptablesTable
+	rawTable     Table
+	mangleTable  Table
+	filterTable  Table
 	ruleRenderer rules.RuleRenderer
 	routeTable   routetable.RouteTableInterface
 	writeProcSys procSysWriter
 	osStat       func(path string) (os.FileInfo, error)
 	epMarkMapper rules.EndpointMarkMapper
+	newMatch     func() generictables.MatchCriteria
```
caseydavenport, line 138: These two fields allow the implementation of endpointManager (and other structs that follow the same pattern) to be agnostic to the underlying match / action implementation.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -423,12 +439,22 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		OnStillAlive:          dp.reportHealth,
 		OpRecorder:            dp.loopSummarizer,
 	}
+	nftablesOptions := nftables.TableOptions{
```
caseydavenport, line 442: Need to evaluate this TableOptions struct to determine which are relevant for nftables. For now, I've mostly just copied the iptables implementation.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -423,12 +439,22 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
+		OnStillAlive:          dp.reportHealth,
+		OpRecorder:            dp.loopSummarizer,
+	}
 
 	if config.BPFEnabled && config.BPFKubeProxyIptablesCleanupEnabled {
 		// If BPF-mode is enabled, clean up kube-proxy's rules too.
-		log.Info("BPF enabled, configuring iptables layer to clean up kube-proxy's rules.")
-		iptablesOptions.ExtraCleanupRegexPattern = rules.KubeProxyInsertRuleRegex
-		iptablesOptions.HistoricChainPrefixes = append(iptablesOptions.HistoricChainPrefixes, rules.KubeProxyChainPrefixes...)
+		if !config.RulesConfig.NFTables {
```
caseydavenport, line 453: Probably need an nftables equivalent here.

felix/nftables/actions_test.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"github.com/projectcalico/calico/felix/environment"
+	"github.com/projectcalico/calico/felix/generictables"
+	. "github.com/projectcalico/calico/felix/nftables"
+
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+var _ = DescribeTable("Actions",
```
caseydavenport, line 26: Need to update all the nftables tests - they are just copies of iptables at the moment.

felix/nftables/ipsets.go
```diff
@@ -0,0 +1,513 @@
...
+	logCxt *log.Entry
+
+	// Optional filter.  When non-nil, only these IP set IDs will be rendered into the dataplane
+	// as Linux IP sets.
+	neededIPSetNames set.Set[string]
+
+	nft knftables.Interface
+}
+
+func NewIPSets(ipVersionConfig *ipsets.IPVersionConfig, nft knftables.Interface) *IPSets {
```
caseydavenport, line 61: This is largely a copy/paste from the existing IPSet dataplane code, but modified to render sets to nftables. Most of member / set tracking logic is exactly the same, and it's just how they are written that differs. Can probably come up with a better way to share code here, but for now this is working.

felix/nftables/match_builder.go
```diff
@@ -0,0 +1,356 @@
...
+func (m nftMatch) InInterface(ifaceMatch string) generictables.MatchCriteria {
+	return append(m, fmt.Sprintf("iifname %s", ifaceMatch))
+}
+
+func (m nftMatch) OutInterface(ifaceMatch string) generictables.MatchCriteria {
+	return append(m, fmt.Sprintf("oifname %s", ifaceMatch))
+}
+
+func (m nftMatch) RPFCheckPassed(acceptLocal bool) generictables.MatchCriteria {
+	// TODO: acceptLocal is not supported in nftables mode.
```
caseydavenport, line 108: As far as I've been able to discover...

e.g., the `iptables-translate` utility doesn't know what to do with it.

felix/nftables/table.go
```diff
@@ -0,0 +1,1293 @@
...
+// rule IDs in place, we can easily distinguish Calico rules from non-Calico rules without needing
+// to know exactly which rules to expect.  To deal with cleanup after upgrade from older versions
+// that did not write rule IDs, we support special-case regexes to detect our old rules.
+//
+// # Thread safety
+//
+// nftablesTable doesn't do any internal synchronization, its methods should only be called from one
+// thread.  To avoid conflicts in the dataplane itself, there should only be one instance of
+// nftablesTable for each table in an application.
+type nftablesTable struct {
```
caseydavenport, line 211: This is also a big copy / paste of the iptables implementation. I have cut out some bits that I think aren't needed / relevant for nftables, and adjusted the logic in a few places but it still needs me to go over it with a fine comb.

felix/rules/endpoints.go
```diff
@@ -385,31 +413,31 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(policyGroups []*PolicyGroup,
 
 	// First set up failsafes.
 	if failsafeChain != "" {
-		rules = append(rules, Rule{
-			Action: JumpAction{Target: failsafeChain},
+		rules = append(rules, generictables.Rule{
+			Match:  r.NewMatch(),
```
caseydavenport, line 417: One of the unfortunate consequences of using an interface type for the match is that leaving the `Match` type `nil` is no longer viable, as it results in nil pointer exceptions when calling bound functions where previously the function would have been called on a `nil` slice.

node/clean-up-filesystem.sh
```diff
@@ -315,9 +321,11 @@ packages_to_remove=$(microdnf repoquery --installed |
 
 
 echo "Removing ${packages_to_remove}"
+
 # Removing one of the packages deletes rc.local, move it out of the way.
 mv /etc/rc.local /etc/rc.local.bak
-rpm -e --nodeps $packages_to_remove
+# rpm -e --nodeps $packages_to_remove
```
caseydavenport, line 327: For now this is a hack because I was struggling to get both iptables and nftables in the same image, and I didn't want to spend too much time on it. Need to revisit.

felix/nftables/table.go
```diff
@@ -0,0 +1,1293 @@
...
+		return nil, nil, err
+	}
+	countNumSaveCalls.Inc()
+
+	hashes = make(map[string][]string)
+	rules = make(map[string][]*knftables.Rule)
+
+	for _, chain := range chains {
+		hashes[chain] = []string{}
+		rulesInChain, err := t.nft.ListRules(context.TODO(), chain)
```
caseydavenport, line 832: Unfortunately making this many nft calls seems to make loading the dataplane even for a pretty small ruleset take >500ms, which doesn't look good for scaling. 

We'll need to update this to be more efficient. We should be able to load the entire table in one API call, but `knftables` doesn't support that (yet, at least) so perhaps they will accept a PR upstream for this one.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -2286,11 +2394,11 @@ func (d *InternalDataplane) loopReportingStatus() {
 	}
 }
 
-// IptablesTable is a shim interface for iptables.Table.
-type IptablesTable interface {
-	UpdateChain(chain *iptables.Chain)
-	UpdateChains([]*iptables.Chain)
-	RemoveChains([]*iptables.Chain)
+// Table is a shim interface for generictables.Table.
+type Table interface {
```
tomastigera, line 2398: nit: What if we move this type now to generictables as say `RuleChains` and embed it in `Table`?

tomastigera (commented):
Would it make sense to rename `generictables` (sounds pretty generic, but it is not so much ;-) ) to say `linuxtables` or `nettables` and place both `iptables` and `nftables` below that? Now we have 3 top level tables packages, which are for linux only and one is just interfaces.

caseydavenport:
> Would it make sense to rename generictables (sounds pretty generic, but it is not so much ;-) ) to say linuxtables or nettables and

@tomastigera yep, I think those are both better!

felix/nftables/table.go
```diff
@@ -0,0 +1,1294 @@
...
+	for chainName, expectedHashes := range t.chainToDataplaneHashes {
+		logCxt := t.logCxt.WithField("chainName", chainName)
+		if t.dirtyChains.Contains(chainName) || t.dirtyBaseChains.Contains(chainName) {
+			// Already an update pending for this chain; no point in flagging it as
+			// out-of-sync.
+			logCxt.Debug("Skipping known-dirty chain")
+			continue
+		}
+		dpHashes := dataplaneHashes[chainName]
+		if !t.ourChainsRegexp.MatchString(chainName) {
```
caseydavenport, line 643: This block is currently not quite behaving as this is written at the moment. Namely, even the base chains are "our chains" in nftables mode, since we own the whole table.

it's still possible that some other actor might add a chain into our table, in which case we should remove it. But the implication that it's a chain we're inserting into isn't right in this mode. I think this logic could be simplified to remove the check for inserted rules, and simply always mark it as dirty.

felix/nftables/table.go
```diff
@@ -0,0 +1,1294 @@
...
+						"Chain had unexpected inserts, marking for resync")
+					t.dirtyBaseChains.Add(chainName)
+				}
+				continue
+			}
+
+			// Re-calculate the expected rule insertions based on the current length
+			// of the chain (since other processes may have inserted/removed rules
+			// from the chain, throwing off the numbers).
+			expectedHashes, _, _ = t.expectedHashesForInsertAppendChain(
```
caseydavenport, line 668: We shouldn't ever hit this block in nftables mode, since the insert / append chains are chains that we own.

felix/nftables/table.go
```diff
@@ -0,0 +1,1294 @@
...
+		}
+		if _, ok := t.chainToDataplaneHashes[chainName]; ok {
+			// Chain expected, we'll have checked its contents above.
+			logCxt.Debug("Skipping expected chain")
+			continue
+		}
+		if !t.ourChainsRegexp.MatchString(chainName) {
+			// Non-calico chain that is not tracked in chainToDataplaneHashes. We
+			// haven't seen the chain before and we haven't been asked to insert
+			// anything into it.  Check that it doesn't have an rule insertions in it
```
caseydavenport, line 708: I think if we just always mark chains that are not ours as dirty, this code block becomes irrelevant as well.

felix/nftables/table.go
```diff
@@ -0,0 +1,1294 @@
...
+
+	// We _think_ we're in sync, check if there are any reasons to think we might
+	// not be in sync.
+	lastReadToNow := now.Sub(t.lastReadTime)
+	if t.refreshInterval > 0 && lastReadToNow > t.refreshInterval {
+		// Too long since we've forced a refresh.
+		t.InvalidateDataplaneCache("refresh timer")
+	}
+
+	// Retry until we succeed.  There are several reasons that updating nftables may fail:
```
caseydavenport, line 897: Comment needs updating.

felix/nftables/match_builder.go
```diff
@@ -0,0 +1,489 @@
...
+	m.clauses = append(m.clauses, fmt.Sprintf("icmpv6 type != %d code != %d", t, c))
+	return m
+}
+
+// VXLANVNI matches on the VNI contained within the VXLAN header.  It assumes that this is indeed a VXLAN
+// packet; i.e. it should be used with a protocol==UDP and port==VXLAN port match.
+//
+// Note: the -m u32 option is not supported on iptables in NFT mode.
+// https://wiki.nftables.org/wiki-nftables/index.php/Supported_features_compared_to_xtables#u32
+func (m nftMatch) VXLANVNI(vni uint32) generictables.MatchCriteria {
```
fasaxc, line 460: We dropped this in iptables mode too, I think? Should we just remove the function entirely?

felix/rules/endpoints.go
fasaxc: Yeah, I was trying to think of a way around that.  Probably not worth it, but you could maybe do it with generics:
```
type Rule[M Match, A Action] {
  Match M
  Action A
}
```
Then the rule renderer could have `[M Match, A Action]` too and pass them through.  Would create a different kind of boilerplate everywhere!  Advantage would be that the compiler would enforce getting it right.

felix/rules/endpoints.go
```diff
@@ -551,12 +565,12 @@ func EndpointChainName(prefix string, ifaceName string) string {
 	return hashutils.GetLengthLimitedID(
 		prefix,
 		ifaceName,
-		MaxChainNameLength,
+		iptables.MaxChainNameLength, // TODO: This is the max chain name length for iptables, but we should use the max for the implementation.
```
fasaxc, line 568: TODO left in

felix/rules/policy.go
```diff
@@ -20,55 +20,59 @@ import (
...
+func (r *DefaultRuleRenderer) ProtoRulesToIptablesRules(protoRules []*proto.Rule, ipVersion uint8, chainComments ...string) []generictables.Rule {
+	var rules []generictables.Rule
 	for _, protoRule := range protoRules {
 		rules = append(rules, r.ProtoRuleToIptablesRules(protoRule, ipVersion)...)
 	}
 	// Strip off any return rules at the end of the chain.  No matter their
 	// match criteria, they're effectively no-ops.
 	for len(rules) > 0 {
-		if _, ok := rules[len(rules)-1].Action.(iptables.ReturnAction); ok {
+		if _, ok := rules[len(rules)-1].Action.(nftables.ReturnAction); ok {
```
fasaxc, line 65: Could use a marker interface for return actions? 
```
type ReturnAction {
    MarkerIsReturnAction()
}
```

felix/fv/bpf_test.go
```diff
@@ -4390,6 +4382,10 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 			})
 
 			It("should have connectivity when DNAT redirects to-host traffic to a local pod.", func() {
+				if NFTMode() {
+					Skip("NFT does not support third-party rules")
```
fasaxc, line 4386: I think this is meant to test what happens if the user creates their own NAT rules, which they could do in NFT mode in a separate table?

felix/nftables/table_test.go
```diff
@@ -0,0 +1,1610 @@
...
+					})
+				})
+			})
+		})
+	})
+})
+
+// })
+//
+// var _ = Describe("Tests of post-update recheck behaviour with refresh timer (nft)", func() {
```
fasaxc, line 755: Commented out...

fasaxc (changes requested):
Overall looks like great work!  I don't think I spotted any major bugs, just a few nits and minor things along with a few places where it might be worth sharing some code between the dataplanes.

The Github UI is really struggling with the size of the PR so my preference would be for you to make the smaller markups in new commits (so I can review as a diff) and then to bank this before doing any further work to share code etc.

A few spec things to tie down:
* Config model; should we make a Dataplane enum and deprecate the old XXXEnabled flags?
* Interleaving rules, I know some customers like to add their own rules in between our "insert" and "append" layers, should we define a way to do that (e.g. can it be done with multiple chains at different priorities or something like that)

felix/ipsets/nft_sets.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fmt"
+
+	"github.com/projectcalico/calico/felix/ip"
+)
+
+// Constants for nftables, which uses alternative set types as well as
+// additional set types that are not supported by iptables.
+const (
+	NFTSetTypeAddr     = "ipv%d_addr"
```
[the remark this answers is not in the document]
caseydavenport: I ended up needing to make new set types for nftables because all of the parsing / canonization of members is keyed off of the type, so a "hash:net" needs different logic than the equivalent nftables set does (as their members are formatted differently).

I agree in general that I'm not really happy with how I've got the sets stuff laid out right now - it feels pretty icky. Perhaps I need to either. The alternative I saw was to make the calc graph aware of iptabes vs. nftables, and that also felt wrong. The way I have be thinking of it is that the "generic" calc graph interface just happens to be the same types / strings as the iptables implementation :sweat_smile:

felix/generictables/match_builder.go
```diff
@@ -0,0 +1,76 @@
...
+	ICMPV6TypeAndCode(t, c uint8) MatchCriteria
+	NotICMPV6TypeAndCode(t, c uint8) MatchCriteria
+	VXLANVNI(vni uint32) MatchCriteria
+}
+
+// NFTMatchCriteria extends the generictables.MatchCriteria interface with nftables-specific methods.
+type NFTMatchCriteria interface {
+	MatchCriteria
+
+	ConntrackStatus(statusNames string) MatchCriteria
```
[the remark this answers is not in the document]
caseydavenport: In iptables mode, I believe these resolve down to the same match, but in nftables mode there are two separate matches - `ct state` and `ct status` for checking different types of conntrack information. `ct state` matches the state of the connection in conntrack (e.g., established, invalid, etc.) whereas `ct status` matches a number of other metadata bits related to the connection - like whether NAT is performed, or flow table offload is being used for this connection.

https://wiki.nftables.org/wiki-nftables/index.php/Matching_connection_tracking_stateful_metainformation#Matching_conntrack_metadata

felix/nftables/ipsets.go
```diff
@@ -0,0 +1,697 @@
...
+	start := time.Now()
+
+	// Create a new transaction to update the IP sets.
+	tx := s.nft.NewTransaction()
+
+	if s.setNameToProgrammedMetadata.Dataplane().Len() == 0 {
+		// Use the total number of IP sets that we believe we have programmed as a proxy for whether
+		// or not the table exists. If this is the first time we've programmed IP sets, make sure we
+		// create the table as part of the transaction as well.
+		tx.Add(&knftables.Table{})
```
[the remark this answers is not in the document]
danwinship: > Presumably this will fail if it does already exist but it was somehow empty?

no, `tx.Add` is a no-op if the object already exists

felix/nftables/ipsets.go
```diff
@@ -0,0 +1,697 @@
...
+			})
+			return nil
+		})
+		if err != nil {
+			return fmt.Errorf("failed to read set memebers: %w", err)
+		}
+
+		// Mark us as having seen the programmed IP set.
+		// TODO: Ideally we'd extract this information from the data plane itself, but it's not exposed
+		// via knftables at the moment.
```
danwinship, line 415: hm, what does that mean?

felix/nftables/match_builder.go
```diff
@@ -0,0 +1,489 @@
...
+}
+
+func (m nftMatch) NotSourceIPSet(name string) generictables.MatchCriteria {
+	m.clauses = append(m.clauses, fmt.Sprintf("IPV saddr != @%s", LegalizeSetName(name)))
+	return m
+}
+
+func (m nftMatch) SourceIPPortSet(name string) generictables.MatchCriteria {
+	// IPPort sets include the IP, protocol, and port, in that order.
+	// Note that "th dport" is only compatible with protocols that have their destination port in
```
danwinship, line 313:
```suggestion
	// Note that "th sport" is only compatible with protocols that have their destination port in
```

and in the next function. (The code is right, it's just the comment that's wrong.)

felix/nftables/match_builder.go
```diff
@@ -0,0 +1,489 @@
...
+func (m nftMatch) VXLANVNI(vni uint32) generictables.MatchCriteria {
+	// TODO: Not supported in nftables mode.
+	return m
+}
+
+// Converts a list of ports to a multiport set suitable for inline use in nftables rules.
+func PortsToMultiport(ports []uint16) string {
+	if len(ports) == 1 {
+		return fmt.Sprintf("%d", ports[0])
+	}
```
danwinship, line 469: fwiw, if you write `tcp sport { 80 }`, nft will optimize it to `tcp sport 80` anyway, so you could remove the optimization here

felix/nftables/table.go
```diff
@@ -0,0 +1,1173 @@
...
+	ipVersion uint8
+	nft       knftables.Interface
+
+	// For rendering rules and chains.
+	render generictables.NFTRenderer
+
+	// featureDetector detects the features of the dataplane.
+	featureDetector environment.FeatureDetectorIface
+
+	// TODO: There are no kernel chains in nftables. We should simplify this for nftables.
```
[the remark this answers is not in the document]
danwinship: > With no kernel chains to share, is there a way for us to nicely interleave with user-provided rules; for example, our inserted and appended rules could map to two different priorities?

"well-defined priorities" is what we were planning for kube-proxy. (btw, @caseydavenport did you end up needing to depend on any assumptions about how kube-proxy behaves anywhere? I'm on the hook to document that for 1.33...)

felix/nftables/table.go
```diff
@@ -0,0 +1,1173 @@
...
+		t.gaugeNumRules.Sub(float64(len(oldChain.Rules)))
+		t.maybeDecrefReferredChains(name, oldChain.Rules)
+		delete(t.chainNameToChain, name)
+		if t.chainIsReferenced(name) {
+			t.dirtyChains.Add(name)
+
+			// Defensive: make sure we re-read the dataplane state before we make updates.  While the
+			// code was originally designed not to need this, we found that other users of
+			// nftables can still clobber out updates so it's safest to re-read the state before
+			// each write.
```
danwinship, line 466: comment feels like it was just `s/iptables/nftables/` and may not actually be true

felix/nftables/table.go
```diff
@@ -0,0 +1,1173 @@
...
+		}
+		return nil, nil, err
+	}
+	for _, chain := range allChains {
+		hashes[chain] = []string{}
+		rules[chain] = []*knftables.Rule{}
+	}
+
+	// List rules and extract the hashes.
+	allRules, err := t.nft.ListRules(ctx, "")
```
danwinship, line 714: `""` only works by coincidence here; you should submit a knftables patch to document/test that (and update `Fake` to support it).

(I eventually want to totally redo the List APIs, but I'm not sure how yet, so it's probably not worth trying to make the API any nicer to support this case yet.)

felix/nftables/table.go
```diff
@@ -0,0 +1,1173 @@
...
+			// but might appear if someone else has modified our table.
+			hashes[rule.Chain] = append(hashes[rule.Chain], "")
+		}
+	}
+	return
+}
+
+// isNotFound helps distinguish errors resulting from the table not yet being programmed.
+func isNotFound(err error) bool {
+	return strings.Contains(err.Error(), "no such table") ||
```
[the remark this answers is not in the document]
danwinship: this already exists upstream; `knftables.IsNotFound(err)`

danwinship (commented):
(quick skim of just the nftables-specific bits...)

felix/nftables/table.go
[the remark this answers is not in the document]
caseydavenport: I haven't thought much about it - I suspect different priorities would do the trick here.

For "deny" rules, users can just put them after all of our rules since later tables still get a chance to deny traffic even if our table accepted it. But yeah, for other types of interaction might want separate priorities. Will raise a task to track this.

felix/ip/trie.go
```diff
@@ -314,6 +325,33 @@ func (t *CIDRTrie) Visit(f func(cidr CIDR, data interface{}) bool) {
 	t.root.visit(f)
 }
 
+// Descendants returns a list of CIDRs representing the closest descentents of the given CIDR in the trie that
```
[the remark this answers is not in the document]
caseydavenport: Yeah, it's definitely a bit purpose built for the use-case of finding the next largest CIDR when masking / unmasking. The rename / example sounds good.

felix/ip/trie.go
```diff
@@ -314,6 +325,33 @@ func (t *CIDRTrie) Visit(f func(cidr CIDR, data interface{}) bool) {
...
+// have data associated with them.
+func (t *CIDRTrie) Descendants(parent CIDR) []CIDR {
+	// Get the node representing this CIDR. We include intermediate nodes so that
```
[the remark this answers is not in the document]
caseydavenport: Hopefully new wording is better. Meant to suggest that this call cares about nodes without any data associated with them.

go.mod
```diff
@@ -331,4 +332,7 @@ replace (
 	k8s.io/mount-utils => k8s.io/mount-utils v0.28.9
 	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.28.9
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.28.9
+
+	// Replace with forked knftables.
+	sigs.k8s.io/knftables => github.com/caseydavenport/knftables v0.0.0-20240510175904-fc87cd388608
```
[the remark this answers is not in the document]
caseydavenport: Oh yeah, I forgot that I had changes in flight here. 

My fork currently only has changes to the test structures - I've already upstreamed the actual "product" code PRs.

Once this goes in I can switch back to upstream: https://github.com/kubernetes-sigs/knftables/pull/12

felix/nftables/ipsets.go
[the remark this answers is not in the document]
caseydavenport: Yeah, this will succeed even if the table exists. But I wanted this function to do nothing if nothing needed to be done.

I suppose I could instead always add the table, and then only execute the transaction if NumOperations is >1 instead of >0.

felix/nftables/ipsets.go
```diff
@@ -0,0 +1,697 @@
...
+
+	for attempt := 0; attempt < 10; attempt++ {
+		if attempt > 0 {
+			s.logCxt.Info("Retrying after an ipsets update failure...")
+		}
+		if s.resyncRequired {
+			// Compare our in-memory state against the dataplane and queue up
+			// modifications to fix any inconsistencies.
+			s.logCxt.Debug("Resyncing ipsets with dataplane.")
+			// s.opReporter.RecordOperation(fmt.Sprint("resync-ipsets-v", s.IPVersionConfig.Family.Version()))
```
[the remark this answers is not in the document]
caseydavenport: Fixed the op recording, will do the metrics when I come back to prometheus metrics for nftables as a whole - I suspect we'll want to introduce new nftables specific metrics for these.

felix/nftables/ipsets.go
```diff
@@ -0,0 +1,697 @@
...
+	setNameToAllMetadata map[string]ipsets.IPSetMetadata
+
+	// setNameToProgrammedMetadata tracks the IP sets that we want to program and
+	// those that are actually in the dataplane.  It's Desired() map is the
+	// subset of setNameToAllMetadata that matches the neededIPSetNames filter.
+	// Its Dataplane() map contains all IP sets matching the IPVersionConfig
+	// that we think are in the dataplane.  This includes any temporary IP
+	// sets and IP sets that we discovered on a resync (neither of which will
+	// have entries in the Desired() map).
+	setNameToProgrammedMetadata *deltatracker.DeltaTracker[string, ipsets.IPSetMetadata]
```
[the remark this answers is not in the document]
caseydavenport: Yep, I agree. I think that will be best saved for a follow-on PR though.

felix/nftables/table.go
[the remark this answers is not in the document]
caseydavenport: @danwinship IIRC, I wasn't able to use `knftables.IsNotFound` here because the Fake implementation doesn't use it - I had a mental note to fix that but I think it got lost :sweat_smile:

felix/nftables/table.go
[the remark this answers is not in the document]
caseydavenport: @danwinship regarding your comment (which magically appeared after my last comment) - I haven't needed to do anything special to interact with kube-proxy so far, which is great. Simply using the default / recommended priorities for each hook has worked well.

felix/nftables/table.go
caseydavenport: It's not _quite_ coincidence :grin: 

https://github.com/kubernetes-sigs/knftables/pull/7

caseydavenport: I missed the fake support in the first PR, though: https://github.com/kubernetes-sigs/knftables/pull/12

felix/nftables/ipsets.go
caseydavenport: The in-memory metadata we keep for each set includes some things like the max size of the set. This function is attempting to synchronize our in-memory view of the set with what actually is programmed, but it's just copying the "desired" Size / Range metadata instead of reading it from the programmed set. I couldn't easily find a way to get this information from knft (can just get set names, and elements). 

I think it's OK and actually not especially relevant, considering right now the nftables code doesn't actually use any of these pieces of metadata when programming the set.

felix/nftables/ipsets_test.go
```diff
@@ -0,0 +1,46 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nftables_test
+
+// import (
```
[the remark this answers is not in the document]
caseydavenport: Ah yep, I missed filling out this UT suite somehow.

felix/nftables/match_builder.go
```diff
@@ -0,0 +1,489 @@
...
+	return m
+}
+
+func (m nftMatch) RPFCheckPassed(acceptLocal bool) generictables.MatchCriteria {
+	// TODO: acceptLocal is not supported in nftables mode.
+	m.clauses = append(m.clauses, "fib saddr . mark . iif oif != 0")
+	return m
+}
+
+func (m nftMatch) RPFCheckFailed(acceptLocal bool) generictables.MatchCriteria {
```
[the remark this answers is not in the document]
caseydavenport: Actually, looks like our code always sets this to false anyway: https://github.com/projectcalico/calico/blob/616bad52431ff8aa4e2d6cfe9ec1a471045fad3e/felix/rules/static.go#L1290-L1294

So perhaps best to just rip it out?

felix/ip/trie.go
```diff
@@ -325,31 +325,56 @@ func (t *CIDRTrie) Visit(f func(cidr CIDR, data interface{}) bool) {
...
-func (t *CIDRTrie) Descendants(parent CIDR) []CIDR {
-	// Get the node representing this CIDR. We include intermediate nodes so that
-	// we can traverse them to find the closest descendants.
+// ClosestDescendants returns a list of CIDRs representing the closest descendants of the given CIDR in the trie that
+// have data associated with them. It does not return a full list of all descendants - only any descendants that are tied for
+// being the closest to the given CIDR.
+//
+// For example, given the following trie, where * indicates an intermediate node with no data associated with it:
+//
+//	  		  10.0.0.0/16
```
caseydavenport, line 334: This looked a lot better in my editor than it does here :sweat_smile:

fasaxc: Looks like it's got mixed tabs and spaces

felix/ip/trie.go
```diff
@@ -325,31 +325,56 @@ func (t *CIDRTrie) Visit(f func(cidr CIDR, data interface{}) bool) {
...
+func (t *CIDRTrie) ClosestDescendants(buf []CIDR, parent CIDR) []CIDR {
+	// Get the node representing this CIDR. Include the node even if it has no associated data, so that
+	// we can traverse it to find the closest descendants.
 	node := t.root.getNode(parent, true)
 	if node == nil {
 		return nil
 	}
-	children := make([]CIDR, 0, 2)
+	if buf == nil {
+		buf = make([]CIDR, 0)
```
fasaxc, line 359: Any reason to actually make a slice instead of just leaving it nil?  I'd say it's the norm in Go to return nil rather than allocate a slice if there's nothing to return.

felix/labelindex/named_port_index.go
```diff
@@ -316,23 +316,27 @@ type SelectorAndNamedPortIndex struct {
 	lastLiveReport time.Time
 }
 
-func NewSelectorAndNamedPortIndex() *SelectorAndNamedPortIndex {
+func NewSelectorAndNamedPortIndex(deduplicate bool) *SelectorAndNamedPortIndex {
```
fasaxc, line 319: I might call this `suppressOverlaps` just to avoid confusion with deduplicating _exact_ duplicate entries, which it already does (IIRC)

felix/nftables/table.go
danwinship: doh, forgot about that PR and then checked my current checked-out copy of that code, which predates your PR, and I saw that it didn't look like it would work and wasn't documented as working. :slightly_smiling_face:

felix/generictables/nft_renderer.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if rule.Match != nil {
+		matchFragment := rule.Match.(NFTMatchCriteria).IPVersion(r.ipv).Render()
+		if matchFragment != "" {
+			fragments = append(fragments, matchFragment)
+		}
+	}
+
+	if rule.Action != nil {
+		// Include a counter action on all rules.
```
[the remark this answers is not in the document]
caseydavenport: Yeah, we probably do. For now, having a counter on each rule has helped me to diagnose subtle bugs since it's a bit easier to track exactly where packets are going.

caseydavenport:
> Config model; should we make a Dataplane enum and deprecate the old XXXEnabled flags?
> Interleaving rules, I know some customers like to add their own rules in between our "insert" and "append" layers, should we define a way to do that (e.g. can it be done with multiple chains at different priorities or something like that)

I think I'm in favor of both of these, but they should probably each be separate PRs

felix/nftables/table.go
```diff
@@ -0,0 +1,1173 @@
...
+	contextTimeout time.Duration
+}
+
+type TableOptions struct {
+	// NewDataplane is an optional function to override the creation of the knftables client,
+	// used for testing.
+	NewDataplane func(knftables.Family, string) (knftables.Interface, error)
+
+	RefreshInterval   time.Duration
+	PostWriteInterval time.Duration
```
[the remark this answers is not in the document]
caseydavenport: I ended up ripping out the postWrite logic from nftables since it seems pretty unnecessary here. Can always revisit adding it via a common implementation if we need it!

felix/nftables/table.go
```diff
@@ -0,0 +1,1173 @@
...
+
+	// Make a copy of our full rules map and keep track of all changes made while processing dirtyBaseChains.
+	// When we've successfully updated nftables, we'll update our cache of chainToFullRules with this map.
+	newChainToFullRules := map[string][]*knftables.Rule{}
+	for chain, rules := range t.chainToFullRules {
+		newChainToFullRules[chain] = make([]*knftables.Rule, len(rules))
+		copy(newChainToFullRules[chain], rules)
+	}
+
+	// TODO: Should these be treated any differently in nftables mode? They are owned by us.
```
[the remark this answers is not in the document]
caseydavenport: Raised a backlog item for this one

felix/generictables/nft_renderer.go
```diff
@@ -0,0 +1,117 @@
...
+func (r *nftRenderer) comment(hash string, rule Rule) *string {
+	fragments := []string{}
+
+	if r.hashCommentPrefix != "" && hash != "" {
+		// Include the rule hash in the comment.
+		fragments = append(fragments, fmt.Sprintf(`%s%s;`, r.hashCommentPrefix, hash))
+	}
+
+	// Add in any comments. Each individual comment must be small enough such that the
+	// total comment is less than 128 characters.
```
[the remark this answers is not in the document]
caseydavenport: Calico's own use of comments shouldn't ever be close to this limit, but we do support user-defined metadata attached to rules which gets stored in the comments and could end up being big.

felix/nftables/table.go
```diff
@@ -0,0 +1,1173 @@
...
+}
+
+func (t *nftablesTable) chainIsReferenced(name string) bool {
+	return t.chainRefCounts[name] > 0
+}
+
+// maybeIncrefReferredChains checks whether the named chain is referenced;
+// if so, it increfs all child chains.  If a child chain becomes newly
+// referenced, its children are increffed recursively.
+func (t *nftablesTable) maybeIncrefReferredChains(chainName string, rules []generictables.Rule) {
```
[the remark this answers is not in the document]
caseydavenport: Have a backlog task for this one as well

felix/ipsets/nft_sets.go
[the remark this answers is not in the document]
fasaxc: The factoring that occurs to me is to change this method
```
func (t IPSetType) CanonicaliseMember(member string) IPSetMember
```
into function (or a method of a dataplane-specific object):
```
func CanonicaliseMember(t IPSetType, member string) IPSetMember
```
and then have two copies, one for iptables and one for nftables?

felix/nftables/ipsets.go
fasaxc: Does nftables limit the size of sets by default? There's an option to set a limit, but perhaps the default is "no limit", which would be good for our use case.  It's an annoying footgun that iptables ipsets have a hard limit on size.

.semaphore/semaphore.yml.d/blocks/20-felix.yml
```diff
@@ -197,6 +197,16 @@
           - make check-wireguard
           - ../.semaphore/run-and-monitor fv-${SEMAPHORE_JOB_INDEX}.log make fv-no-prereqs FV_BATCHES_TO_RUN="${SEMAPHORE_JOB_INDEX}" FV_NUM_BATCHES=${SEMAPHORE_JOB_COUNT}
         parallelism: 3
+      - name: nftables FV Test matrix
+        execution_time_limit:
+          minutes: 120
+        env_vars:
+        - name: FELIX_FV_NFTABLES
+          value: "true"
```
fasaxc, line 205: Should this be `Enabled` now?

Also noticed that Semaphore test results only seemed to have one set of results from the FVs.

What about BPF mode on an nftables system?  Do we need some tests for that?

felix/nftables/ipsets.go
caseydavenport: To my knowledge, no it doesn't enforce a default limit.

fasaxc (approved):

fasaxc (approved):
:shipit:

Issue #8620, opened 2024-03-15
Suggestion for Calico eBPF: Introduction of Direct Routing Mode


pf2rich:
In our current use of Calico's eBPF mode, we have noticed a significant drop in maximum bandwidth—over 50% less compared to when in IPVS mode—for traffic entering the Kubernetes cluster from outside. This reduction occurs as traffic moves via VXLAN from the ingress node to the workload node. We find this decrease in performance to be unreasonable. Is there any plan in the community to implement a Direct Routing (DR) mode, similar to Cilium's DR mode, to route traffic directly to the target pod instead of encapsulating packets in VXLAN?

tomastigera:
@pf2rich I am not familiar with cilium's direct routing. Would you have any pointers or some explanation of it?

50% reduction is quite a bit and certainly worth investigation? Do you observe some increased load on only certain (one) CPU on the communicating nodes? I would be searching for some perf degradation due to locality of the vxlan nodeport traffic.

pf2rich:
@tomastigera 
Based on my understanding of Cilium, when accessing services through NodePort, as traffic reaches node A, Cilium modifies the packet's destination IP address/port to the Pod's IP address/port, and records the original IP and port information in the IP option field. Then, it is routed to the Pod through layer 3 routing. The packets returned by the Pod are subject to SNAT operations through eBPF.

tomastigera:
@pf2rich do you have bonded NICs on your nodes by any chance?

pf2rich:
Yes, we have configured the two physical network cards as bond4.

tomastigera:
It seems that the 50% drop is due to the fact that only one of the devices is used. And since since you saturate the link in/out you can only use half the bandwidth. That should be the same for ebpf or not. However, with bond, you _may_ and possibly _should_ be able to use half the bandwidth of the bond.

(1) in ebpf mode you would have by default the ebpf program on each of the phys devices. It is arguable whether that is a good thing and whether it should be on the bond itself. If you use strict rpf, it must be on the bond.
(2) even when the program is hooked on the bond, it seems that only one of the phys devices is used for both in/out traffic
(3) iptables/ipvs do not have that problem with calico afaict as routing traffic back (forwarded to another node) is done much later
(4) the vxlan tunnel itself does not seem to be the issue for the 50% drop
(5) If vxlan overlay is used for pods (which it often is) forwarding external traffic to a service via the vxlan tunnel is actually faster than using the overlay.

[...]

PR #8790, opened 2024-05-06
[BPF] Variable source port for VXLAN NodePort tunnel

tomastigera:
To efficiently use bonded network cards or multiple CPUs on the receiving node, it is wrong to use a single VXLAN stream for forwarding services (node ports) from one node to a backend on another node. We need to preserve the notion of separate flows by varying the source port for each of the flowsi - as recommended by the RFC. We do so be XORing the source ports of the original stream to create the source port for the vxlan stream that carries it.

refs https://github.com/projectcalico/calico/issues/8620

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -1053,12 +1053,22 @@ nat_encap:
 		}
 	}
 
-	if (vxlan_encap(ctx, &STATE->ip_src, &STATE->ip_dst)) {
+	/* Trivial hash to use multiple vxlan flows. Note that any value will do
+	 * as long as it is a constatnt for this direction of the flow. Does not
+	 * even need to be the same for both directions.
+	 *
+	 * ICMP does not have ports, but there is little to no worries about a
+	 * possible out-of-order processing in relation to the related flow.
+	 */
+	__u16 vxlan_src_port = STATE->sport ^ STATE->dport;
+
+	if (vxlan_encap(ctx, &STATE->ip_src, &STATE->ip_dst, vxlan_src_port)) {
 		deny_reason(ctx, CALI_REASON_ENCAP_FAIL);
 		goto  deny;
 	}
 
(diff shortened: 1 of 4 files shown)


sridhartigera (approved):
LGTM

felix/bpf-gpl/tc.c
```diff
@@ -1053,12 +1053,22 @@ static CALI_BPF_INLINE enum do_nat_res do_nat(struct cali_tc_ctx *ctx,
 		}
 	}
 
-	if (vxlan_encap(ctx, &STATE->ip_src, &STATE->ip_dst)) {
+	/* Trivial hash to use multiple vxlan flows. Note that any value will do
+	 * as long as it is a constatnt for this direction of the flow. Does not
```
fasaxc, line 1057:
```suggestion
	 * as long as it is a constant for this direction of the flow. Does not
```

felix/bpf/ut/nat_encap_test.go
```diff
@@ -106,7 +106,6 @@ func checkVxlan(pktR gopacket.Packet) gopacket.Packet {
 	udpL := pktR.Layer(layers.LayerTypeUDP)
 	Expect(udpL).NotTo(BeNil())
 	udpR := udpL.(*layers.UDP)
-	Expect(udpR.SrcPort).To(Equal(layers.UDPPort(testVxlanPort)))
```
fasaxc, line 109: Would be nice to have some check; e.g. could check that two packets with same source port hash to same but different source ports hash differently?

fasaxc (changes requested):
One typo and one nice-to-have on the testing side

tomastigera:
Flake addressed by this PR https://github.com/projectcalico/calico/pull/8795

PR #8795, opened 2024-05-07
Address some bpf fv flakes

tomastigera:
## Description

    [BPF] ignore map entries that have non-existent values
    
    Fixes this flake in FVs:
    
    Unexpected error:
        <*fmt.wrapError | 0xc00274d380>:
        cannot parse json output: json: cannot unmarshal object into Go struct field mapEntry.value of type []string
        [{
                "key": ["0x01","0x00","0x00","0x00","0x00","0x00","0x00","0x00"
                ],
                "value": {
                    "error": "No such file or directory"
                }
            },{
                "key": ["0x02","0x00","0x00","0x00","0x00","0x00","0x00","0x00"
                ],
                "value": ["0x0a","0x41","0x00","0x02","0x77","0x1f","0x00","0x00"
                ]
            },{
                "key": ["0x00","0x00","0x00","0x00","0x00","0x00","0x00","0x00"
                ],
                "value": ["0xac","0x11","0x00","0x03","0x2b","0x19","0x00","0x00"
                ]
            }
        ]

    Increase fv timeout for felix to become healthy. Takes about 6s

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/bpf.go
+++ b/felix/bpf/bpf.go
@@ -384,7 +384,70 @@ type getnextEntry struct {
 type mapEntry struct {
 	Key   []string `json:"key"`
 	Value []string `json:"value"`
-	Err   string   `json:"error"`
+}
+
+func (me *mapEntry) UnmarshalJSON(data []byte) error {
+	type entry struct {
+		Key   []string `json:"key"`
+		Value any      `json:"value"`
+	}
+
+	if string(data) == "null" {
+		return nil
+	}
+
+	var e entry
+	err := json.Unmarshal(data, &e)
+	if err != nil {
+		// bad json
+		return err
+	}
+
+	v, ok := e.Value.([]any)
+	if !ok {
+		// the value is not what it should be, likely an error like the entry is
+		// now missing (race) so we just ignore it. It is still a valid json.
+		// Return an empty entry which we will filter out.
+		return nil
+	}
+
+	hexbytes := make([]string, len(v))
+	for i, x := range v {
+		hexbytes[i] = x.(string)
+	}
+
+	*me = mapEntry{
+		Key:   e.Key,
+		Value: hexbytes,
+	}
+
+	return nil
+}
+
+type hexMap []mapEntry
+
+func (hm *hexMap) UnmarshalJSON(data []byte) error {
+	if string(data) == "null" {
+		return nil
+	}
+
+	var m []mapEntry
+	err := json.Unmarshal(data, &m)
+	if err != nil {
+		return err
+	}
+
+	var res hexMap
+
+	for _, v := range m {
+		if len(v.Key) != 0 {
+			res = append(res, v)
+		}
+	}
+
+	*hm = res
+
+	return nil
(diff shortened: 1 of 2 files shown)


marvin-tigera:

felix/bpf/bpf.go
```diff
@@ -387,6 +387,69 @@ type mapEntry struct {
...
+	}
+
+	if string(data) == "null" {
+		return nil
+	}
+
+	var e entry
+	err := json.Unmarshal(data, &e)
+	if err != nil {
+		return nil
```
fasaxc, line 404: Why suppress this one?

felix/bpf/bpf.go
```diff
@@ -387,6 +387,69 @@ type mapEntry struct {
 	Err   string   `json:"error"`
 }
 
+func (me *mapEntry) UnmarshalJSON(data []byte) error {
+	type entry struct {
+		Key   []string    `json:"key"`
+		Value interface{} `json:"value"`
+		Err   string      `json:"error"`
```
fasaxc, line 394: I don't see how Err can get set, looks like value would be an object with an error inside it?

Parsing mixed type json is a pain but I might do:
```
var untyped struct {
  Key []string
  Value any
}

err := json.Unmarshal(data, &untyped)
if err != nil {
  ...
}

switch typed := untyped.Value.(type) {
  case map[string]any:
    // An error?
    err = fmt.Errorf("value error: %v", typed["error"])
  case []any:
    // A value
  ...
}
```

felix/bpf/bpf.go
```diff
@@ -387,6 +387,69 @@ type mapEntry struct {
...
+
+func (hm *hexMap) UnmarshalJSON(data []byte) error {
+	if string(data) == "null" {
+		return nil
+	}
+
+	var m []mapEntry
+	err := json.Unmarshal(data, &m)
+	if err != nil {
+		return nil
```
fasaxc, line 437: Why suppress the error?

fasaxc (changes requested):

felix/bpf/bpf.go
tomastigera: we cannot skip over some entries, so a poor mans solution saying, the entry does not look like we expect so just return an empty entry which we later on filter out. But I guess I will go with proper parsing as you suggest above.

felix/bpf/bpf.go
tomastigera:
```
  case map[string]any:
    // An error?
    err = fmt.Errorf("value error: %v", typed["error"])
```
I think we should not return an error when we see an error on an entry.

> I don't see how Err can get set

The output might have changed and is not longer settable. Apparently an error is delivered in a map[string]string as "error" now

marvin-tigera:

fasaxc (approved):

PR #8837, opened 2024-05-17
Add option to disable announcement of aggregated LoadBalancerIP ranges

Lucaber:
## Description

This MR allows to disable the default announcement of all LoadBalancerIP ranges by setting `serviceLoadBalancerRouteAggregationEnabled: false` in the BGPConfiguration.

We are advertising IPs of the same subnets (eg /24) in multiple clusters. LoadBalancer services with externalTrafficPolicy Local are reachable because additional /32 route are being announced only from the cluster where the LoadBalancer exists. But LoadBalancers with externalTrafficPolicy Cluster are not working since the /24 route is being announced from multiple clusters and traffic not reaching the correct cluster.
To mitigate this we would have to dynamically add every ip as a /32 to the BGPConfiguration because /32 routes are not advertised by default.

This MR adds an additional option to the BGPConfiguration to disable the default announcement of all prefixes in the `serviceLoadBalancerIPs` list. In which case we are also announcing /32 routes for LoadBalancers with externalTrafficPolicy Cluster. The `serviceLoadBalancerIPs` list still acts as a filter.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
@Lucaber thanks for the PR! I'm mulling this one over. 

In the meantime, I wonder if you have tried using the BGPFilter resource to achieve the same result? 

https://docs.tigera.io/calico/latest/reference/resources/bgpfilter

I believe it would allow you to configure export filters for Calico that block the advertisement of the aggregated ranges (albeit in a more verbose way).

Not necessarily against something like this PR, but ideally would be good to avoid implementing two ways to achieve the same end result. Let me know what you think.

Lucaber:
Hi @caseydavenport ,
thanks for the suggestion, i didn't know about the BGPFilter resource.
Blocking the announcement via a BGPFilter would only partially solve the problem.

First, calico creates `cali-cidr-block` block rules for all loadbalancerservice cidrs which i'm now preventing in this
[commit](https://github.com/projectcalico/calico/pull/8837/commits/877d9719d5f31c1dada88c907635b91e9608ba39). So traffic from inside the cluster would be blocked by these iptables rules.
Btw: Shouldn't we also block the creation of these rules for `/32` cidrs, as `/32` cidrs are only being announced when a service exists (see [here](https://github.com/projectcalico/calico/blob/ea0608d1dc2fba96261eb57fba4547fb6daa5490/confd/pkg/backends/calico/client.go#L1433))

Secondly, with the BGPFilter, the routes would probably still be created locally on all nodes, so traffic originating from inside the cluster would not be routed via the default route.

caseydavenport:
> Secondly, with the BGPFilter, the routes would probably still be created locally on all nodes

Calico shouldn't actually program the Service LoadBalancer IP addresses / CIDRs into the routing table of the node, so I don't think they would impact routing on the cluster nodes - unless I'm misunderstanding something :sweat_smile: 

> First, calico creates cali-cidr-block block rules for all loadbalancerservice cidrs which i'm now preventing in this
[commit](https://github.com/projectcalico/calico/pull/8837/commits/877d9719d5f31c1dada88c907635b91e9608ba39). So traffic from inside the cluster would be blocked by these iptables rules.

Yep, this is fair and something I hadn't spotted. So, IIUC, we need to:

- Prevent export of the full CIDR (can be achieved with BGPFilter)
- Remove the service loop prevention rules that drop packets to LoadBalancer IPs (configurable via the FelixConfiguration ServiceLookPrevention option)

So I believe that this is already achievable today, albeit with two pieces of configuration. 

> Btw: Shouldn't we also block the creation of these rules for /32 cidrs, as /32 cidrs are only being announced when a service exists (see [here](https://github.com/projectcalico/calico/blob/ea0608d1dc2fba96261eb57fba4547fb6daa5490/confd/pkg/backends/calico/client.go#L1433))

I think there is still a theoretical case where traffic ends up on a node that isn't hosting that service, and will be routed using the host's default gateway which isn't necessarily a loop but we probably still want to prevent. This might just be in theory though, and I agree for the most part we can expect /32 IP address to be real services that kube-proxy will be handling (at least on nodes that are hosting that service).

Lucaber:
Yes you are right but I forgot a third change that is implemented by this PR.
We need to export /32 routes for LoadBalancers with `externalTrafficPolicy: cluster`. These are normally handled by the full cidr route on every node.

caseydavenport:
Sorry for the delay on this.. Right, yes that makes sense to me. So there are a collection of three changes - two of which are expressible using existing config options, but one of them isn't:

- Calico needs to advertise /32 routes for services with traffic policy "Cluster"
- Calico needs to stop advertising the full LoadBalancer IP blocks
- Calico needs to stop programming service loop prevention rules.

> So traffic from inside the cluster would be blocked by these iptables rules.

Perhaps I have forgotten the context here... why do pods within the cluster want to talk via the LoadBalancer service IP? Just trying to remember / understand why we need to stop programming the service loop prevention rules.

caseydavenport:
Unfortunately this one seems to have rotted a bit (largely due to my pickiness trying to determine the minimal set of changes we need here). 

I'm going to close this for now, but happy to pick it back up again if there is interest to keep working on it.

PR #8876, opened 2024-06-03
Support for service loop prevention - ebpf

sridhartigera:
## Description

This PR adds support for service-loop-prevention in ebpf mode.

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: support for service loop prevention
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/nat_lookup.h
```diff
@@ -58,9 +58,18 @@ static CALI_BPF_INLINE struct calico_nat_dest* calico_nat_lookup(ipv46_addr_t *i
...
 	if (!nat_lv1_val) {
-		struct cali_rt *rt;
+		struct cali_rt *rt = cali_rt_lookup(ip_dst);
 
 		CALI_DEBUG("NAT: Miss.\n");
+
+		/* If the traffic is from outside the cluster, to a blocked cidr, drop the packet to
+		 * prevent the packet from looping.
+		 */
+		if (CALI_F_FROM_HEP && !from_tun && rt && cali_rt_is_blackhole(rt)) {
```
tomastigera, line 68: > If the traffic is from outside the cluster

This `CALI_F_FROM_HEP && !from_tun` does not mean that the traffic from outside the cluster. It only means it is not from this node and did not come through the tunnel. In fact, traffic coming through the tunnel is most likely from outside the cluster. To make sure that traffic is from outside the cluster, you need to check `ip_src` for route miss.

This branch in tc.c handle the case where we had NAT miss. I wonder if it would make more sense to do the blackhole lookup there or somewhere we already do a route lookup so that we do not need to do it 2x https://github.com/projectcalico/calico/blob/master/felix/bpf-gpl/tc.c#L424

tomastigera: Perhaps here, because we already do the dst lookup for anyway. https://github.com/projectcalico/calico/blob/master/felix/bpf-gpl/tc.c#L545

felix/bpf-gpl/globals.h
```diff
@@ -52,6 +52,8 @@ enum cali_globals_flags {
 	CALI_GLOBALS_RESERVED7			= 0x00000040,
 	CALI_GLOBALS_NO_DSR_CIDRS		= 0x00000080,
 	CALI_GLOBALS_LO_UDP_ONLY		= 0x00000100,
+	CALI_GLOBALS_SVC_LOOP_DROP	 	= 0x00000200,
+	CALI_GLOBALS_SVC_LOOP_REJECT	 	= 0x00000400,
```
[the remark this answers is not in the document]
sridhartigera: We have 3 config options. ```DROP```, ```REJECT``` and ```DISABLED```. Hence we need 2 flags.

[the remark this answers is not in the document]
tomastigera: OK, that makes sense

felix/bpf-gpl/tc.c
```diff
@@ -559,6 +559,34 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 		goto do_policy;
 	}
 
+	/* If the dest route is a blackhole route and packet is from an external source,
+	 * drop the packet, if service loop prevention is configured to either DROP or REJECT.
+	 */
+	if (cali_rt_is_blackhole(dest_rt) && !(ctx->state->flags & CALI_ST_SKIP_FIB)) {
+		if (CALI_F_FROM_HEP && !(cali_rt_lookup(&ctx->state->ip_src))) {
```
tomastigera, line 566:
```suggestion
	if (CALI_F_FROM_HEP && cali_rt_is_blackhole(dest_rt) && !(ctx->state->flags & CALI_ST_SKIP_FIB)) {
		if (!(cali_rt_lookup(&ctx->state->ip_src))) {
```

Why is the `!(ctx->state->flags & CALI_ST_SKIP_FIB)` important?

felix/bpf-gpl/tc.c
```diff
@@ -559,6 +559,34 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
...
+			struct cali_rt *src_rt = cali_rt_lookup(&ctx->state->ip_src);
```
tomastigera, line 567: Why do you test above `if (!(cali_rt_lookup(&ctx->state->ip_src))) {` and then you do the same lookup here again :thinking:

felix/bpf/ut/icmp_port_unreachable_test.go
```diff
@@ -198,3 +200,77 @@ func checkICMPv6PortUnreachable(pktR gopacket.Packet, ipv6 *layers.IPv6) {
...
+	_, ipv6, _, _, pktBytesV6, err := testPacketV6(nil, ipv6Default, nil, nil)
+	Expect(err).NotTo(HaveOccurred())
+
+	rtKeyV6 := routes.NewKeyV6(dstV6CIDR).AsBytes()
+	rtValV6 := routes.NewValueV6WithIfIndex(routes.FlagBlackHole, 1).AsBytes()
+	err = rtMapV6.Update(rtKeyV6, rtValV6)
+	Expect(err).NotTo(HaveOccurred())
+
+	defer func() {
+		fmt.Println("sridhar defer")
```
tomastigera, line 225: debug

felix/dataplane/linux/bpf_route_mgr.go
```diff
@@ -379,11 +385,15 @@ func (m *bpfRouteManager) calculateRoute(cidr ip.CIDR) routes.ValueInterface {
...
 		// It may be a localhost IP that is not assigned to a device like an
 		// k8s ExternalIP. Route resolver knew that it was assigned to our
 		// hostname.
-		flags |= routes.FlagsLocalHost
+		if cgRoute.Type == proto.RouteType_LOCAL_HOST {
+			flags |= routes.FlagsLocalHost
+		} else {
+			flags = routes.FlagBlackHole
+		}
 		fallthrough
```
tomastigera, line 397: woudn't it be much cleaner to get rid of the fallthrough and either call `route = m.bpfOps.NewValue(flags)` directly with the right flags for each case or take the default branch out and fo `if route == nil && flags != 0 { ... }`

felix/dataplane/linux/bpf_route_mgr.go
```diff
@@ -645,6 +655,44 @@ func (m *bpfRouteManager) onWorkloadEndpointRemove(update *proto.WorkloadEndpoin
...
+		cidr, err := ip.ParseCIDROrIP(cidrStr)
+		if err != nil {
+			log.WithError(err).WithField("cidr", cidr).Error(
+				"Failed to parse cidr.")
+		}
+		if uint8(m.ipFamily) != cidr.Version() {
+			continue
+		}
+
+		if _, ok := m.cidrToRoute[cidr]; ok {
```
tomastigera, line 678: what if it was previously a cidr for a different route? Don't you need to make sure it is a black hole? doesn't it also stay in the cidrsToDel so it get removed later on?

felix/fv/bpf_test.go
```diff
@@ -5509,16 +5509,20 @@ func bpfWaitForPolicy(felix *infrastructure.Felix, iface, hook, policy string) s
...
-	var (
-		out string
-		err error
-	)
-
 	if felix.TopologyOptions.EnableIPv6 {
-		out, err = felix.ExecOutput("calico-bpf", "-6", "routes", "dump")
-	} else {
-		out, err = felix.ExecOutput("calico-bpf", "routes", "dump")
+		return bpfDumpRoutesV6(felix)
```
tomastigera, line 5513: I'd keep this the way it is . It only burries the error deeper. Perhaps we should use Expect with offset in the original code to know where we are failing in the test.

felix/fv/service_loop_prevention_test.go
```diff
@@ -168,12 +180,11 @@ var _ = infrastructure.DatastoreDescribe("service loop prevention; with 2 nodes"
...
 			// Tcpdump should see just 1 packet, the request, with no response (because
 			// we DROP) and no looping.
-			Eventually(countServiceIPPackets).Should(BeNumerically("==", 1))
+			Eventually(countServiceIPPackets).Should(BeNumerically("==", count))
 		}
 	}
 
-	It("programs iptables as expected to block service routing loops", func() {
-
+	It("programs iptables/routes as expected to block service routing loops", func() {
```
tomastigera, line 187:
```suggestion
	It("programs dataplane as expected to block service routing loops", func() {
```

felix/fv/service_loop_prevention_test.go
```diff
@@ -208,13 +229,25 @@ var _ = infrastructure.DatastoreDescribe("service loop prevention; with 2 nodes"
...
-			Eventually(getCIDRBlockRules(felix, "iptables-save"), "8s", "0.5s").Should(ConsistOf(
-				MatchRegexp("-A cali-cidr-block -d 10\\.96\\.0\\.0/17 .* -j REJECT"),
-			))
-			Eventually(getCIDRBlockRules(felix, "ip6tables-save"), "8s", "0.5s").Should(ConsistOf(
-				MatchRegexp("-A cali-cidr-block -d fd5f::/119 .* -j REJECT"),
-			))
+		for id, felix := range tc.Felixes {
+			if BPFMode() {
+				curId := tcProgID[id]
+				Eventually(func() int {
```
tomastigera, line 235: there is a function to make sure programs are attached

felix/fv/service_loop_prevention_test.go
```diff
@@ -223,14 +256,22 @@ var _ = infrastructure.DatastoreDescribe("service loop prevention; with 2 nodes"
...
-		for _, felix := range tc.Felixes {
-			Eventually(getCIDRBlockRules(felix, "iptables-save"), "8s", "0.5s").Should(BeEmpty())
-			Eventually(getCIDRBlockRules(felix, "ip6tables-save"), "8s", "0.5s").Should(BeEmpty())
+		for id, felix := range tc.Felixes {
+			if BPFMode() {
+				curId := tcProgID[id]
+				Eventually(func() int {
+					tcProgID[id] = getTcProgID(felix, "eth0")
+					return tcProgID[id]
+				}, "15s", "1s").ShouldNot(BeNumerically("==", curId))
```
tomastigera, line 265: shouldn't you rather test that the blackhore rules are cleared?

felix/proto/felixbackend.proto
```diff
@@ -573,6 +573,7 @@ enum RouteType {
   LOCAL_HOST = 4;
   REMOTE_TUNNEL = 5;
   LOCAL_TUNNEL = 6;
+  BLACK_HOLE = 7;
```
tomastigera, line 576: This is a little unfortunate as itis local to the bpf ep manager :man_shrugging:

felix/bpf-gpl/tc.c
sridhartigera: We set ```CALI_ST_SKIP_FIB``` for node local DNS, which will hit the same blackhole route.

felix/fv/service_loop_prevention_test.go
sridhartigera: Blackhole routes are cleared only when the BGPConfig is deleted. There is ```ServiceLoopPrevention``` felix config, which is modified in this test. Config change triggers a felix restart. I tried using ```ensureAllNodesBPFProgramsAttached``` but that did not work well because the test did not wait for the felix to restart. Is there a way for the test to know if felix is restarting?

sridhartigera: Found a way.

felix/bpf-gpl/tc.c
tomastigera: Iirc that is only set when from WEP.

felix/fv/service_loop_prevention_test.go
```diff
@@ -223,20 +199,14 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ service loop prevention; wi
...
-		for id, felix := range tc.Felixes {
-			if BPFMode() {
-				curId := tcProgID[id]
-				Eventually(func() int {
-					tcProgID[id] = getTcProgID(felix, "eth0")
-					return tcProgID[id]
-				}, "15s", "1s").ShouldNot(BeNumerically("==", curId))
-			} else {
+		if BPFMode() {
+			ensureAllNodesBPFProgramsAttached(tc.Felixes)
```
tomastigera, line 207: is this enough? Shoudn't we also check for the right routes to be programmed?

sridhartigera: The routes won't change as there is no change in the BGPConfig. Felix config changes. Hence it is enough if we wait for the programs to be attached after felix restart.

felix/bpf-gpl/tc.c
```diff
@@ -562,25 +562,23 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 	/* If the dest route is a blackhole route and packet is from an external source,
 	 * drop the packet, if service loop prevention is configured to either DROP or REJECT.
 	 */
```
tomastigera: The comment is no longer correct

tomastigera (approved):
LGTM, thanks!

felix/fv/service_loop_prevention_test.go
```diff
@@ -168,12 +156,11 @@ var _ = infrastructure.DatastoreDescribe("service loop prevention; with 2 nodes"
 		} else {
 			// Tcpdump should see just 1 packet, the request, with no response (because
```
tomastigera, line 157: 1 in the comment is no longer correct since you introduced count

tomastigera: does the count match how many pings you send?

felix/fv/service_loop_prevention_test.go
```diff
@@ -190,62 +179,100 @@ var _ = infrastructure.DatastoreDescribe("service loop prevention; with 2 nodes"
...
+					MatchRegexp("-A cali-cidr-block -d 10\\.96\\.0\\.0/17 .* -j REJECT"),
+				))
+				Eventually(getCIDRBlockRules(felix, "ip6tables-save"), "8s", "0.5s").Should(ConsistOf(
+					MatchRegexp("-A cali-cidr-block -d fd5f::/119 .* -j REJECT"),
+				))
+			}
+		}
+		By("test that we don't get a routing loop and ICMP back to the sender")
+		if BPFMode() {
+			tryRoutingLoop(false, 2)
```
tomastigera, line 228: why don't we do this for iptables?

tomastigera (approved):
LGTM! Thanks! Could you squash all the commits in one?

PR #8904, opened 2024-06-12
Replace GOGC override with a config param.

fasaxc:
## Description

I've seen a lot of pprofs fro the field showing that our GOGC setting is too aggressive, particularly on the enterprise side.  Let's move the GOGC configuration to the FelixConfig and remove the 20% override by default.

## Todos

- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Felix now sets the Go runtime's GC threshold to 40% (instead of the more aggressive 20% used previously). This trades slight extra RAM usage for significantly lower GC CPU usage.  The setting is now exposed in the FelixConfiguration as goGCThreshold, along with goMemoryLimitMB.  To get the old behaviour, set goGCThreshold to 20.  If memory usage is not a concern, the value can be set even higher to reduce CPU usage.
```

## Reminder for the reviewer

Other optional labels:


mazdakn (approved):
LGTM. I just left few nits.

I was reading this part of the code, and was wondering why is this not configurable!:-D

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -648,6 +648,30 @@ type FelixConfigurationSpec struct {
 	// WindowsManageFirewallRules configures whether or not Felix will program Windows Firewall rules. (to allow inbound access to its own metrics ports) [Default: Disabled]
 	// +optional
 	WindowsManageFirewallRules *WindowsManageFirewallRulesMode `json:"windowsManageFirewallRules,omitempty" validate:"omitempty,oneof=Enabled Disabled"`
+
+	// GoGCThreshold Sets the Go runtime's garbage collection threshold.  I.e. the percentage that the heap is
+	// allowed to grow before garbage collection is triggered.  In general, doubling the value halves the CPU time
+	// spent doing GC, but it also increases memory usage by 50%.  A special value of -1 can be used
```
matthewdupre, line 654: I think I'm right in saying doubling the value does halve the GC CPU time regardless of previous value, but for memory it only increases by 50% when going from 100 to 200.

Peak memory usage is much more important than average because the best practice is to set memoryRequest=memoryLimit, so I think we should offer feedback using that.

I think a reasonable simplification would be to say "doubling the value halves the CPU time spent doing GC, but GC memory overhead is proportional to the value" maybe?

felix/config/config_params.go
```diff
@@ -1036,6 +1033,20 @@ func loadParams() {
...
+// mustParseOptionalInt returns dflt if the given value is empty, otherwise parses the value as an int.
+// Panics if the value is not a valid int.
+func mustParseOptionalInt(rawValue string, dflt int, fieldName string) int {
+	rawValue = strings.TrimSpace(rawValue)
+	if rawValue == "" {
+		return dflt
+	}
+	paramMin, err := strconv.Atoi(rawValue)
+	if err != nil {
+		log.Panicf("Failed to parse min value for %v", fieldName)
```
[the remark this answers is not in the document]
fasaxc: Oh yeah, forgot to clean up after refactoring :facepalm:

api/pkg/apis/projectcalico/v3/felixconfig.go
fasaxc: I think your wording is better, I struggled to come up with something clear and concise :+1:

PR #8905, opened 2024-06-12
Fix deprecations in yaml lint workflow

danudey:


matthewdupre:
@danudey What's the right way to check this is good?  The lint CI run on the PR seems happy and I'm comfortable the 2 failures in the main CI are unrelated.

danudey:
> @danudey What's the right way to check this is good? The lint CI run on the PR seems happy and I'm comfortable the 2 failures in the main CI are unrelated.

I've compared the changed YAML files with relevant parsers (`yq` being the main one), and the results are identical (I actually generated the updated files via `yq`). At this point our only warnings are line length and missing doc start, which I think is a good place to be.

I'd love to expand this out to lint other yaml files, etc., but for now I'm happy leaving it here and seeing how it plays going forward.

coutinhop (approved):

PR #8909, opened 2024-06-13
Rebuild containernetworking binaries, so we don't need to create rele…

rene-dekker:
We never moved https://github.com/projectcalico/containernetworking-plugins to the monorepo, because the long-term vision is to use upstream. Because of this, we need to make a new release to this project every time a new golang patch comes out.

I added some automation to download the code of the project and rebuild it with go-build from the monorepo, thereby removing this tedious step of creating new releases and updating the makefile with the new version.

### DIFF
--- a/cni-plugin/Makefile
+++ b/cni-plugin/Makefile
@@ -25,9 +25,6 @@ WINFV_SRCFILES=$(shell find win_tests -name '*.go')
 # fail if unable to download
 CURL=curl -C - -sSf
 
-# The CNI plugin code that will be cloned and rebuilt with this repo's go-build image
-# whenever the cni-plugin image is created.
-CNI_VERSION=v1.1.1-calico+go-1.22.3
 CNI_ARTIFACTS_URL=https://github.com/projectcalico/containernetworking-plugins/releases/download
 FLANNEL_ARTIFACTS_URL=https://github.com/projectcalico/flannel-cni-plugin/releases/download
 FLANNEL_VERSION=v1.2.0-flannel2-go1.22.2
--- a/cni-plugin/Makefile
+++ b/cni-plugin/Makefile
@@ -25,7 +25,8 @@ WINFV_SRCFILES=$(shell find win_tests -name '*.go')
 # fail if unable to download
 CURL=curl -C - -sSf
 
-# Use forked CNI plugin URL and corresponding tagged artifacts.
+# The CNI plugin code that will be cloned and rebuilt with this repo's go-build image
+# whenever the cni-plugin image is created.
 CNI_VERSION=v1.1.1-calico+go-1.22.3
 CNI_ARTIFACTS_URL=https://github.com/projectcalico/containernetworking-plugins/releases/download
 FLANNEL_ARTIFACTS_URL=https://github.com/projectcalico/flannel-cni-plugin/releases/download
@@ -61,7 +62,7 @@ clean: clean-windows
 	rm -rf bin $(DEPLOY_CONTAINER_MARKER) pkg/install/install.test
 	rm -rf config/ dist/
 	-docker image rm -f $$(docker images $(CNI_PLUGIN_IMAGE) -a -q)
-	rm -rf containernetworking-plugins .containernetworking-plugins.created
+	rm -rf containernetworking-plugins .containernetworking-plugins-*
 
 clean-windows: clean-windows-builder
 	rm -rf $(WINDOWS_BIN) $(WINDOWS_DIST)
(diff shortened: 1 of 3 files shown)


cni-plugin/Makefile
```diff
@@ -140,13 +141,31 @@ $(DEPLOY_CONTAINER_FIPS_MARKER): Dockerfile build fetch-cni-bins
 	$(MAKE) retag-build-images-with-registries VALIDARCHES=$(ARCH) IMAGETAG=latest-fips LATEST_IMAGE_TAG=latest-fips
 	touch $@
 
+CN_FLAGS=-ldflags "-X github.com/containernetworking/plugins/pkg/utils/buildversion.BuildVersion=$(GIT_VERSION)"
+
+CONTAINERNETWORKING_PLUGINS_CREATED=.containernetworking-plugins-$(ARCH).created
+CONTAINERNETWORKING_PLUGINS_CLONED=.containernetworking-plugins.cloned
+
+$(CONTAINERNETWORKING_PLUGINS_CLONED):
```
fasaxc, line 149: I think this will get stuck on an old branch when you pull an update to the calico repo that changes CNI_VERSION

I've used this pattern in the past to deal with this:
```
TOUCHFILE=.foo-bar-cloned.$(VERSION)

$(TOUCHFILE):
    rm -rf <the repo>  .foo-bar-cloned.*
    ... clone it ...
    touch $@
```

cni-plugin/Makefile
```diff
@@ -140,13 +141,31 @@ $(DEPLOY_CONTAINER_FIPS_MARKER): Dockerfile build fetch-cni-bins
...
+	git clone --single-branch --branch $(CNI_VERSION) https://github.com/projectcalico/containernetworking-plugins.git
+	touch $@
+
+$(CONTAINERNETWORKING_PLUGINS_CREATED): $(CONTAINERNETWORKING_PLUGINS_CLONED)
```
fasaxc, line 153: Not sure it's any cleaner(!) but there's a way to tell make that this target makes several binaries on a single invocation:
```
.../bin/foo .../bin/bar .../bin/baz &: $(TOUCHFILE)
    ...
```
The `&:` means "one call to this creates all these files"

I used that in the BPF builds to avoid issues when calling through to the BPF makefile.

fasaxc (changes requested):
I think you need to make sure that updates to the CNI_VERSION trigger a rebuild.

rene-dekker:
> I think you need to make sure that updates to the CNI_VERSION trigger a rebuild.

I'll make the changes, but don't think the necessity is that high. The binaries are built on every merge job and so the (hash) releases will always have the latest go. Other than golang updates, this repo has been dead for years. For the same reason do I think that we don't need to ever update the CNI_VERSION any longer.

cni-plugin/Makefile
rene-dekker: I took over your suggestion. I think it does look slightly cleaner now. Thanks for teaching me this feature.

cni-plugin/Makefile
```diff
@@ -25,7 +25,8 @@ WINFV_SRCFILES=$(shell find win_tests -name '*.go')
 # fail if unable to download
 CURL=curl -C - -sSf
 
-# Use forked CNI plugin URL and corresponding tagged artifacts.
+# The CNI plugin code that will be cloned and rebuilt with this repo's go-build image
+# whenever the cni-plugin image is created.
 CNI_VERSION=v1.1.1-calico+go-1.22.3
```
coutinhop, line 30: suggestion (maybe unrelated): how about we move this version to metadata.mk?

coutinhop (approved):
LGTM

fasaxc:
> I'll make the changes, but don't think the necessity is that high. The binaries are built on every merge job and so the (hash) releases will always have the latest go.

Thanks, you're right that release builds should always be clean, but developers use makefiles too, and it's not sustainable for us to maintain an ever growing mental list of "things the makefile can't handle".  I really don't want to have to run `make clean` every time I switch branches.

cni-plugin/Makefile
```diff
@@ -183,7 +203,7 @@ ut-datastore:
 	-v $(CERTS_PATH):/home/user/certs \
 	$(CALICO_BUILD) sh -c '$(GIT_CONFIG_SSH) \
 			cd  /go/src/$(PACKAGE_NAME) && \
-			ginkgo -cover -r -skipPackage pkg/install $(GINKGO_ARGS)'
+			ginkgo -cover -r -skipPackage pkg/install,containernetworking-plugins $(GINKGO_ARGS)'
```
caseydavenport, line 206: Do we really want to run unit tests that aren't "ours" as part of our CI? I would have suggested not, but maybe there is a reason?

rene-dekker: They are part of the skipPackage flag. The CI runs before the release can be made in the upstream repo, so I see no use for running them again. Also, the suite is flaky and sometimes has to run two or three times before it passes successfully.

caseydavenport: d'oh, yeah I shouldn't be reviewing PRs this early in the morning.

fasaxc (approved):

PR #8911, opened 2024-06-14
Update semaphore runners to new f1 instances

danudey:


hjiawei:
f1 runners are gone. Closing.

Issue #8918, opened 2024-06-17
Calico/eBPF node flooding unknown unicast traffic


alessandro-veras:
## Expected Behavior
Unknown unicast traffic with a dst mac address not local to a K8s/Calico node should be dropped.

## Current Behavior
We are having an issue with our Calido/eBPF implementation. When we shut down any VMs in the datacenter, some traffic is flooded as unknown unicast, initially, it’s a small number of ICMP echo requests from a probe. 
Some K8s nodes we have running calico/eBPF accept this traffic ( even if the dst mac in the packet is not local to the node ) and forward this traffic back to the network.  This is creating an amplifying effect as other K8s nodes take this traffic and forward it back to the network. A few packets are becoming 10K-15K.
When we see it in the capture we are doing in the hypervisor, we can also see that the K8s node is changing the src mac address to the address it has in the local interface.

## Possible Solution
Make sure Calico/eBPF drops packets with a dst mac address not local to the node. 

## Steps to Reproduce (for bugs)

1. We shut down a vm that has a status page application sending icmp requests to check if the VM is alive.
2. The forwarding information for that VM in the switches is gone since the VM is down, and now those ICMP packets are flooded to all hosts in the same VLAN.
3.K8s receives this traffic and sends it back to the network (changing the src mac address ) creating an amplification effect, flooding more traffic in the network.

## Context

The issue is impacting the k8s due to the resouces necessary to deal with the flooding and it's also impacting the network and other nodes due the volume of pps being flooded.

## Your Environment

* Calico version: v3.27.3
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes
* Operating System and version: Ubuntu 20.04 LTS
* Link to your project (optional):

aitorpazos:
@tomastigera could it be possible to backport #8921 to 3.27.x ?

PR #8921, opened 2024-06-17
[BPF] skip FIB when dest is unknown at HEP

tomastigera:
## Description

    [BPF] skip FIB when dest is unknown at HEP
    
    * fix when CALI_ST_SKIP_FIB is set on the way to the host, set
      CALI_CT_FLAG_SKIP_FIB on conntrack - not just when from WEP
    
    * add test for ^^^ and issue https://github.com/projectcalico/calico/issues/6450
    
    * In addition to skipping FIB when there is no route to post-dnat
      destination, also skip FIB when there is a route, but it is not local
      while there was no service involved. In that case, we are not
      forwarding a service (NodePort) to another node and we should only
      forward locally. Let the host decide what to do with such a packet.

    [BPF/UT] set RPF to strict as test provision for that.
    
    Make tests execute the path again after we introduced the option to
    disable RPF.

Fixes https://github.com/projectcalico/calico/issues/8918

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: fix - let the node handle packet when we are not sure about the destination
```

## Reminder for the reviewer

Other optional labels:


marvin-tigera:

fasaxc (approved):
LGTM

Issue #8882, opened 2024-06-05
eBPF not issuing TCP RST on unknown connections


sfudeus:
## Expected Behavior

When a node receives a packet for an unknown TCP connection on the host IP or to a closed port, it should by default reject that connection by issuing a packet with the TCP RST bit set. There may be situations where it should be silently dropped, in default it should behave transparently. 

This behaviour should be identical between the eBPF mode and the iptables mode.

## Current Behavior

When sending a TCP packet to a port on the hostIP of a node running calico on eBPF mode, for unknown connections, there is no reply sent, the packet is silently dropped and the client will resume retries until timeout. This happens both for unknown connections against closed ports and open ports. New connections against closed ports are properly rejected.

When sending such an unknown TCP packet to a node running calico on iptables mode, there is an immediate TCP RST packet and the client can immediately terminate (and potentially reestablish a connection). 

## Possible Solution

Even in eBPF mode, a packet with RST bit set should be sent for any connection not in conntrack. 

## Steps to Reproduce (for bugs)

* Send a TCP packet without proper prior handshake to a nodeIP of a node which is eBPF enabled, regardless if the port is open or not. 
* Observe that you won't get a reply. 

Trivial python script to trigger this:
```
#!/usr/bin/env python3

from scapy.all import *
import argparse

def main():
    parser = argparse.ArgumentParser("rst-test")
    parser.add_argument("dest_host", help="the destination host to test against")
    parser.add_argument("dest_port", help="the destination port to test against", type=int)
    args = parser.parse_args()

    packet = IP(dst=args.dest_host) / TCP(sport=13337, dport=args.dest_port, seq=1337, ack=1336, flags="PA") / Raw(load="test")

    response = sr1(packet, timeout=10)

    if response:
        if response.haslayer(TCP) and response.getlayer(TCP).flags & 0x04:
            print("Received RST packet")
        else:
            print("Did not receive RST packet")
    else:
        print("No response received")

if __name__ == "__main__":
    main()
```

## Context

We are running the apiservers on an anycast IP in the hostNetwork, announced via BGP. When one master node goes down and withdraws its announcement, ECMP routing will route the traffic to a remaining control plane node. With that, such a remaining node will see traffic for unknown TCP connections, right after the ECMP change. 

[...]

PR #8933, opened 2024-06-20
[BPF] send RST to midflow TCP packets from unknown streams

tomastigera:
## Description

    For a smooth switch from iptables to ebpf mode, we do not want to
    interrupt existing connections. If we see midflow packets, we pass them
    to the host stack. If the stack can verify that they belong to an
    existing conntrack, we let them through and we learn the conntrack.
    
    We drop the rest. However, there are some situations when we can see a
    stray TCP packet during ebpf mode, for instance when a pod dies and ECMP
    kicks in and sends a packet to a different host.
    
    If such a packet gets dropped, the end of the connections remains stuck.
    This change sends an RST to such a stream instead of just dropping the
    packets so that the end host can break the connection.
    
Fixes https://github.com/projectcalico/calico/issues/8882

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: don't drop, but reject unknown midflow tcp packets with rst 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/int_dataplane.go
+++ b/felix/dataplane/linux/int_dataplane.go
@@ -1453,8 +1453,15 @@ func (d *InternalDataplane) setUpIptablesBPF() {
 				Action:  iptables.AcceptAction{},
 			},
 			iptables.Rule{
+				Match: iptables.Match().
+					MarkMatchesWithMask(tcdefs.MarkSeenFallThrough, tcdefs.MarkSeenFallThroughMask).
+					Protocol("tcp"),
+				Comment: []string{"REJECT/rst packets from unknown TCP flows."},
+				Action:  iptables.RejectAction{With: "tcp-reset"},
+			},
+			iptables.Rule{
 				Match:   iptables.Match().MarkMatchesWithMask(tcdefs.MarkSeenFallThrough, tcdefs.MarkSeenFallThroughMask),
-				Comment: []string{fmt.Sprintf("%s packets from unknown flows.", d.ruleRenderer.IptablesFilterDenyAction())},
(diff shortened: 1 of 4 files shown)


felix/dataplane/linux/int_dataplane.go
```diff
@@ -1452,9 +1452,16 @@ func (d *InternalDataplane) setUpIptablesBPF() {
...
+				Match: iptables.Match().
+					MarkMatchesWithMask(tcdefs.MarkSeenFallThrough, tcdefs.MarkSeenFallThroughMask).
+					Protocol("tcp"),
+				Comment: []string{"REJECT/rst packets from unknown TCP flows."},
+				Action:  iptables.RejectAction{With: "tcp-reset"},
+			},
 			iptables.Rule{
 				Match:   iptables.Match().MarkMatchesWithMask(tcdefs.MarkSeenFallThrough, tcdefs.MarkSeenFallThroughMask),
-				Comment: []string{fmt.Sprintf("%s packets from unknown flows.", d.ruleRenderer.IptablesFilterDenyAction())},
+				Comment: []string{fmt.Sprintf("%s packets from unknown TCP flows.", d.ruleRenderer.IptablesFilterDenyAction())},
```
fasaxc, line 1464:
```suggestion
				Comment: []string{fmt.Sprintf("%s packets from unknown non-TCP flows.", d.ruleRenderer.IptablesFilterDenyAction())},
```

fasaxc (changes requested):

tomastigera:
only a single unrelated failure fixed here https://github.com/projectcalico/calico/pull/8952

fasaxc (approved):

PR #8935, opened 2024-06-22
Start go build and release tooling

radTuti:
## Description

This is the core base of the go release tooling that will replace useage of Makefile and hack/release for a more cohesive process. It tries to adhere to [standard go project layout](https://github.com/golang-standards/project-layout)

The borrows from the town in Bob the Builder: Ready Steady Build

## Related issues/PRs

N/A

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

I'm going to go with not required atm until it gets intergrated

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


fixham/pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,54 @@
...
+// Main is the entrypoint for the goyek build system.
+//
+// It defines common tasks that can be run.
+// It also sets the default task to run when no task is specified.
+//
+// If no default tasks are specified, the DefaultTask runs the "static-checks" task.
+func Main(packageName string, rootVolume string) {
+	flag.StringVar(&bootstapPackageName, "package", LookupEnv("PACKAGE_NAME", packageName), "The package name to use")
+	pkgName = &bootstapPackageName
+	if rootVolume == "" {
```
caseydavenport, line 40: I think we should use `envconfig` to handle reading environment variable configurations and avoid the need to scatter `os.LookupEnv` calls throughout the codebase, which become tricky to track down.

```go
type Config struct {
  PackageName string `envconfig:"PACKAGE_NAME"`
}
```

then

```go
cfg := Config{}
envconfig.Process("", &cfg)
```

e.g., https://github.com/projectcalico/calico/blob/616bad52431ff8aa4e2d6cfe9ec1a471045fad3e/libcalico-go/lib/apiconfig/apiconfig.go#L51-L65

fixham/pkg/ctl/ctl.go
```diff
@@ -0,0 +1,30 @@
+package ctl
+
+import (
+	"github.com/projectcalico/fixham/internal/client"
+	"github.com/projectcalico/fixham/internal/docker"
+)
+
+// Client is used to interact with the ctl package.
```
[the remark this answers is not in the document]
caseydavenport: From what I can tell, this package doesn't even need to exist if we remove these aliases.

fixham/build/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"github.com/goyek/goyek/v2"
+
+	"github.com/projectcalico/fixham/pkg/bootstrap"
+)
+
+const packageName = "github.com/projectcalico/fixham"
+
+var _ = bootstrap.DefineCleanTask([]string{"./bin/*"}, nil, nil)
```
caseydavenport, line 11: I think we should think a bit more about how to structure this / build an API to use in other repositories - there are some global variables being initialized / written at different points in time that make this code tricky to follow, and prone to issues like nil pointer exceptions if things are executed in the wrong order. Getting the code structure as close to "right" as we can will have a huge positive impact on future development.

Here's some pretty rough pseudocode that hopefully illustrates one idea for how to organize this code. If it's not clear, let's hop on a Zoom call to discuss:

```go
package api

// Package represents a package within the projectcalico/calico repository that can have one or more
// tasks associated with it.
type Package interface {
  Path() string
  Tasks() []goyek.Task
}
```

```go
package main

func main() {
  // Load configuration from environment.
  cfg := LoadConfig() 

  // Initialize the package we want to execute tasks for.
  fixham := NewFixham(cfg)
  tasks.Register(fixham)
}
```

```go
package main

func NewFixham(cfg *config.Config) api.Package {
  return &fixham{...}
}

func (f *fixham) Clean() goyek.Task {
  return common.Clean(...)
}

func (f *fixham) Tasks() []goyek.Task {
  return []goyek.Task{f.Clean()}
}
```

```go
package tasks

func Register(p api.Package) error {
  for _, task := range p.Tasks() {
    goyek.Define(task)
  }
}
```

fixham/internal/client/client.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (c *Client) WithRoot(root string) *Client {
+	c.goBuildRunner.WithRepoVolume(root)
+	return c
+}
+
+func (c *Client) Lint(args ...string) {
+	gitConfigSSH := ""
+	if os.Getenv("GIT_USE_SSH") == "true" {
```
caseydavenport, line 37: This should be attached to a common Configuration struct rather than having `os.Getenv` embedded right in - make it harder to lookup what all the supported config options are this way. Then, `NewClient` would be updated like this:

```go
func NewClient(c config.Config) *Client { . . .}
```

And this line would be:

```go
if c.config.GitUseSSH {
}
```

fixham/internal/client/client.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/fixham/internal/docker"
+)
+
+type Client struct {
```
caseydavenport, line 13: What is this a client for? A client generally implies there is a server or service of some sort that this is a frontend for accessing, but I don't think that's the case here.

caseydavenport (changes requested):

fixham/internal/docker/gobuild.go
```diff
@@ -0,0 +1,248 @@
...
+		},
+		hostConfig: &container.HostConfig{
+			NetworkMode: "host",
+			Binds: []string{
+				fmt.Sprintf("%s:%s:rw", gomodCacheDir, modCacheDir),
+				fmt.Sprintf("%s/.go-pkg-cache:%s:rw", repoRootDir, goCacheDir),
+			},
+		},
+	}
+	if currentUser.Uid == "0" {
```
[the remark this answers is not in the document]
radTuti: I don't control that as Uid is how it is defined in the user.User package.

lib.Makefile
```diff
@@ -567,6 +567,10 @@ static-checks: $(LOCAL_CHECKS)
 
 LINT_ARGS ?= --max-issues-per-linter 0 --max-same-issues 0 --timeout 8m
 
+.PHONY: pilchard
```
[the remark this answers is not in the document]
radTuti: oops...i was using it to test something and it shouldn't have been part of this PR.

fixham/internal/docker/gobuild.go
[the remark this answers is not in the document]
caseydavenport: Ick, Go not following their own idiomatic naming conventions :shaking_face:

fixham/pkg/bootstrap/test.go
```diff
@@ -0,0 +1,109 @@
+package bootstrap
+
+import (
+	"github.com/goyek/goyek/v2"
+	"github.com/sirupsen/logrus"
+)
+
+// TestType represents the type of test to run.
```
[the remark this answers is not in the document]
radTuti: This was to accommodate for the section in `lib.Makefile`

```makefile
## Run the tests for the current platform/architecture
test: ut fv st
```

I attempted to allow the ability for "test" to only a type of test if it is defined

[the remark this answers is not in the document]
radTuti: I think each repo is not restricted from definiing other tasks as well

fixham/internal/client/client.go
radTuti: Removing it. Also based on the suggested structure, this will be unnecessary

fixham/internal/docker/gobuild.go
```diff
@@ -0,0 +1,269 @@
...
+type GoBuildImage struct {
+	imageName    string
+	imageVersion string
+}
+
+// NewGoBuild returns a new GoBuildImage
+// with default image name and version
+func NewGoBuild(version string) GoBuildImage {
+	if version == "" {
+		version = defaultGoBuildVersion
```
hjiawei, line 34: It is better to return an error when `version` isn't set rather than returning a default value. calico/go-build v0.91 isn't special. It happens to be the version we current use and will get outdated pretty soon. Setting default values to a system isn't always good when they are not sane.

We frequently update go-build versions in `metadata.mk` file. Can we read from there?

radTuti: I have updated it. This tool will not use Makefiles but the value can be updated in [config](https://github.com/projectcalico/calico/pull/8935/files#diff-a8f99f44b58022c30358fd5f26c2a1ca5c1580c07c1cad437a9bb913b731497aR6)

caseydavenport: >  We frequently update go-build versions in metadata.mk file. Can we read from there?

Agree - we should read this from `metadata.mk`, at least for now. We should not have two places to configure this.

Longer term perhaps we can switch this to go the other way, so that this is defined in the Go code and nowhere else, but we need to make sure we only have a single place this value is configured at any given time.

fixham/go.mod
```diff
@@ -0,0 +1,44 @@
+module github.com/projectcalico/fixham
```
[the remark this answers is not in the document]
radTuti: fixham is a separate module that can be use on its own similar to api

fixham/internal/docker/gobuild.go
radTuti: updated to read from `metadata.mk`

fixham/go.mod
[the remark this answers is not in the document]
caseydavenport: I believe since we're narrowing in on release logic, we don't need to treat fixham as a library to be imported, rather it's a CLI tool that will orchestrate the release. So with that in mind, I think it will be better to use the existing go.mod / go.sum from the repository root.

fixham/Makefile
danudey, line 1: Instead of @'ing every command, you can use the `.SILENT` pseudo-target. It's allegedly obsolete since we can use @ (according to the GNU Make manual) but if you're trying to suppress every command it's a lot cleaner and can be toggled on and off.

https://stackoverflow.com/questions/24005166/gnu-make-silent-by-default

radTuti: as we don't currently use `.SILENT` it is fair to leave it out of this PR as this file will not have a lot more changes to it

fixham/internal/operator/git.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/projectcalico/calico/fixham/internal/command"
+)
+
+const (
+	operatorRepo = "git@github.com:tigera/operator.git"
+)
+
+func operatorDir(repoRootDir string) string {
+	return repoRootDir + "/operator"
```
caseydavenport, line 14: We shouldn't be cloning the operator (or any other intermediate files) into the root of the projectcalico/calico repo.

I think the best thing to do is to use a temporary directory generated for each invocation. Each directory should be unique and include the hashrelease name / version so that there is no chance of accidentally using the wrong clone if, for example, you run multiple hashreleases back-to-back on different branches.

fixham/internal/operator/operator.go
```diff
@@ -0,0 +1,34 @@
...
+	"strings"
+
+	"github.com/projectcalico/calico/fixham/internal/command"
+)
+
+const (
+	ImageName = "tigera/operator"
+)
+
+func GenVersion(rootDir, componentsVersionPath string) error {
```
caseydavenport, line 15: This is more of an optimization for the future, but `gen-versions` currently clones the `github.com/projectcalico/calico` repository as part of the generation. This is obviously a bit over-kill since we're running this from within the Calico repo already!

We should track a follow-on task to remove this sort of nested dependency.

fixham/Makefile
```diff
@@ -0,0 +1,35 @@
...
+build: bin/hashrelease bin/release
+
+clean:
+	@rm -rf ./bin
+
+bin/release:
+	@mkdir -p bin
+	@go build -o bin/release ./release
+
+bin/hashrelease:
```
caseydavenport, line 18: In order to get Make to track rebuilds properly, you should have this depend on the files that you want to trigger rebuilds.

For example:

```
bin/hashrelease: $(shell find . -name "*.go")
```

^ This will rebuild bin/hashrelease whenever a .go file changes.

fixham/internal/operator/operator.go
```diff
@@ -0,0 +1,34 @@
...
+	if _, err := command.MakeInDir(operatorDir(rootDir), []string{"gen-versions"},
+		[]string{"OS_VERSIONS=" + componentsVersionPath, "COMMON_VERSIONS=" + componentsVersionPath}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func ImageAll(archs []string, version string) error {
+	if _, err := command.Make([]string{"image-all"}, []string{fmt.Sprintf("ARCHES=\"%s\"" + strings.Join(archs, " ")), "VERSION=" + version}); err != nil {
```
caseydavenport: I think this needs to be `MakeInDir` and pass the operator directory

fixham/internal/operator/operator.go
```diff
@@ -0,0 +1,34 @@
...
+		return err
+	}
+	return nil
+}
+
+func InitImage(version string) error {
+	if _, err := command.Make([]string{"image-init"}, []string{"VERSION=" + version}); err != nil {
```
caseydavenport: I think this needs to be `MakeInDir` and pass the operator directory

fixham/internal/dict/dict.go
```diff
@@ -0,0 +1,46 @@
+package dict
```
caseydavenport, line 1: Don't need a whole package for this - can just plop it in `hashrelease` probably, since that's the only place it's used.

fixham/internal/docs/dev.go
```diff
@@ -0,0 +1,103 @@
+package docs
```
caseydavenport, line 1: I think docs is an unintuitive name for this package - it seems to actually be responsible for publishing and interacting with the hashrelease artifacts (or at least some of them).

I think this might be better named as "publish" and then we can move all of the publish logic here?

Alternatively, this can all just be moved into the "hashrelease" package since the logic to publish and delete hash releases is very specific to hashreleases.

The release notes generation probably belongs in its own releasenote package?

fixham/pkg/api/builder.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// DockerRunner returns a DockerRunner to be used in the component
+func (c *Builder) DockerRunner() *docker.DockerRunner {
+	runner := docker.MustDockerRunner()
+	return runner
+}
+
+// DockerGoBuildRunner returns a GoBuildRunner to be used in the component
+func (c *Builder) DockerGoBuildRunner() *docker.GoBuildRunner {
```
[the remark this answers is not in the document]
caseydavenport: Same for a number of these functions - I think this "Builder" concept may be less useful now that we are scoping this to be a release / hashrelease tool instead of a generic framework for building and running tasks. Do we need it at all?

fixham/pkg/api/calicobuilder.go
```diff
@@ -0,0 +1,18 @@
+package api
+
+// CalicoBuilder is a component in the Calico project
+type CalicoBuilder struct {
```
caseydavenport, line 4: I'm not sure I understand the point of this structure any more. It's a builder that overrides the Path() implementation, but why?

caseydavenport: This might be another one that isn't needed any more in the newer factoring of this code?

fixham/internal/version/version.go
```diff
@@ -0,0 +1,79 @@
...
+	*v = Version(_v.String())
+	return nil
+}
+
+func (v *Version) String() string {
+	return string(*v)
+}
+
+func (v *Version) FormattedString() string {
+	if *v == "" {
```
caseydavenport, line 30: when is `v` going to be an empty string?

fixham/internal/version/version.go
```diff
@@ -0,0 +1,79 @@
...
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+
+	"github.com/projectcalico/calico/fixham/internal/calico"
+)
+
+type Version string
+
+func (v *Version) Set(value string) error {
+	_v, err := semver.NewVersion(strings.TrimPrefix(value, "v"))
```
caseydavenport, line 17: `_v` stands out a bit in Go - it's not a very command way of defining a variable (unlike Python).

I'd call this `ver` probably

fixham/internal/version/version.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/Masterminds/semver/v3"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+
+	"github.com/projectcalico/calico/fixham/internal/calico"
+)
+
+type Version string
+
+func (v *Version) Set(value string) error {
```
caseydavenport, line 16: This function is never called - is it needed?

It's a pretty odd pattern to have unless there's really a use-case for setting a struct. Typically something like being hyper memory efficient.

I think this probably wants a package-level New() function instead.

```go
func New(val string) Version {
  // Validate, then cast as a Version().
  return Version(val)
}
```

fixham/internal/version/version.go
```diff
@@ -0,0 +1,79 @@
...
+		prerelease := semver.MustParse(_v.Prerelease())
+		nextPrerelease := prerelease.IncMajor()
+		vNextPrerelease := fmt.Sprintf("%v.%v", nextPrerelease.Major(), nextPrerelease.Minor())
+		vNext := semver.New(_v.Major(), _v.Minor(), _v.Patch(), vNextPrerelease, "")
+		return Version(vNext.String())
+	}
+	return Version(_v.IncMinor().String())
+}
+
+func NextBranch(currentVersion string, devTagSuffix, releaseBranchPrefix string) string {
```
caseydavenport, line 69: This function, as well as NextVersion, seem to be unused. Is that intentional?

fixham/pkg/goyek/hashrelease.go
```diff
@@ -0,0 +1,117 @@
+package goyek
+
+import (
+	_goyek "github.com/goyek/goyek/v2"
```
caseydavenport, line 4: this should probably be `goyekv2` instead of `_goyek` 

see my other comment re: the use of underscores in Golang variable and package names.

fixham/internal/version/version.go
```diff
@@ -0,0 +1,79 @@
...
+func (v *Version) FormattedString() string {
+	if *v == "" {
+		return ""
+	}
+	return fmt.Sprintf("v%v", *v)
+}
+
+func (v *Version) Milestone() string {
+	_v := semver.MustParse(string(*v))
+	return fmt.Sprintf("%s v%d.%d.%d", cases.Title(language.English).String(calico.ProductName), _v.Major(), _v.Minor(), _v.Patch())
```
caseydavenport, line 38: Another instance of `_v` - should fix these up throughout.

fixham/pkg/goyek/hashrelease.go
```diff
@@ -0,0 +1,117 @@
...
+
+func HashreleaseBuild(cfg *config.Config) *GoyekTask {
+	return &GoyekTask{
+		Task: _goyek.Task{
+			Name:  hashreleaseBuildTaskName,
+			Usage: "Build hashrelease",
+			Action: func(a *_goyek.A) {
+				tasks.HashreleaseBuild(cfg)
+				tasks.HashreleaseNotes(cfg)
+				// TODO: either validate hashrelease here or in the publish task
```
caseydavenport, line 69: Do we want a separate "validate" task?

fixham/pkg/goyek/hashrelease.go
```diff
@@ -0,0 +1,117 @@
...
+				tasks.HashreleaseClean(cfg)
+				// TODO: restore repo to original state
+			},
+			Parallel: false,
+		},
+		Deps: []string{hashreleaseBuildTaskName},
+	}
+}
+
+func HashreleaseClean(cfg *config.Config) *GoyekTask {
```
caseydavenport, line 93: Maybe "HashreleaseGarbageCollect"? 

"Clean" implies to me like "make clean", i.e.,e removing local files.

metadata.mk
```diff
@@ -22,6 +22,8 @@ UBI_VERSION=8.10
 # Configuration for Semaphore integration.
 ORGANIZATION = projectcalico
 
+REPO_ROOT := $(shell git rev-parse --show-toplevel)
```
caseydavenport, line 25: metadata.mk is intended to be statically configured metadata that varies per-branch / per-product / etc. 

There's already a REPO_ROOT in `lib.Makefile` that we should use instead, and remove this from here.

caseydavenport (commented):
Comments for today - want to push them out before too long.

I'll continue looking at this on Thursday as well (tomorrow is a bit of a wash scheduling wise for me since I have an event in the afternoon).

fixham/Makefile
```diff
@@ -0,0 +1,35 @@
...
+
+include ../lib.Makefile
+
+.PHONY: build
+build: bin/hashrelease bin/release
+
+clean:
+	@rm -rf ./bin
+
+bin/release:
```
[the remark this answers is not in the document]
caseydavenport: After playing with this some more I think I am OK with the way you have factored it. 

Unfortunately goyek doesn't allow for sub-commands.

Long term I think we should ditch goyek and move to a proper CLI library that allows for sub-commands, but it's low priority.

fixham/pkg/api/builder.go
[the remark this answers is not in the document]
radTuti: this function will still be needed once we start to optimize the current logic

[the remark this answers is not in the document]
caseydavenport: @radTuti could you give an example of the optimizations you have in mind that will need this?

I very well could be misremembering, but I recall discussing that this wasn't needed last time we hopped on Zoom.

I would rather we leave it out of this PR if it isn't being used. We can always add it in another PR if there is a use for it, but dead code / functions that aren't called shouldn't be merged.

fixham/internal/version/version.go
radTuti: TBH only from user error 

```go
v := Version()
v.FormattedString() // this will be an empty string
```

fixham/internal/version/version.go
radTuti: straglers from branch cutting.
They will be used in release as we do need to tag a blank commit after release for next version
Also with cutting new branch.

fixham/pkg/api/builder.go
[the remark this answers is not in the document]
radTuti: I will take it out since there is no immediate usage.

I was hoping the gen-version work for operator but I will need to re-think that implentation ATM

fixham/internal/version/version.go
caseydavenport: @radTuti What I meant by this is that we shouldn't be commit any unused code as part of this PR - even if we think it will be used a in future commit. It's generally bad practice to commit "dead" code as part of a PR, so please remove any functions that are not called in this PR itself, so we can merge a minimally viable change set.

Any functions that are likely to be used in a future PR can be stored in another branch for now so that they are not lost, but it adds extra work for the reviewer when a PR has unused code in it.

fixham/internal/registry/registry.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("failed to get bearer token: %s", resp.Status)
+	}
+	var tokenResp struct {
```
caseydavenport, line 68: This is pretty unusual.

I think I'd opt for defining the struct type and variable separately just for more standard looking code. 

```go
type tokenResp struct {
}
var r tokenResp
```

or just using a map[string]interface{]

```go
resp := map[string]interface{}{}
if err := json.NewDecoder(resp.Body).Decode(&resp); err != nil {
  return "", err
}
return resp["token"].(string), nil
```

fixham/internal/release/releasenotes.go
```diff
@@ -0,0 +1,228 @@
...
+		milestones, resp, err := client.Issues.ListMilestones(context.Background(), owner, repo, nil)
+		if err != nil {
+			return -1, err
+		}
+		for _, m := range milestones {
+			if m.GetTitle() == milestone {
+				return m.GetNumber(), nil
+			}
+		}
+		if resp.NextPage == 0 {
```
caseydavenport, line 58: I don't think we're ever passing `NextPage` to the `ListMilestones` call, so we're not actually getting pagination here. 

If the # of milestones is ever more than a single page, we'll loop forever always listing the first page.

We need to pass a `MilestoneListOption` to the ListMilestones call. Ideally with a few things:

- We want to filter to "open" milestones.
- We want to pass the NextPage from the previous call, if there was one. 

https://github.com/google/go-github/blob/master/github/issues_milestones.go#L39-L53

caseydavenport: Looks like you have it correct in `prIssuesByRepo`, so just modelling after that would be good.

fixham/internal/release/releasenotes.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return prIssues, nil
+}
+
+// extractReleaseNoteFromIssue extracts release notes from an issue.
+// It looks for the release note block in the issue body and returns the content
+// between the start and end markers.
+func extractReleaseNoteFromIssue(issue *github.Issue) ([]string, error) {
+	body := issue.GetBody()
+	startMarker := "```release-note"
```
caseydavenport, line 91: The python script uses a regex match to find all release notes in the body, which I think is a more elegant and correct solution here. https://github.com/projectcalico/calico/blob/d3f5e2ad8d5655b9df1f4185e36e849b6273e14a/hack/release/generate-release-notes.py#L58-L68

The logic here isn't actually quite correct - it's possible for a single issue to have multiple distinct "release-note" section.

fixham/internal/release/releasenotes.go
```diff
@@ -0,0 +1,228 @@
...
+	startIndex += len(startMarker)
+	endIndex := strings.Index(body[startIndex:], endMarker)
+	if endIndex == -1 {
+		return nil, fmt.Errorf("end marker not found")
+	}
+	notes := strings.TrimSpace(body[startIndex : startIndex+endIndex])
+	if len(notes) == 0 {
+		return []string{issue.GetTitle()}, fmt.Errorf("no release notes found")
+	}
+	return strings.Split(notes, "\n"), nil
```
caseydavenport, line 106: it's possible for a single release note to span multiple lines, so I don't think we want to split them.

fixham/internal/release/releasenotes.go
```diff
@@ -0,0 +1,228 @@
...
+func extractReleaseNote(repo string, issues []*github.Issue) ([]*ReleaseNoteIssueData, error) {
+	issueDataList := []*ReleaseNoteIssueData{}
+	for _, issue := range issues {
+		notes, err := extractReleaseNoteFromIssue(issue)
+		if err != nil && len(notes) == 0 {
+			logrus.WithError(err).Errorf("Failed to extract release notes for issue %d", issue.GetNumber())
+			return nil, err
+		}
+		for _, note := range notes {
+			_note := strings.TrimSpace(note)
```
caseydavenport, line 119: you can just reassign `note` here.

```go
note = strings.TrimSpace(note)
```

caseydavenport: But if you wanted to use separate vars, a more idiomatic way would be to use `n` and `note` rather than `_note`. 

```go
for _, n := range notes {
  note := strings.TrimSpace(n)
```

fixham/internal/release/releasenotes.go
```diff
@@ -0,0 +1,228 @@
...
+			State:     "closed",
+			Labels:    []string{releaseNoteLabel},
+		}
+		logrus.WithField("repo", repo).Debug("Getting issues")
+		prIssues, err := prIssuesByRepo(githubClient, owner, repo, opts)
+		if err != nil {
+			logrus.WithError(err).Errorf("Failed to get issues for %s", repo)
+			return "", err
+		}
+		_issueDataList, err := extractReleaseNote(repo, prIssues)
```
caseydavenport, line 211: Another `_` var sneaking around

fixham/internal/release/releasenotes.go
```diff
@@ -0,0 +1,228 @@
...
+			return "", err
+		}
+		issueDataList = append(issueDataList, _issueDataList...)
+	}
+	if len(issueDataList) == 0 {
+		logrus.WithField("milestone", milestone).Error("No issues found for milestone")
+		return "", fmt.Errorf("no issues found for milestone %s", milestone)
+	}
+	releaseNoteFilePath := fmt.Sprintf("%s/%s/%s-release-notes.md", outputDir, releaseNotesFolder, releaseVersion.FormattedString())
+	if err := outputReleaseNotes(issueDataList, repoRootDir+"/"+releaseNoteTemplatePath, releaseNoteFilePath); err != nil {
```
caseydavenport, line 223: For building file paths, I'd recommend using `filepath.Join`

fixham/internal/release/releasenotes.go
caseydavenport: I think we want to remove the python script once we've replaced it in Golang also

fixham/internal/utils/files.go
```diff
@@ -0,0 +1,34 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// CreateDir creates a directory if it does not exist.
+func CreateDir(dir string) error {
```
caseydavenport, line 10: I don't think we need a helper for this. We can just do `os.MkdirAll()` - it already handles skipping directories that already exist.

fixham/pkg/tasks/hashrelease.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to get pinned version")
+	}
+	images := pinnedVersion.Components
+	images["operator"] = pinnedVersion.TigeraOperator
+	results := make(map[string]result, len(images))
+
+	for name, component := range images {
+		ch := make(chan result)
+		go imgExists(component.Image+":"+component.Version, component.Registry, ch)
```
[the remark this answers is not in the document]
caseydavenport: I see a lot of `string + "" + string` throughout. 

Generally it's easier to read using `fmt.Sprintf` except for the most basic of uses.

e.g.,

```go
fmt.Sprintf("%s:%s", component.Image, component.Version)
```

In this case, you could also add another bound function to the component though, since it's attached to a struct already anyway.

e.g.,

```go
func (c Component) ImageWithTag() string {
  return fmt.Sprintf("%s:%s", c.Image, c.Tag)
}
```

caseydavenport (commented):
I know this is still in progress, but I had a few more comments so figured I'd lay them out now to discuss.

fixham/internal/release/releasenotes.go
radTuti: i misinterpreted what the python script was trying to do, i will update

release/pkg/tasks/hashrelease.go
```diff
@@ -0,0 +1,174 @@
...
+	if err := hashrelease.PublishHashrelease(name, releaseHash, note, releaseVersion.Stream(), cfg.HashreleaseDir(), sshConfig); err != nil {
+		logrus.WithError(err).Fatal("Failed to publish hashrelease")
+	}
+}
+
+// HashreleaseCleanRemote cleans up old hashreleases on the docs host
+func HashreleaseCleanRemote(cfg *config.Config) {
+	sshConfig := command.NewSSHConfig(cfg.DocsHost, cfg.DocsUser, cfg.DocsKey, cfg.DocsPort)
+	logrus.Info("Cleaning up old hashreleases")
+	if err := hashrelease.DeleteOldHashreleases(sshConfig, -1); err != nil {
```
caseydavenport, line 170: Why do we pass -1 here just to have it defaulted to 400? 

This probably falls into the category of "Dead code making this more complicated than it needs to be".

We should either:
- Remove this argument from DeleteoldHashreleases, and always default to 400
- Remove the defaulting and require the caller pass in a value, and then pass in 400 here.

release/pkg/tasks/hashrelease.go
```diff
@@ -0,0 +1,174 @@
...
+	exists bool
+	err    error
+}
+
+func imgExists(name string, component hashrelease.Component, ch chan imageExistsResult) {
+	r := imageExistsResult{
+		name:  name,
+		image: component.String(),
+	}
+	exists, err := registry.ImageExists(component.Repository(), component.Registry)
```
caseydavenport, line 93: more of a nit, but you could do this:

```go
r.exists, r.err = registry.ImageExists(...)
```

radTuti:
Some outstanding issues that I am hoping to defer to a new PR

- Publish
  - Creating Operator manifest list
  - Using command runner for rsync

Some additional optimization

- Use types/reference for/ instead of the Image struct

release/internal/hashrelease/hashrelease.go
```diff
@@ -97,25 +97,63 @@ func ListHashreleases(sshConfig *command.SSHConfig) ([]hashrelease, error) {
...
+	library, err := getHashreleaseLibrary(sshConfig)
+	if err != nil {
+		return err
+	}
+	scanner := bufio.NewScanner(strings.NewReader(library))
+	var newLibrary []string
+	for scanner.Scan() {
+		line := scanner.Text()
+		for _, name := range hashreleaseNames {
+			if !strings.Contains(line, name) {
```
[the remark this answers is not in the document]
radTuti: I considered that but it will still result in making a diff because the content is in the format "<hash> - <note>" which we do have access to recreate

caseydavenport (approved):
Looks good!

Plenty more to do here but I think we should get this one merged so we can get incremental improvements on top of it.

I think the most important next steps are:
- Actually switching hashreleases over to using this logic for OSS, disabling the old system.
- Incorporating the hack/release logic into this tooling to build "real" releases as well.
- Sharing as much code as possible between hashrelease / release.

In addition, some other things that we should track as tasks:
- Fix that Calico clones operator, which then clones Calico again as part of gen-versions.
- Consider hosting hashreleases elsewhere - e.g., S3

Issue #8960, opened 2024-06-29
Pod in nat-outgoing should not be SNATed when it accesses local cluster hosts


wayne-cheng:
When I enable the setting `NatOutgoing`, I notice that  the pod's traffic is also SNATted when it accessing local cluster hosts.
I think it is unnecessary as it cause some performance degradation.

Below is the result of `tcpdump` capturing `ping` packets from the pod (`177.65.1.1`) to a cluster host (`192.168.1.83`), which confirms this behavior:
Pod (`177.65.1.1`) -> Host (`192.168.1.83`)
The pod (`177.65.1.1`):
```bash
$ tcpdump -vnn -i eth0 icmp
tcpdump: listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
09:27:48.148126 IP (tos 0x0, ttl 64, id 19509, offset 0, flags [DF], proto ICMP (1), length 84)
    177.65.1.1 > 192.168.1.83: ICMP echo request, id 840, seq 0, length 64
09:27:48.148452 IP (tos 0x0, ttl 63, id 49483, offset 0, flags [none], proto ICMP (1), length 84)
    192.168.1.83 > 177.65.1.1: ICMP echo reply, id 840, seq 0, length 64
09:27:49.149174 IP (tos 0x0, ttl 64, id 19588, offset 0, flags [DF], proto ICMP (1), length 84)
    177.65.1.1 > 192.168.1.83: ICMP echo request, id 840, seq 1, length 64
```

The host (`192.168.1.84`) where the pod deployed on:
> You can observe that the traffic is SNATed to the host ip `192.168.1.84` as it leave the host.
```bash
$ tcpdump -vnn -i enp1s0 icmp
tcpdump: listening on enp1s0, link-type EN10MB (Ethernet), capture size 262144 bytes
09:27:48.148208 IP (tos 0x0, ttl 63, id 19509, offset 0, flags [DF], proto ICMP (1), length 84)
    192.168.1.84 > 192.168.1.83: ICMP echo request, id 64813, seq 0, length 64
09:27:48.148420 IP (tos 0x0, ttl 64, id 49483, offset 0, flags [none], proto ICMP (1), length 84)
    192.168.1.83 > 192.168.1.84: ICMP echo reply, id 64813, seq 0, length 64
09:27:49.149229 IP (tos 0x0, ttl 63, id 19588, offset 0, flags [DF], proto ICMP (1), length 84)
[...]

PR #8961, opened 2024-06-29
Pod in nat-outgoing should not be SNATed when it accesses local cluster

wayne-cheng:
## Description

When `natOutgoing` is enabled in a IPPool, packets sent from Calico networked containers in this IP pool to destinations will be SNATed(msaqueraded).

However, we would prefer that the traffic accessing local cluster hosts should not be msaqueraded.

Provide a enum setting `natOutgoingExclusions`(default: `IPPoolsOnly`) in felix config.

Configure which type of destinations is excluded from being masqueraded.
- IPPoolsOnly: destinations outside of this IP pool will be masqueraded.
- IPPoolsAndHostIPs: destinations outside of this IP pool and all hosts will be masqueraded.

To maintain compatibility, this setting is not unconditionally enabled. We can consider switching its default value to `IPPoolsAndHostIPs` in the future version.

## Related issues/PRs

fixes linux part of #8960

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
New Felix config param natOutgoingExclusions allows for configuring which type of destinations is excluded from being masqueraded.
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
I think this is achievable already today by creating a disabled IP pool containing your host IP range which tells Calico these IPs aren't external to the cluster.

This is obviously not ideal for clusters where the host IPs might not fall cleanly into a small set of IP pools.

I do think a setting like this makes a bit more sense on FelixConfiguration than on IP pools - setting this for a single IP pool (at least based on this implementation) is going to enforce the setting on every IP pool with NATOutgoing set, so the config model is a little bit misleading.

@fasaxc WDYT?

tomastigera:
This is likely going to work for eBPF mode as well, however, it would be better to propagate the setting to somewhere [here](https://github.com/projectcalico/calico/blob/master/felix/bpf-gpl/tc.c#L470-L474)

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -436,6 +436,10 @@ type FelixConfigurationSpec struct {
 	// (ie it uses the iptables MASQUERADE target)
 	NATOutgoingAddress string `json:"natOutgoingAddress,omitempty"`
 
+	// When set to true and ip pool setting `natOutgoing` is true, packets sent from Calico networked containers in this pool
+	// to cluster host subnet will not be excluded from being masqueraded.  [Default: false]
+	DisableHostSubnetNATExclusion bool `json:"disableHostSubnetNATExclusion,omitempty"`
+
```
tomastigera, line 442: Too many negatives for me to make much sense out of this. Could you make it a bit easier to read? Perhaps rename it to just `HostSubnetSNATExclusion`. And expand in the PR description on the two behaviours?

wayne-cheng: I have made the adjustments. PTAL!

wayne-cheng:
> This is likely going to work for eBPF mode as well, however, it would be better to propagate the setting to somewhere [here](https://github.com/projectcalico/calico/blob/master/felix/bpf-gpl/tc.c#L470-L474)

In the related issue, I mentioned that Calico for Windows also needs such a setting. I think we can continue to improve it through additional PRs.

fasaxc:
Could we make this an enum; say:
```
NATOutgoingExclusions: IPPoolsOnly | IPPoolsAndHostSubnet 
```
then we can improve over time with new, better options.  For example, we could match on the all-hosts IP set to offer `IPPoolsAndHostIPs`.

felix/rules/nat.go
```diff
@@ -58,6 +58,11 @@ func (r *DefaultRuleRenderer) makeNATOutgoingRuleIPTables(ipVersion uint8, proto
 		SourceIPSet(masqIPsSetName).
 		NotDestIPSet(allIPsSetName)
 
+	if r.Config.HostSubnetNATExclusion {
```
fasaxc, line 61: You are using the all-host IP set so the config param name is misleading since it's not using the subnet at all.

wayne-cheng:
Hi, I updated the code and description. PTAL, thanks! @fasaxc

tomastigera:
@wayne-cheng sorry for completely dropping ball on this PR, I honestly thought that we merged it. I rebased your code here https://github.com/projectcalico/calico/pull/10275 but you are most welcome to rebase your PR if you want to drive it to the finish line (all the yaml conflicts get resolved by running `make generate` after the api conflicts are resolved) 

Any more comments @fasaxc ?

wayne-cheng:
OK, please retest. @tomastigera

tomastigera:
could you `git rm  manifests/ocp/crd.projectcalico.org_felixconfigurations.yaml` we no longer have that file

wayne-cheng:
> could you `git rm manifests/ocp/crd.projectcalico.org_felixconfigurations.yaml` we no longer have that file
Okay, it’s likely that the command `make generate` was executed when the code was incorrect. I’ve removed it now.

wayne-cheng:
> One silly issue, you need to change this value to 162 🤦‍♂️ https://github.com/wayne-cheng/calico/blob/fix-iptables-nat-host/libcalico-go/lib/backend/syncersv1/updateprocessors/configurationprocessor_test.go#L46

@tomastigera OK, please re-run the CI.

caseydavenport (approved):

tomastigera:
@wayne-cheng Thanks for the contribution!

PR #8983, opened 2024-07-04
BPF -  Fix route add/delete 

sridhartigera:
In BPF dataplane, we add routes to UDP services, so that host networked processes 
can reach them. This route is added/deleted in a loop forever because the routetable does not have a deviceRouteSrcAddress configured.

This PR fixes the issue by checking the source of the expected target against the route's source.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fixed continuous addition/deletion of service routes in eBPF mode.
```

## Reminder for the reviewer

Other optional labels:


fasaxc (commented):
LGTM, but would be good to have a test

fasaxc (approved):
LGTM, thanks for fleshing out the fix with UTs!

PR #8995, opened 2024-07-08
[nftables] Handle IP set members of unknown type

caseydavenport:
## Description

Changes to better handle IP sets with members of an unknown type. This is still an imperfect fix, but 
it does ensure that encountering IP sets of an unknown type won't disrupt the handling of IP sets that we do know about.

To fix this properly, we should enhance knftables to return set type information so that we can tell if a set type is one we understand or not.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/nftables/ipsets.go
+++ b/felix/nftables/ipsets.go
@@ -732,8 +732,10 @@ func (s *IPSets) ApplyDeletions() bool {
 
 	// ApplyDeletions() marks the end of the two-phase "apply". Piggyback on that to
 	// update the gauge that records how many IP sets we own.
+	s.gaugeNumSets.Set(float64(s.setNameToProgrammedMetadata.Dataplane().Len()))
+
+	// Determine if we need to be rescheduled.
 	numDeletionsPending := s.setNameToProgrammedMetadata.PendingDeletions().Len()
-	s.gaugeNumSets.Set(float64(numDeletionsPending))
 	if deletedSets.Len() == 0 {
 		// We had nothing to delete, or we only encountered errors, don't
 		// ask to be rescheduled.
--- a/felix/nftables/ipsets.go
+++ b/felix/nftables/ipsets.go
@@ -434,7 +434,7 @@ func (s *IPSets) tryResync() error {
 
 	// We expect a response for every set we asked for.
 	responses := make([]setData, len(sets))
-	for i := range sets {
+	for i := range responses {
 		setData := <-setsChan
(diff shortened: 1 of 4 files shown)


felix/nftables/ipsets.go
```diff
@@ -390,48 +390,79 @@ func (s *IPSets) tryResync() error {
...
-			}
-			setsChan <- setData{setName: name, elems: strElems}
+			setsChan <- setData{setName: name, elems: elems}
 		}(setName)
 	}
 
 	// We expect a response for every set we asked for.
 	responses := make([]setData, len(sets))
-	for range sets {
+	for i := range sets {
```
fasaxc, line 399:
```suggestion
	for i := range responses {
```
Ranging over the input `sets` hints to me that there's some relationship between `sets[i]` and `responses[i]` but that's not the case; responses will be in random order.

fasaxc (changes requested):

felix/nftables/ipsets.go
```diff
@@ -693,7 +722,7 @@ func (s *IPSets) ApplyDeletions() bool {
 
 	// ApplyDeletions() marks the end of the two-phase "apply". Piggyback on that to
 	// update the gauge that records how many IP sets we own.
-	numDeletionsPending := s.setNameToProgrammedMetadata.Dataplane().Len()
+	numDeletionsPending := s.setNameToProgrammedMetadata.PendingDeletions().Len()
```
caseydavenport, line 725: @fasaxc I noticed this line seems wrong, and that I copied it straight from the ipset implementation: https://github.com/projectcalico/calico/blob/ade2595a23c5b44cea5cea012545b1ef67eb7cee/felix/ipsets/ipsets.go#L963-L964

Is it actually just wrong there as well?

caseydavenport: I suppose it's a bit confusing, because it's called `numDeletionsPending` but it's setting the `gaugeNumSets` metric, and being used to return a value indicating whether or not additional deletions need to be made.

So I think we want to switch this to `PendingDeletions()` and then also update the gauge from `Dataplane().Len()` separately.

fasaxc: Yep, your version looks right, no idea what I did there.

fasaxc (approved):
LGTM

PR #8996, opened 2024-07-08
Limit to 100 concurrent set reads

caseydavenport:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/nftables/ipsets.go
+++ b/felix/nftables/ipsets.go
@@ -383,8 +383,15 @@ func (s *IPSets) tryResync() error {
 	setsChan := make(chan setData)
 	defer close(setsChan)
 
-	// Start a goroutine to list the elements of each set.
+	// Start a goroutine to list the elements of each set. Limit to 100 concurrent set reads to
+	// avoid spawning too many goroutines if there are a large number of sets.
+	routineLimit := make(chan struct{}, 100)
+	defer close(routineLimit)
 	for _, setName := range sets {
+		// Wait for room in the limiting channel.
+		routineLimit <- struct{}{}
+
+		// Start a goroutine to read this set.
 		go func(name string) {
 			elems, err := s.nft.ListElements(ctx, "set", name)
 			if err != nil {
@@ -401,6 +408,8 @@ func (s *IPSets) tryResync() error {
 				}
 			}
 			setsChan <- setData{setName: name, elems: strElems}
+			// Indicate that we're done by removing ourselves from the limiter channel.
+			<-routineLimit
 		}(setName)
 	}
 


felix/nftables/ipsets.go
```diff
@@ -401,6 +408,8 @@ func (s *IPSets) tryResync() error {
 				}
 			}
 			setsChan <- setData{setName: name, elems: strElems}
+			// Indicate that we're done by removing ourselves from the limiter channel.
+			<-routineLimit
```
fasaxc, line 412: Suggest putting this in a defer so that, if someone adds an early return later, the code will still be correct.

felix/nftables/ipsets.go
```diff
@@ -383,8 +383,15 @@ func (s *IPSets) tryResync() error {
...
+	routineLimit := make(chan struct{}, 100)
+	defer close(routineLimit)
 	for _, setName := range sets {
+		// Wait for room in the limiting channel.
+		routineLimit <- struct{}{}
+
+		// Start a goroutine to read this set.
 		go func(name string) {
 			elems, err := s.nft.ListElements(ctx, "set", name)
 			if err != nil {
```
fasaxc, line 397: Looks like you should return early in this error handling case?

felix/nftables/ipsets.go
```diff
@@ -383,8 +383,15 @@ func (s *IPSets) tryResync() error {
 	setsChan := make(chan setData)
 	defer close(setsChan)
 
-	// Start a goroutine to list the elements of each set.
+	// Start a goroutine to list the elements of each set. Limit to 100 concurrent set reads to
+	// avoid spawning too many goroutines if there are a large number of sets.
+	routineLimit := make(chan struct{}, 100)
```
fasaxc, line 388: I might have used `runtime.GOMAXPROCS()` to pick a limit that scales with CPU.

fasaxc (changes requested):

caseydavenport:
@fasaxc thanks for the quick review! I think this is ready for another look

fasaxc (approved):
LGTM.  Would be good to have a UT that checks the error path that had a bug?

caseydavenport:
It's a little bit tricky to add a UT to really capture that without a larger code refactor - I'll keep looking.

I did add related UT that actually caught a different bug, so that's nice.

felix/nftables/ipsets.go
caseydavenport: Still digging into this one - for some reason setting this to GOMAXPROCS caused a lot of tests to fail.

PR #9004, opened 2024-07-11
Add value sanity check and Len to DeltaTracker.

fasaxc:
## Description

Add a Len method to the Desired() side of the DeltaTracker.

Panic if trying to use a map in the DeltaTracker.  Found this was particularly hard to get right in another PR.

## Related issues/PRs

Enabling features for #8418

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/deltatracker/delta_tracker.go
+++ b/felix/deltatracker/delta_tracker.go
@@ -43,6 +43,16 @@ import (
 // keys that are in the dataplane map but not in the desired map.  The
 // pending maps are exposed via the IterPendingUpdates and IterPendingDeletions
 // methods.
+//
+// It's not recommended to use mutable objects in a DeltaTracker.  Updating a
+// mutable object while it is in the DeltaTracker will corrupt its internal
+// state.  By design, the DeltaTracker aliases the Desired and Dataplane values
+// if they happen to be equal. So, to safely mutate a value, you must
+// take a copy, mutate the copy and then re-insert the value.  Note: it is not
+// safe to delete the object from one side of the tracker, mutate it and
+// re-insert it.  Deleting an object that is in-sync from the Dataplane side
+// moves the object from the inDataplaneAndDesired map to the desiredUpdates
+// map, keeping a reference.
 type DeltaTracker[K comparable, V any] struct {
 	// To reduce occupancy, we treat the set of KVs in the dataplane and in the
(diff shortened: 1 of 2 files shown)


felix/deltatracker/delta_tracker.go
```diff
@@ -43,6 +43,16 @@ import (
 // keys that are in the dataplane map but not in the desired map.  The
 // pending maps are exposed via the IterPendingUpdates and IterPendingDeletions
 // methods.
+//
+// It's not recommended to use mutable objects in a DeltaTracker.  Updating a
```
nelljerram, line 47: Can you be more precise?  E.g. "It's not recommended to pass a mutable object such as <X> to the <Y> API(s)."

felix/deltatracker/delta_tracker.go
```diff
@@ -43,6 +43,16 @@ import (
...
+// mutable object while it is in the DeltaTracker will corrupt its internal
+// state.  By design, the DeltaTracker aliases the Desired and Dataplane values
+// if they happen to be equal. So, to safely mutate a value, you must
```
nelljerram, line 50: I think you're saying that the DeltaTracker doesn't perform copies (either shallow or deep).  Is the more important point that it doesn't copy the data when initially passed to it, or that it doesn't copy data when moving between Desired and Dataplane?  (Or both equally.)

felix/deltatracker/delta_tracker.go
```diff
@@ -43,6 +43,16 @@ import (
...
+// take a copy, mutate the copy and then re-insert the value.  Note: it is not
+// safe to delete the object from one side of the tracker, mutate it and
+// re-insert it.  Deleting an object that is in-sync from the Dataplane side
```
nelljerram, line 53: I don't follow this.  Does DeltaTracker provide an API for "delete the object from one side of the tracker"?  If it does, surely that ought to be coded safely?

nelljerram (commented):
I'm sorry to have taken a while before these initial questions, but I think I'm not properly understanding yet...

nelljerram (approved):

marvin-tigera:

PR #9011, opened 2024-07-15
Various cleanups

caseydavenport:
## Description

- Felix docker-image uses architecture-agnostic `ln` command.
- Move ipsets to new package, out of `dataplane/common`
- Be consistent in use of pointer vs. non-pointer is match structures.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/docker-image/Dockerfile
+++ b/felix/docker-image/Dockerfile
@@ -67,7 +67,7 @@ RUN update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 
 # Felix is built with RHEL/UBI and links against libpcap.so.1. We need this symbolic link
 # until Debian changes the soname from .0.8 to .1.
-RUN ln -s $(readlink /usr/lib/*-linux-gnu/libpcap.so.0.8) $(ldconfig -v 2>/dev/null | grep '^/lib/' | cut -f1 -d:)/libpcap.so.1
+RUN ln -s $(readlink /usr/lib/*-linux-gnu/libpcap.so.0.8) $(ldconfig -v 2>/dev/null | grep '^/lib/' | head -n 1 | cut -f1 -d:)/libpcap.so.1
 
 COPY felix.cfg /etc/calico/felix.cfg
 COPY calico-felix-wrapper /usr/bin
--- a/felix/docker-image/Dockerfile
+++ b/felix/docker-image/Dockerfile
@@ -67,7 +67,7 @@ RUN update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 
 # Felix is built with RHEL/UBI and links against libpcap.so.1. We need this symbolic link
 # until Debian changes the soname from .0.8 to .1.
-RUN ln -s $(readlink /usr/lib/*-linux-gnu/libpcap.so.0.8) /usr/lib/x86_64-linux-gnu/libpcap.so.1
+RUN ln -s $(readlink /usr/lib/*-linux-gnu/libpcap.so.0.8) lib$(ldconfig -v 2>/dev/null | grep -Po '/lib/(.*): ' | sed -e 's/: //')/libpcap.so.1
 
 COPY felix.cfg /etc/calico/felix.cfg
 COPY calico-felix-wrapper /usr/bin
--- a/felix/docker-image/Dockerfile
+++ b/felix/docker-image/Dockerfile
@@ -67,7 +67,7 @@ RUN update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 
 # Felix is built with RHEL/UBI and links against libpcap.so.1. We need this symbolic link
 # until Debian changes the soname from .0.8 to .1.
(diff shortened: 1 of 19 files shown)


nelljerram (changes requested):
Just some points about the ldconfig change.  Everything else looks great.

felix/docker-image/Dockerfile
```diff
@@ -67,7 +67,7 @@ RUN update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 
 # Felix is built with RHEL/UBI and links against libpcap.so.1. We need this symbolic link
 # until Debian changes the soname from .0.8 to .1.
-RUN ln -s $(readlink /usr/lib/*-linux-gnu/libpcap.so.0.8) /usr/lib/x86_64-linux-gnu/libpcap.so.1
+RUN ln -s $(readlink /usr/lib/*-linux-gnu/libpcap.so.0.8) lib$(ldconfig -v 2>/dev/null | grep -Po '/lib/(.*): ' | sed -e 's/: //')/libpcap.so.1
```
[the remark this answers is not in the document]
nelljerram: Just realized that my point (4) is bogus - I was thinking of awk.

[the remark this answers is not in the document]
caseydavenport: @nelljerram For context, this change is reverting to how this used to work (with only a small adjustment to account for the image being debian 12 instead of debian 11)

Here's the old incantation: https://github.com/projectcalico/calico/blob/release-v3.28/felix/docker-image/Dockerfile#L69

Here's the comment where @hjiawei requested I revert to this: https://github.com/tigera/calico-private/pull/7767#discussion_r1674833185

That said..

> Can it be right to create the symlink in the working directory, instead of in /usr/lib/x86_64-linux-gnu?

Correct - this was a typo on my part, good catch!

> Why the brackets around .* ?

Yeah, probably don't need those.

> What is the ldconfig intended to do? I think it's worth commenting this.

I believe it's just finding the correct path based on the arch. Not 100% sure, I didn't write this originally and I'm just tuning it up for the new base image.

> Doesn't entirely work for me here:

The output of ldconfig varies based on distro and version - the only reason I'm changing this is because the old debian 11 command no longer worked when I switched to deb 12 (needed for nftables work). I'm not too surprised this command doesn't work in your env, but I'd expect it to work if you ran inside of a debian 12 container.

That said, I agree the command you suggested looks much nicer, so I'll swap to that. Thank you!

felix/docker-image/Dockerfile
```diff
@@ -67,7 +67,7 @@ RUN update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 
 # Felix is built with RHEL/UBI and links against libpcap.so.1. We need this symbolic link
 # until Debian changes the soname from .0.8 to .1.
-RUN ln -s $(readlink /usr/lib/*-linux-gnu/libpcap.so.0.8) /usr/lib/x86_64-linux-gnu/libpcap.so.1
+RUN ln -s $(readlink /usr/lib/*-linux-gnu/libpcap.so.0.8) $(ldconfig -v 2>/dev/null | grep '^/lib/' | cut -f1 -d:)/libpcap.so.1
```
nelljerram, line 70: Thanks for the explanations, which I've reviewed.  I'm afraid I'm still unsure on a couple of points.

1. As it stands, `$(ldconfig ...)` will generate multiple lines (or words) of output, whereas I think we only want 1.  In the original incantation there was a `head -n 1` to trim down to just one output line.
2. I'm unclear which library directory we're trying to extract from the ldconfig output, and hence if the `^` is really wanted.  If the desired answer on some platforms is `/usr/lib/x86_64-linux-gnu`, then I was wrong to recommend the `^`.  Running on my laptop, the results without the `^` are:
```
/usr/lib/x86_64-linux-gnu/libfakeroot
/lib/i386-linux-gnu
/usr/local/lib
/lib/x86_64-linux-gnu
/lib
/lib/i386-linux-gnu/sse2
```
and with the `^`:
```
/lib/i386-linux-gnu
/lib/x86_64-linux-gnu
/lib
/lib/i386-linux-gnu/sse2
```

caseydavenport: @nelljerram My understanding is a bit limited, but I don't think it matters much whether it is in /usr/lib or /lib for our purposes. Won't both locations be searched when linking?

For the amd64 debian:12-slim image at least, there is no difference whether or not we restrict to `^/lib` or `/lib` - they both return a single element.

Agree `head -n 1` should be added back. I think in my initial testing of the new command I was only getting a single result anyway, so forgot to include it.

nelljerram: OK, so your thinking is that it doesn't matter where we create the link, so long as it's one of the places that ld.so is going to look in.  That makes sense.  Will approve now and leave any final tweaks up to you!

nelljerram (approved):

felix/docker-image/Dockerfile
caseydavenport: Thanks for the quick and thorough review, @nelljerram!

PR #9021, opened 2024-07-17
Increase default QPS

MichalFupso:
## Description
If no QPS is provided the inherited default is 5, we increase the QPS default to 50 for faster api request processing

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/k8s/k8s.go
+++ b/libcalico-go/lib/backend/k8s/k8s.go
@@ -347,9 +347,11 @@ func CreateKubernetesClientset(ca *apiconfig.CalicoAPIConfigSpec) (*rest.Config,
 	config.AcceptContentTypes = strings.Join([]string{runtime.ContentTypeProtobuf, runtime.ContentTypeJSON}, ",")
 	config.ContentType = runtime.ContentTypeProtobuf
 
-	// Overwrite the QPS if provided. Default QPS is 5.
+	// Overwrite the QPS if provided or set default QPS
 	if ca.K8sClientQPS != float32(0) {
 		config.QPS = ca.K8sClientQPS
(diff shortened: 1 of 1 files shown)


libcalico-go/lib/backend/k8s/k8s.go
```diff
@@ -347,9 +347,11 @@ func CreateKubernetesClientset(ca *apiconfig.CalicoAPIConfigSpec) (*rest.Config,
 	config.AcceptContentTypes = strings.Join([]string{runtime.ContentTypeProtobuf, runtime.ContentTypeJSON}, ",")
 	config.ContentType = runtime.ContentTypeProtobuf
 
-	// Overwrite the QPS if provided. Default QPS is 5.
+	// Overwrite the QPS if provided or set default QPS
 	if ca.K8sClientQPS != float32(0) {
 		config.QPS = ca.K8sClientQPS
+	} else {
```
caseydavenport, line 353: I think this is probably fine, but I also wonder if it makes sense to do this within the apiserver code that creates the client so that the blast radius is slightly reduced?

This will impact all users of libcalico-go, whereas if we only want to impact the Calico API server we could set this here: https://github.com/caseydavenport/calico/blob/fcbecc23d8fc9a852a4db9941612e3a880ad687e/apiserver/pkg/storage/calico/resource.go#L86-L107

caseydavenport: Up to you though, happy to approve as-is since it seems likely to be harmless.

caseydavenport (approved):

Issue #8831, opened 2024-05-16
Calico node error - iptables-legacy-save command failed


Farhanec07:
## Current Behavior
```$ kubectl get pods -n kube-system | grep calico
calico-kube-controllers-6fb59668cc-k746l   0/1     CrashLoopBackOff   38 (4m22s ago)   120m
calico-node-lzwd5                          0/1     Running            0                117m
calico-node-m5pdk                          0/1     Running            11 (72s ago)     117m
calico-node-zngzj                          1/1     Running            1 (51m ago)      117m
calico-typha-5f49b8b8c4-hmspz              1/1     Running            0                120m
calico-typha-5f49b8b8c4-kksv4              1/1     Running            0                120m
```

panic which i observed that its failing to save iptables rules causing pods to crash.
calico-node pod log -
```2024-05-15 17:52:10.964 [PANIC][61396] felix/table.go 784: iptables-legacy-save command failed after retries ipVersion=0x4 table="raw"
panic: (*logrus.Entry) 0xc0005ed420
2024-05-15 17:52:12.344 [WARNING][61470] felix/table.go 840: iptables save failed error=exit status 127
2024-05-15 17:52:12.949 [PANIC][61470] felix/table.go 784: iptables-legacy-save command failed after retries ipVersion=0x4 table="raw"
panic: (*logrus.Entry) 0xc000720000

goroutine 172 [running]:
github.com/sirupsen/logrus.(*Entry).log(0xc0003ada40, 0x0, {0xc00034d640, 0x31}) ...


2024-05-15 17:52:14.307 [WARNING][61545] felix/table.go 840: iptables save failed error=exit status 127
2024-05-15 17:52:14.307 [WARNING][61545] felix/table.go 778: iptables-legacy-save command failed error=exit status 127 ipVersion=0x4 stderr="" table="raw"
2024-05-15 17:52:14.309 [WARNING][61545] felix/table.go 840: iptables save failed error=exit status 127
2024-05-15 17:52:14.310 [WARNING][61545] felix/table.go 778: iptables-legacy-save command failed error=exit status 127 ipVersion=0x4 stderr="" table="nat"
2024-05-15 17:52:14.312 [WARNING][61545] felix/table.go 840: iptables save failed error=exit status 127
2024-05-15 17:52:14.312 [WARNING][61545] felix/table.go 778: iptables-legacy-save command failed error=exit status 127 ipVersion=0x4 stderr="" table="mangle"
[...]

PR #9022, opened 2024-07-18
[CORE-10546] Change rpm install order to fix 'undefined symbol: xtables_strdup'

coutinhop:
`microdnf install iproute-tc nftables` would install `iptables-libs` 1.8.5-11.el8_9 as a dependency, and it provides `/lib64/libxtables.so.12.3.0`. When using that `.so` file with `LD_PRELOAD`, we can consistently reproduce the issue:

```
# LD_PRELOAD=/lib64/libxtables.so.12.3.0 iptables-legacy-save
iptables-legacy-save: symbol lookup error: iptables-legacy-save: undefined symbol: xtables_strdup
```

The problem was, during our build process, we build iptables (and iptables-libs) v1.8.8 and install with `rpm --force`, but that doesn´t remove the outdated v1.8.5. By moving things around in the Dockerfiles and installing iptables-libs 1.8.8 before iproute-tc and nftables, the problematic 1.8.5 package is no longer installed and thus the correct `.so` file should be used (`libxtables.so.12.6.0` at the time of writing this).

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/8831 (tentatively, pending user confirmation)

## Todos

- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Fix 'undefined symbol: xtables_strdup' error when running 'iptables-legacy-save' in the calico-node image.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/Dockerfile.amd64
+++ b/node/Dockerfile.amd64
@@ -119,26 +119,27 @@ RUN microdnf install \
     # Needed for runit startup script
     which \
     # Needed for the cleanup script
-    findutils \
-    # nftables
-    nftables
+    findutils
 
 # Since the ubi repos do not contain all the packages we need (they're missing conntrack-tools),
 # we're using AlmaLinux repos for missing packages.
 COPY almalinux.repo /etc/yum.repos.d/almalinux.repo
 
 RUN microdnf --enablerepo=baseos install \
-    iproute-tc \
     # Needed for conntrack
     libnetfilter_cthelper libnetfilter_cttimeout libnetfilter_queue \
     conntrack-tools
 
-# Install iptables via rpms. The libs must be force installed because the iptables source RPM has the release
-# version '1.8.8-6.el8' conflicts with iptables-libs (pulled in by the iputils package) '1.8.5-11.el8_9'.
-RUN rpm --force -i /tmp/rpms/iptables-libs-${IPTABLES_VER}.el8.x86_64.rpm && \
-    rpm --force -i /tmp/rpms/iptables-legacy-libs-${IPTABLES_VER}.el8.2.x86_64.rpm && \
-    # Install compatible libnftnl version with selected iptables version
-    rpm -i /tmp/rpms/iptables-legacy-${IPTABLES_VER}.el8.2.x86_64.rpm && \
+# Install iptables-libs via rpm. The libs must installed before installing iproute-tc and nftables via 'microdnf install'
+# otherwise they will pull an outdated version of iptables-libs (1.8.5-11.el8_9) as a dependency.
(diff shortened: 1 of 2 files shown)


jonathan-hurley:
I can confirm that this resolves the issue with `xtables_strdup`.

node/Dockerfile.amd64
```diff
@@ -119,26 +119,27 @@ RUN microdnf install \
...
     conntrack-tools
 
-# Install iptables via rpms. The libs must be force installed because the iptables source RPM has the release
-# version '1.8.8-6.el8' conflicts with iptables-libs (pulled in by the iputils package) '1.8.5-11.el8_9'.
-RUN rpm --force -i /tmp/rpms/iptables-libs-${IPTABLES_VER}.el8.x86_64.rpm && \
-    rpm --force -i /tmp/rpms/iptables-legacy-libs-${IPTABLES_VER}.el8.2.x86_64.rpm && \
-    # Install compatible libnftnl version with selected iptables version
-    rpm -i /tmp/rpms/iptables-legacy-${IPTABLES_VER}.el8.2.x86_64.rpm && \
+# Install iptables-libs via rpm. The libs must installed before installing iproute-tc and nftables via 'microdnf install'
+# otherwise they will pull an outdated version of iptables-libs (1.8.5-11.el8_9) as a dependency.
```
hjiawei, line 134: `1.8.5-11.el8_9` version number may change in the future. This comment can get outdated soon-ish. Maybe we can remove the version? How do we handle future cases when `iptables-libs` from the system repository is newer than what we build (1.8.8)?

hjiawei (approved):

node/Dockerfile.arm64
```diff
@@ -129,27 +129,27 @@ RUN microdnf install \
...
+# Install iptables-libs via rpm. The libs must installed before installing iproute-tc and nftables via 'microdnf install'
+# otherwise they will pull a different version of iptables-libs as a dependency.
+RUN rpm -i /tmp/rpms/iptables-libs-${IPTABLES_VER}.el8.aarch64.rpm && \
+    rpm -i /tmp/rpms/iptables-legacy-libs-${IPTABLES_VER}.el8.2.aarch64.rpm
+
+# iproute-tc and nftables depend on iptables-libs and should be installed after it.
+RUN microdnf --enablerepo=baseos install iproute-tc nftables
+
+# Install iptables via rpm. Install compatible libnftnl version with the selected iptables version
+RUN rpm --force -i /tmp/rpms/libnftnl-${LIBNFTNL_VER}.el8.aarch64.rpm && \
```
hjiawei, line 152: remove `--force` ?

PR #9054, opened 2024-07-24
Build resource lists from registered resources

caseydavenport:
## Description

Fix calicoctl command descriptions to include the correct resource lists
based on the full set of registered resources.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix missing resources in calioctl command help text
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/calicoctl/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/calicoctl/resourcemgr/resourcemgr.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"os"
 	"reflect"
-	"sort"
 	"strings"
 	"time"
 
@@ -114,8 +113,8 @@ var (
 	allKinds  = []string{}
 )
 
+// ValidResources returns all registered resource kinds.
 func ValidResources() []string {
-	sort.Strings(allKinds)
 	return allKinds
 }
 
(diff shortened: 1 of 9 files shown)


calicoctl/calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -100,19 +103,26 @@ type resourceHelper struct {
...
+var (
+	helpers   map[schema.GroupVersionKind]resourceHelper
+	kindToRes = make(map[string]ResourceObject)
+	allKinds  = []string{}
+)
+
+func ValidResources() []string {
+	sort.Strings(allKinds)
+	return allKinds
+}
```
MichalFupso, line 120: What do you think about adding ValidResourcesList(), that returns the already sorted and formatted list

`
func ValidResourcesList() string {
  kinds := sort.String(allKinds)
	resourceList := ""
	for _, r := range kinds {
		resourceList += fmt.Sprintf("    - %s\n", r)
	}
	return resourceList
}
`
Then if we ever want to change the format, we can do it in one place. Maybe a better place for this would be https://github.com/projectcalico/calico/blob/master/calicoctl/calicoctl/util/docstring.go

MichalFupso (approved):

PR #9064, opened 2024-07-29
Allow user-defined rules in nftables mode

caseydavenport:
## Description

To simulate the behavior provided in iptables mode, this PR enables user-defined rules to be inserted between Calico's "insert" and "append" rules.

For example, to deny traffic that Calico would have otherwise allowed.

It does so by splitting our nftables hooks into two parts: an insert chain, and an append chain with a slightly higher priority.

Users can then define their own table, hooked between the two chains, to define their own rules.

Note that with this PR, Calico "drop" rules still take priority regardless of any user intevention. Additional changes would be needed to enable this, as nftables drop rules override any previous or subsequent allow actions in other tables.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/nftables/table.go
+++ b/felix/nftables/table.go
@@ -110,8 +110,8 @@ func init() {
 	prometheus.MustRegister(gaugeNumChains)
 	prometheus.MustRegister(gaugeNumRules)
 
-	// Calico uses a single nftables with a variety of hooks.
-	// The top level base chains are laid out below.
+	// Configure the base hook chains needed by Calico. These are split into two parts - the chains we insert into,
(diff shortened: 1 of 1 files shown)


caseydavenport:
Going to leave this open for comment for now, but still pending concrete use-cases for this. 

Nftables inherently gives users opportunities to interleave rules before or after ours using their own table - the question is "do users need the ability to put rules between our mark chains and our accept chains"?

Issue #8897, opened 2024-06-10
Failed to attach bpf program to pod interface


xander-sh:
## Context ##
After upgrading calico to 3.27.2 version in clusters with PostgreSQL managed by Patroni and depoyed by zalando operator, we started to see errors in the pods with DB. 
```bash
2024-06-01T07:40:03+03:00  2024-06-01 04:40:03 UTC [35341]: [1-1] 665aa623.8a0d 0 [unknown] [unknown] [unknown] 10.222.59.107 LOG:  connection received: host=10.222.59.107 port=35708
2024-06-01T07:40:03+03:00 2024-06-01 04:40:03 UTC [35341]: [2-1] 665aa623.8a0d 0 airflow adm [unknown] 10.222.59.107 LOG:  connection authenticated: identity="adm" method=md5 (/home/postgres/pgdata/pgroot/data/pg_hba.conf:12)
2024-06-01T07:40:03+03:00 2024-06-01 04:40:03 UTC [35341]: [3-1] 665aa623.8a0d 0 airflow adm [unknown] 10.222.59.107 LOG:  connection authorized: user=adm database=airflow SSL enabled (protocol=TLSv1.3, cipher=TLS_AES_256_GCM_SHA384, bits=256)
2024-06-01T07:40:03+03:00 2024-06-01 04:40:03 UTC [35341]: [4-1] 665aa623.8a0d 0 airflow adm [unknown] 10.222.59.107 LOG:  disconnection: session time: 0:00:00.008 user=adm database=airflow host=10.222.59.107 port=35708
2024-06-01T07:40:03+03:00 2024-06-01 04:40:03 UTC [35342]: [1-1] 665aa623.8a0e 0 [unknown] [unknown] [unknown] 10.222.20.175 LOG:  connection received: host=10.222.20.175 port=42608
2024-06-01T07:40:03+03:00 2024-06-01 04:40:03 UTC [35342]: [2-1] 665aa623.8a0e 0 airflow adm [unknown] 10.222.20.175 LOG:  connection authenticated: identity="adm" method=md5 (/home/postgres/pgdata/pgroot/data/pg_hba.conf:12)
[...]

PR #9074, opened 2024-07-30
[v3.27] Use the same priority when re-attaching preamble program to interface.

sridhartigera:
## Description

This PR fixes the following issues
1. When the preamble program is attached repeatedly to an interface, we end up using a lot of tc priority and run into a kernel error. Fix - For the first program, use the kernel allocated priority. For subsequent attaches, we use the priority allocated by the kernel. When doing this, we attach to the same priority but the handle keeps toggling between 0x1 and 0x2.
2. When there is a workload update such as update to the labels, annotations, preamble programs are re-attached. Fix - When we first attach to an interface, we store the qdisc info. For subsequent workload updates, we use the stored qdisc info to decide whether to re-attach. This fix makes sure we don't overwrite the qdisc with invalid info.

fixes https://github.com/projectcalico/calico/issues/8897

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
ebpf: Fixed frequently attaching BPF programs when pods annotations/labels change and eventually failing due ro running out of  tc priority.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/ut/attach_test.go
+++ b/felix/bpf/ut/attach_test.go
@@ -514,6 +514,141 @@ func TestAttach(t *testing.T) {
 	})
 }
 
+// This test simulates workload updates like changing labels, annotations.
+// Expectation is that multiple workload updates should still result in
+// preamble program not getting re-attached.
+func TestAttachWithMultipleWorkloadUpdate(t *testing.T) {
+	RegisterTestingT(t)
+
+	bpfmaps, err := bpfmap.CreateBPFMaps(4)
+	Expect(err).NotTo(HaveOccurred())
+
+	programs := bpfmaps.ProgramsMap.(*hook.ProgramsMap)
+	loglevel := "off"
+
+	bpfEpMgr, err := linux.NewTestEpMgr(
+		&linux.Config{
+			Hostname:              "uthost",
+			BPFLogLevel:           loglevel,
+			BPFDataIfacePattern:   regexp.MustCompile("^hostep[12]"),
+			VXLANMTU:              1000,
+			VXLANPort:             1234,
+			BPFNodePortDSREnabled: false,
+			RulesConfig: rules.Config{
+				EndpointToHostAction: "RETURN",
+			},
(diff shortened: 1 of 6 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1897,6 +1901,9 @@ func (m *bpfEndpointManager) wepApplyPolicyToDirection(readiness ifaceReadiness,
 	if readiness != ifaceIsReady {
 		res, err := m.wepAttachProgram(ap)
 		if err != nil {
+			qdisc.valid = false
+			qdisc.prio = 0
+			qdisc.handle = 0
 			return qdisc, fmt.Errorf("attaching program to wep: %w", err)
```
tomastigera, line 1907: Sort of the way to return a "nil" value for errors. The issue here might be that the returned value is not ignored when error is returned. Otherwise it should not matter which "undefined" value we return.

```suggestion
			return qDiscInfo{}, fmt.Errorf("attaching program to wep: %w", err)
```

felix/bpf/ut/attach_test.go
```diff
@@ -514,6 +514,141 @@ func TestAttach(t *testing.T) {
...
+			Id: &proto.WorkloadEndpointID{
+				OrchestratorId: "k8s",
+				WorkloadId:     "workloadep1",
+				EndpointId:     "workloadep1",
+			},
+			Endpoint: &proto.WorkloadEndpoint{Name: "workloadep1"},
+		})
+		err = bpfEpMgr.CompleteDeferredWork()
+		Expect(err).NotTo(HaveOccurred())
+		valid, prio, handle := bpfEpMgr.GetIfaceQDiscInfo("workloadep1")
```
tomastigera, line 602: a comment here that the fact that handle never changes means that attach is not called so that we know what the correct condition is.

tomastigera (changes requested):
some nits but lgtm!

tomastigera (approved):

PR #9082, opened 2024-07-31
Remove the unnecessary check of non-calico ipsets in resynIPSet

mazdakn:
## Description
`resyncIPSet` is only called for Calico IP sets. There is no need to check the same condition inside it.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/ipsets/ipsets.go
+++ b/felix/ipsets/ipsets.go
@@ -559,7 +559,9 @@ func (s *IPSets) resyncIPSet(ipSetName string) error {
 				// Start of a Members entry, following this, there'll be one member per
 				// line then EOF or a blank line.
 
-				// Look up to see if this is one of our none temporary IP sets.
+				// Optimisation: skip parsing temporary IP set members.
+				// We only need to track their metadata to make sure they 
+				// are deleted.
 				if s.IPVersionConfig.IsTempIPSetName(ipSetName) {
 					if debug {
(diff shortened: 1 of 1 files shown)


felix/ipsets/ipsets.go
```diff
@@ -559,8 +559,8 @@ func (s *IPSets) resyncIPSet(ipSetName string) error {
 				// Start of a Members entry, following this, there'll be one member per
 				// line then EOF or a blank line.
 
-				// Look up to see if this is one of our IP sets.
-				if !s.IPVersionConfig.OwnsIPSet(ipSetName) || s.IPVersionConfig.IsTempIPSetName(ipSetName) {
+				// Look up to see if this is one of our none temporary IP sets.
```
fasaxc, line 562:
```suggestion
				// Optimisation: skip parsing temporary IP set members.
				// We only need to track their metadata to make sure they 
				// are deleted.
```

fasaxc (changes requested):
One suggestion, comment wasn't very useful from before.

Could go even further and skip _listing_ them.  We only need to know that they exist, we don't really need to know the rest of their metadata, I think.

fasaxc (approved):

PR #9084, opened 2024-07-31
Fix host IP updates in eBPF

sridhartigera:
## Description
This PR fixes 2 issues related to host IP change.
1. Fix parsing of host IP - hostIP updates can have addresses as ```ip``` or ```ip/mask```. Need to make sure either way the update is parsed correctly.
2. When there is a host IP, mark all interfaces ```not ready``` so that bpf programs can be re-attached with updated priority.

## Todos

- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
ebpf - Fix parsing host IP update and re-attach program on all interfaces when there is a host IP update.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -768,7 +768,11 @@ func (m *bpfEndpointManager) withIface(ifaceName string, fn func(iface *bpfInter
 	m.dirtyIfaceNames.Add(ifaceName)
 }
 
-func (m *bpfEndpointManager) updateHostIP(ip net.IP, ipFamily int) {
+func (m *bpfEndpointManager) updateHostIP(ipAddr string, ipFamily int) {
+	ip, _, err := net.ParseCIDR(ipAddr)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -778,11 +782,14 @@ func (m *bpfEndpointManager) updateHostIP(ip net.IP, ipFamily int) {
 		// Should be safe without the lock since there shouldn't be any active background threads
 		// but taking it now makes us robust to refactoring.
 		m.ifacesLock.Lock()
+		defer m.ifacesLock.Unlock()
```
tomastigera, line 785: the point of no doing defer was not to host the lock across the rest of the function. But it may not matter since this is probably not called in parallel with anything, is it?

tomastigera (approved):

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: I will better remove the defer and put the unlock, back where it was.

PR #9089, opened 2024-08-01
[master] Auto pick #9088: Use the same priority when re-attaching preamble program

sridhartigera:
Cherry pick of #9088 on master.

#9088: Use the same priority when re-attaching preamble program

# Original PR Body below

Cherry pick of #9074 on release-v3.28.

#9074: Use the same priority when re-attaching preamble program

# Original PR Body below

## Description

This PR fixes the following issues
1. When the preamble program is attached repeatedly to an interface, we end up using a lot of tc priority and run into a kernel error. Fix - For the first program, use the kernel allocated priority. For subsequent attaches, we use the priority allocated by the kernel. When doing this, we attach to the same priority but the handle keeps toggling between 0x1 and 0x2.
2. When there is a workload update such as update to the labels, annotations, preamble programs are re-attached. Fix - When we first attach to an interface, we store the qdisc info. For subsequent workload updates, we use the stored qdisc info to decide whether to re-attach. This fix makes sure we don't overwrite the qdisc with invalid info.

fixes https://github.com/projectcalico/calico/issues/8897

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
ebpf: Fixed frequently attaching BPF programs when pods annotations/labels change and eventually failing due ro running out of  tc priority.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/libbpf/libbpf.go
+++ b/felix/bpf/libbpf/libbpf.go
@@ -167,7 +167,7 @@ func DetachClassifier(ifindex, handle, pref int, ingress bool) error {
 }
 
 // AttachClassifier return the program id and pref and handle of the qdisc
-func (o *Obj) AttachClassifier(secName, ifName string, ingress bool) (int, int, int, error) {
+func (o *Obj) AttachClassifier(secName, ifName string, ingress bool, prio int) (int, int, int, error) {
 	cSecName := C.CString(secName)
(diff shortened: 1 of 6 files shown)


tomastigera (approved):
LGTM, thanks!

Issue #8860, opened 2024-05-28
Connection issue for a multiple zone cluster with calico 3.28.0


lzhecheng:
## Expected Behavior

A Node can reach a service whose endpoint is on another Node (different zone) immediately.

## Current Behavior

A Node cannot reach a service whose endpoint is on another Node (different zone) immediately. The first packet is dropped and the second one works.

## Possible Solution

Use calico 3.27.3

## Steps to Reproduce (for bugs)

1. Create a **multiple zone** cluster
2. Create a service with endpoints on nodes of different zones
3. wget from a Node to the service
4. If the endpoint is not on the same node, the first packet is lost

## Context

Details here: https://github.com/kubernetes-sigs/cloud-provider-azure/issues/6293

## Your Environment

* Calico version
* Orchestrator version (e.g. kubernetes, mesos, rkt):
* Operating System and version:
* Link to your project (optional):

matthewdupre:
Sounds like a regression, we'll have a look

sfudeus:
I might have a similar issue and was just about to open a bugreport. For me, this is related to VXLAN checksum offloading. 
For vxlan-tunneled traffic, the first SYN is lost and has to be resent. 
Please advise if I should add my data here or create a dedicated issue. 

@lzhecheng Can you try with disabling checksum offloading (`featureDetectOverride: ChecksumOffloadBroken=true` in FelixConfiguration) to see if it makes a difference for you?

sfudeus:
I'm adding the basics of what I am observing here: 
With VXLAN checksum offloading enabled, I do observe the following: 
For any traffic, which is 
* directed against a NodePort or an external/loadBalancerIP
* forwarded to a pod on a different subnet(i.e. requiring vxlan, likely happening always when using `vxlanMode: Always` instead of `CrossSubnet`)
the first SYN packet is lost.

I could not observe packet loss when directing the traffic against the podIP itself, only via K8s iptables rules for NodePort/LoadBalancer, likely because of NAT?

I observed the packet loss to happen only on the destination node, between the physical interface and the pod interface, i.e. I could still see the first SYN packet VXLAN encapsulated on the physical interface, but only the second SYN popped up on the pod interface (cali*). 

My test client was only running in the hostNetwork, I didn't test from a pod (yet).

matthewdupre:
[...]

PR #9091, opened 2024-08-01
Execute MASQ rules after kube-proxy to avoid conflict - reenable vxlan csum offload

tomastigera:
## Description

    Calico needs to do MASQ for several reasons and kube-proxy does it when
    it forwards nodeports to another node. Calico conflicts with this MASQ
    in case it routes traffic to the destionation via tunnel. This may
    result in a conflict which triggers wrongly calculated checksums in case
    tx csum offloading is enabled on the tunnel device, which lead to
    disabling offloading on the tunnel and thus to significant performance
    hit.
    
    Do MASQ after kube-proxy means that if kube-proxy already did it,
    calico's is not executed - it would be a duplicate anyway. Other MASQ
    and SNAT cases in the calico chain are othogonal to the kube-proxy
    usecase and thus execute as kube-proxy does not do MASQ in that case.
    
    Note that unlik ein other chains where calico may do policy and thus
    needs to go first, calico does not execute any policy in POSTROUTING
    nat.

fixes https://github.com/projectcalico/calico/issues/8860

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/int_dataplane.go
+++ b/felix/dataplane/linux/int_dataplane.go
@@ -1774,7 +1774,17 @@ func (d *InternalDataplane) setUpIptablesNormal() {
 			Match:  d.newMatch(),
 			Action: d.actions.Jump(rules.ChainNATPrerouting),
 		}})
-		t.InsertOrAppendRules("POSTROUTING", []generictables.Rule{{
+		// We must go last not to conflict with kube-proxy MASQ rules. All we do
+		// is MASQ anyway, but if kube-proxy set a mark to do MASQ, we must let it to
+		// clean it up so not to execute some rules after say a tunnel (vxlan)
+		// device - which should not happen.
+		//
+		// MASQ is a terminating target so once one is hit, no other rule gets
+		// executed.
+		//
+		// N.B. ChainFIPSnat does not do MASQ, but does not collide with k8s
+		// service, namely nodeports.
+		t.AppendRules("POSTROUTING", []generictables.Rule{{
 			Match:  d.newMatch(),
 			Action: d.actions.Jump(rules.ChainNATPostrouting),
 		}})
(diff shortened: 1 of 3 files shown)


felix/dataplane/linux/int_dataplane.go
```diff
@@ -1774,7 +1774,17 @@ func (d *InternalDataplane) setUpIptablesNormal() {
...
+		// We must go last not to conflict with kube-proxy MASQ rules. All we do
+		// is MASQ anyway, but if kube-proxy set a mark to do MASQ, we must let it to
+		// clean it up so not to execute some rules after say a tunnel (vxlan)
+		// device - which should not happen.
+		//
+		// MASQ is a terminating target so once one is hit, no other rule gets
+		// executed.
+		//
+		// N.B. ChainFIPSnat does not do MASQ, but does not collide with k8s
+		// service, namely nodeports.
```
fasaxc, line 1786:
```suggestion
		// We must go last to avoid a conflict if both kube-proxy and Calico 
		// decide to MASQ the traffic.  
		//
		// This is because kube-proxy uses a mark bit to trigger its MASQ and 
		// we need the mark bit to get cleared by kube-proxy's chain.  If we 
		// go first, our MASQ rule terminates further processing, and the 
		// mark bit remains set on the packet.  
		// 
		// Leaving the mark set on the packet is a problem when the packet 
		// gets encapped because the mark is copied to the outer encap packet.
		// The outer packet then gets MASQed by kube-proxy's rule.  In turn,
		// that MASQ triggers a checksum offload bug in the kernel resulting
		// in corrupted packets.
		//
		// N.B. ChainFIPSnat does not do MASQ, but does not collide with k8s
		// service, namely nodeports.
```

fasaxc (changes requested):
I had a go at describing the issue in full in the comment.  What about nftables mode; how do we interact with kube-proxy in that case?

fasaxc (approved):

Issue #8376, opened 2023-12-26
BGPFilter: match on prefix length


defo89:
## Expected Behavior
In certain scenarios there might be a need to allow/deny prefixes with variable prefix length.

Example: Allow accepting prefixes within `77.0.0.0/16` but not `/32` host routes, or more complex case: do not accept routes with prefix length greater than or equal `/20` and less than or equal `/24`.

Another use case could be related to https://github.com/projectcalico/calico/issues/8374 (not yet sure if behaviour in the issue is intended): allow prefix advertisement within ClusterCIDR range but disable announcement of `/32` prefixes.

If I understand [bird filters](https://bird.network.cz/doc/bird-5.html) correctly, that would look like this:
```
  exportV4:
    - action: Reject
      matchOperator: In
      cidr: 100.64.0.0/15{32}
```

From the above link:
```
Sets of prefixes are special: their literals does not allow ranges, but allows prefix patterns that are written as ipaddress/pxlen{low,high}. Prefix ip1/len1 matches prefix pattern ip2/len2{l,h} if the first min(len1, len2) bits of ip1 and ip2 are identical and len1 <= ip1 <= len2. A valid prefix pattern has to satisfy low <= high, but pxlen is not constrained by low or high. Obviously, a prefix matches a prefix set literal if it matches any prefix pattern in the prefix set literal.

There are also two shorthands for prefix patterns: address/len+ is a shorthand for address/len{len,maxlen} (where maxlen is 32 for IPv4 and 128 for IPv6), that means network prefix address/len and all its subnets. address/len- is a shorthand for address/len{0,len}, that means network prefix address/len and all its supernets (network prefixes that contain it).

For example, [ 1.0.0.0/8, 2.0.0.0/8+, 3.0.0.0/8-, 4.0.0.0/8{16,24} ] matches prefix 1.0.0.0/8, all subprefixes of 2.0.0.0/8, all superprefixes of 3.0.0.0/8 and prefixes 4.X.X.X whose prefix length is 16 to 24. [ 0.0.0.0/0{20,24} ] matches all prefixes (regardless of IP address) whose prefix length is 20 to 24, [ 1.2.3.4/32- ] matches any prefix that contains IP address 1.2.3.4. 1.2.0.0/16 ~ [ 1.0.0.0/8{15,17} ] is true, but 1.0.0.0/16 ~ [ 1.0.0.0/8- ] is false.

[...]

PR #9114, opened 2024-08-08
BGPFilter: Add prefixLength to BGPFilter to match on min/max prefix length

mstansberry:
## Description
- Add `prefixLength` to BGPFilter struct for more granular matching
- Allows for advertisement of more or less specific addresses from a calico prefix without needing to accept or reject all
- Tested locally in kind as well as unit tests

## Related issues/PRs
fixes https://github.com/projectcalico/calico/issues/8376
documents https://github.com/tigera/docs/pull/1592

## Todos

- [X] Tests
- [X] Documentation
- [X] Release note

## Release Note

```release-note
BGPFilter: Add prefix length matching
```

## Reminder for the reviewer

Other optional labels:


bsalmicw (approved):
LGTM

confd/pkg/resource/template/template_funcs.go
```diff
@@ -111,12 +111,35 @@ var (
 	}
 )
 
-func filterMatchCIDR(cidr string, operator v3.BGPFilterMatchOperator) (string, error) {
+func filterMatchPrefixLength(cidr string, prefixLength v3.BGPFilterPrefixLength) string {
+	cidrIP, cidrNet, _ := net.ParseCIDR(cidr)
```
caseydavenport, line 115: I think it would be worth at least including a warning log for the error here instead of ignoring it entirely. 

I know we _shouldn't_ hit this since the CIDR should be pre-validated elsewhere, but best to play it safe.

confd/pkg/resource/template/template_funcs_test.go
```diff
@@ -58,7 +58,7 @@ func Test_BGPFilterBIRDFuncs(t *testing.T) {
 	testFilter.Spec = v3.BGPFilterSpec{
 		ImportV4: []v3.BGPFilterRuleV4{
 			{Action: "Accept", Source: "RemotePeers", Interface: "vxlan.calico", MatchOperator: "NotIn", CIDR: "55.4.0.0/16"},
-			{Action: "Reject", Source: "RemotePeers", MatchOperator: "NotIn", CIDR: "55.4.0.0/16"},
+			{Action: "Reject", Source: "RemotePeers", MatchOperator: "NotIn", CIDR: "55.4.0.0/16", PrefixLength: v3.BGPFilterPrefixLength{Min: 16, Max: 24}},
```
caseydavenport, line 61: Instead of replacing these test cases, can we copy/paste and add a second test case that includes the PrefixLength? That way we maintain coverage for both with and without the PrefixLength field set.

caseydavenport: I think we should also consider a test case for specifying a PrefixLength without a CIDR.

From what I can tell, it looks like the API will accept that input configuration but the code won't do anything with it. 

If this is a valid user input, then we should add a test here and make sure it puts out the right output. Otherwise, we should look at API validation to ensure that if a PrefixLength is given, so is a CIDR.

caseydavenport (commented):
Thanks for picking this up @mstansberry !

It's looking pretty good - most of my comments are just API conventions type stuff to make sure we have the right optionality set, and to ensure users can't provide invalid input on the API.

confd/pkg/resource/template/template_funcs.go
mstansberry: Added a print here, not sure if we want to bail out and only return `cidr` with no prefixLength, error, etc

confd/pkg/resource/template/template_funcs_test.go
```diff
@@ -59,8 +59,10 @@ func Test_BGPFilterBIRDFuncs(t *testing.T) {
 		ImportV4: []v3.BGPFilterRuleV4{
 			{Action: "Accept", Source: "RemotePeers", Interface: "vxlan.calico", MatchOperator: "NotIn", CIDR: "55.4.0.0/16"},
 			{Action: "Reject", Source: "RemotePeers", MatchOperator: "NotIn", CIDR: "55.4.0.0/16"},
+			{Action: "Reject", Source: "RemotePeers", MatchOperator: "NotIn", CIDR: "55.4.0.0/16", PrefixLength: &v3.BGPFilterPrefixLengthV4{Min: int32Helper(16), Max: int32Helper(24)}},
 			{Action: "Reject", Interface: "eth0", MatchOperator: "NotIn", CIDR: "55.4.0.0/16"},
 			{Action: "Accept", Interface: "eth0", Source: "RemotePeers"},
+			{Action: "Reject", Interface: "eth0", Source: "RemotePeers", PrefixLength: &v3.BGPFilterPrefixLengthV4{Min: int32Helper(16), Max: int32Helper(24)}},
```
caseydavenport, line 65: If I'm reading this correctly, including a PrefixLength has no impact on the resulting filter if there is no CIDR.

I think we should add a check to this function: https://github.com/projectcalico/calico/blob/master/libcalico-go/lib/validator/v3/validator.go#L1396-L1415

That returns an error if PrefixLength is specified by CIDR is not. Best not to allow configurations that don't have any effect in order to avoid user confusion.

api/pkg/apis/projectcalico/v3/bgpfilter.go
```diff
@@ -86,6 +90,16 @@ type BGPFilterRuleV6 struct {
 	Action BGPFilterAction `json:"action" validate:"required,filterAction"`
 }
 
+type BGPFilterPrefixLengthV4 struct {
+	Min *int32 `json:"min,omitempty" validate:"omitempty,bgpFilterPrefixLengthMinV4"`
```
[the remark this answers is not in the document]
mstansberry: should we just use the kubebuilder validations vs the regex matches I added in the validate code? or would you prefer we stick with both?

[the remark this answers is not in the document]
caseydavenport: Ah sorry, I missed that.

Personally I like having both. the kubebuilder validations will ensure that even users who happen to bypass our API server will have this validation enforced, which is nice.

the validation in v3 allows us to do things that kubebuilder doesn't

mstansberry:
Just following up to see if there are any additional changes to make here? Thanks a lot.

caseydavenport (approved):

marvin-tigera:

mstansberry:
squashed based on the label applied, looks like that might have been the wrong move

caseydavenport:
@mstansberry whoops! Sorry, that label is for the bot

caseydavenport:
Just allows me to run CI, apply the label, and the bot will merge once it passes

PR #9125, opened 2024-08-12
Fail FV tests on core dumps and races.

fasaxc:
## Description

We've had detection for races and core dumps for a long time but we weren't failing tests as a result.  This made the detection not very useful: when we did get a race/core dump there was no associated log because logs are hidden unless the test fails.  

* Make tests fail in `AfterEach()` if a race/corefile is detected.  This causes Ginkgo to dump logs, making it much easier to debug the race/panic.
* Make it clear what the result of the test was to avoid confusion between "test failed" and "a race was hit during the test".
* Stop dumping core for test-workload and other test binaries.  Some panic on failure so we were getting lots of core files for no reason!
* Fix race caused by configuring logrus to disable its lock.  Struggled to find a better fix.  The problem is that Felix starts some threads before it does that configuration, and then can log before the lock gets disabled.  It's only safe to disable the lock once our logging configuration is in place.

## Related issues/PRs

CORE-10585

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/int_dataplane.go
+++ b/felix/dataplane/linux/int_dataplane.go
@@ -2303,13 +2303,14 @@ func (d *InternalDataplane) apply() {
 	// Update the routing table in parallel with the other updates.  We'll wait for it to finish
 	// before we return.
 	var routesWG sync.WaitGroup
+	var numBackgroundProblems atomic.Uint64
 	for _, r := range d.routeTableSyncers() {
 		routesWG.Add(1)
 		go func(r routetable.SyncerInterface) {
 			err := r.Apply()
 			if err != nil {
 				log.Warn("Failed to synchronize routing table, will retry...")
-				d.dataplaneNeedsSync = true
+				numBackgroundProblems.Add(1)
 			}
 			d.reportHealth()
 			routesWG.Done()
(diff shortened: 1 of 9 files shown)


felix/logutils/logutils.go
```diff
@@ -140,10 +140,6 @@ func ConfigureLogging(configParams *config.Config) {
 	// hook above to fan out logs to multiple destinations.
 	log.SetOutput(&logutils.NullWriter{})
 
-	// Since we push our logs onto a second thread via a channel, we can disable the
```
fasaxc, line 143: This was causing a race in Felix that's hard to fix because the lock is inside logrus.  I don't think the lock is very heavily contended so it's probably no big deal to re-enable it

felix/dataplane/linux/int_dataplane.go
```diff
@@ -2303,13 +2303,14 @@ func (d *InternalDataplane) apply() {
...
 	// before we return.
 	var routesWG sync.WaitGroup
+	var numBackgroundProblems atomic.Uint64
 	for _, r := range d.routeTableSyncers() {
 		routesWG.Add(1)
 		go func(r routetable.SyncerInterface) {
 			err := r.Apply()
 			if err != nil {
 				log.Warn("Failed to synchronize routing table, will retry...")
-				d.dataplaneNeedsSync = true
```
fasaxc, line 2312: Race detector detects that this variable is written from multiple goroutines.

song-jiang (approved):
LGTM

PR #9132, opened 2024-08-13
Attach xdp to slave devices of a bond

sridhartigera:
## Description

This PR has the changes to attach XDP to bond slave devices when a donottrack policy is attached to a HEP.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
ebpf - Attach XDP to bond slave devices.
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -2550,37 +2568,58 @@ func (d *bpfEndpointManagerDataplane) applyPolicyToWeps(
...
+	attachXDPOnly bool,
+) (*tc.AttachPoint, *tc.AttachPoint, *xdp.AttachPoint, error) {
+	xdpAttachPoint := *apxdp
+
+	var parallelWG sync.WaitGroup
+	var xdpAP *xdp.AttachPoint
+	var ingressAP, egressAP *tc.AttachPoint
+	var xdpErr, tcErr error
+
+	parallelWG.Add(1)
```
tomastigera, line 2609: woudn't it be easier to handle it in the original function, where you would decide whether to add the xdp goroutine in the waitgroup if you start it or not? Now you may have a routine that only starts to finish the wait (confusing) and does nothing and then you have a second waitgroup ...

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4100,6 +4146,25 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	val, ok := m.nameToIface[masterIfName]
+	m.ifacesLock.Unlock()
+
+	if !ok || val.info.ifaceType != IfaceTypeBond {
+		return []string{}
+	}
+
+	masterIfIndex := val.info.ifIndex
+	slaveDevices := []string{}
+	for name, iface := range m.nameToIface {
```
tomastigera, line 4160: wouldn't it be easier to somehow remember the slaves in the master? When you have a large deploynment with zillions of workload ifaces, you iterate through all of them just to figure out is there are some slaves for this device. You could have the slice of slaves in the master and just return that.

felix/fv/donottrack_test.go
```diff
@@ -38,6 +39,196 @@ import (
...
+		}
+	})
+
+	AfterEach(func() {
+		tc.Stop()
+		infra.Stop()
+		externalClient.Stop()
+	})
+
+	Context("after adding eth0 to a bond", func() {
```
tomastigera, line 113: woudn't it be easier to create  `Context("with bond", ...)` inside the current Describe to save all this copy-pasta and just do the BeforeEach to setup the bond?

felix/fv/donottrack_test.go
```diff
@@ -38,6 +39,196 @@ import (
...
+			host1Selector := fmt.Sprintf("name == 'bond0-%s'", tc.Felixes[1].Name)
+
+			By("Having connectivity after installing bidirectional policies")
+			host0Pol := createDonotTrackPolicy("host-0-pol", host0Selector, host1Selector, client, ctx)
+			_ = createDonotTrackPolicy("host-1-pol", host1Selector, host0Selector, client, ctx)
+
+			Consistently(xdpProgramAttached(tc.Felixes[0], "bond0"), "2s", "1s").Should(BeFalse())
+			Consistently(xdpProgramAttached(tc.Felixes[1], "bond0"), "2s", "1s").Should(BeFalse())
+			Eventually(xdpProgramAttached(tc.Felixes[0], "eth0"), "10s", "1s").Should(BeTrue())
+			Eventually(xdpProgramAttached(tc.Felixes[1], "eth0"), "10s", "1s").Should(BeTrue())
```
tomastigera, line 177: In a way this is the only thing that differs from the non-bond tests. Can't you just place the test in a function and parametrize it whether it is with or w/o bond? Connectivity should always be the same, but where the programs are attached is different.

tomastigera (commented):
looks good in general, but have some nits around the code/test structure

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: I feel handling doing this here is easier to handle.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1296,8 +1334,11 @@ func (m *bpfEndpointManager) markEndpointsDirty(ids set.Set[any], kind string) {
 					}
 				}
 			} else {
-				log.Debugf("Mark host iface dirty, for host %v change", kind)
+				log.Debugf("Mark host iface dirty %v, for host %v change", id, kind)
 				m.dirtyIfaceNames.Add(id)
+				for _, slaveIface := range m.getBondSlaves(id) {
```
tomastigera, line 1339: nit: If you returned the Set you could iterate over the set without creating the slice. In this case it does not matter much since there are going to be a handful of slaves, but in general, I'd advise to watch for this.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -3081,6 +3158,9 @@ func (m *bpfEndpointManager) OnHEPUpdate(hostIfaceToEpMap map[string]proto.HostE
 				delete(m.hostIfaceToEpMap, ifaceName)
 			}
 			m.dirtyIfaceNames.Add(ifaceName)
+			for _, slaveIface := range m.getBondSlaves(ifaceName) {
```
tomastigera, line 3161: this should be done only fr bonds

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -3095,6 +3175,10 @@ func (m *bpfEndpointManager) OnHEPUpdate(hostIfaceToEpMap map[string]proto.HostE
 		m.addHEPToIndexes(ifaceName, &newEp)
 		m.hostIfaceToEpMap[ifaceName] = newEp
 		m.dirtyIfaceNames.Add(ifaceName)
+		for _, slaveIface := range m.getBondSlaves(ifaceName) {
```
tomastigera, line 3178: ditto

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4061,6 +4145,15 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
 	return IfaceTypeData
 }
 
+func (m *bpfEndpointManager) getBondSlaves(masterIfName string) []string {
+	slaveDevices := []string{}
```
tomastigera, line 4149: you could assume that if this is called only for bonds there are usually at least 2 slaves. Perhaps 4, you could preallocate capacity for that many and that would make the append a bit faster. But you may not need to allocate it at all and just iterate over the set.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4061,6 +4145,15 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+
+	if val, ok := m.hostIfaceToSlaveDevices[masterIfName]; ok {
+		slaveDevices = append(slaveDevices, val.Slice()...)
```
tomastigera, line 4152: you do not need to append to the empty slice above, just return this slice, right? And return nil if there is no set :thinking:

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -2527,37 +2583,58 @@ func (d *bpfEndpointManagerDataplane) applyPolicyToWeps(
...
 	egressAttachPoint := *ap
-	xdpAttachPoint := *apxdp
 
+	var ingressErr, egressErr error
 	var parallelWG sync.WaitGroup
 	var ingressAP, egressAP *tc.AttachPoint
-	var xdpAP *xdp.AttachPoint
-	var ingressErr, egressErr, xdpErr error
-
-	parallelWG.Add(2)
```
tomastigera, line 2546: cann't you just have `parallelWG.Add(1)` infron of each goroutine here and the other (xdp) covered by `if ifaceType != IfaceTypeBond {` without creating the other new function which only groups attaching TC? They that function waits for both tc and then for xdp, while here it waited for tc and xdp at the same time :thinking:

tomastigera (changes requested):

felix/fv/donottrack_test.go
```diff
@@ -159,6 +161,98 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ do-not-track policy tests;
...
+		}
+
+		host0Selector := fmt.Sprintf("name == '%s-%s'", iface, tc.Felixes[0].Name)
+		host1Selector := fmt.Sprintf("name == '%s-%s'", iface, tc.Felixes[1].Name)
+
+		By("Having connectivity after installing bidirectional policies")
+		host0Pol := createDonotTrackPolicy("host-0-pol", host0Selector, host1Selector, client, ctx)
+		_ = createDonotTrackPolicy("host-1-pol", host1Selector, host0Selector, client, ctx)
+
+		if iface == "bond0" {
```
tomastigera, line 185: should this get BPF only guard?

tomastigera (approved):

felix/fv/donottrack_test.go
sridhartigera: Running the bond test only for BPF.

PR #9135, opened 2024-08-14
[Felix] Add retries when netlink list APIs encounter EINTR.

sridhartigera:
## Description

Add retries to netlink List APIs as it now returns EINTR. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/int_dataplane.go
+++ b/felix/dataplane/linux/int_dataplane.go
@@ -64,6 +64,7 @@ import (
 	"github.com/projectcalico/calico/felix/jitter"
 	"github.com/projectcalico/calico/felix/labelindex"
 	"github.com/projectcalico/calico/felix/logutils"
+	"github.com/projectcalico/calico/felix/netlinkshim"
 	"github.com/projectcalico/calico/felix/nftables"
 	"github.com/projectcalico/calico/felix/proto"
 	"github.com/projectcalico/calico/felix/routerule"
@@ -1106,7 +1107,14 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 // findHostMTU auto-detects the smallest host interface MTU.
 func findHostMTU(matchRegex *regexp.Regexp) (int, error) {
 	// Find all the interfaces on the host.
-	links, err := netlink.LinkList()
+
+	nlHandle, err := netlinkshim.NewRealNetlink()
+	if err != nil {
+		log.WithError(err).Error("failed to created netlink handle. Unable to auto-detect MTU")
+		return 0, err
+	}
+
+	links, err := nlHandle.LinkList()
 	if err != nil {
 		log.WithError(err).Error("Failed to list interfaces. Unable to auto-detect MTU")
 		return 0, err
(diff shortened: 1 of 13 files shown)


felix/dataplane/linux/int_dataplane.go
```diff
@@ -1106,7 +1107,14 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 // findHostMTU auto-detects the smallest host interface MTU.
 func findHostMTU(matchRegex *regexp.Regexp) (int, error) {
 	// Find all the interfaces on the host.
-	links, err := netlink.LinkList()
+
+	nlHandle, err := netlinkshim.NewRealNetlink()
+	if err != nil {
+		log.WithError(err).Error("failed to created netlink handle. Unable to auto-detect MTU")
```
fasaxc, line 1113:
```suggestion
		log.WithError(err).Error("Failed to create netlink handle. Unable to auto-detect MTU.")
```

felix/fv/etcd_restart_test.go
```diff
@@ -60,7 +58,7 @@ var _ = Context("etcd connection interruption", func() {
 		// Wait until the tunl0 device appears; it is created when felix inserts the ipip module
 		// into the kernel.
 		Eventually(func() error {
-			links, err := netlink.LinkList()
+			links, err := utils.LinkList()
```
fasaxc, line 61: `utils` is best avoided, it ends up being a catchall and you could have 100 utils packages with different random stuff in them.

Can we emphasize the retry somehow; important for the user to know that this is not he same as the basic version.

`netlinkretry.LinkList()`
`netlinkutils.LinkListRetryEINTR()`

felix/netlinkshim/netlink.go
```diff
@@ -47,6 +49,156 @@ type Interface interface {
...
 }
 
+type RealNetlink struct {
+	nlHandle *netlink.Handle
+}
+
 func NewRealNetlink() (Interface, error) {
-	return netlink.NewHandle(syscall.NETLINK_ROUTE)
+	nlHandle, err := netlink.NewHandle(syscall.NETLINK_ROUTE)
+	return &RealNetlink{
```
fasaxc, line 58: Prefer to return `nil, err` if there's an error.  Makes incorrect use fail quicker.

felix/netlinkshim/netlink.go
```diff
@@ -47,6 +49,156 @@ type Interface interface {
...
+func (r *RealNetlink) RuleAdd(rule *netlink.Rule) error {
+	return r.nlHandle.RuleAdd(rule)
+}
+
+func (r *RealNetlink) RuleDel(rule *netlink.Rule) error {
+	return r.nlHandle.RuleDel(rule)
+}
+
+func (r *RealNetlink) Delete() {
+	//nolint:staticcheck
```
fasaxc, line 176: What's the linter's problem?

fasaxc (changes requested):
Pragmatic to re-use the existing shim.  Saves proliferation.  Just need to remember that the shim is needed now and that it has function in it rather than being a trivial shim

felix/netlinkshim/netlink.go
sridhartigera: ```netlink.Delete``` is deprecated and I got the below error from the linter. 
```SA1019: r.nlHandle.Delete is deprecated: use Close instead which is in line with typical resource release patterns for files and other resources.```

fasaxc: Delete just calls through to Close so let's just switch to using Close?

fasaxc (approved):

PR #9137, opened 2024-08-15
[BPF] fix applying and removing log filters

tomastigera:
* do not apply an empty filter (matches all) if there is no filter for that device
* do not apply any filters if there the mode is not debug
* do not pre-allocate filter jumps indexes before deciding whether we will apply filters (hence do not take up space if no filter is applied)
* clean up state when a filter is removed. Do not incorrectly program the index in tc program config if there is no filter left.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/ut/attach_test.go
+++ b/felix/bpf/ut/attach_test.go
@@ -815,6 +815,97 @@ func TestRepeatedAttach(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


sridhartigera (approved):
LGTM. Can you please update release notes.

PR #8979, opened 2024-07-04
Rev netlink to pick up RouteListIter

fasaxc:
## Description

I made a PR to the upstream netlink repo that was just merged.  It changes the library's internal read loop to use an iterator, which avoid a lot of allocations (10% speedup + lower allocs in my benchmarks).  There's also a new, more efficient API but I'll tackle that once the dust settles on #8418 .

## Related issues/PRs

https://github.com/vishvananda/netlink/pull/940

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Felix now uses less CPU and memory when listing routes from the kernel.
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #9139, opened 2024-08-15
Switch to netlink.RouteListFilteredIter() API

fasaxc:
## Description

Optimise RouteTable resync methods:
* Add benchmark for resync with various numbers of routes. 
* Switch to netlink.RouteListFilteredIter() API, which avoids a series of large slice allocations.
* Use benchmark to find more wins:

  * Use  deltatracker.WithValuesEqualFn, DeepEquals is the default and it was much slower.
  * Pre-allocate the seenKeys set, saves the expense of copying it multiple times as it grows.
  * Copy the netlink Route once per loop to a scratch variable to avoid escape to the heap.

Overall, significantly reduces allocations from when there are many routes to resync and significantly increases performance.

Benchmarks for resyncing 64k routes; 50% reduction in resync time; 80% saving in bytes allocated; GC-included CPU usage down by 60%.
```
v3.28:

BenchmarkResync65536-12    	       7	 156498055 ns/op	 242857143 ncpu/op	203556697 B/op	  790473 allocs/op

master:

BenchmarkResync65536-12    	       7	 143550698 ns/op	 270000000 ncpu/op	208147700 B/op	  724999 allocs/op

This PR: 

BenchmarkResync65536-12    	      13	  84051862 ns/op	 108461538 ncpu/op	36912350 B/op	  526080 allocs/op
```
## Related issues/PRs

Follow on from #8979 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Felix's route resync logic has been optimised; it now uses 50% less CPU time and 80% less memory.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/routetable/bench_test.go
+++ b/felix/routetable/bench_test.go
@@ -1,0 +1,129 @@
+// Copyright (c) 2024 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package routetable_test
+
+import (
+	"fmt"
+	"math/rand"
+	"os"
+	"runtime"
+	"testing"
+	"time"
+
+	. "github.com/onsi/gomega"
+	"github.com/prometheus/procfs"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+
+	"github.com/projectcalico/calico/felix/dataplane/linux/dataplanedefs"
+	"github.com/projectcalico/calico/felix/fv/utils"
+	"github.com/projectcalico/calico/felix/ip"
+	"github.com/projectcalico/calico/felix/logutils"
+	mocknetlink "github.com/projectcalico/calico/felix/netlinkshim/mocknetlink"
+	. "github.com/projectcalico/calico/felix/routetable"
+	"github.com/projectcalico/calico/felix/routetable/ownershippol"
+)
+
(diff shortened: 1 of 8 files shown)


felix/routetable/bench_test.go
```diff
@@ -0,0 +1,129 @@
...
+			rt.RouteUpdate(RouteClassLocalWorkload, ifaceName, Target{
+				CIDR: ip.MustParseCIDROrIP(fmt.Sprintf("10.0.%d.%d/32", i, j)),
+			})
+			n++
+			if n == numRoutes {
+				break outer
+			}
+		}
+	}
+	if n < numRoutes {
```
aaaaaaaalex, line 101: should there be error handling in the above RouteUpdate, such that the loop has an alternative break condition?
To me it seems this condition will always be false currently, unless a test is set up with more than 65535 `numRoutes`

felix/routetable/route_table.go
```diff
@@ -781,31 +789,51 @@ func (r *RouteTable) doFullResync(nl netlinkshim.Interface) error {
...
+	var err error
+	seenKeys := set.NewSize[kernelRouteKey](r.kernelRoutes.Dataplane().Len())
+	for attempt := 0; attempt < routeListFilterAttempts; attempt++ {
+		// Using the Iter version here saves allocating a large slice of netlink.Route,
+		// which we immediately discard.
+		var scratchRoute netlink.Route
+		err = nl.RouteListFilteredIter(r.netlinkFamily, routeFilter, routeFilterFlags, func(route netlink.Route) bool {
+			// This copy avoids an alloc per iteration.  We leak scratchRoute
+			// once but avoid leaking the function parameter.
+			scratchRoute = route
```
aaaaaaaalex, line 802: :muscle:

aaaaaaaalex (approved):
Just one potential change, might be nothing

felix/routetable/bench_test.go
fasaxc: This is here to prevent someone from adding `BenchmarkResync1000000` without fixing this bit of the code to actually be capable of generating that many.

fasaxc:
Just holding this back until I've got the RouteTable merge to Enterprise done.  Might want to upstream some of the fixes there and I don't want this to conflict...

PR #9140, opened 2024-08-16
Integrate release and hashrelease logic

caseydavenport:
## Description

Note that this DOES change the format of our hashreleases in order to better match the structure of artifacts produced by a real release. Notably, existing hashrelease structure looks like this:

```
files/
  windows/
   calico-windows-<version.zip
manifests/
release-notes/
tigera-operator.tgz
metadata.yaml
```

The new format looks more like this:

```
manifests/
release-notes/
tigera-operator-<version>.tgz
calico-windows-<version>.zip
metadata.yaml
install-calico-windows.ps1
ocp.tgz
SHA256SUMS
```

The main differences being:

- Some additional files, e.g., SHA256SUMS
- No "files/windows" directory structure.
- tigera-operator.tgz -> tigera-operator-<version>.tgz

To handle this, there is some additional logic to copy files into the "legacy" structure so that both formats coexist. Once our CI tools are updated to use the new format, we can remove this redundancy. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


release/pkg/tasks/release.go
```diff
@@ -13,10 +14,7 @@ import (
 
 // ReleaseNotes generates release notes for the current release.
 func ReleaseNotes(cfg *config.Config) {
-	outDir := cfg.RepoRootDir
-	if cfg.IsHashrelease {
-		outDir = cfg.HashreleaseDir()
-	}
+	outDir := filepath.Join(cfg.RepoRootDir, "release", "_output", "hashrelease")
```
radTuti, line 17: this also need to generate release notes for official release

release/pkg/tasks/operator.go
```diff
@@ -14,13 +14,13 @@ import (
...
 func OperatorHashreleaseBuild(runner *registry.DockerRunner, cfg *config.Config) {
-	outputDir := cfg.OutputDir
+	outputDir := cfg.TmpFolderPath()
 	operatorDir := cfg.OperatorConfig.Dir
 	componentsVersionPath, err := hashrelease.GenerateComponentsVersionFile(outputDir)
 	if err != nil {
 		logrus.WithError(err).Fatal("Failed to generate components.yaml")
 	}
-	operatorComponent, err := hashrelease.RetrievePinnedOperator(outputDir)
+	operatorComponent, err := hashrelease.RetrievePinnedOperator(cfg.TmpFolderPath())
```
radTuti, line 23: this change isn't needed since outputDir is already set to the config

release/pkg/tasks/hashrelease.go
```diff
@@ -190,30 +157,30 @@ func HashreleaseValidate(cfg *config.Config, sendImagestoISS bool) {
...
 	if err != nil {
 		logrus.WithError(err).Fatal("Failed to get candidate name")
 	}
-	operatorVersion, err := hashrelease.RetrievePinnedOperatorVersion(outputDir)
+	operatorVersion, err := hashrelease.RetrievePinnedOperatorVersion(cfg.TmpFolderPath())
 	if err != nil {
 		logrus.WithError(err).Fatal("Failed to get operator version")
 	}
-	releaseHash, err := hashrelease.RetrievePinnedVersionHash(outputDir)
+	releaseHash, err := hashrelease.RetrievePinnedVersionHash(cfg.TmpFolderPath())
```
radTuti, line 183: this change isn't needed since outputDir is set to cfg.TmpFolderPath()

release/pkg/tasks/hashrelease.go
```diff
@@ -104,25 +72,24 @@ func imgExists(name string, component hashrelease.Component, ch chan imageExists
 // These images are checked to ensure they exist in the registry
 // as they should have been pushed in the standard build process.
 func HashreleaseValidate(cfg *config.Config, sendImagestoISS bool) {
-	outputDir := cfg.OutputDir
-	name, err := hashrelease.RetrieveReleaseName(outputDir)
+	name, err := hashrelease.RetrieveReleaseName(cfg.TmpFolderPath())
```
radTuti, line 75:
```suggestion
	outputDir := cfg.TmpFolderPath()
	name, err := hashrelease.RetrieveReleaseName(outputDir)
```

release/pkg/tasks/hashrelease.go
```diff
@@ -104,25 +72,24 @@ func imgExists(name string, component hashrelease.Component, ch chan imageExists
...
 		logrus.WithError(err).Fatal("Failed to get candidate name")
 	}
 	parsedProductVersion := version.Version(productVersion)
-	operatorVersion, err := hashrelease.RetrievePinnedOperatorVersion(outputDir)
+	operatorVersion, err := hashrelease.RetrievePinnedOperatorVersion(cfg.TmpFolderPath())
 	if err != nil {
 		logrus.WithError(err).Fatal("Failed to get operator version")
 	}
-	images, err := hashrelease.RetrieveComponentsToValidate(outputDir)
+	images, err := hashrelease.RetrieveComponentsToValidate(cfg.TmpFolderPath())
```
radTuti, line 92: this change shouldn't be needed if outputDir is set to cfg.TmpFolderPath()

radTuti (commented):
I do not think this should change the output of hashrelease currently since that will affect testing as is

release/internal/hashrelease/pinnedversion.go
```diff
@@ -203,6 +203,7 @@ func RetrievePinnedOperator(outputDir string) (OperatorComponent, error) {
 }
 
 // RetrievePinnedOperatorVersion retrieves the operator version from the pinned version file.
+// TODO: We should get this from metadata.yaml instead of pinned_versions.yaml.
```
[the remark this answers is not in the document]
caseydavenport: Yeah, probably not. I'll remove this TODO.

I'd like "pinned_versions.yaml" to go away though. It's a bit of a hack to tie all of this together, but we should be able to determine the operator version and product version in step 1 and pass it down to the places that need it, rather than treating `pinned_versions.yaml` as the source of truth.

release/pkg/tasks/release.go
caseydavenport: At the moment it's only used for hashreleases. I think we must have lost the hook into the release-notes target in the last PR.

radTuti: there's a release-notes target in root Makefile that calls release-notes target in release/Makefile and uses this function

caseydavenport: Yeah, I pushed a commit yesterday that fixed this - that make target was calling a CLI option that didn't exist any more.

release/internal/outputs/release.go
radTuti, line 1: looks like this is an empty file now, delete pls

release/internal/version/version.go
```diff
@@ -24,52 +50,114 @@ func (v *Version) String() string {
...
+	return fmt.Sprintf("%s-%s", releaseBranchPrefix, v.Stream())
+}
+
+// IsDevVersion returns true if the version includes the dev marker.
+// The dev marker can be used in a tag to indicate what the next release version will be. If present,
+// it means the version is of the form vX.Y.Z-<devTag>-<commitsSinceTag>-g<commitHash>, where vX.Y.Z
+// is the _next_ (currently unreleased) product version.
+func IsDevVersion(ver, devTag string) bool {
+	v := Version(ver)
+	pattern := fmt.Sprintf(`^v\d+\.\d+\.\d+(-\d+\.\d+)?-%s-\d+-g[0-9a-f]{12}(-dirty)?$`, devTag)
```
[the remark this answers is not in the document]
caseydavenport: Why should this return true if the version string doesn't include the `devTag`? I thought the entire purpose of this function was to tell if the dev tag was present.

That said, this function is currently unused so I think it can be removed?

release/build/main.go
```diff
@@ -67,46 +115,101 @@ func main() {
...
 
 		// The publish command is used to publish a locally built hashrelease to the hashrelease server.
 		{
 			Name:  "publish",
-			Usage: "Publish hashrelease from output/ to hashrelease server",
+			Usage: "Publish hashrelease from _output/ to hashrelease server",
 			Flags: []cli.Flag{
 				&cli.BoolFlag{Name: skipValidationFlag, Usage: "Skip pre-build validation", Value: false},
 				&cli.BoolFlag{Name: skipImageScanFlag, Usage: "Skip sending images to image scan service.\nIf pre-build validation is skipped, image scanning also gets skipped", Value: false},
+				&cli.PathFlag{Name: pathFlag, Usage: "Path to the hashrelease to publish", Required: true},
```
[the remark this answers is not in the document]
caseydavenport: Yep, good catch. This was a remnant from an earlier iteration where I was considering having hashreleases in their own directory. Have adjusted how the dirs are handled now.

radTuti (approved):
otherwise lgtm

PR #9168, opened 2024-08-22
[release tool] update to get hashreleases working

radTuti:
## Description

This add some tweaks to get the hashrelease process working from promotion:

- Update the hashrelease semaphore file to a working state
- Update slack messaging
- Push operator before validations
- Fix image validation for windows images

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/.semaphore/release/hashrelease.yml
+++ b/.semaphore/release/hashrelease.yml
@@ -11,18 +11,11 @@ global_job_config:
   secrets:
     - name: docs-ssh
     - name: marvin-github-token
-    - name: gcloud-registry-access
     - name: hashrelease-docker-auth
     - name: iss-image-scanning
     - name: releasebot-slack
   prologue:
     commands:
-      - chmod 0600 ~/.keys/*
-      - ssh-add ~/.keys/*
-      # Credentials for accessing gcloud, needed to push images to gcr
-      - export GOOGLE_APPLICATION_CREDENTIALS=${HOME}/keys/.registry-viewer-serviceaccount.json
-      - gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS}
-      - docker login
       - checkout
 
 blocks:
@@ -35,10 +28,10 @@ blocks:
             - ./bin/release hashrelease publish
       prologue:
         commands:
-          - cache restore release-${SEMAPHORE_GIT_SHA}
           - cd release
+          - cache restore release-${SEMAPHORE_GIT_SHA}
       env_vars:
         - name: OPERATOR_BRANCH
           value: master
         - name: IS_HASHRELEASE
-          value: true
+          value: "true"
--- a/release/build/main.go
+++ b/release/build/main.go
@@ -76,16 +76,12 @@ func hashrelaseSubCommands(cfg *config.Config, runner *registry.DockerRunner) []
 			Flags: []cli.Flag{
 				&cli.BoolFlag{Name: skipValidationFlag, Usage: "Skip pre-build validation", Value: false},
 			},
-			Before: func(c *cli.Context) error {
+			Action: func(c *cli.Context) error {
 				configureLogging()
 				if !c.Bool(skipValidationFlag) {
 					tasks.PreReleaseValidate(cfg)
 				}
 				tasks.PinnedVersion(cfg)
-				return nil
-			},
-			Action: func(c *cli.Context) error {
-				configureLogging()
 				tasks.OperatorHashreleaseBuild(runner, cfg)
 				tasks.HashreleaseBuild(cfg)
 				tasks.ReleaseNotes(cfg)
@@ -102,17 +98,14 @@ func hashrelaseSubCommands(cfg *config.Config, runner *registry.DockerRunner) []
 				&cli.BoolFlag{Name: skipValidationFlag, Usage: "Skip pre-build validation", Value: false},
 				&cli.BoolFlag{Name: skipImageScanFlag, Usage: "Skip sending images to image scan service.\nIf pre-build validation is skipped, image scanning also gets skipped", Value: false},
 			},
-			Before: func(c *cli.Context) error {
+			Action: func(c *cli.Context) error {
 				configureLogging()
-				// Push the operator hashrelease first, as it is needed for the main hashrelease.
+				// Push the operator hashrelease first before validaion
+				// This is because validation checks all images exists and sends to Image Scan Service
 				tasks.OperatorHashreleasePush(runner, cfg)
 				if !c.Bool(skipValidationFlag) {
 					tasks.HashreleaseValidate(cfg, c.Bool(skipImageScanFlag))
 				}
-				return nil
-			},
-			Action: func(c *cli.Context) error {
-				configureLogging()
 				tasks.HashreleasePush(cfg)
 				return nil
 			},
--- a/release/build/main.go
+++ b/release/build/main.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/projectcalico/calico/release/internal/config"
 	"github.com/projectcalico/calico/release/internal/registry"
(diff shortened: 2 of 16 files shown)


.semaphore/release/hashrelease.yml
```diff
@@ -35,10 +28,10 @@ blocks:
             - ./bin/release hashrelease publish
       prologue:
         commands:
-          - cache restore release-${SEMAPHORE_GIT_SHA}
           - cd release
+          - cache restore release-${SEMAPHORE_GIT_SHA}
       env_vars:
         - name: OPERATOR_BRANCH
           value: master
         - name: IS_HASHRELEASE
```
caseydavenport, line 36: Do we still need this env var?

release/build/main.go
```diff
@@ -101,14 +104,15 @@ func hashrelaseSubCommands(cfg *config.Config, runner *registry.DockerRunner) []
 			},
 			Before: func(c *cli.Context) error {
 				configureLogging()
+				// Push the operator hashrelease first, as it is needed for the main hashrelease.
```
caseydavenport, line 107: Why does this need to go in the "before" section? What is this logically changing (other than putting it before the validation)?

release/internal/hashrelease/pinnedversion.go
```diff
@@ -48,14 +49,14 @@ func (c Component) String() string {
...
-	BaseDomain    string
-	CalicoVersion string
-	Operator      Component
-	Note          string
-	Hash          string
-	Registry      string
-	ReleaseBranch string
+	ReleaseName    string
+	BaseDomain     string
+	ProductVersion string
```
caseydavenport, line 54: Personally I find this change from calico -> product to be more confusing than helpful. This is meant to be the calico repository version, right?

release/internal/hashrelease/pinnedversion.go
```diff
@@ -115,23 +116,25 @@ func GeneratePinnedVersionFile(rootDir, operatorDir, devTagSuffix, registry, out
...
 		Note: fmt.Sprintf("%s - generated at %s using %s release branch with %s operator branch",
-			releaseName, time.Now().Format(time.RFC1123), calicoBranch, operatorBranch),
-		ReleaseBranch: calicoBranch,
+			releaseName, time.Now().Format(time.RFC1123), productBranch, operatorBranch),
+		ReleaseBranch: parsedProductVersion.ReleaseBranch(releaseBranchPrefix),
 	}
 	if registry != "" {
 		data.Operator.Registry = registry
 		data.Registry = registry
+		data.Operator.Image = strings.ReplaceAll(operator.ImageName, "-", "/")
```
caseydavenport, line 137: Why do we need this ReplaceAll here?

.semaphore/release/hashrelease.yml
radTuti: It is still being used, although the default in config is set to true, I want to have it explictly called out here too

release/build/main.go
radTuti: It needs to happen before validations and since validations are in "Before" that's why I moved it there.
Alternatively, we'll have to move everything in "Before" to "Action"

release/internal/hashrelease/pinnedversion.go
radTuti: This change is to avoid conflicts with close-source since that will have the "CalicoVersion" mean something else

release/build/main.go
caseydavenport: Yeah, I would rather us just keep everything in Action then. I am not sure what splitting into Before() is actually buying us.

I think this is indication that Validate is not factored correctly and might need to be broken down a bit - e.g., "pre-release validation" that validates it's OK to cut a release, and "post-release validation" that validates push happened correctly.

radTuti: Agreed, I will move all to Action.
Once hash release is switched to building its own images, this validation will be obsolete as all that will be needed is image scanning.

release/internal/hashrelease/pinnedversion.go
radTuti: I noticed the output in "pinned-version.yaml" was wrong because of this.

caseydavenport: Could you expand on that? What was wrong about it? What was showing up there?

radTuti: Currently it outputs:

```yaml
  tigera-operator:
    version: v1.35.0-0.dev-93-ga0208eb67b07-2024-08-22-hashrelease-detergent
    image: tigera/operator
    registry: tuticat
```
This change will have it output corrected since the image is `tuticat/tigera-operator`
```yaml
  tigera-operator:
    version: v1.35.0-0.dev-93-ga0208eb67b07-2024-08-22-hashrelease-detergent
    image: tigera-operator
    registry: tuticat
```

caseydavenport: I don't think this is the right fix for that - `tigera/operator` is of the form `registry/image`, which is correctly formatted using a `/`. I think it's just coincidence that the `registry/image` can be converted to `image` using this operation in this case.

The production image we produce is of this form:

- registry: `tigera`
- image: `operator`
- tag: `<version>`

It sounds like we need to make each of these three configurable and then used consistently when referencing the image throughout (when building, pushing, etc).

.semaphore/release/hashrelease.yml
caseydavenport: Gotcha. I will probably remove this in my PR since it's redundant with the subcommand split.

release/internal/operator/operator.go
```diff
@@ -14,6 +14,18 @@ const (
...
 )
 
+// ImageInfo returns the registry and image name for the operator images.
+// When a registry is specified in the config, the image name is modified and the registry is used
+func ImageInfo(registryRpl string) (string, string) {
+	registry := Registry
+	imageName := ImageName
+	if registryRpl != "" {
+		registry = registryRpl
+		imageName = strings.ReplaceAll(imageName, "/", "-")
```
caseydavenport, line 24: I still don't think this is appropriate - why do we need to do this? We should just use the image unmodified. 

```go
const (
  ImageName = "operator"
  Registry = "quay.io/tigera"
)
```

radTuti: Operator images are built and pushed for each hashrelease becasue calico version needs to be specifed in the operator.
While testing in my fork, I need to push operator to something other than `tigera/operator` so i current have this to standardize it to `<repo>/tigera-operator` as the image name can't still be `tigera/operator`

caseydavenport: This ReplaceAll command is just too much magic for my liking and makes it harder to understand what is happening and why. If when testing in your fork you need a different image name, then we should just make that a configuration option so that you can specify what you need.

radTuti: I have created an OperatorConfig to allow configuring operator variables i.e. image, registy, git repo & branch

release/pkg/tasks/release.go
```diff
@@ -34,10 +34,7 @@ func PreReleaseValidate(cfg *config.Config) {
 	match := fmt.Sprintf(`^(%s|%s-v\d+\.\d+(?:-\d+)?)$`, utils.DefaultBranch, cfg.RepoReleaseBranchPrefix)
 	re := regexp.MustCompile(match)
 	if !re.MatchString(releaseBranch) {
-		if cfg.Registry == "" {
```
caseydavenport, line 37: Does this not prevent us from running this code against dev branches?

caseydavenport (approved):

release/pkg/tasks/release.go
radTuti: it is actually only used for operator ATM. 
there is currently no support for dev branches and I will like to make that update in a different PR

PR #9178, opened 2024-08-27
Configure Certificate Rotation for the Kubelet

jxlwqq:
Almost a year ago, I used this script `install-calico-windows.ps1` to join a windows node to a k8s cluster.

Recently, I received a warning about an expired certificate from https://github.com/prometheus-operator/kube-prometheus : `KubeClientCertificateExpiration: Client certificate is about to expire`. 

I think this change will fix this problem.

https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

## Release Note

```release-note
Configure kubelet certificate rotation on manually installed Calico for Windows.
```

### DIFF
--- a/node/windows-packaging/CalicoWindows/kubernetes/kubelet-service.ps1
+++ b/node/windows-packaging/CalicoWindows/kubernetes/kubelet-service.ps1
@@ -58,6 +58,7 @@ $argList = @(`
     "--hairpin-mode=promiscuous-bridge",`
     "--cgroups-per-qos=false",`
     "--enforce-node-allocatable=""""",`
+    "--rotate-certificates=true",`
     "--kubeconfig=""c:\k\config"""`
 )
 


jxlwqq:
Execute the following command to manually rotate the certificate: 

```powershell
rm C:\var\lib\kubelet\pki\*
Restart-Service -Name kubelet
```

jxlwqq:
Another question is, since the `C:\CalicoWindows\calico-kube-config` file has been generated and the service-acount-token type is never expires. why is C:\k\cnofig still used?

```
Environment variable KUBECONFIG is already set: c:\k\config
```

coutinhop:
> Another question is, since the `C:\CalicoWindows\calico-kube-config` file has been generated and the service-acount-token type is never expires. why is C:\k\cnofig still used?
> 
> ```
> Environment variable KUBECONFIG is already set: c:\k\config
> ```

I believe it's the other way around. `C:\CalicoWindows\calico-kube-config` predates using `C:\k\config` and might be due to be cleaned up (@song-jiang please correct me if I'm wrong). Also let me take the opportunity to plug [using the operator to install Calico on Windows](https://docs.tigera.io/calico/latest/getting-started/kubernetes/windows-calico/operator), since it makes this and other debugging/maintenance much more convenient, hit us up in the Calico Users slack if you'd like help with it: https://slack.projectcalico.org/

PR #9183, opened 2024-08-28
Netlink retry for list APIs in cni-plugin

sridhartigera:
## Description

Add retry to netlink List APIs in CNI-Plugin

Netlink list APIs now return EINTR when interrupted. This is due to a version bump in the netlink library we use. Given that this is not an error, list APIs must be retried when it returns EINTR. The retry is already handled in felix. This PR handles the CNI plugin netlink list API calls.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


cni-plugin/pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -362,10 +373,18 @@ func SetupRoutes(hostNlHandle *netlink.Handle, hostVeth netlink.Link, result *cn
...
+				retries = 3
+				for {
+					// Search all routes and report the conflict, search the name of the iface
+					routes, err = hostNlHandle.RouteList(nil, netlink.FAMILY_ALL)
+					if err != nil {
+						if errors.Is(err, unix.EINTR) && retries > 0 {
+							retries--
+							continue
+						}
+						return fmt.Errorf("error listing routes")
```
caseydavenport, line 385: Probably want to keep the original error wrapped up in this one when we return it

cni-plugin/pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -343,9 +345,18 @@ func SetupRoutes(hostNlHandle *netlink.Handle, hostVeth netlink.Link, result *cn
 			// Route already exists, but not necessarily pointing to the same interface.
 			case syscall.EEXIST:
 				// List all the routes for the interface.
-				routes, err := hostNlHandle.RouteList(hostVeth, netlink.FAMILY_ALL)
-				if err != nil {
-					return fmt.Errorf("error listing routes")
+				retries := 3
```
caseydavenport, line 348: Couldn't this be `RouteListRetryEINTR`? It looks very similar at least.

sridhartigera: ```RouteListRetryEINTR``` directly calls ```netlink.RouteList```, whereas this one is via the handle.

libcalico-go/lib/netlinkutils/netlinkutils.go
```diff
@@ -0,0 +1,65 @@
...
+package netlinkutils
+
+import (
+	"errors"
+
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+)
+
+func AddrListRetryEINTR(link netlink.Link, family int) ([]netlink.Addr, error) {
```
caseydavenport, line 25: Would like to see some function-level comments explaining when/why these should be used - what does EINTR mean in the context of these functions so that readers don't need to go look it up and assess for themselves.

cni-plugin/pkg/dataplane/linux/dataplane_linux.go
caseydavenport: Could we provide an optional argument to provide the handle? Or maybe just require that all callers pass in a handle to use? Feels worthwhile to be consistent

caseydavenport: In other words, if we're going to define a library to help us with retries we should make it capable of handling all of the retry scenarios we need.

cni-plugin/internal/pkg/testutils/utils_linux.go
```diff
@@ -43,6 +43,8 @@ import (
 	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
+
+	"github.com/projectcalico/calico/libcalico-go/lib/netlinkutils"
```
caseydavenport, line 47: we should keep this import with the rest of the libcalico-go imports (I would expect static-checks to fail and `make fix` to address this?)

caseydavenport (approved):
One minor comment about imports but otherwise LGTM

PR #9189, opened 2024-08-28
Changes tier's nil order to 100,000

mazdakn:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/clientv3/tier.go
+++ b/libcalico-go/lib/clientv3/tier.go
@@ -46,6 +46,17 @@ type tiers struct {
 // Create takes the representation of a Tier and creates it.  Returns the stored
 // representation of the Tier, and an error, if there is any.
 func (r tiers) Create(ctx context.Context, res *apiv3.Tier, opts options.SetOptions) (*apiv3.Tier, error) {
+	if res != nil {
+		// Since we're about to default some fields, take a (shallow) copy of the input data
+		// before we do so.
+		resCopy := *res
+		res = &resCopy
+	}
+	// Change nil order to the default value, i.e. 100,000
+	if res.Spec.Order == nil {
+		order := apiv3.DefaultTierOrder
+		res.Spec.Order = &order
+	}
 	if err := validator.Validate(res); err != nil {
 		return nil, err
 	}
@@ -59,6 +70,16 @@ func (r tiers) Create(ctx context.Context, res *apiv3.Tier, opts options.SetOpti
 // Update takes the representation of a Tier and updates it. Returns the stored
 // representation of the Tier, and an error, if there is any.
 func (r tiers) Update(ctx context.Context, res *apiv3.Tier, opts options.SetOptions) (*apiv3.Tier, error) {
+	if res != nil {
+		// Since we're about to default some fields, take a (shallow) copy of the input data
+		// before we do so.
+		resCopy := *res
+		res = &resCopy
+	}
+	if res.Spec.Order == nil {
+		order := apiv3.DefaultTierOrder
+		res.Spec.Order = &order
+	}
 	if err := validator.Validate(res); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


libcalico-go/lib/clientv3/tier.go
```diff
@@ -46,6 +46,17 @@ type tiers struct {
 // Create takes the representation of a Tier and creates it.  Returns the stored
 // representation of the Tier, and an error, if there is any.
 func (r tiers) Create(ctx context.Context, res *apiv3.Tier, opts options.SetOptions) (*apiv3.Tier, error) {
+	if res != nil {
+		// Since we're about to default some fields, take a (shallow) copy of the input data
```
caseydavenport, line 50: I think we should make a common "defaulting" function to ensure the logic is the same for both.

libcalico-go/lib/clientv3/tier.go
```diff
@@ -131,7 +152,13 @@ func (r tiers) Delete(ctx context.Context, name string, opts options.DeleteOptio
 func (r tiers) Get(ctx context.Context, name string, opts options.GetOptions) (*apiv3.Tier, error) {
 	out, err := r.client.resources.Get(ctx, opts, apiv3.KindTier, noNamespace, name)
 	if out != nil {
-		return out.(*apiv3.Tier), err
```
caseydavenport, line 134: Sorry, I didn't mean that get would actually update the data in the datastore. Rather, the get would "fill in the field" on the returned object. Here's where we do this for ippools: https://github.com/projectcalico/calico/blob/b5ce29dd4c898cf9b19123a91e6aec21b8f308cc/libcalico-go/lib/clientv3/ippool.go#L244-L287

caseydavenport: The other important thing is that felix / typha / other components don't use the v3 client, so we need to make sure that they treat "nil" to mean "1,000,000" 

Unfortunately, from a quick skim there doens't seem to be a great centralized place to do that, so we either need to add one or we need to make sure those components that look at tier ordering understand that nil -> 1,000,000.

Considering that this logic only really matters for enterprise, where we have pre-existing tiers, and enterprise only supports KDD mode, we can probably get away with augmenting the Get / List / Watch logic in the k8s backend tier client to populate this field if it is nil: https://github.com/projectcalico/calico/blob/b5ce29dd4c898cf9b19123a91e6aec21b8f308cc/libcalico-go/lib/backend/k8s/resources/tiers.go#L32-L47

caseydavenport (changes requested):

libcalico-go/lib/backend/k8s/resources/tiers.go
```diff
@@ -41,7 +43,21 @@ func NewTierClient(c *kubernetes.Clientset, r *rest.RESTClient) K8sResourceClien
...
-		k8sListType:  reflect.TypeOf(apiv3.TierList{}),
-		resourceKind: apiv3.KindTier,
+		k8sListType:      reflect.TypeOf(apiv3.TierList{}),
+		resourceKind:     apiv3.KindTier,
+		versionconverter: Tierv1v3Converter{},
 	}
 }
+
+// Tierv1v3Converter implements VersionConverter interface.
+type Tierv1v3Converter struct{}
```
caseydavenport, line 53: I think this can be a private struct?

libcalico-go/lib/clientv3/tier.go
```diff
@@ -131,7 +146,10 @@ func (r tiers) Delete(ctx context.Context, name string, opts options.DeleteOptio
...
+		res := out.(*apiv3.Tier)
+		// Default values when reading from backend.
+		cresources.DefaultTierFields(res)
```
caseydavenport, line 151: I don't think we need to do this defaulting here if the backend code has already done it

libcalico-go/lib/resources/tier.go
```diff
@@ -0,0 +1,31 @@
...
+package resources
+
+import (
+	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+)
+
+func DefaultTierFields(res *apiv3.Tier) {
+	// nil order was allowed before, and it was used for the default tier.
+	// For the implementation of BaselineAdminNetworkPolicy, we need to add a tier
+	// before the default one. As such, the default tier order is changed to 100,000 from nil.
```
caseydavenport, line 24: "after the default one" instead of "before the default one"

mazdakn: fixed.

libcalico-go/lib/backend/k8s/resources/tiers.go
mazdakn: Fixed.

caseydavenport (approved):

PR #9213, opened 2024-09-06
[BPF] Use bpf_redirect_peer when CALI_CT_ESTABLISHED_BYPASS

tomastigera:
When a packet arrives on a host interface from the network and conntract says that the status is CALI_CT_ESTABLISHED_BYPASS, that means, the packet should be forwarded to another interface (workload) and the BPF program on that interface would just let it in, it is faster to skip that BPF program and redirect is straight to the ingress side of the workload device (veth). bpf_redirect_peer() does it in the same context in which it processes the ingress on the host interface.

This should cover almost all in-cluster traffic when CTLB is enabled since there is no NAT happening (done in connect) and thus we can preapprove traffic on the host device.

Note that this works for IPIP - we do not have FIB and forwarding enabled for IPIP, but when a packet arrives to the host via IPIP, it is fast forwarded to the destination workload as well. However it is not enabled by default.     BPFRedirectToPeerFromL3Device is set to false by default, but it is an option to enabled using bpf_redirect_peer also with ipip and wireguard devicese as there are some pitfalls like broken tcpdump on the peer side if redirection is from an L3 device.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: faster redirection from host interface to workloads for any includer traffic that does not use NAT (CTLB enabled or pod-to-pod)
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/conntrack.h
```diff
@@ -985,14 +985,27 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_lookup(struct cali_tc_c
...
 						src_to_dst->ifindex, ifindex);
 			}
-			/* Do not worry about packets returning from the same direction as
-			 * the outgoing packets.
-			 *
-			 * Do not check if packets are returning from the NP vxlan tunnel.
-			 */
-			if (!same_if && !ret_from_tun && !hep_rpf_check(ctx) && !CALI_F_NAT_IF && !CALI_F_LO) {
+
+			bool recheck = false;
```
fasaxc, line 989: I read "recheck" as "this packet needs to be rechecked" but I think you're using it to mean "result of the recheck is ..."?  How about calling it `rpf_passed`?

fasaxc (approved):
One naming query but LGTM otherwise

PR #9221, opened 2024-09-12
Service LoadBalancer kube-controller

MichalFupso:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add LoadBalancer kube-controller to allow for Service LoadBalancer IPAM
```

## Reminder for the reviewer

Other optional labels:


kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+		parsedIPs = append(parsedIPs, *curr)
+	}
+
+	if ipv6 > 1 || ipv4 > 1 {
+		return nil, errors.New(fmt.Sprintf("At max only one ipv4 and one ipv6 address can be specified. Recieved %d ipv4 and %d ipv6 addresses", ipv4, ipv6))
+	}
+	return parsedIPs, nil
+}
+
+func parseAnnotation(str string) []string {
```
caseydavenport, line 678: I think you can just `json.Unmarshal` here instead?

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// validateAnnotation checks if the ips specified in the calico annotation are valid.
+// Each service can have at most one ipv4 and one ipv6 address
+func validateAnnotation(ipAddrs []string) ([]cnet.IP, error) {
```
caseydavenport, line 655: This function is called `validateAnnotation` but it's actually validating a list of strings.

I'd expect `validateAnnotation` to take the raw annotation - I'd probably call this `parseCIDRs`

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+		result = append(result, cnet.IPNet{IPNet: *cidr})
+	}
+	return result, nil
+}
+
+// isCalicoManagedLoadBalancer returns if Calico should try to assign IP address for the LoadBalancer
+// We assign IPs only if the loadBalancer controller assignIP is set to AllService
+// or in RequestedOnlyServices if the service has calico annotation
+func isCalicoManagedLoadBalancer(svc *v1.Service, assignIPs string) bool {
+	if svc.Spec.Type == v1.ServiceTypeLoadBalancer {
```
caseydavenport, line 637: I'd factor this like so to reduce some indentation complexity:

```go
if svc.Spec.Type != v1.ServiceTypeLoadBalancer {
  return false
}

if assignIPs == api.AllServices {
  return true
}

// Other logic
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+		Num4:        num4,
+		Num6:        num6,
+		IntendedUse: api.IPPoolAllowedUseLoadBalancer,
+		Hostname:    api.VirtualLoadBalancer,
+		HandleID:    &handle,
+		Attrs:       metadataAttrs,
+	}
+
+	if svc.Annotations[annotationIpv4Pools] != "" {
+		args.IPv4Pools, _ = c.resolvePools(parseAnnotation(svc.Annotations[annotationIpv4Pools]), true)
```
caseydavenport, line 508: We shouldn't be ignoring errors anywhere

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+}
+
+// resolvePools valid IPpool range when specific pool is requested by the user
+func (c *loadBalancerController) resolvePools(pools []string, isv4 bool) ([]cnet.IPNet, error) {
```
caseydavenport, line 595: I think for clarity, we should adjust naming of variables to use a consistent scheme throughout depending on the type of object. We have a lot of things called "pools" but they can be of various types, and they are often being cross referenced with each other which makes the code harder to read:

- v3.IPPool objects
- string that represents a CIDR
- A parsed CIDR
- string that represents a pool name

I'd suggest a naming pattern like:

- pools - []v3.IPpool
- poolCIDRs - []string or []cnet.IPNet
- poolNames - []string
- poolIDs - []string (mixed name / cidr)

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+	annotationIpv6Pools      = "projectcalico.org/ipv6pools"
+	annotationLoadBalancerIp = "projectcalico.org/loadBalancerIPs"
+	timer                    = 1 * time.Minute
+)
+
+// loadBalancerController implements the Controller interface for managing Kubernetes services
+// and endpoints, syncing them to the Calico datastore as NetworkSet.
+type loadBalancerController struct {
+	informer        cache.Controller
+	lbResourceCache rcache.ResourceCache
```
caseydavenport, line 61: I don't think we should be using the `rcache.ResourceCache` interface here (I think I called this out earlier in the design document as well)

It's a really old concept that comes from a time before we had established better patterns. It also has some syncing logic built in that I don't think we really want here.

It looks like you're using it for Services - we should use the indexer/informer pattern for this which comes with its own cache anyway. 

You can create a new Service informer like the ones here: https://github.com/caseydavenport/calico/blob/22d6a666f4d6f7036422bd960d5194871fa623c5/kube-controllers/cmd/kube-controllers/main.go#L431-L432

Each controller that needs can add their own handlers to the existing informer like so: https://github.com/caseydavenport/calico/blob/master/kube-controllers/pkg/controllers/node/controller.go#L127-L131

You can access the informer's indexer like so: https://github.com/caseydavenport/calico/blob/22d6a666f4d6f7036422bd960d5194871fa623c5/kube-controllers/pkg/controllers/node/controller.go#L83

And access a "Lister" interface to access the underlying cache like so: https://github.com/caseydavenport/calico/blob/22d6a666f4d6f7036422bd960d5194871fa623c5/kube-controllers/pkg/controllers/node/ipam.go#L140-L141

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+	syncStatus      bapi.SyncStatus
+	syncChan        chan interface{}
+	ipamBlocks      map[string]model.KVPair
+	ipPools         map[string]api.IPPool
+}
+
+// NewLoadBalancerController returns a controller which manages Service LoadBalancer objects.
+func NewLoadBalancerController(ctx context.Context, clientset *kubernetes.Clientset, calicoClient client.Interface, cfg config.LoadBalancerControllerConfig) controller.Controller {
+	// set up service informer
+	svcWatcher := cache.NewListWatchFromClient(clientset.CoreV1().RESTClient(), "services", "", fields.Everything())
```
caseydavenport, line 77: See my other comment about using a SharedInformer following the pattern of the Pod / Node informers instead of the rcache.ResourceCache interface - it would get rid of this svcWatcher / listFunc logic here.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+		LogTypeDesc: "Service",
+	}
+	ccache := rcache.NewResourceCache(cacheArgs)
+
+	svcHandler := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			log.Debugf("Got ADD event for service")
+			svc := obj.(*v1.Service)
+			if isCalicoManagedLoadBalancer(svc, cfg.AssignIPs) {
+				ccache.Set(createHandle(svc), svc)
```
caseydavenport, line 112: Once the ResourceCache stuff is pulled out, we shouldn't be doing any work on any goroutine except for the main worker - all of our event handler should simply queue the received event on a channel for the main worker routine to read and then return.

Much like how you have implemented the `onUpdate` function for Calico resources, we should simply send these service updates through to the main routine via a channel (and probably via a helper structure that contains the update type ADD / DEL / UPD)

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+	}
+}
+
+// Run starts the controller.
+func (c *loadBalancerController) Run(stopCh chan struct{}) {
+	defer uruntime.HandleCrash()
+
+	// Start the Kubernetes informer, which will start syncing with the Kubernetes API.
+	log.Info("Starting LoadBalancer controller")
+	go c.informer.Run(stopCh)
```
caseydavenport, line 164: Once you switch to a shared informer, this will be handled here: https://github.com/caseydavenport/calico/blob/22d6a666f4d6f7036422bd960d5194871fa623c5/kube-controllers/cmd/kube-controllers/main.go#L479-L482

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+	// Wait until we are in sync with the Kubernetes API before starting the
+	// resource cache.
+	log.Debug("Waiting to sync with Kubernetes API (Service)")
+	for !c.informer.HasSynced() {
+		time.Sleep(100 * time.Millisecond)
+	}
+	log.Debug("Finished syncing with Kubernetes API (Service)")
+	c.lbResourceCache.Run(timer.String())
+
+	for i := 0; i < c.cfg.NumberOfWorkers; i++ {
```
caseydavenport, line 175: We don't need multiple workers in this controller - in fact, doing the logic of this controller in parallel will break things like IPAM GC because each worker will be unaware of the actions the other workers are taking, so we should just get rid of this.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+		go wait.Until(c.runWorker, time.Second, stopCh)
+	}
+
+	log.Info("LoadBalancer controller is now running")
+
+	// Register to the syncer datafeed for updates to ippools and ipamBlocks
+	c.RegisterWith(c.dataFeed)
```
caseydavenport, line 182: I'd expect this registration to happen in the `New` function for the controller

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+		},
+	}
+	_, informer := cache.NewIndexerInformer(svcWatcher, &v1.Service{}, 0, svcHandler, cache.Indexers{})
+
+	return &loadBalancerController{
+		calicoClient:    calicoClient,
+		ctx:             ctx,
+		cfg:             cfg,
+		clientSet:       clientset,
+		dataFeed:        node.NewDataFeed(calicoClient),
```
caseydavenport, line 148: The intention of the DataFeed structure is to be a shared resource to prevent multiple parallel syncers running, which is wasteful.

I think we probably want to construct the DataFeed higher up and then pass it down to this controller and the Node controller, instead of both controllers creating a new instance.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+		if host != nil && *host == fmt.Sprintf("host:%s", api.VirtualLoadBalancer) {
+			c.ipamBlocks[kvp.Key.String()] = kvp
+		}
+	} else {
+		delete(c.ipamBlocks, kvp.Key.String())
+	}
+}
+
+func (c *loadBalancerController) handleIPPoolUpdate(kvp model.KVPair) {
+	log.Infof("Addind IPPOOL")
```
caseydavenport, line 269:
```suggestion

```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+			kick(c.syncChan)
+		}
+		return
+	case model.KVPair:
+		switch update.Key.(type) {
+		case model.ResourceKey:
+			switch update.Key.(model.ResourceKey).Kind {
+			case api.KindIPPool:
+				log.Infof("Running periodic IPAM sync of Service LoadBalancer")
+				c.handleIPPoolUpdate(update)
```
caseydavenport, line 246:
```suggestion
				c.handleIPPoolUpdate(update)
				kick(c.syncChan)
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+			return
+		}
+	}
+
+}
+
+func (c *loadBalancerController) handleBlockUpdate(kvp model.KVPair) {
+	if kvp.Value != nil {
+		host := kvp.Value.(*model.AllocationBlock).Affinity
+		if host != nil && *host == fmt.Sprintf("host:%s", api.VirtualLoadBalancer) {
```
caseydavenport, line 260: For virtual affinities, we really want to be using something other than "host"

i.e., the IPAM code should be writing affinities like this for LB IPs:

```
affinity: "virtual:loadbalancer"
```

And this for "real" IPs:

```
affinity: "host:node"
```

The "prefix" before the `:` is meant to indicate what type of thing this is affine to. If we just use `host:` like we do for real IPs, then anyone can create a real Node in the cluster named `virtual-load-balancer` and it will confuse our code.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,700 @@
...
+							log.Errorf("Error assigning IP to service %s: %s", service.Name, err)
+						}
+					}
+				}
+			}
+		}
+	}
+}
+
+func (c *loadBalancerController) runWorker() {
```
caseydavenport, line 348: I think we can get rid of this entire call chain - all of the Service IP allocation should be happening within `syncIPAM`

It's imperative that we use a single goroutine for both allocating service IPs, as well as IPAM garbage collection, so that the two are always in sync with each other.

caseydavenport (changes requested):
@MichalFupso first pass done - there are some structural changes we need to make to the LB controller in kube-controllers but I don't think they are going to be difficult. 

Mostly a matter of aligning patterns to get rid of the ResourceCache and instead consistently use `SharedIndexInformer` for Services and a common `DataFeed` for Calico resources. 

I'm going to keep reviewing in more detail, but wanted to get the larger changes back to yoru ASAP.

api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
```diff
@@ -84,6 +84,9 @@ type ControllersConfig struct {
 
 	// Namespace enables and configures the namespace controller. Enabled by default, set to nil to disable.
 	Namespace *NamespaceControllerConfig `json:"namespace,omitempty"`
+
+	// LoadBalancer enables and configures the LoadBalancer controller. Enabled by default, set to nil to disable.
```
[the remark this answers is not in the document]
caseydavenport: Checked on this - it should be OK (but kube controllers config is a bit weird)

libcalico-go/lib/clientv3/ippool.go
```diff
@@ -283,7 +283,8 @@ func convertIpPoolFromStorage(pool *apiv3.IPPool) error {
 		pool.Spec.VXLANMode = apiv3.VXLANModeNever
 	}
 
-	if pool.Spec.AssignmentMode == "" {
+	assignmentMode := pool.Spec.AssignmentMode
+	if &assignmentMode == nil {
```
caseydavenport, line 287: I think something is off here - it doesn't make sense to check if the address of a variable is `nil`.

libcalico-go/lib/ipam/ipam.go
```diff
@@ -320,6 +307,9 @@ func (c ipamClient) determinePools(ctx context.Context, requestedPoolNets []net.
 	// We only want to use IP pools which actually match this node, so do a filter based on
 	// selector.
```
caseydavenport, line 308: Comment needs updating here as well

libcalico-go/lib/ipam/ipam.go
```diff
@@ -1693,6 +1683,17 @@ func (c ipamClient) releaseByHandle(ctx context.Context, blockCIDR net.IPNet, op
 		if err = c.ensureConsistentAffinity(ctx, block.AllocationBlock); err != nil {
 			logCtx.WithError(err).Warn("Error ensuring consistent affinity but IP already released. Returning no error.")
 		}
+
+		// If this is loadBalancer we delete the block without waiting
+		if *block.Affinity == "virtual:virtual-load-balancer" {
```
caseydavenport, line 1688: I think we should define this as a constant.

libcalico-go/lib/ipam/ipam_block_reader_writer.go
```diff
@@ -211,7 +211,11 @@ func (rw blockReaderWriter) claimAffineBlock(ctx context.Context, aff *model.KVP
 	logCtx := log.WithFields(log.Fields{"host": host, "subnet": subnet})
 
 	// Create the new block.
-	affinityKeyStr := "host:" + host
+	prefix := "host:"
+	if host == v3.VirtualLoadBalancer {
```
caseydavenport, line 215: I don't think this quite solves the problem - we're still effectively keying off the hostname in order to determine this.

So, someone can still make a Kubernetes node in their cluster that triggers this code.

I think we need to plumb this knowledge down through the IPAM code from the caller somehow, so that we can make sure that only the LB controller is ever generating affinities in this way.

caseydavenport: For example, we could:

- use the `IntendedUse` field of the `AutoAssignArgs` (and add it to `AssignIPArgs`)
- infer this from the provided `Attrs` map on `AutoAssignArgs` / `AssignIPArgs`
- add a new `AffinityType: Host | Virtual` option to those XArgs structs.

I think my preference is the first of those three (at least for now, while we only have a single scenario where it's possible for have a non-host affinity). 

If we ever end up making this more flexible, then option #3 might be better - but until then best not to enable combinations that we don't want.

caseydavenport: I think you will also need to look into these references to `"host:"` and see what the correct resolution is:

- https://github.com/caseydavenport/calico/blob/09903342c9c9927065afad5cbd5503947fbd29e7/libcalico-go/lib/ipam/ipam_block.go#L191-L194
- https://github.com/caseydavenport/calico/blob/09903342c9c9927065afad5cbd5503947fbd29e7/libcalico-go/lib/ipam/ipam_block.go#L196-L201
- https://github.com/caseydavenport/calico/blob/09903342c9c9927065afad5cbd5503947fbd29e7/libcalico-go/lib/ipam/ipam_block_reader_writer.go#L323-L330

caseydavenport: One thing to consider might be to replace all of the `host string` arguments passed around to something more generic - e.g.,

```go
type affinityConfig struct {
  type affinityType

  // The affinity value - a hostname for "host" type affinities. 
  value string
}
```

```go
func (c ipamClient) ensureBlock(ctx context.Context, rsvdAttr *HostReservedAttr, requestedPools []net.IPNet, version int, host string) (*net.IPNet, error) {
```
Would become

```go
func (c ipamClient) ensureBlock(ctx context.Context, rsvdAttr *HostReservedAttr, requestedPools []net.IPNet, version int, affinityCfg affinityConfig) (*net.IPNet, error) {
```

This would at least help ensure we're passing the necessary context wherever it might be needed.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -30,157 +32,143 @@ import (
...
-			// Service changed type, we need to release the addresses used by the service
+			handle, err := createHandle(svcNew)
+			if err != nil {
+				log.WithError(err).Error("Error creating load balancer handle")
+				return
+			}
 			if svcNew.Spec.Type != v1.ServiceTypeLoadBalancer &&
 				svcOld.Spec.Type == v1.ServiceTypeLoadBalancer {
-				ccache.Delete(createHandle(svcOld))
+				lbc.syncerUpdates <- serviceObject{
```
caseydavenport, line 127: `syncerUpdates` implies that the updates are coming from the syncer, but these are actually coming from the informer

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -30,157 +32,143 @@ import (
...
+					updateType: serviceUpdateTypeDELETE,
+				}
 			} else if svcOld.Annotations[annotationIpv4Pools] != svcNew.Annotations[annotationIpv4Pools] ||
```
caseydavenport, line 130: I think it's going to be a lot cleaner to keep all of the logic about whether or not we should act on updates within the reconcile loop - when we get an update, let's just send it through as an update. The other side of the channel should be doing all of the processing.

So basically, this branch should look something like this:

```go
UpdateFunc: func(oldObj, newObj interface{}) {
  // Tell the main loop that this service has changed.
  lbc.serviceUpdates <- NamespacedName{Name: newObj.Name, Namespace: newObj.Namespace}
}
```

The main loop can decide if it cares about this update, it can decide to batch actions, etc., etc.

It doesn't even need to cache the object - the informer already has a backing cache that is accessible and updated at this point, so we just need to tell the main loop "something changed, here's the context"

MichalFupso: @caseydavenport I think that we should do a check here to see if we care about the update. It's mostly because we want to update any IPs if the annotations have changed. If we just trigger an update here without checking if the annotations are different it will be harder for us to know if we should re-assign the IP or the change was not relevant to us. I agree that we do not need to pass the service, but rather just use the informer

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -277,144 +269,124 @@ func (c *loadBalancerController) handleIPPoolUpdate(kvp model.KVPair) {
...
-		svc := obj.(*v1.Service)
-		// We don't have IP, assign one
-		if svc.Status.LoadBalancer.Ingress == nil {
-			return c.assignIP(svc)
-		} else {
-			// Service was updated
-			return c.checkStatus(svc)
+	if len(c.servicesToUpdate) != 0 {
+		for key, svcObj := range c.servicesToUpdate {
+			switch svcObj.updateType {
```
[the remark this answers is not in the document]
caseydavenport: This is somewhat hastily written out, so may not be bulletproof, but just as something to run past you to see what you think - the following pseudocode is probably where I would start from if trying to figure out this logic myself. 

---

The controller would have some internal state that it uses to efficiently track IP allocation state. It would probably need to initialize these maps on startup with IPAM state, and then make incremental changes to them as it allocates and releases IP addresses.

Could use a helper structure to encapsulate that logic, which might make it a bit tidier:

```go
type allocationTracker struct {
  // Keep track of IP allocations by Service, as well as a 
  // reverse lookup of service by IP. These maps are populated based on the IPAM DB state 
  // and NOT by the Service objects.
  servicesByIP map[string]string
  ipsByService map[string]map[string]bool
}

func (t *allocationTracker) AssignAddress(svc *v1.Service, ip string)
func (t *allocationTracker) ReleaseAddress(svc *v1.Service, ip string)
func (t *allocationTracker) DeleteService(svc *v1.Service)
```

The `syncIPAM` function would have something like this:

```go
// Go through each Service that we know about and make sure it is properly in sync.
for _, svc := range c.informer.List() {
  if err := c.syncService(svc); err != nil {
    logrus.WithError(err).Warn("warning")
  }
}
```

```go
// syncService does the following:
// - Releases any IP addresses in the IPAM DB associated with the Service that are not in the Service status.
// - Allocates any addresses necessary to satisfy the Service LB request
// - Updates the controllers internal state tracking of which IP addresses are allocated.
// - Updates the IP addresses in the Service Status to match the IPAM DB.
func syncService(svc *v1.Service) error {
  // TODO
  return nil
}
```

you could even have "event driven" logic accept incoming service updates funneled into the same function, simply calling `syncService` when we get service updates:

```go
select {
  case <-kickChan:
    syncIPAM();
  case <- syncerUpdates:
    handleUpdate(upd)
  case svcID <- serviceUpdates:
    svc := c.informer.Get(svcID)
    syncService(svc)
}
```
}

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -277,144 +269,124 @@ func (c *loadBalancerController) handleIPPoolUpdate(kvp model.KVPair) {
...
+			case serviceUpdateTypeUPDATE:
+				err = c.releaseIP(svcObj.handle)
+				if err != nil {
+					log.WithError(err).Error("Error releasing Service IP")
+					continue
+				}
+				svc, err := c.clientSet.CoreV1().Services(svcObj.service.Namespace).Get(context.Background(), svcObj.service.Name, metav1.GetOptions{})
```
caseydavenport, line 358: We don't want to be using the clientset to Get() services - we have an informer / indexer that we can use to access the latest version of the Service what we know about from the Watch

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -625,69 +580,116 @@ func (c *loadBalancerController) resolvePools(pools []string, isv4 bool) ([]cnet
...
+	_, err := hasher.Write([]byte(handle))
+	if err != nil {
+		log.WithError(err).Panic("Failed to generate hash from handle")
+		return "", err
+	}
+
+	hash := base64.RawURLEncoding.EncodeToString(hasher.Sum(nil))
+	regex := regexp.MustCompile("([-_.])")
+	hash = regex.ReplaceAllString(hash, "")
+	handle = prefix + hash
```
caseydavenport, line 687: This is fine, and no need to change I don't think - but what I was suggesting was that we only add the hash if the handle exceeds the name limit.

e.g.,

```go
handle := strings.ToLower(fmt.Sprintf("%s-%s-%s", svc.Name, svc.Namespace, svc.UID))

if len(handle) > k8svalidation.DNS1123LabelMaxLength {
  // Handle is too long - trim off the end, leaving enough room for 8 bytes of hash.
  hash := generateHash(handle)
  handle = handle[:k8svalidation.DNS1123LabelMaxLength-8]
  handle = fmt.Sprintf("%s%s", handle, hash[:8])
}
```

This  just makes the actual handle IDs human readable in most cases.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
caseydavenport: It might work - I need to think about this one a little bit more. My fear is that comparing the annotations between the "new" and the "old" version of the Service isn't strictly speaking the thing this controller cares about.

The controller wants to compare the annotations, the `LoadBalancer.Ingress` values, and the actual IPAM allocation database to determine whether we need to allocate or release IP addresses.

Comparing the annotations is a proxy for that, and probably catches _most_ cases but isn't guaranteed to catch all of the edge cases we care about.

For example, what happens if another controller modifies the Service and overwrites the status.LoadBalancer values that were previously written by this controller? We could try to write special logic to catch that here, but we may not be able to foresee every scenario that requires an update on the channel.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -30,157 +32,143 @@ import (
...
+	"k8s.io/apimachinery/pkg/labels"
 	uruntime "k8s.io/apimachinery/pkg/util/runtime"
+	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/kubernetes"
+	v1lister "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 )
 
 const (
 	annotationIpv4Pools      = "projectcalico.org/ipv4pools"
```
caseydavenport, line 51:
```suggestion
	annotationIPv4Pools      = "projectcalico.org/ipv4pools"
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -30,157 +32,143 @@ import (
...
 	annotationIpv6Pools      = "projectcalico.org/ipv6pools"
```
caseydavenport, line 52:
```suggestion
	annotationIPv6Pools      = "projectcalico.org/ipv6pools"
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -30,157 +32,143 @@ import (
...
 	annotationLoadBalancerIp = "projectcalico.org/loadBalancerIPs"
```
caseydavenport, line 53:
```suggestion
	annotationLoadBalancerIP = "projectcalico.org/loadBalancerIPs"
```

caseydavenport (changes requested):
@MichalFupso  About to go into meetings, but wanted to get another round in - only reviewed the libcalico-go changes in this one and didn't get to the controller yet.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,952 @@
...
+// - Updates the IP addresses in the Service Status to match the IPAM DB.
+func (c *loadBalancerController) syncService(svcObj serviceObject) error {
+	var err error
+	var svc *v1.Service
+	serviceUpdated := false
+	reassignIP := false
+	svcStatus := make(map[string]v1.LoadBalancerIngress)
+
+	switch svcObj.updateType {
+	case serviceUpdateTypeDELETE:
```
caseydavenport, line 408: I still dislike our sync loop being aware of event types - it shouldn't care about whether we received a DELETE event or not. Instead, it should care about whether or not the service actually exists.

I would prefer if this function took a `types.NamespacedName` and used it to look for the service object in the cache.

```go
func (c *loadBalancerController) syncService(key types.NamespacedName) error {

  // Check if the service exists.
  svc, err := c.serviceLister.Services(key.Namespace).Get(key.Name)
  if err != nil { . . . }

  if svc == nil {
    // Service doesn't exist - clean up state and release any IPs
  } else {
    // Service exists - sync it.
  }
}
```


I don't think this changes the logic you have written here in any substantial way, except:
- It simplifies the `serviceObject` type - it's only used on the channel to signal events. In fact, we could remove the type entirely and just send a `NamespacedName` (it doesn't need to specify an event type, because we can use the presence / absence of the service in the cache to determine what course of action to take).
- It correctly signals that this code is not event driven, rather it's operating based on cluster state.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,952 @@
...
+
+	if svc.Annotations[annotationIPv4Pools] != "" ||
+		svc.Annotations[annotationIPv6Pools] != "" ||
+		svc.Annotations[annotationLoadBalancerIP] != "" {
+		return true
+	}
+	return false
+}
+
+// validateAnnotation checks if the ips specified in the calico annotation are valid.
```
caseydavenport, line 831: comment is out of date

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,952 @@
...
+	hash = regex.ReplaceAllString(hash, "")
+	handle = prefix + hash
+	if len(handle) > k8svalidation.DNS1123LabelMaxLength {
+		handle = handle[:k8svalidation.DNS1123LabelMaxLength]
+	}
+
+	return handle, nil
+}
+
+func isIpInIppool(ipAddr string, cidrs []cnet.IPNet) bool {
```
caseydavenport, line 918:
```suggestion
func isIPInIPPools(ipAddr string, cidrs []cnet.IPNet) bool {
```

(acrynyms same-case, plural since multiple pools)

Or even just `poolsContain(cidrs, ipAddr)`

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,952 @@
...
+	ip := net.ParseIP(ipAddr)
+	for _, cidr := range cidrs {
+		if cidr.Contains(ip) {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *loadBalancerController) localIppoolFromIp(ipAddr string) (*api.IPPool, error) {
```
caseydavenport, line 931:
```suggestion
func (c *loadBalancerController) poolForIP(ipAddr string) (*api.IPPool, error) {
```

What does "local" mean in this context?

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,952 @@
...
+
+			if ipv6 > 1 || ipv4 > 1 {
+				return nil, nil, nil, fmt.Errorf("At max only one ipv4 and one ipv6 address can be specified. Recieved %d ipv4 and %d ipv6 addresses", ipv4, ipv6)
+			}
+		}
+	}
+	return loadBalancerIPs, ipv4Pools, ipv6Pools, nil
+}
+
+// createHandle Returns a handle to use for IP allocation for the service
```
caseydavenport, line 895:
```suggestion
// createHandle returns a handle to use for IP allocation for the service
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,952 @@
...
+
+	// Load services and assigned IPs into cache
+	svcObjs, err := c.getServiceObjectList()
+	if err != nil {
+		log.WithError(err).Fatal("Failed to get service objects")
+	}
+	for _, svcObj := range svcObjs {
+		if IsCalicoManagedLoadBalancer(svcObj.service, c.cfg.AssignIPs) {
+			for _, ingress := range svcObj.service.Status.LoadBalancer.Ingress {
+				c.allocationTracker.assignAddress(svcObj.service, ingress.IP)
```
caseydavenport, line 235: We need to update the allocation tracker based on the IPAM DB, and not based on the LoadBalancer.Ingress fields. This is becuase we use the `allocationTracker` as a faster / cached lookup of the contents of the IPAM DB.

If we use the service Status, we are prone to bugs like these:

- Someone puts a bogus entry into the Status in order to confuse us (intentionally or accidentally - other controllers might mess with this field)
- We have IPs allocated, but the Service is deleted while the pod is down (we'll not realize). 

You can get this "for free" if you use the `onBlockUpdate` function to populate the allocationTracker.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,952 @@
...
+		err = c.syncService(svcObj)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to sync service %s/%s", svcObj.service.Namespace, svcObj.service.Name)
+			continue
+		}
+	}
+
+	for _, block := range c.ipamBlocks {
+		b := block.Value.(*model.AllocationBlock)
+		for key := range b.SequenceNumberForAllocation {
```
caseydavenport, line 373: I think it's a tiny bit more consistent to iterate through the `b.Allocations` array instead - mostly because it's how other places we do this in the code already work.

https://github.com/projectcalico/calico/blob/6c8512e193ed50e71fdf41e3f8f7e93f2f75e4e3/kube-controllers/pkg/controllers/node/ipam.go#L428-L432

What you've written here works though, and has the benefit of only iterating alloacted IPs (the other approach will alocate the whole block) so I am inclined to leave it this way, even if it's a bit inconsistent.

The reason we can't switch other loops to this approach is because the `SequenceNumberForAllocation` field hasn't always existed, meaning this won't work for allocations that pre-date the introduction of that field.

I think in your particular case, it's OK because the LB controller will only ever be interacting with "new" blocks that have this SeuqenceNumberForAllocation field set.

Maybe worth a small comment explaining all of this?

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,952 @@
...
+			ordinal, err := strconv.Atoi(key)
+			if err != nil {
+				log.WithError(err).Errorf("Failed to parse Ordinal for block %s", key)
+			}
+			ip := b.OrdinalToIP(ordinal)
+
+			if _, ok := c.allocationTracker.servicesByIp[ip.String()]; !ok {
```
caseydavenport, line 380: At least according to how I was imagining this would work, the `allocationTracker` was to be a reflection of the data from the IPAM blocks. We already have the Service data available to us in the Service indexer / lister. So, comparing `c.ipamBlocks` with `c.allocationTracker` isn't a useful comparison. I see that right now you've used `c.allocationTracker` more as a reflection of the `Status` field, but I worry (as mentioned elsewhere) that the Status field is an unreliable source of truth for which IP addresses are actually allocated in the cluster.

So, I would expect this code to iterate the IPAM data, use the metadata to reverse engineer the Service NamespacedName, and then compare the IPAM data to the actual Service data (or the allocation tracker, which should be equivalent).

One way to do that might be something like this? (needs a bit more thought)

```go
// Track all services to sync across the k8s API and our IPAM data base.
servicesToCheck := map[string]bool{}

// Add services we know about from the API.
for _, svc := range serviceLister.List() {
  servicesToCheck[namespacedName(svc)] = true
}

// Add services we know about from our IPAM data, if not present in the API.
for svc := range c.allocationTracker.ipsByService {
  servicesToCheck[namespacedName(svc)] = true
}

// Sync all of the services we know about (whether they exist or not). 
// syncService handles cleaning up any IPs for the service if it doesn't exist, and allocates / releases IPs 
// as needed for those that do.
for svc := servicesToCheck {
  syncService(svc)
}
```

caseydavenport (commented):
OK, so, while there are a lot of comments in this round I think this is actually progressing pretty quickly to the right place.

There are a couple of larger requests in this round - I think they may be coming from some confusion on my end or maybe a disconnect on which data sets should be represented by which caches! So might be worth a Zoom call to discuss once you're back in the office.

I'm also a bit delirious / sick today, so hopefully I'm not too far off the mark!

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,915 @@
...
+				num4 = 0
+			}
+			if ip.To16() != nil {
+				num6 = 0
+			}
+		}
+	}
+
+	if num4 == 0 && num6 == 0 {
+		return nil, errors.New("No new IPs to assign, Service already has ipv4 and ipv6 address")
```
caseydavenport, line 659: I think this error is not worded correctly - this could happen if the Service is IPv4 only and already has an IPv4 address, for example.

```suggestion
		return nil, errors.New("No new IPs to assign, Service already has desired LB addresses")
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,915 @@
...
+			c.allocationTracker.assignAddressToService(namespacedName, assignment.IP.String())
+		}
+	}
+
+	return assignedIPs, nil
+}
+
+// releaseIPByHandle tries to release all IPs allocated with the Service unique handle
+func (c *loadBalancerController) releaseIPByHandle(namespacedName types.NamespacedName, handle string) error {
+	log.Info("Service type LoadBalancer deleted, releasing assigned IP address")
```
caseydavenport, line 704: This doesn't necessarily mean the service was deleted, right? It could be that it is no longer a LoadBalancer type?

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,915 @@
...
+
+// syncService does the following:
+// - Releases any IP addresses in the IPAM DB associated with the Service that are not in the Service status.
+// - Allocates any addresses necessary to satisfy the Service LB request
+// - Updates the controllers internal state tracking of which IP addresses are allocated.
+// - Updates the IP addresses in the Service Status to match the IPAM DB.
+func (c *loadBalancerController) syncService(svcObj serviceObject) error {
+	svc, err := c.serviceLister.Services(svcObj.namespacedName.Namespace).Get(svcObj.namespacedName.Name)
+
+	if err != nil {
```
caseydavenport, line 423: I think we need to check if the error IsNotFound here - we don't want to do this for other generic errors that may occur.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,915 @@
...
+		if c.allocationTracker.ipsByService[svcObj.namespacedName] != nil {
+			err = c.releaseIPByHandle(svcObj.namespacedName, svcObj.handle)
+			if err != nil {
+				log.WithError(err).Errorf("Failed to release IP for %s/%s", svcObj.namespacedName.Namespace, svcObj.namespacedName.Name)
+				return err
+			}
+		}
+		kick(c.syncChan)
```
caseydavenport, line 431: Why do we need to kick the channel here?

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,915 @@
...
+		}
+
+		for ip := range c.allocationTracker.ipsByService[svcObj.namespacedName] {
+			if _, ok := statusIPs[ip]; !ok {
+				break
+			}
+			return nil
+		}
+
+		_, err = c.assignIP(svc)
```
caseydavenport, line 544: Won't we hit this even if IPs were allocated based on annotations above? I think the structuring of this logic is a bit tricky to navigate (for me at least), as you need to track the path of each service all the way through a fairly large and complicated function.

It might make sense to decompose this a bit into helper functions - e.g.,

```go
func assignFromAnnotations(svc *ServiceObject) error { }
func assignAutomatically(svc *ServiceObject) error { }
func updateStatusIfNeeded(svc *ServiceObject) error { }
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -0,0 +1,915 @@
...
+			if ipv6 > 1 || ipv4 > 1 {
+				return nil, nil, nil, fmt.Errorf("At max only one ipv4 and one ipv6 address can be specified. Recieved %d ipv4 and %d ipv6 addresses", ipv4, ipv6)
+			}
+		}
+	}
+	return loadBalancerIPs, ipv4Pools, ipv6Pools, nil
+}
+
+// createHandle returns a handle to use for IP allocation for the service
+func createHandle(svc *v1.Service) (string, error) {
```
caseydavenport, line 854: We should write some extensive UTs for this if you haven't already.

Same probably for the `allocationTracker` - both pretty well scoped utilities with clear inputs / outputs.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -28,6 +28,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
+	errors2 "k8s.io/apimachinery/pkg/api/errors"
```
caseydavenport, line 31:
```suggestion
	apierrors "k8s.io/apimachinery/pkg/api/errors"
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -468,114 +457,150 @@ func (c *loadBalancerController) syncService(svcObj serviceObject) error {
...
+func (c *loadBalancerController) updateServiceStatus(svc *v1.Service, svcObj serviceObject) error {
+	var svcIngress []v1.LoadBalancerIngress
+	for ip := range c.allocationTracker.ipsByService[svcObj] {
+		svcIngress = append(svcIngress, v1.LoadBalancerIngress{
+			IP: ip,
+		})
+	}
+	svc.Status.LoadBalancer.Ingress = svcIngress
+
+	_, err := c.clientSet.CoreV1().Services(svc.Namespace).UpdateStatus(context.Background(), svc, metav1.UpdateOptions{})
```
caseydavenport, line 562:
```suggestion
        ctx, cancel := context.WithTimeout(context.Background(), 10 * time.Second)
        defer cancel()
	_, err := c.clientSet.CoreV1().Services(svc.Namespace).UpdateStatus(ctx, svc, metav1.UpdateOptions{})
```

libcalico-go/lib/ipam/ipam_block_reader_writer.go
```diff
@@ -109,10 +109,10 @@ func findContainingPool(pools []v3.IPPool, addr net.IP) (*v3.IPPool, error) {
 //
 // Note that the block may become claimed between receiving the CIDR from this function and attempting to claim the corresponding
 // block as this function does not reserve the returned IPNet.
-func (rw blockReaderWriter) findUsableBlock(ctx context.Context, host string, version int, pools []v3.IPPool, reservations addrFilter, config IPAMConfig) (*cnet.IPNet, error) {
+func (rw blockReaderWriter) findUsableBlock(ctx context.Context, affinityCfg AffinityConfig, version int, pools []v3.IPPool, reservations addrFilter, config IPAMConfig) (*cnet.IPNet, error) {
 	// If there are no pools, we cannot assign addresses.
 	if len(pools) == 0 {
-		return nil, fmt.Errorf("no configured Calico pools for node %s", host)
+		return nil, fmt.Errorf("no configured Calico pools for node %s", affinityCfg.Host)
```
caseydavenport, line 115: "for node %s" needs to vary based on whether or not this is a virtual affinity or a host affinity, no?

libcalico-go/lib/ipam/ipam_block_reader_writer.go
```diff
@@ -153,7 +153,7 @@ func (rw blockReaderWriter) findUsableBlock(ctx context.Context, host string, ve
 			if info, ok := exists[subnet.String()]; !ok {
 				log.Infof("Found free block: %+v", *subnet)
 				return subnet, nil
-			} else if info.affinity == host && info.numFree != 0 {
+			} else if info.affinity == affinityCfg.Host && info.numFree != 0 {
```
caseydavenport, line 156: I think this check needs to also check that the affinity type is the same, otherwise we encounter the same kind of bugs where a host with name `virtual-loadbalancer` will confuse our code.

The purpose of passing an `AffinityConfig` down through these functions is so that they can be aware of the `Type` field.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -296,56 +295,69 @@ func (c *loadBalancerController) handleUpdate(update interface{}) {
...
+		c.allocationTracker.deleteBlock(kvp.Key.String())
+		return
+	}
+
+	allocatedIPs := make(map[string]bool)
+
+	for i := range block.Allocations {
+		if block.Allocations[i] != nil {
+			if _, ok := block.Attributes[*block.Allocations[i]].AttrSecondary[ipam.AttributeNamespace]; !ok {
+				continue
```
caseydavenport, line 324: Might want a warning log here and below - these are noteworthy branches.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -296,56 +295,69 @@ func (c *loadBalancerController) handleUpdate(update interface{}) {
 			}
 		case model.BlockKey:
 			c.handleBlockUpdate(update)
+			kick(c.syncChan)
 			return
 		}
 	}
 }
 
 func (c *loadBalancerController) handleBlockUpdate(kvp model.KVPair) {
```
caseydavenport, line 304: We should make sure to write good UTs for this as well (i.e., sending a series of block updates and checking that the internal allocationTracker maps are updated correctly after each).

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -369,38 +381,19 @@ func (c *loadBalancerController) syncIPAM() {
...
-					handle:         *handle,
-					namespacedName: namespacedName,
-				}
-
-				svcObjs[svcObj] = true
-			}
+	for block := range c.allocationTracker.ipsByBlock {
+		for ip := range c.allocationTracker.ipsByBlock[block] {
+			svcObj := c.allocationTracker.servicesByIP[ip]
+			svcObjs[svcObj] = true
```
caseydavenport, line 396:
```suggestion
	for svcObj := range c.allocationTracker.ipsByService {
		svcObjs[svcObj] = true
```

I think we'll have already condensed the IPAM database into a map of service objects that we can iterate quicker, right? Or is there a reason we want to iterate all of the IPs here?

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -468,114 +457,150 @@ func (c *loadBalancerController) syncService(svcObj serviceObject) error {
...
 	} else {
 		// No annotations are specified, check that the IPs assigned aren't from Manual pool from earlier assignment
-		for ip := range c.allocationTracker.ipsByService[svcObj.namespacedName] {
+		for ip := range c.allocationTracker.ipsByService[svcObj] {
 			pool, err := c.poolForIP(ip)
 			if err != nil {
 				return err
 			}
 			if pool != nil {
 				if pool.Spec.AssignmentMode == api.Manual {
```
caseydavenport, line 494:
```suggestion
			if pool != nil && pool.Spec.AssignmentMode == api.Manual {
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -468,114 +457,150 @@ func (c *loadBalancerController) syncService(svcObj serviceObject) error {
...
+	if len(svc.Status.LoadBalancer.Ingress) != len(c.allocationTracker.ipsByService[svcObj]) {
+		return true
+	}
+
+	for _, ingress := range svc.Status.LoadBalancer.Ingress {
+		if _, ok := c.allocationTracker.ipsByService[svcObj][ingress.IP]; !ok {
+			break
 		}
-		svc.Status.LoadBalancer.Ingress = svcIngress
+		return false
```
caseydavenport, line 546: I don't think this is quite right. It's only going to check the first entry in the `Ingress` list before returning. This works correctly if the first entry in the status is not allocated, but doesn't work correctly when the first entry is allocated and the second is not (since we'll fall through and return `false` before ever executing the loop a second time). 

We need to check both that:
- All IPs that we own are present in the status
- All IPs that we own in the status are actually assigned in our IPAM DB

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -468,114 +457,150 @@ func (c *loadBalancerController) syncService(svcObj serviceObject) error {
...
+	return true
+}
+
+// updateServiceStatus updates the status of the service with IPs from our IPAM storage
+func (c *loadBalancerController) updateServiceStatus(svc *v1.Service, svcObj serviceObject) error {
+	var svcIngress []v1.LoadBalancerIngress
+	for ip := range c.allocationTracker.ipsByService[svcObj] {
+		svcIngress = append(svcIngress, v1.LoadBalancerIngress{
+			IP: ip,
+		})
```
caseydavenport, line 558:
```suggestion
		svcIngress = append(svcIngress, v1.LoadBalancerIngress{IP: ip})
```

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -589,42 +614,30 @@ func (c *loadBalancerController) assignIP(svc *v1.Service) ([]string, error) {
...
-				if lbingress == addrs.String() {
-					skipAddrs = true
-				}
-			}
-
-			if skipAddrs {
+		for _, addr := range loadBalancerIPs {
+			// We must be trying to assign missing address due to an error,
+			// skip this assignment as it's already assigned and move onto the next one
+			if _, exists := c.allocationTracker.ipsByService[*svcObj][addr.String()]; exists {
```
caseydavenport, line 622:
```suggestion
			if _, exists := c.allocationTracker.ipsByService[*svcObj][addr.String()]; exists {
				// We must be trying to assign missing address due to an error,
		    		// skip this assignment as it's already assigned and move onto the next one
```

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -98,7 +98,10 @@ func (c blockAffinityClient) toV1(kvpv3 *model.KVPair) (*model.KVPair, error) {
 	return &model.KVPair{
 		Key: model.BlockAffinityKey{
 			CIDR: *cidr,
-			Host: kvpv3.Value.(*libapiv3.BlockAffinity).Spec.Node,
+			AffinityCfg: model.AffinityConfig{
+				AffinityType: "",
```
caseydavenport, line 102: Shouldn't we be setting the type here as well?

libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -372,7 +374,7 @@ func (c *blockAffinityClient) listV1(ctx context.Context, list model.BlockAffini
 		if err != nil {
 			return nil, err
 		}
-		if host == "" || v1kvp.Key.(model.BlockAffinityKey).Host == host {
+		if host == "" || v1kvp.Key.(model.BlockAffinityKey).AffinityCfg.Host == host {
```
caseydavenport, line 377: All of these checks need to also compare the `AffinityCfg.Type` 

for example:

```go
host := v1kvp.Key.(model.BlockAffinityKey).AffinityCfg.Host
affType := v1kvp.Key.(model.BlockAffinityKey).AffinityCfg.Type
if host == "" || (list.Type == affType && host == list.Host) {

Comment (caseydavenport) [reply libcalico-go/lib/backend/k8s/resources/ipam_affinity.go:377]: Without this, we still have the chance of hitting the case where the hostname matches but the affinity type does not match, which would be a problem.

File: libcalico-go/lib/backend/model/block_affinity.go [hunk 57]
```diff
@@ -39,9 +39,22 @@ const (
...
+	AffinityTypeHost    AffinityType = "host"
+	AffinityTypeVirtual AffinityType = "virtual"
+)
+
 type BlockAffinityKey struct {
-	CIDR net.IPNet `json:"-" validate:"required,name"`
-	Host string    `json:"-"`
+	CIDR        net.IPNet      `json:"-" validate:"required,name"`
+	Host        string         `json:"-"`
+	AffinityCfg AffinityConfig `json:"-"`
```
caseydavenport, line 57: I think this is a bit of a blurring of responsibilities here.

the `type BlockAffinityKey` is already effectively the model representation of an `AffinityConfig`.

I would have expected:

- AffinityConfig stays in the `ipam` package as the IPAM code's way of representing / passing around affinity configuration in its API and internally.
- `BlockAffinityKey` simply gets a new field - `Type` - to represent the new (previously implied to be "host") data that identifies a unique block.

libcalico-go/lib/backend/model/block_affinity.go
```diff
@@ -50,11 +63,12 @@ type BlockAffinity struct {
 }
 
 func (key BlockAffinityKey) defaultPath() (string, error) {
-	if key.CIDR.IP == nil || key.Host == "" {
+	if key.CIDR.IP == nil || key.AffinityCfg.Host == "" {
```
caseydavenport, line 66: This needs to assert on the `key.AffinityCfg.AffinityType` being missing as well.

libcalico-go/lib/backend/model/block_affinity.go
```diff
@@ -105,6 +119,7 @@ func (options BlockAffinityListOptions) KeyFromDefaultPath(path string) Key {
 		return nil
 	}
 	host := r[0][1]
+	affinityType := AffinityType(r[0][0])
```
caseydavenport, line 122: Think we need to update the regex for this to work?

libcalico-go/lib/ipam/ipam_block.go
```diff
@@ -190,22 +190,28 @@ func (b *allocationBlock) assign(affinityCheck bool, address cnet.IP, handleID *
...
+	} else if strings.HasPrefix(*block.Affinity, fmt.Sprintf("%s:", model.AffinityTypeVirtual)) {
+		return *block.Affinity == fmt.Sprintf("%s:", model.AffinityTypeVirtual)+host
 	}
 	return false
 }
 
-func getHostAffinity(block *model.AllocationBlock) string {
-	if block.Affinity != nil && strings.HasPrefix(*block.Affinity, fmt.Sprintf("%s:", AffinityTypeHost)) {
-		return strings.TrimPrefix(*block.Affinity, fmt.Sprintf("%s:", AffinityTypeHost))
+func getHostAffinity(block *model.AllocationBlock) *model.AffinityConfig {
```
caseydavenport, line 201:
```suggestion
func getAffinityConfig(block *model.AllocationBlock) *model.AffinityConfig {
```

libcalico-go/lib/ipam/ipam_block_reader_writer.go
```diff
@@ -319,9 +320,9 @@ func (rw blockReaderWriter) releaseBlockAffinity(ctx context.Context, host strin
 	b := allocationBlock{obj.Value.(*model.AllocationBlock)}
 
 	// Check that the block affinity matches the given affinity.
-	if b.Affinity != nil && !hostAffinityMatches(host, b.AllocationBlock) {
+	if b.Affinity != nil && !hostAffinityMatches(affinityCfg.Host, b.AllocationBlock) {
```
caseydavenport, line 323: `hostAffinityMatches` needs to be aware of `affinityCfg.Type` to make sure we don't release blocks that are of the wrong type. Otherwise there is a risk that we release a block of type `virtual` when we want to release a `host` block (and vice-versa).

libcalico-go/lib/ipam/ipam.go
```diff
@@ -1277,12 +1282,6 @@ func (c ipamClient) ClaimAffinity(ctx context.Context, cidr net.IPNet, affinityC
 		return nil, nil, invalidSizeError(estr)
 	}
 
-	// Determine the hostname to use.
```
caseydavenport, line 1244: hm, why did we need to remove this?

libcalico-go/lib/ipam/ipam.go
```diff
@@ -1393,9 +1397,12 @@ func (c ipamClient) ReleaseBlockAffinity(ctx context.Context, block *model.Alloc
 		logCtx.Info("Block is already released")
 		return nil
 	}
-	hostname := getHostAffinity(block)
+	affinityCfg := getHostAffinity(block)
+	if affinityCfg == nil {
+		return nil
```
caseydavenport, line 1402: I would assume that this branch is an error case? Is there any valid scenario where we expect this to happen?

libcalico-go/lib/ipam/ipam.go
```diff
@@ -1505,19 +1514,21 @@ func (c ipamClient) ReleasePoolAffinities(ctx context.Context, pool net.IPNet) e
 // and removes all host-specific IPAM data from the datastore.
 // RemoveIPAMHost does not release any IP addresses claimed on the given host.
 // If an empty string is passed as the host, then the hostname is automatically detected.
-func (c ipamClient) RemoveIPAMHost(ctx context.Context, host string) error {
+func (c ipamClient) RemoveIPAMHost(ctx context.Context, affinityCfg model.AffinityConfig) error {
 	// Determine the hostname to use.
-	hostname, err := decideHostname(host)
+	hostname, err := decideHostname(affinityCfg.Host)
```
caseydavenport, line 1519:
```suggestion
        var err error
	affinityCfg.Host, err = decideHostname(affinityCfg.Host)
```

caseydavenport: Can do this here and avoid the extra line below.

libcalico-go/lib/ipam/ipam.go
```diff
@@ -1564,8 +1575,8 @@ func (c ipamClient) RemoveIPAMHost(ctx context.Context, host string) error {
 	return errors.New("Max retries hit")
 }
 
-func (c ipamClient) hostBlockPairs(ctx context.Context, pool net.IPNet) (map[string]string, error) {
-	pairs := map[string]string{}
+func (c ipamClient) hostBlockPairs(ctx context.Context, pool net.IPNet) (map[string]model.AffinityConfig, error) {
```
caseydavenport, line 1578: Whenever we change a function to be in terms of AffinityConfig instead of hostnames, we should be updating the name of the function to represent it's new meaning.

```suggestion
func (c ipamClient) affinityConfigsByBlock(ctx context.Context, pool net.IPNet) (map[string]model.AffinityConfig, error) {
```

libcalico-go/lib/ipam/ipam.go
```diff
@@ -1993,12 +2004,15 @@ func (c ipamClient) convertBackendToIPAMConfig(cfg *model.IPAMConfig) *IPAMConfi
 func (c ipamClient) ensureConsistentAffinity(ctx context.Context, b *model.AllocationBlock) error {
 	// Retrieve node for this allocation. We do this so we can clean up affinity for blocks
 	// which should no longer be affine to this host.
-	host := getHostAffinity(b)
-	logCtx := log.WithFields(log.Fields{"cidr": b.CIDR, "host": host})
+	affinityCfg := getHostAffinity(b)
```
caseydavenport, line 2007: Seems like you opted to keep this function scoped only to `host:` type affinities, which I think makes any functions that call it (`ensureConsistentAffinity` and `ReleaseBlockAffinity`) silently fail to work for virtual affinities. 

I think I would rather we:
- Update `getHostAffinity` to handle virtual affinities as well.
- Have `ensureConsistentAffinity` and `ReleaseBlockAffinity` explicitly decide whether they support virtual affinities by checking the `Type` on the returned `AffinityConfig` and taking action based on that. 

It will make this code easier to follow.

caseydavenport: Worth calling out - I think it is correct to skip this function for virtual affinities (probably with a debug log as well).

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -521,45 +497,45 @@ func (c *loadBalancerController) syncService(svcObj serviceObject) error {
...
 // needsIPsAssigned determines if service IPFamilyPolicy is requirement is fulfilled by number of assigned IPs in IPAM storage
-func (c *loadBalancerController) needsIPsAssigned(svc *v1.Service, svcObj serviceObject) bool {
+func (c *loadBalancerController) needsIPsAssigned(svc *v1.Service, svcKey serviceKey) bool {
 	switch *svc.Spec.IPFamilyPolicy {
 	case v1.IPFamilyPolicySingleStack:
-		return len(c.allocationTracker.ipsByService[svcObj]) == 1
+		return len(c.allocationTracker.ipsByService[svcKey]) != 1
 	case v1.IPFamilyPolicyRequireDualStack, v1.IPFamilyPolicyPreferDualStack:
-		return len(c.allocationTracker.ipsByService[svcObj]) == 2
+		return len(c.allocationTracker.ipsByService[svcKey]) != 2
```
caseydavenport, line 505: This isn't a perfect check - ideally we'd also check that for dual stack services, the IPs in our IPAM DB are actually dual-stack.

Otherwise, we might return "true" here when there are two IPv4 IPs in our database (which technically necessitates a new IPv6 address).

this isn't super urgent though - I'd expect that our code would spot the extra IPv4, release it, and then the next sync we'd fail this check.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -521,45 +497,45 @@ func (c *loadBalancerController) syncService(svcObj serviceObject) error {
...
 
+	needsStatusUpdate := false
 	for _, ingress := range svc.Status.LoadBalancer.Ingress {
-		if _, ok := c.allocationTracker.ipsByService[svcObj][ingress.IP]; !ok {
-			break
+		if _, ok := c.allocationTracker.ipsByService[svcKey][ingress.IP]; !ok {
+			needsStatusUpdate = true
 		}
-		return false
 	}
```
caseydavenport, line 523: We also should iterate the `ipsByService` and check that they are all present in the status, not just that all of the IPs in the status are allocated in IPAM.

caseydavenport (commented):
A few comments - I think the libcalico-go changes are moving the right direction but need to go a little bit further in a few places.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
MichalFupso: I think the check to see if length of `ipsByService` and the `Status.LoadBalancer.Ingress` is the same should prevent us from having to iterate over the ipsByService

libcalico-go/lib/ipam/ipam_block.go
```diff
@@ -189,26 +189,21 @@ func (b *allocationBlock) assign(affinityCheck bool, address cnet.IP, handleID *
...
 
 // hostAffinityMatches checks if the provided host matches the provided affinity.
-func hostAffinityMatches(host string, block *model.AllocationBlock) bool {
-	if strings.HasPrefix(*block.Affinity, fmt.Sprintf("%s:", model.AffinityTypeHost)) {
-		return *block.Affinity == fmt.Sprintf("%s:", model.AffinityTypeHost)+host
-	} else if strings.HasPrefix(*block.Affinity, fmt.Sprintf("%s:", model.AffinityTypeVirtual)) {
-		return *block.Affinity == fmt.Sprintf("%s:", model.AffinityTypeVirtual)+host
-	}
-	return false
+func hostAffinityMatches(affinityCfg AffinityConfig, block *model.AllocationBlock) bool {
```
caseydavenport, line 192:
```suggestion
func affinityMatches(affinityCfg AffinityConfig, block *model.AllocationBlock) bool {
```

libcalico-go/lib/ipam/ipam_block_reader_writer_test.go
```diff
@@ -384,8 +384,8 @@ var _ = testutils.E2eDatastoreDescribe("IPAM affine block allocation tests", tes
 						defer wg.Done()
 
 						testhost := "same-host"
-						affinityCfg := AffinityConfig{
```
caseydavenport: We should add tests here and/or in ipam_test.go that check the edge cases related to virtual allocations.

e.g.,

- Allocate IPs to a host named `virtual-loadbalancer`, allocate actual virtual load balancer IPs.
- Release affinities in various orders, make sure only the desired affinities are released.
- Once affinities are released, also delete IPs from the blocks - make sure state is correct for each.

caseydavenport: We should also add tests here that simply check reading / writing virtual affinities and blocks works as expected - i.e., write a block, read it back, compare the two and make sure they match exactly.

caseydavenport (commented):
A few more comments before my morning meetings

libcalico-go/lib/ipam/ipam.go
```diff
@@ -1669,6 +1705,17 @@ func (c ipamClient) releaseByHandle(ctx context.Context, blockCIDR net.IPNet, op
 		if err = c.ensureConsistentAffinity(ctx, block.AllocationBlock); err != nil {
 			logCtx.WithError(err).Warn("Error ensuring consistent affinity but IP already released. Returning no error.")
 		}
+
+		// If this is loadBalancer we delete the block without waiting
+		if *block.Affinity == loadBalancerAffinityHost {
+			block = allocationBlock{obj.Value.(*model.AllocationBlock)}
+			if block.empty() {
+				err = c.blockReaderWriter.deleteBlock(ctx, obj)
+				if err != nil {
```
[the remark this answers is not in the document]
caseydavenport: I think when we last spoke about this, we said we needed to at a minimum add a big comment block explaining why this was OK?

@MichalFupso do you recall the outcome of this?

libcalico-go/lib/ipam/ipam.go
```diff
@@ -2164,13 +2216,13 @@ func getMaxPrefixLen(version int, attrs *HostReservedAttr) (int, error) {
 	return maxPrefixLen, nil
 }
 
-func (c ipamClient) ensureBlock(ctx context.Context, rsvdAttr *HostReservedAttr, requestedPools []net.IPNet, version int, host string) (*net.IPNet, error) {
+func (c ipamClient) ensureBlock(ctx context.Context, rsvdAttr *HostReservedAttr, requestedPools []net.IPNet, version int, affinityCfg AffinityConfig) (*net.IPNet, error) {
 	// This function is similar to autoAssign except it does not allocate ips.
 
-	logCtx := log.WithFields(log.Fields{"host": host})
+	logCtx := log.WithFields(log.Fields{"host": affinityCfg.Host})
```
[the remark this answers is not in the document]
caseydavenport: hehe, looks like I had already commented on this @MichalFupso

caseydavenport (approved):
:shipit: :rocket:

PR #9232, opened 2024-09-14
Introduce default action in tiers

mazdakn:
## Description

Allow users to specify the default action in a tier. The action is applied to workloads selected by policies in the tier, but with no policy deciding the fate of the workload's traffic. This PR introduces a new field to the Tier CRD, named `DefaultAction`. The value can be either `Deny` [default] or `Pass`.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Allow users to specify the default action in a tier.
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/tier.go
```diff
@@ -48,6 +48,10 @@ type TierSpec struct {
 	// last.  Tiers with identical order will be applied in alphanumerical order based
 	// on the Tier "Name".
 	Order *float64 `json:"order,omitempty" protobuf:"bytes,1,opt,name=order"`
+	// EndOfTierAction specifies the action applied to workloads selected by a policy in the tier,
+	// but not rule matched the workload's traffic.
+	// [Default: Deny]
+	EndOfTierAction Action `json:"endOfTierAction,omitempty" validate:"omitempty,oneof=Deny Pass"`
```
mazdakn, line 54: I thin `DefaultAction` is a better name for this field. @caseydavenport WDYT?

caseydavenport: Yeah, I like DefaultAction :+1:

felix/fv/tiered_policy_test.go
```diff
@@ -256,8 +281,28 @@ var _ = infrastructure.DatastoreDescribe("connectivity tests with policy tiers _
 		cc.ExpectNone(ep2_2, ep1_1) // denied by np1-1
 		cc.ExpectNone(ep2_3, ep1_1) // denied by gnp2-2
 
+		return cc
+	}
+
+	It("should test connectivity between workloads with policy tiers", func() {
```
fasaxc, line 287:
```suggestion
	It("should test connectivity between workloads with tier with Pass default action", func() {
```

fasaxc (changes requested):
Looks sound but I'd prefer if you passed the action down as a string through to the dataplane.  Can see us adding another action later (Allow, LogAndDrop, etc) and the translation to a bool adds friction to that.  As a general rule, I prefer to keep all new additions to the model API in-sync with the v3 API to avoid surprises; I'd like to get rid of the model API but keeping it strictly in sync with v3 for anything new at least minimises its impact.

fasaxc:
Oh, one more thing: should there be a change to Dikastes (app-policy)?

api/pkg/apis/projectcalico/v3/tier.go
```diff
@@ -49,6 +49,10 @@ type TierSpec struct {
 	// last.  Tiers with identical order will be applied in alphanumerical order based
 	// on the Tier "Name".
 	Order *float64 `json:"order,omitempty" protobuf:"bytes,1,opt,name=order"`
+	// DefaultAction specifies the action applied to workloads selected by a policy in the tier,
+	// but not rule matched the workload's traffic.
+	// [Default: Deny]
```
caseydavenport, line 54:
```suggestion
	// [Default: Deny]
	// +kubebuilder:validation:Enum=Pass;Deny
```

caseydavenport: This ensures the underlying CRD enforces this as well (just in case someone edits that directly instead of going via our apiserver or calicoctl).

caseydavenport: Strictly speaking I think this should also be a pointer since it's an optional field

libcalico-go/lib/backend/model/tier.go
```diff
@@ -88,5 +88,6 @@ func (options TierListOptions) KeyFromDefaultPath(path string) Key {
 }
 
 type Tier struct {
-	Order *float64 `json:"order,omitempty"`
+	Order         *float64 `json:"order,omitempty"`
+	DefaultAction string   `json:"defaultAction,omitempty"`
```
caseydavenport, line 92: Perhaps I misinterpreted Shaun's comment, but I think he meant for `DefaultAction` to be a `v3.Action` type and not strictly a `string` type on the model API.

>  I prefer to keep all new additions to the model API in-sync with the v3 API to avoid surprises

I think the main thing was that it shouldn't be a `bool`

caseydavenport (approved):

PR #9246, opened 2024-09-18
Various fixes to (hash)release builder

caseydavenport:
## Description

- Fix cni-plugin release-verify to follow the same pattern as other directories.
- Add developer guide for building local releases.
- Add option to specify custom image registry.
- Rename cfg.ValidArches to match underlying implementation (VALIDARCHES is sematnically different)
- Don't error out if pinned_version.yaml already exists (make it easier to develop)
- Fix `make release-windows` being run in the wrong directory (IMPORTANT, NEEDS CHERRY PICK)
- Move hashrelease form-factor adjustment out of After() since we don't want it to run if the build fails. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/build/main.go
+++ b/release/build/main.go
@@ -23,7 +23,6 @@ import (
 	"gopkg.in/natefinch/lumberjack.v2"
 
 	"github.com/projectcalico/calico/release/internal/config"
-	"github.com/projectcalico/calico/release/internal/hashrelease"
 	"github.com/projectcalico/calico/release/internal/registry"
 	"github.com/projectcalico/calico/release/internal/utils"
 	"github.com/projectcalico/calico/release/internal/version"
@@ -39,6 +38,8 @@ const (
 	skipImageScanFlag  = "skip-image-scan"
 	publishBranchFlag  = "git-publish"
 	buildImagesFlag    = "build-images"
+
+	imageRegistryFlag = "dev-registry"
 
 	// Configuration flags for the release publish command.
 	skipPublishImagesFlag    = "skip-publish-images"
@@ -141,6 +142,7 @@ func hashreleaseSubCommands(cfg *config.Config, runner *registry.DockerRunner) [
 			Flags: []cli.Flag{
 				&cli.BoolFlag{Name: skipValidationFlag, Usage: "Skip pre-build validation", Value: false},
 				&cli.BoolFlag{Name: buildImagesFlag, Usage: "Build images from local codebase. If false, will use images from CI instead.", Value: false},
+				&cli.StringFlag{Name: imageRegistryFlag, Usage: "Specify image registry to use, for development", Value: ""},
 			},
 			Action: func(c *cli.Context) error {
 				configureLogging("hashrelease-build.log")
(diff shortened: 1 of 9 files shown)


release/pkg/tasks/hashrelease.go
```diff
@@ -161,6 +162,11 @@ func HashreleaseValidate(cfg *config.Config, skipISS bool) {
 // CheckIfHashReleasePublished checks if the hashrelease has already been published.
 // If it has, the process is halted.
 func CheckIfHashReleasePublished(cfg *config.Config, hash string) {
+	if cfg.DocsHost == "" || cfg.DocsUser == "" || cfg.DocsKey == "" || cfg.DocsPort == "" {
+		logrus.Info("Missing hashrelease server configuration, skipping remote hashrelease check")
+		return
+	}
```
radTuti, line 168: i think this should error out unless it is a dev publish

release/internal/config/config.go
```diff
@@ -29,8 +29,8 @@ type Config struct {
 	// OperatorConfig is the configuration for Tigera operator
 	OperatorConfig operator.Config
 
-	// ValidArchs are the OS architectures supported for multi-arch build
-	ValidArchs []string `envconfig:"VALID_ARCHES" default:"amd64,arm64,ppc64le,s390x"`
+	// Arches are the OS architectures supported for multi-arch build
+	Arches []string `envconfig:"ARCHES" default:"amd64,arm64,ppc64le,s390x"`
```
radTuti, line 33: `VALID_ARCHES` is based on the old hashreleases

release/build/main.go
```diff
@@ -172,38 +180,9 @@ func hashreleaseSubCommands(cfg *config.Config, runner *registry.DockerRunner) [
...
-				}
-
-				// Copy the operator tarball to tigera-operator.tgz
-				operatorTarball := filepath.Join(dir, fmt.Sprintf("tigera-operator-%s.tgz", ver))
-				operatorTarballDst := filepath.Join(dir, "tigera-operator.tgz")
-				if err := utils.CopyFile(operatorTarball, operatorTarballDst); err != nil {
-					return err
-				}
+				// Adjsut the formatting of the generated outputs to match the legacy hashrelease format.
+				tasks.ReformatHashrelease(cfg, dir)
```
radTuti, line 185: probably skip this for dev build since this is a legacy support.

release/internal/config/config.go
caseydavenport: In the Calico repo, VALIDARCHES is an existing option that is the output from `ARCHES` and `EXCLUDE_ARCHES`

https://github.com/projectcalico/calico/blob/master/lib.Makefile#L34

So, it might have been correct in the old repo but here VALIDARCHES has a different meaning, and the behavior of this field is more aligned with the existing "ARCHES" config option.

release/build/main.go
caseydavenport: I'm not sure about that - for one, I think I'd need to add another flag to configure that. Also, if I am building this as a developer I want to make sure that what I am running is as close to the "real" code as possible - what if my changes break this logic?

release/pkg/tasks/hashrelease.go
caseydavenport: How would you suggest I determine this is a dev publish? I am using the presence of these fields to determine if we're running in the CI system or not.

Also, the existing logic doesn't error out if these fields are not present (it logs an error and continues).

radTuti: to check if we are running in a CI, you can check for the env variable `CI` so something like
```suggestion
	if cfg.DocsHost == "" || cfg.DocsUser == "" || cfg.DocsKey == "" || cfg.DocsPort == "" {
		log := logrus.Info
		if os.Getenv("CI") == "true" {
			log=logrus.fatal
		}
		log("Missing hashrelease server configuration, skipping remote hashrelease check")
		return
	}
```

caseydavenport: cool, thanks for the tip!

radTuti (approved):

PR #9254, opened 2024-09-19
Support AdminNetworkPolicy named ports

mazdakn:
## Description

Add support for AdminNetworkPolicy named ports.

We default to `tcp` protocol for named ports: https://github.com/projectcalico/calico/blob/1e7b062f779b5fbae151c0a5a42dc1ba27dde304/felix/calc/rule_scanner.go#L388
and we match it differently for named ports here: https://github.com/projectcalico/calico/blob/1e7b062f779b5fbae151c0a5a42dc1ba27dde304/felix/labelindex/named_port_index.go#L122

## Related issues/PRs
AdminNetworkPolicy core features: https://github.com/projectcalico/calico/pull/9206

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add support for AdminNetworkPolicy named ports.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/k8s/conversion/conversion.go
+++ b/libcalico-go/lib/backend/k8s/conversion/conversion.go
@@ -600,10 +600,10 @@ func unpackANPPorts(k8sPorts *[]adminpolicy.AdminNetworkPolicyPort) (map[string]
 		}
 
 		// Named ports do not have protocol
-		if protocol == nil {
+		/*if protocol == nil {
 			protocolPorts[""] = []numorstring.Port{*calicoPort}
 			continue
-		}
+		}*/
 
 		pStr := protocol.String()
 		// treat nil as 'all ports'
@@ -763,6 +763,11 @@ func k8sAdminPolicyPortToCalicoFields(port *adminpolicy.AdminNetworkPolicyPort) 
 	if port.NamedPort != nil {
 		// For named ports, protocol is nil.
 		dstPort, err = k8sAdminPolicyNamedPortToCalico(*port.NamedPort)
+		if err != nil {
+			return
+		}
+		proto := numorstring.ProtocolFromString(numorstring.ProtocolTCP)
+		protocol = &proto
 		return
 	}
 	return


libcalico-go/lib/backend/k8s/conversion/conversion.go
```diff
@@ -545,11 +546,12 @@ func k8sANPEgressRuleToCalico(rule adminpolicy.AdminNetworkPolicyEgressRule) ([]
 		}
 
 		// Based on specifications at least one Peer is set.
-		var selector, nsSelector string
 		for _, peer := range rule.To {
+			var selector, nsSelector string
+			var nets []string
+			// One and only one of the following fileds is set (based on specification).
```
mazdakn, line 552: Fields

libcalico-go/lib/backend/k8s/conversion/conversion.go
```diff
@@ -558,6 +560,20 @@ func k8sANPEgressRuleToCalico(rule adminpolicy.AdminNetworkPolicyEgressRule) ([]
...
+					if err != nil {
+						return nil, err
+					}
+					nets = append(nets, ipNet.String())
+				}
+				found = true
+			}
+			if peer.Nodes != nil {
+				// TODO: Can we translate nodes to auto-hep just with a selector?
+				selector = k8sSelectorToCalico(peer.Nodes, SelectorNode)
```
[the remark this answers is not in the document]
fasaxc: Yes, IIUC, the auto-hep controller passes node labels through to the HEP and it adds an extra "created by auto-hep controller" label.  So, if you match on the labels as written _and_ you add an extra clause to match that created-by label, it should "just work" as long as you've got auto-HEPs enabled. 

Will need to document that it only works with auto-HEPs enabled, which is a little bit confusing.  Perhaps we should document that we may change the implementation to not require auto-HEPs in future.

libcalico-go/lib/backend/k8s/conversion/conversion.go
```diff
@@ -558,6 +559,16 @@ func k8sANPEgressRuleToCalico(rule adminpolicy.AdminNetworkPolicyEgressRule) ([]
 				nsSelector = k8sSelectorToCalico(&peer.Pods.NamespaceSelector, SelectorNamespace)
 				found = true
 			}
+			if len(peer.Networks) != 0 {
+				for _, n := range peer.Networks {
+					_, ipNet, err := cnet.ParseCIDR(string(n))
+					if err != nil {
+						return nil, err
```
fasaxc, line 566:
```suggestion
						return nil, fmt.Errorf("invalid CIDR in ANP rule: %w", err)
```

libcalico-go/lib/backend/k8s/conversion/conversion.go
```diff
@@ -633,10 +638,25 @@ func k8sAdminPolicyPortToCalicoFields(port *adminpolicy.AdminNetworkPolicyPort)
 		protocol = k8sProtocolToCalico(&proto)
 		return
 	}
-	// TODO: Add support for NamedPorts
+	if port.NamedPort != nil {
+		dstPort, err = k8sAdminPolicyNamedPortToCalico(*port.NamedPort)
+		if err != nil {
+			return
+		}
+		proto := kapiv1.ProtocolTCP
```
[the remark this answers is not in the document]
fasaxc: I don't think we should need a protocol for a named port.  In NP there's an implicit default of TCP but there's no field in ANP so I think the idea is to match the named port, whatever protocol it uses.  so, you could match "dns" and that will match pods that use TCP or UDP in their named port.   so, I think you should set proto to nil and handle that further up the stack if needed.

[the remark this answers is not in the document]
fasaxc: I'm not sure if the dataplane can handle named ports with no protocol match.  Is an IP set port match allowed when there's no protocol specified?

We could work around  that by generating a rule per protocol for UDP, TCP, SCTP but obviously that wouldn't be ideal.

fasaxc (changes requested):

libcalico-go/lib/backend/k8s/conversion/conversion.go
```diff
@@ -417,6 +417,12 @@ func k8sANPIngressRuleToCalico(rule adminpolicy.AdminNetworkPolicyIngressRule) (
 			break
 		}
 
+		// Named ports do not have protocol
+		if protocol == nil {
+			protocolPorts[""] = []numorstring.Port{*calicoPort}
+			continue
+		}
+
```
mazdakn, line 425: @fasaxc Originally I handled it like this, but if a rule combines named ports with other ports, this way a separate rule would generate for named ports.

fasaxc: I _think_ that's the correct interpretation.  Question is: does the dataplane handle that correctly (I'm not sure if you can match on ports without specifying a protocol so possible we fail or default it to TCP or something)?

mazdakn: The conformance tests are passing, and those tests include named ports with both udp and tcp. From what I've seen, it seems the protocol does not matter for named ports. The underlying struct does not include a protocol value.

mazdakn: It looks we default to TCP here: https://github.com/projectcalico/calico/blob/d377c71518daf297fa7eb90711d7cfe2a8f92999/felix/calc/rule_scanner.go#L390

Also the conformance tests only include tcp protocols case. I think we should default to tcp here, and explicitly state that in our docs. WDYT?

libcalico-go/lib/backend/k8s/conversion/conversion.go
```diff
@@ -754,7 +754,15 @@ func k8sAdminPolicyPortToCalicoFields(port *adminpolicy.AdminNetworkPolicyPort)
 		protocol = k8sProtocolToCalico(&proto)
 		return
 	}
-	// TODO: Add support for NamedPorts
+	if port.NamedPort != nil {
+		dstPort, err = k8sAdminPolicyNamedPortToCalico(*port.NamedPort)
+		proto := numorstring.ProtocolFromString(numorstring.ProtocolTCP)
```
mazdakn, line 759: We default to `tcp` in Felix for named ports.

libcalico-go/lib/backend/k8s/conversion/conversion.go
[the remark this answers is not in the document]
mazdakn: Auto heps will be covered in another PR.

PR #9255, opened 2024-09-19
add pipeline file validation

radTuti:
## Description

This ensure that all pipeline files are valid. As some are not run until promotions or via tasks, this catches any error early on.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


hack/pipeline/main.go
```diff
@@ -0,0 +1,125 @@
...
+	return dirs, err
+}
+
+func getYAMLFiles(dir string) ([]string, error) {
+	var files []string
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		// Skip the YAML .semaphore/semaphore.yml.d directory
```
caseydavenport, line 36: We want code comments to be providing additional context - it's obvious from the code that this line skips the semaphore.yml.d directory. The important thing the comment should explain is _why_ the code skips that directory.

Makefile
```diff
@@ -35,11 +35,23 @@ clean:
...
 	$(MAKE) check-language
 	$(MAKE) generate
 	$(MAKE) check-dirty
 
+pipeline-validator: hack/pipeline/*.go
+	docker run --rm \
+		-v $(CURDIR):/calico:rw \
+		-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
+		-w /calico/hack/pipeline \
+		$(CALICO_BUILD) go build -o /calico/bin/pipeline-validator .
```
caseydavenport, line 49: we could just make this `go run` instead of `go build` and it would save us a `make` target.

hack/pipeline/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/projectcalico/calico/hack/pipeline
```
caseydavenport, line 1: I don't think we should be using a separate go.mod / go.sum - both of these dependencies are already in our root go module.

danudey (approved):
lgtm!

Makefile
```diff
@@ -35,11 +35,23 @@ clean:
 	rm -rf ./bin
 
 ci-preflight-checks:
+	$(MAKE) check-pipelinefiles
```
caseydavenport, line 38: I don't think it makes sense to run this on every PR

We only need to execute this when we detect that pipeline files change (and strictly speaking, only the pipline files that have actually changed, if we can determine that)

caseydavenport (changes requested):

Makefile
radTuti: it will be a lot more cumbersome trying to check which pipeline files have changed as they are spread out across the repo.
It does not take a long time to validate all the pipeline files

caseydavenport: I think we should do whatever work we need to avoid that cumbersome logic, then - e.g., standardizing on names + locations. We try pretty hard to avoid running CI jobs that test unmodified code, so we should be doing the same here.

caseydavenport: How many directories would we need to check?

PR #9288, opened 2024-09-27
[BPF] decap VXLAN at host ingress - skip vxlan.calico

tomastigera:
## Description

    When we see a vxlan packet entering a node and if the packet belongs to
    calico, that is, it came from a calico node to a calico node with the
    right VNI, we decap the packet and deliver it where it belongs. This
    allows us to bypass the vxlan.calico device on ingress to the system,
    avoid UDP processing on the host, processing on the vxlan device and
    redirection from there. If packet is from a cluster node or a pod
    without any nat, it can go straight to the peer side of the pod iface.
    
    Note that this would break RPF checks if BPFEnforceRPF=Strict, therefore
    it is not performed when that option is set to Strict and packets take
    the regular path.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


tomastigera:
Closing in favor of https://github.com/projectcalico/calico/pull/9965 as that addressed the performance issues without the need to decap in ebpf.

PR #9307, opened 2024-10-04
[BPF]Fix memory leak when there is a pod churn

sridhartigera:
## Description

When there is a workload update, we add an entry to the nameToIface map. When there is a workload remove, we set the interface info to 0 and delete the entry. In 3.28, new fields were added to interface info, which were not reset to 0 on workload remove. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fixed memory leak in BPF endpoint manager.
```

## Reminder for the reviewer

Other optional labels:


tomastigera (approved):
thanks

PR #9323, opened 2024-10-09
Update dispatch logic in nftables

caseydavenport:
## Description

Update dispatch logic in nftables mode to use verdict maps instead of a list of dispatch chains.

Using verdict maps allows us to look up the correct policy rules to execute in constant time, instead of traversing a list of chains proportional to the number of locally running pods. 

I expect the savings here to be minimal in most cases - iptables mode already mitigates this by using a dispatch trie to minimize the number rules traversed. Still worthwhile, especially as the number of pods per node some users are pushing grows over time.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/int_dataplane.go
```diff
@@ -2292,6 +2307,25 @@ func (d *InternalDataplane) apply() {
 		}(ipSets)
 	}
 
+	// Track if we need to perform additional map updates after tables have been programmed.
+	// This is because there is a bidirectional dependency between maps and rules.  We need to
+	// program maps in case any rule references them, and we also need to update map members which
+	// reference rules after the rules have been programmed.
```
caseydavenport, line 2313: Could probably get rid of this is we moved the maps programming logic into the table struct instead of having a separate component doing it - then we'd program everything as a single transaction.

That's probably preferable to this, but this was easier to mock up.

caseydavenport: I coded up this modification and then accidentally checked out over the code :sob: 

Main thing is that order of operations in the transaction needs to be correct -

- Create table
- Create empty maps as needed
- Create rules
- Add map elements that reference rules

And cleanup of course would be the reverse.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -904,6 +904,11 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		}
 	}
 
+	var nftMaps nftables.MapsDataplane
+	if config.RulesConfig.NFTables {
+		nftMaps = nftablesV4RootTable.(nftables.MapsDataplane)
```
fasaxc, line 909: Think I'd have `nftables.NewTable()` return the concrete type so you don't need to cast; not like  `nftablesV4RootTable` could be an iptables table, it's explicitly the nftables one.

felix/nftables/maps.go
```diff
@@ -0,0 +1,677 @@
...
+
+	// mapNameToAllMetadata contains an entry for each map that has been
+	// added by a call to AddOrReplaceMap (and not subsequently removed).
+	mapNameToAllMetadata map[string]MapMetadata
+
+	// mapNameToProgrammedMetadata tracks the maps that we want to program and
+	// those that are actually in the dataplane.  It's Desired() map is the
+	// subset of mapNameToAllMetadata.
+	// Its Dataplane() map contains all maps matching the IPVersionConfig
+	// that we think are in the dataplane.  This includes any temporary IP
```
fasaxc, line 70:
```suggestion
	// that we think are in the dataplane.  This includes any temporary
```
Few bits of IP set terminology left over?

felix/nftables/maps.go
```diff
@@ -0,0 +1,677 @@
...
+		// If it's not in the dataplane, clean it up immediately.
+		logrus.WithField("id", mapName).Debug("map to remove not in the dataplane.")
+		delete(s.mainSetNameToMembers, mapName)
+	}
+	s.updateDirtiness(mapName)
+}
+
+// nameForMainMap takes the given map ID and returns the name of the map as seen in nftables. This
+// helper should be used to sanitize any map IDs, ensuring they are a consistent format.
+func (s *Maps) nameForMainMap(setID string) string {
```
fasaxc, line 238: setID/mapID?

felix/nftables/maps.go
```diff
@@ -0,0 +1,677 @@
...
+		go func(name string) {
+			// Make sure to indicate that we're done by removing ourselves from the limiter channel.
+			defer func() { <-routineLimit }()
+
+			elems, err := s.nft.ListElements(ctx, "map", name)
+			if err != nil {
+				mapsCh <- mapData{name: name, err: err}
+				return
+			}
+			mapsCh <- mapData{name: name, elems: elems}
```
fasaxc, line 358: Think this will deadlock if there are >100 maps.  There's nothing reading from this chan until after the loop finishes.

Worth checking the sets code for similar if this is copied from there.

felix/nftables/maps.go
```diff
@@ -0,0 +1,677 @@
...
+			return mapData.err
+		}
+
+		// TODO: We need to be able to extract the map type from the dataplane, otherwise we cannot
+		// tell whether or not the map has the correct type.
+		metadata, ok := s.mapNameToAllMetadata[mapName]
+		if !ok {
+			// Programmed in the data plane, but not in memory. We should still load any members of this map in order
+			// to perform map deletion logic (delete members, delete map).
+			logCxt.Info("Map in dataplane but not in memory, will remove it.")
```
fasaxc, line 383: Was expecting to see you add something to `mapNameToProgrammedMetadata.Dataplane()` to tell the later logic that there's something we're not expecting.

fasaxc (changes requested):

felix/nftables/maps.go
```diff
@@ -0,0 +1,677 @@
...
+	// have entries in the Desired() map).
+	mapNameToProgrammedMetadata *deltatracker.DeltaTracker[string, MapMetadata]
+
+	// mainSetNameToMembers contains entries for all maps that are in
+	// mapNameToAllMetadata along with entries for "main" (non-temporary)
+	// maps that we think are still in the dataplane. For maps that are in mapNameToAllMetadata, the
+	// Desired() side of the tracker contains the members that we've been told
+	// about.  Otherwise, Desired() is empty.  The Dataplane() side of the
+	// tracker contains the members that are thought to be in the dataplane.
+	mainSetNameToMembers map[string]*deltatracker.SetDeltaTracker[MapMember]
```
fasaxc, line 81: I think using a `SetDeltaTracker` is subtly broken here.  If the value of the KV pair changes then that will be a distinct entry in the set, but only one entry in the kernel due to the shared key.  That matters when loading data from the kernel, you'll get two entries in the Dataplane side of the tracker instead of one and you'll queue up an add + deletion instead of an update.

Only spotted it because I started looking into https://github.com/projectcalico/calico/issues/9450 today and I fell into a very similar trap there.

felix/dataplane/linux/endpoint_mgr.go
```diff
@@ -679,6 +684,12 @@ func (m *endpointManager) resolveWorkloadEndpoints() {
 		delete(m.activeWlEndpoints, id)
 	}
 
+	if m.bpfEnabled && m.maps != nil {
+		// Remove nftables maps if running in BPF mode, as dispatch chains are not used.
+		m.maps.RemoveMap(rules.NftablesFromWorkloadDispatchMap)
```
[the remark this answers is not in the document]
caseydavenport: Hm, yeah I think it should. This may be a holdover from an earlier version...

felix/nftables/maps.go
```diff
@@ -0,0 +1,677 @@
...
+	if dt == nil {
+		dt = deltatracker.NewSetDeltaTracker[MapMember]()
+		s.mainSetNameToMembers[mainMapName] = dt
+	}
+	return dt
+}
+
+// RemoveMap queues up the removal of an map, it need not be empty.  The maps will be
+// removed on the next call to ApplyDeletions().
+func (s *Maps) RemoveMap(setID string) {
```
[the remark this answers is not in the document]
caseydavenport: In this current iteration, no I don't think so.

felix/nftables/maps.go
caseydavenport: @fasaxc I'm not sure that's actually a problem in this code though? Maybe I am misunderstanding, but sounds like it will generate both an add + delete, but since we program those in the same nftables transaction, and also order them such that the delete comes before the add, the dataplane should be updated correctly.

fasaxc: Ah yep, ordering the updates covers the dataplane side of it, I think.  The VXLAN FDB did the updates the other way around.

Could we trip ourselves up on the Desired side of the map in future?  I'm thinking that the caller changes the verdict that they want and that results in both `key -> verdict1` and `key -> verdict2` in the set (when the dataplane can only handle one of those at a time).

fasaxc: Answer is no right now because the API doesn't allow individual updates, I think.

felix/nftables/ipsets.go
```diff
@@ -413,23 +413,26 @@ func (s *IPSets) tryResync() error {
...
-			elems, err := s.nft.ListElements(ctx, "set", name)
-			if err != nil {
-				setsChan <- setData{setName: name, err: err}
-				return
-			}
-			setsChan <- setData{setName: name, elems: elems}
-		}(setName)
-	}
+
+	go func() {
```
fasaxc, line 417: This works; if I was writing it from scratch, I'd probably use the `errgroup` package now that's in the std lib.

fasaxc (commented):
Looking good, think you were planning some more tests?

felix/nftables/ipsets.go
caseydavenport: Neat, I had missed this one. Updated to use errgroup now.

caseydavenport:
@fasaxc thanks for the reviews! I've updated to use `errgroup` as well as adding some new UTs for the map code + table integration.

felix/nftables/ipsets.go
```diff
@@ -404,47 +405,46 @@ func (s *IPSets) tryResync() error {
...
 	defer close(routineLimit)
-	for _, setName := range sets {
+
+	// Create an errgroup to wait for all the set reads to complete.
+	g, ctx := errgroup.WithContext(ctx)
+	responses := make([]setData, len(sets))
+
+	for i, name := range sets {
 		// Wait for room in the limiting channel.
 		routineLimit <- struct{}{}
```
fasaxc, line 421: errgroup has a built-in limiter, just need to use `g.SetLimit()` and then you can get rid of the custom channel

felix/nftables/ipsets.go
```diff
@@ -404,47 +405,46 @@ func (s *IPSets) tryResync() error {
...
 
+		// Capture the name in a closure.
+		name := name
```
fasaxc, line 424: I don't think this is needed as of Go 1.22?

felix/nftables/ipsets.go
```diff
@@ -404,50 +405,46 @@ func (s *IPSets) tryResync() error {
...
 	// avoid spawning too many goroutines if there are a large number of sets.
 	routineLimit := make(chan struct{}, 100)
 	defer close(routineLimit)
 
-	go func() {
-		for _, setName := range sets {
-			// Wait for room in the limiting channel.
-			routineLimit <- struct{}{}
+	// Create an errgroup to wait for all the set reads to complete.
+	g, ctx := errgroup.WithContext(ctx)
```
fasaxc, line 416:
```suggestion
	g, egCtx := errgroup.WithContext(ctx)
```
I think you should treat the errgroup's context as a child of the timeout one; it gets cancelled after an error so it might not outlive the errgroup.  (Doesn't actually matter here since you're returning early after the errgroup error but it seems a bit off to keep using the errgroup context after we're done with the errgroup.)

felix/nftables/maps.go
```diff
@@ -318,50 +323,45 @@ func (s *Maps) LoadDataplaneState() error {
...
+	// Create an errgroup to manage the fleet of goroutines.
+	g, ctx := errgroup.WithContext(ctx)
 	responses := make([]mapData, len(maps))
-	for i := range responses {
-		mapData := <-mapsCh
-		responses[i] = mapData
+
+	for i, name := range maps {
+		// Wait for room in the limiting channel.
+		routineLimit <- struct{}{}
```
fasaxc, line 339: Same here, can use the built-in limit

fasaxc (changes requested):

fasaxc (approved):

PR #9277, opened 2024-09-25
Formatting code with gofmt

cuishuang:
## Description

Formatting code with gofmt

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Formatting code with gofmt
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #9339, opened 2024-10-14
Add tooling to consistently format Go imports

fasaxc:
## Description

- Split `make fix` into `make fix-changed` and `make fix-all`
  - Use `fix-all` in CI, it runs no slower on semaphore than the old version.
  - `fix-changed` should be faster when run locally.
- `make fix-changed` only formats files that have changed on this branch. 
  - The parent branch is guessed based on distance to the merge base with master and release branches.
  - The remote name is auto-detected from metadata.mk.
- Remove check-fmt target in favour of running a format pass and checking for dirty in CI (hard to do check-fmt now that two commands need to be run in sequence).

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


nelljerram:
@fasaxc While you're in the area, I wonder if you have any thoughts about https://github.com/projectcalico/calico/pull/9277 - i.e. about why our `make fix` doesn't make the changes that are proposed in that PR.

hack/format-all-files.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+hack_dir="$(dirname $0)"
+repo_dir="$(dirname $hack_dir)"
+
+find . -iname "*.go" ! -wholename "./vendor/*" -print0 | \
+  xargs -0 go run "${repo_dir}/hack/cmd/coalesce-imports" -w
+find . -iname "*.go" ! -wholename "./vendor/*" -print0 | \
+  xargs -0 goimports -w -local github.com/projectcalico/calico/
```
caseydavenport, line 11: This is going to need to change in enterprise no? Can we / should we leverage metadata.mk to handle this since we've already got conditional logic for the GIT_REMOTE?

api/Makefile
```diff
@@ -84,6 +84,10 @@ gen-files .generate_files: lint-cache-dir clean-generated
 	touch .generate_files
 	$(MAKE) fix
 
+fix:
```
caseydavenport, line 87: Why do we need a distinct fix target her instead of using the common one? Might be worth a comment at least since this seems like an exceptional case.

caseydavenport (approved):
LGTM but a couple small comments/questions.

hack/format-all-files.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+hack_dir="$(dirname $0)"
+repo_dir="$(dirname $hack_dir)"
+
+find . -iname "*.go" ! -wholename "./vendor/*" -print0 | \
```
fasaxc, line 8: @rene-dekker points out that we need an exclusion for the `third-party` folder too

hack/format-all-files.sh
fasaxc: Ah yep, good catch

fasaxc: No, I think enterprise has the same Go package for calico itself, it's only the API that thinks of itself as `tigera/api`

api/Makefile
fasaxc: The new scripts get broken by extracting the API into its own repo (and the builds of the API deliberately _don't_ mount the API into the build container).  I thought about copying the scripts over but it seemed quite messy to duplicate them all.

hack/format-all-files.sh
fasaxc: Oh, there's the third-party directory that Rene pointed out, I'll just add that in here since it should do no harm

PR #9351, opened 2024-10-16
[release tool] support for forks

radTuti:
## Description

This adds the last bits for being able to use developer forks.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/internal/config/config.go
+++ b/release/internal/config/config.go
@@ -44,6 +44,9 @@ type Config struct {
 	// GitRemote is the remote for the git repository
 	GitRemote string `envconfig:"GIT_REMOTE" default:"origin"`
 
+	// GitRepo is the repository name
+	GitRepo string `envconfig:"GIT_REPO_NAME" default:"calico"`
+
 	// Operator is the configuration for Tigera operator
 	Operator OperatorConfig
 
--- a/release/pkg/manager/branch/manager.go
+++ b/release/pkg/manager/branch/manager.go
@@ -132,7 +132,7 @@ func (b *BranchManager) CutReleaseBranch() error {
 		return err
 	}
 	if b.publish {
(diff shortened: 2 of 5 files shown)


release/internal/config/config.go
```diff
@@ -44,6 +44,9 @@ type Config struct {
 	// GitRemote is the remote for the git repository
 	GitRemote string `envconfig:"GIT_REMOTE" default:"origin"`
 
+	// GitRepo is the repository name
+	GitRepo string `envconfig:"GIT_REPO_NAME" default:"calico"`
```
caseydavenport, line 48: I think this should be a CLI argument

release/pkg/manager/branch/manager.go
```diff
@@ -132,7 +132,7 @@ func (b *BranchManager) CutReleaseBranch() error {
 		return err
 	}
 	if b.publish {
-		if _, err := b.git("push", b.mainBranch); err != nil {
+		if _, err := b.git("push", "-u", b.remote, b.mainBranch); err != nil {
```
caseydavenport, line 135: If we're specifying the remote, I don't think we need `-u`?

release/internal/config/config.go
radTuti: As usage of this value do eventually require the Organization (which isn't a CLI arg), I will prefer to keep this 
This is similar to what we have with operator presently.

caseydavenport: I think they should just both be CLI args? We can use the env var support of the CLI tool framework if this is needed in CI (but I don't think it is?)

CLI arguments are self-documenting, so using those where possible should be a priority.

caseydavenport (approved):

PR #9356, opened 2024-10-17
[BPF] Do conntrack expiry using BPF program

fasaxc:
## Description

Move conntrack cleanup to a dedicated BPF program, which is called synchronously from Felix.  the program:

- Iterates the conntrack map.
- Deletes expired "normal" entries as they are encountered.
- Copies NAT entries to a scratch map so that the forward/reverse entries can be deleted together, minimising the window where only one entry is present.

## Related issues/PRs

CORE-10594

## Todos

- [x] Use correct debug level program
- [x] Handle older kernels with no iteration support
- [x] Tests
  - [x] UT 
  - [x] Check that BPF expiry is actually used
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
eBPF: (on kernels that support it) conntrack expiry scanning is now done from a dedicated BPF program rather than from userspace. This is more efficient and reduces the possibility of races with the dataplane programs.
```

## Reminder for the reviewer

Other optional labels:


tomastigera (commented):
looks pretty good, some suggestions

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -0,0 +1,299 @@
...
+		struct calico_ct_key, // key = NAT rev key
+		struct calico_ct_key, // value = NAT fwd key
+		100000,
+		BPF_F_NO_PREALLOC
+);
+
+// sub_age calculates now-then assuming that the difference is less than
+// 1<<63.  Values larger than that are assumed to have wrapped (then>now) and
+// 0 is returned in that case.
+static __u64 sub_age(__u64 now, __u64 then) {
```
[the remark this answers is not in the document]
fasaxc: I have a mild preference for same line (just saves vertical space) but happy to be consistent.  Would love to have a C code formatter like we do for Go.

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -0,0 +1,299 @@
...
+	CALI_DEBUG("Scanning conntrack map for expired non-NAT entries...");
+	bpf_for_each_map_elem(&CT_MAP_V, process_ct_entry, &ictx, 0);
+	CALI_DEBUG("First pass complete, expired %d entries so far of %d total.", ictx.num_expired, ictx.num_seen);
+	CALI_DEBUG("Processing NAT entries...");
+	bpf_for_each_map_elem(&CCQ_MAP_V, process_ccq_entry, &ictx, 0);
+	CALI_DEBUG("Conntrack cleanup complete: expired %d entries of %d total.", ictx.num_expired, ictx.num_seen);
+
+	// Give detailed stats back to userspace.
+	ictx.end_time = bpf_ktime_get_ns();
+	bpf_skb_store_bytes(skb, 0, &ictx, sizeof(ictx), 0);
```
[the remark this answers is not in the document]
fasaxc: Seemed like a good use of the SKB to me :shrug:

felix/bpf/conntrack/bpf_scanner.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	configuredGlobals := false
+	pinnedCTMap := false
+	for m, err := obj.FirstMap(); m != nil && err == nil; m, err = m.NextMap() {
+		// In case of global variables, libbpf creates an internal map <prog_name>.rodata
+		// The values are read only for the BPF programs, but can be set to a value from
+		// userspace before the program is loaded.
+		mapName := m.Name()
+		if m.IsMapInternal() {
+			if strings.HasPrefix(mapName, ".rodata") {
```
tomastigera, line 141: I noticed recently that there was a change in naming and `Contains` is now required

felix/bpf/conntrack/bpf_scanner.go
```diff
@@ -0,0 +1,305 @@
...
+	program, err := s.ensureBPFExpiryProgram()
+	if err != nil {
+		return fmt.Errorf("failed to load BPF program: %w", err)
+	}
+	fd, err := program.ProgramFD("conntrack_cleanup")
+	if err != nil {
+		return fmt.Errorf("failed to look up BPF program section: %w", err)
+	}
+
+	// The BPF program returns its result in the packet buffer, size it accordingly.
```
tomastigera, line 255: This comment relates to line 260

felix/bpf/ut/bpf_ct_cleanup_test.go
```diff
@@ -0,0 +1,86 @@
...
+		if maps.IsNotExists(err) {
+			deletedEntries = append(deletedEntries, k)
+		} else {
+			Expect(err).NotTo(HaveOccurred(), "unexpected error from map lookup")
+		}
+	}
+	return deletedEntries
+}
+
+func clearCTMap() {
```
tomastigera, line 81: There is already `func resetMap(m maps.Map) {`

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -0,0 +1,276 @@
...
+// can be cleaned up together).
+__attribute__((section("tc"))) int conntrack_cleanup(struct __sk_buff *skb)
+{
+	struct ct_iter_ctx ictx = {};
+	bpf_skb_load_bytes(skb, 0, &ictx, sizeof(ictx));
+	if (ictx.now == 0) {
+		// Caller didn't provide a fixed time (as used in tests), use current
+		// time.
+		ictx.now = bpf_ktime_get_ns();
+	}
```
tomastigera, line 258: If it is used in UTs only, perhaps this could be more expressive? But you would need to list it in `bpf-gpl/list-ut-objs` and compile a test version. So whatever is your preference.

```suggestion
#ifndef UNITTEST
		ictx.now = bpf_ktime_get_ns();
	}
#endif
```

felix/fv/bpf_test.go
```diff
@@ -86,11 +86,11 @@ var (
 	_ = describeBPFTests(withTunnel("ipip"), withProto("udp"))
 	_ = describeBPFTests(withProto("tcp"), withDSR())
 	_ = describeBPFTests(withProto("udp"), withDSR())
-	_ = describeBPFTests(withTunnel("ipip"), withProto("tcp"), withDSR())
-	_ = describeBPFTests(withTunnel("ipip"), withProto("udp"), withDSR())
+	_ = describeBPFTests(withTunnel("ipip"), withProto("tcp"), withDSR(), withConntrackCleanupMode("BPFProgram"))
```
tomastigera, line 89: Any benefit of forcing BPFProgram? In our test env, it should always use BPFProgram. Unless you want it to fail in some envs where it does not load?

tomastigera (commented):
Lookg sgreat, mostly nits and considerations.

I would expect that BPFProgram mode would always fall back to userspace, kind of the same as Auto :thinking:

fasaxc:
> I would expect that BPFProgram mode would always fall back to userspace, kind of the same as Auto

I mainly added that for testing, so I could force it into that mode to make sure we don't always fall back.

felix/fv/bpf_test.go
fasaxc: I wanted to force it so that the test would fail if the program failed to load.

felix/bpf-gpl/conntrack_cleanup.c
fasaxc: Decided not to build a UT binary.  Just seemed like overkill for one minor tweak (especially when the conntrack cleaner is not on critical packet path).

PR #9374, opened 2024-10-22
Prometheus iptables rules metric only counts programmed rules.

aaaaaaaalex:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
The `felix_iptables_rules` Prometheus metric now only counts rules within referenced Iptables chains, no longer counts candidate rules.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/pre_dnat_test.go
+++ b/felix/fv/pre_dnat_test.go
@@ -17,6 +17,7 @@
 package fv_test
 
 import (
+	"fmt"
 	"strconv"
 	"time"
 
@@ -24,6 +25,7 @@ import (
 	. "github.com/onsi/gomega"
 	api "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	"github.com/projectcalico/api/pkg/lib/numorstring"
+	"github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/calico/felix/fv/connectivity"
 	"github.com/projectcalico/calico/felix/fv/containers"
@@ -32,6 +34,7 @@ import (
 	"github.com/projectcalico/calico/felix/fv/workload"
 	"github.com/projectcalico/calico/libcalico-go/lib/apiconfig"
 	client "github.com/projectcalico/calico/libcalico-go/lib/clientv3"
+	"github.com/projectcalico/calico/libcalico-go/lib/options"
 )
 
 // Setup for planned further FV tests:
@@ -63,6 +66,8 @@ var _ = infrastructure.DatastoreDescribe("pre-dnat with initialized Felix, 2 wor
 		options := infrastructure.DefaultTopologyOptions()
 		// For variety, run this test with IPv6 disabled.
 		options.EnableIPv6 = false
+		options.ExtraEnvVars["FELIX_PrometheusMetricsEnabled"] = "true"
+
 		tc, client = infrastructure.StartSingleNodeTopology(options, infra)
 
 		// Install a default profile that allows all ingress and egress, in the absence of any Policy.
@@ -209,16 +214,28 @@ var _ = infrastructure.DatastoreDescribe("pre-dnat with initialized Felix, 2 wor
 						policy.Spec.ApplyOnForward = true
 						protocol := numorstring.ProtocolFromString("tcp")
 						ports := numorstring.SinglePort(8055)
+						metricsPort := numorstring.SinglePort(9091)
+
 						policy.Spec.Ingress = []api.Rule{{
 							Action:   api.Allow,
 							Protocol: &protocol,
 							Destination: api.EntityRule{Ports: []numorstring.Port{
 								ports,
+								metricsPort,
 							}},
 						}}
 						policy.Spec.Selector = "has(host-endpoint)"
 						_, err := client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
 						Expect(err).NotTo(HaveOccurred())
+					})
+
+					AfterEach(func() {
+						if CurrentGinkgoTestDescription().Failed {
+							logrus.WithFields(logrus.Fields{
+								"filter": tc.Felixes[0].IPTablesChains("filter"),
+								"mangle": tc.Felixes[0].IPTablesChains("mangle"),
+							}).Debug("dumping iptables")
+						}
 					})
 
 					It("external client cannot connect", func() {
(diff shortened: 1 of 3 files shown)


felix/fv/pre_dnat_test.go
```diff
@@ -229,6 +326,74 @@ var _ = infrastructure.DatastoreDescribe("pre-dnat with initialized Felix, 2 wor
...
+						curFilterRulesMetric, err := filterRulesMetric.Int()
+						Expect(err).NotTo(HaveOccurred(), "Failed to fetch filter rules prom metric")
+
+						mangleRulesMetricDelta := curMangleRulesMetric - baselineMangleRulesMetric
+						filterRulesMetricDelta := curFilterRulesMetric - baselineFilterRulesMetric
+
+						// The number of cali rules in the dumped tables should increase by the same delta as the metrics
+						Eventually(dumpMangleTable).Should(HaveLen(len(baselineMangleTableIptablesSave) + mangleRulesMetricDelta))
+						Eventually(dumpFilterTable).Should(HaveLen(len(baselineFilterTableIptablesSave) + filterRulesMetricDelta))
+					})
```
fasaxc, line 396: I recommend doing some more steps after this:
* Change the type of the policy from pre-DNAT to normal.  That's the most fiddly change to handle since it goes from referenced in one table to referenced in the other.
* Delete the policy, should go back to the count with the hep.
* Delete the hep, should get back to the base count.

felix/fv/pre_dnat_test.go
```diff
@@ -200,6 +206,87 @@ var _ = infrastructure.DatastoreDescribe("pre-dnat with initialized Felix, 2 wor
...
+						}
+						return grep
+					}
+					dumpMangleTable := func() []string { return dumpIPTables("mangle") }
+					dumpFilterTable := func() []string { return dumpIPTables("filter") }
+
+					// Returns nil when the value of f() hasn't changed for the duration settledPeriod.
+					// Returns error if f() never settles up to the timeout.
+					// Immediately returns error if timeout is lower than settledPeriod.
+					waitToSettle := func(f func() int, settledPeriod, timeout time.Duration) error {
```
fasaxc, line 233: As discussed on slack, would prefer not to rely on settling, if there's something explicit that we can wait for.  It's not a bad approach, it's just a bit slow (and if it flakes, you have to bump the settled timeout so it gets slower...)

If we do go down this path, I suggest making it a general utility method.  Kubernetes has a `wait` package with similar utility funcs in it.

felix/fv/pre_dnat_test.go
```diff
@@ -200,6 +206,87 @@ var _ = infrastructure.DatastoreDescribe("pre-dnat with initialized Felix, 2 wor
...
+								if last == cur {
+									settledTimer = time.NewTimer(settledPeriod)
+								}
+							}
+							select {
+							case <-timeoutTimer.C:
+								return errors.New("Timed out waiting for IPTables to settle.")
+							default:
+								last = cur
+								time.Sleep(1 * time.Second)
```
fasaxc, line 268: I typically reach for 100ms for a sleep loop.  We don't want to tight loop but probing 1x per second isn't normally a big deal.

felix/fv/pre_dnat_test.go
```diff
@@ -200,6 +206,87 @@ var _ = infrastructure.DatastoreDescribe("pre-dnat with initialized Felix, 2 wor
...
+
+							if settledTimer != nil {
+								if last != cur {
+									if !settledTimer.Stop() {
+										<-settledTimer.C
+									}
+									settledTimer = nil
+								} else {
+									select {
+									case <-settledTimer.C:
```
fasaxc, line 251: With the way you're using the settledTimer, you're not actually using it as a timer at all!  You only select on it in a non-blocking way.  I think this code would be clearer if you just recorded the time since the last change and did 
```
last := f()
lastChange := time.Now()
for {
  cur := f()
  if cur != last {
    lastChange := time.Now()
  }
  last = cur
  if time.Since(lastChange) > settledPeriod {
    // we're done
  }
  ...
}
```

felix/fv/pre_dnat_test.go
```diff
@@ -200,6 +206,87 @@ var _ = infrastructure.DatastoreDescribe("pre-dnat with initialized Felix, 2 wor
...
+										// Still settled, but not for long enough. Carry on...
+									}
+								}
+							} else {
+								if last == cur {
+									settledTimer = time.NewTimer(settledPeriod)
+								}
+							}
+							select {
+							case <-timeoutTimer.C:
```
fasaxc, line 264: Using a sleep and a timeout channel is slightly mixing paradigms.  While you''re asleep you're not checking the timeout channel, which could be a bug (or not matter at all).  In this case, it means that, if you just miss a timeout, it'll be >1s before you check the timeout again and you'll do an extra poll of `f()`; not clear if that was intended (it probably doesn't matter for this use case).

Best to jump one way or the other to make your intent clear.  Either:
* Use Sleep() and record the start time and do `if time.Since(startTime) > timeoutPeriod { ... timeout ...}`
* Use channels for the sleep.  `case <-time.After(delay): ...` so that the sleep can be interrupted.

fasaxc: If you also wanted the settled timer to interrupt the sleep (not clear that you do, since you probably want to run f() one more time before declaring it settled), then... A powerful technique is to mask channels in your main select block.  To mask a channel, set it to nil; that prevents its case from firing.  So you could do:
```
var settledC chan <-time.Time

for {
  select {
  case <- time.After(time.Second):
    cur := f()
    if cur == last {
      if settledC == nil { 
        settledC = time.After(...)
      }
    } else {
      settledC  = nil
    }
  case <- settledC:
    // done!
  case <-timeoutC:
    // failed
  }
}
```

fasaxc (changes requested):
The ref counting looks correct :tada: some comments on the test.

Thinking about it, you could have tested this at UT level pretty effectively if there's a way to get the prom gauge.  Then you'd be in control of the exact number of rules and wouldn't need approximates.  Not bad to have a bit more coverage of the real gauges though so I'm OK with FV too.

aaaaaaaalex:
@fasaxc done away with the "settled" business now, instead checking for a particular IPTables chain to determine when the dataplane got programmed at start-of-day, and then using "Eventually" calls to wait until additional changes occur.

felix/fv/pre_dnat_test.go
```diff
@@ -229,6 +246,168 @@ var _ = infrastructure.DatastoreDescribe("pre-dnat with initialized Felix, 2 wor
...
+								curMangleRulesMetric, err := mangleRulesMetric.Int()
+								Expect(err).NotTo(HaveOccurred())
+
+								_, err = client.HostEndpoints().Create(utils.Ctx, hep, utils.NoOptions)
+								Expect(err).NotTo(HaveOccurred(), "Failed to create HEP")
+
+								appliedGNP, err = client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
+								Expect(err).NotTo(HaveOccurred(), "Couldn't create pre-DNAT GNP")
+
+								Eventually(mangleRulesMetric.Int, "5s").ShouldNot(BeEquivalentTo(curMangleRulesMetric), "Mangle rules metric never changed following change of GNP")
```
fasaxc, line 327:
```suggestion
								Eventually(mangleRulesMetric.Int, "5s").Should(BeNumerically(">", curMangleRulesMetric), "Mangle rules metric never changed following change of GNP")
```
Could use `>`

felix/fv/pre_dnat_test.go
```diff
@@ -229,6 +246,168 @@ var _ = infrastructure.DatastoreDescribe("pre-dnat with initialized Felix, 2 wor
...
+						operations := []operation{
+							{"Creating a HEP and pre-DNAT GNP", func() {
+								var err error
+								curMangleRulesMetric, err := mangleRulesMetric.Int()
+								Expect(err).NotTo(HaveOccurred())
+
+								_, err = client.HostEndpoints().Create(utils.Ctx, hep, utils.NoOptions)
+								Expect(err).NotTo(HaveOccurred(), "Failed to create HEP")
+
+								appliedGNP, err = client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
```
fasaxc, line 324: You should create the GNP first.  Felix won't program the GNP unless there's a HEP but it'll program the HEP without a GNP.  So, if you do HEP then GNP, you might see an intermediate step, whereas it'll be atomic the other way.

fasaxc (changes requested):

fasaxc (approved):

Issue #8883, opened 2024-06-05
Helm chart does not have proper updates from operator repo for operator types


tmjd:
## Expected Behavior

The helm chart and manifest should have updates pulled from the [operator repo](https://github.com/tigera/operator/) before each release.

## Current Behavior

Master and Calico v3.28 are not updated with the proper definitions for the operator types. A specific example looking at the v3.28 release, is `csi-node-driver` is the [only allowed container name for the csiNodeDriverDaemonSet](https://github.com/projectcalico/calico/blob/release-v3.28/charts/tigera-operator/crds/operator.tigera.io_installations_crd.yaml#L6869) but `csi-node-driver-registrar` and `csi-node-driver` should also be allowed (based on the [CRD in the operator repo](https://github.com/tigera/operator/blob/release-v1.34/pkg/crds/operator/operator.tigera.io_installations.yaml#L5526)). See https://github.com/tigera/operator/issues/3356 for context.

## Possible Solution

Ensure the release process properly pulls in updates from the operator repo before creating new releases. It will also be good to pull in updates to the master branch periodically or anytime creating new releases.

## Steps to Reproduce (for bugs)

This is an issue with getting the manifests/helm chart updated properly for releases, so no steps necessary to reproduce.

## Context

https://github.com/tigera/operator/issues/3356

## Your Environment

Not Applicable

mashiofu-cafex:
Originated from https://github.com/projectcalico/calico/issues/8853 which is now closed to monitor this issue

muckelba:
We've stumbled across the same problem. I've tried to use the original yaml from the operator repo, but that's not perfect yaml:
```sh
kubectl apply --server-side --force-conflicts -f https://github.com/tigera/operator/blob/master/pkg/crds/operator/operator.tigera.io_installations.yaml
error: error parsing https://github.com/tigera/operator/blob/master/pkg/crds/operator/operator.tigera.io_installations.yaml: error converting YAML to JSON: yaml: line 212: mapping values are not allowed in this context
```

TomBillietKlarrio:
Any updates on this? Just stumbled upon it too.

mashiofu-cafex:
Any update on this effort?

caseydavenport:
There's a makefile helper to do this: https://github.com/projectcalico/calico/blob/bb0c3b99e9b746c7e93fd7c7cc468c6eb494dd46/Makefile#L65-L70

[...]

PR #9377, opened 2024-10-22
Check operator CRDs as part of ci preflight checks

caseydavenport:
## Description

Fixes https://github.com/projectcalico/calico/issues/8883

This PR will fail CI if the operator CRDs are not in sync.

This is potentially annoying, as it means we may need to push some
unrelated changes into PRs to get CI working. But I think it is worth
the friction to avoid these getting out of sync.

We have encountered release problems in the past where we shipped
operator CRDs that were out of date.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ check-language:
 
 generate:
 	$(MAKE) gen-semaphore-yaml
+	$(MAKE) get-operator-crds
 	$(MAKE) -C api gen-files
 	$(MAKE) -C libcalico-go gen-files
 	$(MAKE) -C felix gen-files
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -61,12 +62,15 @@ gen-manifests: bin/helm
...
-# Get operator CRDs from the operator repo, OPERATOR_BRANCH_NAME must be set
-get-operator-crds: var-require-all-OPERATOR_BRANCH_NAME
+# Get operator CRDs from the operator repo, OPERATOR_BRANCH must be set
+get-operator-crds: var-require-all-OPERATOR_BRANCH
+	@echo ================================================================
+	@echo === Pulling new operator CRDs from branch ${OPERATOR_BRANCH} ===
+	@echo ================================================================
 	cd ./charts/tigera-operator/crds/ && \
-	for file in operator.tigera.io_*.yaml; do echo "downloading $$file from operator repo" && curl -fsSL https://raw.githubusercontent.com/tigera/operator/${OPERATOR_BRANCH_NAME}/pkg/crds/operator/$${file%_crd.yaml}.yaml -o $${file}; done
+	for file in operator.tigera.io_*.yaml; do echo "downloading $$file from operator repo" && curl -fsSL https://raw.githubusercontent.com/tigera/operator/${OPERATOR_BRANCH}/pkg/crds/operator/$${file%_crd.yaml}.yaml -o $${file}; done
```
danudey, line 71: Looks like this will only update files we have already. Is there any chance we'll need to pull in new files that we don't have locally?

caseydavenport: There is that possibility, but I don't know how we can predict what those files are and the operator repo doesn't (currently) have a way of telling us this information (there are CRDs there that we _don't_ want to pull in for example).

coutinhop, line 71:
```suggestion
	for file in operator.tigera.io_*.yaml; do echo "downloading $$file from operator repo" && curl -fsSL https://raw.githubusercontent.com/tigera/operator/$(OPERATOR_BRANCH)/pkg/crds/operator/$${file%_crd.yaml}.yaml -o $${file}; done
```

Makefile
```diff
@@ -61,12 +62,15 @@ gen-manifests: bin/helm
...
 	cd ./manifests/ocp/ && \
-	for file in operator.tigera.io_*.yaml; do echo "downloading $$file from operator repo" && curl -fsSL https://raw.githubusercontent.com/tigera/operator/${OPERATOR_BRANCH_NAME}/pkg/crds/operator/$${file%_crd.yaml}.yaml -o $${file}; done
+	for file in operator.tigera.io_*.yaml; do echo "downloading $$file from operator repo" && curl -fsSL https://raw.githubusercontent.com/tigera/operator/${OPERATOR_BRANCH}/pkg/crds/operator/$${file%_crd.yaml}.yaml -o $${file}; done
```
coutinhop, line 73:
```suggestion
	for file in operator.tigera.io_*.yaml; do echo "downloading $$file from operator repo" && curl -fsSL https://raw.githubusercontent.com/tigera/operator/$(OPERATOR_BRANCH)/pkg/crds/operator/$${file%_crd.yaml}.yaml -o $${file}; done
```

Makefile
coutinhop: we need `$()` to refer to the Makefile var, not an env var (which was also incorrect, it would have needed to have `$${}`)

coutinhop: so @danudey told me I'm wrong and `${}` is supported (TIL, thanks!), but I still believe we should try to stick to `$()`

caseydavenport: I can swap (although I will note that these predate my involvement in this area of our Makefile!)

metadata.mk
```diff
@@ -46,3 +46,6 @@ WINDOWS_VERSIONS ?= 1809 ltsc2022
 # whenever the cni-plugin image is created.
 CNI_VERSION=master
 FLANNEL_VERSION=main
+
+# The operator branch corresponding to this branch.
+OPERATOR_BRANCH=master
```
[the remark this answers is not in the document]
caseydavenport: Yep - we already needed this reminder though (it was hard coded deep in our semaphore config files)

release/RELEASING.md
```diff
@@ -103,6 +103,7 @@ When starting development on a new minor release, the first step is to create a
 
    - charts/calico/values.yaml
    - charts/tigera-operator/values.yaml
+   - metadata.mk (OPERATOR_VERSION)
```
coutinhop, line 106: `OPERATOR_BRANCH`?

coutinhop (approved):

release/RELEASING.md
caseydavenport: whoopsie

PR #9384, opened 2024-10-23
fix creating release notes for release

radTuti:
## Description

This fixes the current failure with generating release notes. Also updated the logs

```
time="2024-10-23T16:43:25-07:00" level=info msg="Running git command" cmd="/usr/bin/git describe --tags --always --long --abbrev=12 --dirty" dir=.
time="2024-10-23T16:43:25-07:00" level=info msg="Current git describe" out=v3.30.0-0.dev-113-g6281d1b3224b
time="2024-10-23T16:43:25-07:00" level=fatal msg="Failed to parse version" error="Invalid Semantic Version" version=v3.30.0-
```

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/internal/version/version.go
+++ b/release/internal/version/version.go
@@ -114,7 +114,7 @@ func DetermineReleaseVersion(v Version, devTagSuffix string) (Version, error) {
 	if HasDevTag(v, devTagSuffix) {
 		// This is the first release from this branch - we can simply extract the version from
 		// the dev tag.
-		return New(strings.Split(gitVersion, devTagSuffix)[0]), nil
+		return New(strings.Split(gitVersion, "-"+devTagSuffix)[0]), nil
 	} else {
 		// This is a patch release - we need to parse the previous, and
 		// bump the patch version.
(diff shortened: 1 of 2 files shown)


danudey (approved):

release/internal/version/version.go
```diff
@@ -114,7 +114,7 @@ func DetermineReleaseVersion(v Version, devTagSuffix string) (Version, error) {
 	if HasDevTag(v, devTagSuffix) {
 		// This is the first release from this branch - we can simply extract the version from
 		// the dev tag.
-		return New(strings.Split(gitVersion, devTagSuffix)[0]), nil
+		return New(strings.Split(gitVersion, "-"+devTagSuffix)[0]), nil
```
caseydavenport, line 117: So, this PR broke this right here: https://github.com/projectcalico/calico/commit/5bcda09698e2ec73a8ec2ac06d8ca8cfb1a18a9f#diff-9c76a8ea05e8e2f8e1d03d5e04ac386dce3d9ba16e0f410ee965bdd63f5133eaL91-R114

We used to have logic that prepended the prefix if it didn't exist. I think we probably still want that? It makes this code more reliable (in case someone passes a prefix that already has a "-" on it. 

Basically instead of always including the "-", it should have logic like this:

```go
	if !strings.HasPrefix(devTagSuffix, "-") {
		// The dev tag marker should start with a hyphen.
		// For example in "v3.15.0-0.dev-1-g1234567", we want to split on the "-0.dev" part.
		devTagSuffix = "-" + devTagSuffix
	}
```

caseydavenport (approved):

PR #9388, opened 2024-10-24
[release tool] conflict simplification and enhancements

radTuti:
## Description

Minor changes to simplify release tool and avoid conflict with Enterprise

- define projectcalico & tigera organizations as constants
- PinnedVersion
  - add fn for creating hashrelease url to PinnedVersionData
  - update struct tag to have `omitempty` for fields that aren't in both pinned version and pinned components
- Version
  - define regex for branch as a variable to use in multiple places
- Calico manager
  - add `modifyHelmValues` function
  - clean up formatting in Calico's manager
- Operator manager
  - remove passing function argument to `Build`

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


release/pkg/manager/operator/manager.go
```diff
@@ -98,16 +102,16 @@ func NewManager(opts ...Option) *OperatorManager {
...
 }
 
-func (o *OperatorManager) Build(outputDir string) error {
+func (o *OperatorManager) Build() error {
 	if !o.isHashRelease {
 		return fmt.Errorf("operator manager builds only for hash releases")
 	}
 	if o.validate {
-		if err := o.PreBuildValidation(outputDir); err != nil {
+		if err := o.PreBuildValidation(o.tmpDir); err != nil {
```
caseydavenport, line 110: `o.PreBuildValidation` already has access to fields of `o`, no need to pass as an argument here.

release/pkg/manager/calico/manager.go
```diff
@@ -411,28 +412,35 @@ func (r *CalicoManager) TagRelease(ver string) error {
 	logrus.WithFields(logrus.Fields{"branch": branch, "version": ver}).Infof("Creating Calico release from branch")
 	_, err := r.git("tag", ver)
 	if err != nil {
-		return fmt.Errorf("Failed to tag release: %s", err)
+		return fmt.Errorf("failed to tag release: %s", err)
+	}
+	return nil
+}
+
+func (r *CalicoManager) modifyHelmValues() error {
```
caseydavenport, line 420: This change seems fine to me, I just don't quite understand why we're making it - is there a motivation behind this?

release/pkg/manager/calico/manager.go
```diff
@@ -514,14 +522,14 @@ func (r *CalicoManager) releasePrereqs() error {
 		return fmt.Errorf("cannot cut release from branch: %s", branch)
 	}
 
-	// Make sure we have a github token - needed for publishing to GH.
+	// Make sure we have the right token needed for publishing.
```
caseydavenport, line 525: I don't understand this comment change - seems fine, but the other text was also fine and more specific.

release/pkg/manager/operator/manager.go
```diff
@@ -272,6 +276,6 @@ func (o *OperatorManager) Clone() error {
 	return utils.Clone(fmt.Sprintf("git@github.com:%s/%s.git", o.githubOrg, o.repoName), o.branch, o.dir)
 }
 
-func (o *OperatorManager) make(target string, env []string) (string, error) {
-	return o.runner.Run("make", []string{"-C", o.dir, target}, env)
+func (o *OperatorManager) makeWithNoOutput(target string, env ...string) error {
```
caseydavenport, line 279: Why are we dropping the output from all of these commands? It just means we won't have debug logs available if something goes wrong no?

caseydavenport (changes requested):
Just some small comments, but otherwise seems fine

PR #9392, opened 2024-10-25
Fix untracked policy mark clearance.

fasaxc:
## Description

Was using SetMark(0), which ends up using a mask of 0, so it is a no-op.

## Related issues/PRs

CORE-10758

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
BPF: Fix that we'd fail to clear mark bits after applying do-not-track policy.  Use dedicated mark for XDP bypass traffic.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/rules/static.go
+++ b/felix/rules/static.go
@@ -1165,7 +1165,7 @@ func (r *DefaultRuleRenderer) StaticBPFModeRawChains(ipVersion uint8,
 	rawRules = append(rawRules,
 		generictables.Rule{
 			// Return, i.e. no-op, if bypass mark is not set.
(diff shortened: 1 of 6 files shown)


felix/rules/static.go
```diff
@@ -1271,7 +1271,7 @@ func (r *DefaultRuleRenderer) StaticBPFModeRawChains(ipVersion uint8,
 			// logic because no one else's iptables should have had a chance to execute
 			// yet.
 			{
-				Action: r.SetMark(0),
+				Action: r.SetMaskedMark(0, tcdefs.MarksMask),
```
tomastigera, line 1274: the `(` started on line 1265 in the comment above never closes

tomastigera (approved):

Issue #7558, opened 2023-04-17
Issue with BGP advertisement using /32 serviceExternalIPs


StephanR0:
When we advertise a /32 cidr address to our bgppeer, it gets advertised as blackhole.
If we do the same thing with a /28 or /29 cidr, it does get advertised correctly.

We have a loadbalancer service on a deployment with 3 replicas. We want this service to be available on a public address.

Because of the incorrect advertisement, http requests don't arrive on the correct nodes and the sites become unavailable.

## Expected Behavior

Advertise the /32 address via the correct peer addresses where the pods run on kubernetes nodes with addresses 192.168.1.65, 192.168.1.66 and 192.168.1.68.
We want this service to get an external IP through calico from the serviceExternalIPs pool.
I would expect the 180.xx.xx.87/32 address to be advertised the same way the 180.xx.xx.80/28 range gets advertised.

So in this example:
```
180.xx.xx.87/32    via 192.168.1.65 on eth0 [Mesh_10_168_1_65 2023-04-14] * (100/0) [i]
                   via 192.168.1.66 on eth0 [Mesh_10_168_1_66 2023-04-14] (100/0) [i]
                   via 192.168.1.68 on eth0 [Mesh_10_168_1_68 2023-04-14] (100/0) [i]
```

## Current Behavior

If we define the bgpconfigation with a /32 adress, like this:
```
$ kubectl get bgpconfigurations default -o yaml
apiVersion: projectcalico.org/v3
kind: BGPConfiguration
metadata:
  name: default
spec:
  asNumber: 65003
  logSeverityScreen: Info
  nodeToNodeMeshEnabled: true
  serviceExternalIPs:
  - cidr: 180.xx.xx.87/32
```

We see the routing advertisement 180.xx.xx.87/32 set as 'blackhole' (self?) and routable via all other nodes (see '>' marks)
The 180.xx.xx.87/32 gets routed to blackhole and ALL other nodes, even though these nodes don't all listen for the same services.

```
$ kubectl -n calico-system exec -it daemonset/calico-node -c calico-node -- birdcl -s /run/calico/bird.ctl show route
BIRD v0.3.3+birdv1.6.8 ready.
0.0.0.0/0          via 192.168.1.1 on eth0 [Global_10_168_1_1 2023-04-03] ! (100/0) [AS8455i]
                   via 192.168.1.1 on eth0 [kernel1 2023-04-03] (10)
[...]

PR #9422, opened 2024-10-30
Fix: announcement of /32 and /128 entries for serviceExternalIPs

tanujd11:
## Description

Similar to https://github.com/projectcalico/calico/pull/6282 but for externalIP, I kind of ran into this and it would be great if externalIPs populated as single IP addresses stop advertising itself as global routes and only expose itself for services with externalTrafficPolicy as Cluster.

## Related issues/PRs

Fixes https://github.com/projectcalico/calico/issues/7558

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that single-IP entries on BGPConfiguration serviceExternalIPs were not advertised according to external traffic policy. 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/confd/pkg/backends/calico/routes.go
+++ b/confd/pkg/backends/calico/routes.go
@@ -377,6 +377,32 @@ func (rg *routeGenerator) isSingleLoadBalancerIP(loadBalancerIP string) bool {
 	return false
 }
 
+// isSingleExternalIP determines if the given IP is in the list of
+// allowed ExternalIP CIDRs given in the default bgpconfiguration
+// and is a single IP entry (/32 for IPV4 or /128 for IPV6)
+func (rg *routeGenerator) isSingleExternalIP(externalIP string) bool {
+	if externalIP == "" {
+		log.Debug("Skip empty service External IP")
+		return false
+	}
+	ip := net.ParseIP(externalIP)
+	if ip == nil {
+		log.Errorf("Could not parse service External IP: %s", externalIP)
+		return false
+	}
+
+	for _, allowedNet := range rg.client.GetExternalIPs() {
+		if allowedNet.Contains(ip) {
+			if ones, bits := allowedNet.Mask.Size(); ones == bits {
+				return true
+			}
+		}
+	}
+
+	// Guilty until proven innocent
+	return false
+}
+
 // addFullIPLength returns a new slice, with the full IP length appended onto every item.
 func addFullIPLength(items []string) []string {
 	res := make([]string, 0)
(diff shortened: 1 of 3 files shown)


caseydavenport:
@tanujd11 thanks for the PR - before we can do anything you'll need to sign the CLA in order to make the bot happy :pray:

tanujd11:
Done @caseydavenport.

confd/pkg/backends/calico/routes.go
```diff
@@ -423,10 +449,24 @@ func (rg *routeGenerator) advertiseThisService(svc *v1.Service, ep *v1.Endpoints
...
+	// - LoadBalancer with a single IP.
+	// - Any one of externalIPs in service of type LoadBalancer or NodePort with a single IP.
+	if svc.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeCluster {
+		if svc.Spec.Type == v1.ServiceTypeLoadBalancer && rg.isSingleLoadBalancerIP(svc.Spec.LoadBalancerIP) {
+			logc.Debug("Advertising load balancer of type cluster because of single IP definition")
+			return true
+		}
+
+		if svc.Spec.Type == v1.ServiceTypeLoadBalancer || svc.Spec.Type == v1.ServiceTypeNodePort {
+			for _, extIP := range svc.Spec.ExternalIPs {
```
caseydavenport, line 463: If I understand correctly, this is going to advertise ALL ExternalIPs (even those which don't meet the criteria of matching a configured ExternalIP range exactly) if any of the ExternalIPs matches `isSingleExternalIP` - I think this means that any Service with one isSingleExternalIP will result in all of its ExternalIPs being advertised from every node, even if they are covered by an existing advertised route?

This is probably OK - I think it's a niche enough scenario that we shouldn't fuss too much (yet) about it, but might be worth confirming and perhaps commenting in the code that this is a known quirk of this codepath.

caseydavenport:
@tanujd11 thanks for the PR! This looks pretty good to me - I left one comment, and I also think we should add unit tests for this scenario much like the original PR for LBs did.

confd/pkg/backends/calico/routes.go
tanujd11: I have a confusion here. Won't [this](https://github.com/projectcalico/calico/blob/21f120f1e8bcbce5877cb481dc38572e23003384/confd/pkg/backends/calico/routes.go#L240-L249) codepath chose which externalIPs to actually advertise. The one you mentioned is for if the service is needed to be exposed. Even if this comes to be true, while advertising the routes the code I pasted will see if all the externalIP are allowed by BGPConfiguration. 

Please correct me if I am wrong though. Pretty new to the project.

caseydavenport: @tanujd11 you're absolutely correct - it's been a while since I've looked at this part of the code! Thanks for this :)

caseydavenport (approved):

marvin-tigera:

PR #9423, opened 2024-10-30
[BPF] always use bpf_redir_neigh after FIB lookup

tomastigera:
[BPF] use dst as nh when redirecting to workload
    
    When redirecting to a workload when NAt is required, usually from HEP
    because of node port, us the destination IP as the next hop in
    bpf_redir_neigh() to avoid FIB lookup.

    [BPF] if you know where to, skip explicit FIB lookup
    
    bpf_redirect_neigh() will do it.

    [BPF] skip neigh table lookup in fib lookup
    
    bpf_redir_neigh will do it

    [BPF] split fib.h into co-re and legacy
    
    Going forward, we will maintain co-re and will eventually deprecate and
    remove the legacy one. That will get us back to a single fib.h.

    [BPF] always use bpf_redir_neigh after FIB lookup
    
    There is not need to copy and fill in the mac addresses, bpf_redir_neigh
    will do that for us.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: alwyas use bpf_redirect_neigh when possible - reduce FIB lookups to minimum
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/conntrack_types.h
```diff
@@ -53,6 +53,8 @@ struct calico_ct_leg {
 
 	__u32 opener:1;
 
+	__u32 workload:1;
```
fasaxc, line 56: Can imagine forgetting that this means "the source of this leg is a local workload iface" at some point. Worth a comment just to make ti really clear?

fasaxc (changes requested):
Few small nits but great to see the code getting modernised!

There's still quite a lot of shared logic between the two fib impls (I diffed them locally to get a better feel for it) but happy to go with your judgement on the split.  Not like we can't roll it back if we find the duplication awkward.

fasaxc (approved):

PR #9429, opened 2024-11-01
Update to allowing building and publishing images in hashrelease

radTuti:
## Description

This allows building and publishing images in hashreleases.

Semaphore allows the ability to distinguish how a workflow is started using [`SEMAPHORE_WORKFLOW_TRIGGERED_BY_SCHEDULE`](https://docs.semaphoreci.com/reference/env-vars#workflow-triggered-by-schedule) environment variable.

This allow cleans up some of the hashrelease tasks and move into calico manager or build/main.go

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/Makefile
+++ b/release/Makefile
@@ -26,8 +26,8 @@ ci: static-checks
 ###############################################################################
 .PHONY: hashrelease
 hashrelease: bin/release var-require-all-GITHUB_TOKEN
-	@bin/release hashrelease build
-	@bin/release hashrelease publish
+	@bin/release hashrelease build ${BUILD_ARGS}
+	@bin/release hashrelease publish ${PUBLISH_ARGS}
 
 ###############################################################################
 # Release
--- a/release/build/main.go
+++ b/release/build/main.go
@@ -59,9 +59,10 @@ const (
 	newBranchFlag    = "new-branch-version"
 
 	// Configuration flags for the release publish command.
-	skipPublishImagesFlag    = "skip-publish-images"
-	skipPublishGitTag        = "skip-publish-git-tag"
-	skipPublishGithubRelease = "skip-publish-github-release"
+	skipPublishImagesFlag        = "skip-publish-images"
+	skipPublishGitTagFlag        = "skip-publish-git-tag"
+	skipPublishGithubReleaseFlag = "skip-publish-github-release"
+	skipPublishHashreleaseFlag   = "skip-publish-hashrelease-server"
 )
 
 var (
@@ -161,7 +162,7 @@ func hashreleaseSubCommands(cfg *config.Config) []*cli.Command {
 				&cli.BoolFlag{Name: skipValidationFlag, Usage: "Skip all pre-build validation", Value: false},
 				&cli.BoolFlag{Name: skipBranchCheckFlag, Usage: "Skip check that this is a valid release branch.", Value: false},
 				&cli.BoolFlag{Name: buildImagesFlag, Usage: "Build images from local codebase. If false, will use images from CI instead.", Value: false},
-				&cli.StringFlag{Name: imageRegistryFlag, Usage: "Specify image registry to use", Value: ""},
+				&cli.StringFlag{Name: imageRegistryFlag, Usage: "Specify image registry to use", EnvVars: []string{"REGISTRIES", "DEV_REGISTRIES"}, Value: ""},
 				&cli.StringFlag{Name: operatorOrgFlag, Usage: "Operator git organization", EnvVars: []string{"OPERATOR_GIT_ORGANIZATION"}, Value: config.OperatorDefaultOrg},
 				&cli.StringFlag{Name: operatorRepoFlag, Usage: "Operator git repository", EnvVars: []string{"OPERATOR_GIT_REPO"}, Value: config.OperatorDefaultRepo},
 				&cli.StringFlag{Name: operatorImageFlag, Usage: "Specify the operator image to use", EnvVars: []string{"OPERATOR_IMAGE"}, Value: config.OperatorDefaultImage},
@@ -285,6 +286,9 @@ func hashreleaseSubCommands(cfg *config.Config) []*cli.Command {
 			Name:  "publish",
 			Usage: "Publish hashrelease from _output/ to hashrelease server",
 			Flags: []cli.Flag{
+				&cli.StringFlag{Name: imageRegistryFlag, Usage: "Specify image registry to use", EnvVars: []string{"REGISTRIES", "DEV_REGISTRIES"}, Value: ""},
+				&cli.BoolFlag{Name: skipPublishImagesFlag, Usage: "Skip publishing of container images to registry", Value: true},
+				&cli.BoolFlag{Name: skipPublishHashreleaseFlag, Usage: "Skip publishing to hashrelease server", Value: false},
 				&cli.BoolFlag{Name: latestFlag, Usage: "Promote this release as the latest for this stream", Value: true},
 				&cli.BoolFlag{Name: skipValidationFlag, Usage: "Skip pre-build validation", Value: false},
 				&cli.BoolFlag{Name: skipImageScanFlag, Usage: "Skip sending images to image scan service.", Value: false},
@@ -298,17 +302,20 @@ func hashreleaseSubCommands(cfg *config.Config) []*cli.Command {
 					return fmt.Errorf("%s must be set if %s is set", skipImageScanFlag, skipValidationFlag)
 				}
 
-				// Extract the version from pinned-version.yaml.
-				hash, err := pinnedversion.RetrievePinnedVersionHash(cfg.TmpFolderPath())
+				// Extract the pinned version as a hashrelease.
+				hashrel, err := pinnedversion.AsHashrelease(cfg.RepoRootDir, cfg.TmpFolderPath(), dir)
 				if err != nil {
 					return err
 				}
+				if c.Bool(latestFlag) {
+					hashrel = hashrel.AsLatest()
+				}
 
 				// Check if the hashrelease has already been published.
-				if published, err := tasks.HashreleasePublished(cfg, hash); err != nil {
+				if published, err := tasks.HashreleasePublished(cfg, hashrel.Hash); err != nil {
 					return err
 				} else if published {
-					return fmt.Errorf("hashrelease %s has already been published", hash)
+					return fmt.Errorf("%s hashrelease (%s) has already been published", hashrel.Name, hashrel.Hash)
 				}
 
 				// Push the operator hashrelease first before validaion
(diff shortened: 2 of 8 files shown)


release/pkg/manager/calico/manager.go
```diff
@@ -477,6 +468,9 @@ func (r *CalicoManager) releasePrereqs() error {
 
 // Prerequisites specific to publishing a release.
 func (r *CalicoManager) publishPrereqs() error {
+	if r.isHashRelease {
+		return nil
```
caseydavenport, line 472: Do we not have any other hashrelease prereqs? e.g., if it already exists on the server?

I think we're checking that elsewhere, but why not here? Seems better to consolidate into one location. As it stands, it's not obvious when something belongs in the CalicoManager vs. when it just goes into main.go.

release/pkg/manager/calico/manager.go
```diff
@@ -708,7 +708,8 @@ func (r *CalicoManager) buildContainerImages(ver string) error {
 	return nil
 }
 
-func (r *CalicoManager) publishGithubRelease(ver string) error {
+func (r *CalicoManager) publishGithubRelease() error {
+	ver := r.calicoVersion
```
caseydavenport, line 712: We should just swap `ver` for `r.calicoVersion` instead of using a local variable.

release/pkg/manager/calico/manager.go
```diff
@@ -659,7 +653,11 @@ func (r *CalicoManager) buildReleaseTar(ver string, targetDir string) error {
 	return nil
 }
 
-func (r *CalicoManager) buildContainerImages(ver string) error {
+func (r *CalicoManager) buildContainerImages() error {
+	if !r.buildImages {
```
caseydavenport, line 657: We seem to be inconsistent in this file about whether we check the boolean within the function like this, or if we check it before calling the function.

Personally, I think that calling a function and having it secretly do nothing because of configuration makes the code harder to read, so I'd prefer a model like this:

```go
if r.buildImages {
  r.buildImages()
}
```

We can still have safeguards in the function:

```go
func buildImages() error {
  if !r.buildImages {
    return fmt.Errorf("BUG: Attempted to build images, but not configured to do so")
  }
}
```

caseydavenport (changes requested):

release/Makefile
```diff
@@ -26,8 +26,8 @@ ci: static-checks
 ###############################################################################
 .PHONY: hashrelease
 hashrelease: bin/release var-require-all-GITHUB_TOKEN
-	@bin/release hashrelease build
-	@bin/release hashrelease publish
+	@bin/release hashrelease build ${BUILD_ARGS}
```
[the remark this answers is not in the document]
radTuti: will switch to ENV.

I am using the make target `make hashrelease` so that the variables from `metdata.mk` file gets used

release/build/main.go
```diff
@@ -160,7 +162,7 @@ func hashreleaseSubCommands(cfg *config.Config) []*cli.Command {
 				&cli.BoolFlag{Name: skipValidationFlag, Usage: "Skip all pre-build validation", Value: false},
 				&cli.BoolFlag{Name: skipBranchCheckFlag, Usage: "Skip check that this is a valid release branch.", Value: false},
 				&cli.BoolFlag{Name: buildImagesFlag, Usage: "Build images from local codebase. If false, will use images from CI instead.", Value: false},
-				&cli.StringFlag{Name: imageRegistryFlag, Usage: "Specify image registry to use", Value: ""},
+				&cli.StringFlag{Name: imageRegistryFlag, Usage: "Specify image registry to use", EnvVars: []string{"REGISTRIES", "DEV_REGISTRIES"}, Value: ""},
```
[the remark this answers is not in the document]
radTuti: will switch to just `REGISTRIES`

release/pkg/manager/calico/manager.go
```diff
@@ -48,10 +49,6 @@ func NewManager(opts ...Option) *CalicoManager {
 	b := &CalicoManager{
 		runner:          &command.RealCommandRunner{},
 		validate:        true,
-		buildImages:     true,
```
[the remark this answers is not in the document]
radTuti: the options only allow setting true. 
With release we already pass the value as true

release/pkg/manager/calico/manager.go
radTuti: I can add the check here. However it is in main so that operator images do not get pushed if the hashrelease already exist

release/pkg/manager/calico/manager.go
radTuti: I put this in here to make it more consistent which is similar to what `publishContainerImages` does

release/pkg/manager/calico/options.go
```diff
@@ -63,11 +66,41 @@ func WithOutputDir(outputDir string) Option {
 	}
 }
 
-func WithPublishOptions(images, tag, github bool) Option {
+func WithPublishOptions(opt ...PublishOptions) Option {
```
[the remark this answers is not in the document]
caseydavenport: >  As well as it works better for close source.

Could you explain this a bit more? It's not obvious to me how this is better for closed source code.

release/pkg/manager/calico/manager.go
caseydavenport: OK, we can fix up any other inconsistencies in another PR then.

release/internal/pinnedversion/templates/calico-version.yaml.gotmpl
radTuti, line 1: need a way to note the custom registry

release/pkg/manager/calico/manager.go
```diff
@@ -462,22 +511,199 @@ func (r *CalicoManager) releasePrereqs() error {
...
+}
 
-	// TODO: Make sure the environment isn't dirty.
+// Check that the images exists with the correct version.
+func (r *CalicoManager) assertImageVersions() error {
+	for _, img := range images {
+		imageName := strings.TrimPrefix(img, "calico/")
+		switch img {
+		case "calico/apiserver":
+			for _, reg := range r.imageRegistries {
```
caseydavenport, line 613: I think the main reasons for the switch statement is because the command to check the version is different, right?

Could we factor out all of the common code here?

```go
for _, img := range images {
  switch img {
    case "calico/apiserver":
      cmd = []string{...}
    case "calico/cni": 
      cmd = []string{...}
  }
  for _, reg := r.imageRegistries {
    // Run command determined above. 
  }
}
```

release/pkg/manager/calico/manager.go
```diff
@@ -462,22 +511,199 @@ func (r *CalicoManager) releasePrereqs() error {
...
+				out, _ := r.runner.Run("docker", []string{"run", "--rm", fmt.Sprintf("%s/%s:%s", reg, imageName, r.calicoVersion)}, nil)
+				if !strings.Contains(out, r.calicoVersion) {
+					return fmt.Errorf("version does not match for image %s/%s:%s", reg, imageName, r.calicoVersion)
+				}
+			}
+		case "calico/cni": // cni-plugin
+			for _, reg := range r.imageRegistries {
+				for _, cmd := range []string{"calico", "calico-ipam"} {
+					out, err := r.runner.Run("docker", []string{"run", "--rm", fmt.Sprintf("%s/%s:%s", reg, imageName, r.calicoVersion), cmd, "-v"}, nil)
+					if err != nil || !strings.Contains(out, r.calicoVersion) {
```
caseydavenport, line 623: `err != nil` should be handled distinctly from not finding the version in the output - they are two different root causes. We def. need to log out the error if we hit it otherwise the code may fail in a confusing or non obvious way.

release/pkg/manager/calico/manager.go
```diff
@@ -462,22 +511,199 @@ func (r *CalicoManager) releasePrereqs() error {
...
+				}
+			}
+		case "calico/node":
+			for _, reg := range r.imageRegistries {
+				out, err := r.runner.Run("docker", []string{"run", "--rm", fmt.Sprintf("%s/%s:%s", reg, imageName, r.calicoVersion), "versions"}, nil)
+				if err != nil || !strings.Contains(out, r.calicoVersion) {
+					return fmt.Errorf("version does not match for image %s/%s:%s", reg, imageName, r.calicoVersion)
+				}
+			}
+		case "calico/node-driver-registrar": // pod2daemon
```
caseydavenport, line 665: // pod2daemon 

Can be removed?

release/pkg/manager/calico/manager.go
```diff
@@ -612,7 +838,6 @@ func (r *CalicoManager) buildReleaseTar(ver string, targetDir string) error {
 		fmt.Sprintf("%s/cni:%s", registry, ver):                          filepath.Join(imgDir, "calico-cni.tar"),
 		fmt.Sprintf("%s/kube-controllers:%s", registry, ver):             filepath.Join(imgDir, "calico-kube-controllers.tar"),
 		fmt.Sprintf("%s/pod2daemon-flexvol:%s", registry, ver):           filepath.Join(imgDir, "calico-pod2daemon.tar"),
-		fmt.Sprintf("%s/key-cert-provisioner:%s", registry, ver):         filepath.Join(imgDir, "calico-key-cert-provisioner.tar"),
```
caseydavenport, line 615: why this change?

caseydavenport (changes requested):
A few small comments / questions but overall this looks very good to me.

release/build/main.go
```diff
@@ -159,8 +161,8 @@ func hashreleaseSubCommands(cfg *config.Config) []*cli.Command {
 				&cli.StringFlag{Name: repoFlag, Usage: "Git repository", EnvVars: []string{"GIT_REPO"}, Value: config.DefaultRepo},
 				&cli.BoolFlag{Name: skipValidationFlag, Usage: "Skip all pre-build validation", Value: false},
 				&cli.BoolFlag{Name: skipBranchCheckFlag, Usage: "Skip check that this is a valid release branch.", Value: false},
-				&cli.BoolFlag{Name: buildImagesFlag, Usage: "Build images from local codebase. If false, will use images from CI instead.", Value: false},
-				&cli.StringFlag{Name: imageRegistryFlag, Usage: "Specify image registry to use", Value: ""},
+				&cli.BoolFlag{Name: buildImagesFlag, Usage: "Build images from local codebase. If false, will use images from CI instead.", EnvVars: []string{"BUILD_IMAGES"}, Value: false},
+				&cli.StringFlag{Name: imageRegistryFlag, Usage: "Specify image registry to use", EnvVars: []string{"REGISTRIES"}, Value: ""},
```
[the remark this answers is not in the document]
radTuti: we should support multiple, I'll update accordingly

release/build/main.go
```diff
@@ -288,30 +291,43 @@ func hashreleaseSubCommands(cfg *config.Config) []*cli.Command {
 			Name:  "publish",
 			Usage: "Publish hashrelease from _output/ to hashrelease server",
 			Flags: []cli.Flag{
+				&cli.StringFlag{Name: orgFlag, Usage: "Git organization", EnvVars: []string{"ORGANIZATION"}, Value: config.DefaultOrg},
+				&cli.StringFlag{Name: repoFlag, Usage: "Git repository", EnvVars: []string{"GIT_REPO"}, Value: config.DefaultRepo},
+				&cli.StringFlag{Name: imageRegistryFlag, Usage: "Specify image registry to use", EnvVars: []string{"REGISTRIES"}, Value: ""},
+				&cli.BoolFlag{Name: skipPublishImagesFlag, Usage: "Skip publishing of container images to registry", EnvVars: []string{"SKIP_PUBLISH_IMAGES"}, Value: true},
```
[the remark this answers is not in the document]
radTuti: `--skip-publish-images` is an existing flag.
the `--skip-publish-hashrelease-server` is new and following existing pattern.
I do not think we should change that now in this PR. Perharps a discussion and a different PR

release/build/main.go
```diff
@@ -357,16 +400,13 @@ func releaseSubCommands(cfg *config.Config) []*cli.Command {
 		{
 			Name:  "generate-release-notes",
 			Usage: "Generate release notes for the next release",
-			Flags: []cli.Flag{
```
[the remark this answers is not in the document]
radTuti: release notes should be generated using projectcalico/calico so there is no need to provide the option of specifying the org

release/pkg/manager/calico/manager.go
radTuti: key-cert-provisioner does not need to build a release tar. It was accidentally added here and Dan has corrected that for 3.29 branch

caseydavenport: Why is it not needed within the release tar?

release/pkg/manager/calico/manager.go
radTuti: Not really, it may be more cumbersome and confusing for a few reasons:

- apiserver: always return an error for the version check as there is no kube server
- each command is formatted based on the registry so i will have to return a cmdFornat instead of command.
- cni: we check the version of both calico and calico-ipam which will require returning an array of command
- key-cert-provisioner images (key-cert-provisioner & test-signer) do not have versions in them atm

Having it separated as-is is more readable IMO instead

caseydavenport (approved):

PR #9431, opened 2024-11-01
Add key-cert-provisioner to release process [v3.29]

danudey:
Add key-cert-provisioner to v3.29 release process. Cherry-pick of #9285 and #9463


radTuti (changes requested):
this won't work as key-cert-provisioner does not have the right targets for the `buildContainerImages` and `publishContainerImages`

radTuti (approved):

Issue #9394, opened 2024-10-25
add CALICO_DISABLE_TUNNEL env


gojoy:
## Expected Behavior
A environment variable is needed to control whether process **calico-node -allocate-tunnel-addrs** is started when
run the calico pod in k8s.

## Possible Solution

add the a judgment by **CALICO_DISABLE_TUNNTL** env in the [rc.local](https://github.com/projectcalico/calico/blob/master/node/filesystem/etc/rc.local#L44)
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context

In our scenario, the Kubernetes cluster fully uses the calico BGP mode and no longer starts process **calico-node -allocate-tunnel-addrs**  to reduce the pressure of Calico on the backend

## Your Environment

* Calico version v3.27.2,
* Calico dataplane (iptables, windows etc.) bgp
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes
* Operating System and version:linux
* Link to your project (optional):

fasaxc:
Are you running Typha?  The allocate-tunnel-ips program will use typha if enabled, which is our fan-out proxy.  Using that removes the load from the API server.

caseydavenport:
Yeah, agree that Typha is likely the solution here. If there's a reason that even with Typha this is causing load, that is most likely a bug in the `allocateip.go` code - I don't think we need / should have an env var to control whether this code is enabled assuming the code is written efficiently.

cyclinder:
Some users don't use Typha to install calico, we still need the env var until we figure out a potential bug in the `allocateip.go`, do you think so? @caseydavenport

caseydavenport:
@cyclinder if there truly is a bug in allocateip.go that is causing excessive resource usage, I'd rather we just fix that - that way, everyone benefits and not just those who discover this new arcane environment variable.

> In our scenario, the Kubernetes cluster fully uses the calico BGP mode and no longer starts process calico-node -allocate-tunnel-addrs to reduce the pressure of Calico on the backend

@gojoy do you have evidence that allocate-tunnel-addrs was causing load in your cluster? If so, could you provide that evidence so we can use it to diagnose what might be going wrong?

gojoy:
> @cyclinder if there truly is a bug in allocateip.go that is causing excessive resource usage, I'd rather we just fix that - that way, everyone benefits and not just those who discover this new arcane environment variable.
> 
[...]

PR #9443, opened 2024-11-05
add env var to disable tunnel in calico node

gojoy:
## Description
When the env var CALICO_DISABLE_TUNNEL is set, the process allocate-tunnel-addrs is not started.

In our environment, we only use BGP, didn't use tunnel, we found when create/delete/update a IPPool, all calico-node will get and update node many times, it will make k8s apiserver load very high.Even if the Typha component is deployed, these problems cannot be alleviated.
We have also noticed that there is already an environment variable `CALICO_DISABLE_FELIX` that has a similar function.
So add an another environment variable `CALICO_DISABLE_TUNNEL` that can control whether to start the process allocate-tunnel-addrs  or not.

## Related issues/PRs
Fixs #9394 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/filesystem/etc/rc.local
+++ b/node/filesystem/etc/rc.local
@@ -20,8 +20,11 @@ fi
 NODENAME=$(cat /var/lib/calico/nodename)
 export NODENAME
 
-# If possible pre-allocate any tunnel addresses.
-calico-node -allocate-tunnel-addrs -allocate-tunnel-addrs-run-once || exit 1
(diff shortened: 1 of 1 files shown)


caseydavenport:
I believe the root cause this is attempting to fix was resolved as part of this PR: https://github.com/projectcalico/calico/pull/9506

Issue #9437, opened 2024-11-04
Unable to patch network policies after Calico 3.29.0 upgrade


kashook:
After upgrading to Calico 3.29.0, we noticed that all `caliconetworkpolices` and `globalnetworkpolices` were renamed to be prefixed with `default.` and found the [documentation](https://docs.tigera.io/calico/latest/network-policy/policy-tiers/tiered-policy#the-default-tier) related to the new tier functionality that explains why.

Despite the policy resources being renamed, it does appear that it is generally possible to perform kubernetes API operations against them using the original, non-prefixed names. For example:

```
 $ kubectl get gnp testpol

NAME              CREATED AT
default.testpol   2024-11-04T20:44:22Z

```

```
 $ kubectl delete gnp testpol
globalnetworkpolicy.projectcalico.org "testpol" deleted
```

However, if you attempt to perform a patch using its original, non-prefixed name, you get an error. For example:

```
 $ kubectl patch gnp testpol -p '{"metadata":{"labels":{"testlabel":"value"}}}'
Error from server (BadRequest): the name of the object (default.testpol) does not match the name on the URL (testpol)
```

```
 $ kubectl label gnp testpol testlabel=value
Error from server (BadRequest): the name of the object (default.testpol) does not match the name on the URL (testpol)
```

This also affects any helm chart that contains networkpolicy/globalnetworkpolicy resources. An example from one of the helm releases we deploy in our cluster:

```
Error: UPGRADE FAILED: cannot patch "egress-athens-proxy-http" with kind GlobalNetworkPolicy: the name of the object (default.egress-athens-proxy-http) does not match the name on the URL (egress-athens-proxy-http) && cannot patch "athens-proxy" with kind NetworkPolicy: the name of the object (default.athens-proxy) does not match the name on the URL (athens-proxy)
```

Because of this, all our helm charts, or anything else that attempts to patch polices are broken until they are updated to include the `default.` prefix on the name.

Is this intended behavior? I didn't see a breaking change mentioned in the [release notes](https://docs.tigera.io/calico/3.29/release-notes/). The fact that you can perform other operations (such as get and delete) without the prefix makes me hope that this is maybe just a bug and that we won't have to modify all of our helm charts to add the prefix on all the network policies.

caseydavenport:
@kashook thanks for raising this - will take a look. I don't think this is intended behavior.

ffilippopoulos:
[...]

PR #9445, opened 2024-11-05
[Prototype] Fix for PATCH without tier prefix

caseydavenport:
## Description

This is one attempt at fixing https://github.com/projectcalico/calico/issues/9437

Basically, we try the update as provided. If it fails, then we retry it removing the "tier" from the object. This allows us to call applyPatch on an object without the tier prefix, which passes the name validation.

Once it passes the applyPatch name validation, it continues on to PrepareForUpdate, at which point we need to re-canonicalize the objects to include the tier prefix on them for storage...

I am not sure this is a viable solution - we need to think about what happens if there actually are policies in different tiers and someone tries to send a PATCH. I think it might just be OK, because in the second case you would need to specify the tier name (it's only for default that this is really a problem?)

At a minimum, for this change to go in we'd want to also:

- Only apply this logic for policies in the default tier
- Only apply this logic if we get the precise error back that we expect in the mismatched tier case.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/apiserver/pkg/registry/projectcalico/networkpolicy/strategy.go
+++ b/apiserver/pkg/registry/projectcalico/networkpolicy/strategy.go
@@ -47,7 +47,10 @@ func (policyStrategy) PrepareForCreate(ctx context.Context, obj runtime.Object) 
 }
 
 func (policyStrategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
-	obj.(*calico.NetworkPolicy).Name = canonicalizePolicyName(old)
+	// Canonicalize the old object's name as well to make sure it's consistent.
+	// This typically shouldn't be needed for update requests, but is necessary in PATCH
+	// requests because we may need to strip the tier earlier in the pipeline to pass earlier validation.
+	old.(*calico.NetworkPolicy).Name = canonicalizePolicyName(old)
 }
 
 func canonicalizePolicyName(obj runtime.Object) string {
--- a/apiserver/pkg/storage/calico/resource.go
+++ b/apiserver/pkg/storage/calico/resource.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"reflect"
 	"strconv"
+	"strings"
 	"time"
 
 	aapierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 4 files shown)


apiserver/pkg/registry/projectcalico/networkpolicy/strategy.go
```diff
@@ -47,7 +47,10 @@ func (policyStrategy) PrepareForCreate(ctx context.Context, obj runtime.Object)
 }
 
 func (policyStrategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
-	obj.(*calico.NetworkPolicy).Name = canonicalizePolicyName(old)
+	// Canonicalize the old object's name as well to make sure it's consistent.
+	// This typically shouldn't be needed for update requests, but is necessary in PATCH
+	// requests because we may need to strip the tier earlier in the pipeline to pass earlier validation.
+	old.(*calico.NetworkPolicy).Name = canonicalizePolicyName(old)
 }
```
caseydavenport, line 54: Assuming this approach even has legs, we'll need to make this change for global policies as well.

apiserver/pkg/storage/calico/resource.go
```diff
@@ -417,11 +419,27 @@ func (rs *resourceStore) GuaranteedUpdate(
 			klog.Errorf("checking preconditions (%s)", err)
 			return err
 		}
+
 		// update the object by applying the userUpdate func & encode it
 		updatedObj, ttl, err := userUpdate(curState.obj, *curState.meta)
 		if err != nil {
```
caseydavenport, line 425: Probably want to check the exact error message here.

apiserver/pkg/storage/calico/resource.go
```diff
@@ -417,11 +419,27 @@ func (rs *resourceStore) GuaranteedUpdate(
...
-			klog.Errorf("applying user update: (%s)", err)
-			return err
+			// Try updating the object name to remove the tier - it's possible that the user send the
+			// object without the tier in the name, in which case we need to retry the update against the
+			// object without the tier in the name.
+			switch curState.obj.(type) {
+			case *v3.GlobalNetworkPolicy:
```
caseydavenport, line 430: Should probably only execute this logic if the tier is "default"

caseydavenport:
Closing this as @MichalFupso is working on a better fix.

Issue #9444, opened 2024-11-05
calico-typha CrashLoopBackOff version 3.29: cannot create resource "tiers" in API group "crd.projectcalico.org" at the cluster scope


jcurtis789:
Hello,

Installing Calico v3.29 using the "Manifest" approach outlined on this document results in an error with the calico-typha Deployment.
https://docs.tigera.io/calico/latest/getting-started/kubernetes/self-managed-onprem/onpremises#install-calico-with-kubernetes-api-datastore-more-than-50-nodes

I'm specifically using the section for "Install Calico with Kubernetes API datastore, more than 50 nodes".

This is the raw YAML that I am applying:
https://raw.githubusercontent.com/projectcalico/calico/v3.29.0/manifests/calico-typha.yaml

## Expected Behavior
I expect the calico-typha Pods to become healthy.

## Current Behavior
The calico-typha Pods remain in a CrashLoopBackOff state. The output of 'kubectl logs' mentions the following failure:

`
2024-11-05 18:16:15.865 [INFO][1] typha/daemon.go 312: Initializing the datastore (if needed).

2024-11-05 18:16:15.868 [INFO][1] typha/client.go 253: Unable to initialize default Tier error=connection is unauthorized: tiers.crd.projectcalico.org is forbidden: User "system:serviceaccount:kube-system:calico-node" cannot create resource "tiers" in API group "crd.projectcalico.org" at the cluster scope

2024-11-05 18:16:15.868 [INFO][1] typha/client.go 258: Unable to initialize adminnetworkpolicy Tier error=connection is unauthorized: tiers.crd.projectcalico.org is forbidden: User "system:serviceaccount:kube-system:calico-node" cannot create resource "tiers" in API group "crd.projectcalico.org" at the cluster scope

2024-11-05 18:16:15.868 [ERROR][1] typha/daemon.go 318: Failed to initialize datastore error=connection is unauthorized: tiers.crd.projectcalico.org is forbidden: User "system:serviceaccount:kube-system:calico-node" cannot create resource "tiers" in API group "crd.projectcalico.org" at the cluster scope
`

## Possible Solution
In the YAML located in the documentation, the calico-typha Deployment uses the serviceAccountName 'calico-node' and, by extension, the 'calico-node' ClusterRole. I was able to resolve this issue rather quickly by adding the 'create' verb to the list of resources (including tiers) under the 'crd.projectcalico.org' apiGroup.

[...]

Issue #9442, opened 2024-11-05
Missing RBAC permission in charts/calico to create default Tiers causing typha to not start


sfudeus:
3.29.0 introduced the new policy tiers. According to the docs, the default tier and the AdminNetworkPolicy tier are supposed to be auto-created. calico-kube-controllers even has the permissions to do so and seemingly has created these. 

I'm aware that using https://github.com/projectcalico/calico/tree/master/charts/calico is not a supported means for installing and maintaining Calico, but it causes incomplete manifests even just for documentation. 

## Expected Behavior
The new tiers are auto-created by a component which has the permissions to do so, the others just consume them. 

## Current Behavior

Seemingly calico-kube-controllers does create the new tier objects. But typha as well tries to, and fails with permission issues. 
This seems to be caused by the fact the existence is checked by creation and reaction to an `ErrorResourceAlreadyExists` error (see client.go#408 and 429). 
This causes typha to fail as long as it hasn't create permissions. 

```
2024-11-05 10:07:45.616 [ERROR][1] typha/daemon.go 318: Failed to initialize datastore error=connection is unauthorized: tiers.crd.projectcalico.org is forbidden: User "system:serviceaccount:kube-system:calico-node" cannot create resource "tiers" in API group "crd.projectcalico.org" at the cluster scope
2024-11-05 10:07:46.623 [INFO][1] typha/client.go 253: Unable to initialize default Tier error=connection is unauthorized: tiers.crd.projectcalico.org is forbidden: User "system:serviceaccount:kube-system:calico-node" cannot create resource "tiers" in API group "crd.projectcalico.org" at the cluster scope
2024-11-05 10:07:46.625 [INFO][1] typha/client.go 258: Unable to initialize adminnetworkpolicy Tier error=connection is unauthorized: tiers.crd.projectcalico.org is forbidden: User "system:serviceaccount:kube-system:calico-node" cannot create resource "tiers" in API group "crd.projectcalico.org" at the cluster scope
```

In charts/calico, Typha is set up running with `serviceAccountName: calico-node`. calico-node is only set up with read,list,watch permissions for tiers. (https://github.com/projectcalico/calico/blob/master/charts/calico/templates/calico-node-rbac.yaml#L113). 
Likely this makes sense, to not allow calico-node to manipulate/create tiers in normal operations, but if they do have logic to recreate them on demand/on migration, they need the permissions to do so - or the logic to migrate must be adapted to cope with not having permission. 

## Possible Solution

Either
[...]

PR #9446, opened 2024-11-06
Improve robustness of data store initialisation when creating tiers

mazdakn:
## Description

Make sure initialising data store is not stopped because of the failure in creating `default` and `adminnetworkpolicy` tiers. Calico node and typha do not have `create` permission, and as such attempts to create those tiers result in `unauthorised` errors which in the end result in:
- Typha not starting up, like this GH issues: 
 https://github.com/projectcalico/calico/issues/9442
 https://github.com/projectcalico/calico/issues/9444
- Node not to create `felixConfig` which results in other deployment errors.

This PR fixes the issue in two ways:
- Detect and log `unathorised` error during attempts to create those tiers. Then, ignore the error as other components like calico-kube-controller would eventually create them.
- Add `create` permission to calico node and typha permission to calico manifests. This is similar to operator approach like https://github.com/tigera/operator/blob/1f9a7a9afbadd33d2c216e033360ccbd540a4421/pkg/render/node.go#L482.

The PR also enables deploying typha as part of kind cluster. Having it enabled, would have allowed detecting the issue earlier.

## Related issues/PRs
GH issue: https://github.com/projectcalico/calico/issues/9442 https://github.com/projectcalico/calico/issues/9444

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Do not fail data store initialisation when unauthorised error happen while creating default and adminnetworkpolicy tiers. Those tiers eventually get created by another component. 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/clientv3/client.go
+++ b/libcalico-go/lib/clientv3/client.go
@@ -247,12 +247,14 @@ func (c client) EnsureInitialized(ctx context.Context, calicoVersion, clusterTyp
 		errs = append(errs, err)
 	}
 
-	if err := c.ensureDefaultTierExists(ctx); err != nil {
+	err := c.ensureTierExists(ctx, names.DefaultTierName, v3.Deny, v3.DefaultTierOrder)
+	if err != nil {
 		log.WithError(err).Info("Unable to initialize default Tier")
 		errs = append(errs, err)
 	}
 
-	if err := c.ensureAdminNetworkPolicyTierExists(ctx); err != nil {
+	err = c.ensureTierExists(ctx, names.AdminNetworkPolicyTierName, v3.Pass, v3.AdminNetworkPolicyTierOrder)
+	if err != nil {
 		log.WithError(err).Info("Unable to initialize adminnetworkpolicy Tier")
 		errs = append(errs, err)
 	}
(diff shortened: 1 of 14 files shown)


coutinhop (approved):

libcalico-go/lib/clientv3/client.go
```diff
@@ -393,43 +395,29 @@ func (c client) ensureClusterInformation(ctx context.Context, calicoVersion, clu
...
 // tier resource already exists.
-func (c client) ensureAdminNetworkPolicyTierExists(ctx context.Context) error {
-	order := v3.AdminNetworkPolicyTierOrder
-	actionPass := v3.Pass
-	anpTier := v3.NewTier()
-	anpTier.ObjectMeta = metav1.ObjectMeta{Name: names.AdminNetworkPolicyTierName}
-	anpTier.Spec = v3.TierSpec{
+func (c client) ensureTierExists(ctx context.Context, name string, defaultAction v3.Action, order float64) error {
+	tier, err := c.Tiers().Get(ctx, name, options.GetOptions{})
+	if err == nil && tier != nil {
```
caseydavenport, line 404: What if err != nil? We don't seem to handle that case - I think we need to check specifically for `IsNotFound` errors, and then log out any other types of errors.

caseydavenport:
@mazdakn I'm not sure I understand how this fixes the issue - Typha and calico/node are still going to try to create this on their first go around, because the Tiers won't exist yet - right?

kube-controllers won't start until after Typha / node have started, so the tiers won't exist yet. 

What if we just modified the existing logic to check for NotAuthorized errors as well?

mazdakn:
@caseydavenport in calico manifests, Typha has permission to `list`, `get` and `watch` tiers. This is similar to permission of Calico nodes.
By checking if the tier exists, using `get`, Typha will realise those tiers are available, and initialisation won't fail.

caseydavenport:
> By checking if the tier exists, using get, Typha will realise those tiers are available, and initialisation won't fail. I tested it in a cluster with Typha, and it fixed the issue.

@mazdakn I guess I don't understand why Typha will see that the tiers exist? On a fresh install, those tiers won't exist - right?

libcalico-go/lib/clientv3/client.go
mazdakn: Removed it, since it feels duplicate since we check tier existence few lines after.

node/Makefile
```diff
@@ -345,6 +345,10 @@ test_image: .calico_test.created
 calico-node.tar: $(NODE_CONTAINER_CREATED)
 	docker save --output $@ $(NODE_IMAGE):latest-$(ARCH)
 
+calico-typha.tar: ../go.mod $(shell find ../apiserver -name '*.go') $(shell find ../libcalico-go -name '*.go')
```
caseydavenport, line 348: should be `typha` instead of `apiserver` here

caseydavenport (approved):
Think we should update the RBAC to include the specific tier names, but otherwise LGTM

e.g., 

```
resourceNames:
- default
- adminnetworkpolicy
```

node/Makefile
mazdakn: ah, right. fixed it.

mazdakn:
I'll try to limit the create permission to only default and adminnetworkpolicy tiers in another PR.

PR #9451, opened 2024-11-06
[BPF] exclude Calico's special devices from dataIfaceRegex

tomastigera:
It is confusing when users edit the regex what extra needs to be included. Now the user can focus on the true interfaces and forget about the once that calico uses for its own functionality.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: calico special interfaces excluded from dataIfaceRegex
```

## Reminder for the reviewer

Other optional labels:


fasaxc (commented):
I don't think we should have an option for this one; but

- Some of the interface names are configurable, I think (that was likely a mistake!).  So we should derive the name from the config.
- The tunl0 one only belongs to us if IPIP is enabled.  So, perhaps that should only be included in our automatic list if IPIP is enabled (and otherwise, you need to include it in your regex).

I can't think of any reason that you'd want to exclude one of our interfaces form getting a program; it just seems like that'd break things!?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -529,6 +528,32 @@ func newBPFEndpointManager(
...
+		specialInterfaces = append(specialInterfaces, config.RulesConfig.WireguardInterfaceName)
+	}
+	if config.RulesConfig.WireguardEnabledV6 {
+		specialInterfaces = append(specialInterfaces, config.RulesConfig.WireguardInterfaceNameV6)
+	}
+
+	exp := "(" + config.BPFDataIfacePattern.String() + ")|("
+	for _, d := range specialInterfaces {
+		exp += "^" + d + "$|"
+	}
```
fasaxc, line 551:
```suggestion
	for i, d := range specialInterfaces {
		specialInterfaces[i] = "^" + regexp.Quote(d) + "$"
	}
	exp := config.BPFDataIfacePattern.String() + "|" + strings.Join(specialInterfaces, "|")
```
Recommend quoting the names (since some have dots in? `strings.Join` is a nice way to deal with the delimiter.

felix/docs/config-params.md
```diff
@@ -1366,10 +1366,10 @@ A regular expression that controls which interfaces Felix should attach BPF prog
 | --- | --- |
```
fasaxc, line 1366: Might want to tweak the description in felixconfig.go to make it clear that tunnels should no longer be included :tada:

fasaxc (changes requested):

fasaxc (approved):

PR #9452, opened 2024-11-07
[BPF] support for policy rules Log action

tomastigera:
It is sometimes nice to see whether packets are dropped and which packets are dropped (or accepted) and where. Whenever a packet matches ANY rule with a Log action, the verdict with packet details will be logged to the trace pipe regardless of the bpfLogLevel setting or bpfLogFilters filtering.

```
cali8d1e69e5f89-E: policy ALLOWED proto 17 src 10.65.0.3:46519 dest 172.18.0.6:8055
cali8d1e69e5f89-E: policy ALLOWED proto 6 src 10.65.0.3:36185 dest 10.65.0.2:8055
cali866cd63afec-I: policy ALLOWED proto 6 src 10.65.0.3:36185 dest 10.65.0.2:8055
cali866cd63afec-E: policy ALLOWED proto 6 src 10.65.0.2:43553 dest 10.65.0.3:8055
cali8d1e69e5f89-I: policy DENIED  proto 6 src 10.65.0.2:43553 dest 10.65.0.3:8055
cali8d1e69e5f89-E: policy ALLOWED proto 6 src 10.65.0.3:46519 dest 172.18.0.6:8055
```
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: support for log action in policy rules
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/skb.h
+++ b/felix/bpf-gpl/skb.h
@@ -166,4 +166,44 @@ static CALI_BPF_INLINE void skb_set_mark(struct __sk_buff *skb, __u32 mark)
 
 #define skb_mark_equals(skb, mask, val) (((skb)->mark & (mask)) == (val))
 
+static CALI_BPF_INLINE void skb_log(struct cali_tc_ctx *ctx, bool accepted)
+{
+	if (ctx->state->flags & CALI_ST_LOG_PACKET) {
+#ifdef BPF_CORE_SUPPORTED
+		if (bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_trace_vprintk)) {
+#if CALI_F_XDP
+#define DIR_STR "X"
+#elif ((CALI_COMPILE_FLAGS) & CALI_TC_INGRESS)
+#define DIR_STR "I"
+#else
+#define DIR_STR "E"
+#endif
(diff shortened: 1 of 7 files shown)


felix/bpf-gpl/skb.h
```diff
@@ -166,4 +166,44 @@ static CALI_BPF_INLINE void skb_set_mark(struct __sk_buff *skb, __u32 mark)
...
+				ctx->state->sport,
+				(__u64) &ctx->state->ip_dst,
+				ctx->state->dport,
+			};
+			bpf_trace_vprintk(fmt, sizeof(fmt), args, sizeof(args));
+			return;
+		}
+#endif
+		if (accepted) {
+			bpf_log("ACCEPTED");
```
fasaxc, line 202: Nit: we use Allow and Deny in policy so we should stick to the same language.

fasaxc (commented):
I _think_ this is a good idea but we should put warnings in the docs about the performance hit (which ISTR can be significant) and the fact that you'll get a warning from the kernel that the debug buffer is being used.

log-and-xxx actions are an enterprise-only feature I think.

fasaxc (approved):

PR #9458, opened 2024-11-08
nftables flowtable support

caseydavenport:
## Description

Adds nftables flowtable support to Calico. flowtables let established connections bypass large parts of the Linux networking stack, which should improve performance, especially for longer-lived connections.

It's controlled by a new `nftablesFlowTableOffload` FelixConfiguration field and is enabled by default. It only takes effect on the nftables data plane, so iptables-mode clusters are unaffected. When active, Calico programs a flowtable over the overlay/tunnel and local workload interfaces and offloads established flows from the forward path.

By default this covers pod and overlay traffic. To also offload traffic forwarded across a node's external interfaces (e.g. host NICs carrying pod traffic), set `nftablesFlowTableDataIfacePattern` to a regex matching those interfaces. It works the same way as the BPF data interface pattern, and an empty value (the default) leaves external interfaces out.

A few things worth discussing in review:

- Now that it's on by default, how does it perform in clusters with lots of short-lived connections? Does the flowtable overhead actually hurt there?
- The on/off switch and the interface pattern live in FelixConfiguration today. Do we want finer-grained control (e.g. per-endpoint or via policy) later?
- Enabling by default touches the FelixConfiguration CRD default, which the operator also ships. This likely needs a corresponding operator change so operator-managed installs pick up the new default.

## Related issues/PRs

Kube-proxy implementation: https://github.com/kubernetes/kubernetes/pull/128392

## Release Note

```release-note
nftables flowtable offload, which improves performance for established connections (including traffic forwarded over a node's external interfaces), is now available when using the nftables data plane. Disabled by default, see FelixConfiguration.
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -678,6 +690,20 @@ type FelixConfigurationSpec struct {
 	// +kubebuilder:default=Auto
 	NFTablesMode *NFTablesMode `json:"nftablesMode,omitempty"`
 
+	// NFTablesFlowTableOffload controls whether nftables flowtable offload is enabled for
+	// improved forwarding performance. When enabled, established connections accepted by
+	// Calico policy are offloaded to the kernel's flowtable fast path. Only applies when
+	// nftables mode is active. [Default: Disabled]
+	// +kubebuilder:default=Disabled
+	NFTablesFlowTableOffload *NFTablesFlowTableOffload `json:"nftablesFlowTableOffload,omitempty"`
```
mazdakn, line 698: Is there any reason to not enable it by default?

felix/dataplane/linux/dataplanedefs/dataplane_defs.go
```diff
@@ -25,4 +25,9 @@ const (
 
 	BPFInDev  = "bpfin.cali"
 	BPFOutDev = "bpfout.cali"
+
+	// FlowtableName is the name of the nftables flowtable used for offloading established
+	// flows. The rule renderer references it via "flow offload @<name>" and the nftables
+	// table programs the flowtable object under the same name; the two must stay in lockstep.
+	FlowtableName = "calico"
```
mazdakn, line 32: why this can't be declared in the nftables package itself?

felix/fv/flowtable_test.go
```diff
@@ -0,0 +1,143 @@
...
+		if CurrentSpecReport().Failed() {
+			logNFTDiags(tc.Felixes[0])
+		}
+		for i := range w {
+			if w[i] != nil {
+				w[i].Stop()
+			}
+		}
+		tc.Stop()
+		infra.Stop()
```
mazdakn: I believe these steps are now done by infra itself.

felix/fv/flowtable_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	BeforeEach(func() {
+		if !NFTMode() {
+			Skip("Flowtable offload is an nftables-only feature.")
+		}
+
+		infra = getInfra()
+		opts := infrastructure.DefaultTopologyOptions()
+		// Enable flowtable offload at Felix startup rather than via a mid-test config
+		// update, which would trigger a reprogram that races the connectivity checks.
```
mazdakn: nit: I think this comment is unnecessary.

felix/rules/endpoints.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2025 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2026 Tigera, Inc. All rights reserved.
```
mazdakn: should revert this since there is no change here.

felix/rules/endpoints_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2025 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2026 Tigera, Inc. All rights reserved.
```
mazdakn: same.

felix/generictables/actions.go
```diff
@@ -36,6 +36,7 @@ type ActionFactory interface {
 	LimitPacketRate(rate int64, burst int64, mark uint32) Action
 	LimitNumConnections(num int64, rejectWith RejectWith) Action
 	DSCP(value uint8) Action
+	FlowOffload(table string) Action
```
mazdakn: nit: if we only support `calico` table, maybe we should just hard code it.

felix/iptables/actions.go
```diff
@@ -129,6 +129,10 @@ func (a *actionFactory) DSCP(value uint8) generictables.Action {
 	}
 }
 
+func (s *actionFactory) FlowOffload(ft string) generictables.Action {
+	return nil
+}
```
mazdakn, line 137: please add update unit tests.

felix/nftables/actions.go
```diff
@@ -525,3 +529,16 @@ func (a DSCPAction) ToFragment(features *environment.Features) string {
...
+	TypeFlowOffload struct{}
+}
+
+func (c FlowOffloadAction) ToFragment(features *environment.Features) string {
+	return fmt.Sprintf("flow offload @%s", c.FlowTable)
+}
+
+func (c FlowOffloadAction) String() string {
+	return fmt.Sprintf("FlowOffload:%s", c.FlowTable)
+}
```
mazdakn, line 544: please add update unit tests.

felix/rules/static.go
```diff
@@ -609,6 +609,20 @@ func (r *DefaultRuleRenderer) failsafeOutChain(table string, ipVersion uint8) *g
 func (r *DefaultRuleRenderer) StaticFilterForwardChains() []*generictables.Chain {
 	rules := []generictables.Rule{}
 
+	if r.nft && r.NFTablesFlowTableOffload == "Enabled" {
```
mazdakn: Instead of using `Enabled` we should use the defined const in the API, i.e. `NFTablesFlowTableOffloadEnabled`
If we don't want to import API defs here, we should just change the type to bool.

felix/dataplane/linux/dataplanedefs/dataplane_defs.go
mazdakn: I see. This is used in the `rules` package. Left a comment there that if we should just hard code the table name. This is just a nit, and have no preference.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1150,8 +1151,29 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 	}
 
 	var filterMaps nftables.MapsDataplane
+	var ifceHandlerV4 nftables.InterfaceHandler
+	var flowtableHandlers []nftables.InterfaceHandler
 	if nftablesEnabled {
 		filterMaps = filterTableV4.(nftables.MapsDataplane)
+
+		if config.RulesConfig.NFTablesFlowTableOffload == "Enabled" {
```
mazdakn: Better to use the declared API constant .

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1399,8 +1422,23 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
+
+				var overlayDevicesV6 []string
+				if config.RulesConfig.VXLANEnabledV6 {
+					overlayDevicesV6 = append(overlayDevicesV6, dataplanedefs.VXLANIfaceNameV6)
+				}
+				if config.RulesConfig.WireguardEnabledV6 && len(config.RulesConfig.WireguardInterfaceNameV6) > 0 {
+					overlayDevicesV6 = append(overlayDevicesV6, config.RulesConfig.WireguardInterfaceNameV6)
+				}
+				nftablesV6RootTable.SetOverlayDevices(overlayDevicesV6)
+				flowtableHandlers = append(flowtableHandlers, nftablesV6RootTable)
```
mazdakn: should we do this only if `len(overlatDevicesV6) > 0`?

felix/nftables/table.go
```diff
@@ -482,6 +524,58 @@ func (n *NftablesTable) IPVersion() uint8 {
 	return n.ipVersion
 }
 
+// SetOverlayDevices sets the overlay/tunnel device names that should be included in the
+// flowtable device list. Called during initialization from int_dataplane.go.
+func (t *NftablesTable) SetOverlayDevices(devices []string) {
+	t.overlayDevices = devices
```
mazdakn: probably it's safer to copy the device list.

felix/nftables/table.go
```diff
@@ -48,6 +50,16 @@ const (
 	objectTypeMap   = "map"
 )
 
+type InterfaceHandler interface {
```
mazdakn: nit: should we call this `flowTableHandler` instead? `InterfaceHandler` sounds more generic. I was a bit confused.

mazdakn:
Overall looks good to me. Left some comments/nits.

mazdakn:
Regarding the questions in the description:

> How does it perform in clusters with lots of short-lived connections? Does the flowtable overhead actually hurt there?
IMO, we should execute a performance test, and verify the impact of shot/long lived connections. But anyway this is a useful feature.

The on/off switch and the interface pattern live in FelixConfiguration today. Do we want finer-grained control (e.g. per-endpoint or via policy) later?
I personally think we should try to NOT over complicate our config more than this. But to open future addition, maybe we should change the config model from `Enabled/Disabled` to something that allows additions in future. Maybe should use `All/Disabled` instead. In future, we can introduce more consts to more granular control.

felix/dataplane/linux/int_dataplane.go
caseydavenport: This list feeds the external-interface flowtable manager below (only registered when the data iface pattern is set), which is independent of the overlay devices. A non-overlay cluster can still want external-interface offload, and workload interfaces come in via the endpoint manager's handler either way. So I don't think we want to gate registration on overlay device count here.

felix/nftables/table.go
```diff
@@ -482,6 +524,58 @@ func (n *NftablesTable) IPVersion() uint8 {
...
+}
+
+// recalcFlowtableDevices combines overlay and workload interfaces into a sorted device list.
+// If the list changed, it marks the flowtable as dirty for the next Apply().
+func (t *NftablesTable) recalcFlowtableDevices() {
+	combined := make([]string, 0, len(t.overlayDevices)+len(t.workloadInterfaces)+len(t.externalDevices))
+	combined = append(combined, t.overlayDevices...)
+	combined = append(combined, t.workloadInterfaces...)
+	combined = append(combined, t.externalDevices...)
+	sort.Strings(combined)
```
mazdakn, line 580: Worth using `slices.Compact` to remove duplicate element (if happens for whatever reasons)

felix/iptables/actions.go
```diff
@@ -129,6 +129,12 @@ func (a *actionFactory) DSCP(value uint8) generictables.Action {
 	}
 }
 
+// FlowOffload is a no-op in iptables mode; flowtable offload is nftables-only. The rule
+// renderer only emits it when nftables is active, so this is never reached in practice.
+func (s *actionFactory) FlowOffload() generictables.Action {
+	return nil
```
mazdakn, line 136: Think we should add a panic here. It's basically a programming issue if it's called in iptables. WDYT?

felix/nftables/actions.go
```diff
@@ -525,3 +530,15 @@ func (a DSCPAction) ToFragment(features *environment.Features) string {
 func (a DSCPAction) String() string {
 	return fmt.Sprintf("DSCP %d", a.Value)
 }
+
+type FlowOffloadAction struct {
+	TypeFlowOffload struct{}
```
mazdakn, line 535: seems unused?

mazdakn (commented):
Left a few more comments.
Also do we need to add a cleanup logic if the config is changed or nftables is switched to something else?

caseydavenport:
@mazdakn thanks! And good catch on the cleanup logic. 

I think it is OK at the moment - the rules should be reconciled away so traffic won't actually hit the flowtable - but nothing deletes the table itself IIUC

felix/nftables/table.go
```diff
@@ -482,6 +524,58 @@ func (n *NftablesTable) IPVersion() uint8 {
...
+// flowtable device list. Called during initialization from int_dataplane.go.
+func (t *NftablesTable) SetOverlayDevices(devices []string) {
+	t.overlayDevices = slices.Clone(devices)
+	t.enableFlowtable()
+}
+
+// SetWorkloadInterfaces updates the set of active workload interfaces for the flowtable.
+// Called by the endpoint manager when workload endpoints change.
+func (t *NftablesTable) SetWorkloadInterfaces(ifces []string) {
+	t.workloadInterfaces = ifces
```
mazdakn: we should also copy here.

felix/nftables/table.go
```diff
@@ -482,6 +524,58 @@ func (n *NftablesTable) IPVersion() uint8 {
...
+	t.enableFlowtable()
+}
+
+// SetExternalDevices updates the set of host/external interfaces for the flowtable. Called by
+// the flowtable manager as matching interfaces come up and go down.
+func (t *NftablesTable) SetExternalDevices(ifces []string) {
+	t.externalDevices = ifces
```
mazdakn: same.

libcalico-go/lib/clientv3/felixconfig_e2e_test.go
```diff
@@ -53,6 +53,7 @@ var _ = testutils.E2eDatastoreDescribe("FelixConfiguration tests", testutils.Dat
 		BPFConnectTimeLoadBalancing:    &ctlbEnabled,
 		BPFHostNetworkedNATWithoutCTLB: &hostNetworkedNATDisabled,
 		NFTablesMode:                   ptr.To(apiv3.NFTablesModeAuto),
+		NFTablesFlowTableOffload:       ptr.To(apiv3.NFTablesFlowTableOffloadEnabled),
```
mazdakn: nit: you can use `new(apiv3.NFTablesFlowTableOffloadEnabled)` instead. A new change in golang 1.26.

felix/dataplane/linux/endpoint_mgr.go
```diff
@@ -159,6 +159,8 @@ type endpointManager struct {
 	arpTable Table
 	arpMaps  nftables.MapsDataplane
 
+	ifceHandler nftables.FlowTableHandler
```
mazdakn: nit: flowtableHandler?

felix/nftables/table.go
caseydavenport: Good call, done.

felix/iptables/actions.go
caseydavenport: Agreed, done. It can only be reached by a programming error since the renderer guards on nftables mode, so a loud panic beats a silent nil that would blow up later at render time.

felix/nftables/actions.go
caseydavenport: This matches the marker-field convention the other actions here use (TypeAccept, TypeDrop, etc.), none of which are referenced directly. I left it in for consistency rather than making this one action an outlier.

mazdakn: hmm interesting. Do you know what's the purpose?

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1150,8 +1151,29 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 	}
 
 	var filterMaps nftables.MapsDataplane
+	var ifceHandlerV4 nftables.FlowTableHandler
```
mazdakn: nit: there are still some places like this one that this var is named `ifaceHandler`  but should be `flowtableHandler` for sake of readability. 
These are mostly in `felix/dataplane/linux/` files.

mazdakn (approved):
Thanks. LGTM.
Just a few more nits.

caseydavenport:
Handled the cleanup. When offload is disabled, Felix now deletes the leftover flowtable object on the next resync (the forward rule that referenced it is already reconciled away). Added a unit test plus an FV that flips offload off and checks the flowtable is removed. 446256d

felix/dataplane/linux/int_dataplane.go
caseydavenport: Done, renamed these to flowtableHandlerV4/V6 (and the field/params in endpoint_mgr.go).

felix/dataplane/linux/endpoint_mgr.go
caseydavenport: Done.

felix/nftables/table.go
caseydavenport: Done.

felix/nftables/table.go
caseydavenport: Done.

libcalico-go/lib/clientv3/felixconfig_e2e_test.go
caseydavenport: Tried it, but new() still needs a type on our current toolchain (go1.26.4), and NFTablesFlowTableOffloadEnabled is a constant, so it doesn't compile. Kept ptr.To for now, which also matches the surrounding lines. Happy to switch over once that lands in our build image.

felix/nftables/actions.go
caseydavenport: Honestly not sure of the original intent. Every action struct here carries one of these Type* markers and none are referenced anywhere, so I think it's vestigial (probably from the original iptables port). I kept it just to match the others. Could strip them all in a separate cleanup if we want.

mazdakn:
@caseydavenport nice. Thanks for taking care of the cleaning part. Just reviewed it, LGTM.

felix/nftables/flowoffload_detect.go
```diff
@@ -0,0 +1,69 @@
...
+	tx := nft.NewTransaction()
+	tx.Add(&knftables.Table{})
+	tx.Add(&knftables.Flowtable{
+		Name:     "probe",
+		Priority: &prio,
+	})
+	if err := nft.Run(ctx, tx); err != nil {
+		logrus.WithError(err).Debug("Kernel rejected the flowtable probe; nftables flowtable offload is unsupported.")
+		return false
+	}
```
mazdakn, line 58: Does it make sense to try to load the kernel module here?

caseydavenport:
@mazdakn heads up - I switched the flowtable offload default back to Disabled for now. Offloaded flows take the fast path and skip the mangle/filter hooks, so DSCP marking and connection limits do not apply while a flow is offloaded (the nftables FV caught it). Figured it is better to ship this opt-in and revisit the default once it has had some real-world miles.

PR #9463, opened 2024-11-12
Add release-build and release-publish targets to key-cert-provisioner

danudey:
We've added key-cert-provisioner to release publishing, but it doesn't include the actual release targets we need to build and publish key-cert-provisioner images. This PR adds those images.

### DIFF
--- a/key-cert-provisioner/Makefile
+++ b/key-cert-provisioner/Makefile
@@ -108,9 +108,3 @@ release-publish: release-prereqs .release-$(VERSION).published
 .release-$(VERSION).published:
 	$(MAKE) push-images-to-registries push-manifests IMAGETAG=$(VERSION) RELEASE=$(RELEASE) CONFIRM=$(CONFIRM)
 	touch $@
-
-## Pushes a github release and release artifacts produced by `make release-build`.
-release-publish: release-prereqs .release-ksp-$(VERSION).published
-.release-ksp-$(VERSION).published:
-	$(MAKE) push-images-to-registries push-manifests IMAGETAG=$(VERSION) RELEASE=$(RELEASE) CONFIRM=$(CONFIRM)
-	touch $@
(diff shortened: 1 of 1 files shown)


key-cert-provisioner/Makefile
```diff
@@ -55,11 +55,9 @@ $(BINDIR)/test-signer-$(ARCH):
 # BUILD IMAGE
 ###############################################################################
 .PHONY: image-all
-image-all: $(addprefix sub-image-,$(VALIDARCHES)) sub-image-fips-amd64
```
radTuti, line 58: why is FIPS taken out? pretty sure we still publish FIPS images for OSS

key-cert-provisioner/Makefile
```diff
@@ -93,3 +91,26 @@ clean:
...
+release-publish: release-prereqs .release-$(VERSION).published
+.release-$(VERSION).published:
+	$(MAKE) push-images-to-registries push-manifests IMAGETAG=$(VERSION) RELEASE=$(RELEASE) CONFIRM=$(CONFIRM)
+	touch $@
+
+## Pushes a github release and release artifacts produced by `make release-build`.
+release-publish: release-prereqs .release-ksp-$(VERSION).published
+.release-ksp-$(VERSION).published:
+	$(MAKE) push-images-to-registries push-manifests IMAGETAG=$(VERSION) RELEASE=$(RELEASE) CONFIRM=$(CONFIRM)
+	touch $@
```
radTuti, line 116: this is duplicated

radTuti (changes requested):

key-cert-provisioner/Makefile
danudey: Theoretically we do, but from what I've been able to tell we've literally never published a FIPS KSP image; there weren't any on quay.io or docker hub, so it seems as though it's never happened, and adding in FIPS building for KSP broke the test image that we also build but which does not build a FIPS version. @rene-dekker confirmed that we don't need it (but that if a user wanted it it's easy for them/us to build).

key-cert-provisioner/Makefile
radTuti, line 116: you do not need both. the last target seems redundant

radTuti (changes requested):

radTuti (approved):

PR #9479, opened 2024-11-15
Add setNodeNetworkUnavailable option for calico node

yuzhiquan:
## Description

New feature, add an option `setNodeNetworkUnavailable`, default is true, control the behavior after calico-node is ready
and the calico node to set `NodeNetworkUnavailable = false` or not.
Some times, in our cluster, we have noticed that the node is ready and `NodeNetworkUnavailable = false`. but our network is not ready to accept the requests, so we want to add this option to control this.
And we also can find the options in other network projects like flannel or cilium.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add setNodeNetworkUnavailable option for calico node
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/cmd/calico-node/main.go
+++ b/node/cmd/calico-node/main.go
@@ -53,6 +53,7 @@ var monitorAddrs = flagSet.Bool("monitor-addresses", false, "Monitor change in n
 var runAllocateTunnelAddrs = flagSet.Bool("allocate-tunnel-addrs", false, "Configure tunnel addresses for this node")
 var allocateTunnelAddrsRunOnce = flagSet.Bool("allocate-tunnel-addrs-run-once", false, "Run allocate-tunnel-addrs in oneshot mode")
 var monitorToken = flagSet.Bool("monitor-token", false, "Watch for Kubernetes token changes, update CNI config")
+var setNodeNetworkUnavailable = flagSet.Bool("set-node-network-unavailable", true, "Set nodeNetworkUnavailable after calico startup")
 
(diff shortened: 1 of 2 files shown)


yuzhiquan:
> Is it Calico that is unable to accept requests or something else in the network?

We believe the root cause is the number of our nodes has achieved the quota of full-mesh mode, so we have plan to enable RR in the future, this pr just wants to give another choice in this condition, we can mark the node as ready when we make sure the network is ready for receive requests.

> How do you plan to configure this CLI flag - I don't think there's a way currently to set this when installing e.g., via the operator?

Sorry, I'm not familiar with this part. Do you have any suggestions?

caseydavenport:
> Sorry, I'm not familiar with this part. Do you have any suggestions?

Yeah, it would need to be plumbed through the operator code here: https://github.com/tigera/operator/blob/master/pkg/render/node.go#L1280-L1310

Likely by introducing a a new API field to the `Installation` API: https://github.com/tigera/operator/blob/master/api/v1/installation_types.go#L32

e.g.,

```
spec:
  manageNodeConditions: Disabled
```

It would likely need to be passed through via an environment variable rather than a CLI flag (since the calico/node container configures each process using env vars typically). 

I think it's probably OK to add this as an option.. although I think it's just working around the real problem(s):

- Sounds like you're using BGP full mesh beyond its intended scale.
- We don't have good heuristics to determine when the node is available or not.

PR #9498, opened 2024-11-18
Enhance bpf interface autodetection

sridhartigera:
## Description

This PR enhances the host interface auto-detection logic for the eBPF dataplane.
with the new approach, BPF endpoint manager builds a set of trees based on the netlink information of each host interface.

For example, if we have 2 interfaces in a bond and vlan on top of the bond, vlan interface is at the root of the tree, bond is at level 1 and physical interfaces are at the leaves. In the case of single standalone interfaces, each interface is a single node tree. Attaching Tc,XDP programs follows the below logic
1. If the interface is a root node, attach Tc.
2. If the interface is a leaf node, attach XDP.
3. If the interface is neither root nor leaf, it is ignored.

In the above process, XDP and Tc programs are removed from the interfaces when the interface moves down the tree.

```
bpftool net
xdp:
enp1s0f0np0(2) driver id 965
enp1s0f1np1(4) driver id 968

tc:
lo(1) clsact/ingress cali_tc_preambl:[804] id 804
lo(1) clsact/egress cali_tc_preambl:[805] id 805
bond0.1001(6) clsact/ingress cali_tc_preambl:[974] id 974
bond0.1001(6) clsact/egress cali_tc_preambl:[976] id 976
bpfout.cali(7) clsact/ingress cali_tc_preambl:[844] id 844
bpfout.cali(7) clsact/egress cali_tc_preambl:[843] id 843
cali662a7526402(9) clsact/ingress cali_tc_preambl:[894] id 894
cali662a7526402(9) clsact/egress cali_tc_preambl:[895] id 895
vxlan.calico(10) clsact/ingress cali_tc_preambl:[823] id 823
vxlan.calico(10) clsact/egress cali_tc_preambl:[822] id 822
cali596936e74c1(11) clsact/ingress cali_tc_preambl:[896] id 896
cali596936e74c1(11) clsact/egress cali_tc_preambl:[889] id 889
calie500d5708cd(12) clsact/ingress cali_tc_preambl:[890] id 890
calie500d5708cd(12) clsact/egress cali_tc_preambl:[891] id 891
```

xdp attached to physical interface and tc to bond.vlan

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
ebpf: improved autodetection of device hierarchy like bonds and vlans. tc is attached to the root and xdp to the leaves.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -212,6 +212,16 @@ var zeroIface bpfInterface = func() bpfInterface {
 	return i
 }()
 
+type bpfIfaceNode struct {
+	name        string
+	index       int
+	masterIndex int
+	parentIface *bpfIfaceNode
+	children    map[int]*bpfIfaceNode
+}
+
+type bpfIfaceTrees map[int]*bpfIfaceNode
+
 type bpfInterfaceInfo struct {
 	ifIndex       int
 	isUP          bool
@@ -305,6 +315,7 @@ type bpfEndpointManager struct {
 	ifStateMap              *cachingmap.CachingMap[ifstate.Key, ifstate.Value]
 	removeOldJumps          bool
 	legacyCleanUp           bool
+	hostIfaceTrees          bpfIfaceTrees
 
 	jumpMapAlloc     *jumpMapAlloc
 	xdpJumpMapAlloc  *jumpMapAlloc
@@ -331,10 +342,9 @@ type bpfEndpointManager struct {
 	updatePolicyProgramFn func(rules polprog.Rules, polDir string, ap attachPoint, ipFamily proto.IPVersion) error
 
 	// HEP processing.
-	hostIfaceToEpMap        map[string]proto.HostEndpoint
-	wildcardHostEndpoint    proto.HostEndpoint
-	wildcardExists          bool
-	hostIfaceToSlaveDevices map[string]set.Set[string]
+	hostIfaceToEpMap     map[string]proto.HostEndpoint
+	wildcardHostEndpoint proto.HostEndpoint
+	wildcardExists       bool
 
 	// UT-able BPF dataplane interface.
 	dp bpfDataplane
@@ -482,6 +492,7 @@ func newBPFEndpointManager(
 		policiesToWorkloads:     map[proto.PolicyID]set.Set[any]{},
 		profilesToWorkloads:     map[proto.ProfileID]set.Set[any]{},
 		dirtyIfaceNames:         set.New[string](),
+		hostIfaceTrees:          make(bpfIfaceTrees),
 		bpfLogLevel:             config.BPFLogLevel,
 		logFilters:              config.BPFLogFilters,
 		hostname:                config.Hostname,
@@ -506,13 +517,12 @@ func newBPFEndpointManager(
 		// Note: the allocators only allocate a fraction of the map, the
 		// rest is reserved for sub-programs generated if a single program
 		// would be too large.
-		jumpMapAlloc:            newJumpMapAlloc(jump.TCMaxEntryPoints),
-		xdpJumpMapAlloc:         newJumpMapAlloc(jump.XDPMaxEntryPoints),
-		ruleRenderer:            iptablesRuleRenderer,
-		onStillAlive:            livenessCallback,
-		hostIfaceToEpMap:        map[string]proto.HostEndpoint{},
-		hostIfaceToSlaveDevices: map[string]set.Set[string]{},
-		opReporter:              opReporter,
+		jumpMapAlloc:     newJumpMapAlloc(jump.TCMaxEntryPoints),
+		xdpJumpMapAlloc:  newJumpMapAlloc(jump.XDPMaxEntryPoints),
+		ruleRenderer:     iptablesRuleRenderer,
+		onStillAlive:     livenessCallback,
+		hostIfaceToEpMap: map[string]proto.HostEndpoint{},
+		opReporter:       opReporter,
 		// ipv6Enabled Should be set to config.Ipv6Enabled, but for now it is better
 		// to set it to BPFIpv6Enabled which is a dedicated flag for development of IPv6.
 		// TODO: set ipv6Enabled to config.Ipv6Enabled when IPv6 support is complete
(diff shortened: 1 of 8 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1122,87 +1131,37 @@ func (m *bpfEndpointManager) onInterfaceUpdate(update *ifaceStateUpdate) {
 		return
 	}
 
+	// Should be safe without the lock since there shouldn't be any active background threads
```
tomastigera, line 1134: why did you move this down here?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1122,87 +1131,37 @@ func (m *bpfEndpointManager) onInterfaceUpdate(update *ifaceStateUpdate) {
...
+			// update the ifaceType, master ifindex if bond slave.
+			link, err := m.dp.getIfaceLink(update.Name)
 			if err != nil {
-				log.WithError(err).Warn("Failed to get master interface name. Slave devices not updated")
-			} else {
-				// Slave interface update comes first.
-				if _, ok := m.hostIfaceToSlaveDevices[netiface.Name]; !ok {
-					m.hostIfaceToSlaveDevices[netiface.Name] = set.New[string]()
+				log.Errorf("Failed to get interface information via netlink '%s'", update.Name)
+				curIfaceType = IfaceTypeL3
```
tomastigera, line 1148: why do we default to L3?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4201,10 +4211,246 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+			eintf.name = intf.name
+			eintf.index = intf.index
+			eintf.masterIndex = intf.masterIndex
+			eintf.parentIndex = intf.parentIndex
+		}
+		if val, exists := m.hostIfaces[attrs.Index]; !exists {
+			m.hostIfaces[attrs.Index] = intf
+		} else {
+			val.name = intf.name
+			val.index = intf.index
```
tomastigera, line 4255: it must already have the same index ;-)

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -212,6 +212,15 @@ var zeroIface bpfInterface = func() bpfInterface {
 	return i
 }()
 
+type bpfHostIface struct {
+	name        string
+	index       int
+	masterIndex int
+	parentIndex int
```
tomastigera, line 219: can the interface have both master and parent in the tree? It feels to me that that the system ifaces can have either or, but not both and it is just a naming convention. I wonder if it would make thing simple if we maintained only a parent for both in our internal tree. Or maybe I am confused since there is just one parent pointer.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4201,10 +4211,246 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+			m.hostIfaces[attrs.Index] = intf
+		} else {
+			val.name = intf.name
+			val.index = intf.index
+			val.masterIndex = intf.masterIndex
+			val.parentIndex = intf.parentIndex
+		}
+	} else if attrs.MasterIndex != 0 {
+		m.deleteHostIface(intf.name)
+		// If the master interface is in the tree, add it as a slave.
```
tomastigera, line 4261:
```suggestion
		// If the master interface is a root, add it as a slave.
```

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4201,10 +4211,246 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+		if val, exists := m.hostIfaces[attrs.MasterIndex]; exists {
+			intf.parentIface = val
+			val.children[attrs.Index] = intf
+			m.hostIfaces[attrs.MasterIndex] = val
+		} else {
+			// Need to search slave devices.
```
tomastigera, line 4267:
```suggestion
			// Need to search the tree to find the master
```

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4201,10 +4211,246 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+				intf.parentIface = masterIf
+				masterIf.children[attrs.Index] = intf
+			} else {
+				masterIface := &bpfHostIface{index: attrs.MasterIndex, children: make(map[int]*bpfHostIface)}
+				intf.parentIface = masterIface
+				masterIface.children[attrs.Index] = intf
+				m.hostIfaces[attrs.MasterIndex] = masterIface
+			}
+		}
+		delete(m.hostIfaces, attrs.Index)
```
tomastigera, line 4279: why are you deleting it if you just added it and it could be the new root?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4201,10 +4211,246 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	} else if attrs.ParentIndex != 0 {
+		// bond0 already in list
```
tomastigera, line 4281: bond0 comment is irrelevant

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4201,10 +4211,246 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	}
+}
+
+func isLeafIface(hostIf *bpfHostIface) bool {
+	for _, child := range hostIf.children {
+		if child.name != "" {
+			return false
+		}
+	}
+	return true
```
tomastigera, line 4435:
```suggestion
	return len(hostIf.children) == 0
```
perhaps?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4201,10 +4211,246 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	for _, iface := range m.hostIfaces {
+		slaveIf, _ := findInIfaceTree(iface, -1, name, false)
+		if slaveIf != nil {
+			return slaveIf
+		}
+	}
+	return nil
+}
+
+func getAllIfaces(root *bpfHostIface, nodes *[]string) {
```
tomastigera, line 4419:
```suggestion
func getAllIfaces(root *bpfHostIface) []string {
```
perhaps?

tomastigera (changes requested):

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: In our FVs, eth0 is a veth which has a parent index. Hence the parent interface will be added as a child but it wont have a name, since felix wont get the update. This results in len(hostIf.children) > 0 but its a invalid child. Hence I did not do this way.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4201,10 +4211,246 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
 }
 
 func (m *bpfEndpointManager) getBondSlaves(masterIfName string) set.Set[string] {
-	if val, ok := m.hostIfaceToSlaveDevices[masterIfName]; ok {
+	slaveDevices := set.New[string]()
+	link, err := m.dp.getIfaceLink(masterIfName)
+	if err != nil {
+		log.Errorf("Error getting link for interface %s, err = %v", masterIfName, err)
+		return slaveDevices
```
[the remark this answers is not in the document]
sridhartigera: Have changed to return an empty slice of string.

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: When adding an interface with masterIndex !=0 (which means the intf is a slave intf), we need to delete the interface from the tree before adding it because the interface might have moved from standalone to a bond. Hence the deletion.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -305,6 +314,7 @@ type bpfEndpointManager struct {
 	ifStateMap              *cachingmap.CachingMap[ifstate.Key, ifstate.Value]
 	removeOldJumps          bool
 	legacyCleanUp           bool
+	hostIfaces              map[int]*bpfHostIface
```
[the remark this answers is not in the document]
sridhartigera: changed it to hostIfaceTrees and introduced a new type bpfIfaceTrees

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4201,10 +4211,246 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	attrs := link.Attrs()
+	intf := &bpfHostIface{name: attrs.Name,
+		index:       attrs.Index,
+		masterIndex: attrs.MasterIndex,
+		parentIndex: attrs.ParentIndex,
+		children:    make(map[int]*bpfHostIface)}
+
+	if attrs.MasterIndex == 0 && attrs.ParentIndex == 0 {
+		eintf := m.findHostIfaceByIndex(attrs.Index)
+		if eintf != nil {
```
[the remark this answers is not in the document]
sridhartigera: I just rewrite the name so that I don't overwrite the children information. For example, when felix gets the child information first, we create the child node and also the parent node with master index. Now when felix gets the update of the parent interface, the interface with that index is already there in the tree. Just need to update the name.

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: when we don't get the link information, I default the type to L3 and if the interface matches in `dataIfacePattern`, I change it to Data.

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: I take the lock in `cleanupOldAttach` to get the policyIdx info from the state. Hence moved the lock in the calling function below the block where we call `cleanupOldAttach`

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4371,62 +4420,25 @@ func (m *bpfEndpointManager) getRootInterface(hostIf *bpfHostIface) *bpfHostIfac
...
-	}
-}
-
-func isLeafIface(hostIf *bpfHostIface) bool {
+// isLeafIface returns if the interface does not have a valid child.
+// This can be as simple as return len(hostIf.children) == 0.
+// In the case of FVs, eth0 is a veth which has a parent interface.
+// Though eth0 is the only valid interface in the tree, we add the
+// parent interface as a child, Hence we need to be sure if there are
+// valid child interfaces.
```
tomastigera, line 4440:
```suggestion
// However, the main interfaces may have a parent in a different namespace, like a veth,
// and we are not going to see and update from that parent.
```

tomastigera (changes requested):
I have come mostly naming nits, lgtm in general

tomastigera (approved):

PR #9514, opened 2024-11-21
Reduce memory used to store inactive policies/profiles

fasaxc:
## Description

The `Rule` struct that is used in policies and profiles has grown very large over time.  With allocation overheads, policies end up using approximately 1KB per rule!

We were previously storing pointers to the policies in two places.  We need to change both in order to realise any savings:

- The active rules calculator needs the full policy struct to be available when it determines that a policy is active.
- The policy resolver/policy sorter only need to know the name and order of the policy.

We tackle them in two different ways:

- In the active rules calculator, store inactive policies in a compressed format; first encoding them as JSON, then compressing with "snappy" (a fast compression algorithm).  The data is already marked up for JSON serialisation (as used by Typha to serialise the stream to Felix) so this should be very safe.  Unpack the policies on demand. Notes:

  - Unpacks should happen rarely, only when a policy goes from inactive to active.
  - We could optimise the storage of active policies in future by changing from a sparse struct to something like a slice of match criteria.  But using the encoding we already had seemed like an easy win.

- In the policy resolver, avoid storing the whole policy and extract the needed metadata in a compact struct.  Since we're transforming the data anyway, map policy order "default" to float `+Inf`.  This sorts exactly as desired and avoids awkward special case logic.

I measured the saving using 1,000 namespaces, each with one k8s NetworkPolicy that allowed DNS traffic (UDP and TCP to a handful of CIDRs), traffic within the namespace and traffic from an external namespace.  (This was a policy supplied by a user who was running into occupancy problems with 10s of thousands of namespaces.)

* Before the changes, each namespace+policy used ~25KiB (heap inuse space).
* After the changes, that's down to 7.2KiB.

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Felix now compresses network policies and namespaces in memory, reducing RAM usage in clusters with many policies/namespaces that are not active on a given node.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/packedmap/packedmap.go
+++ b/libcalico-go/lib/packedmap/packedmap.go
@@ -43,7 +43,7 @@ func Make[K comparable, VIn, VPacked any](encoder Encoder[VIn, VPacked]) RawPack
 func MakeCompressedJSON[K comparable, V any]() Map[K, V] {
 	return Map[K, V]{
 		Make[K, V, string](
-			SnappyEncoderWrapper[V]{
+			SnappyEncoderWrapper[V, JSONEncoder[V]]{
 				encoder: JSONEncoder[V]{},
 			},
 		),
@@ -61,7 +61,7 @@ type Map[K comparable, V any] struct {
 func MakeDedupedCompressedJSON[K comparable, V any]() Deduped[K, V] {
 	pm := Make[K, V, unique.Handle[string]](
 		DedupingEncoderWrapper[V, string]{
-			encoder: SnappyEncoderWrapper[V]{
+			encoder: SnappyEncoderWrapper[V, JSONEncoder[V]]{
 				encoder: JSONEncoder[V]{},
 			},
 		},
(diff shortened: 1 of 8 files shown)


libcalico-go/lib/packedmap/packedmap.go
```diff
@@ -0,0 +1,160 @@
...
+// SnappyEncoderWrapper is an Encoder that wraps another Encoder and compresses
+// the packed values using snappy.
+type SnappyEncoderWrapper[V any] struct {
+	encoder JSONEncoder[V]
+}
+
+func (p SnappyEncoderWrapper[V]) Pack(val V) string {
+	buf := p.encoder.Pack(val)
+	var arr [1024 * 16]byte
+	packed := snappy.Encode(arr[:], []byte(buf))
```
tomastigera, line 129: packed would be a subslice of arr if arr was large enough. Does it mean that it always takes 16k at least?

tomastigera (approved):
LGTM with one question, looks like s straight forward refactoring

libcalico-go/lib/packedmap/packedmap.go
fasaxc: Tweaked it to re-use the same cached buffer over and over.  That will force the compiler to allocate the string, as desired.  (Right now, the go compiler does allocate but I couldn't find anything in the language spec to say it always will.)

marvin-tigera:

Issue #9517, opened 2024-11-22
Classic Calico-Installation  (kubectl apply) should have a smaller footprint 


erkules:
In CAPI there is a way to deploy K8s-objects using ConfigMaps (via ClusterRessourceSets).
The K8s-obejects to deploy Calico are above that 1MB limit for storing Objects in etcd.

~~~
$ du -sh tigera-operator.yaml 
1,6M    tigera-operator.yaml
~~~

Cilium uses not so much:

~~~
$ du -sh cilium.yaml 
64K     cilium.yaml
~~~

It would be nice to have a smaller footprint to deploy calico via ClusterRessourceSets in CAPI.

caseydavenport:
This PR removes all of the CRD content from tigera-operator.yaml: https://github.com/projectcalico/calico/pull/9518

PR #9518, opened 2024-11-22
Remove CRDs from tigera-operator.yaml

caseydavenport:
## Description

Trying this out. It removes the CRDs from tigera-operator.yaml to reduce the size of the manifest.

Instead, it configures the operator pod itself to create these CRDs on startup.

The main consequences I can see are:

- operator-init container in OCP will try to create resources before CRDs exist. May need to adjust the code.
- users will need to wait for operator pod to be running before applying `custom-resources.yaml`

But I think this is probably worth trying out.

## Related issues/PRs

Fixes https://github.com/projectcalico/calico/issues/9517

Requires: https://github.com/tigera/operator/pull/3610

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
To reduce manifest size, tigera-operator.yaml no longer includes CustomResourceDefinitions. Instead, the tigera-operator pod will deploy these resources on startup.
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
Actually looking at it, I think this doesn't impact OCP since we install CRDs as distinct manifests in that env.

caseydavenport:
@lwr20 I think there is mileage in that, although I'd need to think about what work is required on our end to make that happen. Not obvious how we get our generation code to do that nicely.

There's a part of me that says "Let's just roll this out and see how it fares" - it's still possible that even including our operator CRDs can exceed maximum sizes in some cases (if not now, then eventually)

coutinhop:
LGTM, one question about `manifests/ocp/policy.networking.k8s.io_adminnetworkpolicies.yaml`: that's not a calico CRD per se, but it might be nice to have the operator take care of it and to remove the yaml from here as well, what do you think?

charts/tigera-operator/templates/tigera-operator/02-tigera-operator.yaml
```diff
@@ -88,6 +91,14 @@ spec:
             defaultMode: 0777
             name: install-resources-script
       initContainers:
+        # Install CRDs first, so that APIs exist for calicoctl to use.
+        - name: bootstrap-crds
+          image: {{ template "tigera-operator.image" .Values.tigeraOperator}}
+          command:
+            - operator
+          args:
+            - -bootstrap-crds
```
coutinhop, line 100: need to add `-variant` arg from https://github.com/tigera/operator/pull/3610 ?

coutinhop (approved):

PR #9528, opened 2024-11-25
Use longer (and thus more descriptive) nftables chain names

caseydavenport:
## Description

nftables supports longer chain names than iptables. This means we don't need to 
truncate / hash chain names as much in nftables mode.

Until now, we've been enforcing the same character length of 28. This PR bumps it up
to 256 - which should result in much more human readable chain names.

Chain names now look something like this:

```
chain filter-cali-po-default/namespace/tier.policy-name {
```

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
In nftables dataplane, use human readable policy chain names where possible. 
```

## Reminder for the reviewer

Other optional labels:


sridhartigera (approved):
LGTM

PR #9538, opened 2024-11-28
[v3.29] Free dedupe buffer's tracking map after shrink

fasaxc:
## Description

Go doesn't free map blocks, even after a map shrinks considerably. The dedupe buffer tends to store a lot of keys for a start-of-day snapshot, make sure we clean up the leaked map capacity once we're back down to zero.

## Related issues/PRs

Upstream issue: https://github.com/golang/go/issues/20135
Pick of #9526 

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Felix: fix that a map used to cache loaded datastore keys would always use RAM proportional to the total number of keys rather than shrinking when no longer needed.
```

## Reminder for the reviewer

Other optional labels:


mazdakn (approved):
LGTM

PR #9539, opened 2024-11-28
Configurable CNI token refresh duration

### LINKED ISSUE #9235: CNI token refresh can take extra 6hrs to happen after expiration
## Expected Behavior

Expired tokens are refreshed within an acceptable timeframe, e.g. few minutes. 

## Current Behavior

Currently, the token refresher sleeps for (exp - now)/4 = max. 6hrs. When the refresher is asleep, if the token expires, it can only be refreshed by restarting all pods, otherwise the cluster networking is down until the sleep is completed. These values are hardcoded in, see https://github.com/projectcalico/calico/blob/3bca3221116db2fb368cfce2508517acd05579e9/node/pkg/cni/token_watch.go#L24. 

Techically, other times could be inserted with the serviceaccount/token file, but that is only used if the token API is down, see https://github.com/projectcalico/calico/blob/3bca3221116db2fb368cfce2508517acd05579e9/node/pkg/cni/token_watch.go#L104. 

## Possible Solution

This could be fixed by 1. making the hardcoded values configurable, either with a separate configuration or by (optionally) preferring the configuration from file or 2. refreshing straight after a failed request. 

1. Get a cluster with calico installed
2. Stop NTP service 
3. Shift time forward by more than 24hrs
4. No pods are coming up or down, calico plugin connection is unauthorized

From the code we can deduce that calico is sleeping and will refresh in 6hrs or so.

## Context

This issue was noticed in testing, but it's still causing the tests to take 6hrs to complete. The same behavior can happen in production if the token for some reason expires earlier than expected. 

## Your Environment

For the reproduction I just put up a fresh VM with Ubuntu 24, installed chrony and followed this https://docs.tigera.io/calico/latest/getting-started/kubernetes/kind to create a Kind cluster with calico

huutomerkki:
## Description

~This PR changes the dynamic timing to constant once per minute refresh rate for the CNI token, as per the discussion in slack about the possible solution for the linked issue.~

Adds a configuration option CNI_TOKEN_REFRESH_INTERVAL to control how frequently CNI token should be refreshed. 

I have manually verified that following the reproduction steps with these changes included in the calico-node image fixes the issues described in #9235. 

## Related issues/PRs

Fixes #9235 

## Todos

- [?] Tests
- [?] Documentation
- [?] Release note

## Release Note

```release-note
Configurable CNI token refresh duration
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/pkg/cni/token_watch.go
+++ b/node/pkg/cni/token_watch.go
@@ -125,7 +125,6 @@ func (t *TokenRefresher) Stop() {
 
 func (t *TokenRefresher) Run() {
 	var nextExpiration time.Time
-	rand := rand.New(rand.NewSource(time.Now().UnixNano()))
 	for {
 		tu, err := t.UpdateToken()
 		if err != nil {
@@ -141,16 +140,8 @@ func (t *TokenRefresher) Run() {
 			}
 
 		}
-		now := time.Now()
-		var sleepTime time.Duration
 		// Do some basic rate limiting to prevent flooding the kube apiserver with requests
-		if nextExpiration.Before(now.Add(t.minTokenRetryDuration * t.defaultRefreshFraction)) {
-			sleepTime = t.minTokenRetryDuration
-		} else {
-			sleepTime = nextExpiration.Sub(now) / t.defaultRefreshFraction
-		}
-		jitter := rand.Float32() * float32(sleepTime)
-		sleepTime += time.Duration(jitter)
+		sleepTime := t.getSleepTime(&nextExpiration)
 		if logrus.IsLevelEnabled(logrus.DebugLevel) {
 			logrus.Debugf("Going to sleep for %s", sleepTime.String())
 		}
@@ -160,6 +151,29 @@ func (t *TokenRefresher) Run() {
 			return
 		}
 	}
+}
+
+func (t *TokenRefresher) getSleepTime(nextExpiration *time.Time) time.Duration {
+	now := time.Now()
+	sleepTime := nextExpiration.Sub(now) / t.defaultRefreshFraction
+	const cniTokenRefreshIntervalName = "CNI_TOKEN_REFRESH_INTERVAL"
+	cniTokenRefreshInterval := os.Getenv(cniTokenRefreshIntervalName)
+	if cniTokenRefreshInterval != "" {
+		duration, err := time.ParseDuration(cniTokenRefreshInterval)
+		if err == nil {
+			logrus.WithField("interval", duration).Debugf("Detected a valid %s", cniTokenRefreshIntervalName)
+			sleepTime = duration
+		} else {
+			logrus.WithError(err).WithField(cniTokenRefreshIntervalName, cniTokenRefreshInterval).Errorf("Detected an invalid %s.", cniTokenRefreshIntervalName)
+		}
+	}
+	if nextExpiration.Before(now.Add(t.minTokenRetryDuration * t.defaultRefreshFraction)) {
+		sleepTime = t.minTokenRetryDuration
+	}
+	rand := rand.New(rand.NewSource(time.Now().UnixNano()))
+	jitter := rand.Float32() * float32(sleepTime)
+	sleepTime += time.Duration(jitter)
+	return sleepTime
 }
 
 func (t *TokenRefresher) tokenRequestSupported(clientset *kubernetes.Clientset) bool {


caseydavenport:
@huutomerkki thanks for raising this PR!

I want to push a little bit on the aggressive timer here - for some of our larger scale users, this is going to result in a large increase  in the number of TokenRequest API calls that Calico makes.

We have users running 5k node clusters (and even larger!), which will result 5k requests per 60 seconds, or 83 requests per second across the entire cluster! I think this is probably just not workable for that scale.

Some thoughts (in no particular order):

- I think refreshing after a failed request is probably the ideal approach, but we need to be careful that we don't allow the CNI plugin to indefinitely propagate its own token. It would need to be some sort of privileged communication back from the CNI plugin to calico/node to trigger this, which might be tricky.
- Perhaps just a larger timeout (say, every 1h instead of every 6h) would satisfy my concerns, but I think that wouldn't meet your goals IIRC.
- Perhaps we just need this to be configurable, so that it can maintain its reasonable default of 6h but allow users with stricter requirements to configure it to be smaller. I don't love the idea of exposing config for this, as it's a fairly low level thing that most users don't need to care about, but sometimes that is required.

I think my preferred option would be to find a way to communicate back to the token refresher that it needs to generate a new token, but I think exposing such a channel is risky and has potential for abuse if we're not careful (imagine an attacker getting access and requesting non-stop token refresh attempts. We'd need a backstop to rate limit this for sure). This trigger would likely need to be something like a restricted UDS or file-based trigger that the CNI plugin can write to when needed in response to a permissions error - WDYT?

caseydavenport:
> Adding a config for this would fulfill my goal, and I think it would be something that I could implement.

Great, lets go with this as an approach for now and I will see about a more proactive refresh token in my spare time if I can manage it :)

For a config option, I suspect it would look something like this:

- A new environment variable - e.g., `CNI_TOKEN_REFRESH_DURATION`
- When unset, the existing behavior (1/4 of the token expiry) would be used.
- When set, the variable would be respected instead.

Then we'd also want a corresponding change to the `Installation` API in the `tigera/operator` repository: https://github.com/tigera/operator/blob/master/api/v1/installation_types.go#L32-L211

This would be the end-user facing API that would configure the environment variable - we can talk about the specifics of what that looks like when we get there :)

kashifest:
@huutomerkki would you wish to continue to work on this ?

caseydavenport (approved):

node/pkg/cni/token_watch.go
```diff
@@ -162,6 +154,21 @@ func (t *TokenRefresher) Run() {
 	}
 }
 
+func (t *TokenRefresher) getSleepTime(nextExpiration *time.Time) time.Duration {
+	now := time.Now()
+	sleepTime := nextExpiration.Sub(now) / t.defaultRefreshFraction
+	if duration, err := time.ParseDuration(os.Getenv("CNI_TOKEN_REFRESH_DURATION")); err == nil {
+		logrus.Debugf("Detects a valid CNI_TOKEN_REFRESH_DURATION")
```
caseydavenport, line 161:
```suggestion
		logrus.WithField("interval", duration).Debugf("Detected a valid CNI_TOKEN_REFRESH_DURATION")
```

caseydavenport: ^ We should log out the value

node/pkg/cni/token_watch.go
```diff
@@ -162,6 +154,21 @@ func (t *TokenRefresher) Run() {
 	}
 }
 
+func (t *TokenRefresher) getSleepTime(nextExpiration *time.Time) time.Duration {
+	now := time.Now()
+	sleepTime := nextExpiration.Sub(now) / t.defaultRefreshFraction
+	if duration, err := time.ParseDuration(os.Getenv("CNI_TOKEN_REFRESH_DURATION")); err == nil {
```
caseydavenport, line 160: I think we should refactor this so that if there is an error, we log an Error log indicating such. We want to avoid users setting this to an invalid value and not seeing any feedback.

caseydavenport, line 160:
```suggestion
	if duration, err := time.ParseDuration(os.Getenv("CNI_TOKEN_REFRESH_INTERVAL")); err == nil {
```

A bit of a nit, but INTERVAL is a more appropriate term here - we're measuring the time between two events, not the length of a singular event.

caseydavenport (changes requested):
Thanks for the update! Just a few small comments but largely this looks correct to me.

mquhuy:
Hi @caseydavenport. I appreciate your thoughtful review. I've incorporated your feedback and updated the pull request. Please let me know if there's anything else that needs attention!

Btw, I noticed now that as I'm not the original author of this PR, I cannot edit its title and description. Could you please suggest if we should create a new PR, or is it possible for me to update these information somehow?

caseydavenport:
@mquhuy I believe the tests that are failing have been fixed in master - could you try merging / rebasing this PR on the latest master?

mquhuy:
@caseydavenport Thank you very much! I've rebased it now

caseydavenport:
Thanks @mquhuy !

PR #9543, opened 2024-12-01
feat: added calico-vxlan-inter-workload.yaml manifest

faizan-exe:
## Description

Type of Fix: New Feature

Why this PR should be merge?
While installing Calico CNI with VXLAN Encapsulation, people often tend to forget whether it is CrossSubnet or Inter-workload configuration. By adding a separate file for inter-workload configuration, it will be easy for people to differentiate between the two and use the configuration as required instead of modifying the calico-vxlan.yaml file manually for interworkload configurartion

Components affected:
- manifest

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/manifests/calico-vxlan-inter-workload.yaml
+++ b/manifests/calico-vxlan-inter-workload.yaml
@@ -1,0 +1,6590 @@
+---
+# Source: calico/templates/calico-kube-controllers.yaml
+# This manifest creates a Pod Disruption Budget for Controller to allow K8s Cluster Autoscaler to evict
+
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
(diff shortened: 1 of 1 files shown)


caseydavenport:
@faizan-exe thanks for the PR - however, I don't think this aligns with how we want to manage manifest files in this repository.

From experience, it becomes very difficult to maintain a fleet of manifests for every possible configuration that a user may want to set. So, I don't think we can accept this as-is. 

I would recommend looking at some of the existing mechanisms we have in place for configuration of Calico. Namely, the tigera/operator provides an easy way to customize the configuration of your install without needing an entirely new manifest.

Similarly, we ship a helm chart that allows configuration of this via values.yaml without an entirely separate manifest.

Issue #9491, opened 2024-11-18
IPv4-Only Pods in Dual-Stack


renyunkang:
If I've enabled dual-stack networking, how can I configure specific pods to use only IPv4 (single-stack)?

Is there a configuration method to achieve this, or is it currently unsupported?

mazdakn:
@renyunkang Have you tried `'cni.projectcalico.org/ipAddrsNoIpam'` annotation. More information here: https://docs.tigera.io/calico/latest/reference/configure-cni-plugins#requesting-a-specific-ip-address

renyunkang:
> @renyunkang Have you tried `'cni.projectcalico.org/ipAddrsNoIpam'` annotation. More information here: https://docs.tigera.io/calico/latest/reference/configure-cni-plugins#requesting-a-specific-ip-address

Thank you for your response. @mazdakn  "NoIpam" is not the desired outcome. I still want Calico to automatically manage IP addresses (IPAM), but I need a configuration option to specify which namespaces or pods should only use IPv4, instead of both IPv4 and IPv6, even with dual-stack enabled.

caseydavenport:
Yeah, I agree - the `ipAddrsNoIpam` is exceptionally niche and not most for 99.99% of users :grin:

caseydavenport:
I think an annotation to expose configuration of IP families is appropriate. One thing to think about is whether or not we want to unconditionally allow application developers (i.e., people deploying pods) to control which IP families they have access to regardless of cluster configuration. 

Generally for these types of annotations, you want this to be opt-in within the CNI config file itself in some way. Otherwise a cluster admin can configure Calico to behave one way (i.e., requiring dual stack) and then have app devs override that configuration when it's not desirable.

We may want to make this annotation available only if the cluster admin has deferred this capability to application developers - see the existing CNI config "feature_control" section which is used for this sort of thing: https://github.com/projectcalico/calico/blob/840cb655c51f40782cfd49f688d1939e7e050abf/cni-plugin/pkg/types/types.go#L33-L37

caseydavenport:
Would be something like `IPFamilySpecification bool` added to that struct, and then checked before parsing the annotation.

caseydavenport:
On second thought.... maybe that's too paranoid. The `Service` struct already allows application devs to configure how their services work. This seems like a fair parallel.

PR #9544, opened 2024-12-02
supports custom ipFamilies

renyunkang:
## Description

After enabling Calico dual-stack networking (IPv4/IPv6) in a Kubernetes cluster, while all newly created Pods will receive both IPv4 and IPv6 addresses, Calico's configuration prevents the cluster from supporting IPv4-only or IPv6-only single-stack modes. This means that even legacy applications using only IPv4 require a restart to obtain dual-stack addresses, even though they don't need IPv6. Enabling dual-stack in a long-running single-stack cluster can therefore introduce several issues:

- Unnecessary Restarts: Forcing a restart on legacy applications that don't require IPv6 increases unnecessary operational burden and potential downtime.

- Increased IP Address Management Complexity: Even if legacy applications restart and obtain IPv6 addresses, these addresses are wasted if the application doesn't support or use IPv6. This adds complexity to IP address management.

Ideally, we'd like Calico to still allow existing Pods to use IPv4-only or IPv6-only single-stack modes after enabling dual-stack.

~~This can be achieved by adding a `cni.projectcalico.org/ipFamilies` annotation, ~~supporting values `ipv4Only`, `ipv6Only`, and `dualStack`~~ supporting values `IPv4`, `IPv6`.~~

This is configured using the `cni.projectcalico.org/ipFamilies` annotation. This annotation must be set to a JSON array. Valid values are: `["IPv4"]` for IPv4-only addressing, `["IPv6"]` for IPv6-only addressing, and `["IPv4", "IPv6"]` (or `["IPv6", "IPv4"]`) for dual-stack addressing. If a value other than these three is set, the annotation parsing will fail, and Pod creation will fail. If the annotation is not set, the existing ipam configuration from the CNI configuration file will be used.

Support for this annotation should be provided at both the `namespace` and `Pod` levels, with Pod-level annotations taking precedence over namespace-level annotations.

## Related issues/PRs

https://github.com/projectcalico/calico/issues/9491

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Support for a new cni.projectcalico.org/ipFamilies annotation to configure which IP families to use on a per-pod or per-Namespace basis.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/pkg/k8s/k8s.go
+++ b/cni-plugin/pkg/k8s/k8s.go
@@ -212,11 +212,12 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 		// Check for calico IPAM specific annotations and set them if needed.
 		if conf.IPAM.Type == "calico-ipam" {
 
-			var v4pools, v6pools string
+			var v4pools, v6pools, ipFamilies string
 
 			// Sets  the Namespace annotation for IP pools as default
 			v4pools = annotNS["cni.projectcalico.org/ipv4pools"]
 			v6pools = annotNS["cni.projectcalico.org/ipv6pools"]
+			ipFamilies = annotNS["cni.projectcalico.org/ipFamilies"]
 
 			// Gets the POD annotation for IP Pools and overwrites Namespace annotation if it exists
 			v4poolpod := annot["cni.projectcalico.org/ipv4pools"]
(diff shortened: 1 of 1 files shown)


cni-plugin/pkg/k8s/k8s.go
```diff
@@ -227,8 +228,12 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 			if len(v6poolpod) != 0 {
 				v6pools = v6poolpod
 			}
+			ipFamiliesPod := annot["cni.projectcalico.org/ipFamilies"]
```
caseydavenport, line 231: What do you think about making this symmetrical with the Services API, which has an identical concept?

> You can set .spec.ipFamilies to any of the following array values:
> ["IPv4"]
> ["IPv6"]
> ["IPv4","IPv6"] (dual stack)
> ["IPv6","IPv4"] (dual stack)

e.g.,

```
cni.projectcalico.org/ipFamilies: '["IPv4"]'
```

renyunkang: > What do you think about making this symmetrical with the Services API, which has an identical concept?
> 
> > You can set .spec.ipFamilies to any of the following array values:
> > ["IPv4"]
> > ["IPv6"]
> > ["IPv4","IPv6"] (dual stack)
> > ["IPv6","IPv4"] (dual stack)
> 
> e.g.,
> 
> ```
> cni.projectcalico.org/ipFamilies: '["IPv4"]'
> ```

I agree. Making it symmetrical with the Services API makes sense.

renyunkang:
PTAL. @caseydavenport

caseydavenport (approved):
This LGTM!

Would you mind submitting a PR to github.com/tigera/docs in order to update this page to document the new option?

https://docs.tigera.io/calico/latest/networking/ipam/use-specific-ip

Starttoaster:
Was this ever released? Looks like it never made it to the docs, but I also don't see it in release notes anywhere. Just curious, trying to use this now on 3.29.6 but it doesn't appear to be working.

caseydavenport:
Yeah, this should have been in v3.30.

caseydavenport:
Guessing there was never a docs PR raised..

PR #9546, opened 2024-12-02
fix: docs for upgrades to include CRD upgrade notes and added upgrade…

venkatamutyala:


charts/tigera-operator/README.md
```diff
@@ -76,7 +86,13 @@ ownership of the helm resources to the new chart location.
 
 ## All other upgrades
 
-1. Run the helm upgrade:
+1. Upgrade CRDs:
```
caseydavenport, line 89: It's worth noting that as of just a few days ago, we merged a PR that removes the need for this step.

For reference: https://github.com/projectcalico/calico/pull/9518

It moves management of these CRDs out of the helm chart and into our operator code, so that users won't need to perform this step :tada:

caseydavenport: This will only take effect in Calico v3.30 though, so what you have here is appropriate for older versions.

venkatamutyala: How long until 3.30 is released? Would it make sense for me to break my PR into two different PRs? I suspect the note for v3.28+ will still apply after v3.30 is released?

venkatamutyala: Just created: 
https://github.com/projectcalico/calico/pull/9557 - `master`
https://github.com/projectcalico/calico/pull/9555 - `release-v3.27`
https://github.com/projectcalico/calico/pull/9556 - `release-v3.28`

venkatamutyala: I believe this PR can be closed out now.

PR #9581, opened 2024-12-10
[BPF] Enhance conntrack map flexibility with CPU-based scaling

ioworker0:
## Description

- new feature

Previously, BPFMapSizeConntrack was set to a fixed value, which lacked
flexibility across different spec of machines. Now, with the introduction
of BPFMapSizePerCPUConntrack, the conntrack map size can be adjusted
dynamically as BPFMapSizeConntrackPerCPU * (Number of CPUs). This allows
for larger map sizes on high-spec machines and smaller map sizes on
lower-spec machines, optimizing resource usage accordingly.

BTW, recently, we added several high-spec machines to the data center,
which led to the conntrack map size being filled up on these machines. It
could have been avoided with BPFMapSizePerCPUConntrack IHMO.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Enhance conntrack map flexibility with CPU-based scaling.
```

## Reminder for the reviewer

Other optional labels:


tomastigera:
Might be related to https://calicousers.slack.com/archives/CPEPF833L/p1733131626596509

lwr20:
re. the lack of counters, in the thread that Tomas linked, @fasaxc said:
> I think we do have counters on master: https://github.com/projectcalico/calico/blob/master/felix/bpf/conntrack/bpf_scanner.go#L43
> I added them when implementing the new BPF-based conntrack cleaner

lwr20:
I have some nervousness about this approach to auto-scaling - it seems that the intent here is to scale the map size based on CPU size.  But map size is about memory, isn't it?  And _free_ memory on a node depends entirely about how tightly you pack workload pods onto the node.  So I'm not entirely sure about the basic premise of this PR.  I'm worried that we create an API that we have to support forever, but isn't a good experience for most users.

> Previously, BPFMapSizeConntrack was set to a fixed value, which lacked
flexibility across different spec of machines.

The "lacking flexibility" part is not entirely true.  The `default` felixconfig applies globally, but it can be overridden for particular nodes by making felixconfigs called `node.<nodename>`.  See https://docs.tigera.io/calico/latest/reference/resources/felixconfig#felixconfig

> Calico automatically creates a resource named default containing the global default configuration settings for Felix. You can use calicoctl to view and edit these settings
> The resources with the name node.<nodename> contain the node-specific overrides, and will be applied to the node <nodename>. When deleting a node the FelixConfiguration resource associated with the node will also be deleted.

Now clearly if you have a lot of large nodes, that's going to be a pain to configure.  ISTR @caseydavenport proposed a system where we add matchlabels to Felixconfigs, and you would then label the nodes so that the correct felixconfig applies to them, and you would only need a felixconfig per node-type in your cluster.  That might solve this issue in a way which works for more users?

ioworker0:
> re. the lack of counters, in the thread that Tomas linked, @fasaxc said:
> 
> > I think we do have counters on master: https://github.com/projectcalico/calico/blob/master/felix/bpf/conntrack/bpf_scanner.go#L43
> > I added them when implementing the new BPF-based conntrack cleaner

Ah... I didn't notice them before, and they are really useful as well, thanks!

However, when it comes to troubleshooting issues with the conntrack map, we still need counters for conntrack creation failure events, IHMO ;)

ioworker0:
> felixconfig per node-type

Thanks a lot for the points!

As a CNI like Calico, I believe that auto-scaling is appropriate. It might also be better to adjust the map size based on available memory, as you suggested, similar to how network stack parameters such as TCP buffer sizes are dynamically adjusted according to system memory in Linux kernel, IIRC.

Moreover, maintaining Felix configurations for each node-type could be a little challenging; nobody dislikes having less config, I guess. Kubernetes aims to free us from managing individual machines, allowing us to focus less on the specific types of machines within a cluster, IHMO.

what do you think?

ioworker0:
> I have some nervousness about this approach to auto-scaling - it seems that the intent here is to scale the map size based on CPU size. But map size is about memory, isn't it? And _free_ memory on a node depends entirely about how tightly you pack workload pods onto the node. So I'm not entirely sure about the basic premise of this PR. I'm worried that we create an API that we have to support forever, but isn't a good experience for most users.
> 
> > Previously, BPFMapSizeConntrack was set to a fixed value, which lacked
> > flexibility across different spec of machines.
> 
> The "lacking flexibility" part is not entirely true. The `default` felixconfig applies globally, but it can be overridden for particular nodes by making felixconfigs called `node.<nodename>`. See https://docs.tigera.io/calico/latest/reference/resources/felixconfig#felixconfig
> 
> > Calico automatically creates a resource named default containing the global default configuration settings for Felix. You can use calicoctl to view and edit these settings
> > The resources with the name node. contain the node-specific overrides, and will be applied to the node . When deleting a node the FelixConfiguration resource associated with the node will also be deleted.
> 
> Now clearly if you have a lot of large nodes, that's going to be a pain to configure. ISTR @caseydavenport proposed a system where we add matchlabels to Felixconfigs, and you would then label the nodes so that the correct felixconfig applies to them, and you would only need a felixconfig per node-type in your cluster. That might solve this issue in a way which works for more users?

Thanks a lot for the points!

As a CNI like Calico, I believe that auto-scaling is appropriate. It might also be better to adjust the map size based on available memory, as you suggested, similar to how network stack parameters such as TCP buffer sizes are dynamically adjusted according to system memory in Linux kernel, IIRC.

Moreover, maintaining Felix configurations for each node-type could be a little challenging; nobody dislikes having less config, I guess. Kubernetes aims to free us from managing individual machines, allowing us to focus less on the specific types of machines within a cluster, IHMO.

what do you think?

ioworker0:
@tomastigera 

I'm just wondering if it is possible for you to take a review?

ioworker0:
> @tomastigera

@tomastigera PING ~

felix/dataplane/linux/int_dataplane.go
```diff
@@ -796,6 +797,10 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 	bpfconntrack.SetCleanupMapSize(config.BPFMapSizeConntrackCleanupQueue)
 	bpfifstate.SetMapSize(config.BPFMapSizeIfState)
 
+	if config.BPFMapSizePerCPUConntrack > 0 {
+		bpfconntrack.SetMapSize(config.BPFMapSizePerCPUConntrack * bpfmaps.NumPossibleCPUs())
```
fasaxc, line 801: Please adjust so that `bpfconntrack.SetMapSize` is only called once after calculating which of the two settings should apply.  Easy to miss this second call when reading the code later.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -724,6 +724,12 @@ type FelixConfigurationSpec struct {
 	// an entry for each active connection.  Warning: changing the size of the conntrack map can cause disruption.
 	BPFMapSizeConntrack *int `json:"bpfMapSizeConntrack,omitempty"`
 
+	// BPFMapSizePerCPUConntrack determines the size of connectrack map based on the number of CPUs. If set it to a
+	// non-zero value, BPFMapSizeConntrack will be override as BPFMapSizePerCPUConntrack * (Number of CPUs).
+	// This map must be large enough to hold an entry for each active connection.  Warning: changing the size of the
+	// conntrack map can cause disruption.
```
fasaxc, line 730:
```suggestion
	// BPFMapSizePerCPUConntrack determines the size of conntrack map based on the number of CPUs. If set to a
	// non-zero value, overrides BPFMapSizeConntrack with `BPFMapSizePerCPUConntrack * (Number of CPUs)`.
	// This map must be large enough to hold an entry for each active connection.  Warning: changing the size of the
	// conntrack map can cause disruption.
```

fasaxc (changes requested):
I think this is a reasonable incremental improvement on what we have that doesn't do any harm if you don't actively use it.

Ideally, we'd auto-size the conntrack map based on load.  If we see it get full, double its size, but that's a lot more work.

It'd also be nice to have a config model that allows for configuring groups of similar nodes by selector.

ioworker0:
Thanks a lot for taking time to review!

> I think this is a reasonable incremental improvement on what we have that doesn't do any harm if you don't actively use it.

Yeah, this is a simple/harmless change that effectively solves the issue we're facing ;)

> 
> Ideally, we'd auto-size the conntrack map based on load. If we see it get full, double its size, but that's a lot more work.

Totally agreed. It's actually more like future-proofing work, so let's do it separately ;p

> 
> It'd also be nice to have a config model that allows for configuring groups of similar nodes by selector.

However, IMHO, it might be a bit challenging, as we currently lack a clear way to know how many groups of nodes there are, whether they are faster or slower ;(

api/pkg/apis/projectcalico/v3/felixconfig.go
ioworker0: Yep, It would be better.

felix/dataplane/linux/int_dataplane.go
ioworker0: Good catch! Will adjust as you suggested ;)

ioworker0:
> take a review

@fasaxc everything is ready, could please take a review again?

ioworker0:
> Thanks a lot for taking time to review!
> 
> > I think this is a reasonable incremental improvement on what we have that doesn't do any harm if you don't actively use it.
> 
> Yeah, this is a simple/harmless change that effectively solves the issue we're facing ;)
> 
> > Ideally, we'd auto-size the conntrack map based on load. If we see it get full, double its size, but that's a lot more work.
> 
> Totally agreed. It's actually more like future-proofing work, so let's do it separately ;p
> 
> > It'd also be nice to have a config model that allows for configuring groups of similar nodes by selector.
> 
> However, IMHO, it might be a bit challenging, as we currently lack a clear way to know how many groups of nodes there are, whether they are faster or slower ;(

@fasaxc everything is ready, could please take a review again?

ioworker0:
> /sem-approve

Okay, I cannot launch the workflow :(

fasaxc:
You need to rev the count of expected config params in `configurationprocessor_test.go`.

ioworker0:
> You need to rev the count of expected config params in `configurationprocessor_test.go`.

Ah, I see. IIUC, revving that to 153, right?

ioworker0:
> Waiting for status to be reported

Done.

ioworker0:
> > You need to rev the count of expected config params in `configurationprocessor_test.go`.
> 
> Ah, I see. IIUC, revving that to 153, right?

Done.

ioworker0:
What happened when doing 'Tests a /32 LB service with externalTrafficPolicy=Local using a RR'?

It seems that it has nothing to do with this PR :(

[1] https://tigera.semaphoreci.com/jobs/11d060de-4b84-4208-8e29-c435440df9c9#L1208

fasaxc:
Just a flake, it passed on re-run.

fasaxc:
Merged, thank you!

ioworker0:
> Merged, thank you!

Thanks a lot for your time! It was a wonderful journey ~

PR #9593, opened 2024-12-12
Watch bookmarks.

fasaxc:
## Description

Add support for [watch bookmarks](https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks) to the Kubernetes Datastore Driver and then use them in the watchersyncer.  This prevents a wasteful re-List() every time the API server does a compaction of a little-updated resource.

While researching watch bookmarks, I noticed that we were mishandling various errors relative to the Kubernetes reflector/informer machinery.  In particular, watch bookmarks are only useful if we don't reset the revision when we get "expected" errors such as "connection refused" or "too many requests".

Includes a refactoring of the main watchercache loop (all in the final commit):

- Split the god method and use defer to clean up the watcher instead of doing it on the various error paths.
- Resync/recreate watcher at the top of the loop only, not in error paths.  I find this style a lot easier to grok!
- Tune the logging: re-use an event-scoped logger to avoid allocs and make the stringification of a WatchEvent a lot more useful.

Note: this does **not** expose watch bookmarks on the "v3" watch API.  The challenge to doing that is that the watch bookmark is conveyed as an "empty" k8s resource with only a `resourceVersion` populated and translating that to a similar Calico object is non-trivial (because the existing conversion functions expect a valid, populated object as input).  

## Related issues/PRs

Builds on #9577, which added the options struct needed to plumb through the AllowWatchBookmarks flag. 
CORE-10820

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico now takes advantage of "watch bookmarks" when watching Kubernetes resources.  This prevents "resource version too old" errors, which result in unnecessary resyncs (which delay updates and increase API server CPU load).
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/api/api.go
+++ b/libcalico-go/lib/backend/api/api.go
@@ -221,6 +221,34 @@ type WatchEvent struct {
 	Error error
 }
 
+func (w WatchEvent) String() string {
+	switch w.Type {
+	case WatchAdded:
+		if w.New == nil {
+			break // Malformed event, but still want to be able to print it!
+		}
+		return fmt.Sprintf("Added:%v@%v", w.New.Key, w.New.Revision)
+	case WatchModified:
+		if w.New == nil {
+			break
+		}
+		return fmt.Sprintf("Modified:%v@%v", w.New.Key, w.New.Revision)
+	case WatchDeleted:
+		if w.Old == nil {
+			break
+		}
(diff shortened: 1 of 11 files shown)


caseydavenport (approved):
Looks like a great change! Easy to follow and LGTM

fasaxc:
Fixes #9622 by detecting "too new" errors and triggering a resync.

PR #9595, opened 2024-12-13
[release tool] refactor pinnedversion

radTuti:
## Description

This refactors the pinnedversion package to allow for different pinnedversion per product.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/internal/pinnedversion/pinnedversion.go
+++ b/release/internal/pinnedversion/pinnedversion.go
@@ -31,10 +31,11 @@ import (
 	"github.com/projectcalico/calico/release/internal/registry"
 	"github.com/projectcalico/calico/release/internal/utils"
 	"github.com/projectcalico/calico/release/internal/version"
+	"github.com/projectcalico/calico/release/pkg/manager/calico"
 )
 
-//go:embed templates/calico-version.yaml.gotmpl
-var calicoVersionTemplateData string
+//go:embed templates/calico-versions.yaml.gotmpl
+var calicoTemplate string
 
 const (
 	pinnedVersionFileName      = "pinned-version.yaml"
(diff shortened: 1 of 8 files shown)


release/pkg/manager/calico/manager.go
```diff
@@ -203,6 +202,7 @@ type CalicoManager struct {
 	// image scanning configuration.
 	imageScanning       bool
 	imageScanningConfig imagescanner.Config
+	imageComponents map[string]registry.Component
```
caseydavenport, line 205: formatting looks off here - goimports?

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -47,167 +48,222 @@ var noImageComponents = []string{
...
 
-	// Registry is the registry for Tigera operator
+type OperatorConfig struct {
+	Dir      string
+	Branch   string
 	Registry string
-
-	// Image is the image for Tigera operator
-	Image string
+	Image    string
```
caseydavenport, line 60: I think it was good to remove the other comments. Image is a tricky one - might be worth a comment explaining the expected format.

e.g., does this take `tigera/operator`? Or does it take `quay.io/tigera/operator`? Or does it take `tigrera/opreator:tag`? etc.

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -47,167 +48,222 @@ var noImageComponents = []string{
...
+	Components     map[string]registry.Component `yaml:"components"`
 }
 
-// PinnedVersionData represents the data needed to generate the pinned version file from the template.
-type PinnedVersionData struct {
-	// ReleaseName is the name of the release.
-	ReleaseName string
+func NewCalicoVersionData(productVersion, operatorVersion string) *CalicoPinnedVersionData {
+	return &CalicoPinnedVersionData{
+		calico:   version.New(productVersion),
```
caseydavenport, line 90: It seems a little odd to me to parse these into `Version` structs, and then throughout the `CalicoPinnedVersionData` implementation we exclusively call `FormattedString` to convert back into a string. Is there a reason?

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -47,167 +48,222 @@ var noImageComponents = []string{
...
+	return fmt.Sprintf("https://%s.%s", d.ReleaseName, d.BaseDomain)
+}
 
 func pinnedVersionFilePath(outputDir string) string {
 	return filepath.Join(outputDir, pinnedVersionFileName)
 }
 
-func operatorComponentsFilePath(outputDir string) string {
-	return filepath.Join(outputDir, operatorComponentsFileName)
+type CalicoPinnedVersions struct {
```
caseydavenport, line 139: Confusing to have structs named `CalicoPinnedVersions` and `CalicoPinnedVersionData` in the same file without any comments explaining the separation - what is the intended scope of these two types?

release/internal/pinnedversion/pinnedversion.go
caseydavenport: I'm actually not sure what the purpose of this structure - `CalicoPinnedVersionData` - is.

I originally thought it was going to implement the `PinnedVersions` interface, but looks like that is actually done by `CalicoPinnedVersions` instead.

This struct seems to effectively take two strings and then provide methods to access those strings, which seems like it might be overcomplicated rather than just passing the strings around where needed. Unless there's an interface that this is intended to satsify with an alternative implementation?

Basically, I think the code is missing comments justifying why this construct should exist.

caseydavenport (changes requested):
A few minor comments / questions

release/pkg/manager/calico/manager.go
radTuti: i missed adding it after i ran `make fix-changed`

release/internal/pinnedversion/pinnedversion.go
radTuti: I have moved this struct to `version` package. It implements `version.Data` interface

PR #9599, opened 2024-12-13
Fix List() calls that didn't fill in revision

fasaxc:
## Description

The watchercache requires the KVList to contain the collection resource version for correctness.  Without it, it can miss updates from the datastore.

- Fix the resources that were missing the revision entirely.
- Retire code in the Kubernetes datastore driver that would turn `List()` calls into `Get()` calls when only one resource was needed.  This code was incorrect because `Get()` doesn't return the "collection resource version", which is what a `List()` call _should_ return (and must return for a subsequent `Watch()` to function correctly).
- Instead of `Get()` use a `List()` with a `FieldSelector` that matches on the name of the resource.  This returns the same resource but it wraps it in a List object that carries the needed collection version.
- As a special case, maintain the Get() logic for `WorkloadEndpoint` behind an opt-in `Context` flag (with a nice long name to make users of the hacky flag feel bad :-P).  The CNI plugin relies on that logic to avoid needing `list` RBAC permissions for Pods.

## Related issues/PRs

CORE-10904

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that libcalico-go would not always fill in the revision when listing certain resources (or single instances of certain resources).  This could result in missed watch events in components such as Typha.
```

## Reminder for the reviewer

Other optional labels:


caseydavenport (approved):

caseydavenport (approved):

cni-plugin/pkg/plugin/plugin.go
```diff
@@ -260,7 +261,21 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
 	}
 
 	// Check if there's an existing endpoint by listing the existing endpoints based on the WEP name prefix.
-	endpoints, err := calicoClient.WorkloadEndpoints().List(ctx, options.ListOptions{Name: wepPrefix, Namespace: wepIDs.Namespace, Prefix: true})
+
+	// We know that, in KDD, even though there may be >1 endpoint, we're only
+	// looking up one backing Pod.  Send it a hint that we really want it to
+	// do a Get instead of a list.  CNI plugin only has RBAC permissions to do
+	// a get on Pod resources.  The default used to be to do a Get if possible,
```
caseydavenport, line 268: >  CNI plugin only has RBAC permissions

Any reason to not just switch to a List() with the field selector specified, and give the CNI plugin the necessary permissions?

caseydavenport (approved):

cni-plugin/pkg/plugin/plugin.go
fasaxc: Mainly just thinking that it was quite a big loosening to give CNI plugin list permissions when it could make do with get before.  That, and the fact that I'd need to adjust manifests and operator to loosen permissions, which felt like extra work to go in the wrong direction.

caseydavenport: I suppose I just don't see a huge security distinction between "Get any pod in the cluster" vs "List any pod in the cluster" 

But I'm happy to leave it as is.

Issue #9372, opened 2024-10-21
TCP timeout in idle session on Calico eBPF.


xander-sh:
## Expected Behavior
* close idle TCP session, or network socket.  

## Current Behavior
We noticed that if there is a long period of inactivity in packet exchange within a TCP session, newly sent packets (also within this TCP session) do not reach the recipient. The packets are visible as outgoing on the source side but are absent on the other side. This leads to packet retransmissions (TCP retransmissions) and a prolonged period of session closure from the packet source (RTO). Meanwhile, the socket through which this session was established remains open. 

Additionally, the debug logs of calico-node show the following messages:  
_**Updating/creating an entry in calico contrack.**_
```bash
2024-10-08T16:46:11.803233776+03:00 stdout F 2024-10-08 13:46:11.803 [DEBUG][45] felix/scanner.go 103: Examining conntrack entry entry=Entry{Type:0, Created:42928133306698124, LastSeen:42928133353918870, Flags: <none> Data: {A2B:{Bytes:0 Packets:0 Seqno:2327455211 SynSeen:true AckSeen:true FinSeen:false RstSeen:false Approved:true Opener:true Ifindex:23} B2A:{Bytes:0 Packets:0 Seqno:257842865 SynSeen:true AckSeen:true FinSeen:false RstSeen:false Approved:true Opener:false Ifindex:11} OrigDst:0.0.0.0 OrigSrc:0.0.0.0 OrigPort:0 OrigSPort:0 TunIP:0.0.0.0}} key=ConntrackKey{proto=6 10.222.41.81:37738 <-> 10.222.2.82:50051}
```

_**Delete map records from calico cntrack**_
```bash
[...]

PR #9607, opened 2024-12-17
[BPF] make conntrack timeouts configurable

tomastigera:
BPFConntrackTimers overides the default values for the specified conntrack timer if
    set. It is a key-value make, where each value can be either a duration or `auto` to
    pick the value from a Linux conntrack timeout.
    
    Possible values for the keys are: CreationGracePeriod, TCPPreEstablished,
    TCPEstablished, TCPFinsSeen, TCPResetSeen, UDPLastSeen, GenericIPLastSeen,
    ICMPLastSeen.
    
    Unset or incorrect values are replaced by the default values with a warning log for
    incorrect values.
    
    Current auto mappings:
    
    TCPPreEstablished: nf_conntrack_tcp_timeout_syn_sent
    TCPEstablished:    nf_conntrack_tcp_timeout_established
    TCPFinsSeen:       nf_conntrack_tcp_timeout_time_wait
    GenericIPLastSeen: nf_conntrack_generic_timeout
    ICMPLastSeen:      nf_conntrack_icmp_timeout
    
    If there is no mapping, 'auto' is replaced by the default value.
    
    [Default:
         CreationGracePeriod: 10s
         TCPPreEstablished:   20s
         TCPEstablished:      1h
         TCPFinsSeen:         auto (30s is default)
         TCPResetSeen:        40s
         UDPLastSeen:         60s
         GenericIPLastSeen:   10m
         ICMPLastSeen:        5s
    ]

refs https://github.com/projectcalico/calico/issues/9372

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: conntrack timers are configurable, closed TCP connections are closed now by default after /proc/sys/net/netfilter/nf_conntrack_tcp_timeout_time_wait seconds (120 by default) up from 30s.
```

## Reminder for the reviewer

Other optional labels:


sridhartigera (approved):
LGTM

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -611,6 +611,39 @@ type FelixConfigurationSpec struct {
...
+	// taken from /proc/sys/net/netfilter/<name> files.
+	//
+	// Unset or incorrect values are replaced by the default values with a warning log for
+	// incorrect values.
+	//
+	// [Default:
+	//	CreationGracePeriod: 10s
+	//	TCPPreEstablished:   20s
+	//	TCPEstablished:      1h
+	//	TCPFinsSeen:         nf_conntrack_tcp_timeout_time_wait
```
fasaxc, line 638: Not sure about allowing arbitrary sysctl lookups; feels a bit too powerful (and could allow user to trick felix into reading something that we don't want)?  Is there an obvious mapping from sysctls to our values so that we could just do `auto` and that means "use sysctl from the look-up-table"?

Failing that, we should at least have an explicit allow-list of sysctls.

tomastigera: yes, `auto` sounds good, I can hardcode the auto values :+1:

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -611,6 +611,40 @@ type FelixConfigurationSpec struct {
...
+	//
+	// Current auto mappings:
+	//
+	// TCPPreEstablished: nf_conntrack_tcp_timeout_syn_sent
+	// TCPEstablished:    nf_conntrack_tcp_timeout_established
+	// TCPFinsSeen:       nf_conntrack_tcp_timeout_time_wait
+	// GenericIPLastSeen: nf_conntrack_generic_timeout
+	// ICMPLastSeen:      nf_conntrack_icmp_timeout
+	//
+	// If there is no mapping, 'auto' is replaced by the default value.
```
caseydavenport, line 633: Now that we have defined these as pointers I don't think we need a special "auto" field - `nil` means "use the default value", right? Or am I misunderstanding the semantics of "auto"?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -611,6 +611,40 @@ type FelixConfigurationSpec struct {
 	// [Default: Auto]
 	BPFConntrackCleanupMode *BPFConntrackMode `json:"bpfConntrackMode,omitempty" validate:"omitempty,oneof=Auto Userspace BPFProgram"`
 
+	// BPFConntrackTimers overides the default values for the specified conntrack timer if
+	// set. It is a key-value make, where each value can be either a duration or `auto` to
+	// pick the value from a Linux conntrack timeout.
+	//
+	// Possible values for the keys are: CreationGracePeriod, TCPPreEstablished,
```
caseydavenport, line 618:
```suggestion
	// Configurable timers are: CreationGracePeriod, TCPPreEstablished,
```

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -611,6 +611,40 @@ type FelixConfigurationSpec struct {
 	// [Default: Auto]
 	BPFConntrackCleanupMode *BPFConntrackMode `json:"bpfConntrackMode,omitempty" validate:"omitempty,oneof=Auto Userspace BPFProgram"`
 
+	// BPFConntrackTimers overides the default values for the specified conntrack timer if
+	// set. It is a key-value make, where each value can be either a duration or `auto` to
+	// pick the value from a Linux conntrack timeout.
```
caseydavenport, line 616:
```suggestion
	// BPFConntrackTimers overides the default values for the specified conntrack timer if
	// set. Each value can optionally be set to a duration. If unset, default values will be used.
```

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -611,6 +611,40 @@ type FelixConfigurationSpec struct {
...
+	//
+	// Possible values for the keys are: CreationGracePeriod, TCPPreEstablished,
+	// TCPEstablished, TCPFinsSeen, TCPResetSeen, UDPLastSeen, GenericIPLastSeen,
+	// ICMPLastSeen.
+	//
+	// Unset or incorrect values are replaced by the default values with a warning log for
```
caseydavenport, line 622: What's an incorrect value? Is there some value that can byapss our validation?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -611,6 +611,40 @@ type FelixConfigurationSpec struct {
...
+	// incorrect values.
+	//
+	// Current auto mappings:
+	//
+	// TCPPreEstablished: nf_conntrack_tcp_timeout_syn_sent
```
caseydavenport, line 627: Would it make sense to call this `TCPSynSent` to align with the underlying timer?

caseydavenport: And same in general. I'm not an expert on these, so take with a grain of salt, but seems more intuitive if we don't need to refer to this mapping to know what timers are being set.

caseydavenport (commented):
@tomastigera I'm just reviewing the API bits (others have reviewed the implementation IIUC). 

Looking pretty good, but had a few questions / suggestions.

api/pkg/apis/projectcalico/v3/felixconfig.go
tomastigera: Yes we need auto to distinguish between "use the default" aka nil/unset and use the value from linux if available aka auto

api/pkg/apis/projectcalico/v3/felixconfig.go
tomastigera: Old comment, not valid anymore :+1:

api/pkg/apis/projectcalico/v3/felixconfig.go
tomastigera: I would make sense to rename the TCP timer, but the linux timers are not quite 1:1 to our timers so I thought it would be better if we spell out which timer do match. I originally allowed users to pick their own mapping but we decided that it is not the best and we hardcoded the mappings.

api/pkg/apis/projectcalico/v3/felixconfig.go
caseydavenport: Got it - in that case, I think the description here isn't clear enough. That behavior wasn't obvious to me. I'll add a suggested edit.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -947,6 +981,20 @@ type ProtoPort struct {
 	Net string `json:"net,omitempty"`
 }
 
+// +kubebuilder:validation:Pattern=`^(([0-9]*(\.[0-9]*)?(ms|s|h|m|us)+)+|auto)$`
+type BPFConntrackTimeout string
+
+type BPFConntrackTimeouts struct {
+	CreationGracePeriod *BPFConntrackTimeout `json:"creationGracePeriod,omitempty"`
```
caseydavenport, line 988: Each of these fields should have a comment explaining what it does and what its default value is. That way our docs generation will build good human readable docs. I think as you've structured it now, this is all front-loaded on the `BPFConntrackTimeouts` structure instead of on the fields.

caseydavenport, line 988:
```suggestion
        // CreationGracePeriod configures . . . . If set to "Auto" the default value is learned from Linux. If nil, Calico will use its own default value. [Default: 10s]
	CreationGracePeriod *BPFConntrackTimeout `json:"creationGracePeriod,omitempty"`
```

e.g., this field "CreationGracePeriod" isn't fully described in the BPFConntrackTimeouts so it's not obvious what it does from reading this.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -611,6 +611,40 @@ type FelixConfigurationSpec struct {
...
+	//
+	// TCPSynSent: nf_conntrack_tcp_timeout_syn_sent
+	// TCPEstablished:    nf_conntrack_tcp_timeout_established
+	// TCPFinsSeen:       nf_conntrack_tcp_timeout_time_wait
+	// GenericIPLastSeen: nf_conntrack_generic_timeout
+	// ICMPLastSeen:      nf_conntrack_icmp_timeout
+	//
+	// If there is no mapping, 'auto' is replaced by the default value.
+	//
+	// [Default:
```
caseydavenport, line 635: We don't need to specify a default value for structs - we should put defaults on the fields within.

api/pkg/apis/projectcalico/v3/felixconfig.go
caseydavenport: Actually, CreationGracePeriod doesn't support "auto" does it? i.e., based on the top level comment, there is no linux mapping for this field?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -947,6 +981,20 @@ type ProtoPort struct {
 	Net string `json:"net,omitempty"`
 }
 
+// +kubebuilder:validation:Pattern=`^(([0-9]*(\.[0-9]*)?(ms|s|h|m|us)+)+|auto)$`
+type BPFConntrackTimeout string
+
+type BPFConntrackTimeouts struct {
+	CreationGracePeriod *BPFConntrackTimeout `json:"creationGracePeriod,omitempty"`
+	TCPSynSent          *BPFConntrackTimeout `json:"tcpSynSent,omitempty"`
```
caseydavenport, line 989:
```suggestion
        // TCPSynSent configures . . . .. If set to "Auto", the value from nf_conntrack_tcp_timeout_syn_sent is used. If nil, Calico uses its own default value. [Default: 20s].
	TCPSynSent          *BPFConntrackTimeout `json:"tcpSynSent,omitempty"`
```

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -611,6 +611,40 @@ type FelixConfigurationSpec struct {
...
+	//
+	// Current auto mappings:
+	//
+	// TCPSynSent: nf_conntrack_tcp_timeout_syn_sent
+	// TCPEstablished:    nf_conntrack_tcp_timeout_established
+	// TCPFinsSeen:       nf_conntrack_tcp_timeout_time_wait
+	// GenericIPLastSeen: nf_conntrack_generic_timeout
+	// ICMPLastSeen:      nf_conntrack_icmp_timeout
+	//
+	// If there is no mapping, 'auto' is replaced by the default value.
```
caseydavenport, line 633:
```suggestion
```

I think all of this is better expressed as field-level comments, as each field has its own behavior. See my suggested format for the per-field comments below.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -611,6 +611,40 @@ type FelixConfigurationSpec struct {
...
+	// [Default:
+	//	CreationGracePeriod: 10s
+	//	TCPSynSent:   20s
+	//	TCPEstablished:      1h
+	//	TCPFinsSeen:         auto (30s is default)
+	//	TCPResetSeen:        40s
+	//	UDPLastSeen:         60s
+	//	GenericIPLastSeen:   10m
+	//	ICMPLastSeen:        5s
+	// ]
```
caseydavenport, line 644:
```suggestion
```

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -947,6 +981,20 @@ type ProtoPort struct {
 	Net string `json:"net,omitempty"`
 }
 
+// +kubebuilder:validation:Pattern=`^(([0-9]*(\.[0-9]*)?(ms|s|h|m|us)+)+|auto)$`
```
caseydavenport, line 984:
```suggestion
// +kubebuilder:validation:Pattern=`^(([0-9]*(\.[0-9]*)?(ms|s|h|m|us)+)+|Auto)$`
```

I think this should be "Auto" based on k8s API conventions for constant values: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

caseydavenport (approved):

PR #9615, opened 2024-12-18
Use policy name used during create

MichalFupso:
## Description

Update how polices in the default tier are managed. Policy created without the default. tier prefix can no longer be managed by the name with default. tier prefix added, if policy was originally created with the default. prefix, it's also no longer possible to manage the policy without the prefix. 

This ensures that the Calico API server always returns the policy name the same as the one the policy was created with. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Policies part of the default tier can be managed only with the original name they were created with. Policies are no longer able to be managed interchangeably with or without the default. tier prefix
```

## Reminder for the reviewer

Other optional labels:


apiserver/cmd/apiserver/server/tier_upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+
+func removeDefaultTierPrefix(name string) string {
+	if strings.HasPrefix(name, "default.") {
+		name = strings.TrimPrefix(name, "default.")
+	}
+	return name
+}
+
+func updateNeeded(k8sClientset *kubernetes.Clientset) bool {
+	tierMigration, err := k8sClientset.CoreV1().ConfigMaps("calico-apiserver").Get(context.Background(), "api-server", metav1.GetOptions{})
```
caseydavenport, line 75: Strictly speaking we may be running in a different namespace in some environments - I believe you can learn what namespace you are in based on a projected file mounted into the container at `/var/run/secrets/kubernetes.io/serviceaccount/namespace` though. Might already be loaded by the go-client? I can't remember.

libcalico-go/lib/backend/etcdv3/etcdv3.go
```diff
@@ -568,3 +611,71 @@ func parseRevision(revs string) (int64, error) {
...
+func defaultPolicyName(d *model.KVPair) error {
+	if _, ok := d.Value.(*apiv3.NetworkPolicy); ok {
+		value := d.Value.(*apiv3.NetworkPolicy)
+
+		annotations, err := storePolicyName(value.Name, value.Annotations)
+		if err != nil {
+			return err
+		}
+
+		value.Annotations = annotations
```
caseydavenport, line 641:
```suggestion
                // First, capture the policy name that was used on the v3 API and store it as an annotation. 
                // This allows us to recreate the original object in List() and Watch() calls correctly, returning the object as expected by the client. 
		annotations, err := storePolicyName(value.Name, value.Annotations)
		if err != nil {
			return err
		}
		value.Annotations = annotations
```

Here's an example of the type of comment I'm looking for.

libcalico-go/lib/backend/etcdv3/etcdv3.go
```diff
@@ -568,3 +611,71 @@ func parseRevision(revs string) (int64, error) {
...
+
+		polName, err := names.BackendTieredPolicyName(value.Name, value.Spec.Tier)
+		if err != nil {
+			return err
+		}
+		value.Name = polName
```
caseydavenport, line 647:
```suggestion
                // Now that we've captured the original name, canonicalize the name for storage, 
                // adding the tier prefix to ensure policies with the same name in different tiers do not conflict.
		polName, err := names.BackendTieredPolicyName(value.Name, value.Spec.Tier)
		if err != nil {
			return err
		}
		value.Name = polName
```

libcalico-go/lib/backend/k8s/k8s_test.go
```diff
@@ -383,7 +383,7 @@ var _ = testutils.E2eDatastoreDescribe("Test UIDs and owner references", testuti
 		npUIDv1, err := conversion.ConvertUID(npUID)
 		Expect(err).NotTo(HaveOccurred())
 
-		name := "test-owner-ref-policy"
+		name := "default.test-owner-ref-policy"
```
caseydavenport, line 386: If we didn't change this, wouldn't the test still pass? I'd expect the backend to store the name as an annotation, and canonicalize the name used for the backing CRD?

libcalico-go/lib/backend/k8s/resources/resources.go
```diff
@@ -138,7 +140,6 @@ func ConvertCalicoResourceToK8sResource(resIn Resource) (Resource, error) {
 	// the metadata annotation.
 	romCopy := &metav1.ObjectMeta{}
 	rom.(*metav1.ObjectMeta).DeepCopyInto(romCopy)
-	romCopy.Name = ""
```
caseydavenport, line 141: I think we probably want to only store the name in the annotation for NetworkPolicy / GlobalNetworkPolicy? i.e., objects that belong to a Tier.

libcalico-go/lib/backend/k8s/resources/resources.go
```diff
@@ -175,6 +176,17 @@ func ConvertCalicoResourceToK8sResource(resIn Resource) (Resource, error) {
 	// Name, Namespace, ResourceVersion, UID, and Annotations (built above).
 	meta := &metav1.ObjectMeta{}
 	meta.Name = rom.GetName()
+	switch resIn.GetObjectKind().GroupVersionKind().Kind {
+	// For NetworkPolicy and GlobalNetworkPolicy, we need to prefix the name with the tier name.
```
caseydavenport, line 180:
```suggestion
	// For NetworkPolicy and GlobalNetworkPolicy, we need to prefix the name with the tier name.
	// This ensures two policies with the same name, but in different tiers, do not resolve to the same backing object.
```

caseydavenport: A bit of a nit, but could you move the switch statement to the end of the metadata populating section? It's breaking up a bunch of similar lines and makes the code a tiny bit harder to follow. 

e.g.,

```go
meta.Name = rom.GetName()
meta.Namespace = rom.GetNamespace()
meta.ResourceVersion = rom.GetResourceVersion()

switch resIn.GetObjectKind().GroupversionKind().Kind() {
 . . .
```

libcalico-go/lib/backend/k8s/resources/resources.go
```diff
@@ -293,7 +305,9 @@ func ConvertK8sResourceToCalicoResource(res Resource) error {
 
 	// Manually write in the data not stored in the annotations: Name, Namespace, ResourceVersion,
 	// so that they do not get overwritten.
-	meta.Name = rom.GetName()
+	if meta.Name == "" {
```
caseydavenport, line 308:
```suggestion
	if meta.Name == "" {
	  // We store the original v3 Name in our annotation when writing NetworkPolicy and GNP objects. If that's present, use it.
	  // Otherwise, fill in the name from the underlying CRD.
```

libcalico-go/lib/backend/model/keys.go
```diff
@@ -23,13 +23,19 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 
+	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	"github.com/projectcalico/calico/libcalico-go/lib/json"
 	"github.com/projectcalico/calico/libcalico-go/lib/namespace"
 	"github.com/projectcalico/calico/libcalico-go/lib/net"
 )
 
+const (
+	metadataAnnotation = "projectcalico.org/metadata"
```
caseydavenport, line 36: We have this defined in a few places now - should we put it in a common location instead?

libcalico-go/lib/clientv3/globalnetworkpolicy_e2e_test.go
```diff
@@ -155,7 +156,7 @@ var _ = testutils.E2eDatastoreDescribe("GlobalNetworkPolicy tests", testutils.Da
 			Expect(outError.Error()).To(Equal("resource already exists: GlobalNetworkPolicy(" + tieredGNPName(name1, tier) + ")"))
 
 			By("Getting GlobalNetworkPolicy (name1) and comparing the output against spec1")
-			res, outError := c.GlobalNetworkPolicies().Get(ctx, name1, options.GetOptions{})
+			res, outError := c.GlobalNetworkPolicies().Get(ctx, tieredGNPName(name1, tier), options.GetOptions{})
```
caseydavenport, line 159: I'm a bit surprised to see all of these test changes - I would have expected this change to be compatible with our existing v3 client tests?

caseydavenport (changes requested):
Mostly my comments revolve around... adding more comments :D 

This behavior is a bit weird and needs to be very cleanly documented in the code.

The other thing that I don't understand are the test changes - seems like we're changing a bunch of tests that use the name to use tier.name instead? That seems fishy to me. I would have expected most tests to stay the same, but to introduce a new set of tests that verify

1. Policies can be created with either name format.
2. Policies are returned using the same name format they were created with (e.g., via List and Watch operations)

apiserver/cmd/apiserver/server/tier_upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+
+	client "github.com/projectcalico/calico/libcalico-go/lib/clientv3"
+	"github.com/projectcalico/calico/libcalico-go/lib/options"
+	"github.com/projectcalico/calico/libcalico-go/lib/winutils"
+)
+
+func migratePolicyNames() error {
```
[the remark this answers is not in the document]
caseydavenport: I also wonder if we need a way to explicitly turn this off - I guess the user can create the configmap and it will do it?

Have you tried this on a mock cluster with say, 10k policy objects?

apiserver/cmd/apiserver/server/tier_upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+	tierMigration, err := k8sClientset.CoreV1().ConfigMaps("calico-apiserver").Get(context.Background(), "api-server", metav1.GetOptions{})
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			configMap := v1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "tier-migration",
+					Namespace: "calico-apiserver",
+				},
+				Data: map[string]string{
+					"tierMigration": "true",
```
caseydavenport, line 84:
```suggestion
					"policyNameMigration": "true",
```

caseydavenport: We should probably define a constant for this.

apiserver/cmd/apiserver/server/tier_upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+				},
+			}
+			_, err = k8sClientset.CoreV1().ConfigMaps("calico-apiserver").Create(context.Background(), &configMap, metav1.CreateOptions{})
+			if err != nil {
+				klog.Errorf("Failed to create api-server configmap: %s", err)
+				return false
```
caseydavenport, line 90: Why does a failure to create the configmap mean we don't need to update the tier names?

apiserver/cmd/apiserver/server/tier_upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+			return true
+		} else {
+			klog.Errorf("Failed to get api-server configmap: %s", err)
+			return false
+		}
+	}
+	if tierMigration.Data["tierMigration"] == "false" {
+		return false
+	}
+	return true
```
caseydavenport, line 102:
```suggestion
	return tierMigration.Data["tierMigration"] != "false"
```

apiserver/cmd/apiserver/server/tier_upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+	for _, policy := range globalNetworkPolicies.Items {
+		policy.Name = removeDefaultTierPrefix(policy.Name)
+		_, err = calicoClient.GlobalNetworkPolicies().Update(context.Background(), &policy, options.SetOptions{})
+		if err != nil {
+			return err
+		}
+	}
+
```
caseydavenport, line 63: Don't we need to update the ConfigMap after a successful tier migration has occurred so we don't run through this again?

apiserver/cmd/apiserver/server/tier_upgrade.go
caseydavenport: Worth verifying that we have the necessary RBAC in our manifests / operator to get, create, and update this configmap.

caseydavenport (changes requested):

PR #9616, opened 2024-12-18
[release-tool] refactor update branch command and manager

radTuti:
## Description

This update the `branch` subcommands and branch manager.

As an extension, version package is also updated to support EPs and allow avoiding conflicts in closed source.

## Related issues/PRs

- split from #9563 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/internal/version/version.go
+++ b/release/internal/version/version.go
@@ -100,15 +100,19 @@ func (v *Version) Milestone() string {
 	return fmt.Sprintf("%s v%d.%d.%d", utils.DisplayProductName(), ver.Major(), ver.Minor(), ver.Patch())
 }
 
-// Stream returns the "release stream" of the version
-// typically it is the major and minor version without the patch version.
+// Stream returns the "release stream" of the version.
+// Typically it is the major and minor version without the patch version.
 // For example, for version "v3.15.0", the stream is "v3.15".
-// For EP 1, appends "-1" to the stream.
+//
+// Early preview versions are handled differently.
+// For example, for version "v3.15.0-1.0", the stream is "v3.15-1".
+// For version "v3.15.0-2.0", the stream is "v3.15" (same as v3.15.1+).
 func (v *Version) Stream() string {
-	ver := semver.MustParse(string(*v))
+	ver := v.Semver()
+	ep, epVer := IsEarlyPreviewVersion(ver)
 	stream := fmt.Sprintf("v%d.%d", ver.Major(), ver.Minor())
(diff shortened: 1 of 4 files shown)


release/internal/version/version.go
```diff
@@ -100,17 +100,36 @@ func (v *Version) Milestone() string {
...
-// Stream returns the "release stream" of the version, i.e., the major and minor version without the patch version.
+// Stream returns the "release stream" of the version
+// typically it is the major and minor version without the patch version.
+// For example, for version "v3.15.0", the stream is "v3.15".
+// For EP 1, appends "-1" to the stream.
 func (v *Version) Stream() string {
 	ver := semver.MustParse(string(*v))
-	return fmt.Sprintf("v%d.%d", ver.Major(), ver.Minor())
+	stream := fmt.Sprintf("v%d.%d", ver.Major(), ver.Minor())
+	if ver.Prerelease() != "" && strings.HasPrefix(ver.Prerelease(), "1.") {
```
caseydavenport, line 110: Should we have an explicit flag passed in to say when something is an EP release? 

My concern is that this approach (using "secret" knowledge about the syntax of a version string) is brittle and not super intuitive.

At a minimum, we should have an `IsEarlyPreviewVersion(v string) bool` function that we can reuse, but still think a flag would be ideal.

release/internal/version/version.go
```diff
@@ -123,30 +142,38 @@ func GitVersion() Version {
...
 	devTagSuffix = strings.TrimPrefix(devTagSuffix, "-")
-	re := regexp.MustCompile(fmt.Sprintf(`^v\d+\.\d+\.\d+-%s-\d+-g[0-9a-f]{12}(-dirty)?$`, devTagSuffix))
+	re := regexp.MustCompile(fmt.Sprintf(`^v\d+\.\d+\.\d+(-\d+\.\d+)?-%s-\d+-g[0-9a-f]{12}(-dirty)?$`, devTagSuffix))
 	return re.MatchString(string(v))
 }
 
 // DetermineReleaseVersion uses historical clues to figure out the next semver
 // release number to use for this release based on the current git revision.
-//
-// OSS Calico uses the following rules:
```
caseydavenport, line 136: I think there is value in separating the version rules that OSS uses vs enterprise?

radTuti: At the moment, the way the rules are structured apply with OSS and Enterprise although Enterprise is only the first scenario since it creates a new dev tag after every release.

caseydavenport (approved):

Issue #7841, opened 2023-07-05
calico-kube-controller 3.26 high cpu usage frequent ipam syncs


juliantaylor:
We updated calico in one of our larger clusters from 3.24.5 to 3.26.1 and since then the kube-controller has a comparably high cpu usage and seems to perform many ipam syncs.
3.24 in an cluster of about 200 nodes and 10000 pods uses about 200m cpu while 3.26.1 uses a full cpu.

The underlying cause appears to be a change probably added in 3.24.6 which made the allocationIsValid function actually use the cache which reduced the cache sync time from a few minutes to a couple seconds.
So essentially a problem was fixed but maybe the high cpu usage is a fixable issue or the sync frequency is too high.

a pprofile showed it spends lots of time in defaultWorkloadEndpointConverter in checkAllocations

```
         0     0%     0%    159.53s 67.54%  github.com/projectcalico/calico/kube-controllers/pkg/controllers/node.(*ipamController).acceptScheduleRequests
     0.64s  0.27%  0.27%    153.88s 65.15%  github.com/projectcalico/calico/kube-controllers/pkg/controllers/node.(*ipamController).checkAllocations
         0     0%  0.27%    153.88s 65.15%  github.com/projectcalico/calico/kube-controllers/pkg/controllers/node.(*ipamController).syncIPAM
     1.44s  0.61%  0.88%    129.67s 54.90%  github.com/projectcalico/calico/kube-controllers/pkg/controllers/node.(*ipamController).allocationIsValid
         0     0%  0.88%     99.52s 42.13%  github.com/projectcalico/calico/libcalico-go/lib/backend/k8s/conversion.defaultWorkloadEndpointConverter.PodToWorkloadEndpoints
     8.49s  3.59%  4.48%     99.52s 42.13%  github.com/projectcalico/calico/libcalico-go/lib/backend/k8s/conversion.defaultWorkloadEndpointConverter.podToDefaultWorkloadEndpoint

```

## Current Behavior

```
grep -E "Synchronizing IPAM|sync com|Triggering sync after batch of updates "
calico-kube-controllers-7cc79d76fd-n5526 calico-kube-controllers 2023-07-05 15:46:54.246 [DEBUG][1] ipam.go 968: Synchronizing IPAM data
calico-kube-controllers-7cc79d76fd-n5526 calico-kube-controllers 2023-07-05 15:46:57.021 [DEBUG][1] ipam.go 1009: IPAM sync completed
[...]

PR #9654, opened 2024-12-30
Don't check every IPAM allocation on every sync.

caseydavenport:
## Description

Fixes https://github.com/projectcalico/calico/issues/7841

The IPAM GC controller currently iterates every IP address allocation in the cluster every time an IP address is assigned or released. This can result in excessive CPU usage on larger / more active clusters. 

Instead, this PR changes the behavior such that we limit the allocations we check to:

- Those on nodes whose IPAM state has changed since the last sync.
- Those on nodes who have had a Pod Deletion Event since the last sync.

The idea is that this set should be much smaller than every node, and will allow us to incrementally scan the cluster instead of checking the entire cluster on every syncIPAM() call. 

We will have a periodic (by default every 5m) sync that will check every node, as a backstop, and it's worth noting that this change on its own means we may be relying on the periodic sync slightly more often in order to determine when IPAM allocations have leaked, as it is possible leaks can occur on nodes without the controller receiving the requisite notifications to notice it.

We might be able to enhance this further by tracking when individual allocations are dirty (and, for example, only clearing them after the grace period proves they are OK or leaked). But I think that's a larger change and this one should do the trick I hope. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Improve CPU efficiency of IP address garbage collection controller
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -138,6 +138,9 @@ func NewIPAMController(cfg config.NodeControllerConfig, c client.Interface, cs k
 		podLister:  v1lister.NewPodLister(pi),
 		nodeLister: v1lister.NewNodeLister(ni),
 
+		nodeDeletionChan: make(chan *v1.Node, batchUpdateSize),
+		podDeletionChan:  make(chan *v1.Pod, batchUpdateSize),
+
 		// Buffered channels for potentially bursty channels.
 		syncerUpdates: make(chan interface{}, batchUpdateSize),
 
@@ -211,6 +214,10 @@ type ipamController struct {
 	// Cache datastoreReady to avoid too much API queries.
 	datastoreReady bool
 
+	// channels for indicating that Kubernetes nodes or pods have been deleted.
+	nodeDeletionChan chan *v1.Node
+	podDeletionChan  chan *v1.Pod
+
 	// For unit testing purposes.
 	pauseRequestChannel chan pauseRequest
 }
@@ -243,10 +250,14 @@ func (c *ipamController) onUpdate(update bapi.Update) {
 	}
 }
 
-func (c *ipamController) OnKubernetesNodeDeleted() {
-	// When a Kubernetes node is deleted, trigger a sync.
+func (c *ipamController) OnKubernetesNodeDeleted(node *v1.Node) {
 	log.Debug("Kubernetes node deletion event")
-	kick(c.syncChan)
+	c.nodeDeletionChan <- node
+}
+
+func (c *ipamController) OnKubernetesPodDeleted(pod *v1.Pod) {
+	log.WithField("pod", pod.Name).Debug("Kubernetes pod deletion event")
+	c.podDeletionChan <- pod
 }
 
 // acceptScheduleRequests is the main worker routine of the IPAM controller. It monitors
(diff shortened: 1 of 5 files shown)


caseydavenport:
This isn't perfect - it's possible for example that on larger clusters with lots of churn, this will still end up iterating a LOT of IPAM allocations. We could probably make this even better if we tracked dirty-ness at the allocation level rather than at the node level. Doing this would definitely want us to also be marking allocations as dirty based on Pod updates as well.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -250,6 +275,21 @@ func (c *ipamController) acceptScheduleRequests(stopCh <-chan struct{}) {
...
+			kick(c.syncChan)
+		case pod := <-c.podDeletionChan:
+			// When a pod is deleted, mark its node as dirty and schedule a sync.
+			// TODO: We could be tracking dirtiness at the allocation level instead.
+			log.WithFields(log.Fields{
+				"pod":  pod.Name,
+				"ns":   pod.Namespace,
+				"node": pod.Spec.NodeName,
+			}).Debug("Received pod deletion event")
+			c.allocationState.markDirty(pod.Spec.NodeName)
```
caseydavenport, line 291: Should do batching here as well

caseydavenport:
After writing this PR up, I've been thinking a bit more about this and I think I'd like to adjust the strategy here.

Here are the 4 main responsibilities of this IPAM controller:

1. Periodic resync. This is meant to catch leaks, and so should scan the entire IPAM data set. 
2. Whenever one or more nodes are deleted, we clean up blocks that belonged to the deleted node (by scanning the whole IPAM data set).
3. Whenever an IPAM block is updated, we scan the whole IPAM data set (indexed in a different way than the above) in order to generate IPAM metrics for Prometheus. 
4. When we see that a Node has more blocks than it needs, we clean them up.

For (1), we need to scan the whole IPAM data set. But we can do this relatively infrequently. 

For (2), we don't necessarily need to scan the entire IPAM data set - we could just clean up the known affine blocks. Scanning the whole set has some advantages - we might catch a leak slightly sooner if it was in a borrowed IPAM block, but the full scan should catch that eventually anyway.

For (3), we do need to scan the entire IPAM data set, but we can probably stop doing it so frequently. Right now we do it every batch of updates, but Prometheus is polling anyway so there's really no value in doing this more frequently than the Prometheus poll timer. We could probably change this to be every 15s (and only if something changed) without any real UX problem.

For (4), deleting unused IPAM blocks that are affine to existing nodes can be (and already are) performed without needing to scan the entire IPAM data set. We track empty blocks as updates come in, so we know exactly what the set of empty blocks is on demand. We can do an empty block check every sync without much cause for concern, so no need to change this.

I think I would propose:

1. Scan the full IPAM data set to check for leaks on a periodic resync (i.e., every 5 minutes).
2. Scan the full IPAM data set to check blocks for deleted nodes whenever one or more nodes are deleted (could be further optimized to check for particular nodes, but not necessary IMO).
3. Only perform a metrics scan of the dataset if IPAM data has changed, and no more frequently than every 15s.

fasaxc:
@caseydavenport for (3) do we scan an in-memory cache, or do we refresh the whole list from the API server?  Seems like we could certainly use a cache for that less critical work.

caseydavenport:
@fasaxc we use an in-memory cache for all of the above! So at least we have that going :sweat_smile:

caseydavenport:
The problem with (3)  I just realized is that we rely on the full IPAM scan to determine which IPs are GC candidates / leaks, which then feeds into the metrics.. so metrics will end up being gated by the full IPAM scan interval anyway!

caseydavenport:
My most recent changes keep the "immediate" sync behavior, where we trigger a sync for each block update, each node deletion update, and each pod deletion update (in order to maintain quick metrics updates). What I've done instead is back to the original approach:

- Only iterate over nodes that have changed since the last sync.
- When we receive an update, wait a little bit longer to see if any more updates can be added to the batch so that we don't call syncIPAM so much.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -250,18 +293,58 @@ func (c *ipamController) acceptScheduleRequests(stopCh <-chan struct{}) {
 	for {
 		// Wait until something wakes us up, or we are stopped.
 		select {
+		case <-c.nodeDeletionChan:
```
fasaxc, line 296: Might be good to service both channels during the consolidation window?  (But this is likely a lot better than it was!)

fasaxc (approved):
Couple of thoughts but perhaps we should bank it!

PR #9661, opened 2024-12-31
[release tool] fix hashrelease failure

radTuti:
## Description

Hashrelease was failing because the grep command to check if a hash already exists in server was exiting with 1 and was being propagated up.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/internal/hashreleaseserver/server.go
+++ b/release/internal/hashreleaseserver/server.go
@@ -85,13 +85,10 @@ func remoteReleasesLibraryPath(user string) string {
 	return filepath.Join(RemoteDocsPath(user), "all-releases")
 }
 
-func HasHashrelease(hash string, cfg *Config) (bool, error) {
+func HasHashrelease(hash string, cfg *Config) bool {
 	logrus.WithField("hash", hash).Debug("Checking if hashrelease exists")
 	out, err := runSSHCommand(cfg, fmt.Sprintf("cat %s | grep %s", remoteReleasesLibraryPath(cfg.User), hash))
-	if err == nil {
-		return strings.Contains(out, hash), nil
-	}
-	return false, err
+	return err != nil && strings.Contains(out, hash)
 }
 
 // SetHashreleaseAsLatest sets the hashrelease as the latest for the stream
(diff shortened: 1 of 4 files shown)


sridhartigera (approved):

release/internal/hashreleaseserver/server.go
```diff
@@ -85,13 +85,10 @@ func remoteReleasesLibraryPath(user string) string {
...
 
-func HasHashrelease(hash string, cfg *Config) (bool, error) {
+func HasHashrelease(hash string, cfg *Config) bool {
 	logrus.WithField("hash", hash).Debug("Checking if hashrelease exists")
 	out, err := runSSHCommand(cfg, fmt.Sprintf("cat %s | grep %s", remoteReleasesLibraryPath(cfg.User), hash))
-	if err == nil {
-		return strings.Contains(out, hash), nil
-	}
-	return false, err
+	return err != nil && strings.Contains(out, hash)
```
danudey, line 91: So now `HasHashrelease` returns `true` if we checked for a hash release and found one, and `false` if we didn't find one or if we had an error checking for one?

I see the logic there, but perhaps in addition we should log either that we had an error or that we found an existing hashrelease, just so we know which one happened to aid in debugging later? Something like:

```go
        if err != nil {
            logrus.WithField("hash", hash).Warn("Error checking hashrelease details") // Add error details here maybe
            return false
        }
        if strings.Contains(out, hash) {
            logrus.WithField("hash", hash).Info("Hashrelease already exists")
            return false
        }
	return true
```

danudey (commented):
lgtm

danudey (approved):

release/internal/hashreleaseserver/server.go
caseydavenport: Should definitely be checking the error, and at a minimum logging it out.

caseydavenport: If we encounter an error, shouldn't we assume that the HR exists?

Seems like a better fix for this might be to make the command we run not exit with 1 on success... I think we're asking for future trouble if we're assuming "err != nil" is a success scenario.

radTuti: Right, I have updated the section to log errors and return the error when it is not expected

caseydavenport (approved):

PR #9663, opened 2024-12-31
Refactor: Use interface to configure libbpf global data when loading bpf programs.

sridhartigera:
## Description

We have different set of global data one for Tc programs, one for CTLB and one for conntrack cleanup when loading bpf programs. Configuring each global has a separate API. This PR provides a single API to configure global data for bpf programs.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/bpf.go
+++ b/felix/bpf/bpf.go
@@ -38,7 +38,8 @@ import (
 	"golang.org/x/sys/unix"
 
 	"github.com/projectcalico/calico/felix/bpf/bpfdefs"
-	"github.com/projectcalico/calico/felix/bpf/hook"
+	"github.com/projectcalico/calico/felix/bpf/libbpf"
+	"github.com/projectcalico/calico/felix/bpf/maps"
 	"github.com/projectcalico/calico/felix/bpf/utils"
 	"github.com/projectcalico/calico/felix/environment"
 	"github.com/projectcalico/calico/felix/labelindex"
@@ -2247,7 +2248,7 @@ func PolicyDebugJSONFileName(iface, polDir string, ipFamily proto.IPVersion) str
 	return path.Join(RuntimePolDir, fmt.Sprintf("%s_%s_v%d.json", iface, polDir, ipFamily))
 }
 
-func MapPinDir(typ int, name, iface string, h hook.Hook) string {
+func MapPinDir() string {
 	PinBaseDir := path.Join(bpfdefs.DefaultBPFfsPath, "tc")
 	subDir := "globals"
 	return path.Join(PinBaseDir, subDir)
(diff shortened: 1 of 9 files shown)


felix/bpf/libbpf/libbpf.go
```diff
@@ -492,6 +485,20 @@ func XDPSetGlobals(
 	return err
 }
 
+func SetGlobalData(m *Map, data interface{}) error {
```
tomastigera, line 488:
```suggestion
func SetGlobalData(m *Map, data any) error {
```

felix/bpf/libbpf/libbpf.go
```diff
@@ -492,6 +485,20 @@ func XDPSetGlobals(
...
+	switch v := data.(type) {
+	case *TcGlobalData:
+		return tcSetGlobals(m, v)
```
tomastigera, line 491: Why not to have this function to be a method of each of the globals structs and then define an interface with that  method (if need be) and call that method rather than using _any_ and this switch?

felix/bpf/conntrack/bpf_scanner.go
```diff
@@ -145,17 +145,17 @@ func (s *BPFProgLivenessScanner) ensureBPFExpiryProgram() (*libbpf.Obj, error) {
 				continue
 			}
 
-			err := libbpf.CTCleanupSetGlobals(
+			err := libbpf.SetGlobalData(
```
tomastigera, line 148: Could this be?

```suggestion
			err := &libbpf.CTCleanupGlobalData{...}.Set(m)
```

sridhartigera: Prefer doing it this way. Will update the PR shortly.

felix/bpf/tc/attach.go
```diff
@@ -30,7 +30,8 @@ import (
 	"github.com/projectcalico/calico/felix/bpf/bpfdefs"
 	"github.com/projectcalico/calico/felix/bpf/hook"
 	"github.com/projectcalico/calico/felix/bpf/libbpf"
-	"github.com/projectcalico/calico/felix/bpf/maps"
+
+	//"github.com/projectcalico/calico/felix/bpf/maps"
```
tomastigera, line 34: remove

felix/bpf/tc/attach.go
```diff
@@ -406,8 +373,8 @@ func (ap *AttachPoint) Config() string {
 	return fmt.Sprintf("%+v", ap)
 }
 
-func (ap *AttachPoint) ConfigureProgram(m *libbpf.Map) error {
-	globalData := libbpf.TcGlobalData{
+func (ap *AttachPoint) ConfigureProgram() *libbpf.TcGlobalData {
```
tomastigera, line 376:
```suggestion
func (ap *AttachPoint) Configuration() *libbpf.TcGlobalData {
```

It would make sense to name it like this to allign with XDP, also it just returns configuration

felix/bpf/tc/attach.go
```diff
@@ -481,13 +448,12 @@ func (ap *AttachPoint) ConfigureProgram(m *libbpf.Map) error {
 		globalData.JumpsV6[tcdefs.ProgIndexPolicy] = uint32(ap.PolicyIdxV6)
 	}
 
-	return ConfigureProgram(m, ap.Iface, &globalData)
+	ConfigureProgram(ap.Iface, globalData)
+	return globalData
 }
 
-func ConfigureProgram(m *libbpf.Map, iface string, globalData *libbpf.TcGlobalData) error {
+func ConfigureProgram(iface string, globalData *libbpf.TcGlobalData) {
```
tomastigera, line 455: you can remove the `func ConfigureProgram` and do the little bit right in the ap

felix/bpf/bpf.go
```diff
@@ -2349,3 +2350,67 @@ func IterPerCpuMapCmdOutput(output []byte, f func(k, v []byte)) error {
 	}
 	return nil
 }
+
+func ConfigureAndLoad(file string, data libbpf.GlobalData, mapsToBePinned ...string) (*libbpf.Obj, error) {
```
tomastigera, line 2354:
```suggestion
func LoadFile/Object(file string, data libbpf.GlobalData, mapsToBePinned ...string) (*libbpf.Obj, error) {
```
My slight preference would be to avoid somethingAndSomethingElse. We are loading the object file with given parameters, it is a function after all ;-)

tomastigera (commented):
LGTM, some nits

tomastigera (approved):

PR #9669, opened 2025-01-02
Modify IPSet Setfilter to make it suitable for filtering ipsets dynamically

sridhartigera:
## Description

This PR modifies `SetFilter` in ipsets to make it suitable to filter ipsets dynamically. The changes are
1. Change SetFilter to set a func which takes ipset name and returns whether it needs to be filtered or not.
2. Provide another API `MarkDirty` which takes a set of ipset names and marks it dirty. This is used for rawEgressPolicy when in BPF mode.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/ipsets/ipsets.go
+++ b/felix/bpf/ipsets/ipsets.go
@@ -57,6 +57,8 @@ type bpfIPSets struct {
 	opRecorder logutils.OpRecorder
 
 	lg *log.Entry
+
+	filterIPSet func(string) bool
 }
 
 func NewBPFIPSets(
@@ -370,9 +372,11 @@ func (m *bpfIPSets) markIPSetDirty(data *bpfIPSet) {
 	m.dirtyIPSetIDs.Add(data.ID)
 }
 
-func (m *bpfIPSets) SetFilter(ipSetNames set.Set[string]) {
-	// Not needed for this IP set dataplane.  All known IP sets
-	// are written into the corresponding BPF map.
+func (m *bpfIPSets) SetFilter(fn func(ipSetName string) bool) {
+	m.filterIPSet = fn
+}
+
+func (m *bpfIPSets) MarkDirty(ipsetNames set.Set[string]) {
 }
 
 type bpfIPSet struct {
--- a/felix/dataplane/linux/int_dataplane.go
+++ b/felix/dataplane/linux/int_dataplane.go
@@ -770,7 +770,12 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		bpfutils.RemoveBPFSpecialDevices()
 	} else {
 		// In BPF mode we still use iptables for raw egress policy.
-		dp.RegisterManager(newRawEgressPolicyManager(rawTableV4, ruleRenderer, 4, ipSetsV4.SetFilter, config.RulesConfig.NFTables))
+		mgr := newRawEgressPolicyManager(rawTableV4, ruleRenderer, 4, ipSetsV4.MarkDirty, config.RulesConfig.NFTables)
+		ipSetsV4.SetFilter(func(ipSetName string) bool {
+			neededIPSets := mgr.GetNeededIPSets()
+			return neededIPSets.Contains(ipSetName)
+		})
+		dp.RegisterManager(mgr)
 	}
 
 	interfaceRegexes := make([]string, len(config.RulesConfig.WorkloadIfacePrefixes))
@@ -1035,7 +1040,12 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 				config.MaxIPSetSize))
 			dp.RegisterManager(newPolicyManager(rawTableV6, mangleTableV6, filterTableV6, ruleRenderer, 6, config.RulesConfig.NFTables))
 		} else {
-			dp.RegisterManager(newRawEgressPolicyManager(rawTableV6, ruleRenderer, 6, ipSetsV6.SetFilter, config.RulesConfig.NFTables))
+			mgr := newRawEgressPolicyManager(rawTableV6, ruleRenderer, 6, ipSetsV6.MarkDirty, config.RulesConfig.NFTables)
+			ipSetsV6.SetFilter(func(ipSetName string) bool {
+				neededIPSets := mgr.GetNeededIPSets()
+				return neededIPSets.Contains(ipSetName)
+			})
+			dp.RegisterManager(mgr)
 		}
 
 		dp.RegisterManager(newEndpointManager(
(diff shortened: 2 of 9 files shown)


felix/ipsets/ipsets.go
```diff
@@ -1092,31 +1092,29 @@ func (s *IPSets) updateDirtiness(name string) {
...
 	}
 
 	// We are filtering down, so compare against the needed set.
-	return s.neededIPSetNames.Contains(name)
+	return s.filterIPSet(name)
+}
+
+func (s *IPSets) MarkDirty(ipsetNames set.Set[string]) {
+	for name, meta := range s.setNameToAllMetadata {
+		if ipsetNames.Contains(name) {
```
fasaxc, line 1111: This is now the wrong check because `s.filterIPSet` should govern whether the IP set is programmed.  You should remove the `ipsetNames` parameter and use the filter.  Then this method can be called `OnFilterUpdated` or something like that and it just triggers a recheck of the filter for all IP sets.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1035,7 +1040,12 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 				config.MaxIPSetSize))
 			dp.RegisterManager(newPolicyManager(rawTableV6, mangleTableV6, filterTableV6, ruleRenderer, 6, config.RulesConfig.NFTables))
 		} else {
-			dp.RegisterManager(newRawEgressPolicyManager(rawTableV6, ruleRenderer, 6, ipSetsV6.SetFilter, config.RulesConfig.NFTables))
+			mgr := newRawEgressPolicyManager(rawTableV6, ruleRenderer, 6, ipSetsV6.MarkDirty, config.RulesConfig.NFTables)
+			ipSetsV6.SetFilter(func(ipSetName string) bool {
+				neededIPSets := mgr.GetNeededIPSets()
```
fasaxc, line 1045: This is a very inefficient way to do this check.  Every call to the filter will build the full "needed sets" `Set` and then throw it away again.  That is `O(n^2)` in the number of IP sets, which is not acceptable.

However, now you've switched to a "pull" model for this, you should be able to do a lot better.  What if you made the API `mgr.IPSetNeeded(name)`?  Then you could do a trivial map lookup inside the manager instead of building the unneeded set.

felix/bpf/ipsets/ipsets.go
```diff
@@ -370,9 +380,11 @@ func (m *bpfIPSets) markIPSetDirty(data *bpfIPSet) {
...
 }
 
-func (m *bpfIPSets) SetFilter(ipSetNames set.Set[string]) {
-	// Not needed for this IP set dataplane.  All known IP sets
-	// are written into the corresponding BPF map.
+func (m *bpfIPSets) SetFilter(fn func(ipSetName string) bool) {
+	m.filterIPSet = fn
+}
+
+func (m *bpfIPSets) MarkDirty(ipsetNames set.Set[string]) {
```
fasaxc, line 387: Feel a bit uneasy that we have a flexible filter function but we're not implementing the rescan here.  Someone could come along later and use the filter function without realising this isn't implemented.

Should at least comment "Not implemented because the only filter we use with this IP set is 100% static, based on the IP set name."

felix/dataplane/windows/ipsets/ipsets.go
```diff
@@ -193,6 +193,8 @@ func (m *IPSets) ApplyDeletions() bool {
 	return false
 }
 
-func (s *IPSets) SetFilter(ipSetNames set.Set[string]) {
-	// Not needed for Windows.
+func (m *IPSets) SetFilter(fn func(ipSetName string) bool) {
```
fasaxc, line 196: Worth commenting that these are deliberately not implemented

felix/nftables/ipsets.go
```diff
@@ -781,15 +781,13 @@ func (s *IPSets) updateDirtiness(name string) {
...
-	s.logCxt.Debugf("Filtering to needed IP set names: %v", ipSetNames)
-	s.neededIPSetNames = ipSetNames
+func (s *IPSets) SetFilter(fn func(ipSetName string) bool) {
+	s.filterIPSet = fn
+}
+
+func (s *IPSets) MarkDirty(ipsetNames set.Set[string]) {
 	for name, meta := range s.setNameToAllMetadata {
-		if s.ipSetNeeded(name) {
+		if ipsetNames.Contains(name) {
```
fasaxc, line 790: Same here, new filter func should govern the check.

felix/nftables/ipsets.go
```diff
@@ -97,7 +97,7 @@ type IPSets struct {
 
 	// Optional filter.  When non-nil, only these IP set IDs will be rendered into the dataplane
```
fasaxc, line 98: Out of date comment

felix/bpf/ipsets/ipsets.go
```diff
@@ -57,6 +57,8 @@ type bpfIPSets struct {
 	opRecorder logutils.OpRecorder
 
 	lg *log.Entry
+
+	filterIPSet func(string) bool
```
fasaxc, line 61: Worth a comment saying what this does.  In particular:
* What does `nil` mean?
* Is `true` "filter in" or "filter out"

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1035,7 +1040,12 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 				config.MaxIPSetSize))
 			dp.RegisterManager(newPolicyManager(rawTableV6, mangleTableV6, filterTableV6, ruleRenderer, 6, config.RulesConfig.NFTables))
 		} else {
-			dp.RegisterManager(newRawEgressPolicyManager(rawTableV6, ruleRenderer, 6, ipSetsV6.SetFilter, config.RulesConfig.NFTables))
+			mgr := newRawEgressPolicyManager(rawTableV6, ruleRenderer, 6, ipSetsV6.MarkDirty, config.RulesConfig.NFTables)
+			ipSetsV6.SetFilter(func(ipSetName string) bool {
```
fasaxc, line 1044: Worth a comment here saying what this filter stuff is about.  
```suggestion
			// When in BPF mode, we still program egress do-not-track rules into
			// (ip|nf)tables; set a filter on the IP sets we program so that only the 
			// IP sets needed for do-not-track are programmed.
			ipSetsV6.SetFilter(func(ipSetName string) bool {
```

fasaxc (changes requested):
I think this is the right sort of change but there are a couple of problems with the API.

I'm a bit wary that the filtering logic is implemented differently in each IP sets implementation.  Could we extract out a new struct that manages the IP set filtering and then re-use it in all the implementations so that they all have a "full" (but shared) implementation of filtering.  Feels like a bit of a footgun that the API is there but not implemented (or partially implemented) in some cases.

PR #9674, opened 2025-01-03
Implement partial resync after IP set update failures

fasaxc:
## Description

Any failure to update an IP set must either be spurious (EBUSY) or caused by a problem with _that_ IP set.  Historically, we've done an expensive full resync if we hit either type of error.

This PR changes that:

- After the first few failures, it does a partial resync, covering only the IP sets that we had already tried to update.  These are the only IP sets that we might have caused to be out of sync.
- After that, it does a full resync as before, just in case.

## Related issues/PRs

Might help with CORE-10804 CORE-11003

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
After failures to program IP sets, the dataplane now does a partial resync of only the affected IP sets. This should reduce the overhead of dealing with a transient failure.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/ipsets/ipsets.go
+++ b/felix/ipsets/ipsets.go
@@ -73,7 +73,8 @@ type IPSets struct {
 	nextTempIPSetIdx       uint
 	ipSetsWithDirtyMembers set.Set[string]
 
-	resyncRequired bool
+	ipSetsRequiringResync set.Typed[string]
+	fullResyncRequired    bool
 
 	// Factory for command objects; shimmed for UT mocking.
 	newCmd cmdFactory
@@ -134,7 +135,8 @@ func NewIPSetsWithShims(
 		mainSetNameToMembers: map[string]*deltatracker.SetDeltaTracker[IPSetMember]{},
 
 		ipSetsWithDirtyMembers: set.New[string](),
-		resyncRequired:         true,
+		ipSetsRequiringResync:  set.New[string](),
+		fullResyncRequired:     true,
 
 		newCmd: cmdFactory,
 		sleep:  sleep,
@@ -281,7 +283,8 @@ func (s *IPSets) RemoveMembers(setID string, removedMembers []string) {
 // QueueResync forces a resync with the dataplane on the next ApplyUpdates() call.
 func (s *IPSets) QueueResync() {
 	s.logCxt.Debug("Asked to resync with the dataplane on next update.")
-	s.resyncRequired = true
+	s.fullResyncRequired = true
+	s.ipSetsRequiringResync.Clear()
 }
 
 func (s *IPSets) GetIPFamily() IPFamily {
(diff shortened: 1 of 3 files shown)


felix/ipsets/ipsets.go
```diff
@@ -767,11 +797,13 @@ func (s *IPSets) tryUpdates(ipsetFilter func(ipSetName string) bool, programmedI
 
 	// Ask each dirty IP set to write its updates to the stream.
 	var writeErr error
+	var touchedIPSets []string
 	for _, setName := range dirtyIPSets {
 		// Ask IP set to write its updates to the stream.
 		if log.IsLevelEnabled(log.DebugLevel) {
 			log.WithField("setName", setName).Debug("Writing updates to IP set.")
 		}
+		touchedIPSets = append(touchedIPSets, setName)
```
tomastigera, line 806: shoudn't this be after the filter below? Or the filter below only means we are not interested in reporting back whether these IPs were programmed?

felix/ipsets/ipsets_test.go
```diff
@@ -753,6 +757,43 @@ var _ = Describe("IP sets dataplane", func() {
...
+
+					apply()
+					// The _other_ IP set shouldn't get fixed due to the
+					// targeted resync.
+					dataplane.ExpectMembers(map[string][]string{
+						v4MainIPSetName:  {},
+						v4MainIPSetName2: {"10.0.0.1", "10.0.0.3", "10.0.0.4"},
+					})
+
+					// But the next timed resync should catch it.
```
tomastigera, line 775:
```suggestion
					// But the next time resync should catch it.
```

tomastigera:
> * After the first few failures, it does a partial resync of the IP sets that we were attempting to update. This covers all cases I can think of that can cause an IP set update to fail.
> * After that, it does a full resync as before, just in case.

You mean that if the errors persists, it triggers a full resync, right?

felix/ipsets/ipsets.go
```diff
@@ -801,6 +833,9 @@ func (s *IPSets) tryUpdates(ipsetFilter func(ipSetName string) bool, programmedI
 			"stderr":     s.stderrCopy.String(),
 			"input":      s.restoreInCopy.String(),
 		}).Warning("Failed to complete ipset restore, IP sets may be out-of-sync.")
+		for _, setName := range touchedIPSets {
```
tomastigera, line 836: Doesn't it mean, that you then retry all IP sets up to the failure ...  but what about those that were not touched yet? Don't all of them remain dirty so they all would be retried anyway?

tomastigera (commented):
I might not fully understand the goal, but is it that if there are changes in N ipsets we keep retrying only those N instead of all? Or do we retry only a subset of those N? It feels like it is the former - if so then it LGTM

felix/ipsets/ipsets.go
fasaxc: Intention is to do a resync of the ones that we "touched" before we finish updating all the dirty IP sets.  We don't need to resync all dirty IP sets because the ones that we haven't touched yet should still be in the expected state.  The problem with the IP sets protocol is that we can't tell which changes were successful in the batch that we just tried so we have to resync all the IP sets that we might have modified.

felix/ipsets/ipsets.go
fasaxc: The filter is only about the reporting.

tomastigera (approved):

PR #9676, opened 2025-01-04
Remove unnecessary docker restart commands from Semaphore configs

hjiawei:
## Description

This changeset removes unnecessary docker restart steps from sempahore prologue. This steps may be useful for semaphore e1 generation machines which have limited resources. As we have migrated to newer f1 generation machine with better cpu and bigger disk spaces, we can remove these steps and reduce churns to the system during setup.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/.semaphore/push-images/alp.yml
+++ b/.semaphore/push-images/alp.yml
@@ -18,12 +18,6 @@ global_job_config:
   prologue:
     commands:
       - checkout
-      # Semaphore mounts a copy-on-write FS as /var/lib/docker in order to provide a pre-loaded cache of
-      # some images. However, the cache is not useful to us and the copy-on-write FS is a big problem given
-      # how much we churn docker containers during the build.  Disable it.
-      - sudo systemctl stop docker
-      - sudo umount /var/lib/docker && sudo killall qemu-nbd || true
-      - sudo systemctl start docker
       # Semaphore is doing shallow clone on a commit without tags.
       # unshallow it for GIT_VERSION:=$(shell git describe --tags --dirty --always)
       - retry git fetch --unshallow
--- a/.semaphore/release/release.yml
+++ b/.semaphore/release/release.yml
@@ -5,7 +5,7 @@ agent:
     type: f1-standard-4
     os_image: ubuntu2204
 execution_time_limit:
-  minutes: 800
+  hours: 14
 blocks:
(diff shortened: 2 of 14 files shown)


.semaphore/release/release.yml
```diff
@@ -24,26 +24,9 @@ blocks:
           # Load the github access secrets.  First fix the permissions.
           - chmod 0600 /home/semaphore/.keys/git_ssh_rsa
           - ssh-add /home/semaphore/.keys/git_ssh_rsa
-          # For some reason, /mnt is 100 GB and has a qemu-nbd image file.
-          # Let's delete it and use it for our own purposes (building calico
-          # without running out of space)
```
hjiawei, line 29: This is probably true for older e1 gen semaphore VMs but not for f1 we currently use. On f1, the 100GB cached container image disk `/dev/sdc1` is mounted directly into `/var/lib/docker`. There is no qemu-nbd image under `/mnt`. I believe the following commands essentially created a folder on the local physical volume and bind mount to the home `calico` folder.

.semaphore/push-images/alp.yml
```diff
@@ -18,12 +18,6 @@ global_job_config:
   prologue:
     commands:
       - checkout
-      # Semaphore mounts a copy-on-write FS as /var/lib/docker in order to provide a pre-loaded cache of
-      # some images. However, the cache is not useful to us and the copy-on-write FS is a big problem given
-      # how much we churn docker containers during the build.  Disable it.
-      - sudo systemctl stop docker
-      - sudo umount /var/lib/docker && sudo killall qemu-nbd || true
```
hjiawei, line 25: On f1 generation VMs, a third disk `/dev/sdc1` is mounted under `/var/lib/docker`. It has 100GB of storage space and pre-loaded images from semaphore.

```
$ mount | grep sdc
/dev/sdc1 on /var/lib/docker type ext4 (rw,noatime,nobarrier)
```

Once unmounted, we are using the first 2 physical volumes (`/dev/sda3` ~25GB + `/dev/sdb1` ~30GB) for Calico image caches. This is not ideal because `/dev/sdc1` is much bigger.

danudey (approved):

PR #9681, opened 2025-01-07
[release tool] fix hashrelease hanging on image checks

radTuti:
## Description

The components were not being pased on for hashrelease publish leading to the goroutine getting stuck waiting for results.

This adds timeout to each image check, includes more logs and validation checks during that process. 

## Related issues/PRs

- needs to be part of #9636 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/cmd/hashrelease.go
+++ b/release/cmd/hashrelease.go
@@ -209,8 +209,10 @@ func hashreleaseSubCommands(cfg *Config) []*cli.Command {
 					operator.WithValidate(!c.Bool(skipValidationFlag.Name)),
 					operator.WithTempDirectory(cfg.TmpDir),
 				)
-				if err := o.Publish(); err != nil {
-					return err
+				if !c.Bool(skipOperatorFlag.Name) {
+					if err := o.Publish(); err != nil {
+						return err
+					}
 				}
 
 				opts := []calico.Option{
@@ -230,6 +232,13 @@ func hashreleaseSubCommands(cfg *Config) []*cli.Command {
 				}
 				if reg := c.StringSlice(registryFlag.Name); len(reg) > 0 {
 					opts = append(opts, calico.WithImageRegistries(reg))
+				}
+				if !c.Bool(publishHashreleaseImageFlag.Name) {
+					components, err := pinnedversion.RetrieveImageComponents(cfg.TmpDir)
+					if err != nil {
+						return fmt.Errorf("failed to retrieve images for the hashrelease: %v", err)
+					}
+					opts = append(opts, calico.WithComponents(components))
 				}
 				r := calico.NewManager(opts...)
 				if err := r.PublishRelease(); err != nil {
--- a/release/cmd/hashrelease.go
+++ b/release/cmd/hashrelease.go
@@ -233,6 +233,8 @@ func hashreleaseSubCommands(cfg *Config) []*cli.Command {
 				if reg := c.StringSlice(registryFlag.Name); len(reg) > 0 {
 					opts = append(opts, calico.WithImageRegistries(reg))
 				}
+				// Note: We only need to check that the correct images exist if we haven't built them ourselves.
+				// So, skip this check if we're configured to build and publish images from the local codebase.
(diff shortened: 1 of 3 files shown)


release/cmd/hashrelease.go
```diff
@@ -209,8 +209,10 @@ func hashreleaseSubCommands(cfg *Config) []*cli.Command {
 					operator.WithValidate(!c.Bool(skipValidationFlag.Name)),
 					operator.WithTempDirectory(cfg.TmpDir),
 				)
-				if err := o.Publish(); err != nil {
-					return err
+				if !c.Bool(skipOperatorFlag.Name) {
```
radTuti, line 212: also noticed this flag wasn't being used

release/cmd/hashrelease.go
```diff
@@ -231,6 +233,13 @@ func hashreleaseSubCommands(cfg *Config) []*cli.Command {
 				if reg := c.StringSlice(registryFlag.Name); len(reg) > 0 {
 					opts = append(opts, calico.WithImageRegistries(reg))
 				}
+				if !c.Bool(publishHashreleaseImageFlag.Name) {
```
caseydavenport, line 236: Shouldn't this be `if c.Bool` instead of `if !c.Bool`?

release/pkg/manager/calico/manager.go
```diff
@@ -568,36 +626,19 @@ func (r *CalicoManager) hashreleasePrereqs() error {
...
+
 	if r.publishImages {
 		return r.assertImageVersions()
 	} else {
-		results := make(map[string]imageExistsResult, len(r.imageComponents))
-		ch := make(chan imageExistsResult)
-		for name, component := range r.imageComponents {
-			go imgExists(name, component, ch)
-		}
-		for range images {
```
caseydavenport, line 579: I think this should be `for range r.imageComponents` - the thing is that we're waiting for each started goroutine to finish, and we know we launched `r.imageComponents` goroutines. That would probably fix the original deadlock IIUC

release/cmd/hashrelease.go
coutinhop, line 236: do we really want to execute this if publishHashreleaseImageFlag is NOT true? or is the `!` a typo?

coutinhop (approved):

release/cmd/hashrelease.go
radTuti: `publishHashreleaseImageFlag` is for publishing images built locally by `hashrelease build`.
We need the list of components when the images have already been published else where

radTuti: The list of components is needed when not publishing images built by the hashrelease locally.

caseydavenport: Let's add a comment saying that then - this is super unintuitive (and caught both me and @coutinhop on this one!)

```
// We only need to check that the correct images exist if we haven't built them ourselves. So, skip this check
// if we're configured to build and publish images from the local codebase.
```

release/pkg/manager/calico/manager.go
```diff
@@ -552,13 +555,68 @@ type imageExistsResult struct {
...
+	for name, component := range r.imageComponents {
+		wg.Add(1)
+		go func(name string, component registry.Component, ch chan imageExistsResult) {
+			defer wg.Done()
+			ctx, cancel := context.WithTimeout(context.Background(), timeout)
+			defer cancel()
+
+			resultCh := make(chan imageExistsResult, 1)
+
+			go func() {
```
caseydavenport, line 581: Like we discussed in Slack, I'd like to see this simplified a lot. I think this is probably more concurrency / complexity than we need for this task, and will be prone to future bugs.

I think we can keep the previous implementation here and revert this change, and then make the change suggested here: https://github.com/projectcalico/calico/pull/9681/files#r1907632859

caseydavenport (changes requested):

caseydavenport (approved):

PR #9696, opened 2025-01-10
fix operator version for hashrelease

radTuti:
## Description

Current operator version for hashrelease is incorrect as it is missing the hashrelease it was built for which can cause issues when operator has not changed but there have been changes in calico.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/internal/version/version.go
+++ b/release/internal/version/version.go
@@ -27,8 +27,8 @@ import (
 	"github.com/projectcalico/calico/release/internal/utils"
 )
 
-// Data is the interface that provides version data for a release.
-type Data interface {
+// Versions is the interface that provides version data for a hashrelease or release.
+type Versions interface {
 	Hash() string
 	ProductVersion() string
 	OperatorVersion() string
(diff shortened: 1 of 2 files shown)


release/internal/version/version.go
```diff
@@ -36,25 +36,27 @@ type Data interface {
...
+	releaseName string
 }
 
 func (v *CalicoVersionData) ProductVersion() string {
 	return v.calico.FormattedString()
 }
 
 func (v *CalicoVersionData) OperatorVersion() string {
-	return v.operator
+	return fmt.Sprintf("%s-%s", v.operator, v.releaseName)
```
caseydavenport, line 59: Couldn't the operator version just include the productVer and that would be unique?

Is this function called for real releases at all? This format would be wrong for a real release (I don't think it is... but wanted to check)

radTuti: the previous process did use the release name. Although using the product version is much better than an arbitrary release name.

This function/ struct is only used for hashreleases.

caseydavenport (approved):

PR #9698, opened 2025-01-10
Program routes for VXLAN tunnel IPs assigned from /32 or /128 blocks

caseydavenport:
## Description

When VXLAN tunnel addresses are assigned from /32 or /128 blocks, the L3 route resolver was masking the REMOTE_WORKLOAD type route with REMOTE_TUNNEL. As a result, the routes weren't being programmed (because we only program routes for REMOTE_TUNNEL if they are borrowed IPs).

This PR modifies the RouteUpdate message to accept multiple types, allowing the L3 route resolver to mark the route with all of the matching types.

The VXLAN data plane now also identifies VXLAN routes that are both REMOTE_TUNNEL and REMOTE_WORKLOAD and programs directly connected routes for them.

This fixes an issue where host <-> pod connectivity was not functional when assigning VXLAN tunnel addresses from /32 or /128 IPAM blocks.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix route programming for VXLAN tunnel IPs assigned from a /32 or /128 IPAM block.
```

## Reminder for the reviewer

Other optional labels:


felix/calc/l3_route_resolver.go
```diff
@@ -843,6 +852,7 @@ func (c *L3RouteResolver) flush() {
 				}
 			}
 		}
+		rt.Types = types.Slice()
```
fasaxc, line 855: Good idea to sort the resulting to make it deterministic?

felix/proto/felixbackend.proto
```diff
@@ -605,7 +605,9 @@ message TunnelType {
 }
 
 message RouteUpdate {
+  // Deprecated - use RouteTypes instead.
```
fasaxc, line 608:
```suggestion
  // Deprecated - use types instead.
```

felix/proto/felixbackend.proto
```diff
@@ -617,6 +619,7 @@ message RouteUpdate {
   bool local_workload = 9;
   TunnelType tunnel_type = 10;
   bool borrowed = 11;
+  repeated RouteType types = 12;
```
fasaxc, line 622: I'd move this to line 610.  The id number will be out of order, but that's kindof the reason to have the id number: so that the order in the file can be "natural" while assigning unique IDs as we go.

felix/proto/felixbackend.proto
```diff
@@ -605,7 +605,9 @@ message TunnelType {
 }
 
 message RouteUpdate {
+  // Deprecated - use RouteTypes instead.
   RouteType type = 1;
```
fasaxc, line 609: Maybe remove this field (and mark IP ad reserved)?  You're no longer populating it so removing it will break code that uses it accidentally (at least when any consumers regenerate their protobuf).

fasaxc (commented):
Think I might have reached for a bit field and or'ed in the different flags but using a slice is probably ok.  (I have a little worry that l3rr might be on the critical path for larger clusters that don't use our IPAM.)

fasaxc (approved):
LGTM

PR #9708, opened 2025-01-13
Add initial flow aggregation code and API

caseydavenport:
## Description

This PR introduces a new component responsible for ingesting, aggregating, and the servering flow log data from the individual felix daemons deployed in a Calico cluster. This is still an early sketch of the component, but the major functional pieces are in place so it's a good place to start.

It is made of a few parts:

- a gRPC API with which each Felix is able to push flow log data to the aggregator.
- an aggregator component, which groups / aggregates data from across the cluster into time-windowed buckets
- a gRPC API which allows querying of the aggregated data
- an HTTP emitter component, that emits further time-aggregated data to a configurable endpoint.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Introduces `calico/goldmane`, a new component for aggregating observability data across nodes.
```

## Reminder for the reviewer

Other optional labels:


goldmane/cmd/main.go
```diff
@@ -0,0 +1,165 @@
...
+	NumBucketsToCombine int `json:"num_buckets_to_combine" envconfig:"NUM_BUCKETS_TO_COMBINE" default:"20"`
+
+	// PushIndex is the index of the bucket which triggers pushing to the emitter. A larger value
+	// will increase the latency of emitted flows, while a smaller value will cause the emitter to emit
+	// potentially incomplete flows.
+	PushIndex int `json:"push_index" envconfig:"PUSH_INDEX" default:"30"`
+
+	// NoCache disables caching of emitted flow state, allowing the emitter
+	// to run without need for a k8s API server.
+	NoCache bool `json:"no_cache" envconfig:"NO_CACHE"`
```
mazdakn, line 89: If we might extend this, maybe we should called it something like `CachingStrategy` with a more extensible type.

goldmane/cmd/main.go
```diff
@@ -0,0 +1,165 @@
...
+			emitter.WithCACertPath(cfg.CACertPath),
+			emitter.WithClientKeyPath(cfg.ClientKeyPath),
+			emitter.WithClientCertPath(cfg.ClientCertPath),
+			emitter.WithServerName(cfg.ServerName),
+		)
+		aggOpts = append(aggOpts, aggregator.WithSink(logEmitter))
+		go logEmitter.Run(stopCh)
+	}
+
+	// Create an aggregator and collector, and connected the collector to the aggregator.
```
mazdakn, line 142: connected -> connect

goldmane/docker/flowgen/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:latest
```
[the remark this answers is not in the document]
caseydavenport: Oh we can, I think this was just an autocomplete or copy/paste error :D

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+func (a *LogAggregator) Stop() {
+	close(a.done)
+}
+
+func (a *LogAggregator) rollover() {
+	// Add a new bucket at the start and remove the last bucket.
+	start := time.Unix(a.buckets[0].EndTime, 0)
+	end := start.Add(a.rolloverTime)
```
caseydavenport, line 275: I have some small fear that small discrepancies in rollover time can cause our internal bucket array to eventually fall behind the current time.

If a rollover() is delayed by a second, for example due to a long query, then when the rollover occurs the new bucket will be offset by a second from the actual time. If this happens enough, then eventually new bucket end times may actually be in the past! We'll potentially get flow updates for `time.Now()` that don't match any of the buckets in our array. 

Need to think about how to resolve this - we could try to spot when this has happened and add buckets to compensate?

goldmane/cmd/main.go
caseydavenport: Yeah, I need to revisit this one - I only added it as a quick fix for some testing I wanted to do.

goldmane/pkg/aggregator/aggregator.go
caseydavenport: I've got a local commit that addresses this by using a variable, calculated rollover time to make sure we catch up / slow down our rollovers as needed.

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -0,0 +1,307 @@
...
+	// will cause the emitter to emit incomplete flows.
+	//
+	// Latency-to-emit is roughly (pushIndex * rolloverTime).
+	pushIndex int
+}
+
+func NewLogAggregator(opts ...Option) *LogAggregator {
+	// Establish default aggregator configuration.
+	a := &LogAggregator{
+		rolloverTime:       1 * time.Minute,
```
mazdakn, line 90: If we want 5min aggregation with 20 buckets, shouldn't this be 15sec?

goldmane/pkg/client/client.go
```diff
@@ -0,0 +1,236 @@
...
+		inChan:     make(chan *proto.Flow, 5000),
+		cache:      cache,
+	}
+}
+
+// flowKey wraps the canonical FlowKey type with a start and end time, since this cache
+// stores flows across multiple aggregation intervals.
+type flowKey struct {
+	StartTime int64
+	EndTime   int64
```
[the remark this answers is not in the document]
caseydavenport: They are used implicitly as identifying information as part the key in the map

goldmane/pkg/client/client.go
```diff
@@ -0,0 +1,236 @@
...
+		})
+		if err != nil {
+			b.Wait()
+			continue
+		}
+
+		// Send new Flows as they are received.
+		for flog := range c.inChan {
+			// Add the flow to our cache. It will automatically be expired in the background.
+			c.cache.Add(flog)
```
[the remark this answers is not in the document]
caseydavenport: I think that would defeat the purpose of this cache - we want to cache them specifically so that we can re-send flows after the server restarts (so we want to hold on to flows that we've already sent, rather than clean them up quicker)

goldmane/pkg/aggregator/types.go
```diff
@@ -0,0 +1,174 @@
...
+		EndTime:   end.Unix(),
+		Flows:     make(map[types.FlowKey]*types.Flow),
+		RuleIndex: make(map[string]set.Set[types.FlowKey]),
+	}
+}
+
+// merge merges the flows from b2 into b.
+func (b *AggregationBucket) merge(b2 *AggregationBucket) {
+	for k, v := range b2.Flows {
+		v := v
```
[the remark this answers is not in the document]
caseydavenport: doh, yeah still just muscle memory.

goldmane/pkg/aggregator/aggregator.go
caseydavenport: Yeah, we actually override this default value with one in cmd/main.go... but it's a bit confusing to have a non-functional default here. I'll update to match what we actually use.

goldmane/pkg/emitter/emitter.go
```diff
@@ -0,0 +1,264 @@
...
+	// Query the latest configmap.
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+
+	cm := &corev1.ConfigMap{}
+	if err := e.kcli.Get(ctx, configMapKey, cm); err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("error getting configmap: %v", err)
+	} else if errors.IsNotFound(err) {
+		// Configmap doesn't exist, nothing to load.
+		return nil
```
mazdakn, line 249: I think it worth adding  a debug log here.

goldmane/pkg/emitter/emitter.go
```diff
@@ -0,0 +1,264 @@
...
+	if err := e.loadCachedState(); err != nil {
+		logrus.Errorf("Error loading cached state: %v", err)
+	}
+
+	// This is the main loop for the emitter. It listens for new batches of flows to emit and emits them.
+	for {
+		// Get pending work from the queue.
+		key, quit := e.q.Get()
+		if quit {
+			return
```
mazdakn, line 106: Maybe we should log (debug) shutting down the go routine.

goldmane/pkg/emitter/emitter.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+		e.q.Done(key)
+
+		// Emit the bucket.
+		bucket, _ := e.buckets.get(key)
```
mazdakn, line 111: Should we handle if key does not exists? I'm not sure if that's possible, but it feels we should even if it's rare case or maybe if it's programmer mistake (which it should panic). WDYT?

goldmane/pkg/emitter/emitter.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	// Update the timestamp in the configmap.
+	cm.Data["latestTimestamp"] = fmt.Sprintf("%d", e.latestTimestamp)
+
+	if cm.ResourceVersion == "" {
+		// Create the configmap.
+		if err := e.kcli.Create(context.Background(), cm); err != nil {
+			return fmt.Errorf("error creating configmap: %v", err)
+		}
```
mazdakn, line 225: Also worth adding an info log that the config map is created.

mazdakn (commented):
Left some comments about emitter package.

goldmane/pkg/emitter/emitter.go
caseydavenport: Yeah, definitely should consider this case. I _think_ it can only happen with programming error? But I'll log an error and handle it to be safe.

goldmane/pkg/aggregator/aggregator_test.go
```diff
@@ -0,0 +1,538 @@
...
+					BytesIn:               100 * int64(test.expectedNumConnectionsStarted),
+					BytesOut:              200 * int64(test.expectedNumConnectionsStarted),
+					PacketsIn:             10 * int64(test.expectedNumConnectionsStarted),
+					PacketsOut:            20 * int64(test.expectedNumConnectionsStarted),
+					NumConnectionsStarted: int64(test.expectedNumConnectionsStarted),
+				}
+				ExpectFlowsEqual(t, &exp, flows[0])
+			} else {
+				// Should consistently return no flows.
+				for i := 0; i < 10; i++ {
```
mazdakn, line 374: Maybe replace this with a `consistently` assertion.

goldmane/pkg/aggregator/aggregator_test.go
```diff
@@ -0,0 +1,538 @@
...
+		BytesOut:              200,
+		PacketsIn:             10,
+		PacketsOut:            20,
+		NumConnectionsStarted: 1,
+	}
+	agg.Receive(&proto.FlowUpdate{Flow: fl})
+
+	// Expect the aggregator to have received it.
+	var flows []*proto.Flow
+	require.Eventually(t, func() bool {
```
mazdakn, line 89: The same comment that if we should instead use gomega.

mazdakn (approved):
I left a few more comments. But also approved the changes, as it looks great for initial PR.

goldmane/pkg/emitter/emitter_test.go
```diff
@@ -0,0 +1,312 @@
...
+		StartTime:             18,
+		EndTime:               28,
+		BytesIn:               100,
+		BytesOut:              200,
+		PacketsIn:             10,
+		PacketsOut:            20,
+		NumConnectionsStarted: 1,
+	}
+	expectedBody, err := json.Marshal(flow)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
caseydavenport: Oh, yeah I started using this because it's used in some of the other Enterprise tests that I modeled this after and I sort of like this a bit better. But I could switch.

PR #9719, opened 2025-01-15
apiserver: better logging of change-handling routines for extension-apiserver-authentication

aaaaaaaalex:
…on-apiserver-authentication ConfigMap

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix a bug where the Calico APIserver would restart for all updates to the `extension-apiserver-authentication`, regardless of whether the ConfigMap's data actually changed.
```

## Reminder for the reviewer

Other optional labels:


apiserver/cmd/apiserver/server/watch.go
```diff
@@ -81,6 +84,7 @@ func WatchExtensionAuth(ctx context.Context) (bool, error) {
 					n := new.(*corev1.ConfigMap)
 					// Only detect as changed if the version has changed
```
fasaxc, line 85: Do we definitely care about every revision change here?  Seems like we could easily check the data for changes instead:
```
if len(n.Data) != len(o.Data) { changed = true ... }
for k, v := range n.Data {
  if ov, ok := o.Data[k] !ok || ov != v { changed = true ... }
}
// Same for n.BinaryData ...
```
(It's common for an operator to tweak annotations on the object every few seconds, but we should look at the data we care about and ignore anything else.)

aaaaaaaalex: The intention of this change is to see if our cache is stale and re-iterating over the same revisions post-restart, rather than investigating the change itself. Though, I suppose if we confirm that our cache is not stale, inspecting the changes themselves is the next step, so I'll put both in, thanks!

aaaaaaaalex: This is done now

apiserver/cmd/apiserver/server/watch.go
```diff
@@ -81,6 +84,19 @@ func WatchExtensionAuth(ctx context.Context) (bool, error) {
...
+						if changedKey != "" {
+							oldVal = o.Data[changedKey]
+							newVal = o.Data[changedKey]
+						}
+						logrus.WithFields(logrus.Fields{
+							"oldResourceVersion": o.ResourceVersion,
+							"newResourceVersion": n.ResourceVersion,
+							"changedDataKey":     changedKey,
+							"oldKeyValue":        oldVal,
+							"newKeyValue":        newVal,
```
matthewdupre, line 98: We mustn't log sensitive values, and I think these come from a secret which by default makes them sensitive (unless carefully reasoned otherwise).

aaaaaaaalex: Sorted now, thanks

coutinhop (approved):

fasaxc (approved):

PR #9723, opened 2025-01-15
make dataplanes dual (internal plus external together)

VadimEisenberg:
## Description

Currently, for Felix it is possible to use either the Internal dataplane driver or an external one. In NeuReality, our hardware, employs a dual network stack for different types of traffic. One is the standard Linux stack, for which the internal Felix dataplane is used. For the other network stack, we developed our own dataplane driver (external).

We propose to change Calico code to enable using two dataplane drivers simultaneously. The internal dataplane driver will be designated as the primary one, the external dataplane driver will be designated as the secondary one. Felix will send messages to both dataplanes, but will receive messages from the primary dataplane only.

The change should not be disruptive to the current users of Calico since they will continue to specify either internal or external dataplane driver. In case both drivers will be specified, the dual dataplane mode will be employed.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/decorator.go
+++ b/felix/dataplane/decorator.go
@@ -1,0 +1,32 @@
+// Copyright © 2024 NeuReality Ltd., and its licensors. All rights reserved.
+// This software contains proprietary information of NeuReality Ltd.
+// You shall use such proprietary information only as may be permitted in writing by NeuReality Ltd.
+// All materials contained herein are the property of NeuReality Ltd.
+
+// THIS SOFTWARE IS PROVIDED BY NEUREALITY “AS IS” AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL NEUREALITY OR ITS LICENSORS
+// BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+package dataplane
+
+type dataplaneDriverDecorator struct {
+     primaryDriver DataplaneDriver
+     secondaryDriver DataplaneDriver 
+}
+
+func (decorator dataplaneDriverDecorator) SendMessage(msg interface{}) error {
+     decorator.secondaryDriver.SendMessage(msg)
+     return decorator.primaryDriver.SendMessage(msg) // return message from the primary driver only
+}
+
+func (decorator dataplaneDriverDecorator) RecvMessage() (msg interface{}, err error) {
+     return decorator.primaryDriver.RecvMessage() // receive message from the primary driver only
+}
--- a/felix/dataplane/driver.go
+++ b/felix/dataplane/driver.go
@@ -67,6 +67,11 @@ func StartDataplaneDriver(
 		return &inactive.InactiveDataplane{}, nil
 	}
 
+	var primaryDataplaneDriver DataplaneDriver
+	var primaryCmd *exec.Cmd
+	var secondaryDataplaneDriver DataplaneDriver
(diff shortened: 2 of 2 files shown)


caseydavenport:
@VadimEisenberg thanks for the PR, very cool! One thing - you'll need to sign the CLA in order to make the bot happy.

felix/dataplane/decorator.go
```diff
@@ -0,0 +1,32 @@
...
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+package dataplane
+
+type dataplaneDriverDecorator struct {
```
tomastigera, line 20: perhaps?

```suggestion
type dataplaneDriverMultiplexer struct {
```

tomastigera (commented):
What happens if the secondary driver has and error? Is it OK to ignore it?

felix/dataplane/decorator.go
VadimEisenberg: @tomastigera Great points, thank you! Let me rename the decorator, and return an error in case the secondary driver returns an error.

VadimEisenberg:
@caseydavenport Thank you! I have signed the CLA.

image

felix/dataplane/decorator.go
```diff
@@ -0,0 +1,32 @@
...
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL NEUREALITY OR ITS LICENSORS
+// BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
matthewdupre, line 16: You'll need to use the Apache license header here (similar to in the other files in the repo).  You're welcome to keep your copyright line (without "All rights reserved", which is typically used when no license is provided).

Please let me know if you have any specific concerns about submitting this under the Apache 2.0 license.

matthewdupre (changes requested):

felix/dataplane/decorator.go
VadimEisenberg: @matthewdupre Thank you for pointing my attention to this, I have fixed the header. I have no concerns about submitting this PR under the Apache 2.0 license.

VadimEisenberg:
@tomastigera I have applied your changes, please review. More comments are welcome.

tomastigera:
@VadimEisenberg there is already a mechanism to support multiple dataplanes. Afaik that mechanism is used to push policies to applicationlayer policy engine (Dikastes) Who could we unify these mechanisms. Having 2 seems like and overkill. Feels like your approach would not need separate gorutines to read from the per-dp channels, otoh your approach makes one dp wait for the other as the message is dispatched synchronously. Idk whether it is a requirement of your usecase.

cc @fasaxc @caseydavenport 

https://github.com/projectcalico/calico/blob/master/felix/daemon/daemon.go#L462

VadimEisenberg:
@tomastigera Using policy sync server will work for our needs, thank you! However, I have the second part of dual dataplanes issue, the one that deals with the CNI plugin dataplanes https://github.com/projectcalico/calico/pull/9743. Do you have a similar solution for CNI plugin dataplanes?

Let's discuss the two PRs together, what should be the right design for a dual network stack, considering Felix and CNI plugin dataplanes. Once we agree on the design, we can discuss technical details - using goroutines for primary and secondary dataplanes, etc.

tomastigera:
@VadimEisenberg sent you amessage on CU slack, where it might be easier to discuss.

VadimEisenberg:
Closing in favor of using policy sync server.

The policy sync server opened by Felix daemon: https://github.com/projectcalico/calico/blob/master/felix/daemon/daemon.go#L479 .

An example of listening to the policy sync server is dikastes: [calico/app-policy/cmd/dikastes/dikastes.go at master · projectcalico/calico](https://github.com/projectcalico/calico/blob/master/app-policy/cmd/dikastes/dikastes.go).

PR #9732, opened 2025-01-17
[BPF] BPFMapSizeConntrackScaling=DoubleIfFull

tomastigera:
When the conntrack table is running out of space, it doubles its size automatically. It is hard to pick up a good default so we start with something "good enough" and if it is not, then we double it and perhaps double again if need be. The memory is not preallocated and we periodically clean ut the table so if the usage goes down, memory should also go down.

This mode is the new default, but it can be turned off.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: conntrack table doubles in size if it is running out of space
```

## Reminder for the reviewer

Other optional labels:


ioworker0:
Nice~ It's really a great feature, and exactly what I was looking for. But, I'm also considering whether we should disable it by default since increaing the size can be quite disruptive, as Shaun mentioned[1].

Hmm.. It would be better to let users explicitly enable this mode after they fully understand the potential impacts, especially for newcomers, IMHO.

BTW, if there are any updates on conntrack, could you keep me informed? I'm really interested in it ;p

[1] https://calicousers.slack.com/archives/CPEPF833L/p1733907124977759?thread_ts=1733131626.596509&cid=CPEPF833L

ioworker0:
On second thought, what’s really important is that users can easily check the calico-node logs to identify where something went wrong. For example, if the nf_conntrack table is full, the kernel logs an error about it in dmesg.

While there’s no ideal default value, in practice, a single line of clear log is often more valuable than anything else. In production, nothing is more urgent than quickly detecting problems. There is room for improvement, so let’s work on it ;p

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -748,10 +748,17 @@ type FelixConfigurationSpec struct {
...
+	// BPFMapSizeConntrack*. 'DoubleIfFull' doubles the size when the map is pretty much full even
+	// after cleanups. [Default: DoubleIfFull]
+	// +kubebuilder:validation:Pattern=`^(?i)(Disabled|DoubleIfFull)?$`
+	BPFMapSizeConntrackScaling string `json:"bpfMapSizeConntrackScaling,omitempty"`
+
 	// BPFMapSizeConntrackCleanupQueue sets the size for the map used to hold NAT conntrack entries that are queued
 	// for cleanup.  This should be big enough to hold all the NAT entries that expire within one cleanup interval.
 	// +kubebuilder:validation:Minimum=1
-	BPFMapSizeConntrackCleanupQueue *int `json:"bpfMapSizeConntrackCleanupQueue,omitempty"  validate:"omitempty,gte=1"`
+	BPFMapSizeConntrackCleanupQueue *int `json:"bpfMapSizeConntrackCleanupQueue,omitempty"  validate:"omitempty,gte=v1"`
```
fasaxc, line 761: `v1`?

felix/bpf/conntrack/bpf_scanner.go
```diff
@@ -71,9 +74,14 @@ func registerConntrackMetrics() {
 // Note: the tests for this object are largely in the bpf/ut package, since
 // we require a privileged environment to test the BPF program.
 type BPFProgLivenessScanner struct {
-	ipVersion int
-	timeouts  Timeouts
-	logLevel  BPFLogLevel
+	ipVersion                    int
+	timeouts                     Timeouts
+	logLevel                     BPFLogLevel
+	scaling                      bool
```
fasaxc, line 80:
```suggestion
	autoScale                      bool
```
think this would be more guessable

felix/bpf/conntrack/bpf_scanner.go
```diff
@@ -217,18 +235,61 @@ func (s *BPFProgLivenessScanner) RunBPFExpiryProgram(opts ...RunOpt) error {
...
+
+func (s *BPFProgLivenessScanner) writeNewSizeFile() error {
+	// Make sure directory exists.
+	if err := os.MkdirAll("/var/lib/calico", os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create directory /var/lib/calico: %s", err)
+	}
+
+	newSize := 2 * s.maxEntries
+
+	// Write the smallest MTU to disk so other components can rely on this calculation consistently.
```
fasaxc, line 287: MTU?

felix/fv/bpf_map_resize_test.go
```diff
@@ -162,6 +163,110 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ Felix bpf test configurable
...
+		tc    infrastructure.TopologyContainers
+
+		w [2]*workload.Workload
+	)
+
+	BeforeEach(func() {
+		infra = getInfra()
+		opts := infrastructure.DefaultTopologyOptions()
+		opts.ExtraEnvVars["FELIX_BPFMapSizeConntrack"] = "100"
+		opts.ExtraEnvVars["FELIX_BPFConntrackLogLevel"] = "debug"
```
fasaxc, line 184: Set the "disable log dropping" setting to avoid flakes where your regex misses.

felix/bpf/conntrack/bpf_scanner.go
```diff
@@ -217,18 +235,61 @@ func (s *BPFProgLivenessScanner) RunBPFExpiryProgram(opts ...RunOpt) error {
...
+		s.higherCount = 0
+	}
+	s.liveEntries = newLiveEntries
+
+	full := float64(newLiveEntries) / float64(s.maxEntries)
+	log.Debugf("full %f, total %d, totalDeleted %d", full, total, totalDeleted)
+	// If the ct map keeps filling up and gets over 90% full or if it hits 95%
+	// no matter what, resize the map.
+	if s.higherCount >= 3 && full > 0.90 || full > 0.95 {
+		if err := s.writeNewSizeFile(); err != nil {
```
fasaxc, line 269: Suggest logging a warning here: "Conntrack table nearly full, initiating resize from x to y; will restart to apply new size..."

That'll let the user clearly see what triggered any disruption that they investigate later.

fasaxc (changes requested):
Few small queries but I think it'll be a good change.  I think leaving it on by default is probably for the best.  Might cause disruption, but so does a full conntrack table.

fasaxc (approved):

ioworker0:
> Few small queries but I think it'll be a good change. I think leaving it on by default is probably for the best. Might cause disruption, but so does a full conntrack table.

There is one more concern: auto-scaling could disrupt all existing connections, while a full conntrack table only blocks new connections. In production, AFAIK, the conntrack map typically fills up gradually, so the effects are not the same. wdyt?

tomastigera:
@ioworker0 you are right. Lets say that the ideal case is (and most people imo fall into this cathegory) that the size of the conntrack table is big enough. It is configurable. This automatic rezise is a safety valve and can be turned off. Also note that resizing the conntrack table is a global setting, while doubling the size of the table is local to the node where the ct is running out of space. Also note that some users experience overfill due to inactive udp/dns connections or short lived tcp connections so any disruption due to the processes of resizing is much smaller.

ioworker0:
> @ioworker0 you are right. Lets say that the ideal case is (and most people imo fall into this cathegory) that the size of the conntrack table is big enough. It is configurable. This automatic rezise is a safety valve and can be turned off. Also note that resizing the conntrack table is a global setting, while doubling the size of the table is local to the node where the ct is running out of space. Also note that some users experience overfill due to inactive udp/dns connections or short lived tcp connections so any disruption due to the processes of resizing is much smaller.

Thanks a lot for the lesson!

It makes sense to me now, and I see that I missed something important before ;p

ioworker0:
> If the default setting proves to be a problem, we can change it in the future ;-)

Yep, totally agree ;)

PR #9741, opened 2025-01-22
[BPF] Add support for tracking conntrack metrics in Userspace mode

ioworker0:
## Description

When BPFConntrackCleanupMode is set to BPFConntrackModeUserspace, or in cases where the kernel does not support BTF (unfortunately, this is the case with some of our clusters), we are missing metrics related to the conntrack map. These metrics are obviously very useful, so let's also track these metrics under the BPFConntrackModeUserspace setting.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


ioworker0:
CC @tomastigera @fasaxc

ioworker0:
> CC @tomastigera @fasaxc

Oops, the commit 'Adding flow log support #9737 ' made similar changes before me. But, I'm just wondering why we use different metrics for BPFConntrackModeUserspace and BPFConntrackModeBPFProgram. Since they both *actually* do the same task, wouldn't it be better to use the same metric for both as I did? This way would likely be more user-friendly IHMO. wdyt?

tomastigera:
well, the reason is that the flow log supports pulls in something that precede the bpf cleaner and it stats and should have been merge long time ago. Afaict we can unify those two and ditch the ones that flow logs PR brought over. Feel free to resolve the conflicts in the favor of your PR.

The userspace cleaner will be deprecated sooner or later anyway since the bpf one is superior and the default.

tomastigera:
The userspace cleaner is not going anywhere anytime soon

ioworker0:
> /sem-approve

I think the conflicts are resolved, and the CI issues should be fixed as well ;)

ioworker0:
> > /sem-approve
> 
> I think the conflicts are resolved, and the CI issues should be fixed as well ;)

Ping @fasaxc @tomastigera ;p

tomastigera:
Closing this since we merge both ebpf modeand userspace mode together

Issue #9724, opened 2025-01-15
Enable Calico dataplanes to operate in a dual mode (internal plus external together)


VadimEisenberg:
Currently, for Felix it is possible to use either the Internal dataplane driver or an external one. In NeuReality, our hardware employs a dual network stack for different types of traffic. One is the standard Linux stack, for which the internal Felix dataplane is used. For the other network stack, we developed our own dataplane driver (external).

We propose to change Calico code to enable using two dataplane drivers simultaneously. The internal dataplane driver will be designated as the primary one, the external dataplane driver will be designated as the secondary one. Felix will send messages to both dataplanes, but will receive messages from the primary dataplane only.

The change should not be disruptive to the current users of Calico since they will continue to specify either internal or external dataplane driver. In case both drivers will be specified, the dual dataplane mode will be employed.

The change is very simple, here is the PR - https://github.com/projectcalico/calico/pull/9723.

## Expected Behavior
When both internal dataplane driver and an external dataplane driver are specified, both should run in dual mode: sending messages to both dataplane drivers but receiving messages from the internal dataplane driver only (the primary one).

## Current Behavior
Currently, only one dataplane driver can be specified – either the internal one or an external one.

## Possible Solution

Here is a possible PR to implement it https://github.com/projectcalico/calico/pull/9723.

## Context
Our hardware employs a dual network stack for different types of traffic. One is the standard Linux stack, for which the internal Felix dataplane is used. For the other network stack, we developed our own dataplane driver (external).

## Your Environment

* Calico dataplane: the internal Linux dataplane driver + our custom one
* Orchestrator version Kubernetes
* Link to your project (optional): [NeuReality | We Make AI Easy](https://www.neureality.ai/)

VadimEisenberg:
Closing in favor of using policy sync server.

The policy sync server opened by Felix daemon: https://github.com/projectcalico/calico/blob/master/felix/daemon/daemon.go#L479 .

An example of listening to the policy sync server is dikastes: [calico/app-policy/cmd/dikastes/dikastes.go at master · projectcalico/calico](https://github.com/projectcalico/calico/blob/master/app-policy/cmd/dikastes/dikastes.go).

PR #9743, opened 2025-01-22
enable dual CNI plugin dataplanes

VadimEisenberg:
## Description

Currently, for CNI plugin it is possible to use either the default system dataplane or a grpc one. In NeuReality, our hardware, employs a dual network stack for different types of traffic. One is the standard Linux stack, for which the default system dataplane is used. For the other network stack, we developed our own dataplane (external).

We propose to change Calico code to enable using two dataplanes simultaneously. The default system dataplane will be designated as the primary one, the external dataplane will be designated as the secondary one. Calico will send messages to both dataplanes.

The change should not be disruptive to the current users of Calico since they will continue to specify either default system dataplane or a grpc dataplane. In case a dataplane option `"useAsSecondary": "true"` in addition to `"type": "grpc"`, two dataplanes will be used simultaneously.

## Related issues/PRs

* https://github.com/projectcalico/calico/issues/9724
* https://github.com/projectcalico/calico/pull/9723

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

### DIFF
--- a/cni-plugin/pkg/dataplane/dataplane.go
+++ b/cni-plugin/pkg/dataplane/dataplane.go
@@ -50,10 +50,28 @@ func GetDataplane(conf types.NetConf, logger *logrus.Entry) (Dataplane, error) {
 	if !ok {
 		return getDefaultSystemDataplane(conf, logger)
 	}
+
+	var externalDataplane Dataplane
+
 	switch name {
 	case "grpc":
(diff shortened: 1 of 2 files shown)


VadimEisenberg:
This functionality can be handled by CNI plugin chaining, by adding an additional CNI plugin.

PR #9745, opened 2025-01-22
Refactor internals of Goldmane

caseydavenport:
## Description

This is still somewhat a WIP, but thought it was far enough along to be a reasonable increment.

The main changes here:

- Switch the aggregation bucket array to be a circular ring, rather than needing to append / remove buckets on rollover.
- Introduce the concept of a Cascade to track a particular Flow's stats over time.
- Introduce Index pattern for more efficient queries based on various sorts.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


goldmane/pkg/aggregator/aggregator.go
```diff
@@ -34,19 +35,32 @@ const (
...
 
+type flowResponse struct {
+	flows []*proto.Flow
+	err   error
+}
+
 type LogAggregator struct {
-	buckets []AggregationBucket
+	// indices allow for quick handling of flow queries sorted by various methods.
+	indicies map[proto.SortBy]Index[string]
```
caseydavenport, line 51: This is just by string right now since it's the only kind we have. Probably want separate maps for each sortable type we want, eventually:

```go
  stringIndicies map[proto.SortBy]Index[string]
  intIndicies map[proto.SortBy]Index[int]
```

Having them as a map makes it simple to lookup which index satisfies the given request, and allows us to iterate all indicies in places we need to - like adding/removing DiachronicFlow objs.

goldmane/pkg/aggregator/index.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (idx *index[E]) List(opts IndexFindOpts[E]) []*types.Flow {
+	var matchedFlows []*types.Flow
+
+	// Find the index of the DiachronicFlow from which to start the search. If the cursor is 0, we start from the beginning.
+	// Otherwise, we start from the DiachronicFlow with the ID equal to the cursor.
+	var i int
+	if opts.cursor > 0 {
+		i = sort.Search(len(idx.diachronics), func(i int) bool {
+			return idx.diachronics[i].ID > opts.cursor
```
caseydavenport, line 65: @Brian-McM you had this in your original implementation of this against my in-progress branch, but looking at it now I am not sure this works. Namely because `sort.Search` is a binary search and `idx.diachronics` is sorted by the `keyFunc` and not by `ID` 

I think this means the cursor needs to be the same value as the sort key (i.e., a DestName for a DestName index, or a DestNamespace for a DestNamespace index). WDYT?

goldmane/pkg/aggregator/index_test.go
```diff
@@ -0,0 +1,155 @@
...
+		ID: 3,
+		Key: types.FlowKey{
+			DestName:      "a",
+			DestNamespace: "ns2",
+		},
+	}
+	idx.Add(trickyFlow)
+
+	// Verify the DiachronicFlows are ordered correctly. The two flows with the same DestName should be adjacent,
+	// sorted by their ID.
```
caseydavenport, line 118: I should add a test case that verifies subsequent updates with lower ID values get sorted properly as well.

goldmane/pkg/aggregator/index.go
Brian-McM: Ya there's a bug here, my mind wasn't working quite right when I was sorting this out. Basically to speed up searching, the index is sorted by the main key (dest) then by ID, so if you have thousands of flows for the same destination then the look up is quicker.

goldmane/pkg/aggregator/index.go
```diff
@@ -0,0 +1,181 @@
...
+		k1 := idx.keyFunc(&idx.diachronics[i].Key)
+		k2 := idx.keyFunc(&d.Key)
+		if k1 > k2 {
+			// The key of the DiachronicFlow at index i greater than the key of the flow.
+			return true
+		}
+		if k1 == k2 {
+			// The field(s) this Index is optimized for considers these keys the same.
+			// Sort based on the key's ID to ensure a deterministic order.
+			// TODO: This will result in different ordering on restart. Should we sort by FlowKey fields instead
```
Brian-McM, line 170: It might not actually matter if the order changes after restart, since this is just for efficient lookups. The only thing that would be affected is the slight time difference certain lookups take for specific flows, i.e. looking up a flow that is in middle of the or ends of the index will be found in 1 or 2 checks, but that could move to somewhere that takes logn checks.

goldmane/pkg/aggregator/types.go
```diff
@@ -137,28 +100,245 @@ func GetStartTime(interval int) int64 {
...
+
+// indexSubtract subtracts n from idx, wrapping around if necessary.
+func (r *BucketRing) indexSubtract(idx, n int) int {
+	return (idx - n + len(r.buckets)) % len(r.buckets)
+}
+
+func (r *BucketRing) findBucket(time int64) (int, *AggregationBucket) {
+	// Find the bucket that contains the given time.
+	// TODO: We can do this without iterating over all the buckets by simply calculating
+	// the index based on the time. It's a very small win though - there aren't that many buckets to iterate.
```
Brian-McM, line 252: But we can at least brag about how smart we are :)

Brian-McM (approved):

PR #9754, opened 2025-01-24
update manifests for dev releases

radTuti:
## Description

The maifests generated when the registry is overwritten are currently incorrect since the generate manifests currently does not update the registries. This attempts to fix that by:

- update calico chart values to separate registry from image to make it easier to change
- update tigera-operator values to separate registry from image for calicoctl
- update calico & tigera-operator chart templates to accomodate this new separation

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


manifests/generate.sh
```diff
@@ -11,17 +11,32 @@ HELM=${HELM:-../bin/helm}
...
+OPERATOR_IMAGE=${OPERATOR_IMAGE:-$defaultOperatorImage}
+
 NON_HELM_MANIFEST_IMAGES="calico/apiserver calico/windows calico/ctl calico/csi calico/node-driver-registrar calico/dikastes calico/flannel-migration-controller"
 
 echo "Generating manifests for Calico=$CALICO_VERSION and tigera-operator=$OPERATOR_VERSION"
 
+##########################################################################
+# Update images in the values file
+##########################################################################
+if [[ $REGISTRY != "calico" ]]; then
```
caseydavenport, line 33:
```suggestion
if [[ $REGISTRY != $defaultRegistry ]]; then
```

manifests/generate.sh
```diff
@@ -32,47 +47,48 @@ metadata:
...
+  --set calicoctl.tag=$CALICO_VERSION \
+  ../charts/tigera-operator >>tigera-operator.yaml
 
 ##########################################################################
 # Build CRD manifest.
 #
 # This manifest is used in "Calico the hard way" documentation.
 ##########################################################################
-echo "# CustomResourceDefinitions for Calico the Hard Way" > crds.yaml
+echo "# CustomResourceDefinitions for Calico the Hard Way" >crds.yaml
```
caseydavenport, line 64:
```suggestion
echo "# CustomResourceDefinitions for Calico the Hard Way" > crds.yaml
```

caseydavenport: Few of these spacing changes aren't right

manifests/generate.sh
```diff
@@ -96,15 +112,18 @@ done
...
+  ../charts/tigera-operator/ \
+  --output-dir ocp \
+  --no-hooks \
+  --set installation.kubernetesProvider=OpenShift \
+  --set installation.enabled=false \
+  --set apiServer.enabled=false \
+  --set tigeraOperator.image=$OPERATOR_IMAGE \
+  --set tigeraOperator.version=$OPERATOR_VERSION \
+  --set tigeraOperator.registry=$OPERATOR_REGISTRY \
+  --set calicoctl.image=$REGISTRY/ctl \
```
caseydavenport, line 125: Seems wrong that `calicoctl.image` is of the form REGISTRY/CTL whereas the `tigeraOperator.image` is just of the form `tigera/operator` with the registry configured separately - we should be using consistent meanings for image / version / registry.

release/pkg/manager/calico/manager.go
```diff
@@ -891,7 +897,7 @@ func (r *CalicoManager) generateManifests() error {
 }
 
 func (r *CalicoManager) resetManifests() {
-	if _, err := r.runner.RunInDir(r.repoRoot, "git", []string{"checkout", "manifests"}, nil); err != nil {
+	if _, err := r.runner.RunInDir(r.repoRoot, "git", []string{"checkout", "manifests", "charts"}, nil); err != nil {
```
caseydavenport, line 900: What change happens in the charts diretory? I can't see where we're modifying that

release/pkg/manager/calico/manager.go
```diff
@@ -358,13 +358,13 @@ func (r *CalicoManager) getRegistryFromManifests() (string, error) {
...
 	imgs := strings.Split(out, "\n")
 	for _, i := range imgs {
-		if strings.Contains(i, "operator") {
-			continue
-		} else if strings.Contains(i, "tigera/") {
-			splits := strings.SplitAfter(i, "/tigera/")
-			registry := splits[0]
-			logrus.WithField("registry", registry).Debugf("Using registry from image %s", i)
-			return registry, nil
+		if strings.HasPrefix(i, "calico/") {
```
caseydavenport, line 361: I don't quite understand this change - what does it do? We're ignoring "calico/" images? 

The old implementation had a log that emitted what registry was chosen and why - I think we should at least keep that.

caseydavenport (changes requested):

manifests/generate.sh
radTuti: the calicoctl images are pushed as <registry>/ctl

sadly we currently do not follow a consistent pattern b/w operator and monorepo images, that will require a bigger overhaul which is beyond the scope of this PR

release/pkg/manager/calico/manager.go
radTuti: it is not ignoring calico images. The images are currently listed as `calico/<image_name>` so when the image is set as `calico/<image_name>`, this passes an empty string as a result.

release/pkg/manager/calico/manager.go
radTuti: the images in `charts/calico/values.yaml` have their registies updated by `manifests/generate.sh` [here](https://github.com/projectcalico/calico/pull/9754/files#diff-3c90671bf6dbbd7436fde8f5512b677ebc4846700d80dc63635a30b4b58a68afR34)

manifests/generate.sh
caseydavenport: > The calicoctl images are pushed as /ctl

Surely they must be pushed as `docker.io/calico/ctl` no?

Operator has two separate config options - `registry` and `image`

Calicoctl has a single option - `image` - that appears to be a merged version of registry and image.

release/pkg/manager/calico/manager.go
caseydavenport: Ah, right. I need to think about that one for a moment. I think it's a change in scope of the `generate.sh` script that I might not be completely comfortable with. That script is meant to deterministically generate manifests based on values.yaml, not be responsible for modifying values.yaml itself.

release/pkg/manager/calico/manager.go
caseydavenport: I don't understand why we would want to do that though. 

Maybe you could explain what was wrong with the previous implementation and that will help me understand the change.

manifests/generate.sh
radTuti: For my dev images, it is created as `docker.io/tuticat/ctl`

When the default registry is not overwritten, this be set to `calico/ctl` which matches the current state of `image: calico/ctl:master`

manifests/generate.sh
```diff
@@ -32,49 +42,60 @@ metadata:
...
+  --set installation.enabled=false \
+  --set apiServer.enabled=false \
+  --set whisker.enabled=false \
+  --set goldmane.enabled=false \
+  --set tigeraOperator.version=$OPERATOR_VERSION \
+  --set tigeraOperator.image=$OPERATOR_IMAGE \
+  --set tigeraOperator.registry=$OPERATOR_REGISTRY \
+  --set calicoctl.tag=$CALICO_VERSION \
+  --set calicoctl.registry=$REGISTRY \
+  ../charts/tigera-operator >>tigera-operator.yaml
```
danudey, line 55: The whitespace changes makes it difficult to review the non-whitespace changes. If changing indents, it would be best to leave this to a separate, whitespace-specific commit so that functional changes can be distinguished from cosmetic changes.

danudey: (I know Github has an option to hide whitespace, but it also makes it difficult to look at changes on the command line, to see relevant changes in `git blame`, etc.)

charts/calico/values.yaml
```diff
@@ -2,25 +2,34 @@
 version: master
 
 # Configure the images to use when generating manifests.
+# If a new image is added, update manifests/generate.sh
 node:
-  image: quay.io/calico/node
+  image: node
+  registry: quay.io/calico
```
caseydavenport, line 8: I think that everywhere else, we consider the registry to be `quay.io` and the image to be `calico/node` - we should be consistent here as well.

charts/tigera-operator/values.yaml
```diff
@@ -81,7 +81,8 @@ tigeraOperator:
   version: master
   registry: quay.io
 calicoctl:
-  image: quay.io/calico/ctl
+  image: ctl
+  registry: quay.io/calico
```
caseydavenport, line 85: Same here - you can see that the registry for `tigera/operator` is `quay.io` - we should follow that definition of what a registry is.

release/pkg/manager/calico/manager.go
```diff
@@ -39,9 +39,10 @@ import (
 // Global configuration for releases.
 var (
 	// Default defaultRegistries to which all release images are pushed.
+	// The first registry is what the manifests use. Currently it is quay.io/calico.
```
caseydavenport, line 42: This isn't what I expected - I'd have expected that `gen-manifests` should use the values specified in values.yaml unless the user has provided an override. 

I think we should keep that behavior - otherwise, the CI checks that verify manifests are up-to-date are testing the wrong thing.

radTuti: this does not negate the CI checks that verify manifests.

Below, the first registry in image registries is used to generate the manifests. If a user does not override the default, this is stated to keep it aligned with what happens when `gen-manifests` is run separately.

charts/calico/values.yaml
radTuti: Actually, monorepo considers registry to be `quay.io/calico`. See [defaultRegistries](https://github.com/projectcalico/calico/blob/master/release/pkg/manager/calico/manager.go#L42-L48)

There is discrepancy with operator which does use `quay.io` as registry and `tigera/operator` as image but aligning them is beyond the scope of this PR.

charts/tigera-operator/values.yaml
radTuti: Stated this on a different comment but `tigera/operator` follows a different format from monorepo. This follows the current way the monorepo operates. See [defaultRegistries](https://github.com/projectcalico/calico/blob/master/release/pkg/manager/calico/manager.go#L42-L48)
Changing this from `quay.io/calico` with image as `node` to `quay.io` with image as `calico/node` will require changing a lot more beyond the scope of this PR

release/pkg/manager/calico/manager.go
caseydavenport: > keep it aligned with what happens when gen-manifests is run separately

Keeping something aligned / in-sync is not nearly as good as running it through the same code path.

If `values.yaml` changes, this code will need to be updated as well to match it - and if for whatever reason we forget to do that, it will lead to problems. 

I think the desired behavior is:
- Use the value from values.yaml by default
- If a dev has specified a custom registry, use that instead.

Makefile
```diff
@@ -77,11 +77,8 @@ generate:
 	$(MAKE) gen-manifests
 	$(MAKE) fix-changed
 
-gen-manifests: bin/helm
```
caseydavenport, line 80: Doesn't changing this prevent me from running `make OPERATOR_VERSION=foo gen-manifests`?

charts/calico/values.yaml
caseydavenport: This is what our other values.yaml file looks like: https://github.com/projectcalico/calico/blob/master/charts/tigera-operator/values.yaml#L81-L84

```
tigeraOperator:
  image: tigera/operator
  version: master
  registry: quay.io
```

It looks like the `DEV_REGISTRIES` env var and the `release/` tooling use the word registry to describe `<REGISTRY>/<REPOSITORY>` but that's just incorrect.

We don't need to fix those in this PR, but we shouldn't propagate the incorrect version, especially because there is another counter example in this repository that does it the "right" way.

caseydavenport: calicoctl is also incorrect in the tigera-operator values.yaml, so it's clearly all just a mess:

https://github.com/projectcalico/calico/blob/a31851348d5d65b573a6165354c5eb5cdd9c9704/charts/tigera-operator/values.yaml#L85-L87

```
calicoctl:
  image: quay.io/calico/ctl
  tag: master
```

caseydavenport: So, fine to do as you have because it's all a mess. Just makes me very sad.

node/tests/k8st/infra/calico-kdd.yaml
```diff
@@ -562,7 +562,7 @@ spec:
         # It can be deleted if this is a fresh installation, or if you have already
         # upgraded to use calico-ipam.
         - name: upgrade-ipam
-          image: docker.io/calico/cni:latest-amd64
+          image: quay.io/calico/cni:latest-amd64
```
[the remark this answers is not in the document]
radTuti: this file is auto-generated. Default registry in manifests was switched from docker to quay hence the updates here

Makefile
radTuti: reverting those changes as they are not essential to this PR

release/pkg/manager/calico/manager.go
```diff
@@ -901,6 +902,11 @@ func (r *CalicoManager) generateManifests() error {
 	env := os.Environ()
 	env = append(env, fmt.Sprintf("CALICO_VERSION=%s", r.calicoVersion))
 	env = append(env, fmt.Sprintf("OPERATOR_VERSION=%s", r.operatorVersion))
+	env = append(env, fmt.Sprintf("OPERATOR_REGISTRY=%s", r.operatorRegistry))
+	env = append(env, fmt.Sprintf("OPERATOR_IMAGE=%s", r.operatorImage))
```
caseydavenport, line 906: I need to think about this a little bit more. I'm not very comfortable with the complexity this seems to be introducing. It's a lot more difficult to understand exactly where the image / registry used in a proper release is meant to come from.

I think this falls into the same category of duplicating what `values.yaml` is doing. I think we want `charts/tigera-operator/values.yaml` to be the source of truth for releases and hashreleases, and the CLI flags passed to the release tool are an optional override for dev releases.

caseydavenport (changes requested):
This is looking pretty good, I have two small-ish comments.

The main concern I have is that `values.yaml` is the default used when running `make gen-manifests` but is not used for the default when building a release using the release tool.

We seem to already be breaking this rule a little bit, I noticed by looking at the code here: https://github.com/projectcalico/calico/blob/a31851348d5d65b573a6165354c5eb5cdd9c9704/release/pkg/manager/calico/manager.go#L119-L120

So, we don't necessarily need to address this in this PR. . .

We do need to make sure we're not changing the API contract of `charts/tigera-operator/values.yaml` though - that's a using-facing API for the product so we need to be backwards compatible or we need to add upgrade steps for helm users (and I don't think this change is worth that).

charts/tigera-operator/templates/tigera-operator/02-tigera-operator.yaml
```diff
@@ -108,7 +108,7 @@ spec:
                 optional: true
         # Install any v3 API resources provided in the calico-resources ConfigMap.
         - name: create-initial-resources
-          image: {{.Values.calicoctl.image}}:{{.Values.calicoctl.tag}}
+          image: {{.Values.calicoctl.registry}}/{{.Values.calicoctl.image}}:{{.Values.calicoctl.tag}}
```
caseydavenport, line 111: This will need to be reverted as well in order to avoid breaking users on upgrade.

release/pkg/manager/calico/manager.go
radTuti: > It's a lot more difficult to understand exactly where the image / registry used in a proper release is meant to come from
this section is only used for hashreleases here: https://github.com/projectcalico/calico/blob/master/release/pkg/manager/calico/manager.go#L296-L316

> I think this falls into the same category of duplicating what values.yaml is doing
This section is always using values from the `release/` flags defaults. If we want to define SOT for variables, i reckon it will have to start there which IMO will go beyond the scope of this PR.

I will like us to have clear delineation for where SOT comes from for a lot of the variables used by the release tool esp WRT default values for flags that the tool consumes

caseydavenport: Yeah, I think it might have been a bug that I missed in previous reviews the source of truth was moved to `release/` instead of using the `values.yaml` file. 

To be clear, I'm not asking you to fix in this PR, but right not we're in an uncomfortable place where we have multiple sources of truth for different purposes (per-PR CI uses values.yaml, hashrelease and release use the `release/` package values).

caseydavenport (approved):

PR #9764, opened 2025-01-27
Support for Flow streaming in Goldmane

caseydavenport:
## Description

Adds basic streaming support to Goldmane, allowing clients to receive incremental Flow updates every 15s.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


goldmane/pkg/aggregator/aggregator.go
```diff
@@ -317,6 +322,10 @@ func (a *LogAggregator) rollover() time.Duration {
 		return nil
 	})
 
+	// Tell the stream manager that we've rolled over. This will trigger emission of any
+	// flows that are ready to be emitted to active stream clients.
+	a.streams.rollover(a.buckets, a.diachronics)
```
caseydavenport, line 327: Right now this is synchronous with the rest of the rollover processing - TBD if this is good enough. It has the nice effect of not needing to interact with core LogAggregator memory from another goroutine, but it does means we'll service every stream sequentially from the main loop.

We could also take a snapshot of the memory and send it via a channel, and run the stream manager on a second goroutine, which I believe is closer to how your original commit worked @Brian-McM 

Main reason I didn't just start with that approach is that the streams are tied to the rollover interval (we're emitting aggregated flows), whereas in your original PR we were siphoning off pre-aggregation Flows.

caseydavenport: Ultimately I don't want to over optimize this implementation at this stage - it's still not even clear what the final requirements for API and behavior will be. i.e., will we need the ability to start a stream at a particular point in time and catch-up from there? Will we need configurable aggregation windows? etc.

goldmane/pkg/aggregator/bucket_ring.go
```diff
@@ -335,6 +336,26 @@ func (r *BucketRing) FlowCollection(diachronics map[types.FlowKey]*types.Diachro
 	return flows
 }
 
+// RecentFlows returns the set of FlowKeys from the last N buckets, along with the start and end time of the window.
+// Note: RecentFlows does not include the bucket indicated by the headIndex (as that bucket is in the future), nor
+// the bucket before that (as that bucket is currently being filled).
+func (r *BucketRing) RecentFlows(n int) (set.Set[types.FlowKey], int64, int64) {
```
Brian-McM, line 342: I think it would be nice if we could just somehow pass back the DiachronichFlows that we're ultimately looking for. I think (please correct me if I'm wrong) that we wouldn't every be only interested in the keys, the ultimate goal is to get the flows.

This just seems like a situation where we can pass back a list of objects that can be used, even if the list is something like:

```
type DiachronicFlows struct {
  Key
  DiachronicFlow
  startTime time.Time
  endTime time.Time
}

func (d *DiachronicFlowss) Flows() []DiachronicFlow {
  // do stuff with time stamps and diachronic.
}

```

I'm also wondering if the DiachronichFlow can have a reference to the Key (somehow) so you don't need to pass around both. Any time a look up is done for Diachronic flows the key could be added at that point (and the implementation funneled), or it could be put in there on insert into the Ring (and use the unique package to dedup the FlowKey).

goldmane/pkg/aggregator/stream_manager.go
```diff
@@ -0,0 +1,137 @@
...
+	logrus.WithField("id", id).Debug("Closing stream")
+	close(s.flows)
+	delete(m.streams, id)
+}
+
+func (m *streamManager) rollover(ring *BucketRing, diachronics map[types.FlowKey]*types.DiachronicFlow) {
+	// Each stream may have requested aggregation across a different number of buckets, so the individual stream
+	// objects are responsible for tracking their own state - whether or not it is ready to emit data, and how many buckets
+	// back it wants to go.
+	logrus.Debug("Checking if any stream should emit data")
```
Brian-McM, line 107: What would you think about keeping the logic with the Ring, instead of requiring this coupling between the types?

You could just send the Diachronics on a channel to emit and let the ring figure out what ones it's sending.

We might not need to care about tracking the rollovers between streams, and just keep the iteration constant for all streams. For instance, if we're iterating on 6 buckets, and a stream is created on the 5th bucket, I think it would be fine to send back the 6 buckets worth of data even though it includes flows that were created before the stream started.

At least we could start with this simplish / consistent implementation and get a little more complex if we realise this isn't enough?

Brian-McM (changes requested):

goldmane/pkg/aggregator/stream_manager.go
caseydavenport: > For instance, if we're iterating on 6 buckets, and a stream is created on the 5th bucket, I think it would be fine to send back the 6 buckets worth of data even though it includes flows that were created before the stream started.

The problem is that, in theory, two streams need not have the same aggregation window. 

User 1 logs into the UI, says "Show me flows ever 15s" and user 2 logs in and says "Show me flows every minute" - those result in two separate Stream objects, each sending flows at a different rate and aggregating across different sets of buckets.

Brian-McM (approved):

PR #9769, opened 2025-01-29
Fix EBUSY errors from netlink operations.

fasaxc:
## Description

Hunch is that the change in https://github.com/vishvananda/netlink/pull/1018, will prevent EBUSY by draining the remaining messages from the kernel before returning.

Before https://github.com/vishvananda/netlink/pull/925, all messages would be read but the "interrupted" flag was ignored.  That PR made it so that the dump would return early, but that would leave some messages unconsumed.  There was logic elsewhere to discard these messages, but the kernel has a check for "dump in progress" at the start of a new dump and it returns EBUSY in that case:

```
int __netlink_dump_start(struct sock *ssk, struct sk_buff *skb,
			 const struct nlmsghdr *nlh,
			 struct netlink_dump_control *control)
{
...
	mutex_lock(&nlk->nl_cb_mutex);
	/* A dump is in progress... */
	if (nlk->cb_running) {
		ret = -EBUSY;
		goto error_unlock;
	}
```

I wasn't able to trace the exact path through the kernel code to verify when cb_running would be set, or that the error would be returned on the NLMSG_DONE as we see here but it seemed like a solid bet.

## Related issues/PRs

CORE-10999

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that netlink list operations could fail with "device or resource busy" errors when under load resulting in Felix restarting.
```

## Reminder for the reviewer

Other optional labels:


nelljerram (approved):

sridhartigera (approved):
LGTM

PR #9777, opened 2025-01-31
[RS-2319] Map OpenStack-derived policy to the "default" tier

nelljerram:
## Description

Since the merging of the multiple tier work, OpenStack-derived network policies are wrongly showing up as belonging to an "ossg" tier, whereas they should be in the default tier.  The "ossg" tier is being inferred because when our code translates an OpenStack security group to a corresponding network policy, it generates the name as "ossg.default."  followed by the security group ID.

## Related issues/PRs

https://tigera.atlassian.net/browse/RS-2319

## Release Note

```release-note
Fix: Map OpenStack-derived policy to the "default" tier, not "ossg".
```

### DIFF
--- a/libcalico-go/lib/names/policy.go
+++ b/libcalico-go/lib/names/policy.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 Tigera, Inc. All rights reserved.
+// Copyright (c) 2024-2025 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


mazdakn (approved):
LGTM.
Interesting, I did not know about openstack case. Maybe worth adding a comment to the code for the future references.

mazdakn:
@nelljerram I don't know about the execution path in open stack, just wondering if we need to add tests for other functions in the same file?

nelljerram:
@mazdakn I'm not ignoring your comments - but merging this now so that we can have fresh .debs built and then kick off the calico-test ST suite.  I'll consider your comments in a follow up.

nelljerram:
FTR: The CI failure was unrelated.

PR #9778, opened 2025-01-31
[RS-2319] Map OpenStack-derived policy to the "default" tier

nelljerram:
## Description

Since the merging of the multiple tier work, OpenStack-derived network policies are wrongly showing up as belonging to an "ossg" tier, whereas they should be in the default tier.  The "ossg" tier is being inferred because when our code translates an OpenStack security group to a corresponding network policy, it generates the name as "ossg.default."  followed by the security group ID.

(backport of #9777 to the v3.29 branch)

## Related issues/PRs

https://tigera.atlassian.net/browse/RS-2319

## Release Note

```release-note
Fix: Map OpenStack-derived policy to the "default" tier, not "ossg".
```

### DIFF
--- a/libcalico-go/lib/names/policy.go
+++ b/libcalico-go/lib/names/policy.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 Tigera, Inc. All rights reserved.
+// Copyright (c) 2024-2025 Tigera, Inc. All rights reserved.
 //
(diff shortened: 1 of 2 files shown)


mazdakn (approved):
LGTM.

nelljerram:
The CI failure looks entirely unrelated.

PR #9782, opened 2025-02-03
Markups following https://github.com/projectcalico/calico/pull/9777

nelljerram:


nelljerram:
@mazdakn Could you be more precise about what you had in mind with [this comment](https://github.com/projectcalico/calico/pull/9777#issuecomment-2628104767)?  I'm not sure what other functions would be useful to test with an OpenStack-derived policy name.

mazdakn:
@nelljerram the comment is about this file: https://github.com/projectcalico/calico/blob/master/libcalico-go/lib/names/policy_test.go
It seems there are more tests in there related to different functions. I was just asking if you need to add an entry for openstack security group name as well.

nelljerram:
> @nelljerram the comment is about this file: https://github.com/projectcalico/calico/blob/master/libcalico-go/lib/names/policy_test.go It seems there are more tests in there related to different functions. I was just asking if you need to add an entry for openstack security group name as well.

OK, thanks.  I don't think there are any other cases where this is needed.

mazdakn (approved):
LGTM

PR #9784, opened 2025-02-03
Goldmane API updates

caseydavenport:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


goldmane/proto/api.proto
```diff
@@ -47,15 +34,111 @@ message FlowRequest {
...
+  // Every AggregationInterval the server must send a FlowResult containing the aggregated data for that Flow from a
+  // time interval of width AggregationInterval.
+  //
+  // For a Flow that has continuous traffic, the server should send updates covering the range
+  // [now-2*AggregationInterval, now-AggregationInterval] so that the data is reasonably likely to be complete.
+  int64 aggregation_interval = 3;
+}
+
+// FlowResult wraps a Flow object with additional metadata.
+message FlowResult {
```
Brian-McM, line 71: Why the change to FlowResult?

goldmane/proto/api.proto
```diff
@@ -47,15 +34,111 @@ message FlowRequest {
...
+  // ID is an opaque integer value ID that can be used to identify a Flow, and is 1:1 with the FlowKey.
```
Brian-McM, line 72: What were the issues with adding this ID to the Flow and saying 0 means "not set"?

caseydavenport: A Flow is used on multiple APIs, including those that don't have a concept of an ID - e.g., the push API to Linseed and the collection API used by Felix.

This ID is only a concept on the aggregated List/Stream APIs, and I think it makes sense that those APIs may want a place to include additional metadata in the future anyway.

Brian-McM: I had some other thoughts on this, but realize it probably doesn't matter for this so I'm fine with it as is.

goldmane/proto/api.proto
```diff
@@ -1,36 +1,29 @@
...
+  rpc FilterHints(FilterHintsRequest) returns (stream FilterHint);
 }
 
-// FlowReceipt is a response from the server to a client after publishing a Flow.
-message FlowReceipt {}
-
-// FlowRequest defines a message to request a particular selection of aggregated Flow objects.
-message FlowRequest {
+// ListRequest defines a message to request a particular selection of aggregated Flow objects.
+message ListRequest {
```
Brian-McM, line 26: From a general services design POV, it might be worth it to prefix this with `Flow`. There could be other services with listing abilities.

goldmane/proto/api.proto
```diff
@@ -47,15 +40,142 @@ message FlowRequest {
...
+}
+
+// PolicyMatch defines criteria for matching one or more policy rules within a Flow's
+// policy trace.
+message PolicyMatch {
+  PolicyKind Kind = 1;
+  string Tier = 2;
+  string Namespace = 3;
+  string Name = 4;
+  string Action = 5;
```
Brian-McM, line 137: Just to confirm, to see if a flow was allowed or denied you need to traverse the policies?

Does this always align with if the flow was actually denied or allowed, like with staged policy?

caseydavenport: Nope, there's a top-level Action field in the FlowKey as well that contains the final action and is what this matches against. 

I'm not actually sure what the top-level Action field will display for staged policy actions - I suspect it only shows the "enforced" action and for staged you do need to look at the pending_policies array.

goldmane/pkg/aggregator/bucketing/bucket_ring.go
```diff
@@ -25,7 +25,27 @@ import (
 )
 
 type StreamReceiver interface {
-	ReceiveFlow(*types.DiachronicFlow, int64, int64)
+	Receive(FlowBuilder)
+}
+
+type FlowBuilder interface {
```
Brian-McM, line 31: Could you add a comment or two about the purpose of this? Just stating something like it's used so we don't take up memory building the flow until we need to?

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -195,7 +199,9 @@ func (a *LogAggregator) Run(startTime int64) {
 		case req := <-a.listRequests:
 			req.respCh <- a.queryFlows(req.req)
 		case req := <-a.streamRequests:
-			req.respCh <- a.streams.register(req)
+			stream := a.streams.register(req)
+			req.respCh <- stream
```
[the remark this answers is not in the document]
caseydavenport: Yeah, exactly right. Maybe I should add a comment to that extent, but since this is all happening on the main routine there's no chance of us rolling over while doing the backfill.

goldmane/proto/api.proto
Brian-McM: Ahh right, the key.

goldmane/proto/api.proto
```diff
@@ -47,15 +40,142 @@ message FlowRequest {
...
+  // Note that this ID is not valid across server restarts. Its primary use-case is for correlating FlowResult
+  // updates from a Stream request.
+  int64 id = 1;
+
+  // The Flow object itself.
+  Flow flow = 2;
+}
+
+// Filter defines criteria for selecting a set of Flows based on their parameters.
+message Filter {
```
Brian-McM, line 115: I think we probably want to add action here so we can filter by the actual enforced action?

goldmane/pkg/aggregator/bucketing/bucket_ring.go
Brian-McM: I actually really like this though, I think it's a good practice of not consuming more memory than then needed until we actually need it.

PR #9785, opened 2025-02-03
Node driver registrar update

MichalFupso:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/pod2daemon/Makefile
+++ b/pod2daemon/Makefile
@@ -78,7 +78,7 @@ $(BINDIR)/csi-driver-%: $(SRC_FILES)
 	$(call build_binary, csidriver/main.go, $@)
 
 UPSTREAM_REGISTRAR_PROJECT ?= kubernetes-csi/$(REGISTRAR_IMAGE)
-UPSTREAM_REGISTRAR_TAG     ?= 9b41324b20f609bce4dda9e50cc484c4ef4f5949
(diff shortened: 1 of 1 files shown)


matthewdupre (approved):
Change looks fine, but could you please add a comment saying what this is?  Maybe "// https://github.com/kubernetes-csi/node-driver-registrar/releases/tag/v2.13.0" or something?

That makes me wonder if this needs to be a commit.  a498... isn't a particular tagged release or anything so there's no better way to refer to it, but 9b4 is a tagged release so can we just say v2.13.0 instead?  Hopefully that will "just work", and removes the need for the clarifying comment?

pod2daemon/Makefile
```diff
@@ -78,7 +78,7 @@ $(BINDIR)/csi-driver-%: $(SRC_FILES)
 	$(call build_binary, csidriver/main.go, $@)
 
 UPSTREAM_REGISTRAR_PROJECT ?= kubernetes-csi/$(REGISTRAR_IMAGE)
-UPSTREAM_REGISTRAR_TAG     ?= a498c31b88cd31f8ea7bafd0d715a75f9939fa1f
+UPSTREAM_REGISTRAR_TAG     ?= 9b41324b20f609bce4dda9e50cc484c4ef4f5949 #https://github.com/kubernetes-csi/node-driver-registrar/releases/tag/v2.13.0
```
Behnam-Shobiri, line 81: I think we can just mention the version directly (instead of the hash) and it will work. We only start using hash since upstream didn't have all the CVE fixes that we wanted.

PR #9791, opened 2025-02-04
Add backend API for open source "Whisker" UI

Brian-McM:
This commit adds the backend for the open source Whisker UI. Notes:
- Adds /flows and /flows/_stream endpoints for listing and streaming flows
- Adds lib/httpapimachinery module for portable API / BFF utilities

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


whisker-backend/Makefile
```diff
@@ -0,0 +1,80 @@
...
+release-build: .release-$(VERSION).created
+.release-$(VERSION).created:
+	$(MAKE) clean image-all RELEASE=true
+	$(MAKE) retag-build-images-with-registries RELEASE=true IMAGETAG=$(VERSION)
+	$(MAKE) retag-build-images-with-registries RELEASE=true IMAGETAG=latest
+
+release-verify: release-prereqs
+	@echo "Nothing to do"
+
+release-publish: release-prereqs release-verify .release-$(VERSION).published
```
caseydavenport, line 73:
```suggestion
release-publish: release-prereqs .release-$(VERSION).published
```

whisker-backend/Makefile
```diff
@@ -0,0 +1,80 @@
...
+.release-$(VERSION).published:
+	$(MAKE) push-images-to-registries push-manifests IMAGETAG=$(VERSION) RELEASE=$(RELEASE) CONFIRM=$(CONFIRM)
+	$(MAKE) FIPS=true push-images-to-registries push-manifests IMAGETAG=$(VERSION)-fips RELEASE=$(RELEASE) CONFIRM=$(CONFIRM)
```
caseydavenport, line 76:
```suggestion
```

whisker-backend/Makefile
```diff
@@ -0,0 +1,80 @@
...
+
+	# Push Windows images.
+	$(MAKE) release-windows IMAGETAG=$(VERSION) CONFIRM=$(CONFIRM)
```
caseydavenport, line 79:
```suggestion
```

Don't think we're building Windows versions of this right now

lib/httpapimachinery/pkg/server/adaptors/gorilla/route.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"github.com/gorilla/mux"
+	"net/http"
+
+	"github.com/projectcalico/calico/lib/httpapimachinery/pkg/handler"
+	"github.com/projectcalico/calico/lib/httpapimachinery/pkg/server"
+)
+
+type Router interface {
```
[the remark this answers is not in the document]
caseydavenport: I think this interface might only be used internally? Can it just be private? 

I think I might be getting a little bit confused by the duplicate interfaces named `Router` in this package as well as in the `server` package - seems like they both represent a subset of the capabilities that a `type router struct` exposes.

assuming I am understanding this correctly, which is not a given, I might do something like this:

```go
// router is a local interface that defines the set of methods needed to register APIs with handlers.
type router interface {
}
```

and then in server:

```go
// Registrar is an interface that provides methods for registering APIs for use.
type Registrar interface {
  RegisterAPIs(...) http.Handler
}
```

i.e., naming the interfaces after the set of functions they expose rather than the underlying `type struct` that implements them.

whisker-backend/pkg/apis/v1/flows.go
```diff
@@ -0,0 +1,48 @@
+package v1
+
+import "time"
+
+const (
+	FlowsPath       = sep + "flows"
+	FlowsStreamPath = FlowsPath + sep + "_stream"
```
caseydavenport, line 7: what does the `_` signify our APIs?

lib/httpapimachinery/pkg/handler/stream.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to send event: %w", err)
+	}
+
+	w.Flush()
+	return nil
+}
+
+// Data encodes the given object as json and flushes it down the http stream.
+func (w *jsonEventStreamWriter[E]) Data(e E) error {
```
caseydavenport, line 42: Maybe `WriteData`? or just `Write`? e.g., `stream.Write(data)` - Might just be me, but `Data` doesn't imply an action is being take - rather I'd expect it to return me some Data.

whisker-backend/pkg/handlers/v1/flows.go
```diff
@@ -0,0 +1,114 @@
...
+
+func protoToFlow(flow *proto.Flow) whiskerv1.FlowResponse {
+	return whiskerv1.FlowResponse{
+		StartTime: time.Unix(flow.StartTime, 0),
+		EndTime:   time.Unix(flow.EndTime, 0),
+		Action:    flow.Key.Action,
+
+		SourceName:      flow.Key.SourceName,
+		SourceNamespace: flow.Key.SourceNamespace,
+		SourceLabels:    strings.Join(flow.SourceLabels, " | "),
```
caseydavenport, line 100: Is a single string of Labels really what the UI wants instead of a json list?

whisker-backend/test/cert.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
caseydavenport, line 1: If possible, I think it would be better to generate these on-demand as part of running the tests that need them (and removing on `make clean`) rather than checking them in. Not strictly required, but I think it's a better model and raises fewer eyebrows when it comes to security reviews of our repository.

caseydavenport (commented):
lookin' good! Mostly nits and questions so far

lib/httpapimachinery/pkg/context/context.go
```diff
@@ -0,0 +1,59 @@
...
+	"net/http"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/google/uuid"
+
+	"github.com/projectcalico/calico/lib/httpapimachinery/pkg/header"
+)
+
+// Key is used to ensure we don't have collisions with existing keys.
```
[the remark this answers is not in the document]
Brian-McM: Yes that's exactly it.

whisker-backend/pkg/apis/v1/flows.go
Brian-McM: Generally the API for list and get would be `/resources` and `/resources/{id}`, however, `_stream` wouldn't signal a resource ID. For these cases I like to prefix the path with an underscore so that it would likely never conflict with any sort of resource identifier.

whisker-backend/test/cert.key
Brian-McM: I didn't actually mean to commit this, I'd rather create these on the fly (with a golang library).

whisker-backend/pkg/handlers/v1/flows.go
Brian-McM: No I wouldn't think so, this was just conforming to what Ronan already had.

lib/httpapimachinery/pkg/handler/json.go
```diff
@@ -0,0 +1,64 @@
+package handler
+
+import (
+	"encoding/json"
+	"net/http"
+
+	_ctx "github.com/projectcalico/calico/lib/httpapimachinery/pkg/context"
```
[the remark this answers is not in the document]
Brian-McM: I'd be down for something like `apictx`, or `httpctx`. I don't really want to just use context because I do want to ensure it isn't confused with the base type.

lib/httpapimachinery/pkg/handler/json_test.go
```diff
@@ -0,0 +1,55 @@
...
+	. "github.com/onsi/gomega"
+
+	_ctx "github.com/projectcalico/calico/lib/httpapimachinery/pkg/context"
+	"github.com/projectcalico/calico/lib/httpapimachinery/pkg/handler"
+)
+
+func TestJSONListResponse(t *testing.T) {
+	setupTest(t)
+
+	type Request struct {
```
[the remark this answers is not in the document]
Brian-McM: I'll see if I can unexported it but the fields at least need to be exported for the decoders to work.

lib/httpapimachinery/pkg/handler/response.go
```diff
@@ -0,0 +1,82 @@
+package handler
+
+type ErrorResponse struct {
+	ErrMsg string `json:"error"`
```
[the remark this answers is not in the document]
Brian-McM: Ya I could do that.

whisker-backend/pkg/apis/v1/common.go
```diff
@@ -0,0 +1,7 @@
+package v1
+
+const (
+	sep  = "/"
+	sBrc = "{"
```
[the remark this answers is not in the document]
Brian-McM: This is whisker you're commenting on, did you mean in the example?

I found the constants helped when you have a lot of APIs, I thought I'd just leave it for the users of the API to define instead of making them public in the lib package. I'll just put this in the flow file and remove this though.

whisker-backend/pkg/apis/v1/flows.go
caseydavenport: Kubernetes API uses a pattern like this:

```
GET /resources?watch=true
```

Seems like params might make more sense here instead of using a special value `{id}`?

lib/httpapimachinery/pkg/handler/json.go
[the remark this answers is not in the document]
caseydavenport: I'd still prefer spelling out `context` instead of `ctx` - `ctx` is too commonly used to refer to an instance of a Context object.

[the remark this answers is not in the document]
Brian-McM: Ok I'm good with that, good point.

caseydavenport (approved):

PR #9795, opened 2025-02-04
Enable Debugging of Felix FV Tests in GoLand IDE

skoryk-oleksandr:
## Description

This PR updates the `felix/README.md` with instructions on debugging **Felix FV tests** in **GoLand** IDE. Additionally, it includes a shell script to build the necessary resources for FV tests, ensuring a smoother debugging experience.


matthewdupre (approved):
I don't use goland so couldn't test it, but the change is safe and hopefully useful to people

felix/README.md
```diff
@@ -131,6 +131,48 @@ ginkgo watch -r
...
+
+    - **Name:** Provide a meaningful name for the configuration, e.g., Debug Felix FV Tests.
+
+    - **Test kind:** `package`
+
+    - **Package Path:** set to `github.com/projectcalico/calico/felix/fv`
+
+    - **Working Directory:** Set the working directory to: `/{path to the project root}/felix/fv`
+
+    - **Environment:** Copy the following Env Vars to the field: `FV_BINARY=bin/calico-felix-amd64;ACK_GINKGO_DEPRECATIONS=1.16.5;CERTS_PATH=/{path to the project root}/hack/test/certs;FV_ETCDIMAGE=quay.io/coreos/etcd:v3.5.6-amd64;FV_K8SIMAGE=calico/go-build:1.23.5-llvm18.1.8-k8s1.31.4;PRIVATE_KEY=/{path to the project root}/felix/fv/private.key;FV_CWLOGDIR=/{path to the project root}/felix/fv/cwlogs`
```
matthewdupre, line 156: It would be nice if we didn't have to specify things like the FV_K8SIMAGE (which updates from time to time), but if it's not easy that's OK.

skoryk-oleksandr: Removed FV_K8SIMAGE param. It works without it.

fasaxc: I think it's important to use the same image that the makefile would use.  Just because it works, doesn't mean it's correct (you might pick up an old image and get false sense of confidence in some code that doesn't work on a newer k8s)

felix/build_fv.sh
```diff
@@ -0,0 +1,20 @@
+# Script Name: build_and_test.sh
```
coutinhop, line 1: nit: shouldn't this be the same as the filename?

TBH I feel conflicted about adding this script, it doesn't seem add much value, couldn't the commands in it just be in the debugging guide in the README?

coutinhop: (just saw Shaun's comment about `make fv-prereqs`, that's much better IMHO )

PR #9744, opened 2025-01-22
[CORE-10973] Add QoSControls to WorkloadEndpoint

coutinhop:
Parsed from annotations:
- qos.projectcalico.org/ingressBandwidth:
- qos.projectcalico.org/egressBandwidth:
- qos.projectcalico.org/ingressBurst:
- qos.projectcalico.org/egressBurst:
- qos.projectcalico.org/ingressPacketRate:
- qos.projectcalico.org/egressPacketRate:
- qos.projectcalico.org/ingressConnections:
- qos.projectcalico.org/egressConnections:

Also honor the k8s bandwidth plugin annotations if present (with native Calico annotations taking precedence):
kubernetes.io/ingress-bandwidth:
kubernetes.io/egress-bandwidth:

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #9811, opened 2025-02-07
[RS-2320] Add QoS support in OpenStack code

nelljerram:
Add QoS support in OpenStack code

This is the OpenStack-specific coding that matches up with #9744 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/common/config.py
+++ b/networking-calico/networking_calico/common/config.py
@@ -48,6 +48,12 @@ SHARED_OPTS = [
                help="When in a multi-region OpenStack deployment, a unique "
                     "name for the region that this node (controller or "
                     "compute) belongs to."),
+    # Max connection options, in advance of max connection support being added
+    # properly to the Neutron API.
+    cfg.IntOpt('max_ingress_connections_per_port', default=0,
+               help="If non-zero, a maximum number of ingress connections to impose on each port."),
+    cfg.IntOpt('max_egress_connections_per_port', default=0,
+               help="If non-zero, a maximum number of egress connections to impose on each port."),
 ]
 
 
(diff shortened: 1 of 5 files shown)


networking-calico/networking_calico/common/config.py
```diff
@@ -48,6 +48,12 @@
                help="When in a multi-region OpenStack deployment, a unique "
                     "name for the region that this node (controller or "
                     "compute) belongs to."),
+    # Max connection options, in advance of max connection support being added
+    # properly to the Neutron API.
+    cfg.IntOpt('max_ingress_connections_per_port', default=0,
+               help="If non-zero, a maximum number of ingress connections to impose on each port."),
+    cfg.IntOpt('max_egress_connections_per_port', default=0,
+               help="If non-zero, a maximum number of egress connections to impose on each port."),
```
coutinhop, line 56: apologies for my lack of knowledge in this, but is there a need for a calico configuration (say, in FelixConfig) analogous to this? how exactly is this set by a user?

coutinhop (approved):

networking-calico/networking_calico/common/config.py
nelljerram: So this would be set in an `/etc/neutron/neutron.conf` file on the controller nodes, as:
```
...
[calico]
max_ingress_connections_per_port = 100
max_egress_connections_per_port = 100
...
```
I don't think it makes sense as a FelixConfig setting, because it's really just a stopgap for the Neutron DB currently missing these fields, and so the Neutron driver is the right place for temporary configuration to fill that gap.

If we think about a possible role for FelixConfig independently of OpenStack, then the question would be: given that we now support optional pod annotations for QoS, do we also want FelixConfig fields to set defaults for pods that don't have those annotations?  It's certainly conceivable, but I haven't heard anyone saying that is actually required.  If it _was_ required, I assume it would be required equally for all of the possible QoS fields.

WDYT, does that all make sense?

coutinhop: I get it now, thanks @nelljerram! I agree it only makes sense on the neutron config, as far as I can tell there is no requirement to set defaults when unset. Thanks once again!

PR #9813, opened 2025-02-07
[release tool] hashrelease promotion pipeline

radTuti:
## Description

This adds checking that "Push/Publish ___ images" jobs are done before proceeding with a hashrelease promotion. 

Additionally removed defaulting building & publishing images for task triggered hashrelease as Semaphore support setting variables there.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


release/internal/ci/semaphore.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	logrus.WithField("count", expectedPromotionCount).Debug("expected number of image promotions")
+	parentPipelineID, err := fetchParentPipelineID(orgURL, pipelineID, token)
+	if err != nil {
+		return false, err
+	}
+	if parentPipelineID == "" {
+		logrus.Info("no parent pipeline found, skipping image promotions check")
+		return true, nil
+	}
```
danudey, line 168: If we run a workflow and the first pipeline run has failed, we may re-run the failed jobs. This creates a second pipeline in the workflow. Which one shows up as the parent that we're looking for? Does it depend on how the promotion is triggered? If we trigger the hashrelease promotion while viewing the 'first' pipeline does it break this?

release/internal/ci/semaphore.go
```diff
@@ -0,0 +1,197 @@
...
+		"actual":   promotionsCount,
+	}).Debug("number of image promotions")
+	if promotionsCount < expectedPromotionCount {
+		return false, fmt.Errorf("number of promotions do not match: expected %d, got %d", expectedPromotionCount, promotionsCount)
+	}
+	for _, promotion := range promotions {
+		if promotion.Status != passed {
+			logrus.WithField("promotion", promotion.Name).Error("triggering promotion failed")
+			return false, fmt.Errorf("triggering %q promotion failed, cannot check pipeline result", promotion.Name)
+		}
```
danudey, line 186: If the above doesn't break the promotions count, does the failed promotion (which we do have a successful version of) cause things to fail at this point?

It might be more reliable to do something like:

1. Get the names of expected promotions from `03-promotions.yml` (not just the count)
2. Get the list of promotions which were run
3. Print an error for 'This promotion was not run' or 'This promotion failed to run' for each expected promotion without a corresponding successful promotion.

```
The following image promotions are missing; please run them and try again:
* Push apiserver images
* Push cni-plugin images
* ...

The following image promotions failed:
* Push kube-controllers images:  <url to the failed pipeline>
```

release/internal/ci/semaphore.go
radTuti: by re-run failed jobs, I assume you mean a partial run of the existing pipeline. In that case it does not impact this as the primary pipeline is still considered the parent pipeline for a promotion.

danudey (approved):

PR #9821, opened 2025-02-11
Enable Debugging of Felix FV Tests in GoLand IDE

skoryk-oleksandr:
## Description
This PR brings changes that allow to debug felix fv tests in GoLand ide. It adds new Makefile target called `build-fv-env`.

This target performs the following steps:
**1.** Builds necessary resources required for running FV tests.
**2.** Ensures that all FV test prerequisites are met.
**3.** Exports required environment variables into a `.env` file.

Env vars from prepared `.env` file are being loaded to runtime environment using `godotenv.Load("../.env")` command. After that all env vars needed for FV tests run are available in the runtime. 

This PR also add the instructions how to configure GoLand runtime configuration.


felix/.gitignore
```diff
@@ -70,3 +70,6 @@ htmlcov/
 
 # produced as part of openstack packaging build
 debian/files
+
+#Env Configuration
+.env
```
stevegaossou, line 75: Minor: I think we usually have a next line at the end of the file.

felix/Makefile
```diff
@@ -315,6 +315,24 @@ fv-prereqs: $(REMOTE_DEPS) \
...
+# build-fv-env: Prepare the environment for FV (Function Verification) tests
+#
+# This target performs the following steps:
+# 1. Builds necessary resources required for running FV tests.
+# 2. Ensures that all FV test prerequisites are met.
+# 3. Exports required environment variables into a `.env` file.
+#
+# The `.env` file is used by the Go test runtime to load the necessary
+# configuration into the environment.
+build-fv-env: build fv-prereqs
```
stevegaossou, line 327: The ENV set up in this target is specifically for Goland, correct? Or is it useful .env useful outside of the IDE use case? 

If it's specific for Goland, I think it would be helpful for someone reading the Makefile to have target name mention "goland" in it.

felix/fv/infrastructure/felix.go
```diff
@@ -162,6 +163,12 @@ func RunFelix(infra DatastoreInfra, id int, options TopologyOptions) *Felix {
 	if fvBin == "" {
 		fvBin = fmt.Sprintf("bin/calico-felix-%s", arch)
 	}
+
+	if cwLogDir == "" {
+		var wDir, err = os.Getwd()
+		Expect(err).NotTo(HaveOccurred())
+		cwLogDir = filepath.Join(wDir, "/cwlogs")
```
stevegaossou, line 170: Could you clarify what the `cwLogDir` directory is used for and why we need this for the debugging? Is it related to logs generated through Goland?

stevegaossou (commented):
Added some comments.

felix/fv/infrastructure/felix.go
skoryk-oleksandr: `cwLogDir is not related specifically to the Goland`. When you run tests via Makefile this variable is being set here https://github.com/skoryk-oleksandr/calico/blob/53cc83b4b6f4b01c796bd1b75773ed301bda10da/felix/fv/run-batches#L26
`export FV_CWLOGDIR=`pwd`/cwlogs` which will be `/{path to the project root}/felix/fv/cwlogs`
So basically I added logic that checks if this var is not set it use default value.

felix/Makefile
skoryk-oleksandr: This target can be used for Goland and VS Code as well. It prepares environment for running tests in IDE. `.env` file can be used for VS Code as well.

felix/fv/infrastructure/felix.go
stevegaossou: Okay, I see. So you're basically providing a failsafe (default) value in case `FV_CWLOGDIR` is empty.

felix/.gitignore
```diff
@@ -70,3 +70,6 @@ htmlcov/
 
 # produced as part of openstack packaging build
 debian/files
+
+#Env Configuration
```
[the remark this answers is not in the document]
skoryk-oleksandr: Fixed

felix/.gitignore
skoryk-oleksandr: fixed

felix/fv/utils/utils.go
```diff
@@ -53,6 +54,10 @@ type EnvConfig struct {
 var Config EnvConfig
 
 func init() {
+	envErr := godotenv.Load("../.env")
+	if envErr != nil {
+		log.Warningf("Error loading .env file! Err = %s", envErr)
```
matthewdupre, line 59: The .env file isn't mandatory right?  In which case I think this is more like a Debug log

matthewdupre (approved):
Approved subject to log level change; I don't need to rereview

PR #9827, opened 2025-02-12
Update host end point flow log

mazdakn:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/calc/endpoint_lookup_cache.go
```diff
@@ -124,12 +124,13 @@ type TierData struct {
...
-func (e *EndpointData) IsHostEndpoint() (isHep bool) {
+func (e *EndpointData) IsHostEndpoint() bool {
 	switch e.Key.(type) {
 	case model.HostEndpointKey:
-		isHep = true
+		return true
+	default:
+		return false
 	}
-	return
```
matthewdupre, line 132: Wow, I didn't know you could even do this...

felix/collector/flowlog/reporter.go
```diff
@@ -154,6 +154,9 @@ func (r *FlowLogReporter) Report(u interface{}) error {
 		return fmt.Errorf("invalid metric update")
 	}
 	log.Debug("Flow Logs Report got Metric Update")
+	if mu.SrcEp != nil && mu.SrcEp.IsHostEndpoint() {
```
matthewdupre, line 157: I think this check is for when reporter == src (unless the SrcEp field doesn't match the eventual source fields in the flow log).  We need to skip if dest is HEP when reporter == dst as well.

matthewdupre (changes requested):
I think this change could be made earlier, maybe around https://github.com/projectcalico/calico/blob/master/felix/collector/collector.go#L297 which I think would save some resources inside Felix.  PTAL and let me know if you think that would be a better location.

felix/collector/flowlog/reporter.go
mazdakn: If we do the same thing for dest reporter, then we will remove all hep logs. I thought it was said to only remove source reporter!

matthewdupre: The spec is:
- For source reporter, report only flow logs where the source is a pod
- For dest reporter, report only flow logs where the dest is a pod
I've been wording it carefully because it's quite confusing and there are a lot of cases (bear in mind one Felix could be either or both of the source and dest reporter).

mazdakn: Ok, essentially the idea is to remove HEP reporter either as source or dest reporter.

mazdakn: btw, the proposed changed is basically how it's done in the private repo.

mazdakn:
@matthewdupre I tried to use your proposal, but removing that information there, basically makes the collector to skip the log. Which is different from what we want. We don't want to ignore the log completely, but rather to mask the information.

We may be able to achieve the same thing earlier, in Collector, but might take more time. I suggest to use this solution as it is for the moment since we take the same approach in enterprise. I prefer to apply the change in same place for sake of merge conflicts: https://github.com/tigera/calico-private/blob/aa967f4f5f3da6abbcb5f160cf0257507d04004f/felix/collector/flowlog/reporter.go#L177-L184

Later I can work on improvements. WDYT?

felix/collector/collector.go
```diff
@@ -306,7 +306,13 @@ func (c *collector) getDataAndUpdateEndpoints(t tuple.Tuple, expired bool, packe
 	dstEp := c.lookupEndpoint(t.Src, t.Dst)
 	dstEpIsNotLocal := dstEp == nil || !dstEp.IsLocal
 
-	if !okData {
+	// Ignore HEP reporters.
+	if (srcEp != nil && srcEp.IsHostEndpoint()) ||
```
matthewdupre, line 310: add && srcEp.IsLocal inside the brackets; and similarly && dstEp.IsLocal inside the brackets on the following line.  (Or any refactoring of that logic you prefer).  You can also move this all inside if !exists, which will skip doing the check on non-first updates.

We want to include all flow logs where the reporter side is a pod, so they can have an endpoint (either one) be a hostendpoint

felix/collector/collector.go
```diff
@@ -306,12 +306,18 @@ func (c *collector) getDataAndUpdateEndpoints(t tuple.Tuple, expired bool, packe
...
 
-	if !okData {
+	if !exists {
 		// For new entries, check that at least one of the endpoints is local.
 		if srcEpIsNotLocal && dstEpIsNotLocal {
 			return nil
 		}
 
+		// Ignore HEP reporters.
+		if (srcEp != nil && srcEp.IsLocal && srcEp.IsHostEndpoint()) ||
```
matthewdupre, line 316: I think this change is right.  I'm wondering if it should also be reflected in line 338 though?  I'm not totally sure what the possible circumstances are for an endpoint to change - maybe it's a timing thing on deletion?

Thinking through it, the only problematic case is if somehow a pod could turn into a HEP.  I don't think that's actually possible though.

matthewdupre (approved):

felix/collector/collector.go
mazdakn: For an execution to reach line 338, first the endpoint should exist in the ep map, by reaching line 323, to execute `updateEpStatsCache()`, and then in the next iteration it might reach 338. 

So even, if endpoint change is possible, collector will first need to add it to the ep map first.

PR #9830, opened 2025-02-13
Add new pre-flight check to verify go.mod.

fasaxc:
## Description

- Check that it is listable (as needed by tools such as goland).
- Check it is "go mod tidied"
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,11 +37,15 @@ clean:
 	rm -rf ./bin
 
 ci-preflight-checks:
+	$(MAKE) check-go-mod
 	$(MAKE) check-dockerfiles
 	$(MAKE) check-language
 	$(MAKE) generate
 	$(MAKE) fix-all
 	$(MAKE) check-dirty
+
+check-go-mod:
+	$(DOCKER_GO_BUILD) ./hack/check-go-mod.sh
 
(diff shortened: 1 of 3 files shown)


nelljerram:
Could we look instead at configuring Goland (and other editors) to use the relevant go-build container for any `go ...` operations that they require?  ISTM that that would really be the right thing.

fasaxc:
It runs in the go-build container so there's no new requirement to have go installed.  Of course, if you're `go mod  tidy`ing with your local go install and it is different to the one in go-build then you might check in something that doesn't pass.  But that's one of the things I want to catch(!) that you['ve let your env drift from the one in go-build.

fasaxc:
FWIW, go now downloads the "right" version of the toolchain from your go.mod anyway, so it's actually pretty clever and rarely hits that problem.

nelljerram (approved):
I'm sorry, I completely missed the `$(DOCKER_GO_BUILD)` in the `check-go-mod` target.  So this is all fine then.  Sorry for misreading!

PR #9839, opened 2025-02-13
Remove dataplane rules for staged policies

mazdakn:
## Description

Staged policies are mainly used with flow log to observe the effect of a policy (staged ones). In flow logs, we keep two policy traces:

- **Pending**: policies that will be applied on a flow when staged policies are enforced. We use policy store (in user space) for populating pending traces.
- **Enforced**: policies that are being applied or enforced on a flow. Staged policies have no affect on this trace.

 As such, it's possible to skip programming staged policies in data plane since the value they provide is not used anymore and was replaced by evaluating the "pending/staged" policies in userspace.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/rules/endpoints.go
```diff
@@ -510,6 +511,10 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
 				if polGroup.ShouldBeInlined() {
 					// Group is too small to have its own chain.
 					for _, p := range polGroup.PolicyNames {
+						if model.PolicyIsStaged(p) {
+							logrus.Debugf("Skipping staged policy %v", p)
```
caseydavenport, line 515:
```suggestion
							logrus.Debugf("Skip programming staged policy %v", p)
```

caseydavenport (approved):
policy rule changes look legit :+1:

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -41,6 +41,7 @@ import (
 	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	"github.com/projectcalico/api/pkg/lib/numorstring"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
```
tomastigera, line 44: we usually import it this way

```suggestion
	log "github.com/sirupsen/logrus"
```

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -2968,11 +2968,11 @@ func (m *bpfEndpointManager) extractTiers(tiers []*proto.TierInfo, direction Pol
...
 			for i, polName := range directionalPols {
-				staged := model.PolicyIsStaged(polName)
-
-				if !staged {
-					stagedOnly = false
+				if model.PolicyIsStaged(polName) {
+					logrus.Debugf("SKipping staged policy %v", polName)
+					continue
 				}
+				stagedOnly = false
```
tomastigera, line 2975: why is this always false? To make sure that if there is a tier with staged only policies, we still honor that tier somehow? Do we need to test for this case somehow? Are we testing it already?

mazdakn: If we reach this point, it means we are processing a none staged policy. So, it means this tier is not one with only staged only policies. As such `stagedOnly` is set to false.

This is similar to what it was before. The only change is to skip processing/programming staged policies.

mazdakn: The difference is that if we are handling a tier with only staged policies, the end of tier action would be different: 
```
if endTierDrop && !stagedOnly && tier.DefaultAction != string(apiv3.Pass) {
				polTier.EndRuleID = m.endOfTierDropID(dir, tier.Name)
				polTier.EndAction = polprog.TierEndDeny
			} else {
				polTier.EndRuleID = m.endOfTierPassID(dir, tier.Name)
				polTier.EndAction = polprog.TierEndPass
			}
```

mazdakn: We have a FV for this: https://github.com/projectcalico/calico/blob/3185239e17c50f85d54d48a2abbdb37f60d0968c/felix/fv/tiered_policy_test.go#L377

mazdakn: btw, this change was made with introducing tier's default action: https://github.com/projectcalico/calico/pull/9232

felix/dataplane/linux/bpf_ep_mgr.go
caseydavenport: There are other packages actually named `log`. It's convenient to use the actual package name for things like goimports automatically importing the right package as well.

felix/rules/endpoints.go
```diff
@@ -744,7 +749,13 @@ func (g *PolicyGroup) ChainName() string {
 }
 
 func (g *PolicyGroup) ShouldBeInlined() bool {
-	return len(g.PolicyNames) <= 1
+	var count int
+	for _, name := range g.PolicyNames {
```
dimitri-nicolo, line 753: No need to finish counting all of the SNPs, return false if the count has exceeded the max value.

```
    count := 0
    for _, name := range g.PolicyNames {
        if !model.PolicyIsStaged(name) {
            count++
            if count > 1 {
                return false
            }
        }
    }
    return true
 ```

Comment (mazdakn) [reply felix/rules/endpoints.go:753]: Thanks. Nice hint. Will fix it.

Comment (mazdakn) [reply felix/rules/endpoints.go:753]: Fixed.

File: felix/rules/endpoints.go [hunk 5]
```diff
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
```
dimitri-nicolo, line 25: Since we are using logrus, can we replace the other calls "log", so that we don't import
 `"github.com/sirupsen/logrus"`
 and
`log "github.com/sirupsen/logrus"`

felix/fv/tiered_policy_test.go
```diff
@@ -342,8 +342,8 @@ var _ = infrastructure.DatastoreDescribe("connectivity tests with policy tiers _
 				}
 				return strings.Count(out0, "default.ep1-1-allow-all") > 0 &&
 					strings.Count(out1, "default.ep1-1-allow-all") == 0 &&
-					strings.Count(out0, "default/staged:default.np3-4") == 0 &&
-					strings.Count(out1, "default/staged:default.np3-4") > 0
+					strings.Count(out0, "staged") == 0 &&
```
dimitri-nicolo, line 345: Could you explain this change?

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: My question is that you set it to true, then you skip all staged policies to set it to false.

Shouldn't that for loop be like `for policies { if policy is not staged { stagedOnly = false ; break loop ; } }` because what you are looking for it that there is at least one non-staged policy, right? You write the kinda same thing, but for a tier with only non-staged policies you loop over all of them while you can bail out much sooner. This way it reads weird and perhaps therefore I am reading it wrong :thinking:

tomastigera (dismissed):

felix/fv/tiered_policy_test.go
mazdakn: It's a more strict check. Instead of checking specific staged rule is programmed or not, this checks that no staged policy should be available in dataplane. Based on this PR, the expectation is that no staged policy should appear in dataplane rules.

felix/dataplane/linux/bpf_ep_mgr.go
mazdakn: The loop performs two things:
1. Detecting if there is any non-staged policy in tier.
2. Programming non-staged policies.

If we break out of the loop when encountering non-staged policy, then we won't program non-staged policies.
The confusion is due to the fact that the loop is trying to perform two things in one loop.

felix/rules/policy.go
```diff
@@ -125,7 +118,7 @@ func (r *DefaultRuleRenderer) ProtoRulesToIptablesRules(
 			break
 		}
 	}
-	if len(chainComments) > 0 {
+	if len(chainComments) > 0 && !staged {
```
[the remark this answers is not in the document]
mazdakn: yeah. make sense. Updated.

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: Ohhhh I get it now, sorry, yeah, all good! I misread where you get out of the loop :heavy_check_mark:

tomastigera (commented):
LGTM

felix/dataplane/linux/bpf_ep_mgr.go
mazdakn: To make it more readable we need to break it into two loop, but it would be less efficient.

felix/rules/policy.go
```diff
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/sirupsen/logrus"
```
caseydavenport, line 21: Looks like you've got both imports here now

caseydavenport (approved):
One comment but otherwise LGTM still :D

felix/fv/tiered_policy_test.go
dimitri-nicolo: Thank you!

dimitri-nicolo (approved):
lgtm

felix/rules/policy.go
mazdakn: Fixed.

PR #9847, opened 2025-02-14
Add logging interface to decouple implementation from codebase

Brian-McM:
## Description

This commit adds custom logging interface to wrap our logger implementation (currently logrus) in the lib/std folder and move all logutilities into the lib/log folder. This replaces all references to logrus and libcalico/logutils in the project to use lib/std/log to ensure the logging interface includes everything we need. Some initial benefits are:
- Forces all uses of logging in the repo to use our agreed upon formatting / customizations. This cuts down on code to setup loggers and ensures optimizations are used everywhere.
- Ensures logrus dep conflicts don't happen between main module and lib modules since the only direct import of logrus would be in lib/log.
- Interface ensures tests can easily swap out implementation for mock implementation.
   

I've added testutils/assert library to help with writing tests in the lib modules, as we don't want to import ginkgo or testify in the lib modules as they might cause dependency issues with the main module.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


lib/std/log/formatters.go
```diff
@@ -0,0 +1,539 @@
...
+
+var preComputedInfixByLevelSyslog = make([]string, len(AllLevels))
+
+func init() {
+	for _, level := range AllLevels {
+		preComputedInfixByLevelSyslog[level] = strings.ToUpper(level.String()) + " "
+	}
+}
+
+// FormatForSyslog formats logs in a way tailored for syslogrus.  It avoids logging information that is
```
fasaxc, line 150: `syslogrus` :grin:

PR #9848, opened 2025-02-14
Program nflog rules only when flowlog is enabled

mazdakn:
## Description

To minimise the impact of flow logs on clusters, users should be able to decide if flow log rules are programmed in data plane or not. Currently, flow logs are always programmed. This PR makes it possible that flow log rules get programmed (nflog in iptables and log in nftables) only when flow log feature is enabled.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/policy_mgr.go
```diff
@@ -39,21 +39,38 @@ type policyManager struct {
...
 }
 
 type policyRenderer interface {
-	PolicyToIptablesChains(policyID *types.PolicyID, policy *proto.Policy, ipVersion uint8) []*generictables.Chain
-	ProfileToIptablesChains(profileID *types.ProfileID, policy *proto.Profile, ipVersion uint8) (inbound, outbound *generictables.Chain)
+	PolicyToIptablesChains(
+		policyID *types.PolicyID,
+		policy *proto.Policy,
+		ipVersion uint8,
+		flowLogEnabled bool,
```
mazdakn, line 50: It probably worth aggregating some of the arguments like `flowlogEnabled`, `staged`, `nft`, ... into a common option struct. But I'll try to do it in a separate PR.

felix/rules/policy_test.go
```diff
@@ -507,7 +715,49 @@ var _ = Describe("Protobuf rule to iptables rule conversion", func() {
...
+					"Policy staged:default.foo ingress",
+				},
+			}))
+
+			outbound := chains[1].Rules
+			Expect(outbound).To(HaveLen(1))
+			Expect(outbound[0]).To(Equal(generictables.Rule{
+				Comment: []string{
+					"Policy staged:default.foo egress",
+				},
```
mazdakn, line 735: It seems we always add a rule with comment even when the generated rule is empty: https://github.com/projectcalico/calico/blob/bc844f6f8213d17889d1e55cd1804d0e3f7f7f2d/felix/rules/policy.go#L128

felix/dataplane/driver.go
```diff
@@ -394,7 +394,6 @@ func StartDataplaneDriver(
 			RouteTableManager:                  routeTableIndexAllocator,
 			MTUIfacePattern:                    configParams.MTUIfacePattern,
 			BPFExcludeCIDRsFromNAT:             configParams.BPFExcludeCIDRsFromNAT,
-			NfNetlinkBufSize:                   configParams.NfNetlinkBufSize,
```
tomastigera, line 397: This makes sense, but will also create conflict on merge ... :man_shrugging:

felix/config/config_params.go
```diff
@@ -537,6 +537,10 @@ func (config *Config) TableRefreshInterval() time.Duration {
 	return config.IptablesRefreshInterval
 }
 
+func (config *Config) FlowLogEnabled() bool {
+	return config.FlowLogsGoldmaneServer != ""
```
tomastigera, line 541: Is flowlogS or flowlog_ the prefered form? I'd think the plural is the desired one.

felix/dataplane/linux/policy_mgr.go
tomastigera: Since the policy rendered is an object, it would be greate to be able to create it with these things preset on the struct. But I guess we use the same renderer (which was meant to be plugable) and actually only call the funcs hangging off it :-/

felix/fv/flow_logs_goldmane_test.go
```diff
@@ -533,8 +523,82 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ goldmane flow log tests", [
...
+		})
+
+		It("should not program nflog rules", func() {
+			Eventually(nflogRulesProgrammed, "10s", "1s").Should(BeFalse(),
+				"Unexpected nflog rules to appear on the correct felix instances")
+			Consistently(nflogRulesProgrammed, "10s", "1s").Should(BeFalse(),
+				"Unexpected nflog rules to appear on the correct felix instances")
+		})
+	})
+
```
tomastigera, line 577: Imo this test is redundant. It is already tested in UTs that it programs the rules or not. If anything (imo) it would make more sense to tests that it does not break connectivity and that policies work as expected in both cases if we have any such tests ....

felix/rules/endpoints.go
```diff
@@ -51,6 +51,7 @@ func (r *DefaultRuleRenderer) WorkloadEndpointToIptablesChains(
 	adminUp bool,
 	tiers []TierPolicyGroups,
 	profileIDs []string,
+	flowLogEnabled bool,
```
tomastigera, line 54: if this guy was a global option to the renderer, you would not need to change the signature of every method. As you observed yourself above.

Feels like this PR should not expand on this baggage, especially for internal methods.

felix/rules/endpoints.go
```diff
@@ -561,17 +578,18 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
 					//
 					// For untracked and pre-DNAT rules, we don't do that because there may be
 					// normal rules still to be applied to the packet in the filter table.
-					rules = append(rules, generictables.Rule{
-						Match:  r.NewMatch().MarkClear(r.MarkPass),
-						Action: r.Nflog(nflogGroup, CalculateEndOfTierDropNFLOGPrefixStr(dir, tier.Name), 0),
-					})
-
+					if flowLogEnabled {
```
tomastigera, line 581: in the end of the day, it seems like this is the only place where you need it. It does not need to be plumbed through all the func calls :thinking:

tomastigera (changes requested):

felix/dataplane/driver.go
mazdakn: I think it make sense to have all fields related to flow logs in one place. The conflict will happen once, and can be managed easily.

felix/fv/flow_logs_goldmane_test.go
mazdakn: Yeah, and for that many UTs are added. This one is just to confirm the result in FV. The `testSetup()` function also checks the connectivity and also that the policies are programmed.

mazdakn: Are you suggesting to remove it?

felix/rules/endpoints.go
mazdakn: Make sense. It makes it much cleaner. Thanks.

felix/fv/flow_logs_goldmane_test.go
mazdakn: With current situation, it seems there is no need for this test as stated. In all other FVs, flow logs are disabled, and as such connectivity and policies are tested.

tomastigera (approved):
LGTM

caseydavenport (approved):

PR #9849, opened 2025-02-14
Initial work to opensource guardian

Brian-McM:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


guardian/pkg/server/server.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/projectcalico/calico/guardian/pkg/conn"
+	"github.com/projectcalico/calico/guardian/pkg/tunnel"
+)
+
+type Server interface {
+	ListenAndServeCluster(ctx context.Context) error
+	ListenAndServeManagementCluster(ctx context.Context) error
+}
+
+// Client is the voltron client. It is used by Guardian to establish a secure tunnel connection to the Voltron server and
```
asincu, line 39: I think this is called server now and it is not public.

guardian/pkg/server/server.go
```diff
@@ -0,0 +1,196 @@
...
+	tunnelDialerOptions []tunnel.DialerOption
+}
+
+func New(addr string, opts ...Option) (Server, error) {
+	var err error
+	srv := &server{
+		http: new(http.Server),
+
+		tunnelAddr:        addr,
+		tunnelServerName:  strings.Split(addr, ":")[0],
```
asincu, line 72: What if we have an IPV6 address ?

guardian/pkg/server/server.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	log.Debug("expecting TLS server name: ", srv.tunnelServerName)
+
+	// set the dialer for the tunnel manager if one hasn't been specified
+	tunnelAddress := srv.tunnelAddr
+
+	var dialer tunnel.SessionDialer
+	if srv.tunnelCert == nil {
+		log.Warnf("No tunnel creds, using unsecured tunnel")
```
asincu, line 92: We should probably drop support for unsecure tunnel

guardian/pkg/tunnel/tunnel.go
```diff
@@ -0,0 +1,264 @@
...
+	"sync"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/calico/guardian/pkg/chanutil"
+)
+
+const (
+	tunnelNetwork = "voltron-tunnel"
```
asincu, line 30: I wonder if there is any benefits to configure this at start up instead of hard-coding it. In case we want to change it, it would be better to not hard-coded.

guardian/pkg/server/target.go
```diff
@@ -0,0 +1,169 @@
...
+	HostHeader *string `json:"hostHeader,omitempty"`
+	// AllowInsecureTLS allows https with insecure tls settings
+	AllowInsecureTLS bool `json:"allowInsecureTLS,omitempty"`
+
+	// ClientCertPath and ClientKeyPath can be set for mTLS on the connection
+	// from Voltron to the destination.
+	ClientCertPath string `json:"clientCertPath"`
+	ClientKeyPath  string `json:"clientKeyPath"`
+
+	Unauthenticated bool `json:"unauthenticated,omitempty"`
```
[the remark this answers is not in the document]
Brian-McM: I think so... this seems to only apply to the other side of the tunnel upon closer inspection (and probably not even for the tunnel itself).

guardian/pkg/server/server.go
Brian-McM: This isn't splitting an IP, the address is a domain name.

guardian/pkg/server/server.go
Brian-McM: Ya let's do it, I don't think we have any case (except maybe in testing?) where we configure without creds here.

guardian/pkg/asyncutil/async.go
```diff
@@ -0,0 +1,186 @@
...
+			logrus.Debug("Shutdown signal received, shutting down...")
+			return
+		case cmd := <-executor.cmdChan:
+			logrus.Debugf("Received command.")
+			if !pause {
+				executor.executeCommand(ctx, cmd)
+			} else {
+				backlog = append(backlog, cmd)
+			}
+		case cmd := <-executor.backlogChan:
```
asincu, line 141: Can we have more than 100 tasks in the backlog channel ?

guardian/pkg/config/tigera.go
```diff
@@ -0,0 +1,52 @@
...
+
+	cfg.configureLogging()
+
+	return cfg, nil
+}
+
+type TigeraConfig struct {
+	CalicoConfig
+
+	PacketCaptureCABundlePath string `default:"/certs/packetcapture/tls.crt" split_words:"true"`
```
asincu, line 25: Are these needed by this component ?

guardian/pkg/asyncutil/async.go
```diff
@@ -0,0 +1,186 @@
...
+	return executor.cmdChan
+}
+
+func (executor *commandExecutor[C, R]) ShutdownSignaler() Signaler {
+	return executor.shutdownCompleteSig
+}
+
+func (executor *commandExecutor[C, R]) Send(params C) <-chan Result[R] {
+	cmd, resultChan := NewCommand[C, R](params)
+	executor.cmdChan <- cmd
```
asincu, line 104: Can we send 101 requests ?

guardian/pkg/tunnel/tunnel.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+}
+
+func (t *tunnel) Listener() (net.Listener, error) {
+	return (<-t.getListenerExecutor.Send(nil)).Result()
+}
+
+func (t *tunnel) accept() (net.Conn, error) {
+	result := <-t.acceptConnExecutor.Send(nil)
+	logrus.Debug("Brian Markl.")
```
asincu, line 217: Might want to choose another Debug message.

guardian/pkg/asyncutil/async.go
```diff
@@ -0,0 +1,186 @@
...
+	"context"
+	"errors"
+	"io"
+	"sync"
+
+	"github.com/sirupsen/logrus"
+)
+
+type CommandExecutor[C any, R any] interface {
+	Send(C) <-chan Result[R]
```
asincu, line 31: I saw that in most cases, we call this as Send(nill). Do we really need parameter here ?

guardian/pkg/config/tigera.go
Brian-McM: No I didn't actually mean to commit this, I'll remove this (it's for private).

guardian/pkg/asyncutil/async.go
Brian-McM: Not at the same time, but the other side of the channel shouldn't block so there wouldn't be any issue.

guardian/pkg/config/config.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+
+		serverName, err := extractServerName(pemServerCrt)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		tlsConfig.ServerName = serverName
+	} else {
+		tlsConfig.ServerName = strings.Split(cfg.VoltronURL, ":")[0]
```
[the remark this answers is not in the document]
Brian-McM: This is the logic as it currently is with guardian in production.

guardian/pkg/config/config.go
```diff
@@ -0,0 +1,232 @@
...
+		log.Error("Invalid logging level passed in. Will use default level set to WARN")
+		// Setting default to WARN
+		level = log.WarnLevel
+	}
+
+	log.SetLevel(level)
+}
+
+func (cfg *Config) Cert() (string, *x509.CertPool, error) {
+	if strings.ToLower(cfg.VoltronCAType) == "public" {
```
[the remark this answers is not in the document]
Brian-McM: Not sure tbh, but it's in the current version of guardian so I need to make sure it's in this version.

guardian/pkg/cryptoutils/crypto.go
```diff
@@ -0,0 +1,54 @@
...
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+// GenerateFingerprint returns the sha256 hash for a x509 certificate printed as a hex number
+func GenerateFingerprint(certificate *x509.Certificate) string {
```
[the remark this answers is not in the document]
Brian-McM: It's logged in the dial function on failure.

guardian/pkg/config/config.go
```diff
@@ -0,0 +1,232 @@
...
+
+// Config is the struct to parse the env configuration into.
+type Config struct {
+	LogLevel      string `default:"INFO"`
+	CertPath      string `default:"/certs" split_words:"true" json:"-"`
+	VoltronCAType string `default:"Tigera" split_words:"true"`
+	VoltronURL    string `required:"true" split_words:"true"`
+
+	KeepAliveEnable   bool `default:"true" split_words:"true"`
+	KeepAliveInterval int  `default:"100" split_words:"true"`
```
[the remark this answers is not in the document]
Brian-McM: I'm just keeping the setting here the same as it is in Guardian currently.

guardian/pkg/daemon/daemon.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/projectcalico/calico/guardian/pkg/server"
+	"github.com/projectcalico/calico/guardian/pkg/tunnel"
+)
+
+type configOpts struct {
+	proxyTargets []server.Target
+}
+
+// Run starts the daemon, which configures and starts the services needed for guardian to run.
+func Run(cfg config.Config, proxyTargets []server.Target, opts ...Option) {
```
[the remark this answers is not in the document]
Brian-McM: No not yet, I was going to pass them in as options but they're not actually optional. I'll remove the options and add it in if it's actually needed at some other point.

guardian/pkg/config/config.go
```diff
@@ -0,0 +1,232 @@
...
+
+	TunnelDialRetryAttempts int           `default:"20" split_words:"true"`
+	TunnelDialRetryInterval time.Duration `default:"5s" split_words:"true"`
+	TunnelDialTimeout       time.Duration `default:"60s" split_words:"true"`
+
+	TunnelDialRecreateOnTunnelClose bool          `default:"true" split_words:"true"`
+	ConnectionRetryAttempts         int           `default:"25" split_words:"true"`
+	ConnectionRetryInterval         time.Duration `default:"5s" split_words:"true"`
+
+	Listen     bool   `default:"true"`
```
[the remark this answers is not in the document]
Brian-McM: I'm not exactly sure, but it's in the existing guardian.

guardian/pkg/asyncutil/async.go
Brian-McM: Not at the same time. Since this loop never blocks it's unlikely we'd every reach that level of requests, and if we do it would be cleared out.

guardian/pkg/tunnel/doc.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package tunnel defines an authenticated tunnel API, that allows creating byte
+// pipes in both directions, initiated from either side of the tunnel.
+package tunnel
```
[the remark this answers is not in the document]
Brian-McM: I'm not sure yet, it's only where we need to describe the package in general. I wanted to add some comments for this package, but not every one warrants it.

guardian/pkg/tunnel/tunnel.go
Brian-McM: For now I'm just going to keep it hardcoded (since this is how it's currently done).

guardian/pkg/server/options.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"time"
+
+	"github.com/projectcalico/calico/guardian/pkg/tunnel"
+)
+
+// Option is a common format for New() options
+type Option func(*server) error
+
+// WithProxyTargets sets the proxying targets, can be used multiple times to add
```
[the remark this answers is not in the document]
Brian-McM: Updated it so it does what the comment says.

guardian/pkg/server/server.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	log.Debug("expecting TLS server name: ", srv.tunnelServerName)
+
+	// set the dialer for the tunnel manager if one hasn't been specified
+	tunnelAddress := srv.tunnelAddr
+
+	var dialer tunnel.SessionDialer
+
+	dialer, err = tunnel.NewTLSSessionDialer(tunnelAddress, tlsConfig, srv.tunnelDialerOptions...)
```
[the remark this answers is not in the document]
Brian-McM: Done.

guardian/pkg/asyncutil/async.go
Brian-McM: I'd thought about this some more, and I'd like to just stick with having the parameter here. I could make another interface that doesn't take in a parameter, but it seems like that would be unnecessary given that you just pass nil as the parameter as opposed to having different interfaces to work with.

This is intended to be a general use interface as well, and plan on moving it to somewhere more general. This will at least be used for voltron.

guardian/pkg/server/server.go
Brian-McM: I've taken this out, it's not actually needed here.

guardian/pkg/asyncutil/types.go
```diff
@@ -0,0 +1,121 @@
...
+// Write writes the error to the buffer. If the buffer is full the error is dropped.
+func (b *asyncErrorBuffer) Write(err error) {
+	WriteNoWait(b.errs, err)
+}
+
+func (b *asyncErrorBuffer) Receive() <-chan error {
+	return b.errs
+}
+
+func (b *asyncErrorBuffer) Close() {
```
asincu, line 112: Can close be called twice ?

guardian/pkg/asyncutil/types.go
```diff
@@ -0,0 +1,121 @@
...
+func (s *signaler) Send() {
+	// If the channel is full we don't need to wait to send another signal, there's already an unprocessed signal.
+	WriteNoWait(s.ch, struct{}{})
+}
+
+func (s *signaler) Receive() <-chan struct{} {
+	return s.ch
+}
+
+func (s *signaler) Close() {
```
asincu, line 79: Can close be called twice ?

guardian/pkg/asyncutil/async.go
```diff
@@ -0,0 +1,274 @@
...
+// AsyncCommandExecutor executes commands sent using the Send channel asynchronously. The result is sent back on the <-chan Result[R}
+// channel when the command has executed. If the command output results in an EOF, the command is backlogged. EOF signals
+// that the call must fixe something, meaning it needs to pause, fix whatever is wrong, and then resume execution.
+// Resuming ensures the backlogged cmds are then executed again.
+type AsyncCommandExecutor[C any, R any] interface {
+	Send(C) <-chan Result[R]
+	ExecutionController
+}
+
+// ExecutionController manages multiple CommandExecutors.
```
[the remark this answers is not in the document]
Brian-McM: > This interface seems to handle re-connection in case of an error.

I'll add some better comments, but what is done between DrainAndBacklog and Resume is up to the caller. This is generic, non specific library function. The way it's being used in the tunnel is executing commands that interact with a connection, and re establishing the connection in failure, but that is specific to the tunnel.

guardian/pkg/asyncutil/types.go
Brian-McM: No and if it is it should panic (that's a developer error if this is called more than once).

guardian/pkg/asyncutil/types.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (b *asyncErrorBuffer) Close() {
+	close(b.errs)
+}
+
+// Clear drains the internal buffer and returns when there's nothing left.
+// Not that writing to the error buffer should not be done while clearing, since if writing is happening as quick
+// as clearing is then the buffer will never be cleared.
+func (b *asyncErrorBuffer) Clear() {
```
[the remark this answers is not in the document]
Brian-McM: It's not needed, these are writing too and interacting with channels. Can you point out a specific race condition you're seeing?

guardian/pkg/tunnel/tunnel_test.go
```diff
@@ -0,0 +1,174 @@
...
+			mockDialer := new(tunmocks.SessionDialer)
+			mockSession := new(tunmocks.Session)
+
+			mockDialer.On("Dial").Return(mockSession, nil)
+			tc.setSession(mockSession)
+
+			tun, err := tunnel.NewTunnel(mockDialer)
+			Expect(err).NotTo(HaveOccurred())
+
+			ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
Brian-McM: The TODO isn't to handle right now.

I'll add the timeout though, that's a good idea.

guardian/pkg/tunnel/tunnel.go
```diff
@@ -0,0 +1,286 @@
...
+			}
+		}
+	}()
+
+	var recreateSession <-chan struct{}
+
+	var drainCoordinatorCh <-chan asyncutil.Result[asyncutil.Signaler]
+	// Use a FunctionCallRateLimiter to 1) ensure we don't rapidly retry recreating the session if it's constantly
+	// disconnecting and 2) return an error if this is called to many times within a 30-second time window.
+	drainCoordinator := asyncutil.NewFunctionCallRateLimiter(2*time.Second, 30*time.Second, 5, func(any) (asyncutil.Signaler, error) {
```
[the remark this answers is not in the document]
Brian-McM: Not for right now, I'm not sure if I want this to be configurable or now. I need to think about this more as followup.

guardian/pkg/asyncutil/chan.go
```diff
@@ -0,0 +1,103 @@
...
+	select {
+	case v := <-c:
+		return v, true
+	default:
+		var v R
+		return v, false
+	}
+}
+
+func ReadBatch[R any](c <-chan R, n int) []R {
```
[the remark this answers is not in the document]
Brian-McM: No, I'll remove it.

guardian/pkg/config/config.go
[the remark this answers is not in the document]
Brian-McM: I updated this to parse it as a url and get the hostname.

guardian/pkg/asyncutil/async.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	return coordinator
+}
+
+// NewCommandExecutor creates a new CommandExecutor implementation. It calls the given function f with the command given
+// to Send. Any errors from the function are sent over the errBuff. If an EOF is sent over the error buff, the caller
+// must pause the executor, restart / fix whatever processes need restarting or fixing, then resume execution (using the
+// PauseExecution and ResumeExecution functions). ResumeExecution re runs the commands that failed with EOF.
+func NewCommandExecutor[C any, R any](ctx context.Context, errBuff ErrorBuffer, f func(context.Context, C) (R, error)) CommandExecutor[C, R] {
+	hdlr := &commandExecutor[C, R]{
```
asincu, line 115: typo: mind using executor or commandExecutor  instead of hdlr ?

guardian/pkg/asyncutil/async.go
```diff
@@ -0,0 +1,274 @@
...
+		close(executor.cmdChan)
+
+		// Wait for inflight requests to finish before handling outstanding cmds.
+		executor.inflightCmds.Wait()
+
+		close(executor.backlogChan)
+
+		// Add all outstanding commands in the backlog or cmd channels to the backlog slice.
+		executor.backlog = append(executor.backlog, ReadAll(executor.backlogChan)...)
+		executor.backlog = append(executor.backlog, ReadAll(executor.cmdChan)...)
```
asincu, line 151: I think we need to perform this operation before line 142: https://github.com/projectcalico/calico/pull/9849/files#diff-332e67201f01430847e13e593d0dc866cbb68781014e7dc94babff2de33d1f92R142. Either that or you do not need it.

guardian/pkg/asyncutil/types.go
```diff
@@ -0,0 +1,121 @@
...
+
+	select {
+	case c.resultChan <- Result[R]{err: err}:
+	default:
+		panic("result channel is full, this should never happen since only one result should ever be written")
+	}
+}
+
+// Signaler is an interface used for waiting for and sending simple signals.
+type Signaler interface {
```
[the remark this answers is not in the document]
Brian-McM: I think I'd prefer to keep this as Signaler. This isn't inteaded to be used as a Channel, only send / receive a signal with no information.

guardian/pkg/asyncutil/async.go
Brian-McM: Not at all this was from a previous iteration.

guardian/pkg/asyncutil/async.go
Brian-McM: It doesn't need to be done before, this is actually intentional. I'm closing the channel so nothing more can be sent over it. Closing the channel doesn't "clear" it, all the items will remain on the channel until they're read off.

This basically forces a panic to happen if somebody writes to the channel after it's closed and we're trying to clean up. If they could write to it after reading everything off, then that would result in some commands hanging after close.

caseydavenport (approved):

PR #9678, opened 2025-01-06
factor-out calico/base image to metadata.mk

aaaaaaaalex:
## Description
Parameterizes the calico/base image used to build various Calico components

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #9862, opened 2025-02-18
Changes to build Envoy images for Gateway API in OSS

nelljerram:
I am trying to arrange to build and publish the envoy-{gateway,proxy,ratelimit} images that are needed for Gateway API function on OSS.  To test that I've been using

    release/bin/release hashrelease build --skip-branch-check --build-images=true

With the changes in this PR so far, that succeeds and I have these images built locally:

    asia.gcr.io/projectcalico-org/envoy-{gateway,proxy,ratelimit}:TAG
    eu.gcr.io/projectcalico-org/envoy-{gateway,proxy,ratelimit}:TAG
    gcr.io/projectcalico-org/envoy-{gateway,proxy,ratelimit}:TAG
    us.gcr.io/projectcalico-org/envoy-{gateway,proxy,ratelimit}:TAG
    calico/envoy-{gateway,proxy,ratelimit}:TAG
    quay.io/calico/envoy-{gateway,proxy,ratelimit}:TAG
    envoy-{gateway,proxy,ratelimit}:TAG

where TAG is latest,latest-amd64,latest-arm64,latest-ppc64le,latest-s390x, except that envoy-proxy images aren't built for ppc64le and s390x.

The envoy-specific changes are mixed up with some more general changes to permit local testing of the hashrelease build process.  LMK if you'd prefer me to split pieces out into separate PRs.

Please note, the `third_party/...` changes are all just copied from Enterprise, except for the follow-on changes in commits https://github.com/projectcalico/calico/pull/9862/commits/d36b1212baef47798e789e6464783c91de116059, https://github.com/projectcalico/calico/pull/9862/commits/b6b09632d2e59df112ef44c1e4340fb01e1b195e and https://github.com/projectcalico/calico/pull/9862/commits/b18a136462c7494051b08304f82735bf0fec4f62


caseydavenport (commented):
This largely LGTM! The only thing I'm unsure of is introducing two UBI images - can we not keep the versions aligned?

metadata.mk
```diff
@@ -19,6 +19,7 @@ KINDEST_NODE_VERSION=v1.31.4
 KIND_VERSION=v0.25.0
 PROTOC_VER=v0.1
 UBI_VERSION=8.10
+UBI9_VERSION=9.4
```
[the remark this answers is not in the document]
nelljerram: @hjiawei I see from other PRs that you've been working on the UBI images.  Would you happen to know the answer to Casey's question?  I.e. do we really need two UBI versions, or could we perhaps update OSS across the board to use UBI version 9?

release/pkg/manager/calico/manager.go
radTuti, line 1: can you update the `images` variables so this can be included in metadata file as images for a specific build

release/Makefile
```diff
@@ -33,7 +33,7 @@ ci: static-checks
 .PHONY: hashrelease hashrelease-build hashrelease-publish
 hashrelease: hashrelease-build hashrelease-publish
 
-hashrelease-build: bin/release var-require-all-GITHUB_TOKEN
+hashrelease-build: bin/release
```
radTuti, line 36: `GITHUB_TOKEN` is required to generate release notes. 
any reason for taking it out?

third_party/envoy-proxy/Makefile
```diff
@@ -0,0 +1,59 @@
+include ../../metadata.mk
+
+PACKAGE_NAME ?= github.com/projectcalico/calico/third_party/envoy-proxy
+
+ENVOY_PROXY_IMAGE ?= envoy-proxy
+BUILD_IMAGES ?= $(ENVOY_PROXY_IMAGE)
+
+THIRD_PARTY_REGISTRY ?= gcr.io/unique-caldron-775/cnx/tigera/third-party
```
hjiawei, line 8: OSS sem build secret may not have access to our internal gcr registry. Do we want to open this?

third_party/envoy-gateway/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+# https://github.com/envoyproxy/gateway/blob/v1.1.2/tools/docker/envoy-gateway/Dockerfile
+# but:
+# - using calico/base instead of distroless
+# - using TARGETARCH instead of TARGETPLATFORM.
+
+FROM --platform=linux/amd64 busybox AS source
+# Create the data directory for eg
+RUN mkdir -p /var/lib/eg
+
+FROM calico/base:latest-${TARGETARCH}
```
hjiawei, line 11: OSS calico/base is parameterized in https://github.com/projectcalico/calico/pull/9678. Use the `CALICO_BASE` arg instead. There is no need to specify `TARGETARCH` because docker buildkit will handle it.

metadata.mk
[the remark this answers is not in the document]
hjiawei: I have removed this definition as explained in this [comment](https://github.com/projectcalico/calico/pull/9867/files#r1960945679). When defining the `UBI9_IMAGE` variable in `lib.Makefile`, you can use `latest`.

Regarding UBI 8 vs UBI 9, calico/node continues to use UBI 8 as its base, and our toolchain is also built on RHEL/UBI 8. While it may be possible to update calico/base to UBI9, fully transitioning to UBI 9 would require dropping support RHEL 8.

hjiawei:
Are the new envoy images built as part of the CI process? I could't find the semaphore pipeline (yaml) files. Am I overlooking something?

metadata.mk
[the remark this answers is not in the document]
nelljerram: > I have removed this definition as explained in this comment. When defining the UBI9_IMAGE variable in lib.Makefile, you can use latest.

As long as that also gets merged to Enterprise, that should be fine.  One of my objectives here is to minimize any divergence between OSS and Enterprise.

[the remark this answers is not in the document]
nelljerram: > Regarding UBI 8 vs UBI 9, calico/node continues to use UBI 8 as its base, and our toolchain is also built on RHEL/UBI 8. While it may be possible to update calico/base to UBI9, fully transitioning to UBI 9 would require dropping support for RHEL 8.

@caseydavenport I think that means we have to keep using both UBI 8 (e.g. for calico/node) and 9 (e.g. for the envoy builds).

release/Makefile
nelljerram: Yes, and I tried to explain this in the PR description.  It's so that I can test the release process locally.  Do we strictly need the release notes for a hashrelease?

third_party/envoy-proxy/Makefile
nelljerram: If needed I'm sure we can copy the envoybinary image to a public registry.  But let's first resolve everything else about this PR and then see if we need that extra step.

nelljerram: Based on occurrences of "gcr" in .semaphore/release/hashrelease.yml and .semaphore/release/release.yml, it looks like this will not be an immediate problem because the build does have access to gcr.io.

release/Makefile
danudey: Is GITHUB_TOKEN actually required for OSS releases? All the required information *should* be available through the public API.

radTuti: > Do we strictly need the release notes for a hashrelease?

Having the release notes in the hashrelease gives us a good idea of what is eventually coming out to allow updating if needed.

> Is GITHUB_TOKEN actually required for OSS releases? All the required information should be available through the public API.

publicly available does not mean you do not hit the rate limit. We want consistency.

third_party/envoy-proxy/Makefile
caseydavenport: @danudey hashreleases for OSS are already using quay, so I think we should just use quay for this. No reason to put these in a private registry for OSS?

release/Makefile
caseydavenport: > Is GITHUB_TOKEN actually required for OSS releases? All the required information should be available through the public API.

For ratelimiting purposes, yes it is required.

>  It's so that I can test the release process locally.

Why not just pass a token? Strictly speaking, generating release notes is part of the release process, and so we want to be running that logic when we generate hashreleases and do manual testing of the release process.

metadata.mk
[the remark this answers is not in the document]
caseydavenport: OK. Can we please add a detailed comment in metadata.mk explaining why we have both and when it would be acceptable to consolidate back to a single version, then? Doesn't need to be in this PR. But I think this is an exception to the rule and warrants clear documentation.

third_party/envoy-proxy/Makefile
nelljerram: Agreed, I'll move the envoybinary image to `quay.io/tigera/envoybinary:v1.31.5`.

nelljerram:
> Are the new envoy images built as part of the CI process? I could't find the semaphore pipeline (yaml) files. Am I overlooking something?

Yes, it's different in OSS.  They're built as part of the release process using Golang code under `release`.

release/pkg/manager/calico/manager.go
```diff
@@ -80,6 +83,9 @@ var (
 		"csi",
 		"ctl",
 		"dikastes",
+		"envoy-gateway",
+		"envoy-proxy",
+		"envoy-ratelimit",
```
radTuti, line 88: pls update the components in `release/internal/pinnedversion/templates/calico-versions.yaml.gotmpl` with the images

third_party/envoy-gateway/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# This is a copy of
+# https://github.com/envoyproxy/gateway/blob/v1.1.2/tools/docker/envoy-gateway/Dockerfile
+# but:
+# - using calico/base instead of distroless
+# - using TARGETARCH instead of TARGETPLATFORM.
+
+ARG CALICO_BASE
+
+FROM --platform=linux/amd64 busybox AS source
+# Create the data directory for eg
```
caseydavenport, line 10:
```suggestion
```

Unfinished comment?

radTuti (commented):
from a releasing perspective,

 in `release/internal/pinnedversion/templates/calico-versions.yaml.gotmpl`

- [x] update components list

in `release/pkg/manager/calico/manager.go`

- [x] update variables: `imageReleaseDirs`, `images`
- [ ] update function `assertImageVersions` 

in image/component dir

- [x] release-build Make target
- [x] release-publish Make target

third_party/envoy-gateway/Dockerfile
nelljerram: Actually no, it means "eg" as an abbreviation for envoy gateway.  But I'll improve it!

caseydavenport: ahhh makes sense. Yeah that acronym wasn't obvious to me :sweat_smile:

release/pkg/manager/calico/manager.go
nelljerram: Thanks @radTuti for catching all those!

nelljerram:
The CI jobs that failed in one run passed in the next run, so I feel confident that they are only flakes and are not reflecting a real problem in this PR.

nelljerram:
FYI @radTuti I'm thinking of merging this tomorrow morning (UK) so that I can then click the hashrelease promotion and keep an eye on that, to address any issues that arise.  (As opposed to merging it now and possibly dropping issues on other people overnight.)  I hope that sounds sensible.

PR #9867, opened 2025-02-18
Pin calico/base to release tags

hjiawei:
## Description

This change pins calico/base version to a particular tag instead of `latest`. It resolves issues during a (hash) release when non-amd64 components accidentally pull in amd64 bases.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,7 +206,7 @@ endif
 		python:3 \
 		bash -c '/usr/local/bin/python release/get-contributors.py >> /code/AUTHORS.md'
 
-update-pins: update-go-build-pin
+update-pins: update-go-build-pin update-calico-base-pin
 
 ###############################################################################
 # Post-release validation
--- a/lib.Makefile
+++ b/lib.Makefile
@@ -265,7 +265,7 @@ CERTS_PATH := $(REPO_ROOT)/hack/test/certs
 ifdef USE_UBI_AS_CALICO_BASE
 CALICO_BASE ?= $(UBI_IMAGE)
 else
-CALICO_BASE ?= calico/base
+CALICO_BASE ?= calico/base:$(CALICO_BASE_VER)
 endif
 
 ifndef NO_DOCKER_PULL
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -204,7 +204,7 @@ endif
 		python:3 \
 		bash -c '/usr/local/bin/python release/get-contributors.py >> /code/AUTHORS.md'
 
-update-pins: update-go-build-pin
+update-pins: update-go-build-pin update-calico-base-pin
```
hjiawei, line 207: Normally, we update calico/base together with calico/go-build before each release. This `update-pins` target will be triggered once we have a new calico/go-build release. Both calico/go-build and calico/base versions will be updated.

lib.Makefile
```diff
@@ -199,7 +199,7 @@ ifeq ($(BUILDARCH),amd64)
 	# *-amd64 tagged images for etcd are not available until v3.5.0
 	ETCD_IMAGE = quay.io/coreos/etcd:$(ETCD_VERSION)
 endif
-UBI_IMAGE ?= registry.access.redhat.com/ubi8/ubi-minimal:$(UBI_VERSION)
+UBI_IMAGE ?= registry.access.redhat.com/ubi8/ubi-minimal:latest
```
hjiawei, line 202: We do `microdnf upgrade` when building the container image. It essentially moves the tag to `latest`. We are not really pinning to the `UBI_VERSION` defined in `metadata.mk`.

lib.Makefile
```diff
@@ -374,6 +374,23 @@ define update_go_build_pin
...
+# update_calico_base_pin updates the CALICO_BASE_VER in metadata.mk.
+define update_calico_base_pin
+	$(eval new_ver := $(shell curl -s "https://hub.docker.com/v2/repositories/calico/base/tags/?page_size=100" | jq -r '.results[].name' | grep -E "^ubi8-[0-9]+$$" | sort -r | head -n 1))
+	$(eval old_ver := $(shell grep -E "^CALICO_BASE_VER" $(1) | cut -d'=' -f2 | xargs))
+
+	@echo "current CALICO_BASE_VER=$(old_ver)"
+	@echo "latest CALICO_BASE_VER=$(new_ver)"
+
+	bash -c '\
+		if [[ "$(new_ver)" > "$(old_ver)" ]]; then \
```
coutinhop, line 386: today I learned that this "just works" for strings

coutinhop (approved):

PR #9873, opened 2025-02-19
Disable Collector and lookup caches when flowlog is disabled

mazdakn:
## Description

This PR allows users to disable both Collector and Lookup caches, when flow log is disabled. Fortunately, due to not supporting flow logs in Windows dataplane, we already check the value of these two, before accessing them. As an example: https://github.com/projectcalico/calico/blob/b447c01cffa115ef0d3c28be0db63095e2809a33/felix/calc/calc_graph.go#L482

Also we already have unit tests for constructing dataplane with or without Collector: https://github.com/projectcalico/calico/blob/b447c01cffa115ef0d3c28be0db63095e2809a33/felix/dataplane/linux/int_dataplane_test.go#L124

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -3099,7 +3100,13 @@ func strToByte64(s string) [64]byte {
...
 
 func (m *bpfEndpointManager) ruleMatchIDFromNFLOGPrefix(nflogPrefix string) polprog.RuleMatchID {
-	return m.lookupsCache.GetID64FromNFLOGPrefix(strToByte64(nflogPrefix))
+	if m.lookupsCache != nil {
+		return m.lookupsCache.GetID64FromNFLOGPrefix(strToByte64(nflogPrefix))
+	}
+	// Lookup cache is not available, so generate an ID out of provided prefix.
+	h := fnv.New64a()
+	h.Write([]byte(nflogPrefix))
+	return h.Sum64()
```
mazdakn, line 3109: This is the approach we had before introducing flow logs.

mazdakn: https://github.com/projectcalico/calico/blob/232b72fce37b6dfff316e3977efba27fc5058696/felix/dataplane/linux/bpf_ep_mgr.go#L4159

fasaxc: Do we use this for anything when flow logs are disabled?

fasaxc (changes requested):
Would be good to have an FV that runs with it disabled, just to check the felix doesn't crash.

felix/dataplane/linux/bpf_ep_mgr.go
mazdakn: Yes, we do.

mazdakn:
@fasaxc based on this change, all FVs are executed without lookup caches, except the ones related to flow logs.

PR #9875, opened 2025-02-20
Support nested BGP peering with calico-nodes running in local kubevirt VM pods

song-jiang:
## Description

This PR adds support for allowing `calico-node` to peer with `calico-node` instances running inside KubeVirt VM pods locally, based on the labels of the VM pods.

API changes:
- New field `LocalWorkloadSelector` to BGPPeer resource.
- New field `localWorkloadPeeringIPV4` and `localWorkloadPeeringIPV4` to BGPConfigurations.

Felix changes:
- It watches BGPPeer and calculates local workloads selected by the BGPPeer.
- It populates endpoint status files with peering information.
- It add `localWorkloadPeeringIP` to the network interface of the workload selected by the BGPPeer.

Confd changes
- It watches endpoint status files updated by Felix.
- It reconfigures bird.cfg/bird6.cfg based on the peering information read from endpoint status files.

libcalico-go changes
- Added status-file-writer and status-file-watcher. 

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Support nested BGP peering with calico-nodes running in local kubevirt VM pods.
```

## Reminder for the reviewer

Other optional labels:


libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	log.WithField("dir", w.dir).Info("Started watching directory.")
+
+	// Listen for events
+	for {
+		select {
+		case event, ok := <-w.fsWatcher.Events:
+			if !ok {
+				// Stop if channel is closed.
```
fasaxc, line 95: Is this expected?  I'd have thought you'd want to recreate the watch in this case.

libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -0,0 +1,228 @@
...
+				}
+			}
+
+		case err, ok := <-w.fsWatcher.Errors:
+			if !ok {
+				// Stop if channel is closed.
+				return
+			}
+			log.WithError(err).Error("fsnotify error. Falling back to polling.")
+			go w.pollDirectory()
```
fasaxc, line 129: Seems inconsistent to start polling here but not if the other channel is closed.  I'm not keen on the decision to start polling being spread around throughout the function.  I'd probably have an outer retry loop for recreating the watch and do polling whenever we don't have an active watch.  You probably also need to do a poll at start of day after establishing the watch to make sure you didn't miss anything?

think it might be clearer not to move polling to a separate goroutine, just do the polling on this goroutine, then you can't get confused about whether a poll is running.

libcalico-go/lib/epstatusfile/status_file_writer.go
```diff
@@ -0,0 +1,190 @@
...
+	Mkdir(name string, perm os.FileMode) error
+	ReadDir(name string) ([]os.DirEntry, error)
+	ReadFile(name string) ([]byte, error)
+	WriteFile(name string, data []byte, perm os.FileMode) error
+}
+
+type defaultFilesys struct{}
+
+// Create wraps os.Create.
+func (fs *defaultFilesys) Create(name string) (*os.File, error) {
```
fasaxc, line 58: I think there's a ready-made FS for this.

felix/calc/active_bpgpeer_calculator.go
```diff
@@ -0,0 +1,299 @@
...
+				} else {
+					abp.onBGPPeerUpdate(bgpPeer)
+				}
+			} else {
+				logCxt.Debug("Deleting BGPPeer from abp")
+				abp.onBGPPeerDelete(id.Name)
+				delete(abp.bgpPeersByName, id.Name)
+			}
+		case libv3.KindNode:
+			if update.Value != nil {
```
fasaxc, line 127: Calculation graph should be "memoryless" so this should handle node deletion too.

felix/calc/active_bpgpeer_calculator.go
```diff
@@ -0,0 +1,299 @@
...
+				nodeName := update.Key.(model.ResourceKey).Name
+
+				logCxt := logrus.WithField("node", nodeName)
+
+				if nodeName == abp.hostname {
```
fasaxc, line 132:
```suggestion
				nodeName := update.Key.(model.ResourceKey).Name

				if nodeName != abp.hostname {
				    return
				}
				logCxt := logrus.WithField("node", nodeName)

```
Go idiom is to return early rather than nest more deeply.

felix/calc/active_bpgpeer_calculator.go
```diff
@@ -0,0 +1,299 @@
...
+						// if node labels has been updated, re-evaluate it againt all bgp peers.
+						for name, bgpPeer := range abp.bgpPeersByName {
+							if !abp.ifBgpPeerSelectHost(bgpPeer) {
+								// Trying to delete BGPPeer if it does not select the host.
+								abp.onBGPPeerDelete(name)
+							} else {
+								abp.onBGPPeerUpdate(bgpPeer)
+							}
+						}
+					}
```
fasaxc, line 146: Suggest moving this into `spb.onNodeUpdated()`

felix/calc/active_bpgpeer_calculator.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// Update peer data for the endpoint.
+	delete(abp.peersByID, id)
+
+	// Send the update.
+	logrus.Debugf("Clear BGP Peer data on endpoint %s", id)
+	abp.OnEndpointBGPPeerDataUpdate(id, nil)
+}
+
+func (abp *ActiveBGPPeerCalculator) ifBgpPeerSelectHost(bgpPeer *v3.BGPPeer) bool {
```
fasaxc, line 232:
```suggestion
func (abp *ActiveBGPPeerCalculator) bgpPeerSelectsLocalNode(bgpPeer *v3.BGPPeer) bool {
```

felix/calc/active_bpgpeer_calculator.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	return false
+}
+
+func (abp *ActiveBGPPeerCalculator) onBGPPeerUpdate(bgpPeer *v3.BGPPeer) {
+	logrus.WithField("bgppeer", bgpPeer).Debugf("On BGP Peer update")
+	name := bgpPeer.Name
+
+	abp.bgpPeersByName[name] = bgpPeer
```
fasaxc, line 256: Looks like this is duplicated in the caller?

felix/calc/active_bpgpeer_calculator.go
```diff
@@ -0,0 +1,299 @@
...
+	newSelector, err := sel.Parse(rawSelector)
+	if err != nil {
+		logrus.WithError(err).Errorf("Couldn't parse selector: %s", rawSelector)
+		return
+	}
+
+	// Save new selector for this BGP peer if it is different with the old one.
+	selector, exists := abp.workloadSelectorsByPeerName[name]
+	if exists {
+		if reflect.DeepEqual(selector, newSelector) {
```
fasaxc, line 268: selector has cached fields inside so DeepEqual is incorrect.  I think it has an Equal method?  (Or you can compare the UID)

felix/calc/active_bpgpeer_calculator.go
```diff
@@ -0,0 +1,299 @@
...
+	hostname string
+
+	// Node labels for this host.
+	nodeLabels map[string]string
+
+	// Labels of active local endpoints.
+	labelsByID map[model.WorkloadEndpointKey]map[string]string
+
+	// Active BGP peers.
+	bgpPeersByName map[string]*v3.BGPPeer
```
fasaxc, line 45:
```suggestion
	// All BGP peers.
	allBGPPeersByName map[string]*v3.BGPPeer
```

felix/calc/active_bpgpeer_calculator.go
```diff
@@ -0,0 +1,299 @@
...
+			}
+		case libv3.KindNode:
+			if update.Value != nil {
+				nodeName := update.Key.(model.ResourceKey).Name
+
+				logCxt := logrus.WithField("node", nodeName)
+
+				if nodeName == abp.hostname {
+					node := update.Value.(*libv3.Node)
+					if !reflect.DeepEqual(node.Labels, abp.nodeLabels) {
```
fasaxc, line 134: I think there's `maps.Equal(...)` now, which is faster than DeepEqual

fasaxc (changes requested):

felix/calc/active_bpgpeer_calculator.go
song-jiang: If name of a local node is `my-host`, the bgppeer calculator handles node update on `my-host` only. If `my-host` is deleted, calico-node is gone so we don't need to handle this scenario. Deletion of other nodes does not have any effect on this node.

libcalico-go/lib/epstatusfile/status_file_writer.go
song-jiang: Are you saying something similar to https://github.com/spf13/afero?

felix/calc/active_bpgpeer_calculator.go
song-jiang: Code updated to handle node deletion.

libcalico-go/lib/epstatusfile/status_file_writer.go
song-jiang: Actually we are using our own version of filesystem, because in some test we need to evaluate the timing for file operations. Those functionalities are not provided by an off-the-shelf module.

fasaxc: I mean https://pkg.go.dev/os#DirFS `os.DirFS(...)`

fasaxc (changes requested):

confd/tests/compiled_templates/bgpfilter/export_only/global_peer/bird.cfg
```diff
@@ -94,6 +94,7 @@ protocol bgp Global_10_192_0_3 from bgp_template {
     calico_export_to_bgp_peers(true);
     reject;
   };  # Only want to export routes for workloads.
+  next hop self;
```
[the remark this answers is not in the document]
song-jiang: Yes. There's a use case for this.
we need this config for `VM (AS1) ----- NODE (AS2) ----- TOR (AS2)` model where NODE and TOR has the same ASnumber, and that means we need to configure next hop self on the Calico end of the NODE - TOR peerings. Otherwise TOR would not be able to resolve routes exported from the NODE `child-cluster-cidr-block via VM`. With this config, the route exported from NODE to TOR will be updated to `child-cluster-cidr-block via NODE`.  

This config is enabled whenever `(it isn't emitting next hop keep) AND (remote AS == local AS)`, and that wouldn't break any pre-existing use cases, because the NODE is always the next hop within an iBGP.

libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -0,0 +1,264 @@
...
+		case <-w.stopChan:
+			return nil
+		}
+	}
+}
+
+// Start begins watching the directory.
+func (w *FileWatcher) runWatcher() {
+	// Get current state of the directory and emit initial events.
+	w.scanDirectory()
```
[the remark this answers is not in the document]
song-jiang: Good point!

felix/linkaddrs/link_addrs.go
```diff
@@ -0,0 +1,378 @@
...
+		handlemgr.WithNewHandleOverride(la.newNetlinkHandle),
+	)
+	return &la
+}
+
+func (la *LinkAddrsManager) QueueResync() {
+	la.resyncPending = true
+}
+
+func (la *LinkAddrsManager) SetLinkLocalAddress(ifacename string, addr string) error {
```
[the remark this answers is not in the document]
song-jiang: I used `ip.CIDR` instead of `ip.Addr`. 
This is because when we sync with the dataplane, the format is CIRD instead of a single IP. It also add support of programming link local address with IPNet.

libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -0,0 +1,264 @@
...
+// FileWatcher monitors a directory and reports events (creation, update and deletion) on file updates.
+// It tries to use fsnotify to watch the directory. If fsnotify fails, it switches to polling mode.
+
+type Callbacks struct {
+	OnFileCreation func(fileName string)
+	OnFileUpdate   func(fileName string)
+	OnFileDeletion func(fileName string)
+}
+
+type FileWatcher struct {
```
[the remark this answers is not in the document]
song-jiang: Can do but I think it is better to raise another PR for it.  @aaaaaaaalex may help on this.

libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+
+		if w.fsWatcher != nil {
+			// Run fsnotify watcher loop if possible.
+			err := w.runFsnotifyWatcher(w.fsWatcher)
+			w.fsnotifyActive = false
+			if err != nil {
+				// fall back to polling.
+				log.WithError(err).Info("Start polling directory on updates.")
+				w.pollC <- struct{}{}
```
fasaxc, line 180: I think this send could deadlock:
* First time through the loop, send is OK, leaves pending item on the channel.
* Then `w.pollTicker.C` pops in the select below and select chooses that instead of pollC.
* Loop and hit this error again, now we deadlock.
Not quite sure what this is trying to do.  Could you just `continue` to tool again?  Or should you let it fall through and wait for the ticker?

libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// scanDirectory detects file creations, modifications, and deletions.
+func (w *FileWatcher) scanDirectory() {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	currentState := make(map[string]os.FileInfo)
+
+	err := filepath.Walk(w.dir, func(path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
fasaxc: I'd rather minimize scope given the time pressure on this one!

fasaxc (changes requested):

libcalico-go/lib/epstatusfile/status_file_watcher.go
song-jiang: That's a mistake. I should just let it fall through and wait for the ticker. There's no need for this channel.

libcalico-go/lib/selector/selector.go
```diff
@@ -29,6 +29,7 @@ func init() {
 	if err != nil {
 		panic(fmt.Sprintf("Failed to parse !all(): %v", err))
 	}
+	_ = NoMatch.UniqueID() // Initialise its cached fields.
```
fasaxc, line 32: Sigh, looks like there's LabelRestrictions() that gets cached too.

fasaxc (approved):

PR #9856, opened 2025-02-17
Revert declaration that _our_ plugin provides QoS support

nelljerram:
In fact QoS support is provided by the 'qos' service plugin, which is configured with

    service_plugins = qos

in `/etc/neutron/neutron.conf`.  It's wrong to declare it here as well, and that causes a traceback ending with:

    2025-02-17 12:34:56.469 40638 ERROR neutron   File "/usr/lib/python3/dist-packages/neutron/api/extensions.py", line 492, in __init__
    2025-02-17 12:34:56.469 40638 ERROR neutron     self.check_if_plugin_extensions_loaded()
    2025-02-17 12:34:56.469 40638 ERROR neutron   File "/usr/lib/python3/dist-packages/neutron/api/extensions.py", line 572, in check_if_plugin_extensions_loaded
    2025-02-17 12:34:56.469 40638 ERROR neutron     raise exceptions.ExtensionsNotFound(
    2025-02-17 12:34:56.469 40638 ERROR neutron neutron_lib.exceptions.ExtensionsNotFound: Extensions not found: ['qos'].

Looking at `check_if_plugin_extensions_loaded`, we see

    def check_if_plugin_extensions_loaded(self):
        """Check if an extension supported by a plugin has been loaded."""
        plugin_extensions = self.get_supported_extension_aliases()
        missing_aliases = plugin_extensions - set(self.extensions)
        missing_aliases -= _PLUGIN_AGNOSTIC_EXTENSIONS
        if missing_aliases:
            raise exceptions.ExtensionsNotFound(
                extensions=list(missing_aliases))

so it's clear that this is caused by the wrong `_supported_extension_aliases` declaration in our plugin code.

PR #9877, opened 2025-02-20
OpenStack: Complete what we need for QoS support in our driver

nelljerram:
Sorry, these are more required plumbing changes that I missed in #9811 and #9856.

Also a key fix to query bandwidth limit and packet rate rules directly, instead of via the policy, as the policy object does not really have a 'rules' attribute.

After these, I've verified that I can execute the QoS workflow in https://docs.openstack.org/neutron/latest/admin/config-qos.html#user-workflow.

### DIFF
--- a/networking-calico/networking_calico/plugins/ml2/drivers/calico/qos_driver.py
+++ b/networking-calico/networking_calico/plugins/ml2/drivers/calico/qos_driver.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2025 Tigera, Inc. All rights reserved.
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from networking_calico.compat import log
+
+from neutron_lib import constants
+from neutron_lib.api.definitions import portbindings
+from neutron_lib.db import constants as db_consts
+from neutron_lib.services.qos import base
+from neutron_lib.services.qos import constants as qos_consts
+
+
+LOG = log.getLogger(__name__)
+
+DRIVER = None
+
+SUPPORTED_RULES = {
+    qos_consts.RULE_TYPE_BANDWIDTH_LIMIT: {
+        qos_consts.MAX_KBPS: {
+            'type:range': [0, db_consts.DB_INTEGER_MAX_VALUE]},
+        qos_consts.MAX_BURST: {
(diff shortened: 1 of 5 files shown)


networking-calico/networking_calico/plugins/ml2/drivers/calico/qos_driver.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+LOG = log.getLogger(__name__)
+
+DRIVER = None
+
+SUPPORTED_RULES = {
+    qos_consts.RULE_TYPE_BANDWIDTH_LIMIT: {
+        qos_consts.MAX_KBPS: {
+            'type:range': [0, db_consts.DB_INTEGER_MAX_VALUE]},
```
coutinhop, line 34: quick question: does `0` mean 'unset' in this context? on the felix side (k8s pod annotations), we do mean `0 = unset`, but we do some validations on sane values (https://github.com/projectcalico/calico/blob/a96488d9c297a054e2fb83b262819e18b9574fcb/libcalico-go/lib/backend/k8s/conversion/workload_endpoint_default.go#L40-L46), but I wonder how and where to best have it for openstack...

nelljerram: I think it means that the Neutron will restrict the `max_kbps` field to be between 0 and `db_consts.DB_INTEGER_MAX_VALUE`, in QoS policies that are attached to Calico ports.

Following that interpretation, I've specified the range for `max_burst_kpps` to be between 0 and 0 - i.e. that field must always be zero - because Calico doesn't support a burst packet rate limit.

I should probably also mention that I mostly just copied this file from https://github.com/openstack/neutron/blob/stable/2024.1/neutron/services/qos/drivers/linuxbridge/driver.py

nelljerram: Also, yes, 0 effectively means unset, because of the conditions like
```
                if r['max_kbps'] != 0:
```
in endpoints.py (above).

nelljerram: (I now realize that my first answer was not answering the question that you were really asking!)

nelljerram: Possibly I should update the ranges here to match your `maxBandwidth`, `maxBurst` and `maxPacketRate` values.  But:
- Where and how are your values policed?  Does a setting outside the range get pulled into the range?  Or does it just make the WEP invalid?  Or is that setting just ignored?
- What does `1P` mean?
- For `minPacketRate` I need to experiment more, to check that it's still possible for a field to be unset when there is a non-zero minimum for that field.
- I need to think more about `minNumConnections` and `maxNumConnections`, to decide whether to apply those to constrain the neutron.conf configuration fields.

One approach would be not to check those min/max ranges in the OpenStack code, and leave it for your code to handle.  That might be preferable to avoid duplicating the values and creating more maintenance for later.

In any case, can you let me know if this is your _only_ concern?  If it is, I'd like to merge this PR anyway to facilitate further system testing.  I can raise a follow-on PR to do whatever we think needs doing on this point.

nelljerram:
I'm going to merge this in order to get fresh packages built that include these changes.  That means I'll be able to redeploy a ST rig without having to manual apply these changes afterwards.  I've spent today system testing, and that hasn't thrown up any further issues in this PR's code.

PR #9882, opened 2025-02-21
fix when CertFile and Cert is ""

rxy0210:
## Description

When my etcd is not configured with a file such as cert, this code is problematic.
With the original code, it is predictable that the default user of the code filled in the details of the cert, or the location of the cert file. This results in an etcd with no cert configured and a connection error.
## Related issues/PRs
fixes https://github.com/projectcalico/calico/issues/9883

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/etcdv3/etcdv3.go
+++ b/libcalico-go/lib/backend/etcdv3/etcdv3.go
@@ -99,7 +99,8 @@ func NewEtcdV3Client(config *apiconfig.EtcdConfig) (api.Client, error) {
 			Key:    config.EtcdKey,
 		}
 		tlsConfig, err = tlsInfo.ClientConfigInlineCertKey()
-	} else {
+	}
+	if haveFiles {
 		tlsInfo := &transport.TLSInfo{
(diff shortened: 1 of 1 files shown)


caseydavenport (approved):
Seems fine to me (although I cannot recommend running without enabling TLS)

caseydavenport:
Running tests - but I'd like to understand why you want the ability to run without TLS enabled.

rxy0210:
> 正在运行测试 - 但我想了解您为什么希望能够在未启用 TLS 的情况下运行。

When I started an etcd locally just for debugging purposes, I chose to start without TLS in order to complete the debugging quickly. And etcd itself supports a boot mode without TLS, but we are not compatible, which may be a problem.

caseydavenport:
We have a general policy of avoiding adding non-production code into the project - which this would fall under. I'd recommend simply generating the necessary TLS configuration for your test environment rather than running without TLS enabled.

PR #9885, opened 2025-02-21
Allow users to disable flowlogs in bpf dataplane

mazdakn:
## Description

To minimise the impact of flow logs on clusters, users should be able to decide if flow log rules are programmed in data plane or not. Currently, flow logs are always programmed in eBPF . This PR makes it possible that flow logs rules in eBPF data plane are programmed only when flow log feature is enabled.

## Related issues/PRs
Disable nflog rules in iptables/nftables when flow logs is disabled: https://github.com/projectcalico/calico/pull/9848
Disable lookup caches and collector when flow logs is disabled: https://github.com/projectcalico/calico/pull/9873

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -3809,6 +3776,11 @@ func (m *bpfEndpointManager) loadPolicyProgram(
 		ipSetIDAlloc = m.v6.ipSetIDAlloc
 	}
 
+	// Flow logs functionality is enabled.
+	if m.lookupsCache != nil {
```
tomastigera, line 3780: nit, perhaps wrap the test in a method so the code reads better, rather than needing a comment?
```suggestion
	if m.FlowLogsEnabled() {
```

tomastigera (approved):
LGTM, one nit

felix/dataplane/linux/bpf_ep_mgr.go
mazdakn: make sense. Will fix it.

PR #9928, opened 2025-03-05
Fix hashrelease task failing to generate Slack notification on failure

nelljerram:
The Semaphore log for the last few builds has:

    2025-03-05 12:53:12.607 [ERROR][8343] release/message.go 92: Failed to render message error=invalid character '\n' in string literal
    2025-03-05 12:53:12.607 [ERROR][8343] release/main.go 86: Failed to send slack notification error=invalid character '\n' in string literal

It's because the error message can include literal newlines, which aren't allowed; we need to convert those to escaped newlines.

As part of this:

- Add a `ut` target to `release/Makefile`.

- Add a test to repro the case above.

- Comment out a failing case in the existing `version_test.go`.


release/internal/version/version_test.go
```diff
@@ -33,14 +34,21 @@ func TestDetermineReleaseVersion(t *testing.T) {
...
 		// A dev tag leading up to a minor release should return the minor release number.
 		"v3.29.0-0.dev-424-gfd40f1838223": "v3.29.0",
-		"v3.22.0-0.dev":                   "v3.22.0",
+
+		// This case fails because HasDevTag requires an "-N-gCOMMIT" part after the
+		// "0.dev", so commenting it out for now.  If just "vX.Y.Z-0.dev" is valid,
+		// HasDevTag needs fixing to allow that.  If HasDevTag is correct, this case should
+		// be removed.
+		//
+		// "v3.22.0-0.dev": "v3.22.0",
```
caseydavenport, line 43: Pretty sure this is a valid tag - for example, I see this tag in recent history: `v3.30.0-0.dev` 

@radTuti I think you had made some changes to HasDevTag recently - guessing those broke this test (and we weren't running the tests so we didn't see it!)

caseydavenport (approved):

release/internal/version/version_test.go
radTuti: it is actually not a valid dev tag as all dev tags are in the format `vX.Y.Z(-A.B)-N=gCOMMIT` so on the commit tagged as `v3.30.0-0.dev` returns `v3.30.0-0.dev-0-ga8daadedb3db`.

I think rather than comment it out, it should be removed.

radTuti (approved):

marvin-tigera:

PR #9942, opened 2025-03-06
Refactor initial creation of flow client / connection flow

Brian-McM:
This commit does the following:
- Create the grpc client in the goldmane constructor and return an error if it fails (the client doesn't initiate a connection, just validates the given parameters)
- Return a channel from the "Connect" function (previously Run) so that callers have the option to wait for an initial connection to be established with goldmane
- Change the goldmane daemon to accept the config parameters. This allows for configuring and running the  daemon in tests as it would be run in the container for fv like tests.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/collector/dpstatshelper.go
+++ b/felix/collector/dpstatshelper.go
@@ -57,8 +57,12 @@ func New(
 	goldmaneAddr := configParams.FlowLogsGoldmaneServer
 	if goldmaneAddr != "" {
 		log.Infof("Creating Flow Logs GoldmaneReporter with address %v", goldmaneAddr)
-		gd := goldmane.NewReporter(goldmaneAddr)
-		dispatchers[FlowLogsGoldmaneReporterName] = gd
+		gd, err := goldmane.NewReporter(goldmaneAddr)
+		if err != nil {
+			log.WithError(err).Error("Failed to create Flow Logs GoldmaneReporter, skipping adding reporter.")
+		} else {
+			dispatchers[FlowLogsGoldmaneReporterName] = gd
+		}
 	}
 	if len(dispatchers) > 0 {
 		log.Info("Creating Flow Logs Reporter")
--- a/felix/collector/goldmane/client.go
+++ b/felix/collector/goldmane/client.go
@@ -21,8 +21,6 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/projectcalico/calico/felix/collector/flowlog"
 	"github.com/projectcalico/calico/felix/collector/types/endpoint"
@@ -38,22 +36,22 @@ type GoldmaneReporter struct {
 	once    sync.Once
 }
 
-func NewReporter(addr string) *GoldmaneReporter {
+func NewReporter(addr string) (*GoldmaneReporter, error) {
+	cli, err := client.NewFlowClient(addr)
+	if err != nil {
+		return nil, err
+	}
 	return &GoldmaneReporter{
 		address: addr,
-		client:  client.NewFlowClient(addr),
-	}
+		client:  cli,
+	}, nil
 }
 
 func (g *GoldmaneReporter) Start() error {
 	var err error
 	g.once.Do(func() {
-		var grpcClient *grpc.ClientConn
-		grpcClient, err = grpc.NewClient(g.address, grpc.WithTransportCredentials(insecure.NewCredentials()))
-		if err != nil {
-			return
-		}
-		go g.client.Run(context.Background(), grpcClient)
+		// We don't wait for the initial connection to start so we don't block the caller. This could be changed later.
+		g.client.Connect(context.Background())
 	})
 	return err
 }
--- a/goldmane/pkg/client/client.go
+++ b/goldmane/pkg/client/client.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/projectcalico/calico/goldmane/pkg/internal/flowcache"
 	"github.com/projectcalico/calico/goldmane/proto"
@@ -30,20 +31,35 @@ const (
 	FlowCacheCleanup = 30 * time.Second
 )
 
-func NewFlowClient(server string) *FlowClient {
-	return &FlowClient{
-		inChan: make(chan *proto.Flow, 5000),
-		cache:  flowcache.NewExpiringFlowCache(FlowCacheExpiry),
+// NewFlowClient creates a new client to the goldmane grpc API. It creates the initial grpcClient connection to verify
+// that an initial connection can be created when connect is called (grpc.NewClient doesn't establish an initial connection,
+// just validates that it should be able to with the given parameters).
+//
+// If an error is returned, it means that no amount of retrying will create the client with the same parameters.
+func NewFlowClient(server string) (*FlowClient, error) {
+	grpcClient, err := grpc.NewClient(server, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, err
 	}
+
+	return &FlowClient{
+		inChan:      make(chan *proto.Flow, 5000),
+		cache:       flowcache.NewExpiringFlowCache(FlowCacheExpiry),
+		grpcCliConn: grpcClient,
+	}, nil
 }
 
 // FlowClient pushes flow updates to the flow server.
 type FlowClient struct {
-	inChan chan *proto.Flow
-	cache  *flowcache.ExpiringFlowCache
+	inChan      chan *proto.Flow
+	cache       *flowcache.ExpiringFlowCache
+	grpcCliConn *grpc.ClientConn
 }
 
-func (c *FlowClient) Run(ctx context.Context, grpcClient grpc.ClientConnInterface) {
+// Connect starts the grpc connection to stream flows to goldmane. It returns a channel that closes once the initial
+// connection has been established which gives callers the option to wait for an initial connection before proceeding
+// to use the client.
+func (c *FlowClient) Connect(ctx context.Context) <-chan struct{} {
 	logrus.Info("Starting flow client")
 	defer func() {
 		logrus.Info("Stopping flow client")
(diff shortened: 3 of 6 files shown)


goldmane/pkg/client/client.go
```diff
@@ -52,8 +68,52 @@ func (c *FlowClient) Run(ctx context.Context, grpcClient grpc.ClientConnInterfac
...
+				logrus.WithError(err).Warn("Failed to receive receipt")
+				break
+			}
+		}
+
+		if err := rc.CloseSend(); err != nil {
+			logrus.WithError(err).Warn("Failed to close connection")
+		}
+
+		rc, err = c.connect(ctx)
```
caseydavenport, line 105: Why do we need another call to connect here?

goldmane/pkg/client/client.go
```diff
@@ -52,8 +68,52 @@ func (c *FlowClient) Run(ctx context.Context, grpcClient grpc.ClientConnInterfac
...
 	go c.cache.Run(FlowCacheCleanup)
 
+	startUp := make(chan struct{})
+	go func() {
+		rc, err := c.connect(ctx)
+		// Close this regardless of the error since we don't want the other side of the channel to hang forever.
+		close(startUp)
+		if err != nil {
+			logrus.WithError(err).Warn("Unable to connect to flow server.")
+			return
```
caseydavenport, line 78: If we hit an error how do we retry? I think we just return and the client never retries?

goldmane/pkg/daemon/daemon.go
```diff
@@ -66,14 +66,23 @@ type Config struct {
...
 	if err := envconfig.Process("", &cfg); err != nil {
 		logrus.WithError(err).Fatal("Failed to load configuration from environment")
 	}
 
 	utils.ConfigureLogging(cfg.LogLevel)
+
+	return cfg
+}
+
+func (cfg Config) ConfigureLogging() {
```
caseydavenport, line 81: Don't think we need this new function (and it's never being called)

felix/collector/dpstatshelper.go
```diff
@@ -57,8 +57,12 @@ func New(
 	goldmaneAddr := configParams.FlowLogsGoldmaneServer
 	if goldmaneAddr != "" {
 		log.Infof("Creating Flow Logs GoldmaneReporter with address %v", goldmaneAddr)
-		gd := goldmane.NewReporter(goldmaneAddr)
-		dispatchers[FlowLogsGoldmaneReporterName] = gd
+		gd, err := goldmane.NewReporter(goldmaneAddr)
+		if err != nil {
+			log.WithError(err).Error("Failed to create Flow Logs GoldmaneReporter, skipping adding reporter.")
```
caseydavenport, line 62: I think this is a fatal error, no?

felix/collector/goldmane/client.go
```diff
@@ -38,22 +36,22 @@ type GoldmaneReporter struct {
...
 func (g *GoldmaneReporter) Start() error {
 	var err error
 	g.once.Do(func() {
-		var grpcClient *grpc.ClientConn
-		grpcClient, err = grpc.NewClient(g.address, grpc.WithTransportCredentials(insecure.NewCredentials()))
-		if err != nil {
-			return
-		}
-		go g.client.Run(context.Background(), grpcClient)
+		// We don't wait for the initial connection to start so we don't block the caller. This could be changed later.
```
caseydavenport, line 53:
```suggestion
		// We don't wait for the initial connection to start so we don't block the caller.
```

goldmane/pkg/client/client.go
Brian-McM: If connect returns an error it's a fatal one (context cancelled for example). So this will block until there's a fatal error or a connection is made (using the back off retry).

In the next loop, if an error occurs using rc then we call connect again, which does the same thing (retry until a connection is established or return if there's a fatal error) then continue processing.

goldmane/pkg/daemon/daemon.go
Brian-McM: Sooo this is preemptive (and actually meant to replace the utils.ConfigureLogging up above.

If we're creating the config manually (as opposed to from the env), mostly for tests, then we want to ensure that we can configure logging.

felix/collector/dpstatshelper.go
Brian-McM: Currently it's up to you whether or not you want this to be fatal. Do we crash loop node if the reporter can't be configured?

I want to say "yes" but that's not how this is currently working (this error is swallowed and not even logged https://github.com/projectcalico/calico/pull/9942/files#diff-909a2a8475061194f60bb3e6864203a1fa0390d37f6071fe27c2506f382dd45eL52-L55).

goldmane/pkg/client/client.go
Brian-McM: This is the reconnect logic, if that loop before this exits it means there was an error that requires a reconnect. I think when you wrote this comment I may not have had the outer loop.

goldmane/pkg/daemon/daemon.go
caseydavenport: idk, feels like extra code for no benefit to me. We already have a ConfigureLogging that takes an argument... so why do we need a bound method that calls the utility function?

Ultimately it doesn't really matter, but I don't see the benefit. If we're creating the config manually we can just call the function that already exists?

felix/collector/dpstatshelper.go
caseydavenport: idk, maybe one for @mazdakn

goldmane/pkg/client/client.go
caseydavenport: I don't understand - there's no lop here, so if we hit this error we `return` but the code that called `Connect()` won't see any indicator of that error.

caseydavenport:
```go
for {
    rc, err := c.connect(ctx)
    do.Once(close(startUp))
    if err != nil {
        ...
}
```

We could put it in the loop (and remove the need for both `connect` calls) if we did something like this

caseydavenport: > If connect returns an error it's a fatal one (context cancelled for example). So this will block until there's a fatal error or a connection is made (using the back off retry).

Ok.. I see... you're saying we can never recover from any of the errors that connect returns.

I think we need to either exit the process, or hook into liveness / readiness probes

goldmane/pkg/client/client.go
```diff
@@ -52,8 +68,57 @@ func (c *FlowClient) Run(ctx context.Context, grpcClient grpc.ClientConnInterfac
...
+			if err != nil {
+				logrus.WithError(err).Warn("Failed to reconnect to flow server, will not retry (fatal error).")
+				return
+			}
+		}
+	}()
+
+	return startUp
+}
+
```
caseydavenport, line 118:
```suggestion
// connect establishes a new connection to the server and sends any cached logs. Note that non-fatal errors are retried.
// Any returned error is deemed unrecoverable and demands establishment of a new underlying gRPC connection
```

caseydavenport: or something like that

goldmane/pkg/client/client.go
Brian-McM: Which process do you mean? This does exit the current process (in the client) by returning (the same logic as before).

I think we should hook into some liveness probe or something, but this wasn't what I was trying to iterate on with this PR. I'm not quite trying to change how this works in prod, just the interface in which we interact to make it easier to interact with.

Should I create a ticket to track the liveness probe logic (or process exiting, however this needs to be handled by felix), or did you already have a task for that?

goldmane/pkg/daemon/daemon.go
caseydavenport: > We could but that means the caller needs to know the internal details of this. If we follow this pattern and we have 20 different test suites that create the Config struct then directly call utils.ConfigureLogging 

None of the test suites reference Config at all, because none of the test suites actually run the Daemon.

Brian-McM: Just removed it.

caseydavenport: this function is still here @Brian-McM

goldmane/pkg/client/client.go
caseydavenport: Yeah, I suppose it's no worse than it is today.

But goroutine != process here - this error should result in a new gRPC connection attempt, either by restarting Felix altogether or by exiting and creating a new FlowClient altogether (while leaving Felix running).

felix/collector/dpstatshelper.go
caseydavenport: yeah

caseydavenport: Although I almost prefer the current behavior... presumably it panics if it hits this? 

A single error log and then nothing is pretty hard to spot. Maybe let's just make this a Fatal()

goldmane/pkg/daemon/daemon.go
Brian-McM: Whoops, what did I previously delete lol.

Brian-McM: I think you're looking at the old commit @caseydavenport

felix/collector/dpstatshelper.go
Brian-McM: I don't think it panics currently though, it just returns. Could you point out where that would happen?

goldmane/pkg/client/client.go
Brian-McM: Ya agreed that one of those two things should happen.

felix/collector/dpstatshelper.go
Brian-McM: Oh are you referring to a send on the closed channel or something like that?

caseydavenport: something like that

caseydavenport: an accidental panic, not an intentional one

goldmane/pkg/emitter/emitter.go
```diff
@@ -92,7 +92,7 @@ func NewEmitter(opts ...Option) *Emitter {
 	return e
 }
 
-func (e *Emitter) Run(stopCh chan struct{}) {
+func (e *Emitter) Run(ctx context.Context) {
```
caseydavenport, line 95: hmmm, this isn't used - probably best to at least add a TODO if we're not going to check the context in the loop (my preference)

felix/collector/dpstatshelper.go
Brian-McM: I don't think it actually panics though, the channel isn't closed when it returns.

So even as is, this change stops the caller of Push from being blocked because the channel is full.

goldmane/pkg/emitter/emitter.go
Brian-McM: Ya I can add that, but I was really just replacing what was currently there (which wasn't used).

caseydavenport: Yeah I understand, but the change is pointless if it doesn't do anything. You can do it now (in which case we remember) or you can do it later as part of your FV PR, in which case we might forget and then we'll see odd flaky hard to trace behavior in the FVs.

felix/collector/dpstatshelper.go
caseydavenport: My point is that we should panic if the gRPC config is invalid so that the user finds out - whether that's via an explicit panic (ideally) or an implicit one (trying to access null) I don't really care

Brian-McM: OK so I'll just change this so that 1) logs Fatal here and 2) the channel is properly closed when it should be in the client so a panic is forced.

That said, my points weren't that I disagree with what you're saying and that we shouldn't do that, I just wasn't intending to change / fix any bugs with these changes (ideally I didn't want bug fixing scope creep in this PR).

goldmane/pkg/emitter/emitter.go
Brian-McM: Added a commit to handle the context.

felix/collector/dpstatshelper.go
caseydavenport: Yeah, I get the desired to not add scope creep but if we don't fix these things now when else are we going to? Takes more time to raise a backlog issue than it does to switch Error to Fatal

caseydavenport (approved):

felix/collector/dpstatshelper.go
Brian-McM: Ya I hear ya.

PR #9947, opened 2025-03-07
[CORE-10975] Add QoS packet rate limit

coutinhop:
## Description

Add packet rate limiting QoS control implementation, using iptables (using '-m limit' and a new mark) and nftables (using 'limit rate over') rules to limit ingress and/or egress packet rate based on workload QoSControls (from pod annotations in k8s).

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Add packet rate limiting QoS control to workloads.
```

## Reminder for the reviewer

Other optional labels:


felix/fv/qos_controls_test.go
```diff
@@ -156,6 +164,106 @@ var _ = infrastructure.DatastoreDescribe(
...
+		Context("With packet rate limits", func() {
+			getRules := func(felixId int) string {
+				out, err := tc.Felixes[felixId].ExecOutput("iptables-save", "-c")
+				log.Infof("iptables-save -c output:\n%v", out)
+				Expect(err).NotTo(HaveOccurred())
+				return out
+			}
+
+			It("should limit packet rate correctly", func() {
+				format.MaxLength = 1000000
```
nelljerram, line 177: I think this is going to affect all following tests.  Personally, I'm a fan of putting `format.MaxLength = 0` somewhere, like in libcalico-go/lib/clientv3/tier_e2e_test.go:
```
func init() {
	// Stop Gomega from chopping off diffs in logs.
	format.MaxLength = 0
}
```

nelljerram: (IOW, my only concern is that it _looks_ like this might be a local-only setting, but in fact it won't be.)

felix/fv/qos_controls_test.go
```diff
@@ -156,6 +164,106 @@ var _ = infrastructure.DatastoreDescribe(
...
+
+				By("Starting iperf3 server on workload 0")
+				serverCmd := w[0].ExecCommand("iperf3", "-s")
+				err := serverCmd.Start()
+				Expect(err).NotTo(HaveOccurred())
+
+				By("Running iperf3 client on workload 1")
+				out, err := w[1].ExecOutput("iperf3", "-c", w[0].IP, "-O5", "-M1000", "-J")
+				Expect(err).NotTo(HaveOccurred())
+				baselineRate, err := getRateFromJsonOutput(out)
```
nelljerram, line 187: Note, this is in bits per sec.

felix/fv/qos_controls_test.go
```diff
@@ -156,6 +164,106 @@ var _ = infrastructure.DatastoreDescribe(
...
+				Expect(err).NotTo(HaveOccurred())
+				log.Infof("iperf client rate with no packet rate limit (bps): %v", baselineRate)
+				// Expect the baseline rate to be much greater (>=100x) the bandwidth that we
+				// would get with the packet rate we are going to configure just below. In
+				// practice we see several Gbps here.
+				Expect(baselineRate).To(BeNumerically(">=", 800000.0*100))
```
nelljerram, line 193: Is there a connection between 800000 here and `IngressPacketRate: 100` below?

felix/fv/qos_controls_test.go
```diff
@@ -156,6 +164,106 @@ var _ = infrastructure.DatastoreDescribe(
...
+				By("Setting 100 packets/s limit for ingress on workload 0 (iperf3 server)")
+				w[0].WorkloadEndpoint.Spec.QoSControls = &api.QoSControls{
+					IngressPacketRate: 100,
+				}
+				w[0].UpdateInInfra(infra)
+				Eventually(tc.Felixes[0].ExecOutputFn("ip", "r", "get", "10.65.0.2"), "10s").Should(ContainSubstring(w[0].InterfaceName))
+
+				By("Waiting for the config to appear in 'iptables-save' on workload 0")
+				// ingress config should be present
+				Eventually(getRules(0), "10s", "1s").Should(And(MatchRegexp(`-A cali-tw-`+regexp.QuoteMeta(w[0].InterfaceName)+` .* -m comment --comment "Mark packets within ingress packet rate limit" -m limit --limit `+regexp.QuoteMeta("100/sec")+` -j MARK --set-xmark 0x\d+/0x\d+`), MatchRegexp(`-A cali-tw-`+regexp.QuoteMeta(w[0].InterfaceName)+` .* -m comment --comment "Drop packets over ingress packet rate limit" -m mark ! --mark 0x\d+/0x\d+ -j DROP`)))
```
nelljerram, line 204: In this check, `iptables-save -c` will be called only once, returning a string, and then that same string will be analysed up to 10 times.  If it fails the first time, it won't make any difference to keep trying for the next 10s.

What you really want is:
```
			getRules := func(felixId int) func() string {
                                return func() {
				out, err := tc.Felixes[felixId].ExecOutput("iptables-save", "-c")
				log.Infof("iptables-save -c output:\n%v", out)
				return out
			}
			}
```

felix/fv/qos_controls_test.go
```diff
@@ -156,6 +164,106 @@ var _ = infrastructure.DatastoreDescribe(
...
+				// egress config should not be present
+				Consistently(getRules(0), "10s", "1s").ShouldNot(Or(MatchRegexp(`-A cali-fw-`+regexp.QuoteMeta(w[0].InterfaceName)+` .* -m comment --comment "Mark packets within egress packet rate limit" -m limit --limit `+regexp.QuoteMeta("100/sec")+` -j MARK --set-xmark 0x\d+/0x\d+`), MatchRegexp(`-A cali-fw-`+regexp.QuoteMeta(w[0].InterfaceName)+` .* -m comment --comment "Drop packets over egress packet rate limit" -m mark ! --mark 0x\d+/0x\d+ -j DROP`)))
+
+				out, err = w[1].ExecOutput("iperf3", "-c", w[0].IP, "-O5", "-M1000", "-J")
+				Expect(err).NotTo(HaveOccurred())
+				ingressLimitedRate, err := getRateFromJsonOutput(out)
+				Expect(err).NotTo(HaveOccurred())
+				log.Infof("iperf client rate with ingress packet rate limit on server (bps): %v", ingressLimitedRate)
+				// Expect the limited rate to be below an estimated desired rate (1000 byte packets * 8 bits/byte * 100 packets/s = 800000 bps)
```
nelljerram, line 213: Ah nice.  Please explain that relation above as well.

felix/fv/qos_controls_test.go
```diff
@@ -156,6 +164,106 @@ var _ = infrastructure.DatastoreDescribe(
...
+				Expect(ingressLimitedRate).To(BeNumerically("<=", 800000.0))
```
nelljerram, line 214: Should there be a bit of leeway here, e.g. `* 1.2`

felix/rules/endpoints_test.go
```diff
@@ -1359,6 +1369,82 @@ var _ = Describe("Endpoints", func() {
 					},
 				}))
 			})
+
+			It("should render a workload endpoint with packet rate limiting QoSControls", func() {
+				format.MaxLength = 1000000
```
nelljerram, line 1374: same comment about this as in FV test

manifests/ocp/custom-resources.yaml
nelljerram, line 1: Shouldn't be in this PR.

nelljerram (changes requested):
Looks great.  Just a few minor points to address.

In terms of dataplane support - something that I should have checked before! -
- Is it correct that we're supporting iptables and nftables, but not eBPF?
- Is that the case for the bandwidth control as well?

felix/fv/qos_controls_test.go
coutinhop: Not exactly a connection, just expecting 100x+ faster traffic with no limits than when limited. I'll add a comment for how we arrived at 800000 here too (as below)

felix/fv/qos_controls_test.go
coutinhop: In my experience, this upperbound was never even close to hitting when I ran the test (the max was typically 640kbps), but I suppose it doesn't hurt to have some margin, will add!

manifests/ocp/custom-resources.yaml
coutinhop: oh, I think this showed up after running `make generate`, will remove

felix/dataplane/driver.go
```diff
@@ -123,6 +123,8 @@ func StartDataplaneDriver(
 		markScratch0, _ = markBitsManager.NextSingleBitMark()
 		markScratch1, _ = markBitsManager.NextSingleBitMark()
 
+		markLimitPacketRate, _ = markBitsManager.NextSingleBitMark()
```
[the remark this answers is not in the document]
coutinhop: Considering we invariably use the mark in 2 consecutive iptables rules, I suppose we could reuse an existing 'scratch' mark instead of allocating a new one, but then I think we'd need a 3rd rule to clear the mark (considering it could/would be used for other purposes as well, as we're only using the mark to "bypass" the iptables `-m limit` module limitation of not being able to drop packets _above_ the limit, we could clear the mark in a rule immediately after the one dropping the packets without the mark). @nelljerram do you think that's a better approach?

coutinhop:
> Looks great. Just a few minor points to address.
> 
> In terms of dataplane support - something that I should have checked before! -
> 
>     * Is it correct that we're supporting iptables and nftables, but not eBPF?
> 
>     * Is that the case for the bandwidth control as well?

Thanks so much! Yes, currently all 3 planned controls will only support iptables and nftables, but not eBPF. This control (and the future connection limit control) relies on iptables/nftables itself, and will need to be implemented differently on eBPF. The bandwidth control has an incompatibility between the `tc qdisc` that the control needs and one that the eBPF dataplane uses to attach BPF programs to interfaces. So yes, in short, this will support iptables/nftables, but not eBPF (at least for now).

felix/fv/qos_controls_test.go
coutinhop: this is noted in the `Info` log a few lines below it: https://github.com/projectcalico/calico/blob/483cc0e5bce25fd56d3dcd5c67ae1ac20a7897fc/felix/fv/qos_controls_test.go#L189

should I add a comment in addition to that?

felix/dataplane/driver.go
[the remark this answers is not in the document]
nelljerram: I'd prefer @fasaxc or @tomastigera to give an opinion about this, as I don't know how careful we need to be with mark bit usage.  One possible situation is that there isn't any real shortage in OSS, but there might be when we get around to merging this into calico-private...

felix/fv/qos_controls_test.go
nelljerram: If you add a comment to explain the 800000 just below, I don't think we need a comment for this point as well.  What was confusing was the **combination** of not explaining the 800000 _and_ not indicating the units of `getRateFromJsonOutput`.

felix/fv/qos_controls_test.go
nelljerram: The `*100` part is fine.  It's the `800000` that needs explanation.  I know now that this is explained a bit further below - but I tend to read code forwards!

felix/fv/infrastructure/infra_k8s.go
```diff
@@ -210,8 +210,14 @@ func RunBPFLog() *containers.Container {
...
 
 func (kds *K8sDatastoreInfra) runK8sApiserver() {
+	// Get current working dir as docker does not accept relative paths for mounting volumes
+	pwd, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
 	args := []string{
 		"-v", os.Getenv("CERTS_PATH") + ":/home/user/certs", // Mount in location of certificates.
+		"-v", pwd + "/../../libcalico-go/config/crd:/crds", // Mount in location of CRDs.
```
coutinhop, line 220: I'm not entirely sure but I was having failures almost every time when trying to copy the CRDs into the container (maybe related to low disk space on my machine), mounting as volume fixed it

PR #9965, opened 2025-03-11
[BPF] rework L2 overlay (vxlan) for performance

tomastigera:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/vxlan_mgr.go
```diff
@@ -682,35 +702,44 @@ func (m *vxlanManager) configureVXLANDevice(
...
-		VxlanId:      m.vxlanID,
-		Port:         m.vxlanPort,
-		VtepDevIndex: parent.Attrs().Index,
-		SrcAddr:      ip.FromString(parentDeviceIP).AsNetIP(),
+		LinkAttrs: la,
+		Port:      m.vxlanPort,
+	}
+
+	if m.dpConfig.BPFEnabled {
+		vxlan.FlowBased = true
```
sridhartigera, line 720: This is set for legacy systems as well. We don't set the tunnel_key in bpf for non co-re which can break vxlan with ebpf on legacy environments.

sridhartigera (changes requested):

sridhartigera (approved):
LGTM

felix/bpf-gpl/fib_co_re.h
```diff
@@ -86,18 +105,111 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
...
+				.family = 10, /* AF_INET6 */
+#else
+				.family = 2, /* AF_INET */
+#endif
+				.tot_len = 0,
+				.ifindex = CALI_F_TO_HOST ? ctx->skb->ingress_ifindex : ctx->skb->ifindex,
+				.l4_protocol = state->ip_proto,
+			};
+#ifdef IPVER6
+		ipv6_addr_t_to_be32_4_ip(fib_params(ctx)->ipv6_src, &state->ip_src);
```
fasaxc, line 134: Different indent here?

felix/dataplane/linux/bpf_route_mgr.go
```diff
@@ -325,7 +325,17 @@ func (m *bpfRouteManager) calculateRoute(cidr ip.CIDR) routes.ValueInterface {
...
+		// Handle RouteType_REMOTE_TUNNEL first as borrowed IPs are going to
+		// have RouteType_LOCAL_WORKLOAD as as well.
+		flags |= routes.FlagsRemoteTunneledHost
+		switch cgRoute.IpPoolType {
+		case proto.IPPoolType_VXLAN:
+			flags |= routes.FlagVXLAN
+		}
+		nodeIP := net.ParseIP(cgRoute.DstNodeIp)
+		route = m.bpfOps.NewValueWithNextHop(flags, ip.FromNetIP(nodeIP))
+	} else if rts&proto.RouteType_LOCAL_WORKLOAD == proto.RouteType_LOCAL_WORKLOAD && !cgRoute.Borrowed /* not ours */ {
```
fasaxc, line 338: I don't quite grok this condition; LOCAL_WORKLOAD & borrowed is a local /32 borrowed from a remote /26; don't we want routes for those?

felix/vxlanfdb/vxlan_fdb.go
```diff
@@ -107,6 +108,12 @@ func WithNetlinkHandleShim(newNetlinkHandle func() (netlinkshim.Interface, error
 	}
 }
 
+func WithARPUpdatesOnly() Option {
```
fasaxc, line 111:
```suggestion
func WithNeighUpdatesOnly() Option {
```
IPv6 uses NDP so maybe we should be more generic here.

felix/vxlanfdb/vxlan_fdb.go
```diff
@@ -77,12 +77,13 @@ type VXLANFDB struct {
...
 	arpEntries     *deltatracker.DeltaTracker[ip.Addr, comparableHWAddr]
 	fdbEntries     *deltatracker.DeltaTracker[comparableHWAddr, ip.Addr]
-	logCxt         *log.Entry
+	logCtx         *log.Entry
 	resyncPending  bool
 	logNextSuccess bool
 	nl             *handlemgr.HandleManager
 
 	newNetlinkHandle func() (netlinkshim.Interface, error)
+	arpOnly          bool
```
fasaxc, line 86: Having a bool that turns off half the function isn't ideal.  Could we split this struct into two, one for ARP and one for the FDB?  (And then only instantiate one of them?)

Or, _should_ we be resyncing ARP too but cleaning it out if we're in the mode that doesn't use it?  In which case you could use the bool to prevent population of the Desired() side of the arpEntries tracker and it'd naturally clean everything up.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1129,21 +1137,41 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
 
 		if config.RulesConfig.VXLANEnabledV6 {
-			vxlanFDBV6 := vxlanfdb.New(netlink.FAMILY_V6, dataplanedefs.VXLANIfaceNameV6, featureDetector, config.NetlinkTimeout)
+			vxlanName := dataplanedefs.VXLANIfaceNameV6
+
+			var (
+				fdbOpts     []vxlanfdb.Option
+				vxlanMgrOps []vxlanMgrOption
+			)
+			if config.BPFEnabled {
```
fasaxc, line 1146: You've got a condition on BTFEnabled too up above, is tat not needed here?

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1129,21 +1137,41 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
+				vxlanName = dataplanedefs.VXLANIfaceNameV4
```
fasaxc, line 1147: Worth a comment saying why we're sharing the V4 device here, otherwise it looks like a bug.

felix/dataplane/linux/vxlan_mgr.go
```diff
@@ -78,6 +79,16 @@ type vxlanManager struct {
 	// Log context
 	logCtx     *logrus.Entry
 	opRecorder logutils.OpRecorder
+
+	maintainIPOnly bool
```
fasaxc, line 83: Worth commenting this to explain why it's needed

fasaxc (changes requested):

felix/dataplane/linux/bpf_route_mgr.go
tomastigera: yes, we want to handle borrowed but since the flags are `REMOTE_TUNNEL | REMOTE_WORKLOAD | LOCAL_WORKLOAD` it gets handled in the `REMOTE_TUNNEL` above and here we should handle `LOCAL_WORKLOAD` only if it is not borrowed. Bit of a mess give what what l3_route_resolver reports, but this explanation from @caseydavenport makes some sense to me:

- It will get REMOTE_TUNNEL if it's not a RefTypeWEP and it's not allocated to the local node
- REMOTE_WORKLOAD in this case I think is just saying "This is from within an IPAM block and the allocation is not local to this node", so a duplicate of REMOTE_TUNNEL basically.
- LOCAL_WORKLOAD is likely coming from the IPAM block, meaning that the block this IP belongs to is local to this node?
- Borrowed is saying that the IP is on a host that is not the same as the block that contributed it...
-

felix/dataplane/linux/int_dataplane.go
tomastigera: yes

felix/vxlanfdb/vxlan_fdb.go
tomastigera: It is the opposite, we always need arp/neigh, we do not need fdb in bpf mode :eyes:

tomastigera: will merge as is and created a ticket for this

PR #9974, opened 2025-03-12
Add Filters endpoint

Brian-McM:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


whisker-backend/pkg/apis/v1/flows.go
```diff
@@ -196,3 +226,14 @@ type PolicyHit struct {
 	RuleIndex   int64      `json:"rule_index"`
 	Trigger     *PolicyHit `json:"trigger"`
 }
+
+type FlowFilterHintsRequest struct {
+	// Type represents the filter type to get hints for.
+	// Note that this is a pointer because the 0 value of the Filter type fails the required check.
```
caseydavenport, line 232: What check does it fail?

caseydavenport: Do we need a `FilterTypeUnspecified` like we have for other enums?

caseydavenport (commented):
One minor question

whisker-backend/pkg/apis/v1/flows.go
Brian-McM: It fails the "required" validation check. We could use a `FilterTypeUnspecified`, that would force the required validation check without requiring this to be a pointer.

Would you rather I do that?

caseydavenport: Oh, "the required check" - I read this as "a required check" :laughing: 

Uhh, I think from the gRPC API perspective it's useful to have an explicit Unspecified option - otherwise, when unspecified it just defaults to the zero-value'd option. 

But I'm not too bothered either way. Was just curious if we were working around a limitation somewhere else.

caseydavenport: Should it just not be marked as "required"?

Brian-McM: It is required though from the BFF standpoint, I don't think it makes sense to default to a filter type here.

> Was just curious if we were working around a limitation somewhere else.

Yaaaaa I suppose we are...

caseydavenport (approved):

PR #9978, opened 2025-03-13
Add Goldmane to manifests by default

Brian-McM:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/charts/tigera-operator/templates/crs/custom-resources.yaml
+++ b/charts/tigera-operator/templates/crs/custom-resources.yaml
@@ -28,6 +28,13 @@ spec:
 ---
 
 apiVersion: operator.tigera.io/v1
+kind: Goldmane
+metadata:
+  name: default
+
+---
+
+apiVersion: operator.tigera.io/v1
 kind: Whisker
 metadata:
   name: default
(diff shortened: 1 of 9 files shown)


charts/tigera-operator/templates/crs/custom-resources.yaml
```diff
@@ -27,6 +27,13 @@ spec:
 {{ if .Values.whisker.enabled }}
 ---
 
+apiVersion: operator.tigera.io/v1
```
caseydavenport, line 30: We need to follow the example of the other CRs here and provide a mechanism to enable / disable via the helm chart values.yaml

Brian-McM: OK, I was thinking we could tie this to the whisker flag, but I'm fine to add a different flag.

caseydavenport: One point of having them as separate CRs is that it's possible to deploy Goldmane without Whisker, so we probably want to show that here.

caseydavenport (approved):

PR #9982, opened 2025-03-14
Goldmane emitter improvements

caseydavenport:
## Description

- Reload certificates when they change. This allows us to mount new
  certs without restarting the pod and losing in-memory state.
- Switch config model to be time-based, and adjust the default emission
  window to reduce latency.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/goldmane/pkg/daemon/daemon.go
+++ b/goldmane/pkg/daemon/daemon.go
@@ -63,14 +63,17 @@ type Config struct {
 	// from each node in the cluster.
 	AggregationWindow time.Duration `json:"aggregation_window" envconfig:"AGGREGATION_WINDOW" default:"15s"`
 
-	// The number of buckets to combine when pushing flows to the sink. This can be used to reduce the number
-	// buckets combined into time-aggregated flows that are sent to the sink.
-	NumBucketsToCombine int `json:"num_buckets_to_combine" envconfig:"NUM_BUCKETS_TO_COMBINE" default:"20"`
+	// EmitAfterSeconds is the number of seconds from flow receipt to wait before a flow becomes eligible to be
+	// pushed to the emitter. Increasing this value will increase the latency of emitted flows, while a smaller
+	// value will cause the emitter to emit potentially incomplete flows.
+	//
+	// This must be a multiple of the aggregation window.
+	EmitAfterSeconds int `json:"emit_after_seconds" envconfig:"EMIT_AFTER_SECONDS" default:"30"`
 
-	// PushIndex is the index of the bucket which triggers pushing to the emitter. A larger value
-	// will increase the latency of emitted flows, while a smaller value will cause the emitter to emit
-	// potentially incomplete flows.
-	PushIndex int `json:"push_index" envconfig:"PUSH_INDEX" default:"30"`
+	// EmitterAggregationWindow is the duration of time across which flows are aggregated before being pushed to the emitter.
+	// This must be a multiple of the aggregation window.
+	// Emitted flows will be aggregated from "now-EmitAfterSeconds-EmitterAggregationWindow" to "now-EmitAfterSeconds"
+	EmitterAggregationWindow time.Duration `json:"emitter_aggregation_window" envconfig:"EMITTER_AGGREGATION_WINDOW" default:"5m"`
 }
 
 func ConfigFromEnv() Config {
@@ -116,8 +119,8 @@ func Run(ctx context.Context, cfg Config) {
 	// Track options for log aggregator.
 	aggOpts := []aggregator.Option{
 		aggregator.WithRolloverTime(cfg.AggregationWindow),
-		aggregator.WithBucketsToCombine(cfg.NumBucketsToCombine),
-		aggregator.WithPushIndex(cfg.PushIndex),
+		aggregator.WithBucketsToCombine(int(cfg.EmitterAggregationWindow.Seconds()) / int(cfg.AggregationWindow.Seconds())),
+		aggregator.WithPushIndex(cfg.EmitAfterSeconds / int(cfg.AggregationWindow.Seconds())),
 	}
 
 	if cfg.PushURL != "" {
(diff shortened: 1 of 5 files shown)


goldmane/pkg/emitter/http_client.go
```diff
@@ -71,20 +73,95 @@ func newHTTPClient(caCert, clientKey, clientCert, serverName string) (*http.Clie
...
 func newEmitterClient(url, caCert, clientKey, clientCert, serverName string) (*emitterClient, error) {
+	// Create an initial HTTP client, and a function to help encapsualte the reload logic.
 	client, err := newHTTPClient(caCert, clientKey, clientCert, serverName)
 	if err != nil {
 		return nil, err
 	}
-	return &emitterClient{url: url, client: client}, nil
+
+	updChan := make(chan struct{})
+	getClient := func() (*http.Client, error) {
```
Brian-McM, line 83: I was hoping that this could be something more generic and put in a lib module. This is some pretty useful functionality that could be abstracted to recreate an object based on changes to the file system.

A crude sketch would be something like:
```
type RegenerativeObject[E] interface {
  Get() (E, error)
}

func NewRegenerativeObject[E](newF func() (E, error), fileToWatch string) {
...
}
```

I suppose it could be even more generic and not require a list of files (require some other signaling interface, in which there is a FileWatching implementation).

goldmane/pkg/emitter/http_client.go
```diff
@@ -71,20 +73,95 @@ func newHTTPClient(caCert, clientKey, clientCert, serverName string) (*http.Clie
...
+
+	return func() {
+		// If we exit this function, make sure to close the file watcher and update channel.
+		defer fileWatcher.Close()
+		defer close(updChan)
+		for {
+			select {
+			case event, ok := <-fileWatcher.Events:
+				if !ok {
+					logrus.Info("file watcher closed")
```
Brian-McM, line 135: Nit this could be added as a defer statement.

Brian-McM (changes requested):

goldmane/pkg/daemon/daemon.go
```diff
@@ -63,14 +63,17 @@ type Config struct {
...
-	// PushIndex is the index of the bucket which triggers pushing to the emitter. A larger value
-	// will increase the latency of emitted flows, while a smaller value will cause the emitter to emit
-	// potentially incomplete flows.
-	PushIndex int `json:"push_index" envconfig:"PUSH_INDEX" default:"30"`
+	// EmitAfterSeconds is the number of seconds from flow receipt to wait before a flow becomes eligible to be
+	// pushed to the emitter. Increasing this value will increase the latency of emitted flows, while a smaller
+	// value will cause the emitter to emit potentially incomplete flows.
+	//
+	// This must be a multiple of the aggregation window.
+	EmitAfterSeconds int `json:"emit_after_seconds" envconfig:"EMIT_AFTER_SECONDS" default:"30"`
```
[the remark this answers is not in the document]
caseydavenport: > 	// This must be a multiple of the aggregation window.

Yep :p

goldmane/pkg/emitter/http_client.go
caseydavenport: @Brian-McM `RegenerativeObject` feels way too abstract for what we need here.

A library function that implements a file watcher is fine. But that's what `fsnotify` already does so I don't think we need to re-implement that.

Writing common libraries is a useful tool, but we need to be really careful we're not just taking everything and turning it into an abstracted concept of itself on the chance that we'll need it to be that generic. It makes extra work for us now in order to figure out what the abstracted representation should be, when we may not ever use this for another use-case (or if the use case arrives, we may have gotten the abstraction wrong).

Ultimately, the fact that this client reloads its CA cert is not its defining characteristic and it's losing sight of the big picture if we try to define it by that fact. 

I'm fine to write a generic HTTP client implementation that wraps some of this, that feels like something that might see more use. Just let me know if you think that is the path to take.

Brian-McM: Regardless of whether or not you do this, I can't agree that this isn't a good abstraction.

I think an abstraction like that is extremely useful, and we'll probably be able to use it in multiple place. This is nothing specific to http, and there's no need to make it so. You just want a new object based off of some signaling change. I'm not sure why we'd need a specific http implementation, all we want to do is create a new "object" (in this case an http client) based on some external change, in this case a file changed.

> Writing common libraries is a useful tool, but we need to be really careful we're not just taking everything and turning it into an abstracted concept of itself on the chance that we'll need it to be that generic.

Why this is a _**good**_ abstraction is because any implementation could use it as the heart of the code, and it takes out any channel / go routine management. It can be used here, in any other place needed. You can make the watchFiles something reusible (it returns a channel) and instead of using a getClient function you use the "RegenerativeObject" which has a function to return that exact type.

The caller wouldn't need to worry about setting anything up, just that when there are changes that happen they can know that the object is recreated based off the function given. They don't need to start or manage monitor functions or routines, which is a good thing when that can be abstracted out properly (which it could in this case) because devs who want this ability don't need to know about what's under the hood (and if they want to loop into the abstraction).

> Ultimately, the fact that this client reloads its CA cert is not its defining characteristic and it's losing sight of the big picture if we try to define it by that fact.

This is exactly what the abstraction I suggested doesn't focus on. All you need to do is call a recreation based on a signal.

> I'm fine to write a generic HTTP client implementation that wraps some of this, that feels like something that might see more use. Just let me know if you think that is the path to take.

I'm not really sure what a generic http client implementation would like different to the concept I proposed, except having the generic type be exactly an http.Client, which shows to me that this might as well be a generic fucnction (where you could also implement `interface RegenerativeHttpClient RegenerativeObject[http.Client]` if you felt like doing that.

So feel free not to do this, I'm not blocking on this, but this is a useful abstraction and sometime very generic abstraction aren't forced and bad, especially when they take complexity out of something that is in heart very generic (recreate the object when I get a signal, and do so in a thread safe manner).

Brian-McM (approved):

PR #9986, opened 2025-03-14
Fix a few bugs in policy hit parsing to/from string, and stop panicking

caseydavenport:
## Description

We didn't have a way of bubbling up errors if we were given invalid
policy hits (or there was a bug in the parsing). This PR changes that so
that, and also adds explicit validation to the input and output to
ensure we're not accepting bad data.

In the process, a couple of staged policy parsing bugs were found and
fixed.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/goldmane/proto/methods.go
+++ b/goldmane/proto/methods.go
@@ -20,26 +20,53 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
+	"k8s.io/kubernetes/pkg/apis/core/validation"
 )
 
 // ToString converts a PolicyHit struct into a string label.
 // TODO: This is a temporary solution - we should be pushing the structured PolicyHit representation
 // further down the stack, rather than converting between string / struct.
-func (h *PolicyHit) ToString() string {
+func (h *PolicyHit) ToString() (string, error) {
 	// Format is: "<policy index>|<tier>|<namePart>|<action>|<rule index>"
 	tmpl := "%d|%s|%s|%s|%d"
 
+	if err := h.Validate(); err != nil {
+		return "", err
+	}
+
 	var namePart string
+	var err error
 	if h.Kind == PolicyKind_EndOfTier {
-		namePart = makeNamePart(h.Trigger)
+		namePart, err = makeNamePart(h.Trigger)
 	} else {
-		namePart = makeNamePart(h)
+		namePart, err = makeNamePart(h)
+	}
+	if err != nil {
+		return "", err
 	}
 
 	// Convert action from enum to string.
 	action := strings.ToLower(Action_name[int32(h.Action)])
 
-	return fmt.Sprintf(tmpl, h.PolicyIndex, h.Tier, namePart, action, h.RuleIndex)
+	return fmt.Sprintf(tmpl, h.PolicyIndex, h.Tier, namePart, action, h.RuleIndex), nil
+}
+
+func (h *PolicyHit) Validate() error {
+	if _, ok := Action_name[int32(h.Action)]; !ok {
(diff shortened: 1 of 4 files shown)


goldmane/proto/methods.go
```diff
@@ -20,26 +20,55 @@ import (
...
+	return fmt.Sprintf(tmpl, h.PolicyIndex, h.Tier, namePart, action, h.RuleIndex), nil
+}
+
+func (h *PolicyHit) Validate() error {
+	if _, ok := Action_name[int32(h.Action)]; !ok {
+		return fmt.Errorf("unexpected action: %v", h.Action)
+	}
+
+	switch h.Kind {
+	case PolicyKind_GlobalNetworkPolicy, PolicyKind_AdminNetworkPolicy, PolicyKind_BaselineAdminNetworkPolicy:
```
dimitri-nicolo, line 62: It looks like we're missing `PolicyKind_StagedGlobalNetworkPolicy` from this case

goldmane/proto/methods.go
```diff
@@ -54,7 +83,8 @@ func (h *PolicyHit) fields() logrus.Fields {
 	}
 }
 
-func makeNamePart(h *PolicyHit) string {
+func makeNamePart(h *PolicyHit) (string, error) {
+	logrus.WithFields(h.fields()).Debug("Generating name part from policy hit")
 	var namePart string
 	switch h.Kind {
```
dimitri-nicolo, line 89: For consistency in the naming scheme we've chosen for kubernetes network policy, can we define the kind to `PolicyKind_KubernetesNetworkPolicy` instead of `PolicyKind_NetworkPolicy`?

goldmane/proto/methods.go
```diff
@@ -66,9 +96,9 @@ func makeNamePart(h *PolicyHit) string {
 	case PolicyKind_StagedKubernetesNetworkPolicy:
 		namePart = fmt.Sprintf("%s/staged:knp.default.%s", h.Namespace, h.Name)
 	case PolicyKind_StagedGlobalNetworkPolicy:
-		namePart = fmt.Sprintf("staged:%s.%s", h.Tier, h.Name)
+		namePart = fmt.Sprintf("%s.staged:%s", h.Tier, h.Name)
 	case PolicyKind_StagedNetworkPolicy:
-		namePart = fmt.Sprintf("%s/staged:%s.%s", h.Namespace, h.Tier, h.Name)
+		namePart = fmt.Sprintf("%s/%s.staged:%s", h.Namespace, h.Tier, h.Name)
 	case PolicyKind_AdminNetworkPolicy:
 		namePart = fmt.Sprintf("kanp.adminnetworkpolicy.%s", h.Name)
```
dimitri-nicolo, line 103: Even though it might be less readable, I'm in favor of utilizing constants for `knp`, kanp`, `baselineadminnetworkpolicy` etc.. Might help avoid misspelling these if we were to add any other checks down the line

goldmane/proto/methods.go
caseydavenport: I think we want to leave it like so - the NetworkPolicy kind is the native Kubernetes type, so should get the name :)

goldmane/proto/methods.go
caseydavenport: Yeah... might just be personal preference but I hope that the unit tests will prevent that sort of bug from showing up.

The constants make it so much less readable that I find it easier to write bugs into the code that way. I'll see if I can do it cleanly though....

dimitri-nicolo (approved):
lgtm

PR #9995, opened 2025-03-17
Add dynamic enable/disable of flow emission

caseydavenport:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


goldmane/pkg/emitter/emitter.go
```diff
@@ -92,6 +115,31 @@ func NewEmitter(opts ...Option) *Emitter {
...
+}
+
+func (e *Emitter) Disable() {
+	e.enabledLock.Lock()
+	defer e.enabledLock.Unlock()
+	e.enabled = make(chan struct{})
+}
+
+func (e *Emitter) isEnabled() bool {
+	e.enabledLock.Lock()
```
caseydavenport, line 130: not actually sure I need this locking... but we do potentially reassign `e.enabled` so it seems right...

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -228,16 +240,14 @@ func (a *LogAggregator) maybeEmitFlows() {
...
-		// We've already pushed this bucket, so we can skip it. We'll emit the next flow once
-		// bucketsToAggregate buckets have been rolled over.
-		logrus.Debug("Delaying flow emission, no new flows to emit")
-		return
-	}
-
-	if len(flows.Flows) > 0 {
-		a.sink.Receive(flows)
+	// Iterate the collections in reverse order, as we want to emit the oldest data first.
+	collections := a.buckets.FlowCollections()
```
Brian-McM, line 244: Since it looks like you want the BucketRing calculate what needs to be sent to the sync, it might be better to create a `PushFlowCollections` function. Something like:
```
type flowSink interface {
  Receive(...)
}

func (r *BucketRing) PushFlowCollections(sink flowSink) {
  collection := .....
  for i := len(collections) - 1; i >= 0; i-- {
    c := collections[i]
    if len(c.Flows) > 0 {
      sink.Receive(c)
      for _, b := range fc.buckets {
        b.Pushed = true
      }
    }
  }
}
```

I think it simplifies the interaction between the two interfaces since the BucketRing is supposed to know what to push (so might as well just give it the thing it should use to push), ensures that no follow up is need after PushFlowCollections is called for "completion", and can mark the buckets as pushed as each one is being pushed (that implementation above doesn't do that but it could).

goldmane/pkg/aggregator/bucketing/bucket_ring.go
```diff
@@ -274,15 +274,45 @@ func (r *BucketRing) nowIndex() int {
...
-// for statistics to settle down before publishing.
-func (r *BucketRing) FlowCollection() *FlowCollection {
+func (r *BucketRing) FlowCollections() []*FlowCollection {
 	// Determine the newest bucket in the aggregation - this is always pushAfter buckets back from "now".
 	endIndex := r.indexSubtract(r.nowIndex(), r.pushAfter)
 	startIndex := r.indexSubtract(endIndex, r.bucketsToAggregate)
 
+	// maxCollections defines the maximum number of collections to build for any given call. Each collection is
+	// 5 minutes long, so 12 collections will cover a full hour.
+	maxCollections := 12
```
Brian-McM, line 284: What's the reason for needing a maximum number here?

Also, isn't that 5 minute window configurable (so should this be based off a configurable window)?

goldmane/pkg/aggregator/bucketing/bucket_ring.go
```diff
@@ -274,15 +274,45 @@ func (r *BucketRing) nowIndex() int {
...
+	}
+	logrus.WithField("num", len(collections)).Debug("Built flow collections to emit")
+	return collections
+}
+
+// FlowCollection returns a collection of flows to emit, or nil if we are still waiting for more data.
+// The BucketRing builds a FlowCollection by aggregating flow data from across a window of buckets. The window
+// is a fixed size (i.e., a fixed number of buckets), and starts a fixed period of time in the past in order to allow
+// for statistics to settle down before publishing.
+func (r *BucketRing) FlowCollection(startIndex, endIndex int) *FlowCollection {
```
Brian-McM, line 315: Could you rename this to FlowCollectionsToEmit or something like that? It's not clear by the name that it wouldn't return buckets that were already emitted.

goldmane/pkg/daemon/config.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/calico/goldmane/pkg/aggregator"
+	"github.com/projectcalico/calico/goldmane/pkg/internal/utils"
+)
+
+type goldmaneFileConfig struct {
```
Brian-McM, line 28: Maybe this is preference, but to me configuration is easier to understand when it's all together so I would advocate for adding this logic to the config object that we already have.

I don't think we need multiple config structures just because there's multiple sources.

goldmane/pkg/aggregator/bucketing/bucket_ring.go
caseydavenport: Yeah, it probably should be adjusted to at least be based on the configurable window.

The original reason I added this was to avoid looping forever (since it's a ring!) 

Other ways to do this would just be to break when we loop around one complete ring, which is probably what we want.

goldmane/pkg/daemon/config.go
Brian-McM: I wasn't thinking this should also be configurable via env variables as well, including it in the config struct doesn't mean that.

I don't think the config struct should represent what the source of the configuration IMO, it's just everything that's needed to configure goldmane put into a struct that so the dameon can configure it. I don't feel there's a need to have multiple config structs to configure goldmane.

This said, leave it if you don't agree or even if you just don't want to do it, this is probably close to splitting hairs and in general preference (especially when there's just two structs).

goldmane/pkg/emitter/http_client.go
```diff
@@ -80,7 +81,19 @@ func newEmitterClient(url, caCert, clientKey, clientCert, serverName string) (*e
 		return nil, err
 	}
 
-	updChan := make(chan struct{})
+	// We only reload the client when we need to use it, but the file watcher may generate events at any time - sometimes
+	// several in quick succession. To handle this, we use a small deduplicator goroutine to ensure we're servicing the
+	// channel.
+	updChan := make(chan struct{}, 1)
+	reloadClient := make(chan struct{})
+	go func() {
```
Brian-McM, line 89: Instead of putting this here, what would you think of the file watcher just keeping a hash of file and only sending notifications if the contents has changed?

I think this would be a bit more robust and require less from any of you're code that uses it.

This current implementation doesn't really help deduplicate if the reader side of the channel reads the struct off before the next update happens anyway.

goldmane/pkg/daemon/config.go
```diff
@@ -0,0 +1,115 @@
...
+		f.aggregator.SetSink(f.sink)
+	}
+	logrus.Info("Sink manager started")
+
+	// Start the file watch.
+	go f.watchFn(ctx)
+
+	go func() {
+		// If enablement changes, update the sink.
+		for range f.upd {
```
Brian-McM, line 84: I think given that each update here would trigger a call to emit the flows it would be worth to ensure we're not getting duplicate or unnecessary updates, i.e. ensuring that updates are only received if the file contents have changed.

You could also just track if it's already enabled or not, but this is the second place where we ideally don't want unecessary updates so fixing the issue at the source seems pragmatic.

caseydavenport: The call to emit flows is already de-duplicated, so if there's nothing new to do then no flows will be emitted and minimal cycles will be used. I don't think there's a lot of value de-duplicating here - it will just introduce opportunities for us to get something wrong.

The code that checks if flows need to be emitted already needs to be capable of deduplicating regardless of how this loop behaves.

goldmane/pkg/emitter/http_client.go
caseydavenport: The way that configmap hot reloading works involves multiple temporary files with randomly generated names that all get copied around / deleted / updated. It results in a few successive fsnotify events of varying types, which get de-duplicated here.

We could have the file-watcher try to do that itself by using memory to cache things, but it will be pretty awkward.

If anything, the hash checking should be done on this side of the channel, where we know exactly which files we care about (the file watcher watches a directory by necessity). 

But even that I think is overkill. In normal operation this is something that will happen once, and there's not really a downside to us getting multiple events here. The upside is less code.

goldmane/pkg/daemon/config.go
Brian-McM: I can't say I really understand why you'd want the file watcher to send events that don't need to be emitted because nothing has actually changed, especially given that you're already writing de duplication logic outside the watcher (i.e. in the emitter/http_client) where you could just make this inherent in it.

This just seems like requiring any user of the file watcher to fix a flaw with it.

Anyway, I'll leave it at that.

Brian-McM (approved):

PR #8949, opened 2024-06-27
Replace gogoprotobuf with golang protobuf

hda2:
## Description

This change replaces gogoprotobuf with golang protobuf for Calico components. Gogoprotobuf is [deprecated](https://x.com/awalterschulze/status/1584553056100057088) since Oct 2022 and is not updated. Replacing gogoprotobuf with golang protobuf can bring in new protobuf features/fixes and reduce outdated package dependencies.

To work with golang protobuf, we need (1) a [calico/go-build change](https://github.com/projectcalico/go-build/pull/574) to include protoc in the toolchain image and (2) workaround an issue that protobuf generated types are not comparable. 

## Related issues/PRs

Require https://github.com/projectcalico/go-build/pull/574 to be merged and pick to release branches first.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #9996, opened 2025-03-17
Fix bpf endpoint manager panic on deleting hostendpoints.

sridhartigera:
## Description

As part of protobuf changes in https://github.com/projectcalico/calico/pull/8949, ```hostIfaceToEpMap``` was changed from map[string]proto.HostEndPoint to map[string]*proto.Hostendpoint. As a result, we get a `nil` value when the key is not present and bpf ep manager crashes when trying to dereference it.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -3254,6 +3254,9 @@ func (m *bpfEndpointManager) OnHEPUpdate(hostIfaceToEpMap map[string]*proto.Host
 }
 
 func (m *bpfEndpointManager) addHEPToIndexes(ifaceName string, ep *proto.HostEndpoint) {
+	if ep == nil {
+		return
+	}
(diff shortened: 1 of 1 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -3254,6 +3254,9 @@ func (m *bpfEndpointManager) OnHEPUpdate(hostIfaceToEpMap map[string]*proto.Host
 }
 
 func (m *bpfEndpointManager) addHEPToIndexes(ifaceName string, ep *proto.HostEndpoint) {
+	if ep == nil {
```
fasaxc, line 3257: Looks like `removeHEPFromIndexes` uses a different fix (it calls `GetTiers()` et al, which handle the `nil`).  Please can you bring them into sync?

I think I find your way a bit clearer (and less subject to an external change breaking us).

PR #10004, opened 2025-03-19
Remove fsnotify dependency

caseydavenport:
## Description

fsnotify is no longer maintained, and had some problems watching files
that were updated in place by kubernetes - e.g., configmap projection.

Instead, use a polling file watcher.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/goldmane/pkg/internal/utils/files.go
+++ b/goldmane/pkg/internal/utils/files.go
@@ -18,6 +18,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"time"
 
@@ -37,6 +38,12 @@ func WatchFilesFn(updChan chan struct{}, interval time.Duration, files ...string
 
 	for _, file := range files {
 		logrus.WithField("file", file).Debug("Watching file")
+		hash, err := getFileHash(file)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return nil, err
+		} else if hash != "" {
+			cached[file] = hash
+		}
 	}
 
 	return func(ctx context.Context) {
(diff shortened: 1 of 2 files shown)


goldmane/pkg/internal/utils/files.go
```diff
@@ -15,73 +15,80 @@ package utils
...
-			case <-ctx.Done():
+			if ctx.Err() != nil {
 				return
-			case event, ok := <-fileWatcher.Events:
-				if !ok {
-					logrus.Warn("File watcher events channel closed")
-					return
+			}
+
+			for _, file := range files {
```
Brian-McM, line 52: What would you think of calculating all the hashes first and then sending a single update if any of them has changed?

goldmane/pkg/internal/utils/files.go
```diff
@@ -15,73 +15,80 @@ package utils
...
+					continue
 				}
-				if !chanutil.WriteNonBlocking(updChan, struct{}{}) {
-					logrus.WithField("event", event).Debug("file notification channel is full, dropping update")
+
+				// Hash the file.
+				hash := fmt.Sprintf("%x", sha256.Sum256(data))
+
+				// Compare the hash to the cached hash.
+				if cur, ok := cached[file]; !ok {
```
Brian-McM, line 72: Do we really want to skip a notification if a file was added? I think it could be that it's not configured to start, maybe a blank config map was mounted, then it's added later when it's configured?

We could start mounting empty secrets / config maps to avoid restarts for things like this.

goldmane/pkg/internal/utils/files.go
```diff
@@ -15,73 +15,80 @@ package utils
...
-				if !ok {
-					logrus.Warn("File watcher events channel closed")
-					return
+			}
+
+			for _, file := range files {
+				logCtx := logrus.WithField("file", file)
+
+				// Read the file.
+				data, err := os.ReadFile(file)
```
Brian-McM, line 56: If we're going for a bit of optimizations then the following approach would ensure we don't load the entire file to get the hash:

```
file, err := os.Open(filePath)
if err != nil {
	....
}

defer file.Close()
hash := sha256.New()
if _, err := io.Copy(hash, file); err != nil {
	....
}
```

goldmane/pkg/internal/utils/files.go
caseydavenport: Yeah, I had this thought as well and we probably do want to generate an event. But this made sure that the existing UTs passed :sweat_smile:

Brian-McM (approved):

PR #10015, opened 2025-03-19
[CORE-11082] Install Calico in OCP with eBPF mode as the default

lucastigera:
## Description

This PR creates/changes manifests to install Calico with the eBPF dataplane in an OpenShift cluster by default. 

Changes on the user docs might look like this:

### Install with eBPF mode

- Set the `KUBERNETES_SERVICE_HOST` attribute in `01-configmap-kubernetes-services-endpoint.yaml` file. The values can be accessed using the command below.
```
grep -E '"BaseDomain"|"ClusterName"' .openshift_install_state.json -B 1 -A 1
```
The output would look like this:
```
    "*installconfig.baseDomain": {
        "BaseDomain": "base.domain.com",
        "Publish": ""
--
    "*installconfig.clusterName": {
        "ClusterName": "ocp-cluster-name"
    },

```

You can set the values in the file using the command below:
```
CLUSTER_NAME=`grep -o '"ClusterName": "[^"]*' .openshift_install_state.json | sed 's/"ClusterName": "//'` && \
sed -i "s|<cluster_name>|$CLUSTER_NAME|g" manifests/01-configmap-kubernetes-services-endpoint.yaml && \
BASE_DOMAIN=`grep -o '"BaseDomain": "[^"]*' .openshift_install_state.json | sed 's/"BaseDomain": "//'` && \
sed -i "s|<base_domain>|$BASE_DOMAIN|g" manifests/01-configmap-kubernetes-services-endpoint.yaml
```

- Include `spec.template.spec.dnsConfig.nameservers` value to resolve the apiserver DNS in `02-tigera-operator.yaml` file:
  - For clusters in AWS, the DNS server address is 169.254.169.253.

```
dnsConfig:
  nameservers:
  - 169.254.169.253
```

### Install with Iptables mode

- Change `linuxDataplane` to `Iptables` in `manifests/01-cr-installation.yaml` file.
- Set `deployKubeProxy` to `true` in `cluster-network-operator.yaml` file.
- Comment out the `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` lines in `01-configmap-kubernetes-services-endpoint.yaml` file.
```
sed -i 's/^\(\s*linuxDataplane:\s*\)BPF/\1Iptables/' manifests/01-cr-installation.yaml && \
sed -i 's/^\(\s*deployKubeProxy:\s*\)false/\1true/' manifests/cluster-network-operator.yaml && \
sed -i '/^\s*KUBERNETES_SERVICE_HOST:/s/^\(\s*\)/#\1/' manifests/01-configmap-kubernetes-services-endpoint.yaml && \
sed -i '/^\s*KUBERNETES_SERVICE_PORT:/s/^\(\s*\)/#\1/' manifests/01-configmap-kubernetes-services-endpoint.yaml
```

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
- New manifests to install Calico with the eBPF dataplane in an OpenShift cluster.
```

## Reminder for the reviewer

Other optional labels:


release/pkg/manager/calico/manager.go
```diff
@@ -1051,7 +1057,7 @@ Attached to this release are the following artifacts:
 - {release_tar}: container images, binaries, and kubernetes manifests.
 - {calico_windows_zip}: Calico for Windows.
 - {helm_chart}: Calico Helm v3 chart.
-- ocp.tgz: Manifest bundle for OpenShift.
```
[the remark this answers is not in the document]
lucastigera: Good question, @radTuti .
Since we want to prioritize the eBPF installation over Iptables, I thought it would be a good idea to only "advertise" the `ocp-bpf.tgz`. It’s not a must-have change, though. If you think it doesn’t make sense, I can include both files, as you suggested. What are your thoughts?"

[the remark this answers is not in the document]
sridhartigera: We will still need ocp.tgz for upgrade.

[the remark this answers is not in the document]
lucastigera: I didn’t mean removing the file from the bucket or from the docs. My thought was to avoid including it in the release notes.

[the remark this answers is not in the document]
danudey: IIRC the decision was made in v3.30 to use the eBPF dataplane by default on OCP, meaning that users who want to continue to use the current dataplane would want `ocp.tgz`, but that we will be directing users to `ocp-bpf.tgz` by default.

Unless there's an easy way to make an `ocp.tgz` that installs eBPF by default but doesn't break existing clusters for upgrade, it seems like two archives might be necessary.

manifests/generate.sh
```diff
@@ -110,6 +110,8 @@ ${HELM} template \
 # The first two lines are a newline and a yaml separator - remove them.
 find ocp/tigera-operator -name "*.yaml" | xargs sed -i -e 1,2d
 mv $(find ocp/tigera-operator -name "*.yaml") ocp/ && rm -r ocp/tigera-operator
+# Copy the OCP manifests to a new directory for BPF dataplane, without overwriting existing files in the destination.
+cp -rn ocp/ ocp-bpf/
```
danudey, line 114: This should be `cp -r --update=none`, per the warning issued by GNU coreutils 9.4:

> cp: warning: behavior of -n is non-portable and may change in future; use --update=none instead

danudey (commented):
Minor change in `cp` command for future-proofing.

release/pkg/manager/calico/manager.go
```diff
@@ -1051,7 +1057,7 @@ Attached to this release are the following artifacts:
...
+- ocp-bpf.tgz: Manifest bundle for OpenShift.
```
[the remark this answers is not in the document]
danudey: We could also label the older version differently, like:

- `ocp-iptables.tgz` - Manifest bundle for OpenShift using classic iptables/nftables dataplane
- `ocp.tgz` - Manifest bundle for new OpenShift installs or installs using eBPF

I'm not sure about the wording, but something like that. Since we're trying to make eBPF the default, we might not want to call it out as "OCP but with eBPF" but rather call out the old one as "OCP but with iptables".

Seems like a product/marketing decision though.

manifests/ocp/01-configmap-kubernetes-services-endpoint.yaml
```diff
@@ -0,0 +1,11 @@
...
+# installing Calico on OpenShift with eBPF dataplane.
+# If you are installing Calico on OpenShift with the Iptables dataplane, you can remove it.
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: kubernetes-services-endpoint
+  namespace: tigera-operator
+data:
+  KUBERNETES_SERVICE_HOST: "api.<cluster_name>.<base_domain>"
+  KUBERNETES_SERVICE_PORT: "6443"
```
sridhartigera, line 11: Better to comment these 2 lines, as we will anyways want the user to fill it out.

lucastigera: IMO, it didn’t make sense to have the two other files "ready for BPF" while this one had the attributes commented out. It felt a bit off to me. Despite the need to change it to include the value, it initially seemed more concise.

The downside of this approach is shipping a document with an "invalid/incomplete" configuration. If that's a problem, then I agree leaving it commented out is a better option.

Even though, if the user installs Calico with these lines commented out or with the default (invalid) value, it will fail anyway.

manifests/ocp/01-cr-installation.yaml
```diff
@@ -4,6 +4,10 @@ kind: Installation
 metadata:
   name: default
 spec:
+# Configures Calico networking using the eBPF dataplane. If you want to change the dataplane
+# to iptables, you should set the linuxDataplane parameter below to "Iptables".
```
caseydavenport, line 8: Indentation seems off here

manifests/ocp/cluster-network-operator.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: operator.openshift.io/v1
+kind: Network
+metadata:
+  name: cluster
+spec:
+# The eBPF dataplane doesn't require the kube-proxy to be deployed, so it's set to false.
+# If you change the linuxDataplane to Iptables, the deployKubeProxy parameter must be set to true.
```
caseydavenport, line 7: Same here - comment should align with `deployKubeProxy`

manifests/ocp/01-cr-installation.yaml
caseydavenport: Worth adding a comment here saying "You must also change `deployKubeProxy` within `cluster-network-operator.yaml`

manifests/ocp/cluster-network-operator.yaml
caseydavenport, line 7:
```suggestion
# If you change the linuxDataplane to Iptables in 01-cr-installation.yaml, the deployKubeProxy parameter must be set to true.
```

manifests/ocp/01-configmap-kubernetes-services-endpoint.yaml
caseydavenport: Yeah, I don't think we need to comment these out if we're deploying BPF by default.

But we probably do want to include comments explaining how to determine the correct values for these fields. Is there truly no default value that works for "most" clusters? What if KUBERNETES_SERVICE_HOST was just set to "api"?

lucastigera: The **api.<cluster_name>.<base_domain>** is the template used by OpenShift to set the Kubernetes Service Host. The cluster name is a string defined by the user, and the base domain depends on the user's platform where the cluster will be hosted. Therefore, IMO, there is no way to predict a default or a "most used" value.

I believe that keeping `api.<cluster_name>.<base_domain>` provides better guidance for users to fill in this field.

manifests/tigera-operator-ocp-upgrade.yaml
```diff
@@ -0,0 +1,2866 @@
...
+            - --skip-exists
+            - --skip-empty
+            - -f
+            - /calico-resources
+          volumeMounts:
+            - name: calico-resources
+              mountPath: /calico-resources
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
coutinhop, line 695: we can leave CRDs out of this manifest, the operator will update them

coutinhop (approved):
LGTM

PR #10024, opened 2025-03-20
Add pagination for retrieving hints

Brian-McM:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


goldmane/proto/api.proto
```diff
@@ -19,7 +19,7 @@ service Flows {
   // Namespaces and source / destination names. It allows progressive filtering of criteria based on
   // other filters. i.e., return the flow destinations given a source namespace.
   // Note that this API provides hints to the UI based on past flows and other values may be valid.
-  rpc FilterHints(FilterHintsRequest) returns (stream FilterHint);
+  rpc FilterHints(FilterHintsRequest) returns (FilterHintsResponse);
```
caseydavenport, line 22: This is a change worth talking about I think.

We've been streaming results as a performance measure - presumably this is so that we can include the total number of items in the response?

Alternatives might include adding a new type of `FilterHint` included at the end of the results - e.g., you get a stream of `FilterHint` objects followed by one that just includes the `Total` value and other metadata.

I bring this up because we are going to need to solve this same problem for Flows, which I think have a much more compelling use-case for returning a `stream FlowResult`

caseydavenport: We also use `Result` as the suffix for flows, so probably best to stick with that terminology for consistency.

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -278,102 +303,31 @@ func (a *LogAggregator) Stream(req *proto.FlowStreamRequest) (*Stream, error) {
...
 // List returns a list of flows that match the given request. It uses a channel to
 // synchronously request the flows from the aggregator.
-func (a *LogAggregator) List(req *proto.FlowListRequest) ([]*proto.FlowResult, error) {
+func (a *LogAggregator) List(req *proto.FlowListRequest) ([]*proto.FlowResult, int64, error) {
 	respCh := make(chan *listResponse)
 	defer close(respCh)
 	a.listRequests <- listRequest{respCh, req}
 	resp := <-respCh
-	return resp.flows, resp.err
+	return resp.flows, resp.total, resp.err
```
caseydavenport, line 311: As far as I can tell the `resp.total` here is unused?

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -278,102 +303,31 @@ func (a *LogAggregator) Stream(req *proto.FlowStreamRequest) (*Stream, error) {
...
-	flows, err := a.List(&proto.FlowListRequest{
-		SortBy:       []*proto.SortOption{{SortBy: sortBy}},
-		Filter:       req.Filter,
-		StartTimeGte: req.StartTimeGte,
-		StartTimeLt:  req.StartTimeLt,
-	})
-	if err != nil {
-		return nil, err
+	var hints []*proto.FilterHint
+	for _, key := range resp.keys {
```
caseydavenport, line 323: should this field be `resp.values`? It's going into the `Value: ` field of the response.

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -472,42 +426,83 @@ func (a *LogAggregator) queryFlows(req *proto.FlowListRequest) *listResponse {
...
+	case proto.FilterType_FilterTypeSourceName:
+		sortBy = proto.SortBy_SourceName
+	case proto.FilterType_FilterTypeSourceNamespace:
+		sortBy = proto.SortBy_SourceNamespace
+	}
+
+	// If a sort order was requested, use the corresponding index to find the matching flows.
+	if idx, ok := a.indices[sortBy]; ok {
+		// If a sort order was requested, use the corresponding index to find the matching flows.
+		// We need to convert the FlowKey to a string for the index lookup.
```
caseydavenport, line 493:
```suggestion
```

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -472,42 +426,83 @@ func (a *LogAggregator) queryFlows(req *proto.FlowListRequest) *listResponse {
...
 	})
 
 	// Convert the flows to proto format.
 	flowsToReturn := a.flowsToResult(flows)
-	return &listResponse{flowsToReturn, nil}
+	return &listResponse{int64(totalPages), flowsToReturn, nil}
+}
+
+func (a *LogAggregator) queryFilterHints(req *proto.FilterHintsRequest) *filterHintsResponse {
+	logrus.WithFields(logrus.Fields{"req": req}).Debug("Received flow request")
```
caseydavenport, line 468:
```suggestion
	logrus.WithFields(logrus.Fields{"req": req}).Debug("Received filter hints request")
```

goldmane/pkg/aggregator/index.go
```diff
@@ -30,7 +31,8 @@ type Ordered interface {
 // Index provides efficient querying of Flow objects based on a given sorting function.
 type Index[E Ordered] interface {
 	// TODO: Clients need a way to know how many pages of results exist for a given query.
```
caseydavenport, line 33:
```suggestion
```

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +64,79 @@ type IndexFindOpts struct {
...
 	}).Debug("Listing flows from index")
 
 	var matchedFlows []*types.Flow
+	var totalMatchedCount int
 
-	// Find the index into the DiachronicFlow array from which to start the search based on the provided page number and limit.
-	// We need to iterate through the DiachronicFlows to find the starting point, since not every DiachronicFlow is necessarily a match
-	// for the given options.
-	var i int
-	if opts.page > 0 {
```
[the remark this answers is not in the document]
caseydavenport: I was expecting to see a count of the total number of items within the index that match the filter, not the total number of pages. But they are sort of interchangeable. 

How confident are you that clients prefer number of pages to number of items?

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +64,81 @@ type IndexFindOpts struct {
...
 	}
 
+	return matchedFlows, calculatePageCount(totalMatchedCount, int(opts.pageSize))
+}
+
+// Keys retrieves a unique set of keys matching the given index options.
+func (idx *index[E]) Keys(opts IndexFindOpts) ([]E, int) {
+	logrus.WithFields(logrus.Fields{
+		"opts": opts,
+	}).Debug("Listing flows from index")
```
caseydavenport, line 102:
```suggestion
	}).Debug("Listing keys from index")
```

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +64,81 @@ type IndexFindOpts struct {
...
+	for _, diachronic := range idx.diachronics {
+		flow := idx.evaluate(diachronic, opts)
+		if flow != nil {
+			// increment the count regardless of whether we're including the key, as we need a total matching count.
+			totalMatchedCount++
+
+			// The target represents the index of the last key on the previous page, so we need to ensure that we're above
+			// that index before we start collecting flows.
+			if totalMatchedCount > target {
+				// Only append the key if
```
caseydavenport, line 87: ... if what!?!?

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +64,81 @@ type IndexFindOpts struct {
...
+		flow := idx.evaluate(diachronic, opts)
 		if flow != nil {
-			matchedFlows = append(matchedFlows, flow)
-		}
-
-		if opts.limit > 0 && int64(len(matchedFlows)) == opts.limit {
-			return matchedFlows
+			key := idx.keyFunc(&diachronic.Key)
+			// If the previous key does not equal the current key we know that we haven't seen this key yet, as this
+			// is sorted list and all keys with the same value are together.
```
caseydavenport, line 116: I think it's worth explaining explicitly why we might have duplicate keys - it's not super intuitive, given that Key is generally something meant to be unique.

(In this case, it's because we're not really talking about a key, we're talking about a field of an object used for sorting).

Might be worth doing a pass of this whole Index file and tightening up our terminology for clarity - `keyFunc` -> `getSortValue` or something like that.

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +64,81 @@ type IndexFindOpts struct {
...
+				}
 			}
 		}
 	}
 
+	return matchedFlows, calculatePageCount(totalMatchedCount, int(opts.pageSize))
+}
+
+// Keys retrieves a unique set of keys matching the given index options.
+func (idx *index[E]) Keys(opts IndexFindOpts) ([]E, int) {
```
caseydavenport, line 99: I think Keys is a confusing name for this method of the Index and has thrown me off a few times reading through the code now - a `Key` is a unique identifier for an object, which isn't what this function returns. Plus we already use Key pretty extensively to refer to the FlowKey, which is an identifier for a distinct Flow object. 

Rather, this function returns a deduplicated set of fundamentally non-unique values currently within the index.

Maybe `ValueSet`, or `SortValueSet`?

This is tied with my other comment about tightening up the terminology here for things like `keyFunc`, etc.

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +64,81 @@ type IndexFindOpts struct {
...
-			matchedFlows = append(matchedFlows, flow)
-		}
-
-		if opts.limit > 0 && int64(len(matchedFlows)) == opts.limit {
-			return matchedFlows
+			key := idx.keyFunc(&diachronic.Key)
+			// If the previous key does not equal the current key we know that we haven't seen this key yet, as this
+			// is sorted list and all keys with the same value are together.
+			if previousKey != nil && key == *previousKey {
+				// If we've seen this key match before then continue since it won't count as part of the count.
```
caseydavenport, line 118:
```suggestion
				// If we've seen this key match before then no need to add it to the set.
```

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +64,81 @@ type IndexFindOpts struct {
...
+// Keys retrieves a unique set of keys matching the given index options.
+func (idx *index[E]) Keys(opts IndexFindOpts) ([]E, int) {
+	logrus.WithFields(logrus.Fields{
+		"opts": opts,
+	}).Debug("Listing flows from index")
+
+	var matchedKeys []E
+	var totalMatchedCount int
+
+	target := int(opts.page * opts.pageSize)
```
caseydavenport, line 107:
```suggestion
	pageStart := int(opts.page * opts.pageSize)
```

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +64,81 @@ type IndexFindOpts struct {
...
+			previousKey = &key
+
+			// increment the count regardless of whether we're including the key, as we need a total matching count.
+			totalMatchedCount++
+
+			// The target represents the index of the last key on the previous page, so we need to ensure that we're above
+			// that index before we start collecting keys.
+			if totalMatchedCount > target {
+				// Only append the key if
+				if opts.pageSize == 0 || int64(len(matchedKeys)) < opts.pageSize {
```
caseydavenport, line 130:
```suggestion
                        // Include the value if:
                        // - We're not performing a paginated search
                        // - We are performing a paginated search, and it falls within the page bounds.
			if opts.pageSize == 0 || totalMatchedCount > target || int64(len(matchedKeys)) < opts.pageSize {
```

I find this phrasing of the condition a bit easier to follow, but just preference.

goldmane/proto/api.proto
```diff
@@ -105,6 +105,19 @@ message FilterHintsRequest {
...
+  // to determine page boundaries. Note that pages may change over time as new flow data is collected or expired.
+  // Querying the same page at different points in time may return different results.
+  int64 page = 5;
+
+  // PageSize configures the maximum number of results to return as part of this query.
+  int64 page_size = 6;
+}
+
+message FilterHintsResponse {
+  int64 total = 1;
```
caseydavenport, line 119: Need API documentation comments on this guy.

I expected "total" to be "total number of hints that matched the query" and not "total number of pages of hints that match the query"

goldmane/pkg/aggregator/ring_index.go
```diff
@@ -16,39 +16,46 @@ package aggregator
...
 	"sort"
 
 	"github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/calico/goldmane/pkg/internal/types"
+	"github.com/projectcalico/calico/libcalico-go/lib/set"
 )
 
-func NewRingIndex(a *LogAggregator) *RingIndex {
+type logAggregator interface {
```
caseydavenport, line 27: I think it's worth adding a comment or two about why we have this intermediary interface.

goldmane/pkg/aggregator/ring_index.go
```diff
@@ -106,11 +115,15 @@ func (a *RingIndex) List(opts IndexFindOpts) []*types.Flow {
...
 }
 
 func (r *RingIndex) Add(d *types.DiachronicFlow) {
 }
 
 func (r *RingIndex) Remove(d *types.DiachronicFlow) {
 }
+
+func (a *RingIndex) Keys(opts IndexFindOpts) ([]int64, int) {
+	return []int64{}, 0
```
caseydavenport, line 125: Do we want a warning log here? Seems like a bug if we ever call this?

goldmane/pkg/aggregator/index.go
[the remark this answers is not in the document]
Brian-McM: I'm sort of basing it off of what I need right now. TBH we could probably send both since we're sending back a metadata object.

goldmane/proto/api.proto
Brian-McM: Updated. Let me know what you think.

goldmane/pkg/aggregator/ring_index.go
Brian-McM: I was just going off the precedence set above, but I guess this is different since you expect something valid to be returned.

I think a panic might actually be appropriate, if this is called then it's a dev error.

goldmane/pkg/aggregator/aggregator.go
caseydavenport: @Brian-McM I think this one still doesn't apply - we're not returning "Flows" here and we're not converting FLowKeys?

goldmane/pkg/aggregator/aggregator.go
caseydavenport: same here

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -475,39 +452,138 @@ func (a *LogAggregator) queryFlows(req *proto.FlowListRequest) *listResponse {
...
+		// We need to convert the FlowKey to a string for the index lookup.
+		keys, totalPages := idx.Keys(IndexFindOpts{
+			startTimeGt: req.StartTimeGte,
+			startTimeLt: req.StartTimeLt,
+			pageSize:    req.PageSize,
+			page:        req.Page,
+			filter:      req.Filter,
+		})
+
+		results := make(chan *proto.FilterHintsResult, 100)
```
caseydavenport, line 559: Worth doing this based on the number of results?

Like, if we know there are only 5 results we can do a smaller channel. If we know there are 1k we can do a larger channel (maybe with a hard cap at 1k or something)

goldmane/pkg/aggregator/aggregator_test.go
```diff
@@ -1565,14 +1584,20 @@ func TestFilterHints(t *testing.T) {
...
+				flows, _ := aggList(agg, &proto.FlowListRequest{})
 				return len(flows) == 10
 			}, waitTimeout, retryTime, "Didn't receive all flows").Should(BeTrue())
 
 			// Query for hints using the query from the testcase.
-			hints, err := agg.Hints(tc.req)
+			results, err := agg.StreamHints(tc.req)
 			require.NoError(t, err)
 
+			var hints []*proto.FilterHint
```
caseydavenport, line 1595: Probably worth verifying the metadata message is correct as well

goldmane/pkg/aggregator/aggregator_test.go
```diff
@@ -69,6 +70,22 @@ func ExpectFlowsEqual(t *testing.T, expected, actual *proto.Flow) {
...
 }
 
+func aggList(agg *aggregator.LogAggregator, req *proto.FlowListRequest) ([]*proto.FlowResult, error) {
+	var flows []*proto.FlowResult
+	resultChan, err := agg.List(req)
+	if err != nil {
+		return nil, err
+	}
+
+	for result := range resultChan {
```
caseydavenport, line 80: Probably worth verifying the metadata message is correct as well

goldmane/pkg/aggregator/ring_index.go
```diff
@@ -21,34 +21,40 @@ import (
...
+	// Collect all the flow keys across all buckets that match the request. We will then
 	// use DiachronicFlow data to combine statistics together for each key across the time range.
-	keys := a.agg.buckets.FlowSet(opts.startTimeGt, opts.startTimeLt)
+	keys := a.agg.flowSet(opts.startTimeGt, opts.startTimeLt)
 
 	// Aggregate the relevant DiachronicFlows across the time range.
 	flowsByKey := map[types.FlowKey]*types.Flow{}
 	keys.Iter(func(key types.FlowKey) error {
-		d, ok := a.agg.diachronics[key]
-		if !ok {
```
caseydavenport, line 51: curious if this is a functional or just aesthetic change?

goldmane/pkg/server/flow_service.go
```diff
@@ -48,7 +48,7 @@ func (s *FlowsServer) List(req *proto.FlowListRequest, server proto.Flows_ListSe
 	}
 
 	// Send flows.
-	for _, flow := range flows {
+	for flow := range flows {
```
caseydavenport, line 51: Another naming thing, but I think we need to rename all of these from "flow" and "flows" to "messages" and "message" or something like that - we have a new type of non-flow message that can be sent (metadata) so we should make sure naming reflects that. 

Same for the filter hints.

goldmane/pkg/aggregator/index.go
Brian-McM: Wouldn't you like to know...

goldmane/pkg/aggregator/index.go
Brian-McM: Ya good point, I might opt for renaming key to value as suggested. I'll add a comment after that.

goldmane/pkg/aggregator/index.go
Brian-McM: That's not correct though, it would should be `totalMatchedCount > pageStart && (opts.pageSize == 0 || int64(len(matchedFlows)) < opts.pageSize)` (which I can change it to).

goldmane/pkg/aggregator/ring_index.go
Brian-McM: done

goldmane/pkg/aggregator/ring_index.go
Brian-McM: Neither? If that makes sense?

This is because I added the internal interface and I didn't see a need to return a second bool value, but I can if you think it's a good idea.

caseydavenport: oh, right. Yeah, the interface is throwing me off.

goldmane/pkg/aggregator/ring_index.go
Brian-McM: Added a panic.

goldmane/proto/api.proto
```diff
@@ -105,6 +115,33 @@ message FilterHintsRequest {
...
+  // Querying the same page at different points in time may return different results.
+  int64 page = 5;
+
+  // PageSize configures the maximum number of results to return as part of this query.
+  int64 page_size = 6;
+}
+
+message FilterHintsResult {
+  // ListMetadata specifies information about the filter hints to be returned after this message. On one of ListMetadata
+  // or FilterHint will be specified.
```
caseydavenport, line 130:
```suggestion
  // ListMetadata specifies information about the filter hints to be returned after this message.
```

goldmane/proto/api.proto
```diff
@@ -105,6 +115,33 @@ message FilterHintsRequest {
...
+
+  // FilterHint contains a value that flows can be filtered on.
+  FilterHint value = 2;
+}
+
+// ListMetadata contains information about a returned list of items, such as pagination information (total number of pages
+// and total number of results).
+message ListMetadata {
+  // totalPages is the total number of pages that exist given that a pageSize was specified.
+  int64 totalPages = 1;
```
caseydavenport, line 141:
```suggestion
  // TotalPages is the total number of pages that exist given that a pageSize was specified.
  int64 total_pages = 1;
```

goldmane/proto/api.proto
```diff
@@ -105,6 +115,33 @@ message FilterHintsRequest {
...
+
+  // totalResults are the total number of results that would have been returned if no pagination was specified.
+  int64 totalResults = 2;
```
caseydavenport, line 144:
```suggestion
  // TotalResults are the total number of results that would have been returned if no pagination was specified.
  int64 total_results = 2;
```

caseydavenport (approved):
Please take a look at the latest set of comments - mostly tidy up type things, and a couple missing test assertions.

Otherwise this LGTM, no need to re-review.

goldmane/pkg/client/flowservice.go
```diff
@@ -49,53 +49,64 @@ func NewFlowsAPIClient(host string, opts ...grpc.DialOption) (FlowsClient, error
...
 		// Break if EOF is found (no more data to be returned).
 		if err == io.EOF {
 			break
 		} else if err != nil {
-			return nil, fmt.Errorf("failed to receive flow from stream: %w", err)
+			return nil, nil, fmt.Errorf("failed to receive flow from stream: %w", err)
 		}
 
-		hints = append(hints, flow)
+		if result.Meta != nil {
```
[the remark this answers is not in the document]
Brian-McM: This isn't really assuming we only get one, this is assuming the last value for metadata is the correct one if there are multiple values sent.

I think this is a fine assumption for this endpoint, the logic being the metadata has been updated and needs to be reset.

goldmane/pkg/aggregator/aggregator_test.go
caseydavenport: @Brian-McM you're fundamentally adding new functionality that needs to be unit tested though, this isn't just a matter of refactoring it's a matter of asserting the new functionality works as expected.

matthewdupre: UTs isn't something we should be leaving for a followup PR: we need to be able to trust that merged code is properly tested.  I won't weigh in on what the right UTs are here though.

Brian-McM: So I did modify some of the UTs, but to be clear the reason why I was saying this was because there are other UTs added to the index / ring index code. Maybe that still wasn't good enough, but I wasn't forgoing testing this.

goldmane/proto/api.proto
```diff
@@ -61,13 +61,13 @@ message FlowListRequest {
...
-// FlowListResult is a message containing a FlowResult or ListMetadata (it will only ever contain one or the other).
+// FlowListResult is a message containing a list of FlowResults and ListMetadata.
 message FlowListResult {
-  // ListMetadata specifies information about the filter hints to be returned after this message.
+  // ListMetadata specifies list information about the hints returned.
   ListMetadata meta = 1;
 
-  // FlowResult wraps a Flow object with additional per-flow metadata.
-  FlowResult value = 2;
+  // FlowResult wraps Flow objects with additional per-flow metadata.
```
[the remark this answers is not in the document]
caseydavenport: I think all of these comments are actually subtly wrong - they are written in terms of the `Type` when actually they should be the field names.

e.g.,

```
// Flows is a list of FlowResult objects.
```

goldmane/proto/api.proto
```diff
@@ -61,13 +61,13 @@ message FlowListRequest {
   int64 aggregation_interval = 7;
 }
 
-// FlowListResult is a message containing a FlowResult or ListMetadata (it will only ever contain one or the other).
+// FlowListResult is a message containing a list of FlowResults and ListMetadata.
 message FlowListResult {
-  // ListMetadata specifies information about the filter hints to be returned after this message.
+  // ListMetadata specifies list information about the hints returned.
```
caseydavenport, line 66:
```
Meta specifies metadata about the returned flows
```

^ Same here - the field name is "Meta" and the type is "ListMetadata"

Also, this one still mentions "hints" instead of "flows"

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -424,10 +369,12 @@ func (a *LogAggregator) backfill(stream *Stream, request *proto.FlowStreamReques
 		return
 	}
 
+	flows := resp.results.Flows
```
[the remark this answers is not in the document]
caseydavenport: Not that it matters - this is all going to change in my PR, so we can ignore this one.

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +61,97 @@ type IndexFindOpts struct {
 	filter *proto.Filter
 }
 
-func (idx *index[E]) List(opts IndexFindOpts) []*types.Flow {
+// List returns a list of flows and metadata about the list that's returned.
+// TODO we might want to return a channel instead of a list and make this more memory efficient. The list metadata
```
caseydavenport, line 65: TODOs no longer apply?

goldmane/pkg/aggregator/index.go
```diff
@@ -60,42 +61,97 @@ type IndexFindOpts struct {
...
+			}
+			previousSortValue = &sortValue
+
+			// increment the count regardless of whether we're including the key, as we need a total matching count.
+			totalMatchedCount++
+
+			// The target represents the index of the last key on the previous page, so we need to ensure that we're above
+			// that index before we start collecting keys.
+			if totalMatchedCount > pageStart {
+				// Only append the key if
```
caseydavenport, line 127: .... IF WHAT!?!?

goldmane/pkg/aggregator/index_test.go
```diff
@@ -160,11 +185,404 @@ func TestIndexAddRemove(t *testing.T) {
...
+			expectedTotalPages:   1,
+			expectedTotalResults: 4,
+			expectedNumFlows:     4,
+			filter:               &proto.Filter{DestNamespaces: []*proto.StringMatch{{Value: "ns2"}}},
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.description, func(t *testing.T) {
+			RegisterTestingT(t)
```
caseydavenport, line 307: I don't think we need this - it's in setupTest?

goldmane/pkg/aggregator/index_test.go
```diff
@@ -160,11 +185,404 @@ func TestIndexAddRemove(t *testing.T) {
...
 	Expect(flows).To(HaveLen(0))
 
 	// Remove flows we know aren't in the index, to make sure we're idempotent.
 	for _, flow := range allFlows {
 		idx.Remove(flow)
 	}
 }
+
+func TestIndexPagination_General(t *testing.T) {
+	defer setupTest(t)()
```
caseydavenport, line 198: We should call this within the `t.Run` - it's meant to be called per-test, not per suite.

This can cause problems if we eventually add other per-test state to this function.

goldmane/pkg/aggregator/index_test.go
```diff
@@ -160,11 +185,404 @@ func TestIndexAddRemove(t *testing.T) {
...
+				TotalPages:   tc.expectedTotalPages,
+				TotalResults: tc.expectedTotalResults,
+			}))
+			Expect(len(keys)).To(Equal(tc.expectedNumFlows))
+		})
+	}
+}
+
+func TestRingIndexPagination_General(t *testing.T) {
+	defer setupTest(t)()
```
caseydavenport, line 463: Same here -should be within t.Run

goldmane/pkg/aggregator/index_test.go
```diff
@@ -160,11 +185,404 @@ func TestIndexAddRemove(t *testing.T) {
...
+			expectedTotalPages:   1,
+			expectedTotalResults: 4,
+			expectedNumFlows:     4,
+			filter:               &proto.Filter{DestNamespaces: []*proto.StringMatch{{Value: "ns2"}}},
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.description, func(t *testing.T) {
+			RegisterTestingT(t)
```
caseydavenport, line 573: Should be `defer setupTest(t)()`

goldmane/pkg/aggregator/index_test.go
```diff
@@ -21,9 +21,34 @@ import (
 
 	"github.com/projectcalico/calico/goldmane/pkg/internal/types"
 	"github.com/projectcalico/calico/goldmane/pkg/internal/utils"
+	"github.com/projectcalico/calico/goldmane/proto"
 	"github.com/projectcalico/calico/libcalico-go/lib/logutils"
+	"github.com/projectcalico/calico/libcalico-go/lib/set"
 )
 
+type simpleLogAggregatorStub struct {
+	diachronics []*types.DiachronicFlow
```
caseydavenport, line 30: not really a problem, but seems like this structure would be a lot easier to implement if it just had a `set.Set` here already - one function converts to a Set anyway, and the other does a lookup by iterating the items instead of just using the Set to access the value.

goldmane/pkg/aggregator/index_test.go
caseydavenport: Right, but we should be calling setupTest() within this t.Run anyway

goldmane/pkg/aggregator/index_test.go
Brian-McM: Ya probably, I'll think about adding this now.

goldmane/pkg/aggregator/index.go
Brian-McM: Ya I wasn't suggesting keeping the today (I've removed it), just saying that it might apply for a future use case.

goldmane/pkg/aggregator/index_test.go
Brian-McM: Ya I realized that looking into your other comment.

goldmane/pkg/aggregator/index_test.go
Brian-McM: I think I'm just going to leave this for now to get this through.

goldmane/pkg/client/flowservice.go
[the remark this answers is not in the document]
caseydavenport: It's moot now since we changed how this works.

caseydavenport (approved):

PR #10027, opened 2025-03-21
Use new set.Adaptive type in named port index

fasaxc:
## Description

Add an "adaptive" implementation of `set.Set` that uses a different storage approach depending on number of items.

* 0 items: store a nil pointer.
* 1 item: store pointer to the item
* 2-16 items: store pointer to array of that length rounded up to power of 2
* 17+: store pointer to a proper map.

It is faster and uses a fraction of the RAM for 0-16 items, then its performance becomes similar to the normal set.Typed (with slight overhead).

Use the map in the named port index, where we have many sets that are generally expected to be very small.

It knocks 20% off the `BenchmarkSnapshot200Local250kTotal10000TagPols` benchmark, which has a very large number of small sets.  It gives almost no improvement on the others, but those are dominated by having large numbers of active policies and very few labels.

```
Before:

goos: linux
goarch: amd64
BenchmarkSnapshot200Local250kTotal10000TagPols-12                     	       1	3245931822 ns/op	       307.9 HeapAllocMB	        29.00 IPSets	     10432 Msgs	     10000 Policies	         3.246 s	970140472 B/op	10684941 allocs/op
BenchmarkSnapshot200Local10kTotal1000NetsetPols-12                    	       2	 516745681 ns/op	        78.60 HeapAllocMB	      1000 IPSets	      2403 Msgs	      1000 Policies	         0.5117 s	245642908 B/op	 1909899 allocs/op
BenchmarkSnapshot200Local10kTotal10000NetsetPols-12                   	       1	5866074645 ns/op	       604.9 HeapAllocMB	     10000 IPSets	     20403 Msgs	     10000 Policies	         5.866 s	2525107568 B/op	17306981 allocs/op

After:

goos: linux
goarch: amd64
BenchmarkSnapshot200Local250kTotal10000TagPols-12                     	       1	3230500160 ns/op	       249.8 HeapAllocMB	        29.00 IPSets	     10432 Msgs	     10000 Policies	         3.230 s	929163456 B/op	11895089 allocs/op
BenchmarkSnapshot200Local10kTotal1000NetsetPols-12                    	       2	 510494161 ns/op	        77.94 HeapAllocMB	      1000 IPSets	      2403 Msgs	      1000 Policies	         0.5055 s	245306080 B/op	 1962315 allocs/op
BenchmarkSnapshot200Local10kTotal10000NetsetPols-12                   	       1	5766052793 ns/op	       598.9 HeapAllocMB	     10000 IPSets	     20403 Msgs	     10000 Policies	         5.766 s	2514418280 B/op	17287133 allocs/op
```

Also included: store sets as their concrete type in the places where we store a lot of them; saves a few MB.

## Related issues/PRs

CORE-11400

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Felix now stores its selector index more efficiently resulting in reduced memory usage in large clusters.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/set/adaptive.go
+++ b/libcalico-go/lib/set/adaptive.go
@@ -1,0 +1,311 @@
+// Copyright (c) 2025 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package set
+
+import (
+	"unsafe"
+)
+
+const (
+	adaptiveSetArrayLimit = 16
+	sizeStoredInMap       = -1
+)
+
+var arrCapForSize = [adaptiveSetArrayLimit + 1]int{
+	0, 1, 2, 4, 4, 8, 8, 8, 8, 16, 16, 16, 16, 16, 16, 16, 16,
+}
+
+// Adaptive is a set implementation that uses different underlying data
+// structures depending on the size of the set.  For sets that are typically
+// empty, or, have only one or two elements, it is more than twice as fast and
+// it uses ~10x less memory. Above adaptiveSetArrayLimit it switches to a
+// map-based implementation like set.Typed.
+//
+// The zero value of Adaptive is an empty set (so it may be embedded in other
+// datastructures).
+//
+// Should not be copied after first use. Since the struct carries the size,
+// copying by value and then mutating can result in two Adaptive instances
+// sharing storage but with different sizes; this is not recommended!
+type Adaptive[T comparable] struct {
+	// size is either the number of elements in the set, or sizeStoredInMap
+	// if the set is backed by a map.
+	size int
+
+	// p holds different types depending on size.
+	// if size == 0, p is nil.
+	// if size is in the range [1, adaptiveSetArrayLimit], p is a pointer to
+	//    an array with length size, rounded up to next power of two.
+	// if size == sizeStoredInMap, p is a pointer to a map[T]v
+	p unsafe.Pointer
+}
+
+func NewAdaptive[T comparable]() *Adaptive[T] {
+	return &Adaptive[T]{}
+}
+
+func AdaptiveFromArray[T comparable](items []T) *Adaptive[T] {
+	s := NewAdaptive[T]()
+	for _, t := range items {
+		s.Add(t)
+	}
+	return s
+}
+
+func AdaptiveFrom(items ...int) *Adaptive[int] {
+	return AdaptiveFromArray(items)
+}
+
+func (a *Adaptive[T]) Len() int {
+	if a.size == sizeStoredInMap {
+		return len(*(*map[T]v)(a.p))
+	}
+	return int(a.size)
+}
+
+func (a *Adaptive[T]) Add(item T) {
+	switch a.size {
+	case 0:
+		// Array of one item is just the item.
+		a.p = unsafe.Pointer(&item)
+		a.size = 1
+	default:
+		tPtr := (*T)(a.p)
+		tSlice := unsafe.Slice(tPtr, arrCapForSize[a.size])[:a.size]
+		for _, t := range tSlice {
+			if t == item {
+				// The element is already in the set.
+				return
+			}
+		}
+
+		// If we get here, need to add the element to the set.
+		if a.size < adaptiveSetArrayLimit {
+			// Set is still small enough to use a slice.
+			if len(tSlice) == cap(tSlice) {
+				// Need to grow the slice; we do it manually so we can control
+				// the exact new capacity.
+				tSlice = growSliceCap(tSlice, arrCapForSize[a.size+1])
+				a.p = unsafe.Pointer(&tSlice[0])
+			}
+			tSlice = append(tSlice, item)
+			a.size++
+			return
+		}
+
+		// Need to upgrade to a map.
+		m := make(map[T]v, a.size+1)
+		for _, t := range tSlice {
+			m[t] = emptyValue
+		}
+		m[item] = emptyValue
+		a.p = unsafe.Pointer(&m)
+		a.size = sizeStoredInMap
+	case sizeStoredInMap:
+		m := *(*map[T]v)(a.p)
+		m[item] = emptyValue
+	}
+}
+
+func growSliceCap[T any](in []T, newCap int) []T {
+	out := make([]T, len(in), newCap)
+	copy(out, in)
+	return out
+}
+
+func (a *Adaptive[T]) AddAll(itemArray []T) {
+	for _, v := range itemArray {
+		a.Add(v)
+	}
+}
+
+func (a *Adaptive[T]) AddSet(other Set[T]) {
+	other.Iter(func(item T) error {
+		a.Add(item)
+		return nil
+	})
+}
+
+func (a *Adaptive[T]) Discard(item T) {
+	switch a.size {
+	case 0:
+		return
+	case 1:
+		theOne := (*T)(a.p)
+		if *theOne == item {
+			a.p = nil
+			a.size = 0
+		}
+	case sizeStoredInMap:
+		m := *(*map[T]v)(a.p)
+		delete(m, item)
+		if len(m) <= adaptiveSetArrayLimit {
+			// Downgrade to an array.
+			s := make([]T, 0, arrCapForSize[len(m)])
+			for t := range m {
+				s = append(s, t)
+			}
+			a.p = unsafe.Pointer(&s[0])
+			a.size = len(m)
+		}
+	default:
+		tPtr := (*T)(a.p)
+		tSlice := unsafe.Slice(tPtr, arrCapForSize[a.size])[:a.size]
+		for i, t := range tSlice {
+			if t == item {
+				// Found the element to remove.
+				newSize := a.size - 1
+				newCap := arrCapForSize[newSize]
+				if newCap < cap(tSlice) {
+					// Downgrade to a smaller array.
+					updatedSlice := make([]T, arrCapForSize[newSize])
+					// Copy the elements before and after the removed element.
+					copy(updatedSlice, tSlice[:i])
+					copy(updatedSlice[i:], tSlice[i+1:])
+					a.p = unsafe.Pointer(&updatedSlice[0])
+				} else {
+					// Keep the same slice.  Swap the last element into the
+					// removed element's slot.
+					tSlice[i] = tSlice[newSize]
+					var zeroT T
+					tSlice[newSize] = zeroT
+				}
+				a.size = newSize
+				return
+			}
+		}
+
+	}
+}
+
+func (a *Adaptive[T]) Clear() {
+	a.size = 0
+	a.p = nil
+}
+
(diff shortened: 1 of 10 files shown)


libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,311 @@
...
+
+func AdaptiveFromArray[T comparable](items []T) *Adaptive[T] {
+	s := NewAdaptive[T]()
+	for _, t := range items {
+		s.Add(t)
+	}
+	return s
+}
+
+func AdaptiveFrom(items ...int) *Adaptive[int] {
```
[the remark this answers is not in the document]
nelljerram: Sorry, my mistake, I thought the int parameter was a count.  In fact I think Copilot is right here.

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,311 @@
...
+	return AdaptiveFromArray(items)
+}
+
+func (a *Adaptive[T]) Len() int {
+	if a.size == sizeStoredInMap {
+		return len(*(*map[T]v)(a.p))
+	}
+	return int(a.size)
```
nelljerram: int() cast not needed here?

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+func (a *Adaptive[T]) Add(item T) {
+	switch a.size {
+	case 0:
+		// Array of one item is just the item.
+		a.p = unsafe.Pointer(&item)
+		a.size = 1
+	default:
```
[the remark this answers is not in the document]
fasaxc: SGTM

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,311 @@
...
+				// Found the element to remove.
+				newSize := a.size - 1
+				newCap := arrCapForSize[newSize]
+				if newCap < cap(tSlice) {
+					// Downgrade to a smaller array.
+					updatedSlice := make([]T, arrCapForSize[newSize])
+					// Copy the elements before and after the removed element.
+					copy(updatedSlice, tSlice[:i])
+					copy(updatedSlice[i:], tSlice[i+1:])
+					a.p = unsafe.Pointer(&updatedSlice[0])
```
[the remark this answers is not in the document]
fasaxc: This can't happen, it would go through the `case 1:` arm of the switch.

libcalico-go/lib/set/adaptive.go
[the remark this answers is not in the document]
nelljerram: Presumably need to add a UT to cover this.

libcalico-go/lib/set/adaptive.go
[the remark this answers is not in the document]
nelljerram: I don't follow copilot's concern, but I do think it's unusual to see a `default:` _before_ a non-default `case ...:`.  Copilot's suggested edit would address that.

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,351 @@
...
+		m := make(map[T]v, a.size+1)
+		for _, t := range tSlice {
+			m[t] = emptyValue
+		}
+		m[item] = emptyValue
+		a.p = unsafe.Pointer(&m)
+		a.size = sizeStoredInMap
+	case sizeStoredInMap:
+		m := a.loadMapFromPointer()
+		m[item] = emptyValue
```
nelljerram: Is it guaranteed that the map address never changes here?

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,351 @@
...
+			}
+		}
+
+		// If we get here, need to add the element to the set.
+		if a.size < adaptiveSetArrayLimit {
+			// Set is still small enough to use a slice.
+			if len(tSlice) == cap(tSlice) {
+				// Need to grow the slice; we do it manually so we can control
+				// the capacity and keep it in sync with arrCapForSize (allowing
+				// us to avoid storing it).
```
nelljerram, line 108: I don't follow this "storing it" comment, given that we get the capacity from `cap(tSlice)`.  What are you suggesting we'd need to store?

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,351 @@
...
+		tSlice := a.loadSliceFromPointer()
+
+		// Scan the slice to see if the item is present.
+		for i, t := range tSlice {
+			if t == item {
+				// Found the element to remove.
+				newSize := a.size - 1
+				newCap := arrCapForSize[newSize]
+				if newCap < cap(tSlice) {
+					// Downgrade to a smaller array.
```
nelljerram: Is it enough of a win - in terms of storage saving when discarding items - to justify the code for storage adjustments when reducing the size of the set?  ISTM it could be very defensible to say that we only adjust the underlying storage when _adding_ items.

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,351 @@
...
+		return
+	case 1:
+		// Set has only one item, the pointer will point to that single item,
+		// which is equivalent to a single-entry array.
+		theOne := (*T)(a.p)
+		if *theOne == item {
+			a.p = nil
+			a.size = 0
+		}
+	default:
```
nelljerram, line 172: Same comment as above about seeing a default before another case.

libcalico-go/lib/set/adaptive.go
nelljerram: OTOH that code is already here now, and it's not too complicated, so perhaps not worth worrying about.

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,351 @@
...
+
+		// Scan the slice to see if the item is present.
+		for i, t := range tSlice {
+			if t == item {
+				// Found the element to remove.
+				newSize := a.size - 1
+				newCap := arrCapForSize[newSize]
+				if newCap < cap(tSlice) {
+					// Downgrade to a smaller array.
+					updatedSlice := make([]T, arrCapForSize[newSize])
```
nelljerram: Should this be `make([]T, newSize, arrCapForSize[newSize])` ?  Otherwise I don't see how subsequent `append`s can work correctly.  If I'm right, that suggests we also need some more UT.

Oh, maybe it's OK because we reconstruct the slice each time in `loadSliceFromPointer`, based on `a.size`.  If so, I guess that also answers my querying about "storing the capacity".

nelljerram: Maybe worth some additional commenting to clarify that we only store the array behind the slice and not the slice itself.

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,351 @@
...
+					// Copy the elements before and after the removed element.
+					copy(updatedSlice, tSlice[:i])
+					copy(updatedSlice[i:], tSlice[i+1:])
+					a.p = unsafe.Pointer(&updatedSlice[0])
+				} else {
+					// Keep the same slice.  Swap the last element into the
+					// removed element's slot.
+					tSlice[i] = tSlice[newSize]
+					var zeroT T
+					tSlice[newSize] = zeroT
```
nelljerram, line 203: Like in my last comment, this is another case that looks confusing, because we don't reduce the len of the slice.  What is the need for the `tSlice[newSize] = zeroT` line?

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,351 @@
...
+				}
+				a.size = newSize
+				return
```
nelljerram, line 206: If we're aiming to be fully non-hysteretical with the underlying storage representation, I think we're missing the handling for reducing from size 2 to size 1.

libcalico-go/lib/set/adaptive.go
```diff
@@ -0,0 +1,351 @@
...
+				a.Discard(v)
+			}
+		}
+	}
+}
+
+// loadSliceFromPointer loads the array stored in our unsafe pointer, returning
+// it as a slice with len == a.size and cap == arrCapForSize[a.size].
+func (a *Adaptive[T]) loadSliceFromPointer() []T {
+	if a.size < 1 || a.size > adaptiveSetArrayLimit {
```
nelljerram, line 287: I think `a.size < 1` could be `a.size < 2` if we had the code for downgrading storage from size 2 to size 1.

(Although now wondering if I'm misunderstanding the equivalence between a single value and a slice of len 1.)

libcalico-go/lib/set/adaptive.go
fasaxc: I mean, we don't need a field to store capacity because it's implicit depending on the size.

libcalico-go/lib/set/adaptive.go
fasaxc: A pointer to an array is just a pointer to the first element, so, if there's only one element, there's no difference between a pointer to an array and pointer to an element directly.

libcalico-go/lib/set/adaptive.go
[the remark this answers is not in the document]
fasaxc: I'm not keen on listing all the numbers, for fear of missing a case when maintaining the code.

libcalico-go/lib/set/adaptive.go
fasaxc: Not while we're looking at it; I think.  Maps are defined as reference types, so when you have a map type variable, in practice, it's actually a pointer to some external data.  I'm taking a pointer to that pointer, dereferencing it and then operating on the map.  I think that's all safe and sound.

libcalico-go/lib/set/adaptive.go
fasaxc: I've been bitten by Go's built in map not shrinking after containing a lot of elements in the past; my default is to shrink so that the capacity is kept in sync with `arrCapForSize`.  If we then identify some real code that would benefit from not shrinking, we can revisit.

libcalico-go/lib/set/adaptive.go
fasaxc: I think that "just works".  `arrCapForSize[1] == 1` so we'll make a len == 1 array and copy into it. No real difference vs a dedicated case.

libcalico-go/lib/set/adaptive.go
fasaxc: > Should this be make([]T, newSize, arrCapForSize[newSize])

Yes, that'd be clearer.  Originally I'd written 
```
make([]T, arrCapForSize[newSize], arrCapForSize[newSize])
```
and the linter informed me that `make([]T, x, x)` is equivalent to `make([]T, x)`. Much clearer to use `newSize` instead for the length of the slice.  I'm not sure why I used `arrCapForSize[newSize]` in both arguments before.

libcalico-go/lib/set/adaptive.go
fasaxc: It prevents holding on to a potentially large object.  The GC still sees the whole array, not just our slice of it.

libcalico-go/lib/set/adaptive.go
fasaxc: Let's just put default at the end.  I was trying to hint that this is the numeric order even though I'm using the default case but I've added comments now so let's remove the surprising ordering. I'm not keen on a long list of numeric cases.

nelljerram: That works, thanks.

libcalico-go/lib/set/adaptive.go
nelljerram: Ah right, thanks.

libcalico-go/lib/set/adaptive.go
nelljerram: Got it.  I didn't fully understand that before.

PR #10032, opened 2025-03-21
[release tool] auth and image validation fixups [master]

danudey:
A few minor fixups to docker authentication and image existence validation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -120,6 +120,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require github.com/avast/retry-go/v4 v4.6.1
+
 require (
 	al.essio.dev/pkg/shellescape v1.5.1 // indirect
 	cel.dev/expr v0.19.1 // indirect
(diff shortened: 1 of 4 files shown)


release/pkg/manager/calico/manager.go
```diff
@@ -629,11 +629,14 @@ func (r *CalicoManager) checkHashreleaseImagesPublished() ([]registry.Component,
...
 		result := <-resultsCh
 		if result.err != nil {
-			resultsErr = errors.Join(resultsErr, fmt.Errorf("error checking %s exists: %s", result.image, result.err.Error()))
+			logrus.Error(fmt.Errorf("error checking %s exists: %s", result.image, result.err.Error()))
 		} else if !result.exists {
 			missingImages = append(missingImages, r.imageComponents[result.name])
 		}
 	}
+	if len(missingImages) > 0 {
+		resultsErr = fmt.Errorf("error validating %s images", len(missingImages))
```
radTuti, line 638: resultsErr should state all the missing images as well, simply having a length is not useful error information

go.mod
```diff
@@ -120,6 +120,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require github.com/avast/retry-go/v4 v4.6.1
```
radTuti, line 123: this require should not be separated. put it in the appropirate existing require section.

also, I think it is overkill to add a rety library that is used once.

radTuti (changes requested):

PR #10043, opened 2025-03-24
Use unique.Handle to deduplicate FlowKey fields

caseydavenport:
## Description

FlowKey fields are highly duplicated across flows - use unique.Handle to intern these values and reduce memory usage.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


goldmane/pkg/internal/types/flow.go
```diff
@@ -23,104 +23,102 @@ import (
...
-	// Reporter is either "src" or "dst", depending on whether this flow was generated
-	// at the initiating or terminating end of the connection attempt.
-	Reporter proto.Reporter `protobuf:"varint,13,opt,name=reporter,proto3,enum=goldmane.Reporter" json:"reporter,omitempty"`
-	// Action is the ultimate action taken on the flow.
-	Action proto.Action `protobuf:"varint,14,opt,name=action,proto3,enum=goldmane.Action" json:"action,omitempty"`
-	// Policies includes an entry for each policy rule that took an action on the connections
-	// aggregated into this flow.
-	Policies unique.Handle[PolicyTrace] `protobuf:"bytes,15,opt,name=policies,proto3" json:"policies,omitempty"`
+	Source      unique.Handle[FlowKeySource]
+	Destination unique.Handle[FlowKeyDestination]
```
Brian-McM, line 56: Any reason for these to be exported?

goldmane/pkg/types/flow.go
```diff
@@ -0,0 +1,336 @@
...
+
+type FlowKeySource struct {
+	SourceName      string
+	SourceNamespace string
+	SourceType      proto.EndpointType
+}
+
+type FlowKeyDestination struct {
+	DestName             string
+	DestNamespace        string
```
Brian-McM, line 34: Are we not worried about the overhead of forgoing interning the values here? 

What you've done is definitely the a big chunk, but would there be a downside to interning these values too? The namespaces will at least be a source of constant duplication.

caseydavenport: I think it's hard to say which approach is going to be more effective at de-duplication - interning the entire structures, or interning the individual values. 

My hunch is that that interning the structures will be - on average - more effective, as it can reduce multiple fields down to a single pointer instead of a pointer per-field, and my suspicion is that the nature of Flow data is such that we will see lots of the same sources and destinations, rather than a full distribution of every combination of name / namespace, etc.

Brian-McM: I wasn't suggesting that you remove the struct interning, it's a good idea and if it was a one vs the other I agree that the struct interning might be more valuable. 

I'm just wondering if there's any reason why we can't also intern some or all of the strings here as well, since we know they'll be duplicated.

With generics and the unique package, it seems like interning is quite simple to implement. I'm wondering if there's any reason we shouldn't be a bit more liberal with this interning feature in cases where we know the something might be duplicated. 

Unless we know of some issues with the library if it's overused (slight overhead with the internal map lookups?).

caseydavenport: I'd rather we identify specific areas that would benefit the most from this, and then address them. If profiling comes back and says that the flow keys are still the problem, then we can look at it at as a solution - but the benefit from interning the specific values within the struct is still unclear to me at the moment - a small experiment I ran showed no improvement in doing this, and depending on the nature of the data, interning the individual values could actually be _worse_ than not interning them (although I think it's unlikely we see this show up often in practice).

Brian-McM (approved):

PR #10045, opened 2025-03-24
Update labels controller to utilize syncer for updates

MichalFupso:
## Description
Update labels controller to utilize syncer updates to sync labels between k8s and calico nodes 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


kube-controllers/pkg/controllers/node/labels.go
```diff
@@ -58,161 +104,205 @@ func (c *nodeLabelController) RegisterWith(f *utils.DataFeed) {
...
-// onUpdate receives node objects and maintains the mapping of Kubernetes nodes to Calico nodes.
 func (c *nodeLabelController) onUpdate(update bapi.Update) {
+	switch update.KVPair.Key.(type) {
+	case model.ResourceKey:
+		switch update.KVPair.Key.(model.ResourceKey).Kind {
+		case apiv3.KindNode:
+			c.syncerUpdates <- update.KVPair
+		}
+	}
+}
```
lucastigera, line 118: `switch` with a singe `case` looks odd to me.
Is there a reason to not use `if`?

```
    if key, ok := update.KVPair.Key.(model.ResourceKey); ok && key.Kind == apiv3.KindNode {
        c.syncerUpdates <- update.KVPair
    }
```

kube-controllers/pkg/controllers/node/labels.go
```diff
@@ -58,161 +104,205 @@ func (c *nodeLabelController) RegisterWith(f *utils.DataFeed) {
...
+	case model.KVPair:
+		switch update.Key.(type) {
+		case model.ResourceKey:
+			switch update.Key.(model.ResourceKey).Kind {
+			case apiv3.KindNode:
+				c.handleNodeUpdate(update)
+			}
+		}
+	}
+}
```
lucastigera, line 138: Same here. Nested switches with a single case. Are you thinking of expanding the logic later and paving the road now?

MichalFupso: It was more to have a similar structure as our other controllers. There are no plans to expand the logic as of now.

kube-controllers/pkg/controllers/node/labels.go
```diff
@@ -58,161 +104,205 @@ func (c *nodeLabelController) RegisterWith(f *utils.DataFeed) {
...
-						break
-					}
-				}
-			default:
-				// Shouldn't have any other kinds show up here.
-				logrus.Warnf("Unexpected kind received over syncer: %s", update.KVPair.Key)
+		return
+	}
+
+	n := update.Value.(*apiv3.Node)
```
lucastigera, line 160: Is there a chance that `update.Value` has a type other than `*apiv3.Node`?

MichalFupso: We're calling it from `handleUpdate()` after confirming the KVPair is `apiv3.Node`, there should not be a situation where it's something other than that

kube-controllers/pkg/controllers/node/labels.go
```diff
@@ -58,161 +104,205 @@ func (c *nodeLabelController) RegisterWith(f *utils.DataFeed) {
...
-// getCalicoNode returns the Calico node name for the given Kubernetes node name, as it exists in the syncer's cache,
-// and a boolean indicating cache hit or miss.
-func (c *nodeLabelController) getCalicoNode(kn string) (string, bool) {
-	c.Lock()
-	defer c.Unlock()
-	cn, ok := c.nodemapper[kn]
-	return cn, ok
+func (c *nodeLabelController) syncAllNodesLabels() {
+	if c.syncStatus != bapi.InSync {
+		logrus.WithField("status", c.syncStatus).Debug("Not in sync, skipping node sync")
```
lucastigera, line 216: The message says 'skipping node sync,' but the code doesn’t seem to implement a mechanism to skip. Is there a return missing? Is the message incorrect, or did I misunderstand the code? hehe

kube-controllers/pkg/controllers/node/labels.go
```diff
@@ -58,161 +104,205 @@ func (c *nodeLabelController) RegisterWith(f *utils.DataFeed) {
...
+		logrus.WithError(err).Info("Unable to get corresponding k8s node name, skipping")
+	} else if kn != "" {
+		// Create a mapping from Kubernetes node -> Calico node.
+		logrus.Debugf("Mapping k8s node -> calico node. %s -> %s", kn, n.Name)
+
+		c.k8sNodeMapper[kn] = n.Name
+
+		node, err := c.nodeLister.Get(kn)
+		if err != nil {
+			logrus.WithError(err).WithField("node", kn).Error("Unable to get node")
```
lucastigera, line 176: Question:
Is this actually a real 'Error'? I mean, does the system detect an error but still allow the code to continue? 
Should it be a 'Warning' instead, or should a `return` be added to stop execution and just after logging the error?

kube-controllers/pkg/controllers/node/labels.go
```diff
@@ -58,161 +104,205 @@ func (c *nodeLabelController) RegisterWith(f *utils.DataFeed) {
...
+
+			if node != nil && c.syncStatus == bapi.InSync {
+				// Only trigger the update if the k8s node exists
+				c.k8sNodeUpdate <- node
 			}
+			return
 		}
 
+		c.calicoNodeCache[n.Name] = n
 	}
```
lucastigera, line 192: Just a suggestion to exit early, in order to avoid nested conditions.

```suggestion
	if err != nil {
		logrus.WithError(err).Info("Unable to get corresponding k8s node name, skipping")
		return
	} 
	if kn == "" {
		logrus.WithError(err).Info("Corresponding k8s node name is empty, skipping")
		return
	} 	

	// Create a mapping from Kubernetes node -> Calico node.
	logrus.Debugf("Mapping k8s node -> calico node. %s -> %s", kn, n.Name)

	c.k8sNodeMapper[kn] = n.Name

	node, err := c.nodeLister.Get(kn)
	if err != nil {
		logrus.WithError(err).WithField("node", kn).Error("Unable to get node")
	}

	_, existsCalicoNode := c.calicoNodeCache[n.Name];
	c.calicoNodeCache[n.Name] = n
	if  !existsCalicoNode {
		// If this is new calicoNode trigger sync for corresponding k8s node.
		// As we only sync labels k8s -> calico node, no need to sync when calico node already exists in our cache
		if node != nil && c.syncStatus == bapi.InSync {
			// Only trigger the update if the k8s node exists
			c.k8sNodeUpdate <- node
		}
	}
```

kube-controllers/pkg/controllers/node/labels.go
MichalFupso: You're right, missing a `return` there

kube-controllers/pkg/controllers/node/labels.go
MichalFupso: It is a real error, as the nodeLister should be able to get the node. If we run into an error getting the k8snode I think it's still good to store the calicoNode update, but don't trigger the sync

lucastigera (approved):

PR #10081, opened 2025-03-27
Refactor UTs for endpoint nftable/iptables rules 

mazdakn:
## Description

Refactor UTs for endpoint rules in nftables/iptables to get rid of many duplication.
This PR:
- Adds a rule builder in UTs to prevent copy/pasting all rules in each test case.
- Removes duplicates tests related to flowlogs. Instead run the same tests with and without flowlogs enabled. (no other tests have changed)

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


caseydavenport (approved):
I'm good with the approach here - not without its downsides, but I think it's better than what we have!

I haven't looked in too much detail, just more generally at the strategy.

PR #9937, opened 2025-03-05
make CRD management optional in helm chart

caseydavenport:
## Description

In case anyone wants to opt-out of CRD management by the operator

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add helm values.yaml config option for operator management of CRDs (default: true) 
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #10116, opened 2025-03-31
[EV-5709] Remove descriptions from libcalico-go CRDs

dimitri-nicolo:
## Description

The secret created by helm encapsulates information from every CRD into one secret which exceeds the allowable size limit. The first step in mitigating the issue is to remove the descriptions, by setting the `maxDescLen=0` in the execution of `controller-gen`  in libcalico-go.

Felixconfig CRD will keep its descriptions; removing them causes a build failure.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
CRDs are generated without a description.
```

## Reminder for the reviewer

Other optional labels:


libcalico-go/config/crd/crd.projectcalico.org_ipamblocks.yaml
```diff
@@ -19,48 +19,23 @@ spec:
...
-              description:
-                IPAMBlockSpec contains the specification for an IPAMBlock
-                resource.
               properties:
                 affinity:
-                  description: |-
-                    Affinity of the block, if this block has one. If set, it will be of the form
-                    "host:<hostname>". If not set, this block is not affine to a host.
                   type: string
                 allocations:
```
dimitri-nicolo, line 31: The patch starts at line 31 after the descriptions are removed.

libcalico-go/Makefile
```diff
@@ -53,19 +53,25 @@ gen-files: gen-crds
...
-	# Patch in manual tweaks to the generated CRDs.
-	patch -p2 < patches/0001-Add-nullable-to-IPAM-block-allocations-field.patch
 	# Remove the first yaml separator line.
 	$(DOCKER_GO_BUILD) sh -c 'find ./config/crd -name "*.yaml" | xargs sed -i 1d'
 	# Add K8S AdminNetworkPolicy CRD
 	curl -sSfL --retry 3 $(NETPOL_CRD_URL)/$(NETPOL_ANP_CRD) -o ./config/crd/$(NETPOL_ANP_CRD)
 	curl -sSfL --retry 3 $(NETPOL_CRD_URL)/$(NETPOL_BANP_CRD) -o ./config/crd/$(NETPOL_BANP_CRD)
 	# Run prettier to fix indentation
 	docker run --rm --user $(id -u):$(id -g) -v $(CURDIR)/config/crd/:/work/config/crd/ tmknom/prettier --write --parser=yaml /work
+	# Patch in manual tweaks to the generated CRDs.
```
dimitri-nicolo, line 73: Moved the patch to after the prettification, so that it's applied to the final YAML saved to config/crd

caseydavenport (approved):
I wonder if we can just have the felix documentation generation run against the "with descriptions" version of the CRDs? Then the ones we ship would all have the descriptions stripped?

Minor thing, and I don't think we need to change in this PR - but might be slightly nicer.

james-callahan:
:( this makes `kubectl explain` output much worse.

Consider doing this in *just* the helm chart rather than the crds that everything uses?

caseydavenport:
Yeah, I agree - @james-callahan could you open an issue for us to track?

james-callahan:
> Yeah, I agree - @james-callahan could you open an issue for us to track?

https://github.com/projectcalico/calico/issues/11256

stevehipwell:
@caseydavenport given that Helm shouldn't be used to manage CRDs, and more specifically the fact that the CRDs in the Helm chart regularly change resulting in CRD drift unless they're managed externally; would it not make sense to support the operator managing the Calico CRDs directly instead of having them in the Helm chart in the first place? You could always have an option to still manage the CRDs externally if required and valid.

caseydavenport:
@stevehipwell probably best to discuss on the linked issue, but yes I agree.

>  would it not make sense to support the operator managing the Calico CRDs directly instead of having them in the Helm chart in the first place?

In fact, we already have rolled out a change to get the operator managing CRDs instead of the helm chart - with one small exception, which is that sometimes users want to deploy FelixConfiguration via the helm chart itself, which right now relies on the helm chart createing the FelixConfiguration CRD.

stevehipwell:
@caseydavenport which release supports this?

caseydavenport:
Looks like it was first in v3.30: https://github.com/projectcalico/calico/pull/9937

caseydavenport:
So, setting that option to `true` and passing `skip-crds` to `helm install` should do the trick IIUC.

stevehipwell:
Thanks @caseydavenport.

PR #10132, opened 2025-04-02
Use operator for in-repo e2e tests

caseydavenport:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


node/tests/k8st/infra/values.yaml
```diff
@@ -1,40 +1,16 @@
...
-
-# Configure the images to use when generating manifests.
-node:
-  image: docker.io/library/node
-calicoctl:
-  image: docker.io/calico/ctl
-typha:
-  image: docker.io/calico/typha
+installation:
   enabled: true
```
coutinhop, line 2: this line pertained to `typha:`, the operator takes care of that automatically, right? I think this should be removed...

node/tests/k8st/infra/build-operator.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "Cloning https://github.com/${REPO} @ ${BRANCH}"
+git clone --depth=1 https://github.com/${REPO} -b ${BRANCH} operator
+
+# Modify the versions that are in-use to match our locally built images.
+pushd operator
+make build/_output/bin/gen-versions
+build/_output/bin/gen-versions -os-versions=../calico_versions.yml > pkg/components/calico.go
+
+# Modify pull policy to be "Never".
+find . -name '*.go' | xargs sed -i 's/PullIfNotPresent/PullNever/g'
```
coutinhop, line 21: :heart:

coutinhop (approved):
LGTM

node/tests/k8st/infra/build-operator.sh
```diff
@@ -0,0 +1,30 @@
...
+BRANCH=${BRANCH:-master}
+
+rm -rf operator/
+echo "Cloning https://github.com/${REPO} @ ${BRANCH}"
+git clone --depth=1 https://github.com/${REPO} -b ${BRANCH} operator
+
+# Modify the versions that are in-use to match our locally built images.
+pushd operator
+make build/_output/bin/gen-versions
+build/_output/bin/gen-versions -os-versions=../calico_versions.yml > pkg/components/calico.go
```
matthewdupre, line 18: I'm wondering if there's a practical way to generate this file from tigera/operator:config/calico_versions.yml.

As it stands there are only relatively minor variations from that file (mostly master vs latest-amd64 but it's not quite a simple swap).

For older branches it's also a little annoying: the operator tends to pin a release version (e.g. v3.28.2), but I guess the intention here is to always use the latest-amd64 tag regardless of what branches are in use?  Might be clearer to swap everything to test-build if that's the case?

matthewdupre (commented):
:heart:

node/tests/k8st/infra/values.yaml
```diff
@@ -1,40 +1,24 @@
...
-dikastes:
-  image: docker.io/calico/dikastes
-csi-driver:
-  image: docker.io/calico/csi-driver
 
-# CRDs are installed as part of cluster creation, so don't need them in
-# the test manifest.
-includeCRDs: false
+goldmane:
+  enabled: false
```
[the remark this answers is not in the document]
caseydavenport: Yeah! I was planning on doing this as a part 2 follow-on, but could certainly enable from the get-go.

node/tests/k8st/deploy_resources_on_kind_cluster.sh
```diff
@@ -114,10 +44,44 @@ function wait_pod_ready() {
...
+
 echo "Wait for Calico to be ready..."
-wait_pod_ready -l k8s-app=calico-node -n kube-system
-wait_pod_ready -l k8s-app=calico-kube-controllers -n kube-system
+wait_pod_ready -l k8s-app=calico-node -n calico-system
+wait_pod_ready -l k8s-app=calico-kube-controllers -n calico-system
+wait_pod_ready -l k8s-app=calico-apiserver -n calico-apiserver
 wait_pod_ready -l k8s-app=kube-dns -n kube-system
+wait_pod_ready calicoctl -n kube-system
 echo "Calico is running."
```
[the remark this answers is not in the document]
caseydavenport: Yeah, we could do that instead pretty easily.

node/tests/k8st/infra/build-operator.sh
caseydavenport: yeah, the intention is that the version we build is always going to install the `latest` images that were built as part of the CI / commit under test.

I think this will work for release-branches as well. We could change how we build / tag images for CI so that they match the ones expected for the operator that is being pulled in - that would allow us to remove this I think?

Well-worth doing IMO, but I think that's a phase 2 task.

node/tests/k8st/tests/test_simple.py
```diff
@@ -123,136 +123,3 @@ def test_default_pods_running(self):
...
-        # We do it this way - instead of one-shot pods that are
-        # created, try wget, and then exit - because it takes a
-        # relatively long time (7 seconds?) in this test setup for
-        # Calico routing and policy to be set up correctly for a newly
-        # created pod.  In particular it's possible that connection
-        # from a just-created pod will fail because that pod's IP has
-        # not yet propagated to the IP set for the ingress policy on
-        # the server pod - which can confuse test code that is
-        # expecting connection failure for some other reason.
-        kubectl("run access -n policy-demo" +
```
caseydavenport, line 147: Remove this test for a couple of reasons:

- It deploys a pod with floatingIPs, which isn't supported in the operator.
- However, it doesn't actually use that floating IP (so we're not losing test coverage of it)
- This is a duplicate of tests that are run in the e2e CI target.

PR #10138, opened 2025-04-03
confd: LocalBGPPeerWatcher has no in-sync message.

aaaaaaaalex:
## Description

BGP peerings could get generated without waiting for its input

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/confd/Makefile
+++ b/confd/Makefile
@@ -58,6 +58,7 @@ test-kdd: bin/confd-$(ARCH) bin/kubectl bin/bird bin/bird6 bin/calico-node bin/c
 		-v $(CURDIR)/bin:/calico/bin/ \
 		-v $(CURDIR)/etc/calico:/etc/calico/ \
 		-v $(CURDIR)/../:/go/src/github.com/projectcalico/calico:rw \
+		-v /var/run/calico/endpoint-status \
 		-e GOPATH=/go \
 		-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
 		-e FELIX_TYPHAADDR=127.0.0.1:5473 \
@@ -91,6 +92,7 @@ test-etcd: bin/confd-$(ARCH) bin/etcdctl bin/bird bin/bird6 bin/calico-node bin/
 		-v $(CURDIR)/bin:/calico/bin/ \
 		-v $(CURDIR)/etc/calico:/etc/calico/ \
 		-v $(CURDIR):/go/src/$(PACKAGE_NAME):rw \
+		-v /var/run/calico/endpoint-status \
 		-e GOPATH=/go \
 		-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
 		-e UPDATE_EXPECTED_DATA=$(UPDATE_EXPECTED_DATA) \
--- a/confd/pkg/backends/calico/client.go
+++ b/confd/pkg/backends/calico/client.go
@@ -192,7 +192,7 @@ func NewCalicoClient(confdConfig *config.Config) (*client, error) {
 	c.localBGPPeerWatcher.Start()
 
 	// Create a conditional that we use to wake up all of the watcher threads when there
-	// may some actionable updates.
+	// may be some actionable updates.
 	c.watcherCond = sync.NewCond(&c.cacheLock)
 
 	// Increment the waitForSync wait group.  This blocks the GetValues call until the
@@ -292,9 +292,11 @@ func NewCalicoClient(confdConfig *config.Config) (*client, error) {
 	return c, nil
 }
 
+// Strings for keying in-sync messages by their upstream source.
 var (
-	SourceSyncer         string = "SourceSyncer"
-	SourceRouteGenerator string = "SourceRouteGenerator"
+	SourceSyncer              string = "SourceSyncer"
+	SourceRouteGenerator      string = "SourceRouteGenerator"
+	SourceLocalBGPPeerWatcher string = "LocalBGPPeerWatcher"
 )
 
 // client implements the StoreClient interface for confd, and also implements the
@@ -425,7 +427,7 @@ func (c *client) ExcludeServiceAdvertisement() bool {
 	return false
 }
 
-// OnInSync handles multiplexing in-sync messages from multiple data sources
+// OnSyncChange handles multiplexing in-sync messages from multiple data sources
 // into a single representation of readiness.
 func (c *client) OnSyncChange(source string, ready bool) {
 	c.cacheLock.Lock()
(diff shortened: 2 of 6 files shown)


test.out
```diff
@@ -0,0 +1,1922 @@
+make: Entering directory '/home/alex/Repositories/github.com/projectcalico/calico/confd'
```
fasaxc, line 1: Looks like an unintended check in?

fasaxc (changes requested):

libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -138,6 +134,7 @@ func (w *FileWatcher) runFsnotifyWatcher(watcher *fsnotify.Watcher) error {
 					}
 				}
 			}
+			w.callbacks.OnInSync(true)
```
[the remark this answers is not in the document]
aaaaaaaalex: Just calling it from one place now (after initial poll).

Wanted to move the initial poll up a layer - if for some reason, the watcher was nil, the poll would not happen when its within the 'runFsnotifyWatcher`. Unlikely, but felt diligent.

confd/Makefile
```diff
@@ -52,12 +53,14 @@ test: ut test-kdd test-etcd
...
 	-git clean -fx etc/calico/confd
 	mkdir -p tests/logs
+	mkdir -p tests/endpoint-status
 	docker run --rm --net=host \
 		-v $(CERTS_PATH):/home/user/certs \
 		-v $(CURDIR)/tests/:/tests/ \
 		-v $(CURDIR)/bin:/calico/bin/ \
 		-v $(CURDIR)/etc/calico:/etc/calico/ \
 		-v $(CURDIR)/../:/go/src/github.com/projectcalico/calico:rw \
+		-v $(CURDIR)/tests/endpoint-status:/var/run/calico/endpoint-status \
```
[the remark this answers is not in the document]
fasaxc: Ah

confd/pkg/backends/calico/client.go
```diff
@@ -439,13 +441,13 @@ func (c *client) OnSyncChange(source string, ready bool) {
 	log.Infof("Source %v readiness changed, ready=%v", source, ready)
 
 	// Check if we are fully in sync, before applying this change.
-	oldFullSync := c.sourceReady[SourceSyncer] && c.sourceReady[SourceRouteGenerator]
+	oldFullSync := c.sourceReady[SourceSyncer] && c.sourceReady[SourceRouteGenerator] && c.sourceReady[SourceLocalBGPPeerWatcher]
```
fasaxc, line 444: Abstract as `c.inSync()`?

libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -167,6 +164,9 @@ func (w *FileWatcher) runWatcher() {
 			log.WithError(err).Info("Error initializing fsnotify. Falling back to polling.")
 		}
 
+		// Get current state of the directory and emit initial events.
+		w.scanDirectory()
```
fasaxc, line 168: This ends up scanning the directory twice per loop if we're polling. I think we need a scan _here_ to make sure that we don't miss any events so perhaps just remove the one in the 

```
case <-w.pollTicker.C:
``` 
arm of the select below?  Then the poll ticker will just continue the loop, which will try to make a new watcher and then do this poll.

libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -167,6 +164,9 @@ func (w *FileWatcher) runWatcher() {
...
+		w.callbacks.OnInSync(true)
```
fasaxc, line 169: I _think_ it'd be better to only call this once.  Downstream users getting the event might be confused by getting dupes, or attach meaning to it that we didn't intend.

libcalico-go/lib/epstatusfile/status_file_watcher_test.go
```diff
@@ -125,6 +153,9 @@ var _ = Describe("Workload endpoint status file watcher test", func() {
 
 		filePath := filepath.Join(statusDir, "pod1")
 		Eventually(haveEvents, "5s", "1s").WithArguments(filePath, []string{"create"}).Should(BeTrue())
```
fasaxc, line 155: These assertions look race-prone to me.  haveEvents is reading data that's modified by callback but it doesn't use a mutex

libcalico-go/lib/epstatusfile/status_file_watcher_test.go
```diff
@@ -125,6 +153,9 @@ var _ = Describe("Workload endpoint status file watcher test", func() {
...
+		Eventually(r.inSyncEvents).ShouldNot(BeEmpty())
```
fasaxc, line 156: Eventually with a slice is a smell.  Due to Go's value semantics, this is equivalent to 
```
s := r.inSyncEvents
Eventually(s).ShouldNot(...)
```
So, apart from racing with writes to `inSyncEvents` the code doesn't actually re-read the slice each time through Eventually's check loop.

You should:
- Protect `inSyncEvents` with a mutex
- Pass a func to Eventually

fasaxc (changes requested):

libcalico-go/lib/epstatusfile/status_file_watcher.go
aaaaaaaalex: Sorted. I also noticed we're not closing the watcher anywhere when we re-iterate, so I added that code too under `newFsnotifyWatcher()`

libcalico-go/lib/epstatusfile/status_file_watcher.go
aaaaaaaalex: I believe I only left one call behind now

libcalico-go/lib/epstatusfile/status_file_watcher_test.go
aaaaaaaalex: Locking everything in `eventRecorder` now

libcalico-go/lib/epstatusfile/status_file_watcher.go
fasaxc: I mean once total, not once per loop.

libcalico-go/lib/epstatusfile/status_file_watcher_test.go
```diff
@@ -65,6 +70,24 @@ func (e *eventRecorder) OnFileDeletion(fileName string) {
...
+func (e *eventRecorder) OnInSync(s bool) {
+	e.Lock()
+	defer e.Unlock()
+	e.inSyncEvents = append(e.inSyncEvents, s)
+}
+
+func (e *eventRecorder) Events() map[string][]string {
+	e.Lock()
+	defer e.Unlock()
+	return e.fileNameToEvents
```
fasaxc, line 82: You release the lock before the map is accessed in the caller.  Easiest option is to clone it
```suggestion
	return maps.Clone(e.fileNameToEvents)
```

fasaxc (changes requested):

libcalico-go/lib/epstatusfile/status_file_watcher_test.go
aaaaaaaalex: d'oh!

libcalico-go/lib/epstatusfile/status_file_watcher.go
```diff
@@ -80,6 +81,14 @@ func (w *FileWatcher) Start() {
 }
 
 func (w *FileWatcher) newFsnotifyWatcher() error {
+	oldWatcher := w.fsWatcher
```
fasaxc, line 84: This also gets closed by the defer in `runWatcher()`, might end up closing it twice.

libcalico-go/lib/epstatusfile/status_file_watcher.go
aaaaaaaalex: Missed this one until now. I've wrapped the call in a Once

fasaxc (approved):

PR #10145, opened 2025-04-04
Plumb watch bookmarks through to clientv3

caseydavenport:
## Description

Not sure if we need this yet, but I noticed that the main k8s API server
sets watchBookmarks=true on the watch requests it makes against the
Calico API server, but we aren't respecting that.

Adding this to the clientv3 API would be the first step towards allowing
our apiserver to handle those requests.

I don't think this is causing any problems - clients MUST not expect the
server to generate watch bookmarks anyway, so this would just be a small
optimization if anything.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


fasaxc:
I don't think this will just work.  The problem is the conversion step; all the converters need to be modified to allow an "invalid" bookmark object through.

caseydavenport:
@fasaxc yeah, I threw this up optimistically but pretty quickly realized there was more to do and haven't gotten back to it!

PR #10178, opened 2025-04-09
Fix confd status file churn

fasaxc:
## Description

- Fix that felix would rewrite endpoints status files every resync because the JSON didn't round trip (empty slice returned as nil slice).
- Let confd retry reading the endpoint status file once after 50ms to avoid spurious errors caused by felix being mid-write.
- In confd, only store local endpoints if they actually have a BGPPeer name populated.  Suppress BGPPeer recalculation if the changed endpoint wasn't actually a BGP peer.
- Tweak tests to use a random temp dir so that they can be run outside of the UT container.
- Fix file permissions on the endpoint-status dir.  655 should be 755 (need the execute bit to allow for access to files within the directory).

## Related issues/PRs

CORE-11263

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/epstatusfile/status_file_writer.go
+++ b/libcalico-go/lib/epstatusfile/status_file_writer.go
@@ -181,13 +181,22 @@ func WorkloadEndpointToWorkloadEndpointStatus(ep *proto.WorkloadEndpoint) *Workl
 		peerName = ep.LocalBgpPeer.BgpPeerName
 	}
 	epStatus := &WorkloadEndpointStatus{
-		IfaceName:   ep.Name,
-		Mac:         ep.Mac,
-		Ipv4Nets:    ep.Ipv4Nets,
-		Ipv6Nets:    ep.Ipv6Nets,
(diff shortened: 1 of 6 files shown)


libcalico-go/lib/epstatusfile/status_file_writer.go
```diff
@@ -181,15 +181,24 @@ func WorkloadEndpointToWorkloadEndpointStatus(ep *proto.WorkloadEndpoint) *Workl
...
 	return epStatus
 }
 
+func normaliseZeroLenSlice[T any](nets []T) []T {
+	if len(nets) == 0 {
+		return nil
+	}
+	return nets
+}
+
```
coutinhop, line 201: nit (and a very small one at that): this looks like it could be useful else where, we could at some point in the future move this to a more general utils package (and rename `nets` -> `slice` or something)

coutinhop (approved):

PR #10182, opened 2025-04-09
Refactor stream implementation in Goldmane

caseydavenport:
## Description

The stream implementation had a few bottlenecks that were showing up in larger tests.

1. Registering and closing streams went via the main goroutine, which meant they would be blocked by other operations - including stream backfilling.
1. If a stream was closed by a client during backfill, there was no way to tell the main loop to cancel the backfill operation.
1. Stream writes were usually fast, but if enough back pressure was applied on the way out it could freeze the main log aggregator loop.
1. Unpacking every Flow for each stream was using a lot of memory. 

To resolve / mitigate these:

- Stream registration and unregistration now happens on a separate goroutine, allowing for streams to terminate during other operations. 
- Stream termination cancels a per-stream Context, which can be used to terminate operations writing to the stream.
- Add a buffered channel between the main aggreagtor loop and the stream manager. 
- Increase the channel buffering per-stream based on the request. 100 was not nearly enough for very large clusters.
- Pass references to the data all the way to the point that we convert to protobuf format, allowing us to re-use the proto.FlowUpdate structure for each event, substantially reducing the memory impact of a Stream.

This is really just a short-term mitigation. A larger re-work of this is probably in order. Namely, a couple of problems still exist:

- For really large clusters, or Streams that ask for huge amounts of data (e.g., O(1M) flows), backfilling the stream can still delay the main loop causing dropped flows. This is just a function of iterating a large number of flows.
- Rollover() currently iterates the entire bucket to send flows to the stream. This can cause the rollover() event to impact flow ingestion as well.

To fix these, we probably need a slightly different concurrency model that allows for streams to execute both backfill and rollover logic asynchronously from the main ingestion loop. This will take some care to ensure we're handling data access correctly.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


goldmane/pkg/aggregator/bucketing/bucket_ring.go
```diff
@@ -26,6 +27,8 @@ import (
 	"github.com/projectcalico/calico/libcalico-go/lib/set"
 )
 
+var StopBucketIteration = errors.New("stop bucket iteration")
```
Brian-McM, line 30: Comment for future work, would be nice to have a single constant in a collection package that represents short circuiting iterators.

Actually wondering if we should switch to using `iter.Seq` for our iterators in the future, it handles short circuiting.

goldmane/pkg/aggregator/stream_manager.go
```diff
@@ -90,38 +41,111 @@ type streamManager struct {
 
 	// closedStreamsCh is a channel on which to receive UUIDs of streams that have been closed.
 	closedStreamsCh chan string
+
+	// streamRequests is the channel to receive requests for new streams.
+	streamRequests chan *streamRequest
+
+	// backfillRequests is a channel to send backfill requests on to the log aggregator.
+	backfillRequests chan *Stream
```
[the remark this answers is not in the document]
caseydavenport: I'm not sure I follow - this receives Streams and triggers backfilling of those streams. What else do you see it doing?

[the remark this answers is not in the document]
Brian-McM: Sorry I got confused, please ignore this now.

goldmane/pkg/aggregator/aggregator.go
```diff
@@ -50,31 +55,36 @@ var (
 		Help: "Total number of flows received by Goldmane aggregator.",
 	})
 
-	flowIndexLatency = prometheus.NewHistogram(prometheus.HistogramOpts{
-		Name: "goldmane_aggr_flow_process_latency",
-		Help: "Histogram measuring the time taken to index a flow.",
+	flowIndexLatency = cprometheus.NewSummary(prometheus.SummaryOpts{
```
WilliamTigera, line 58: Why the change to using the cprometheus package?

caseydavenport: @WilliamTigera cprometheus is a Calico wrapper around the prometheus package that auto configures some fields - specifically for Summary types it automatically configures the `Objectives` to reasonable values and avoids me needing to duplicate it here.

https://github.com/projectcalico/calico/blob/ab8ee9a8ee125bb2ad25a19a2d285aec8b2b621c/libcalico-go/lib/prometheus/summary.go#L21-L28

WilliamTigera (approved):

Issue #10112, opened 2025-03-30
Only the 1st Cluster IP Is Advertised via BGP With Dual Stack


nick-oconnor:
## Expected Behavior
On dual stack clusters with external BGP advertisement, I expect that both IPv4 and IPv6 cluster IPs are advertised for `NodePort` type services with `externalTrafficPolicy: Local`.

## Current Behavior
On dual stack BGP clusters, only the 1st `ClusterIP` is advertised via BGP. This is either IPv4 or IPv6, depending on which is given for `clusterIP`, despite `clusterIPs` having two addresses. You can work around the issue by either creating a service per IP family or using external IPs, I would like for this to work.

## Possible Solution
It looks like the [code](https://github.com/projectcalico/calico/blob/v3.29.3/confd/pkg/backends/calico/routes.go#L236) is using `clusterIP`. Is it feasible to add all values from `clusterIPs` instead?

## Steps to Reproduce (for bugs)
1. Setup dual stack + ToR router peered cluster with IPv4 and IPv6 CIDRs
2. Create a `NodePort` service with `ipFamilyPolicy: RequireDualStack` and `externalTrafficPolicy: Local`
3. Inspect the external router's IPv4 and IPv6 routing tables

## Context
I'd like to route service cluster IPs externally with BGP multipath.

## Your Environment
* Calico version: 3.29.3
* Calico dataplane (iptables, windows etc.): eBPF w/ DSR, no encap, no NAT, no kube-proxy, IPv4 and IPv6 BGP peering with ToR router
* Orchestrator version (e.g. kubernetes, mesos, rkt): tigera-operator
* Operating System and version: Ubuntu 24.04.2

nick-oconnor:
Hi @nelljerram! Yep. I have a BGP peering per IP family. This works for services with both IPv4 and IPv6 external IPs. It looks like the code is only using the address from `clusterIP` and ignoring `clusterIPs`.

nick-oconnor:
Sure! On it.

nick-oconnor:
I didn't realize this required a CLA. I've have to submit the CLA for approval.

nick-oconnor:
Approved! Sorry for the PR noise.

PR #10183, opened 2025-04-09
Export cluster IPs for all IP families

nick-oconnor:
## Description

Previously, only the primary cluster IP for a service was exported. This change exports all cluster IPs for a service. Per the k8s [API spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#servicespec-v1-core):

> if clusterIPs is not specified for a service, it will be initialized from the clusterIP field

I expect this is backward compatible with k8s v1.20+ (https://github.com/kubernetes/kubernetes/commit/6675eba3eff1c8e565c4060a9c1396f75da7cc3e).

## Related issues/PRs

Fixes https://github.com/projectcalico/calico/issues/10112

## Release Note

```release-note
Export cluster IPs for all IP families
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/confd/pkg/backends/calico/routes.go
+++ b/confd/pkg/backends/calico/routes.go
@@ -456,7 +456,7 @@ func (rg *routeGenerator) advertiseThisService(svc *v1.Service, ep *v1.Endpoints
 	}
 
 	// also do nothing if no cluster IPs are assigned
-	if len(svc.Spec.ClusterIPs) == 0 {
+	if svc.Spec.ClusterIP == "" || svc.Spec.ClusterIP == "None" {
 		logc.Debug("Skipping service with no cluster IPs")
 		return false
 	}
--- a/confd/pkg/backends/calico/routes.go
+++ b/confd/pkg/backends/calico/routes.go
@@ -245,7 +245,7 @@ func (rg *routeGenerator) getAllRoutesForService(svc *v1.Service) []string {
 	routes := make([]string, 0)
 	if rg.client.AdvertiseClusterIPs() {
 		// Only advertise cluster IPs if we've been told to.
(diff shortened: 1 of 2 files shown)


confd/pkg/backends/calico/routes.go
```diff
@@ -455,9 +455,9 @@ func (rg *routeGenerator) advertiseThisService(svc *v1.Service, ep *v1.Endpoints
 		return false
 	}
 
-	// also do nothing if the clusterIP is empty or None
-	if svc.Spec.ClusterIP == "" || svc.Spec.ClusterIP == "None" {
-		logc.Debug("Skipping service with no cluster IP")
+	// also do nothing if no cluster IPs are assigned
+	if len(svc.Spec.ClusterIPs) == 0 {
```
tomastigera, line 459: should this also include the previous checks for backward compatibility? Specs do not say that only thenew ClusterIPs is valid. Also it may contain the special values 

> Valid values are "None", empty string (""), or a valid IP address. Setting this to "None" makes a "headless service"

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#-strong-service-apis-strong-

nick-oconnor: Great point. I was hoping to avoid having to depend on both `clusterIP` and `clusterIPs`. I think I found the [API server](https://github.com/kubernetes/kubernetes/blob/v1.32.3/pkg/registry/core/service/storage/alloc.go#L347-L361) logic for assigning clusterIPs and it's pretty messy. I've reverted this part since I think it's cleaner and safer than trying to parse it out of `clusterIPs`.

caseydavenport (approved):

caseydavenport:
Thanks @nick-oconnor !

PR #10187, opened 2025-04-10
Add WatchList support with GenericListWatcher abstraction

liuxu623:
## Description

This PR adds WatchList-aware synchronization support for the Kubernetes backend and refactors shared list/watch behavior into a reusable `GenericListWatcher`.

Kubernetes resources that support WatchList can stream initial data through watch events using `SendInitialEvents=true`, watch bookmarks, and `ResourceVersionMatch=NotOlderThan`. Resources that do not support WatchList, older apiservers, and non-Kubernetes backends continue to use List+Watch.

## Motivation

The previous `watcherCache` implementation owned list calls, watch creation, retry throttling, revision tracking, stale-entry detection, and event processing in one place. That made it difficult to add Kubernetes WatchList support without duplicating list/watch behavior or changing etcd semantics.

This change separates the shared state machine from backend-specific behavior:

- `GenericListWatcher` owns common retry, revision, initial sync, and event-loop logic.
- Kubernetes and etcd provide backend-specific `ListWatchBackend` implementations.
- `watcherCache` focuses on converting `EventHandler` callbacks into syncer updates.

WatchList reduces initial-sync memory pressure on Kubernetes clusters that support it because initial objects can be streamed through watch events instead of first materializing a full List response in the watcher cache.

### Before

```text
watcherCache
|-- List/Watch logic embedded
|-- Error handling mixed with event processing
|-- Retry/throttle logic embedded
|-- Revision tracking embedded
|-- oldResources map copied for deletion detection
`-- Tightly coupled to concrete backend watch behavior
```

### After

```text
watcherCache
|-- Implements api.EventHandler
|-- Handles OnAdd / OnUpdate / OnDelete / OnSync / OnError
|-- Uses resyncEpoch for stale-entry deletion detection
`-- Delegates list/watch mechanics to backend ListAndWatch or generic fallback

GenericListWatcher
|-- Common retry/throttle logic
|-- Current revision tracking
|-- Repeated error counting
|-- Initial sync state
`-- Shared event loop and basic watch event handling

k8s.ListWatcher
|-- WatchList initial sync when supported
|-- Bookmark handling
|-- Fallback to List+Watch
`-- Kubernetes-specific error handling

etcdv3.ListWatcher
|-- Traditional List+Watch
`-- etcd-specific watch error handling
```

### Architecture

```mermaid
flowchart TB
    subgraph WatcherSyncer["WatcherSyncer Layer"]
        WS[WatcherSyncer<br/>Coordinates watcher caches]
        WC1[watcherCache #1]
        WC2[watcherCache #2]
        WC3[watcherCache #N]
        WS --> WC1
        WS --> WC2
        WS --> WC3
    end

    subgraph EventHandling["Event Handling"]
        EH{{api.EventHandler}}
        EH1[OnResyncStarted]
        EH2[OnAdd / OnUpdate / OnDelete]
        EH3[OnSync]
        EH4[OnError]
        EH --> EH1
        EH --> EH2
        EH --> EH3
        EH --> EH4
    end

    subgraph ClientInterfaces["Client Interfaces"]
        APIClient[api.Client<br/>List + Watch]
        OptionalLAW[api.ListAndWatchClient<br/>optional ListAndWatch]
        WatchListSupport[api.WatchListSupporter<br/>optional capability]
    end

    subgraph Backends["Backend Implementations"]
        K8SC[k8s.KubeClient]
        ETCDC[etcdv3.Client]
        GenericFallback[clientListWatchBackend<br/>generic List+Watch fallback]
        K8SC -. implements .-> OptionalLAW
        ETCDC -. implements .-> OptionalLAW
        GenericFallback --> APIClient
    end

    subgraph ListWatchers["ListWatcher Implementations"]
        GLW[api.GenericListWatcher]
        K8SLW[k8s.ListWatcher]
        ETCDLW[etcdv3.ListWatcher]
        K8SLW --> GLW
        ETCDLW --> GLW
        GenericFallback --> GLW
    end

    subgraph K8sModes["Kubernetes Sync Modes"]
        WLM[WatchList mode<br/>SendInitialEvents + bookmark]
        LWM[List+Watch fallback]
        K8SLW --> WLM
        K8SLW -. fallback .-> LWM
        WatchListSupport -. controls .-> WLM
    end

    subgraph DataSources["Data Sources"]
        APIServer[(Kubernetes API server)]
        ETCD[(etcd)]
    end

    WC1 -. implements .-> EH
    WC2 -. implements .-> EH
    WC3 -. implements .-> EH

    WC1 -->|prefer optional ListAndWatch| OptionalLAW
    WC2 -->|fallback if unsupported| GenericFallback
    WC3 -->|fallback if unsupported| GenericFallback

    OptionalLAW --> K8SC
    OptionalLAW --> ETCDC

    WLM --> APIServer
    LWM --> APIServer
    ETCDLW --> ETCD

    style OptionalLAW fill:#9f9,stroke:#393
    style WLM fill:#9cf,stroke:#369
    style GLW fill:#fc9,stroke:#963
```

### Data Flow

```mermaid
sequenceDiagram
    participant WS as WatcherSyncer
    participant WC as watcherCache
    participant Client as backend client
    participant LW as backend ListWatcher
    participant GLW as GenericListWatcher
    participant Server as K8s API / etcd

    WS->>WC: run(ctx)

    alt Client implements api.ListAndWatchClient
        WC->>Client: ListAndWatch(ctx, list, handler, options)
        Client->>LW: create backend ListWatcher
    else Client only implements api.Client
        WC->>GLW: create GenericListWatcher
        WC->>LW: create generic clientListWatchBackend
    end

    loop ListAndWatch loop
        GLW->>GLW: RetryThrottleC()

        alt InitialSyncPending and WatchList supported
            GLW->>LW: CreateWatch(useWatchList=true)
            alt WatchList create succeeds
                GLW->>LW: PerformInitialSync(useWatchList=true)
                LW->>WC: OnResyncStarted()
                loop Stream initial events
                    Server-->>LW: Added / Modified / Deleted
                    LW->>GLW: UpdateRevision()
                    LW->>WC: OnAdd / OnUpdate / OnDelete
                end
                Server-->>LW: Bookmark(initial-events=true)
                LW->>GLW: UpdateRevision()
                LW->>WC: OnSync()
            else WatchList unavailable
                GLW->>LW: PerformInitialSync(useWatchList=false)
                LW->>Server: List(empty revision)
                Server-->>LW: KVPairList(revision)
                LW->>WC: OnResyncStarted()
                LW->>WC: OnAdd(kvp) for each result
                LW->>WC: OnSync()
                GLW->>LW: CreateWatch(useWatchList=false)
            end
        else List+Watch mode
            GLW->>LW: PerformInitialSync(useWatchList=false)
            LW->>Server: List(empty revision)
            Server-->>LW: KVPairList(revision)
            LW->>WC: OnResyncStarted()
            LW->>WC: OnAdd(kvp) for each result
            LW->>WC: OnSync()
            GLW->>LW: CreateWatch(useWatchList=false)
        end

        loop Watch events
            Server-->>LW: Watch event / bookmark / error
            LW->>GLW: HandleBasicWatchEvent or backend-specific handling
            GLW->>WC: OnAdd / OnUpdate / OnDelete
        end

        alt Backend-specific ListAndWatch unsupported
            Client-->>WC: ErrorOperationNotSupported
            WC->>GLW: fall back to generic List+Watch
        else Non-context terminal error
            Client-->>WC: error
            WC->>WS: errorSyncBackendError
        end
    end
```

#### Optional backend ListAndWatch

```go
type ListAndWatchClient interface {
    ListAndWatch(
        ctx context.Context,
        list model.ListInterface,
        handler EventHandler,
        opts ...ListWatcherOption,
    ) error
}
```

This is intentionally optional. `api.Client` remains compatible with implementations that only provide `List` and `Watch`.

#### EventHandler

```go
type EventHandler interface {
    OnResyncStarted()
    OnAdd(kvp *model.KVPair)
    OnUpdate(kvp *model.KVPair)
    OnDelete(kvp *model.KVPair)
    OnSync()
    OnError(err error)
}
```

#### ListWatchBackend

```go
type ListWatchBackend interface {
    PerformList(ctx context.Context) (*model.KVPairList, error)
    CreateWatch(ctx context.Context, useWatchList bool) (WatchInterface, error)
    HandleWatchEvent(event WatchEvent) error
    HandleListError(err error)
    HandleWatchError(err error)
    PerformInitialSync(ctx context.Context, g *GenericListWatcher, useWatchList bool) error
}
```

#### WatchListSupporter

```go
type WatchListSupporter interface {
    SupportsWatchList() bool
}
```

Backends or resource clients can implement this to disable WatchList for resources that need special sync semantics.

### 1. GenericListWatcher (`api/listwatcher.go`)

`GenericListWatcher` provides the common list/watch state machine:

- retry throttling through `MinResyncInterval`, `ListRetryInterval`, and `WatchPollInterval`
- connection timeout detection through `WatchRetryTimeout`
- current revision tracking
- repeated error counting through `MaxErrorsPerRevision`
- full resync reset through `ResetForFullResync`
- initial sync tracking through `InitialSyncPending`
- basic add/update/delete watch event handling

The constructor accepts `ListWatcherOption` values and ignores nil options defensively.

### 2. Kubernetes ListWatcher (`k8s/listwatcher.go`)

Kubernetes uses `k8s.ListWatcher`, which embeds `GenericListWatcher`.

WatchList mode:

- uses `SendInitialEvents=true`
- uses empty revision plus `ResourceVersionMatch=NotOlderThan`
- requests bookmarks
- identifies sync completion from the initial-events bookmark
- keeps the Kubernetes object in bookmark events so the final bookmark annotation can be inspected

Fallback behavior:

- if WatchList creation fails, the code falls back to List+Watch for that sync, matching client-go reflector behavior
- if a resource client reports `SupportsWatchList() == false`, initial sync uses List+Watch directly
- if `KubeClient.ListAndWatch` is unsupported for a resource type, `watcherCache` falls back to the generic `api.Client.List` plus `api.Client.Watch` path

Kubernetes-specific error handling:

- missing API / CRD: mark in sync, reset for future full resync, retry later
- resource expired / gone / too-large resource version: reset for full resync
- connection refused / too many requests: retry and reset if the connection timeout is exceeded

### 3. Profile resources

Profile resources explicitly disable WatchList:

```go
func (c *profileClient) SupportsWatchList() bool {
    return false
}
```

Profiles merge Namespace and ServiceAccount watches. The merged watcher must wait for both underlying streams to complete initial events before it can safely report sync completion, so the PR keeps Profiles on List+Watch for now.

### 4. etcd ListWatcher (`etcdv3/listwatcher.go`)

etcd uses `etcdv3.ListWatcher` with traditional List+Watch. It shares `GenericListWatcher` for retry, revision, error-count, connection-timeout, and full-resync behavior, but it does not use WatchList or bookmark sync completion.

### 5. watcherCache (`watchersyncer/watchercache.go`)

`watcherCache` now implements `api.EventHandler` and delegates list/watch mechanics:

- prefer backend-specific `api.ListAndWatchClient`
- fall back to generic List+Watch when backend-specific ListAndWatch returns `ErrorOperationNotSupported`
- forward non-context ListAndWatch errors to the syncer
- handle bookmarks in the generic fallback backend to advance the current revision
- skip malformed nil delete events instead of panicking

#### Resync Epoch Mechanism

```go
type cacheEntry struct {
    revision    string
    key         model.Key
    resyncEpoch uint64
}

type watcherCache struct {
    resyncEpoch            uint64
    lastHandledResyncEpoch uint64
    resources              map[string]cacheEntry
}
```

On resync start:

1. Increment `resyncEpoch` when the cache is non-empty.
2. Process incoming full-resync events, updating entries to the current epoch.
3. After sync, delete entries whose epoch was not refreshed.

```mermaid
sequenceDiagram
    participant Cache as watcherCache
    participant Resources as resources map
    participant Syncer as WatcherSyncer

    Note over Resources: Initial state<br/>pod-a epoch=1<br/>pod-b epoch=1<br/>pod-c epoch=1

    Cache->>Cache: OnResyncStarted()<br/>resyncEpoch = 2

    Note over Cache: Full resync receives pod-a and pod-c<br/>pod-b was deleted while disconnected

    Cache->>Resources: OnAdd(pod-a) sets epoch=2
    Cache->>Resources: OnAdd(pod-c) sets epoch=2
    Note over Resources: pod-a epoch=2<br/>pod-b epoch=1 stale<br/>pod-c epoch=2

    Cache->>Cache: OnSync()
    Cache->>Resources: scan entries with epoch < 2
    Resources-->>Cache: pod-b is stale
    Cache->>Syncer: send delete for pod-b
    Cache->>Resources: remove pod-b

    Note over Resources: Final state<br/>pod-a epoch=2<br/>pod-c epoch=2
```

This avoids copying all cached entries into a separate `oldResources` map for every resync.

### 6. watchersyncer options compatibility

`watchersyncer.New` accepts options again and passes `api.ListWatcherOptions` down into watcher caches and backend list watchers.

The previous exported option remains available:

```go
// Deprecated: use WithListWatcherOptions to set
// api.ListWatcherOptions.WatchRetryTimeout instead.
func WithWatchRetryTimeout(t time.Duration) Option
```

`ResourceType.SendDeletesOnConnFail` is also retained as a deprecated field for external API compatibility.

### Overview

This PR uses empty revision for full resync List calls. That is intentional: fallback List+Watch should list the latest consistent data, then start the following watch from the returned list revision.

WatchList also starts with empty revision and sets `ResourceVersionMatch=NotOlderThan`, allowing Kubernetes to stream initial events and signal completion with a bookmark.

### Revision Lifecycle

```mermaid
stateDiagram-v2
    [*] --> Empty: NewGenericListWatcher
    Empty --> Syncing: InitialSyncPending

    state Syncing {
        [*] --> WatchList: Kubernetes and SupportsWatchList
        [*] --> ListWatch: etcd / fallback / unsupported WatchList

        WatchList --> CreateWatchList: CreateWatch(useWatchList=true)
        CreateWatchList --> ReceiveInitialEvents: watch created
        CreateWatchList --> ListWatch: WatchList create failed
        ReceiveInitialEvents --> UpdateRev: Each event updates revision
        UpdateRev --> ReceiveInitialEvents
        ReceiveInitialEvents --> Synced: initial-events bookmark

        ListWatch --> ListAll: PerformList(empty revision)
        ListAll --> UpdateRevFromList: Use list revision
        UpdateRevFromList --> Synced: OnSync
    }

    Synced --> Watching: CreateWatch(useWatchList=false)

    state Watching {
        [*] --> WaitEvent
        WaitEvent --> ProcessEvent: Event received
        ProcessEvent --> UpdateRevision: Update CurrentRevision
        UpdateRevision --> WaitEvent
    }

    Watching --> Empty: full resync required
    Watching --> Syncing: watch recreated
```

### Revision Update Points

| Scenario | Revision source | Action |
| --- | --- | --- |
| Initial state | `""` | Request latest data |
| List+Watch list completion | `list.Revision` | `UpdateRevision()` |
| WatchList added/modified/deleted event | `event.New.Revision` or `event.Old.Revision` | `UpdateRevision()` |
| WatchList bookmark | `event.New.Revision` | `UpdateRevision()` |
| Generic fallback bookmark | `event.New.Revision` | `UpdateRevision()` |
| ResourceExpired / Gone / too-large RV | none | `ResetForFullResync()` |
| Repeated watch errors | none | `ResetForFullResync()` |

### Revision Flow

```mermaid
sequenceDiagram
    participant GLW as GenericListWatcher
    participant Backend as backend ListWatcher
    participant Server as K8s API / etcd

    Note over GLW: CurrentRevision is empty

    alt WatchList initial sync
        GLW->>Backend: CreateWatch(useWatchList=true)
        Backend->>Server: Watch(empty revision, SendInitialEvents=true, ResourceVersionMatch=NotOlderThan)
        Server-->>Backend: WatchAdded(rev=100)
        Backend->>GLW: UpdateRevision("100")
        Server-->>Backend: WatchAdded(rev=101)
        Backend->>GLW: UpdateRevision("101")
        Server-->>Backend: Bookmark(rev=101, initial-events=true)
        Backend->>GLW: UpdateRevision("101")
        Backend->>GLW: MarkInitialSyncComplete()
    else List+Watch initial sync
        Backend->>Server: List(empty revision)
        Server-->>Backend: KVPairList(rev=102)
        Backend->>GLW: UpdateRevision("102")
        Backend->>GLW: MarkInitialSyncComplete()
        GLW->>Backend: CreateWatch(useWatchList=false)
        Backend->>Server: Watch(revision=102)
    end

    loop Watch events
        Server-->>Backend: WatchModified(rev=103)
        Backend->>GLW: UpdateRevision("103")
        Server-->>Backend: Bookmark(rev=110)
        Backend->>GLW: UpdateRevision("110")
    end

    alt Full resync needed
        Backend-->>GLW: ResourceExpired / too many errors
        GLW->>GLW: ResetForFullResync()
        Note over GLW: CurrentRevision is empty<br/>InitialSyncPending = true
    end
```

## Why Empty Revision Instead of `"0"`

The fallback List+Watch path always lists with an empty revision. This keeps full resync behavior simple and consistent: List asks for the latest data, then Watch starts from the list result revision.

For Kubernetes, empty revision is also the right shape for modern apiservers that support consistent list-from-cache behavior. The implementation does not force `resourceVersion=0` for fallback lists.

For etcd, empty revision means no explicit `WithRev` is passed, so the latest data is returned.

## Compatibility Notes

- `api.Client` remains backward compatible. Backend-specific list/watch support is provided by optional `api.ListAndWatchClient`.
- Existing clients that only implement `List` and `Watch` continue through the generic List+Watch fallback.
- `backend/api.WatchOptions` uses a backend-api-defined `ResourceVersionMatch` type and translates to Kubernetes `metav1.ResourceVersionMatch` in the k8s resource layer.
- `watchersyncer.WithWatchRetryTimeout` remains as a deprecated compatibility wrapper.
- `ResourceType.SendDeletesOnConnFail` remains as a deprecated compatibility field.

## Todos

- [x] Tests
- [x] Documentation not required
- [x] Release note

## Release Note

```release-note
Add WatchList-aware synchronization support for Kubernetes backend watchers. Kubernetes resources that support WatchList can stream initial data through watch events with automatic fallback to List+Watch, while etcd and unsupported resources continue using List+Watch behavior.
```

### DIFF
--- a/libcalico-go/lib/backend/watchersyncer/watchercache.go
+++ b/libcalico-go/lib/backend/watchersyncer/watchercache.go
@@ -211,7 +211,7 @@ func (wc *watcherCache) OnSync() {
 	if wc.resyncEpoch != wc.lastHandledResyncEpoch {
 		// We just completed a resync with a non-empty cache, we need to scan
 		// the cache to look for resources that were deleted during the resync.
-		updates := make([]api.Update, 0)
+		updates := make([]api.Update, 0, len(wc.resources))
 		for k, r := range wc.resources {
 			if r.resyncEpoch == wc.resyncEpoch {
 				// This resource was validated during the resync.
--- a/libcalico-go/lib/backend/watchersyncer/watchercache.go
+++ b/libcalico-go/lib/backend/watchersyncer/watchercache.go
@@ -214,6 +214,10 @@ func (wc *watcherCache) OnUpdate(kvp *model.KVPair) {
 
 // OnDelete handles delete events from ListAndWatch
 func (wc *watcherCache) OnDelete(kvp *model.KVPair) {
+	if kvp == nil {
+		wc.logger.Warn("Received delete event without KVPair, skipping")
+		return
+	}
 	kvp.Value = nil
 	wc.handleWatchListEvent(kvp)
 }
--- a/libcalico-go/lib/backend/watchersyncer/watchercache.go
+++ b/libcalico-go/lib/backend/watchersyncer/watchercache.go
@@ -49,7 +49,7 @@ type watcherCache struct {
 	listWatcherOptions     api.ListWatcherOptions
 }
 
-// cacheEntry is an entry in our cache.  It groups the a key with the last known
+// cacheEntry is an entry in our cache.  It groups a key with the last known
 // revision that we processed.  We store the revision so that we can determine
 // if an entry has been updated (and therefore whether we need to send an update
 // event in the syncer callback).
--- a/libcalico-go/lib/backend/watchersyncer/watchercache.go
+++ b/libcalico-go/lib/backend/watchersyncer/watchercache.go
@@ -46,6 +46,7 @@ type watcherCache struct {
 	results                chan<- any
 	hasSynced              bool
 	resourceType           ResourceType
+	listWatcherOptions     api.ListWatcherOptions
 }
 
 // cacheEntry is an entry in our cache.  It groups the a key with the last known
@@ -59,13 +60,14 @@ type cacheEntry struct {
 }
 
 // Create a new watcherCache.
-func newWatcherCache(client api.Client, resourceType ResourceType, results chan<- any) *watcherCache {
+func newWatcherCache(client api.Client, resourceType ResourceType, results chan<- any, options api.ListWatcherOptions) *watcherCache {
 	return &watcherCache{
-		logger:       logrus.WithField("ListRoot", listRootForLog(resourceType.ListInterface)),
-		client:       client,
-		resourceType: resourceType,
-		results:      results,
-		resources:    make(map[string]cacheEntry),
+		logger:             logrus.WithField("ListRoot", listRootForLog(resourceType.ListInterface)),
+		client:             client,
+		resourceType:       resourceType,
+		results:            results,
+		resources:          make(map[string]cacheEntry),
+		listWatcherOptions: options,
 	}
 }
 
@@ -96,10 +98,10 @@ func (wc *watcherCache) run(ctx context.Context) {
 
 func (wc *watcherCache) listAndWatch(ctx context.Context) error {
 	if client, ok := wc.client.(api.ListAndWatchClient); ok {
-		return client.ListAndWatch(ctx, wc.resourceType.ListInterface, wc)
+		return client.ListAndWatch(ctx, wc.resourceType.ListInterface, wc, api.WithListWatcherOptions(wc.listWatcherOptions))
 	}
 
-	lw := api.NewGenericListWatcher(wc.resourceType.ListInterface, wc)
+	lw := api.NewGenericListWatcher(wc.resourceType.ListInterface, wc, api.WithListWatcherOptions(wc.listWatcherOptions))
 	backend := &clientListWatchBackend{
 		client:      wc.client,
 		list:        wc.resourceType.ListInterface,
--- a/libcalico-go/lib/backend/watchersyncer/watchercache.go
+++ b/libcalico-go/lib/backend/watchersyncer/watchercache.go
@@ -129,12 +129,27 @@ func (b *clientListWatchBackend) HandleWatchEvent(event api.WatchEvent) error {
 	if b.listWatcher.HandleBasicWatchEvent(event) {
 		return nil
 	}
-	if event.Type == api.WatchError {
+
+	switch event.Type {
+	case api.WatchBookmark:
+		b.handleBookmark(event)
+		return nil
+	case api.WatchError:
 		b.HandleWatchError(event.Error)
 		return event.Error
+	default:
+		b.listWatcher.Logger.WithField("eventType", event.Type).Warn("Unexpected watch event type")
+		return nil
 	}
-	b.listWatcher.Logger.WithField("eventType", event.Type).Warn("Unexpected watch event type")
-	return nil
+}
+
+func (b *clientListWatchBackend) handleBookmark(event api.WatchEvent) {
+	if event.New == nil {
+		b.listWatcher.Logger.Warn("Watch bookmark received without revision")
+		return
+	}
+	b.listWatcher.UpdateRevision(event.New.Revision)
+	b.listWatcher.MarkSuccessfulConnection()
 }
 
 func (b *clientListWatchBackend) HandleListError(err error) {
--- a/libcalico-go/lib/backend/watchersyncer/watchercache.go
+++ b/libcalico-go/lib/backend/watchersyncer/watchercache.go
@@ -91,6 +91,7 @@ func (wc *watcherCache) run(ctx context.Context) {
 			return
 		}
 		wc.logger.WithError(err).Error("ListAndWatch failed")
+		wc.results <- errorSyncBackendError{Err: err}
 	}
 }
 
--- a/libcalico-go/lib/backend/watchersyncer/watchercache.go
+++ b/libcalico-go/lib/backend/watchersyncer/watchercache.go
@@ -83,9 +83,8 @@ func (wc *watcherCache) run(ctx context.Context) {
 	// On shutdown, send deletions for all the objects we're tracking.
 	defer wc.sendDeletionsForAllResources()
 
-	// Run ListAndWatch - this will block until context is cancelled or an error occurs
-	// watcherCache directly implements the EventHandler interface
-	if err := wc.client.ListAndWatch(ctx, wc.resourceType.ListInterface, wc); err != nil {
+	// Run ListAndWatch - this will block until context is cancelled or an error occurs.
+	if err := wc.listAndWatch(ctx); err != nil {
 		if ctx.Err() != nil || errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 			wc.logger.WithError(err).Debug("ListAndWatch stopped")
 			return
@@ -93,6 +92,64 @@ func (wc *watcherCache) run(ctx context.Context) {
 		wc.logger.WithError(err).Error("ListAndWatch failed")
 		wc.results <- errorSyncBackendError{Err: err}
 	}
+}
+
+func (wc *watcherCache) listAndWatch(ctx context.Context) error {
+	if client, ok := wc.client.(api.ListAndWatchClient); ok {
+		return client.ListAndWatch(ctx, wc.resourceType.ListInterface, wc)
+	}
+
+	lw := api.NewGenericListWatcher(wc.resourceType.ListInterface, wc)
+	backend := &clientListWatchBackend{
+		client:      wc.client,
+		list:        wc.resourceType.ListInterface,
+		listWatcher: lw,
+	}
+	return lw.ListAndWatchWithBackend(ctx, backend)
+}
+
+type clientListWatchBackend struct {
+	client      api.Client
+	list        model.ListInterface
+	listWatcher *api.GenericListWatcher
+}
+
+func (b *clientListWatchBackend) PerformList(ctx context.Context) (*model.KVPairList, error) {
+	return b.client.List(ctx, b.list, "")
+}
+
+func (b *clientListWatchBackend) CreateWatch(ctx context.Context, useWatchList bool) (api.WatchInterface, error) {
+	return b.client.Watch(ctx, b.list, api.WatchOptions{
+		AllowWatchBookmarks: true,
+		Revision:            b.listWatcher.CurrentRevision,
+	})
+}
+
+func (b *clientListWatchBackend) HandleWatchEvent(event api.WatchEvent) error {
+	if b.listWatcher.HandleBasicWatchEvent(event) {
+		return nil
+	}
+	if event.Type == api.WatchError {
+		b.HandleWatchError(event.Error)
+		return event.Error
+	}
+	b.listWatcher.Logger.WithField("eventType", event.Type).Warn("Unexpected watch event type")
+	return nil
+}
+
+func (b *clientListWatchBackend) HandleListError(err error) {
+	b.listWatcher.HandleListError(err)
+}
+
+func (b *clientListWatchBackend) HandleWatchError(err error) {
+	if b.listWatcher.HandleWatchUnavailableError(err) {
+		return
+	}
+	b.listWatcher.HandleWatchError(err)
+}
+
+func (b *clientListWatchBackend) PerformInitialSync(ctx context.Context, g *api.GenericListWatcher, useWatchList bool) error {
+	return g.PerformListSync(ctx, b)
 }
 
 // OnResyncStarted implements api.EventHandler interface.
@@ -197,9 +254,8 @@ func (wc *watcherCache) handleWatchListEvent(kvp *model.KVPair) {
 	}
 }
 
-// handleWatchBookmark handles a bookmark event from the API server, these
-// update the revision that we should be watching from without sending
-// a KVP.  This prevents datastore compactions from invalidating our watches.
+// handleConvertedWatchEvent handles one converted watch event and sends the
+// appropriate update type to the main watcherSyncer.
 func (wc *watcherCache) handleConvertedWatchEvent(kvp *model.KVPair) {
 	if kvp.Value == nil {
 		wc.handleDeletedUpdate(kvp.Key)
--- a/libcalico-go/lib/backend/watchersyncer/watchercache.go
+++ b/libcalico-go/lib/backend/watchersyncer/watchercache.go
@@ -18,16 +18,15 @@ import (
 	"context"
 	"errors"
 	"strings"
-	"time"
 
 	"github.com/sirupsen/logrus"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	utilnet "k8s.io/apimachinery/pkg/util/net"
 
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/model"
-	cerrors "github.com/projectcalico/calico/libcalico-go/lib/errors"
 )
+
+// Compile-time check that watcherCache implements api.EventHandler interface
+var _ api.EventHandler = (*watcherCache)(nil)
 
 // The watcherCache provides watcher/syncer support for a single key type in the
 // backend.  These results are sent to the main WatcherSyncer on a buffered "results"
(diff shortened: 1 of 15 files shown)


libcalico-go/lib/backend/watchersyncer/watchercache.go
```diff
@@ -231,38 +275,44 @@ func (wc *watcherCache) maybeResyncAndCreateWatcher(ctx context.Context) {
...
-					wc.logger.Info("Resource too old/new error from server, clearing cached watch revision.")
-					wc.resetWatchRevisionForFullResync()
-					// Error is a "layer 7" error; so connection is good!
-					wc.lastSuccessfulConnTime = time.Now()
-				} else if time.Since(wc.lastSuccessfulConnTime) > wc.watchRetryTimeout {
-					// Need to send back an error here for handling. Only callbacks with connection failure handling should actually kick off anything.
-					wc.logger.Warn("Connection to datastore has failed - signaling error to client.")
-					wc.results <- errorSyncBackendError{
-						Err: err,
+			l, err := wc.watchList(ctx)
```
liuxu623, line 278: maybe we should add a config to control whether use WatchList

fasaxc:
D'oh, think we're duplicating some work here; I've been working on the same. #9593 was an enabling PR to be able to add this feature.  You're right; it's a big win for us.  It actually moves the k8s API to work the same way that Calico's internal "Syncer" API works(!)

liuxu623:
> D'oh, think we're duplicating some work here; I've been working on the same. #9593 was an enabling PR to be able to add this feature. You're right; it's a big win for us. It actually moves the k8s API to work the same way that Calico's internal "Syncer" API works(!)
> 
> I need to dust off my work in progress and see if there were any gotchas and compare with your approach.

I noticed #9593 before, WatchList feature depends on bookmark. After support bookmark, a lot of work was reduced.

liuxu623:
@fasaxc Felix add `K8S_USE_WATCH_LIST ` env to controls whether use of WatchList, PTAL.

fasaxc:
Sorry for the radio silence, I finally got a chance to have a look.  A few high level thoughts rather than jumping into line-by line review:

- Overall, this looks like it'd work.

- Can we detect specifically that the server doesn't support the watch-list operation and fall back only in that case?  I think k8s reflector/informer does manage to do this somehow; I think the error we expect from an old API server is distinct.

- I think introducing the `func (wc *watcherCache) watchList(ctx context.Context)` method might be too cautious.  We shouldn't need to turn the first part of the watch into a simulated "list" call because the output of the watchersyncer uses Calico's "Syncer" API, which is actually almost identical to the "send initial events" approach that k8s has now adopted(!). 

  - Record that we're resyncing
  - Start a "send initial events" watch.
  - Handle each event as normal, send it downstream straight away (downstream can handle duplicates).
  - When we get the first bookmark:
     - Use the cache of previously seen resources to figure out any deletions and send those downstream.
     - If this is the first time we've been in sync, send an "in-sync" message downstream.  This is Calico's equivalent to the first bookmark"; it tells the downstream "ok you've got everything now".

- For config, I'd prefer to go through felix/typha's `config_params.go`, which is how those components are generally configured.  This code normally runs in Typha in a large cluster, offloading it from Felix.  That said, there isn't a great way to configure Typha in most deployments because the operator locks down its env vars.

- We need to test this thoroughly.  Bugs in this area tend to be catastrophic.  It'd be good to:

  - Run the test suite both with and without the feature enabled.
  - If we can, test trying to use the feature when it's disabled at the API server so we get an error and test the fallback code.
  - See if we can figure out a way to test corner cases like getting disconnected mid-resync.

- Calico still supports the etcdv3 backend, I think you need to make sure that the etcdv3 backend returns a "not supported" error if passed the new watch options that it doesn't support.

For reference, here's a gist with my proof-of-concept version: https://gist.github.com/fasaxc/79da02a2ab91c3bcd9644761e6abfc79  It was based on a much earlier version of the watch bookmarks PR and it's hard to rebase now but there might be something worth borrowing from it.  Some things I did differently:

- Split the initial resync into two versions, one using list and one not.  I was hoping that the watch list version would be a lot simpler due to being closer to Calico's model.  Looking at it now, I'd move the for loop up into the calling function since it's the same in both.  I added a bunch of documentation while I was trying to understand this code thoroughly again.

- To do the "mark and sweep" GC of deleted keys after a resync, I introduced an epoch counter that was incremented once per resync.  This made it easy to scan the resources map and find all the keys that should be deleted.  If I remember correctly, the `oldResources` map got very fiddly when doing watch-based resyncs due to losing the atomicity of doing a full list.  You've preserved the atomicity by simulating the list operation, but that means that you have to buffer the whole output in memory rather than stream it.

- I introduced a method `wc.client.SupportedWatchOptions()` to let the watchercache ask the client if it supported the feature.  This was how I planned to detect etcd not supporting the feature.  Not sure this panned out, better to just try it and have the backend respond with a "not supported" error, since there's no easy way to detect if k8s supports it without trying.

- I was experimenting with the options pattern on `wc.client.Watch(... WithBookmarks())`, I think this was a dead end so I went with an options struct in the watch bookmarks PR.

liuxu623:
> * Can we detect specifically that the server doesn't support the watch-list operation and fall back only in that case?  I think k8s reflector/informer does manage to do this somehow; I think the error we expect from an old API server is distinct.

Before k8s 1.27, watch not support resourceVersionMatch, ValidateListOptions return err.
https://github.com/kubernetes/kubernetes/blob/release-1.26/staging/src/k8s.io/apimachinery/pkg/apis/meta/internalversion/validation/validation.go#L30
```go
// ValidateListOptions returns all validation errors found while validating the ListOptions.
func ValidateListOptions(options *internalversion.ListOptions) field.ErrorList {
	allErrs := field.ErrorList{}
	if match := options.ResourceVersionMatch; len(match) > 0 {
		if options.Watch {
			allErrs = append(allErrs, field.Forbidden(field.NewPath("resourceVersionMatch"), "resourceVersionMatch is forbidden for watch"))
		}
                 ......
	}
	return allErrs
}
```
After k8s 1.27, sendInitialEvents not support if WatchList feature gate is disabled, ValidateListOptions return err.
https://github.com/kubernetes/kubernetes/blob/release-1.27/staging/src/k8s.io/apimachinery/pkg/apis/meta/internalversion/validation/validation.go#L61
```go
func validateWatchOptions(options *internalversion.ListOptions, isWatchListFeatureEnabled bool) field.ErrorList {
	allErrs := field.ErrorList{}
	match := options.ResourceVersionMatch
	if options.SendInitialEvents != nil {
		if match != metav1.ResourceVersionMatchNotOlderThan {
			allErrs = append(allErrs, field.Forbidden(field.NewPath("resourceVersionMatch"), fmt.Sprintf("sendInitialEvents requires setting resourceVersionMatch to %s", metav1.ResourceVersionMatchNotOlderThan)))
		}
		if !isWatchListFeatureEnabled {
			allErrs = append(allErrs, field.Forbidden(field.NewPath("sendInitialEvents"), "sendInitialEvents is forbidden for watch unless the WatchList feature gate is enabled"))
		}
	}
	......
	return allErrs
}
```
So, the error is same for new and old apiserver, and the error is Invalid error, we can use the error type to determine whether the apiserver supports WatchList.
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/get.go#L202
```go
		metainternalversion.SetListOptionsDefaults(&opts, utilfeature.DefaultFeatureGate.Enabled(features.WatchList))
		if errs := metainternalversionvalidation.ValidateListOptions(&opts, utilfeature.DefaultFeatureGate.Enabled(features.WatchList)); len(errs) > 0 {
			err := errors.NewInvalid(schema.GroupKind{Group: metav1.GroupName, Kind: "ListOptions"}, "", errs)
			scope.err(err, w, req)
			return
		}
```

liuxu623:
> * Calico still supports the etcdv3 backend, I think you need to make sure that the etcdv3 backend returns a "not supported" error if passed the new watch options that it doesn't support.

Yes, I plan to make etcd watch return the same error as apiserver if specifies ResourceVersionMatch or SendInitialEvents.
```go
// Watch entries in the datastore matching the resources specified by the ListInterface.
func (c *etcdV3Client) Watch(cxt context.Context, l model.ListInterface, options api.WatchOptions) (api.WatchInterface, error) {
	allErrs := field.ErrorList{}
	if len(options.ResourceVersionMatch) > 0 {
		allErrs = append(allErrs, field.Forbidden(field.NewPath("resourceVersionMatch"), "resourceVersionMatch is forbidden for etcdv3 backend"))
	}
	if options.SendInitialEvents != nil && *options.SendInitialEvents {
		allErrs = append(allErrs, field.Forbidden(field.NewPath("sendInitialEvents"), "sendInitialEvents is forbidden for etcdv3 backend"))
	}
	if len(allErrs) > 0 {
		return nil, apierrors.NewInvalid(schema.GroupKind{Group: metav1.GroupName, Kind: "ListOptions"}, "", allErrs)
	}
```

liuxu623:
https://github.com/projectcalico/calico/blob/v3.30.0/libcalico-go/lib/backend/etcdv3/watcher.go#L93-L115
```go
if wc.initialRev == 0 {
	// No initial revision supplied, so perform a list of current configuration
	// which will also get the current revision we will start our watch from.
	var kvps *model.KVPairList
	var err error
	if kvps, err = wc.listCurrent(); err != nil {
		log.Errorf("failed to list current with latest state: %v", err)
		// Error considered as terminating error, hence terminate watcher.
		wc.sendError(err)
		return
	}

	// If we're handling profiles, filter out the default-allow profile.
	if len(kvps.KVPairs) > 0 && (key == profilesKey || key == defaultAllowProfileKey) {
		wc.removeDefaultAllowProfile(kvps)
	}

	// We are sending an initial sync of entries to the watcher to provide current
	// state.  To the perspective of the watcher, these are added entries, so set the
	// event type to WatchAdded.
	log.WithField("NumEntries", len(kvps.KVPairs)).Debug("Sending create events for each existing entry")
	wc.sendAddedEvents(kvps)
}
```
But I found etcdv3 watcher will list first if `initialRev == 0`, maybe we needn't return a "not supported" error for etcdv3 backend.

liuxu623:
```
2025-05-08 06:17:46.395 [WARNING][105] felix/watchercache.go 157: API server not support WatchList, falling back to List ListRoot=".../v3/pc.org/workloadendpoints" error=ListOptions.meta.k8s.io "" is invalid: sendInitialEvents: Forbidden: sendInitialEvents is forbidden for watch unless the WatchList feature gate is enabled
```
I have test apiserver disable WatchList featuregate, fallback to list works normal.

liuxu623:
https://kubernetes.io/blog/2025/05/09/kubernetes-v1-33-streaming-list-responses/

```
kube-apiserver disables the beta WatchList feature by default in 1.33 in favor of the StreamingCollectionEncodingToJSON and StreamingCollectionEncodingToProtobuf features.kube-controller-manager no longer opts into enabling the WatchListClient feature in 1.33. (https://github.com/kubernetes/kubernetes/pull/131359, [@deads2k](https://github.com/deads2k)) [SIG API Machinery]
```

@fasaxc WatchList feature has disabled by default in 1.33, because StreamingCollectionEncodingToJSON and StreamingCollectionEncodingToProtobuf appear to work better, maybe we doesn't need WatchList anymore.

liuxu623:
> Agree, let's put it on hold. It's a shame; the streaming decode is better for Calico on the client side. We avoid a big resource spike while we parse/validate the very large List object.

@fasaxc  Kubernetes 1.34 has re-promoted the WatchList feature to Beta, and we can resume this work now.​​

mazdakn:
@fasaxc @liuxu623 fyi, we bumped k8s version to `1.34.1` recently: https://github.com/projectcalico/calico/pull/11200
Just wanted to point it out here as it might have unblocked this PR.

liuxu623:
@fasaxc I apologize for the lack of updates to this PR recently due to work commitments. I've made some refactorings to the ListWatch code; details can be found in the PR description. Looking forward to your feedback.

liuxu623:
@fasaxc @caseydavenport @danudey  PTAL

fasaxc:
Thanks @liuxu623 , this has grown into a large refactor, it might take me a while to process!  Also note that @pasanw has a couple of changes in flight (upstreaming a fix from our enterprise codebase).  We'll want to make sure that doesn't get lost.

PR #10209, opened 2025-04-13
ebpf - Fix configuring arp and rp_filter for bpf nat devices

sridhartigera:
## Description

In version 242 systemd started setting a persistent MAC address on virtual interfaces. Therefore bpfin.cali and bpfout.cali interfaces MAC changes after creation. As a result, we get a wrong mac when reading it for the first time. This results in a wrong ARP entry programmed. Also, rp_filter setting changes. 

Fix - Configure the dataplane entries for bpfin.cali and bpfout.cali when bpf_ep_mgr gets notified that the interfaces are OPER_UP.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf - Fix configuring arp entries for bpf NAT devices for systemd >= 242
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -168,6 +168,7 @@ type bpfDataplane interface {
 	ensureNoProgram(attachPoint) error
 	ensureQdisc(iface string) (bool, error)
 	ensureBPFDevices() error
+	configureBPFDevices() error
 	updatePolicyProgram(rules polprog.Rules, polDir string, ap attachPoint, ipFamily proto.IPVersion) error
 	removePolicyProgram(ap attachPoint, ipFamily proto.IPVersion) error
 	setAcceptLocal(iface string, val bool) error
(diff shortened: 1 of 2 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1117,11 +1118,18 @@ func (m *bpfEndpointManager) onInterfaceUpdate(update *ifaceStateUpdate) {
 					delete(m.initAttaches, update.Name)
 				}
 			}
+			if update.Name == dataplanedefs.BPFInDev {
+				m.ifacesLock.Lock()
+				if err := m.reconcileBPFDevices(dataplanedefs.BPFInDev); err != nil {
+					log.WithError(err).Fatal("Failed to configure BPF devices")
+				}
+				defer m.ifacesLock.Unlock()
```
tomastigera, line 1126: either use defer right after Lock or just Unlock here. The fatal above won't return, it will kill the process. Why don't you handle it below where you handle  Out?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1182,8 +1190,10 @@ func (m *bpfEndpointManager) onInterfaceUpdate(update *ifaceStateUpdate) {
 			// We require host interfaces to be in non-strict RPF mode so that
 			// packets can return straight to host for services bypassing CTLB.
 			switch update.Name {
-			case dataplanedefs.BPFInDev, dataplanedefs.BPFOutDev:
-				// do nothing
+			case dataplanedefs.BPFOutDev:
```
tomastigera, line 1193: what happes to "do nothing" with dataplanedefs.BPFInDev ? Aren't you already handling this case above?

tomastigera (approved):

PR #10223, opened 2025-04-15
[BPF] fix pkt/bytes accounting for VXLAN

tomastigera:
## Description

Fixes an omission introduced in fix https://github.com/tigera/felix-private/pull/1032. CALI_F_TUNNEL represented IPIP only and VXLAN was never distinguished in any way as it was treated as a regular L2 device. That is fixed by the following two fixes originally introduced for https://github.com/projectcalico/calico/pull/9965

    [BPF] compile distinct programs for VXLAN
    
    We need to know that we are forwarding from and L2 overlay.

    [BPF] rename references to tunnel to ipip
    
    IT is a historical artefact and we should name it properly and leave the
    name tunnel for a generic overlay.

Also includes

    [BPF] flowlogs - we now know which side initiated the conn

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


sridhartigera (approved):
LGTM

PR #10246, opened 2025-04-16
[CORE-11251] Openshift: move CR order to after operator has applied CRDs

coutinhop:
Move custom resource yamls from 01-cr-*.yaml to 03-cr-*.yaml, so that they are applied after the operator has been created (02-*.yaml), so that the operator is able to create the CRDs needed for the CRs.

Tested on openshift 4.16 on AWS, and HCP cluster (HCP has some docs changes pending to correctly apply all yamls and wait for CRDs to be created).

## Todos

- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Change OpenShift manifests order to fix an error when creating HCP clusters.
```

## Reminder for the reviewer

Other optional labels:


manifests/ocp/03-cr-apiserver.yaml
```diff
@@ -0,0 +1,7 @@
+# This section configures the Calico API server.
+# For more information, see: https://docs.tigera.io/calico/latest/reference/installation/api#operator.tigera.io/v1.APIServer
```
lucastigera, line 2: This comment was modified. Was that intentional or by mistake?

manifests/ocp/03-cr-installation.yaml
lucastigera, line 1: **Question:**
Doesn't the operator depend on the Installation?
Wouldn't the information below be necessary for the operator to determine things like which resources to install?

```
  calicoNetwork:
    linuxDataplane: BPF
  variant: Calico
```

manifests/generate.sh
```diff
@@ -135,19 +135,21 @@ ${HELM} template \
...
 	--set calicoctl.tag=$CALICO_VERSION
 # The first two lines are a newline and a yaml separator - remove them.
-find ocp/tigera-operator -name "*.yaml" | xargs sed -i -e 1,2d
+find ocp/tigera-operator -name "*.yaml" -print0 | xargs -0 sed -i -e 1,2d
 mv $(find ocp/tigera-operator -name "*.yaml") ocp/ && rm -r ocp/tigera-operator
 
 # Generating the upgrade manifest for OCP.
-# It excludes the CRs (01-*) and the specific BPF files to maintain compatibility with iptables.
-VALUES_FILES=$(ls ocp | grep -v -e '^01-' -e 'cluster-network-operator.yaml' -e '02-configmap-calico-resources.yaml')
+# It excludes files specific to using BPF, CRs (03-cr-*) and CRDs to maintain compatibility with iptables.
```
caseydavenport, line 142: I'm not really sure what "maintain compatibility with iptables" means here - I thought we excluded these files because they are install-time configuration that shouldn't be re-applied?

manifests/generate.sh
```diff
@@ -135,19 +135,21 @@ ${HELM} template \
...
-for FILE in $VALUES_FILES; do
+for FILE in $OCP_VALUES_FILES; do
   cat "ocp/$FILE" >> tigera-operator-ocp-upgrade.yaml
-  echo -e "---" >> tigera-operator-ocp-upgrade.yaml  # Add divisor
+  echo -e "---" >> tigera-operator-ocp-upgrade.yaml  # Add separator
 done
-# Remove the last separator (last line)
-sed -i -e '$ d' tigera-operator-ocp-upgrade.yaml
+# Remove the last separator
+sed -i -e '${/^---$/d}' tigera-operator-ocp-upgrade.yaml
```
caseydavenport, line 150: Is this just cosmetic? or actually important?

manifests/ocp/03-cr-apiserver.yaml
coutinhop: yes, it's intentional, it looked like the file was copied straight from the templates, the URL wasn't valid with unsubstituted values

manifests/ocp/03-cr-installation.yaml
coutinhop: no, the operator doesn't depend on the installation, the installation is what tells the operator to install Calico (and how), the operator will be installed on the `02-*.yaml` step and remain "idle", then when the installation is created in the `03-*.yaml` step, it will install Calico on the cluster

manifests/generate.sh
coutinhop: to be 100% honest, I didn't touch this part of the comment, but I definitely should. will update the reasoning there

manifests/generate.sh
coutinhop: again, I was just trying to make the `sed` expression a little more robust, but yes I do believe it's not invalid yaml, so yes cosmetic. should we leave the last `---` separator?

lucastigera (approved):

manifests/generate.sh
coutinhop: It was cosmetic, I tested. Removed this to keep the script simpler

manifests/generate.sh
coutinhop: I updated the comment, please take a look

caseydavenport (approved):

PR #10248, opened 2025-04-17
skip deleting ippool

liuxu623:
## Description

We shouldn't use a IPPool if it's deletionTimestamp is not zero.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/clientv3/client.go
+++ b/libcalico-go/lib/clientv3/client.go
@@ -201,6 +201,10 @@ type poolAccessor struct {
 
 func (p poolAccessor) GetEnabledPools(ctx context.Context, ipVersion int) ([]v3.IPPool, error) {
 	return p.getPools(ctx, func(pool *v3.IPPool) bool {
+		if !pool.DeletionTimestamp.IsZero() {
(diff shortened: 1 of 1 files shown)


caseydavenport:
@liuxu623 nice, this seems like a good enhancement although I believe the scenario this protects against is meant to be covered by deletion logic already via this logic: https://github.com/projectcalico/calico/blob/522e2d5b3ab1d028a5531ebcfe475bbb6d7e24cb/libcalico-go/lib/clientv3/ippool.go#L169-L187

Either way, I think what you've done here is better since it protects against anyone sneaking around our validation logic.

caseydavenport:
I think a unit test to verify this would be a nice addition, if you have some time.

liuxu623:
> I think a unit test to verify this would be a nice addition, if you have some time.

ok I will add unit test later.

liuxu623:
https://github.com/projectcalico/calico/pull/10260

PR #10264, opened 2025-04-18
Standardize buildinfo package for simplicity

caseydavenport:
## Description

We use the same or very similar LDFLAGS across all of our components,
each redefining a buildinfo package.

This PR centralizes this all into a single location.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


pkg/buildinfo/buildinfo.go
```diff
@@ -12,26 +12,25 @@
...
-var BuildVersion string
+// Filled in by build process.
+var (
+	// Version is the version of the code.
+	Version string
 
-// BuildDate stores the date of the build
-var BuildDate string
+	// GitRevision is the commit hash of the code.
+	GitRevision string
```
fasaxc, line 25: I think the go toolchain embeds this into the binary automatically now?

https://pkg.go.dev/runtime/debug#BuildInfo

Could we grab some of this information from there and remove some of the make machinery?

caseydavenport: @fasaxc neat! I didn't know about that.

Will need to dig a bit - trivial swap doesn't seem to "just work" - it has empty build information for the `main` module. Potentially something with how we are building.

PR #10265, opened 2025-04-18
Remove redundant license files

caseydavenport:
## Description

These are all remnants from when we had separate repositories for this
code - they are all covered by the root LICENSE.md file now.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


coutinhop (approved):

hjiawei:
We also need to remove the license copy from Dockerfiles (for example, https://github.com/projectcalico/calico/blob/master/apiserver/Dockerfile#L22).

hjiawei:
Maybe also check with @lwr20. IIRC, OpenShift certification requires a LICENSE file in the container?

fasaxc:
I think we need to include the licenses in the images but that doesn't mean that we need dozens of copies.

caseydavenport:
> I think we need to include the licenses in the images but that doesn't mean that we need dozens of copies.

Agreed. Unfortunatley our repo structure makes this annoying to do since we need to pass the LICENSE file to the Docker build context. But we can make temporary copies as part of the docker image build.

Our plan to reorganize the repo package structure should make this a lot easier, when we get around to it.

caseydavenport:
@coutinhop @hjiawei @fasaxc OK I think this one is ready for another look - I ended up using the root LICENSE and copying it into the build context of each container.

As I move towards a broader repository refactor, I am hoping this will get simpler (image builds can be done from the root so we won't need to copy things around).

coutinhop (approved):

PR #10267, opened 2025-04-21
list ippools with label selector

gojoy:
## Description
Listing ippools with label selector when assigning ip in the IPAM. When there are a large number of ippools in the k8s cluster, this optimization can lead to a performance improvement.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/pkg/ipamplugin/ipam_plugin.go
+++ b/cni-plugin/pkg/ipamplugin/ipam_plugin.go
@@ -243,15 +243,16 @@ func cmdAdd(args *skel.CmdArgs) error {
 			maxBlocks = 1
 		}
 		assignArgs := ipam.AutoAssignArgs{
-			Num4:             num4,
-			Num6:             num6,
-			HandleID:         &handleID,
-			Hostname:         nodename,
-			IPv4Pools:        v4pools,
-			IPv6Pools:        v6pools,
-			MaxBlocksPerHost: maxBlocks,
-			Attrs:            attrs,
-			IntendedUse:      v3.IPPoolAllowedUseWorkload,
+			Num4:                     num4,
+			Num6:                     num6,
+			HandleID:                 &handleID,
+			Hostname:                 nodename,
+			IPv4Pools:                v4pools,
+			IPv6Pools:                v6pools,
+			MaxBlocksPerHost:         maxBlocks,
+			Attrs:                    attrs,
+			IntendedUse:              v3.IPPoolAllowedUseWorkload,
+			IpPoolsLabelSelectorKeys: conf.IPAM.IpPoolsLabelSelectorKeys,
 		}
 		if runtime.GOOS == "windows" {
 			rsvdAttrWindows := &ipam.HostReservedAttr{
--- a/cni-plugin/pkg/types/types.go
+++ b/cni-plugin/pkg/types/types.go
@@ -70,13 +70,14 @@ type NetConf struct {
 	VXLANMacPrefix string `json:"vxlan_mac_prefix"`
 	VXLANVNI       uint64 `json:"vxlan_vni"`
 	IPAM           struct {
-		Name       string
-		Type       string   `json:"type"`
-		Subnet     string   `json:"subnet"`
-		AssignIpv4 *string  `json:"assign_ipv4"`
-		AssignIpv6 *string  `json:"assign_ipv6"`
-		IPv4Pools  []string `json:"ipv4_pools,omitempty"`
-		IPv6Pools  []string `json:"ipv6_pools,omitempty"`
+		Name                     string
+		Type                     string   `json:"type"`
+		Subnet                   string   `json:"subnet"`
+		AssignIpv4               *string  `json:"assign_ipv4"`
+		AssignIpv6               *string  `json:"assign_ipv6"`
+		IPv4Pools                []string `json:"ipv4_pools,omitempty"`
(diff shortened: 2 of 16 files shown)


caseydavenport:
@gojoy thanks for the PR! I just want to make sure I understand this correctly - the idea is that instead of specifying exact IP pool names in the CNI config,you'd be able to specify a label selector that selects IP pools to use for allocations on that node?

I'm curious if you've actually had problems with the current approach, and if so what level of scale you've seen them at and what the problems were?

One thing I'm a bit wary of is the interaction between this and the `NodeSelector` field on IP pools - it can get a bit confusing if Nodes are selecting IP pools, and IP pools are in turn also selecting nodes.

gojoy:
> @gojoy thanks for the PR! I just want to make sure I understand this correctly - the idea is that instead of specifying exact IP pool names in the CNI config,you'd be able to specify a label selector that selects IP pools to use for allocations on that node?
> 
> I'm curious if you've actually had problems with the current approach, and if so what level of scale you've seen them at and what the problems were?
> 
> One thing I'm a bit wary of is the interaction between this and the `NodeSelector` field on IP pools - it can get a bit confusing if Nodes are selecting IP pools, and IP pools are in turn also selecting nodes.

There are thousands of IP pool resources in our production Kubernetes cluster. During each IP allocation process by the CNI plugin,all IP pool resources need to be retrieved from the API server, which places pressure on the cluster's control plane,It also leads to a long time consumption in the IP allocation process（there will be multiple paginated query requests with a limit of 500.）. It is recommended to utilize label selectors to filter the IP pools in advance on the API server side, thereby reducing the overhead associated with listing operations.

caseydavenport:
@gojoy got it, and to make sure I understand the UX you're proposing:

- Apply the same label to a set of IP pools and a set of nodes - e.g., `pool-group: group-1`
- Tell the CNI plugin to use label key `pool-group`
- The IPAM code will then use any IP pools that have the same value for `pool-group` as the Node performing the allocation.

Is that right?

gojoy:
> @gojoy got it, and to make sure I understand the UX you're proposing:
> 
> * Apply the same label to a set of IP pools and a set of nodes - e.g., `pool-group: group-1`
> * Tell the CNI plugin to use label key `pool-group`
> * The IPAM code will then use any IP pools that have the same value for `pool-group` as the Node performing the allocation.
> 
> Is that right?

yes, that's right. This is the way we can think of to optimize the performance.

fasaxc:
@gojoy I'm curious, what is your reason for having thousands of pools?  Is it just fragmentation (you're using a lot of small pools to get many IPs) or something else?

gojoy:
> @gojoy I'm curious, what is your reason for having thousands of pools? If your cluster huge? Is it just fragmentation (you're using a lot of small pools to get many IPs) or something else?

Our cluster uses the container network mode of BGP+underlay. Therefore, each node has at least one ippools resource (usually a subnet with a mask of 26). When there are thousands of nodes in the cluster, the number of ippools resources will also be extremely large.

caseydavenport:
@gojoy normally for that I'd expect a single IP pool - Calico already splits IP pools into /26 blocks that it assigns to nodes by default, so you shouldn't need to use many IP pools unless you want precise control of which IP subnets are assigned to which nodes or you have a large discontinuous IP space for pods in the cluster.

gojoy:
> @gojoy normally for that I'd expect a single IP pool - Calico already splits IP pools into /26 blocks that it assigns to nodes by default, so you shouldn't need to use many IP pools unless you want precise control of which IP subnets are assigned to which nodes or you have a large discontinuous IP space for pods in the cluster.

Ideally, a single IP pool in the cluster would suffice. However,In our scenario, due to the existence of multiple clusters of different sizes and the dynamic nature of the cluster size, in order to use IP resources more precisely, we have to utilize multiple ippool resources.

caseydavenport:
>  in order to use IP resources more precisely, we have to utilize multiple ippool resources.

Just to make sure I understand - this means that the IP usage requirements of a given cluster cannot be represented efficiently with a single or small number of CIDRs?

gojoy:
> > in order to use IP resources more precisely, we have to utilize multiple ippool resources.
> 
> Just to make sure I understand - this means that the IP usage requirements of a given cluster cannot be represented efficiently with a single or small number of CIDRs?

I'll try to explain in detail why it's not feasible to use a single IP pool:

1.The current IP pool's CIDR is mapped to the Kubernetes cluster, but the number of nodes in the cluster can scale up or down. If we initially allocate a sufficiently large CIDR (e.g., with a /16 mask) for this cluster, a reduction in the number of nodes later would mean that most of the allocated IP addresses go unused, resulting in significant wastage.

2.In an underlay+BGP network model, different PODs/TORs within an IDC are typically assigned distinct CIDRs for container networking(which enables better route aggregation). However, nodes in a single Kubernetes cluster may be distributed across multiple PODs/TORs, making it impossible to use a contiguous CIDR as the cluster's IP pool.

I believe these are common considerations when managing multiple large-scale Kubernetes clusters. Thanks

caseydavenport:
Thanks for putting the effort into this one, but I'm going to close it for now.

I think the approach we want to take for improving IPAM scale is to switch to a deamon-based model, where the local Felix can serve requests for IPAM information much more efficiently than round-tips to the API server. I don't want to fragment the IP pool selection logic much further than it already is for now.

PR #10271, opened 2025-04-22
cni-plugin: merge calico-ipam into calico

liuxu623:
## Description

Merge calico-ipam into calico, use function calls instead of binary file calls to improve performance and code robustness.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/pkg/ipamplugin/ipam_plugin.go
+++ b/cni-plugin/pkg/ipamplugin/ipam_plugin.go
@@ -142,14 +142,6 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return fmt.Errorf("error constructing WorkloadEndpoint name: %s", err)
 	}
 
-	handleID := utils.GetHandleID(conf.Name, args.ContainerID, epIDs.WEPName)
-
-	logger := logrus.WithFields(logrus.Fields{
-		"Workload":    epIDs.WEPName,
-		"ContainerID": epIDs.ContainerID,
-		"HandleID":    handleID,
-	})
-
 	ipamArgs := ipamArgs{}
 	if err = cnitypes.LoadArgs(args.Args, &ipamArgs); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


caseydavenport:
Thanks for the PR @liuxu623 - just wanted to let you know I haven't missed this one, just haven't had time to take a look yet.

caseydavenport:
Reducing the number of API calls required to make an IP allocation is something I'd like to do as well! I think it's a fairly complicated task - unfortunately, our IPAM code does require a fair number of API calls as it is currently implemented. 

More broadly, I think the way to do this is similar to what you described above - switching to a (e.g., gRPC) daemon backed model, where the IPAM plugin can make requests to the Daemon and get IPs back. The Daemon would be using watches to cache data instead of explicit requests (ideally via Typha to avoid API server load) - but this is a really big change and requires a lot of care to ensure we handle edge cases and race conditions appropriately.

I'd be happy to talk through options for moving that direction - I don't think that removing the exec call to the IPAM plugin is going to move the needle here though, and I'd rather not merge changes without an impact - maybe let's start another issue where we can discuss what the future of Calico's IPAM implementation might look like?

liuxu623:
> Reducing the number of API calls required to make an IP allocation is something I'd like to do as well! I think it's a fairly complicated task - unfortunately, our IPAM code does require a fair number of API calls as it is currently implemented.
> 
> More broadly, I think the way to do this is similar to what you described above - switching to a (e.g., gRPC) daemon backed model, where the IPAM plugin can make requests to the Daemon and get IPs back. The Daemon would be using watches to cache data instead of explicit requests (ideally via Typha to avoid API server load) - but this is a really big change and requires a lot of care to ensure we handle edge cases and race conditions appropriately.
> 
> I'd be happy to talk through options for moving that direction - I don't think that removing the exec call to the IPAM plugin is going to move the needle here though, and I'd rather not merge changes without an impact - maybe let's start another issue where we can discuss what the future of Calico's IPAM implementation might look like?

ok, close this PR and discuss in another issue.

PR #10292, opened 2025-04-24
Add basic statistics FV test

caseydavenport:
## Description

Adds a basic FV test of the Statistics gRPC connection.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/goldmane/fv/daemon_test.go
+++ b/goldmane/fv/daemon_test.go
@@ -290,3 +290,72 @@ func TestHints(t *testing.T) {
 		return nil
 	}, 5*time.Second, 1*time.Second).Should(Succeed())
 }
+
+func TestStatistics(t *testing.T) {
+	cfg := daemon.Config{
+		LogLevel:                 "debug",
+		Port:                     8988,
+		AggregationWindow:        time.Second * 1,
+		EmitAfterSeconds:         2,
+		EmitterAggregationWindow: time.Second * 2,
+	}
+	defer daemonSetup(t, cfg)()
+
+	// Generate credentials for the Goldmane client.
+	creds, err := client.ClientCredentials(clientCert, clientKey, clientCA)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to create goldmane TLS credentials.")
+	}
+
+	// Generate credentials for the Goldmane client.
+	statsClient, err := client.NewStatisticsAPIClient(goldmaneURL, grpc.WithTransportCredentials(creds))
+	require.NoError(t, err)
+
+	// Send a request for statistics with no data.
+	Eventually(func() error {
+		res, err := statsClient.List(ctx, &proto.StatisticsRequest{})
+		if err != nil {
+			return err
+		}
+		if res != nil {
+			return fmt.Errorf("statistics returned")
+		}
+		return nil
+	}, 5*time.Second, 1*time.Second).Should(Succeed())
+
+	// Create some flow data.
+	// Create a client to pusher Flows.
+	pusher, err := client.NewFlowClient(goldmaneURL, clientCert, clientKey, clientCA)
+	require.NoError(t, err)
+
+	connected := pusher.Connect(ctx)
+	require.NoError(t, err)
+	Eventually(connected, 5*time.Second, 100*time.Millisecond).Should(BeClosed())
+
+	go func(ctx context.Context) {
+		for {
+			if ctx.Err() != nil {
+				pusher.Close()
+				return
+			}
+			f := testutils.NewRandomFlow(time.Now().Unix())
+			pusher.Push(types.ProtoToFlow(f))
+			time.Sleep(100 * time.Millisecond)
+		}
+	}(ctx)
+
+	// Now statistics should return data.
+	Eventually(func() error {
+		res, err := statsClient.List(ctx, &proto.StatisticsRequest{})
+		if err != nil {
+			return err
+		}
+		if res == nil {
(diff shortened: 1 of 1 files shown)


goldmane/fv/daemon_test.go
```diff
@@ -290,3 +290,72 @@ func TestHints(t *testing.T) {
...
+
+func TestStatistics(t *testing.T) {
+	cfg := daemon.Config{
+		LogLevel:                 "debug",
+		Port:                     8988,
+		AggregationWindow:        time.Second * 1,
+		EmitAfterSeconds:         2,
+		EmitterAggregationWindow: time.Second * 2,
+	}
+	defer daemonSetup(t, cfg)()
```
skoryk-oleksandr, line 302: Looks like `deamonSetup` function initialize global `ctx` field. Might be better to make `deamonSetup` return `ctx` and use local `ctx` variable. It will siplyfy runnig FVs in parallel, if we need it.

goldmane/fv/daemon_test.go
```diff
@@ -290,3 +290,72 @@ func TestHints(t *testing.T) {
...
+		return nil
+	}, 5*time.Second, 1*time.Second).Should(Succeed())
+
+	// Create some flow data.
+	// Create a client to pusher Flows.
+	pusher, err := client.NewFlowClient(goldmaneURL, clientCert, clientKey, clientCA)
+	require.NoError(t, err)
+
+	connected := pusher.Connect(ctx)
+	require.NoError(t, err)
```
skoryk-oleksandr, line 332: Looks like here we check `err` from the `client.NewFlowClient` function call but not from `pusher.Connect(ctx)`

goldmane/fv/daemon_test.go
```diff
@@ -290,3 +290,72 @@ func TestHints(t *testing.T) {
...
+	Eventually(connected, 5*time.Second, 100*time.Millisecond).Should(BeClosed())
+
+	go func(ctx context.Context) {
+		for {
+			if ctx.Err() != nil {
+				pusher.Close()
+				return
+			}
+			f := testutils.NewRandomFlow(time.Now().Unix())
+			pusher.Push(types.ProtoToFlow(f))
```
skoryk-oleksandr, line 342: Should we check error for the `pusher.Push(types.ProtoToFlow(f))` call ?

goldmane/fv/daemon_test.go
```diff
@@ -290,3 +290,72 @@ func TestHints(t *testing.T) {
...
+	statsClient, err := client.NewStatisticsAPIClient(goldmaneURL, grpc.WithTransportCredentials(creds))
+	require.NoError(t, err)
+
+	// Send a request for statistics with no data.
+	Eventually(func() error {
+		res, err := statsClient.List(ctx, &proto.StatisticsRequest{})
+		if err != nil {
+			return err
+		}
+		if res != nil {
```
skoryk-oleksandr, line 320: Is it enough to check only for `nill`? should we also check for empty object? If I'm not mistaken GRPC returns  zero-value struct (not nill) if there’s "no data".

goldmane/fv/daemon_test.go
skoryk-oleksandr, line 302: `createKeyCertPair()` or `daemonSetup()` look like utility functions. Should we add `t.Helper()` into these functions to make stack traces clearer when they fail. Maybe we should also think about moving them to some utility class to make test code cleaner and improve reusability.

caseydavenport: Yeah, it might be possible to run these in parallel with a bit of work - we'd need to commonize a lot more than the context though (for example, we'd need separate TCP ports for each test to bind to and logic to manage allocation of ports for each test).

This is relying on existing test infrastructure and refactoring that is I think outside the scope of this PR. I'm not against us taking time to investigate how difficult it would be to make these tests run in parallel as a separate task, though.

caseydavenport: >  about moving them to some utility class to make test code cleaner and improve reusability.

Maybe? they are already reused within this test suite. If we find a spot that can use them sure, but we should be careful about over engineering utilities. It's not always better to write something as a library function if it's actually only used in one place!

Better to keep it private and then export it later if we end up needing it somewhere else.

goldmane/fv/daemon_test.go
caseydavenport: Yep, good catch. This should be a `len(res)` check.

goldmane/fv/daemon_test.go
caseydavenport: `pusher.Connect` doesn't return an error though?

goldmane/fv/daemon_test.go
caseydavenport: FWIW, I believe this code actually _does_ work based on how the List() function is implemented: https://github.com/projectcalico/calico/blob/7fe4acde84f2450ef720165f1816f2ede87a5325/goldmane/pkg/client/statistics_service.go#L54-L65

Note that if there are no results from the stream we'll return an uninitialized `[]*proto.StatisticsResult`, which will evaluate to `nil`. 

Still better to improve the check though.

skoryk-oleksandr: I see :ok_hand:

skoryk-oleksandr (approved):

PR #10295, opened 2025-04-25
Optimize error prompts

DrAuYueng:
## Description
Improve the error message of `ErrorResourceAlreadyExists` type to avoid losing the original Err.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
NONE
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/errors/errors.go
+++ b/libcalico-go/lib/errors/errors.go
@@ -96,7 +96,7 @@ func (e ErrorResourceAlreadyExists) Unwrap() error {
 }
 
 func (e ErrorResourceAlreadyExists) Error() string {
-	return fmt.Sprintf("resource already exists: %v", e.Identifier)
+	return fmt.Sprintf("resource already exists: %v with error: %v", e.Identifier, e.Err)
 }
 
 // Error indicating a problem connecting to the backend.


caseydavenport:
@DrAuYueng Thanks for the PR - before we can merge you'll need to sign the CLA to make the bot happy.

Is this in service of some other goal? Or just something you noticed? I would expect this error to always be the same thing, so not especially useful to log out.

DrAuYueng:
@caseydavenport IMO, `Err` is defined and should not be swallowed in the subsequent processing, so err seems meaningless, for example:
```
if b.Allocations[ordinal] != nil {
		return cerrors.ErrorResourceAlreadyExists{
			Err:        fmt.Errorf("Address already assigned in block"),
			Identifier: address.String(),
		}
}
```

caseydavenport:
Makes sense, thanks @DrAuYueng

caseydavenport (approved):

marvin-tigera:

DrAuYueng:
@caseydavenport CI check failed due to error judgment problem. I just fixed it. Please take a look again. Thanks.

caseydavenport (approved):

PR #10301, opened 2025-04-25
[CORE-11344] Fix whisker multi-arch manifests [master]

danudey:
A subtle bug in our Makefiles.

We set `PUSH_MANIFEST_IMAGES` in `lib.Makefile` using the `:=` syntax; this causes `make` to evaluate the line immediately and store it as a static string, rather than lazily evaluating the entire line every time the variable is accessed.

By setting the value of `BUILD_IMAGES` after `lib.Makefile` is imported, that variable is never picked up; the `push-manifests` target then has nothing to do and exits silently, and no multi-arch manifests are ever pushed.

Solve this by moving the definition of `BUILD_IMAGES` above the import.

### DIFF
--- a/whisker-backend/Makefile
+++ b/whisker-backend/Makefile
@@ -1,6 +1,8 @@
 include ../metadata.mk
 
 PACKAGE_NAME = github.com/projectcalico/calico/whisker-backend
+
+# Configure variables used by ci/cd common targets from lib.Makefile.
 IMAGE_BUILD_MARKER = whisker-backend-$(ARCH).created
 BUILD_IMAGES = whisker-backend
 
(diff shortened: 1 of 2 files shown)


whisker-backend/Makefile
```diff
@@ -11,7 +12,6 @@ IMAGE_BUILD_MARKER = whisker-backend-$(ARCH).created
 include ../lib.Makefile
 
 # Configure variables used by ci/cd common targets from lib.Makefile.
```
caseydavenport, line 14: I think this comment needs to move with it?

danudey: Fixed.

caseydavenport (approved):

PR #10306, opened 2025-04-25
Support reconnection/resync with Typha

fasaxc:
## Description

- Add optional callback for reconnection-aware clients.

- Adjust Typha discovery to reset after all Typhas have been tried.

- Make the dedupe buffer reconnection-aware. It now

  - Stores off the keys that it had previously seen when it gets the OnTyphaConnectionRestarted() call.
  - Discards those seen keys as the resync progresses.
  - Synthesises deletions for KVs that weren't seen during the resync.
  - Recalculates the UpdateType when sending keys downstream so that the calculation graph sees a resync as a sequence of updates for exisitng keys.

- Refactor the client so that it

  - Does one connecction synchronously (including connection attempts to mutliple Typha instances as before)
  - Reconnects in the background after a failure.
  - Sends WaitForDatastore/ResyncInProgress messages when it's doing a reconnection.  (This should make the `felix_resyncs_started` and `felix_resync_state` Prometheus metrics useful again.
  - Re-uses a single connection attempt tracker so that we cycle through Typha instances on reconnect.

- Varous minor changes:

  - Add "done" channels to various components to avoid "log to testing.T after test finished" errors in new tests.
  - Add 32 bit random value to connection ID. Makes it a lot more greppable in logs.

## Related issues/PRs

CORE-11348

## Todos

The client will still bail out if it can't connect to any Typha.  Not sure if that's desirable or not; we could keep retrying but issues like running out of file handles might be better handled with a restart.

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Scale: Typha clients (such as Felix) now try to reconnect and resync with Typha on connection failure.  This reduces the impact of Typha restart and load rebalancing.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/daemon/daemon.go
+++ b/felix/daemon/daemon.go
@@ -601,8 +601,9 @@ configRetry:
 			break
 		}
 		healthAggregator.Report(healthName, &health.HealthReport{Live: true, Ready: true})
-		// Up-to-date Typha client will refuse to connect unless Typha signals
-		// that it supports node resource updates.
+
+		// Typha client now requires support for node updates and will refuse
+		// to connect to an (ancient) Typha that does not support them.
 		configParams.SetUseNodeResourceUpdates(true)
 
 		go func() {
--- a/typha/fv-tests/reconnection_test.go
+++ b/typha/fv-tests/reconnection_test.go
@@ -23,7 +23,7 @@ func TestReconnection(t *testing.T) {
 	h, recorder, clientStoppedC := setUpReconnectionTest(t)
 
 	// Deliberately configure the servers with different KVs.  When the client
-	// flip flows between the two, it should sync with whichever server it
+	// flip-flops between the two, it should sync with whichever server it
 	// connects to.  This may mean synthesising a deletion for the second key.
 	h0ConfigUpdates := h[0].SendConfigUpdates(1)
 	h[0].SendStatus(api.InSync)
@@ -55,7 +55,7 @@ func TestReconnectionNewServerNotInSync(t *testing.T) {
 	h, recorder, clientStoppedC := setUpReconnectionTest(t)
 
 	// Deliberately configure the servers with different KVs.  When the client
-	// flip flows between the two, it should sync with whichever server it
+	// flip-flops between the two, it should sync with whichever server it
 	// connects to.  This may mean synthesising a deletion for the second key.
 	h0ConfigUpdates := h[0].SendConfigUpdates(2)
 	h[0].SendStatus(api.InSync)
--- a/typha/pkg/discovery/discovery.go
+++ b/typha/pkg/discovery/discovery.go
@@ -55,7 +55,7 @@ func (t Typha) String() string {
 		nodePart = "/" + *t.NodeName
 	}
 	if strings.Contains(t.Addr, t.IP) {
-		// Mainline: IP matches address; avoid printing it twice.
+		// Mainline: IP included in address; avoid printing it twice.
 		return t.Addr + nodePart
 	}
 	return t.Addr + "," + t.IP + nodePart
@@ -291,12 +291,17 @@ func NewConnAttemptTracker(d AddressLoader) *ConnectionAttemptTracker {
 // and then returns the first entry in the list that has not been returned before.  If the list is static,
 // NextAddr() will effectively just iterate through the static list.
 type ConnectionAttemptTracker struct {
-	discoverer         AddressLoader
-	triedCache         bool
+	discoverer AddressLoader
+	triedCache bool
+
+	// triedAddrsLastSeen has en entry for each Typha address that has been
+	// tried.  We use presence of an entry to prevent trying the same address
+	// multiple times before we've tried all available addresses.  The timestamp
+	// is used to clean up stale entries; each time we choose an address, we
+	// refresh the timestamp on all the addresses that still exist.  The map is
+	// reset when we run out of addresses to try.
 	triedAddrsLastSeen map[addrDedupeKey]time.Time
 }
-
-var ErrTriedAllAddrs = fmt.Errorf("tried all available discovered addresses")
 
 func (d *ConnectionAttemptTracker) NextAddr() (Typha, error) {
 	allKnownAddrs, err := d.refreshAddrs()
@@ -340,7 +345,7 @@ func (d *ConnectionAttemptTracker) pickNextTypha(allKnownAddrs []Typha) (out Typ
 	foundUnusedTypha := false
 
 	// Defensive: make sure we don't leak if typha addresses are churning.
-	d.gcLastSeen(allKnownAddrs)
+	d.refreshAndGCLastSeen(allKnownAddrs)
 
 	for _, a := range allKnownAddrs {
 		addrKey := a.dedupeKey()
@@ -365,9 +370,10 @@ func (d *ConnectionAttemptTracker) pickNextTypha(allKnownAddrs []Typha) (out Typ
 	return
 }
 
-// gcLastSeen updates the last-seen times for d.triedAddrsLastSeen and cleans
-// up entries that haven't been seen for a long time.
-func (d *ConnectionAttemptTracker) gcLastSeen(addrs []Typha) {
+// refreshAndGCLastSeen updates the last-seen times for existing entries in
+// d.triedAddrsLastSeen and cleans up entries that haven't been seen for a
+// long time.
+func (d *ConnectionAttemptTracker) refreshAndGCLastSeen(addrs []Typha) {
 	for _, a := range addrs {
 		addrKey := a.dedupeKey()
 		if _, ok := d.triedAddrsLastSeen[addrKey]; ok {
--- a/typha/pkg/syncclient/sync_client.go
+++ b/typha/pkg/syncclient/sync_client.go
@@ -48,6 +48,10 @@ func init() {
 	nextID.Store(rand.Uint64() << 32)
 }
 
+func allocateConnID() uint64 {
+	return nextID.Add(1)
+}
+
 const (
 	defaultReadtimeout  = 30 * time.Second
 	defaultWriteTimeout = 10 * time.Second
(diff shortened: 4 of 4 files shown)


felix/daemon/daemon.go
```diff
@@ -591,15 +591,9 @@ configRetry:
...
-		supportsNodeResourceUpdates, err := typhaConnection.SupportsNodeResourceUpdates(10 * time.Second)
-		if err != nil {
-			time.Sleep(time.Second) // Avoid tight restart loop in case we didn't really wait 10s above.
-			log.WithError(err).Fatal("Did not get hello message from Typha in time")
-			return
-		}
-		log.Debugf("Typha supports node resource updates: %v", supportsNodeResourceUpdates)
-		configParams.SetUseNodeResourceUpdates(supportsNodeResourceUpdates)
+		// Up-to-date Typha client will refuse to connect unless Typha signals
+		// that it supports node resource updates.
```
aaaaaaaalex, line 595: Any issue with version skew on-upgrade?

fasaxc: I think the feature was added in 2019 so, yes, but only if you're skipping a dozen versions.  Even in that case, it will only block new felix from talking to old typha, felix will just keep trying untill it connects to an up-level typha.

nelljerram: It's not clear what I should understand from the "Up-to-date" in this comment.  IIUC we're using Typha client code from the same point in this monorepo, so the level of that code is already completely determined - right?  Would it mean anything different if the comment said just "Typha client will refuse to connect unless ..." ?

libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
```diff
@@ -69,11 +73,39 @@ func New() *DedupeBuffer {
...
+	defer d.lock.Unlock()
+
+	// We're about to be sent a complete new snapshot of the data.  Clear
+	// our in-flight state and make a transient copy of the keys that we have
+	// already sent so that we can figure out if any KVs were deleted while
+	// we were disconnected.
+	log.Info("Typha connection restarted, clearing pending update queue.")
+	clear(d.keyToPendingUpdate)
+	d.pendingUpdates = list.List{}
+	if d.liveKeysNotSeenSinceReconnect == nil {
```
nelljerram, line 87: Is there a hole in the scenario of multiple reconnection, like this?
1. Key X exists and has been sent downstream.
2. Reconnection # 1 and we start resyncing, but don't complete it.  During this phase key X still exists and is included in the new snapshot, so we remove X from `liveKeysNotSeenSinceReconnect`.
3. Reconnection # 2, and at about the same time X was deleted from the DB.  So it isn't included in the new new snapshot.  But we won't send a downstream delete for it because it has already been removed from `liveKeysNotSeenSinceReconnect`.

nelljerram: I guess I haven't got that exactly right, because just below we have
```
	d.liveKeysNotSeenSinceReconnect = d.liveResourceKeys.Copy()
```
regardless of whether it's # 1 or # 2.  But now the question becomes when and how `liveResourceKeys` is updated...

typha/pkg/discovery/discovery.go
```diff
@@ -38,16 +39,30 @@ type Typha struct {
...
+	return addrDedupeKey(fmt.Sprintf("%s/%s/%s", t.Addr, t.IP, node))
+}
+
+func (t Typha) String() string {
+	nodePart := ""
+	if t.NodeName != nil {
+		nodePart = "/" + *t.NodeName
+	}
+	if strings.Contains(t.Addr, t.IP) {
+		// Mainline: IP matches address; avoid printing it twice.
```
nelljerram: Perhaps: "IP is included in the address; avoid printing the IP twice."

typha/pkg/discovery/discovery.go
```diff
@@ -251,76 +279,107 @@ type AddressLoader interface {
 }
 
 func NewConnAttemptTracker(d AddressLoader) *ConnectionAttemptTracker {
-	return &ConnectionAttemptTracker{
-		discoverer:           d,
-		previouslyTriedAddrs: set.New[string](),
+	cat := &ConnectionAttemptTracker{
+		discoverer:         d,
+		triedAddrsLastSeen: map[addrDedupeKey]time.Time{},
```
nelljerram, line 284: Is there a better word than "seen" - not obvious if this means the last time we connected to them, or when that connection ended?

nelljerram: Looks like "picked".

typha/pkg/syncclient/sync_client.go
```diff
@@ -115,87 +123,127 @@ func New(
...
+
+		for cxt.Err() == nil {
+			connectionFinishedWG.Wait()
+			if rac, ok := s.callbacks.(RestartAwareCallbacks); ok {
+				log.Info("Typha connection failed but client callback is restart-aware.  Restarting connection...")
+				s.refreshConnID()
+				rac.OnTyphaConnectionRestarted()
+				s.callbacks.OnStatusUpdated(api.WaitForDatastore)
+			} else {
+				log.Info("Typha client callback is not restart-aware. Exiting...")
```
nelljerram, line 225: In these "return" cases, how does the client learn that there is no longer a working Typha connection?

typha/pkg/discovery/discovery.go
```diff
@@ -251,76 +279,107 @@ type AddressLoader interface {
...
+	}
+	if !foundUnusedTypha {
+		// We've tried them all, reset the tracking set so we'll loop again...
+		logrus.Debug("No unused Typha address found. Resetting.")
+		clear(d.triedAddrsLastSeen)
+		// ...starting with the first in the list.
+		out = allKnownAddrs[0]
+	}
+	logrus.WithField("addr", out).Debug("Next typha to try.")
+	d.triedAddrsLastSeen[out.dedupeKey()] = time.Now()
```
nelljerram, line 368: Do we need this?  IIUC all the time stamps were just set anyway by gcLastSeen.

typha/pkg/discovery/discovery.go
nelljerram: Sorry, no, it's when they were returned from refreshAddrs.

typha/pkg/discovery/discovery.go
```diff
@@ -251,76 +279,107 @@ type AddressLoader interface {
...
-		d.allKnownAddrs = d.discoverer.CachedTyphaAddrs()
-		logrus.WithField("addrs", d.allKnownAddrs).Debug("Using cached typha addresses")
-		if len(d.allKnownAddrs) > 0 {
-			return nil
+func (d *ConnectionAttemptTracker) refreshAddrs() ([]Typha, error) {
+	if !d.triedCache {
+		// Very first time, we expect the discoverer to have a cache of the
+		// addresses it loaded at start-up.  Try that.
+		d.triedCache = true
+		allKnownAddrs := d.discoverer.CachedTyphaAddrs()
```
nelljerram, line 320: Would it be clearer to move the `triedCache` logic down into the discoverer?  It looks like it would feel more natural for the discoverer itself to know whether to use CachedTyphaAddrs or LoadTyphaAddrs.

libcalico-go/lib/backend/syncersv1/dedupebuffer/dedupe_buffer.go
nelljerram: OK I've reviewed `liveResourceKeys` and think this is watertight after all.  Sorry for the noise.  Will leave these comments just in case you want to triple-check.

typha/fv-tests/reconnection_test.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/projectcalico/calico/typha/pkg/syncclient"
+	"github.com/projectcalico/calico/typha/pkg/syncproto"
+	"github.com/projectcalico/calico/typha/pkg/syncserver"
+)
+
+func TestReconnection(t *testing.T) {
+	h, recorder, clientStoppedC := setUpReconnectionTest(t)
+
+	// Deliberately configure the servers with different KVs.  When the client
+	// flip flows between the two, it should sync with whichever server it
```
nelljerram: nit: "flops" I guess

nelljerram: (more occurrences below)

nelljerram (approved):
LGTM; just a few nits and minor queries for your consideration.

typha/pkg/discovery/discovery.go
fasaxc: I'll add a comment, the map is doing double duty:
- Avoid retrying an address before we've tried all addresses.  (For this, it could just be a set.)
- Aging out entries that we haven't seen for a while. (For this we use a timestamp that gets updated even if we don't try the address on a particular iteration.)

typha/pkg/syncclient/sync_client.go
fasaxc: The defer calls `Finished.Done()`, which was already our signal for "typha client failed"

typha/pkg/discovery/discovery.go
fasaxc: Yes, `gcLastSeen` only updates existing timestamps.  I've updated naming/commenting to make it clearer.

typha/pkg/discovery/discovery.go
fasaxc: I think it would but ISTR there was an awkward problem that surfaced when I tried that.  Of course now I'm struggling to remember what it was!  Perhaps an issue with Felix's wireguard bootstrap logic that rippled out. (In any case, felt like this PR was big enough.)

nelljerram: Fair enough; it's fine to leave this to (possibly) later.

Issue #10058, opened 2025-03-25
[BPF] forwarding to peer with kubevirt does not work


Yoda317:
We have Kubevirt and Calico eBPF dataplane as cni. Virtual machine connected with pod network in bridge mode  couldn't communicate with anything outside node where it's running.
If eBPF disabled on this node by felixconfiguration, everything works fine

## Expected Behavior

Virtual machine is reachable from outside node where it is running

## Current Behavior

Virtual machine is reachable only from host network and other vms on this node

## Steps to Reproduce (for bugs)

1. Calico with eBPF enabled 
felixconfig
```
kind: FelixConfiguration
metadata:
  name: default
spec:
  bpfDataIfacePattern: ^((en|wl|ww|sl|ib)[Popsx].*|(wlan|wwan).*|tunl0$|vxlan.calico$|vxlan-v6.calico$|wireguard.cali$|wg-v6.cali$|egress.calico$|(eth|bond)[0-9]+.[0-9]+$)
  bpfEnabled: true
  bpfKubeProxyEndpointSlicesEnabled: true
  bpfKubeProxyIptablesCleanupEnabled: false
  bpfLogLevel: ""
  floatingIPs: Disabled
  logSeverityScreen: Warning
  prometheusGoMetricsEnabled: false
  prometheusProcessMetricsEnabled: false
  reportingInterval: 0s
  usageReportingEnabled: false
  vxlanPort: 4790
  xdpEnabled: false
```
#
2. Kubevirt virtual machine connected to pod network in bridge mode.
vm manifest
```
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  name: vmp-06
  namespace: kubevirt
spec:
  running: false
  template:
    metadata:
      labels:
        kubevirt.io/size: medium 
        kubevirt.io/domain: testvm
    spec:
      domain:
        devices:
          disks:
            - name: disk0 
              disk:
                bus: virtio
            - name: cloudinitdisk
              disk:
                bus: virtio
          interfaces:
          - name: default 
            bridge: {}
        resources:
          requests:
            memory: 2G
      networks:
      - name: default
        pod: {}
      nodeSelector:
        kubernetes.io/hostname: node63144
      volumes:
        - name: disk0
          persistentVolumeClaim:
            claimName: pvc-vmp-06
        - name: cloudinitdisk
          cloudInitNoCloud:
            userData: |-
              #cloud-config
                             
```

## Context

```
[...]

PR #10308, opened 2025-04-25
Do not use redirect_peer when sending packets to kubevirt VMs

sridhartigera:
## Description

In eBPF mode, we use ```bpf_redirect_peer``` when sending packets to a local workload from the host interface. When redirecting to the peer interface of the pod, we don't change the destination MAC before redirecting.

When VMs run as workloads, like kubevirt, redirecting to the workload's peer with an incorrect MAC results in packet gets dropped. This happens when the kubevirt VM's networking is in bridge mode. 

 The challenge is to identify if the local workload is a kubevirt VM. We can rely on a label added by the kubevirt to the pod as mentioned in https://kubevirt.io/user-guide/cluster_admin/annotations_and_labels. 

Fix -
1. Identify the workload as kubevirt VM  using the label ```kubevirt.io=virt-launcher```
2. If the workload is a kubevirt VM, add a flag to the bpf route entry.
3. If the route has the VM flag, add a flag to skip redirection in the conntrack. (First packet will anyways be sent out using fib_lookup).
4. From the 2nd packet, skip redirection if the conntrack has the flag.

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/10058

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
eBPF - Fix forwarding of packets to kubevirt pods when in bridge mode.
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/conntrack_types.h
```diff
@@ -38,6 +38,7 @@ enum cali_ct_type {
 #define CALI_CT_FLAG_NP_LOOP	0x800 /* marks connections that were turned around when accessing nodeport on a local IP */
 #define CALI_CT_FLAG_NP_REMOTE	0x1000 /* marks connections from local host to remote backend of a nodeport */
 #define CALI_CT_FLAG_NP_NO_DSR	0x2000 /* marks connections from a client which is excluded from DSR */
+#define CALI_CT_FLAG_SKIP_REDIR	0x4000 /* marks connections from a client which is excluded from redir */
```
tomastigera, line 41:
```suggestion
#define CALI_CT_FLAG_SKIP_REDIR_PEER	0x4000 /* marks connections that should not be redirected to wl peer side */
```

It should not skip redir completely, right? redirect_neigh is perfectly OK

felix/bpf-gpl/tc.c
```diff
@@ -606,16 +606,20 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
...
-		 * policed. It is ingress into the system and we got a packet, which is
-		 * not for this host, and it wasn't resolved as a service and it is not
-		 * for a local workload either. But we hit a route so it may be some L2
-		 * broadcast, we do not quite know. Let the host route it or dump it.
-		 *
-		 * https://github.com/projectcalico/calico/issues/8918
-		 */
-		ctx->state->flags |= CALI_ST_SKIP_FIB;
+	} else if (CALI_F_FROM_HEP) {
+		if (cali_rt_flags_local_workload_vm(dest_rt->flags)) {
```
tomastigera, line 610: This will be valid from WEP after https://github.com/projectcalico/calico/pull/9965 merges

felix/proto/felixbackend.proto
```diff
@@ -405,6 +405,7 @@ message WorkloadEndpoint {
   map<string, string> annotations = 11;
   QoSControls qos_controls = 12;
   LocalBGPPeer local_bgp_peer = 13;
+  map<string, string> labels = 14;
```
tomastigera, line 408: I'd double check with @fasaxc if adding another non-empty possibly large map is OK just to accommodate one specific usecase. Obviously, having the label might prove handy in other places.

felix/fv/bpf_test.go
```diff
@@ -1469,6 +1470,42 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+				vmw.ConfigureInInfra(infra)
+				cc.Expect(Some, w[1][0], vmw, ExpectWithSrcPort(12345))
+				cc.CheckConnectivity()
+				m := dumpCTMapsAny(4, tc.Felixes[0])
+				k := conntrack.NewKey(6, net.ParseIP(w[1][0].IP).To4(), 12345,
+					net.ParseIP(vmw.IP).To4(), 8055)
+				Expect(m).To(HaveKey(k))
+				val := m[k]
+				flags := val.Flags()
+				Expect(flags & ctv3.FlagNoRedir).To(Equal(ctv3.FlagNoRedir))
```
tomastigera, line 1506: This looks like something that would be good enough to do in a UT. The fv does not guarantee that the packet would be dropped by the simulated workload. I guess you can create the bridge in the namespace and emulate delivering the packet like in the VM case. Than just a connectivity test would be sufficient and would make sense. All that is being tested here is that the label is propagated to conntrack.

felix/proto/felixbackend.proto
fasaxc: That would be OK because we only do it for local workloads (so we're limited to a few hundred label maps).  

That said, I usually prefer to do the decision making in the calc graph so I might have added a dedicated field to indicate the workload type.

We do already pass the annotations through (used by VPP), so I suppose this would be consistent with that.

tomastigera (approved):
thanks

PR #10314, opened 2025-04-28
Kube-controllers updates - syncer, batching

MichalFupso:
## Description
Updates to calico kube-controllers

- syncer adds resourceVersion onto the resource from KVPair version when running in etcd mode
- adds batch processor, general implementation of batch processing to be used by the controllers, allows for a quick emptying of the updateChan and processes updates in batch 
- moves cmdwrapper to calico/pkg/cmdwrapper
- new func for getting default KCC without the need of exporting it in runconfig
- removes warning about unexpected kind received via syncer, syncer is used by majority of the controllers, controllers should expect updates not relevant to them and ignore those

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


kube-controllers/cmd/wrapper/wrapper.go
```diff
@@ -15,7 +15,7 @@
 package main
 
 import (
-	"github.com/projectcalico/calico/typha/pkg/cmdwrapper"
+	"github.com/projectcalico/calico/kube-controllers/pkg/cmdwrapper"
```
caseydavenport, line 18: I was actually thinking this might make sense as a generic binary / package that isn't tied to any one project. We use it in kube-ctrl in OSS, and Typha in enterprise - seems like a generic and well-scoped tool without being tied to any particular project?

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller_ut_test.go
```diff
@@ -64,7 +78,7 @@ var _ = Describe("LoadBalancer controller UTs", func() {
 
 		factory.Start(stopChan)
 		cache.WaitForCacheSync(stopChan, serviceInformer.HasSynced)
-		dataFeed := utils.NewDataFeed(cli)
+		dataFeed := utils.NewDataFeed(cli, "etcdv3")
```
caseydavenport, line 81: Isn't there a constant it should use?

kube-controllers/pkg/controllers/utils/batch_processor.go
```diff
@@ -0,0 +1,51 @@
...
+
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	// BatchUpdateSize length of the buffered channel for the controllers and the max batch size to handle during ProcessBatch
+	BatchUpdateSize = 1000
+
+	// batchWaitDuration maximum time we wait for more updates to be received via channel before starting to process the updates
+	batchWaitDuration = 1 * time.Second
```
caseydavenport, line 14: not sure if it matters - some of the places you've replaced with this function didn't have a wait duration set prior to this, they just read until the channel was empty. That might just be an unimportant difference, but have you considered if there's any impact of that change?

kube-controllers/pkg/controllers/utils/syncer.go
```diff
@@ -91,18 +97,48 @@ func (d *DataFeed) OnStatusUpdated(status bapi.SyncStatus) {
...
+
 	// For each consumer registered for this type, send an update.
-	for _, f := range c.registrations[t] {
+	for _, f := range d.registrations[t] {
 		f(update)
 	}
 }
+
+// updateResourceVersion updates the resourceVersion of the resource when we run in etcd mode. The resource version is revision on the KVPair
+func (d *DataFeed) updateResourceVersion(update bapi.Update) {
```
caseydavenport, line 121: Did we talk about doing this within the syncer itself? Or no? I can't remember...

That seems like the natural place for this - otherwise any other consumers of the syncer will need to re-implement this same logic.

kube-controllers/pkg/controllers/utils/syncer.go
```diff
@@ -27,10 +27,14 @@ import (
 	client "github.com/projectcalico/calico/libcalico-go/lib/clientv3"
 )
 
+const (
+	Etcdv3 = "etcdv3"
```
caseydavenport, line 31: Do we not need a k8s const as well?

kube-controllers/pkg/controllers/utils/batch_processor.go
MichalFupso: For the autoHEP and labels controller, I think we can expect that we could receive multiple node updates in a short time, so it's beneficial to empty the chan quickly. I don't think we're going to see much of this for the loadbalancer controller, but at most it increases the wait time by 2 seconds. Maybe we don't use the batch_processor for syncer updates as those should be fairly quick

kube-controllers/pkg/controllers/utils/syncer.go
MichalFupso: So far we only care if we're running in etcd mode to update the resourceVersion. In k8s more it's already part of the resource

kube-controllers/pkg/controllers/utils/syncer.go
MichalFupso: In the future we should update the backend syncer directly so the resources contain the resourceVersion. This is a workaround till then

kube-controllers/pkg/controllers/utils/syncer.go
```diff
@@ -91,18 +97,51 @@ func (d *DataFeed) OnStatusUpdated(status bapi.SyncStatus) {
...
 	}
 }
+
+// updateResourceVersion updates the resourceVersion of the resource when we run in etcd mode. The resource version is revision on the KVPair
+// This is a workaround for a fact that the backend syncer api does not update the resourceVersion.
+// Kube-controller syncer should not have to be aware of the datastore and the backend syncer should correctly update the resourceVersion
+// This can be removed once the backend syncer code is updates
+func (d *DataFeed) updateResourceVersion(update bapi.Update) {
+	if update.Value == nil {
+		// We received delete, we don't have to update the resourceVersion as the resource does not exists
```
caseydavenport, line 126:
```suggestion
		// We received delete, we don't have to update the resourceVersion as the resource does not exist
```

caseydavenport (approved):

PR #10319, opened 2025-04-29
Workaround error with gRPC trailers over proxies

caseydavenport:
## Description

gRPC uses trailers to send status information for streamed requests. We
have found that some proxies don't handle this super well when there is
no body in the response. Likely because they are failing to wait long
enough to proxy the trailers to the client, and preemptively closing the
proxied connection.

This change sends an empty status response, but explicitly sets the gRPC
status code to "Not found" - this works around the empty body issue with
proxies, while still appearing as an empty response to clients.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/goldmane/pkg/client/statistics_service.go
+++ b/goldmane/pkg/client/statistics_service.go
@@ -20,6 +20,8 @@ import (
 	"io"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


doucol (approved):
LGTM

PR #10320, opened 2025-04-29
Build images for Envoy Gateway v1.3.2, to address CVEs

nelljerram:
- envoy-gateway code is updated to v1.3.2.  Removed patch because the upstream version now already includes that change.

- Per the [gateway release notes](https://github.com/envoyproxy/gateway/releases/tag/v1.3.2), envoy-ratelimit code is updated to 0141a24f.  Removed patch because the upstream version now already includes that change.

- envoy-proxy code should be updated to v1.33.1, but we'll do that in a separate step once we've been able to build that version.  We _think_ Gateway API will still work anyway with the older envoy-proxy code, and the other updates will allow us to reassess the CVE position.

Also add CI blocks to test envoy image builds pre-merge.

Also link to doc in the operator repo about the process for updating these versions.


Behnam-Shobiri:
Maybe not for this PR, but we need a new patch update x/net to latest

nelljerram:
@Behnam-Shobiri In which build do you mean?  (gateway, proxy or ratelimit?)

Behnam-Shobiri:
I believe we need it for gateway and ratelimit, since x/net min safe version is v0.38.0 (latest v0.39.0)

nelljerram:
Thanks @Behnam-Shobiri I'll do those in a following PR.

coutinhop (approved):

PR #10330, opened 2025-04-30
[CORE-11367] openstack: Add workaround for limited /boot partition space when an initramfs update is needed

coutinhop:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/.semaphore/semaphore-scheduled-builds.yml
+++ b/.semaphore/semaphore-scheduled-builds.yml
@@ -934,6 +934,15 @@ blocks:
             # this was causing the DevStack setup to fail, because pip doesn't know how to uninstall
             # or replace the existing version.  Happily we do know that, so let's do it upfront here.
             - sudo apt-get remove -y python3-wrapt || true
+            # Install all the packages that would trigger an initramfs update using a workaround
+            # for limited /boot partition space in the ubuntu2004 image
+            - sudo apt update
+            - sudo rsync -av /boot/ /boot2/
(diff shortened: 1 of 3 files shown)


lucastigera (approved):
LGTM. Given that an attempt to upgrade the Ubuntu image caused many other jobs to fail, this workaround seems like a better option for now.

nelljerram:
@coutinhop Can you add more detail here about what happened to require this change?  The PR title says "when an initramfs update is needed", but **is** an initramfs update needed?  At what point in the job does that happen?  Did something about the Semaphore machine setup change so that it **does** now happen, whereas previously it didn't?

coutinhop:
> @coutinhop Can you add more detail here about what happened to require this change?  The PR title says "when an initramfs update is needed", but **is** an initramfs update needed?  At what point in the job does that happen?  Did something about the Semaphore machine setup change so that it **does** now happen, whereas previously it didn't?

@nelljerram sorry for the delay. It's been a while, but from what I remember the initramfs update is triggered by certain packages being updated when running `apt upgrade`. I forget the exact reason but I believe we do need to update some package(s). Through trial and error I figured which packages triggered it and explicitly updated them with the workaround.

As far as I can tell, the initramfs update was always happening, the issue is the disk space for the `/boot` partition became insufficient as updates have gotten larger.

I'll dig for more details tomorrow and update the PR description.

PR #10333, opened 2025-04-30
Improve performance of IPAM GC controller

caseydavenport:
## Description

This fixes a few performance issues:

- IPAM allocations were done serially. Instead, do them block-by-block to reduce the number of API calls required.
- Don't artificially rate limit node affinity release.
- Don't perform a GET to the API server on every IP release - this was dreadfully slow, and we can be reasonably confident in our cache.
- Increase the QPS for the Calico client.

Hopefully this is an easy win for larger clusters. The only potentially iffy change is the change to no longer force an API request for each Pod, instead checking with our cache fed by the informer. I think this should be relatively safe - we already have a leak grace period to handle timing issues - and it **substantially** improves the performance of the GC code in my testing.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Improve IPAM garbage collection performance by reducing the number of API calls required to release batches of IP addresses. 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/kube-controllers/cmd/kube-controllers/main.go
+++ b/kube-controllers/cmd/kube-controllers/main.go
@@ -334,7 +334,17 @@ func startCompactor(ctx context.Context, interval time.Duration) {
 // getClients builds and returns Kubernetes and Calico clients.
 func getClients(kubeconfig string) (*kubernetes.Clientset, client.Interface, error) {
 	// Get Calico client
-	calicoClient, err := client.NewFromEnv()
+	config, err := apiconfig.LoadClientConfigFromEnvironment()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Increase the client QPS on the Calico client.
+	// - For one, this client is shared across a number of different controllers, so will need a higher request count.
+	// - Secondly, the IPAM GC controller can potentially generate a very large number of requests.
+	config.Spec.K8sClientQPS = 500
+
+	calicoClient, err := client.New(*config)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to build Calico client: %s", err)
 	}
--- a/kube-controllers/cmd/kube-controllers/main.go
+++ b/kube-controllers/cmd/kube-controllers/main.go
@@ -359,6 +359,7 @@ func getClients(kubeconfig string) (*kubernetes.Clientset, client.Interface, err
 	// Increase the QPS of the Kubernetes client as well. This is also used heavily by the IPAM GC controller
 	// in some circumstances.
 	k8sconfig.QPS = 100
+	k8sconfig.Burst = 200
 
 	// Get Kubernetes clientset
 	k8sClientset, err := kubernetes.NewForConfig(k8sconfig)
--- a/kube-controllers/cmd/kube-controllers/main.go
+++ b/kube-controllers/cmd/kube-controllers/main.go
@@ -356,6 +356,10 @@ func getClients(kubeconfig string) (*kubernetes.Clientset, client.Interface, err
 		return nil, nil, fmt.Errorf("failed to build kubernetes client config: %s", err)
 	}
 
+	// Increase the QPS of the Kubernetes client as well. This is also used heavily by the IPAM GC controller
+	// in some circumstances.
+	k8sconfig.QPS = 100
+
 	// Get Kubernetes clientset
 	k8sClientset, err := kubernetes.NewForConfig(k8sconfig)
 	if err != nil {
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -1075,11 +1075,22 @@ func (c *IPAMController) syncIPAM() error {
 
 	c.allocationState.syncComplete()
 	log.Debug("IPAM sync completed")
+
+	// If there is still dirty state, then we need to do another pass.
+	if len(c.confirmedLeaks) > 0 {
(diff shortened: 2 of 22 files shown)


kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -123,11 +121,6 @@ func NewIPAMController(cfg config.NodeControllerConfig, c client.Interface, cs k
 		client:    c,
 		clientset: cs,
 		config:    cfg,
-		rl: workqueue.NewTypedMaxOfRateLimiter(
```
caseydavenport, line 126: tbh I'm not really sure why we had this rate limiting in place to begin with. I think removing it is OK to do? It was artificially slowing down our ability to release IPAM block affinities.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1050,42 +1048,35 @@ func (c *IPAMController) syncIPAM() error {
...
+	var opts []ipam.ReleaseOptions
+	leaks := map[ipam.ReleaseOptions]*allocation{}
 	for id, a := range c.confirmedLeaks {
 		logc := log.WithFields(a.fields())
 
-		// Final check that the allocation is leaked, this time ignoring our cache
-		// to make sure we're working with up-to-date information.
-		if c.allocationIsValid(a, false) {
+		// Final check that the allocation is leaked.
+		if c.allocationIsValid(a, true) {
```
caseydavenport, line 1079: This is one change I'd like a second opinion on - it was taking a LOT of time to query each Pod individually, and switching to a cache check will potentially reduces hundreds to thousands of API calls made by this controller when - for example - tearing down a large number of nodes at once.

caseydavenport: But, if for some reason our informer cache doesn't match the actual state of the cluster, then we may take actions that are incorrect (like releasing an IP that we don't need to). I think this is unlikely though.

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -334,7 +334,17 @@ func startCompactor(ctx context.Context, interval time.Duration) {
...
-	calicoClient, err := client.NewFromEnv()
+	config, err := apiconfig.LoadClientConfigFromEnvironment()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Increase the client QPS on the Calico client.
+	// - For one, this client is shared across a number of different controllers, so will need a higher request count.
+	// - Secondly, the IPAM GC controller can potentially generate a very large number of requests.
+	config.Spec.K8sClientQPS = 500
```
coutinhop, line 345: just curious: what was the default before changing here?

kube-controllers/pkg/controllers/node/ipam.go
coutinhop: the "This is only for retry speed" makes me think this is to implement some kind of exponential backoff for retries? should be useful to keep if so, but not sure if that's the only thing this did...

kube-controllers/pkg/controllers/node/ipam.go
coutinhop: I'm not very familiar with this, so I went digging into the podLister that we use as cache (please let me know if I'm way off), but it seems like it does a k8s watch on pods? So I do agree with you, seems pretty unlikely that we'd do something wrong here due to the cache being outdated...

kube-controllers/pkg/controllers/node/ipam.go
caseydavenport: Yeah, but I think it was an outdated comment or just wrong. The ratelimited would block cleanup of other nodes if a single node hit an issue, and didn't seem to be related to retries.

Plus, the second argument here applies to every request - not just failed ones - which enforced an intrinsic 10 per second limit.

coutinhop (approved):

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1098,27 +1088,81 @@ func (c *IPAMController) garbageCollectKnownLeaks() error {
...
+		}
 	}
 	return nil
 }
 
+func (c *IPAMController) releaseNodes(nodes []string) error {
+	var storedErr error
+	if len(nodes) > 0 {
+		log.WithField("num", len(nodes)).Info("Found a batch of nodes to release")
+	}
```
mazdakn, line 1150: I know with len(nodes) == 0, the following for loop is a no-op, but I think it's more idiomatic to have:
```
if len(nodes) == 0 {
  return nil
}
log.WithField("num", len(nodes)).Info("Found a batch of nodes to release")
```

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1098,27 +1088,81 @@ func (c *IPAMController) garbageCollectKnownLeaks() error {
...
+	for _, cnode := range nodes {
+		logc := log.WithField("node", cnode)
+
+		// Potentially rate limit node cleanup.
+		logc.Info("Cleaning up IPAM affinities for deleted node")
+		if err := c.cleanupNode(cnode); err != nil {
+			// Store the error, but continue. Storing the error ensures we'll retry.
+			logc.WithError(err).Warnf("Error cleaning up node")
+			storedErr = err
+			continue
```
mazdakn, line 1160: nit: `continue` is redundant here.

mazdakn (approved):
LGTM from code standpoint.  just left a few nits.
Also it worth updating the copyright years.

PR #10334, opened 2025-04-30
update license bin dir path

radTuti:
## Description

follow-up to things missed in #10331 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


apiserver/Makefile
```diff
@@ -73,7 +73,7 @@ endif
 .PHONY: build
 build: $(BINDIR)/apiserver-$(ARCH)
 
-bin/LICENSE: ../LICENSE.md
+$(BINDIR)/LICENSE: ../LICENSE.md
```
caseydavenport, line 76: I'd prefer if we used `bin/` here like it was before and have the Dockerfile look in `bin` - BINDIR varies based on architecture, and we don't need a license per architecture.

apiserver/Dockerfile
```diff
@@ -17,9 +17,10 @@ ARG CALICO_BASE
 FROM scratch AS source
 
 ARG BIN_DIR
+ARG LICENSE_DIR
 ARG TARGETARCH
 
-COPY ${BIN_DIR}/LICENSE /licenses/LICENSE
+COPY ${LICENSE_DIR}/LICENSE /licenses/LICENSE
```
caseydavenport, line 23: We need an extra argument for this since it is always the same value. Let's just keep things simple.

```suggestion
COPY bin/LICENSE /licenses/LICENSE
```

caseydavenport (approved):

PR #10346, opened 2025-05-02
API server - Policy WatchManager 

MichalFupso:
## Description
This change introduces a WatchManager that keeps track of all active Watch channels and stops the watch when new tier is added. 

This fixes an issue when new Tier is added the Watch is not receiving events about policies in the new Tier. This is because when a Watch() is created and no selectors were specified by the consumer, we default the selector to contain all Tiers that the consumer has authorization for. When new Tier as added, it's possible that the user has access to those policies as well, and not receiving them. 

WatchManager keeps track of all active Watch events for policies, when new Tier is added the WatchManager cancels all Watch that it has in map. This forces the consumer to reestablish the Watch, this new Watch will include the newly added Tier.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/apiserver/pkg/apiserver/apiserver.go
+++ b/apiserver/pkg/apiserver/apiserver.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/projectcalico/calico/apiserver/pkg/rbac"
 	calicorest "github.com/projectcalico/calico/apiserver/pkg/registry/projectcalico/rest"
+	"github.com/projectcalico/calico/apiserver/pkg/registry/projectcalico/util"
 	"github.com/projectcalico/calico/apiserver/pkg/storage/calico"
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/model"
@@ -74,6 +75,7 @@ type ProjectCalicoServer struct {
 	GenericAPIServer      *genericapiserver.GenericAPIServer
 	RBACCalculator        rbac.Calculator
 	CalicoResourceLister  CalicoResourceLister
+	WatchManager          *util.WatchManager
 	SharedInformerFactory informers.SharedInformerFactory
 }
 
@@ -119,6 +121,10 @@ func (c completedConfig) New() (*ProjectCalicoServer, error) {
 	// implement our own syncer-based cache.
 	calicoLister := NewCalicoResourceLister(cc)
 
+	// Create the manager for policy watches. It keeps track of established watches and makes sure they are canceled
+	// in case of on update that the watch should contain
+	watchManager := util.NewWatchManager(cc)
+
 	// Create the RBAC calculator,
 	calculator, err := c.NewRBACCalculator(calicoLister)
 	if err != nil {
@@ -129,11 +135,12 @@ func (c completedConfig) New() (*ProjectCalicoServer, error) {
 		GenericAPIServer:      genericServer,
 		RBACCalculator:        calculator,
 		CalicoResourceLister:  calicoLister,
+		WatchManager:          watchManager,
 		SharedInformerFactory: c.GenericConfig.SharedInformerFactory,
 	}
 
 	apiGroupInfo.VersionedResourcesStorageMap["v3"], err = calicostore.NewV3Storage(
-		Scheme, c.GenericConfig.RESTOptionsGetter, c.GenericConfig.Authorization.Authorizer, calicoLister)
+		Scheme, c.GenericConfig.RESTOptionsGetter, c.GenericConfig.Authorization.Authorizer, calicoLister, watchManager)
 	if err != nil {
 		return nil, err
 	}
--- a/apiserver/pkg/apiserver/apiserver.go
+++ b/apiserver/pkg/apiserver/apiserver.go
@@ -122,7 +122,7 @@ func (c completedConfig) New() (*ProjectCalicoServer, error) {
 	calicoLister := NewCalicoResourceLister(cc)
 
 	// Create the manager for policy watches. It keeps track of established watches and makes sure they are canceled
-	// in case of on update that the watch should contain
+	// in case of on update that the watch necessitates recreation of the watch.
 	watchManager := util.NewWatchManager(cc)
 
 	// Create the RBAC calculator,
--- a/apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
+++ b/apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
@@ -173,14 +173,14 @@ func (r *REST) Watch(ctx context.Context, options *metainternalversion.ListOptio
 	}
 
 	record := util.WatchRecord{
-		Id:    uuid.New().String(),
+		ID:    uuid.New().String(),
 		Kind:  calico.KindGlobalNetworkPolicy,
 		Watch: w,
 		Ctx:   ctx,
 	}
 	r.watchManager.AddWatch(record)
 
-	return w, err
+	return w, nil
 }
 
 func (r *REST) ShortNames() []string {
--- a/apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
+++ b/apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
@@ -16,6 +16,7 @@ package globalpolicy
 
 import (
 	"context"
+
 	"github.com/google/uuid"
 	calico "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	"k8s.io/apimachinery/pkg/api/meta"
--- a/apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
+++ b/apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
@@ -16,7 +16,7 @@ package globalpolicy
 
 import (
 	"context"
-
+	"github.com/google/uuid"
 	calico "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metainternalversion "k8s.io/apimachinery/pkg/apis/meta/internalversion"
@@ -38,8 +38,9 @@ import (
 type REST struct {
 	*genericregistry.Store
 	rbac.CalicoResourceLister
-	authorizer authorizer.TierAuthorizer
-	shortNames []string
+	authorizer   authorizer.TierAuthorizer
+	watchManager *util.WatchManager
+	shortNames   []string
 }
 
 // EmptyObject returns an empty instance
@@ -53,7 +54,7 @@ func NewList() runtime.Object {
 }
 
 // NewREST returns a RESTStorage object that will work against API services.
-func NewREST(scheme *runtime.Scheme, opts server.Options, calicoResourceLister rbac.CalicoResourceLister) (*REST, error) {
+func NewREST(scheme *runtime.Scheme, opts server.Options, calicoResourceLister rbac.CalicoResourceLister, watchManager *util.WatchManager) (*REST, error) {
 	strategy := NewStrategy(scheme)
 
 	prefix := "/" + opts.ResourcePrefix()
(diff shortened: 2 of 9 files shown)


apiserver/pkg/registry/projectcalico/util/watch_manager.go
```diff
@@ -0,0 +1,100 @@
...
+func NewWatchManager(cc api.Client) *WatchManager {
+	return &WatchManager{
+		client:       cc,
+		sync:         make(chan struct{}),
+		watchRecords: make(map[string]WatchRecord),
+	}
+}
+
+func (m *WatchManager) Start() {
+	m.syncer = watchersyncer.New(
```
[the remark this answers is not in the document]
MichalFupso: As you said, it's for Tiers that were not created by the API server

apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
```diff
@@ -165,7 +167,20 @@ func (r *REST) Watch(ctx context.Context, options *metainternalversion.ListOptio
...
+
+	record := util.WatchRecord{
+		ID:    uuid.New().String(),
+		Kind:  calico.KindGlobalNetworkPolicy,
+		Watch: w,
+		Ctx:   ctx,
+	}
+	r.watchManager.AddWatch(record)
+
+	return w, nil
```
lucastigera, line 183: Would it make sense to integrate this new code directly into the r.Store.Watch function?
Since the additional code is currently repeated after all four calls to this method, would it be worthwhile to add a kind string parameter to r.Store.Watch to reduce repetition?

This block change would then looks like: `return r.Store.Watch(ctx, options, calico.KindGlobalNetworkPolicy)`

lucastigera: This comment also applies to the other three similar changes in this PR. =)

apiserver/test/integration/clientset_test.go
```diff
@@ -60,6 +60,123 @@ func TestGroupVersion(t *testing.T) {
...
+		defer done.Done()
+		for i := 0; i < 1; i++ {
+			select {
+			case e := <-w.ResultChan():
+				events = append(events, e)
+			case <-timeout:
+				timeoutErr = fmt.Errorf("timed out waiting for events")
+				return
+			}
+		}
```
lucastigera, line 115: Why is a `for` loop used if we're only performing a single iteration?
Wouldn't it be clearer to place the `select` statement directly outside the loop and remove the loop altogether?

apiserver/test/integration/clientset_test.go
```diff
@@ -60,6 +60,123 @@ func TestGroupVersion(t *testing.T) {
...
+			})
+			defer shutdownServer()
+			if err := testPolicyWatch(client); err != nil {
+				t.Fatal(err)
+			}
+		}
+	}
+
+	if !t.Run(name, rootTestFunc()) {
+		t.Errorf("test-policywatch test failed")
```
lucastigera, line 79: Is it safe to assume the reason for the failure and include it in the error message to help with debugging?

apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
MichalFupso: The Store.Watch is used by all of our resources, we want to keep track of Watches only for policies, I think it's better to keep it within the policy API

apiserver/test/integration/clientset_test.go
MichalFupso: The actual more specific test failure is coming from the `testPolicyWatch` bellow.

apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
lucastigera: Oh! I didn’t realize that Store is a k8s package. During a quick review, I mistakenly thought it was a component we created and used only for those 4 calls. :grimacing:

apiserver/pkg/registry/projectcalico/util/watch_manager.go
```diff
@@ -0,0 +1,104 @@
...
+	if status == api.InSync {
+		close(m.sync)
+	}
+}
+
+func (m *WatchManager) OnUpdates(updates []api.Update) {
+	for _, u := range updates {
+		if u.KVPair.Value == nil {
+			// We do not need to cancel watches for delete as the original watch is still valid with other records
+			return
```
lucastigera, line 71: Is this `return` exiting the entire function? Shouldn't we `continue` instead? Something like:

```
func (m *WatchManager) OnUpdates(updates []api.Update) {
	// We do not need to cancel watches for delete as the original watch is still valid with other records
	if u.Key.(model.ResourceKey).Kind == v3.KindTier && u.KVPair.Value != nil {
		// New Tier added, we need to stop all watches in case there is a user that can watch policies in the new Tier
		// When the watch is re-established it will contain policies in the newly created Tier
		logrus.WithField("Tier", u.Key.(model.ResourceKey).Name).Debug("New Tier added, closing all policy watches")
		for _, record := range m.watchRecords {
			record.Watch.Stop()
		}
		m.watchRecords = map[string]WatchRecord{}
	}
}
```

apiserver/pkg/registry/projectcalico/util/watch_manager.go
```diff
@@ -0,0 +1,104 @@
...
+// AddWatch adds a watch to our map and triggers monitoring for watch closure by the consumer or API server
+func (m *WatchManager) AddWatch(record WatchRecord) {
+	logrus.WithFields(logrus.Fields{"id": record.ID, "Kind": record.Kind}).Debug("Adding WatchRecord")
+	m.watchRecords[record.ID] = record
+	go m.monitorWatch(record)
+}
+
+// minitorWatch waits to see if the context is done signaling that the watch has been ended. We can remove the watch from our map
+func (m *WatchManager) monitorWatch(record WatchRecord) {
+	for {
```
lucastigera, line 95: Is this `for` loop necessary?
Wouldn't the function return after the context is done?

apiserver/pkg/registry/projectcalico/util/watch_manager.go
```diff
@@ -0,0 +1,104 @@
...
+			// We do not need to cancel watches for delete as the original watch is still valid with other records
+			return
+		}
+
+		if u.Key.(model.ResourceKey).Kind == v3.KindTier {
+			// New Tier added, we need to stop all watches in case there is a user that can watch policies in the new Tier
+			// When the watch is re-established it will contain policies in the newly created Tier
+			logrus.WithField("Tier", u.Key.(model.ResourceKey).Name).Debug("New Tier added, closing all policy watches")
+			for _, record := range m.watchRecords {
+				record.Watch.Stop()
```
skoryk-oleksandr, line 79: Does it make sense to log also what policy watches have we stopped?

apiserver/pkg/registry/projectcalico/util/watch_manager.go
```diff
@@ -0,0 +1,104 @@
...
+	m.watchRecords[record.ID] = record
+	go m.monitorWatch(record)
+}
+
+// minitorWatch waits to see if the context is done signaling that the watch has been ended. We can remove the watch from our map
+func (m *WatchManager) monitorWatch(record WatchRecord) {
+	for {
+		<-record.Ctx.Done()
+		// Watch has been closed, we should remove it from our map
+		logrus.WithFields(logrus.Fields{"id": record.ID, "Kind": record.Kind}).Debug("Stopping WatchRecord")
```
skoryk-oleksandr, line 98: Should we also highlight what is the reason of the stopping WatchRecord?

apiserver/test/integration/clientset_test.go
```diff
@@ -60,6 +60,119 @@ func TestGroupVersion(t *testing.T) {
...
+				return
+			case <-timeout:
+				timeoutErr = fmt.Errorf("timed out waiting for events")
+				return
+			}
+		}
+	}()
+
+	// Creating a new Tier should force the watch to be closed
+	_, err = client.ProjectcalicoV3().Tiers().Create(context.Background(), tier, metav1.CreateOptions{})
```
skoryk-oleksandr, line 128: Should we wait here for Policy creation event is received before creating a tier (`done.Wait()`)? Is it possible sittuation when Tier could be created before the policy event is processed by the watcher? If that happens, the WatchManager will close the watch, and the goroutine will hit a timeout instead of receiving the event?

skoryk-oleksandr (approved):

apiserver/pkg/apiserver/apiserver.go
```diff
@@ -119,6 +121,10 @@ func (c completedConfig) New() (*ProjectCalicoServer, error) {
 	// implement our own syncer-based cache.
 	calicoLister := NewCalicoResourceLister(cc)
 
+	// Create the manager for policy watches. It keeps track of established watches and makes sure they are canceled
+	// in case of on update that the watch should contain
```
caseydavenport, line 125:
```suggestion
	// in case of on update that necessitates recreation of the watch. 
```

apiserver/pkg/registry/projectcalico/util/watch_manager.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// AddWatch adds a watch to our map and triggers monitoring for watch closure by the consumer or API server
+func (m *WatchManager) AddWatch(record WatchRecord) {
+	logrus.WithFields(logrus.Fields{"id": record.ID, "Kind": record.Kind}).Debug("Adding WatchRecord")
+	m.watchRecords[record.ID] = record
+	go m.monitorWatch(record)
+}
+
+// minitorWatch waits to see if the context is done signaling that the watch has been ended. We can remove the watch from our map
```
caseydavenport, line 90:
```suggestion
// monitorWatch waits to see if the context is done signaling that the watch has been ended. We can remove the watch from our map
```

apiserver/pkg/registry/projectcalico/util/watch_manager.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+			m.watchRecords = map[string]WatchRecord{}
+		}
+	}
+}
+
+// AddWatch adds a watch to our map and triggers monitoring for watch closure by the consumer or API server
+func (m *WatchManager) AddWatch(record WatchRecord) {
+	logrus.WithFields(logrus.Fields{"id": record.ID, "Kind": record.Kind}).Debug("Adding WatchRecord")
+	m.watchRecords[record.ID] = record
```
caseydavenport, line 86: You need to wrap `watchRecords` with the mutex here as well. Anywhere we access the map should be protected.

apiserver/pkg/registry/projectcalico/util/watch_manager.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (m *WatchManager) OnUpdates(updates []api.Update) {
+	for _, u := range updates {
+		// We do not need to cancel watches for delete as the original watch is still valid with other records
+		if u.Key.(model.ResourceKey).Kind == v3.KindTier && u.KVPair.Value != nil {
+			// New Tier added, we need to stop all watches in case there is a user that can watch policies in the new Tier
+			// When the watch is re-established it will contain policies in the newly created Tier
+			logrus.WithField("Tier", u.Key.(model.ResourceKey).Name).Debug("New Tier added, removing all WatchRecords")
+			for _, record := range m.watchRecords {
```
caseydavenport, line 74: You'll need a mutex call around this access to watchRecords as well.

caseydavenport (changes requested):
One pattern I like to use for mutexing around a map is to force all access of the map via a wrapper structure. Helps avoid missing cases.

For example:

```go
type mutexedMap struct {
  sync.Mutex
  m map[string]string
}

func (m *mutexedMap) Set(k, v string) {
  m.Lock()
  defer m.Unlock()
  m.m[k] = v
}
```

And similar functions for `Get`, `Remove`, `Iter`

caseydavenport (approved):

PR #10350, opened 2025-05-02
Goldmane Stream refactor

caseydavenport:
## Description

The main change in this PR is to adjust how data is send through streams. Previously, Goldmane would iterate each Flow and push it to the stream manager. This was exceptionally inefficient at scale, and resulted in long periods of time where the main loop was occupied feeding streams.

This PR adjusts that implementation such that Goldmane sends a pointer to the bucket to be streamed, allowing each stream to iterate the bucket at its own pace, on its own goroutine. This leaves moves congestion off of the main loop, allowing it to do more important things like ingest flows.

This does introduce a RWMutex into each bucket to prevent concurrent access. This does have some potential to block ingestion still, but is much less likely to happen because we only stream flows after some time has passed, so new flows should very rarely be indexed into a bucket that is actively being streamed. 

## Release Note

```release-note
Improve Goldmane flow streaming efficiency
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/goldmane/Makefile
+++ b/goldmane/Makefile
@@ -73,6 +73,16 @@ gen-mocks:
 	$(DOCKER_RUN) $(CALICO_BUILD) sh -c 'mockery'
 
 ###############################################################################
+# Run load generator against a real cluster using your KUBECONFIG env var.
+###############################################################################
+load:
+	kubectl get secret -n calico-system node-certs --template='{{index .data "tls.key"}}' | base64 -d > tls.key
+	kubectl get secret -n calico-system node-certs --template='{{index .data "tls.crt"}}' | base64 -d > tls.crt
+	kubectl get secret -n calico-system goldmane-key-pair --template='{{index .data "tls.crt"}}' | base64 -d > ca.crt
+	kubectl port-forward -n calico-system svc/goldmane 7443:7443 &
+	SERVER=goldmane:7443 CLIENT_CERT=tls.crt CLIENT_KEY=tls.key CA_FILE=ca.crt go run ./cmd/flowgen/main.go
+
+###############################################################################
 # Release
 ###############################################################################
 ## Deploys images to registry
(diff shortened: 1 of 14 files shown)


goldmane/Makefile
```diff
@@ -72,6 +72,16 @@ ut:
 gen-mocks:
 	$(DOCKER_RUN) $(CALICO_BUILD) sh -c 'mockery'
 
+###############################################################################
```
caseydavenport, line 75: This is just cruft from testing - can remove this.

skoryk-oleksandr (commented):
Overall, these changes look good to me. My only comment is that, since `buckets` will now be accessed concurrently, it would be good to cover this with unit tests. I’d recommend covering the following cases:
- `bucket.AddFlow()` while `bucket.Iter()` is running.
- Call `.Reset()` on a bucket while it's being streamed (i.e., .`Iter()` is executing)
- Cancel the stream context while the stream is consuming a bucket.
- Two or more streams simultaneously process the same bucket.
- Push a bucket to a stream before calling `.markReady()`. Check that the .Iter() function skips unready buckets properly.

skoryk-oleksandr (approved):

maximusprime3367:
The recent Goldmane stream refactor significantly improves efficiency by optimizing how data is transmitted through streams. Previously, Goldmane processed each Flow sequentially, pushing it directly to the stream manager—a method that caused bottlenecks, especially at scale, and slowed down the main loop. The new update introduces a smarter approach: instead of handling each Flow individually, Goldmane now sends a pointer to the data bucket, allowing streams to process information independently on separate goroutines. This enhancement reduces congestion, freeing up the main loop for critical tasks like flow ingestion. For seamless streaming performance, check out [Yacine TV](https://www.yacinetvweb.com/), the premier platform for high-quality, uninterrupted live streams. Whether you're watching sports, movies, or TV shows, Yacine TV ensures smooth playback with optimized backend efficiency.

Issue #10338, opened 2025-05-02
Parsing value for FailsafeInboundHostPorts: "[]" (from environment variable) panic: runtime error: slice bounds out of range [:-1]


logabot:
Try to set `FailsafeInboundHostPorts` and `FailsafeOutboundHostPorts` to empty list via environment variables. 

## Expected Behavior
FailsafeInboundHostPorts: []
FailsafeOutboundHostPorts: []

## Current Behavior
Pod had env:
```yaml
    - name: FELIX_FAILSAFEINBOUNDHOSTPORTS
       value: '[]'
    - name: FELIX_FAILSAFEOUTBOUNDHOSTPORTS
       value: '[]'
```
or 
```yaml
    - name: FELIX_FAILSAFEINBOUNDHOSTPORTS
       value: "'[]'"
    - name: FELIX_FAILSAFEOUTBOUNDHOSTPORTS
       value: "'[]'"
```
get error
```
2025-05-01 16:22:38.160 [INFO][25] tunnel-ip-allocator/config_params.go 695: Parsing value for FailsafeInboundHostPorts: [] (from environment variable)
panic: runtime error: slice bounds out of range [:-1]
```

## Steps to Reproduce (for bugs)
Kubespray:
1. Set 
  ```yaml
  calico_node_extra_envs:
    FELIX_FAILSAFEINBOUNDHOSTPORTS: "[]"
    FELIX_FAILSAFEOUTBOUNDHOSTPORTS: "[]"
  ```
2. deploy cluster

## Context
I try to disable all default port after cluster setup and globalNetworkPolicies applied 

## Your Environment
* Calico version 3.29.1
* Calico dataplane (iptables, windows etc.) iptables
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes via kubespray
* Operating System and version: Ubuntu 20.04

sridhartigera:
@logabot Given that ```FailsafeInboundHostPorts``` is a list, ```[]``` is a valid value when set in the felixconfiguration resource. When using ENVVAR, the value is a string and ```""``` to indicate an empty list.

caseydavenport:
Yep, I think this is just a documentation being wrong issue

skoryk-oleksandr:
@logabot Let us know if using this syntaxt helped to fix the problem 
```
calico_node_extra_envs:
  FELIX_FAILSAFEINBOUNDHOSTPORTS: ""
[...]

PR #10354, opened 2025-05-05
Fix panic when Failsafe configs are set to empty list

sridhartigera:
## Description

FailSafeInBoundPorts and FailSafeOutBoundPorts can be set to an empty list ```"[]"```. This results in felix panicking. 
This PR fixes https://github.com/projectcalico/calico/issues/10338

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/config/param_types_test.go
+++ b/felix/config/param_types_test.go
@@ -61,6 +61,27 @@ var _ = DescribeTable("CIDR list parameter parsing",
 	Entry("Mix of IP and CIDRs", "1.1.1.1/24, 2.2.2.2", []string{"1.1.1.0/24", "2.2.2.2/32"}, true),
 )
 
+var _ = DescribeTable("Port list parameter parsing",
+	func(raw string, expected []config.ProtoPort, expectSuccess bool) {
+		p := config.PortListParam{config.Metadata{
+			Name: "PortLists",
+		}}
+		actual, err := p.Parse(raw)
+		if expectSuccess {
+			Expect(err).To(BeNil())
+			Expect(actual).To(Equal(expected))
+		} else {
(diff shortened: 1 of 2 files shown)


felix/config/param_types_test.go
```diff
@@ -61,6 +61,27 @@ var _ = DescribeTable("CIDR list parameter parsing",
...
+		}}
+		actual, err := p.Parse(raw)
+		if expectSuccess {
+			Expect(err).To(BeNil())
+			Expect(actual).To(Equal(expected))
+		} else {
+			Expect(err).To(BeNil())
+		}
+	},
+	Entry("Empty", "", nil, true),
```
caseydavenport, line 77: Interesting, I would have expected an empty string to be set the default - but I guess we have logic to detect "unset" vs "set but empty"? 

If that's the case, I wonder if the right fix is just in documentation to set `""` instead of `"[]"` - WDYT?

caseydavenport: It doesn't really, based on the env var syntax. 

`tcp:22,udp:1.1.1.1:22,3033` - this is a valid value, and it has no `[` or `]` in it at all!

It's a valid YAML representation, but this isn't YAML we're talking about.

sridhartigera: Better to fix the docs.

sridhartigera: It makes sense to set the value as ```[]``` in the felix config but when configured as an env var it must be set to "".

coutinhop (approved):

PR #10356, opened 2025-05-05
Move vxlan and ipip manager duplicate code into route manager

mazdakn:
## Description

This PR refactors vxlan and ipip managers by moving their duplicate logic into an embedded router manager which is responsible for:
- All sort of routes
- Configuring tunnel devices

IPIP manager UT is also cleaned up. Merging similar UT/FV tests are intentionally left out of this PR to make it easier for review. Follow up PRs will improves those tests.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/route_mgr.go
```diff
@@ -0,0 +1,686 @@
...
+	}
+
+	if m.maintainIPOnly {
+		if err := m.ensureAddressOnLink(addr, link); err != nil {
+			return fmt.Errorf("failed to ensure address of interface: %s", err)
+		}
+		return nil
+	}
+
+	if m.vxlanEnabled() {
```
tomastigera, line 584: why did you put this whole "keep device i sync" here as well? It seems like there quite a bit of a difference betwee the two tunnels. With this it also feel more like a tunnelBaseMgr then just route mgr. It should probably be called tunnelRouteMgr as it ia quite a specific type of route manager

mazdakn: Actually there is not much difference. The Only difference is in the link itself, otherwise all other operation related to making sure the device is created, and configured is the same.

mazdakn: In the end, route management is tightly integrated with interface management. I did another prototype to make them separate, but the result was not clean. Both ipip and vxlan managers have a keep in sync and tunnel configuration methods with around 90% code duplicate. This is the reason, I moved it into this manager.

This check is the only place that we have to check specifically for vxlan. Though this one can also be improved to be generic.

We can definitely change the names as we think appropriate.

mazdakn: btw, I initially named this manager `tunnelManager`, but then changed it to `routeManager` to use it later for unencap routing.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -680,17 +680,23 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
 		)
-		dp.vxlanParentC = make(chan string, 1)
+		dp.parentInterfaceC = make(chan string, 1)
 		vxlanMTU := config.VXLANMTU
 		if config.BPFEnabled && bpfutils.BTFEnabled {
 			vxlanMTU = 0
 		}
-		go dp.vxlanManager.KeepVXLANDeviceInSync(context.Background(), vxlanMTU, dataplaneFeatures.ChecksumOffloadBroken, 10*time.Second, dp.vxlanParentC)
+		go dp.vxlanManager.KeepVXLANDeviceInSync(
+			context.Background(),
```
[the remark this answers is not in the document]
caseydavenport: Good to keep IMO - in case we need to signal goroutine shutdown, etc. in the future.

felix/dataplane/linux/route_mgr.go
```diff
@@ -0,0 +1,686 @@
...
+	isBlock = isType(msg, proto.RouteType_REMOTE_WORKLOAD)
+
+	if isRemoteTunnel && msg.Borrowed {
+		// If we receive a route for a borrowed VXLAN tunnel IP, we need to make sure to program a route for it as it
+		// won't be covered by the block route.
+		return true
+	}
+	if isRemoteTunnel && isBlock {
+		// This happens when tunnel addresses are selected from an IP pool with blocks of a single address.
+		// These also need routes of the form "<IP> dev vxlan.calico" rather than "<block> via <VTEP>".
```
caseydavenport, line 149:
```suggestion
		// These also need routes of the form "<IP> dev vxlan.calico" rather than "<block> via <TunnelEndpoint>".
```

felix/dataplane/linux/ipip_mgr.go
```diff
@@ -86,206 +70,126 @@ type ipipManager struct {
...
+	m.routeMgr.setTunnelRouteFunc(m.route)
+}
+
+func (m *ipipManager) OnUpdate(protoBufMsg interface{}) {
+	switch msg := protoBufMsg.(type) {
 	case *proto.HostMetadataUpdate:
 		m.logCtx.WithField("hostname", msg.Hostname).Debug("Host update/create")
 		if msg.Hostname == m.hostname {
-			m.setLocalHostAddr(msg.Ipv4Addr)
+			m.routeMgr.updateDataIfaceAddr(msg.Ipv4Addr)
```
caseydavenport, line 157: `routeMgr` has its own `OnUpdate` method - any reason not to just pass all messages directly down by calling `m.routeMgr.OnUpdate(protoBufMsg)`?

felix/dataplane/linux/route_mgr.go
```diff
@@ -0,0 +1,686 @@
...
+	return msg.Types&t == t
+}
+
+func (m *routeManager) routeIsLocalBlock(msg *proto.RouteUpdate) bool {
+	// RouteType_LOCAL_WORKLOAD means "local IPAM block _or_ /32 of workload" in IPv4.
+	// It means "local IPAM block _or_ /128 of workload" in IPv6.
+	if !isType(msg, proto.RouteType_LOCAL_WORKLOAD) {
+		return false
+	}
+	// Only care about a specific ippool
```
caseydavenport, line 244:
```suggestion
	// Only care about IPPools which match the encap type of the manager.
```

felix/dataplane/linux/route_mgr.go
```diff
@@ -0,0 +1,686 @@
...
+			if addr.IPNet.IP.String() == dataAddr {
+				m.logCtx.Debugf("Found data interface: %s", link)
+				return link, nil
+			}
+		}
+	}
+	return nil, fmt.Errorf("Unable to find data interface with address %s", dataAddr)
+}
+
+// checks that it is still correctly configured.
```
caseydavenport, line 459:
```suggestion
// KeepDeviceInSync runs in a loop and checks that the device is still correctly configured, and updates it if necessary.
```

felix/dataplane/linux/int_dataplane.go
```diff
@@ -2140,12 +2153,12 @@ func (d *InternalDataplane) loopUpdatingDataplane() {
...
 			d.onIfaceMonitorMessage(ifaceUpdate)
-		case name := <-d.vxlanParentC:
-			d.vxlanManager.OnParentNameUpdate(name)
-		case name := <-d.vxlanParentCV6:
-			d.vxlanManagerV6.OnParentNameUpdate(name)
-		case name := <-d.noEncapDeviceC:
-			d.ipipManager.OnNoEncapDeviceUpdate(name)
+		case name := <-d.dataInterfaceC:
+			d.ipipManager.routeMgr.OnDataDeviceUpdate(name)
+		case name := <-d.parentInterfaceC:
```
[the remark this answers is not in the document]
mazdakn: I use `dataInterface` as generic term, and `parentInterface` only for vxlan manager. Given that we used `parentInterface` for vxlan, using the same term for ipipManger caused confusion. For that I tried to find a generic term. If we think `parentInterface` is generic and more clear, I'll re-use it for all these 3 managers.

felix/dataplane/linux/ipip_mgr.go
```diff
@@ -86,206 +70,126 @@ type ipipManager struct {
...
+	mtu int,
 	dpConfig Config,
 	opRecorder logutils.OpRecorder,
 	nlHandle netlinkHandle,
-	ipVersion uint8,
 ) *ipipManager {
+
 	if ipVersion != 4 {
-		logrus.Errorf("IPIP manager only supports IPv4")
+		logrus.Errorf("IPIP manager is only supports in IPv4")
```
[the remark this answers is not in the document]
mazdakn: Reverted the change.

felix/dataplane/linux/ipip_mgr.go
```diff
@@ -86,206 +70,126 @@ type ipipManager struct {
...
+		tunnelDevice:       tunnelDevice,
+		tunnelDeviceMTU:    mtu,
+		ipVersion:          ipVersion,
+		externalNodeCIDRs:  dpConfig.ExternalNodesCidrs,
+		ipSetDirty:         true,
+		dpConfig:           dpConfig,
+		routeProtocol:      calculateRouteProtocol(dpConfig),
+		logCtx: logrus.WithFields(logrus.Fields{
+			"ipVersion":     ipVersion,
+			"tunnel device": tunnelDevice,
```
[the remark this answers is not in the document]
mazdakn: Fixed.

felix/dataplane/linux/ipip_mgr.go
```diff
@@ -86,206 +70,126 @@ type ipipManager struct {
...
+			m.routeMgr.OnUpdate(msg)
+		}
 	}
 }
 
-func (m *ipipManager) getLocalHostAddr() string {
-	m.myAddrLock.Lock()
-	defer m.myAddrLock.Unlock()
-	return m.hostAddr
+func (m *ipipManager) mayUpdateRoutes() {
```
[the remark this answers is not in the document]
mazdakn: Fixed.

felix/dataplane/linux/ipip_mgr.go
mazdakn: We can definitely do that. There are 2 reason I did no go that path:
1 - I hoped to process updates only once.
2 - If we want to send `HostMetadataUpadate` down to route manager to let it find its own data interface, then we should do that same thing for vxlan manager by sending `VTEP` updates as well. Then to not mix up, we need to make sure each update is processed by correct version of route manager by checking ip pool.
This can only be done by adding multiple checks if vxlan or ipip tunnel is enabled, making the code less readable.

Does this make sense?

caseydavenport: Yeah, that sounds reasonable to me - thanks!

felix/dataplane/linux/int_dataplane.go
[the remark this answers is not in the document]
caseydavenport: I think either is fine really, just found it non-obvious why we had different names for the same "concept" in different encap modes.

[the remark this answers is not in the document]
mazdakn: Make sense to use one term. I changed all `data interface` to `parent interface'

felix/dataplane/linux/ipip_mgr.go
```diff
@@ -86,206 +70,123 @@ type ipipManager struct {
...
+			tunnelDevice,
+			ipVersion,
+			mtu,
+			dpConfig,
+			opRecorder,
+			nlHandle,
+		),
+	}
+
+	m.routeMgr.routeClassTunnel = routetable.RouteClassIPIPTunnel
```
caseydavenport, line 141: I think I'm still not sure why these aren't arguments to `newRouteManager` - presumably these are required fields, right?

caseydavenport (approved):
One comment but overall LGTM

felix/dataplane/linux/ipip_mgr.go
mazdakn: I'll address this in the next PR.

PR #10370, opened 2025-05-07
Add OCP-required labels to new v3.30+ images [master]

danudey:


third_party/envoy-gateway/Dockerfile
```diff
@@ -22,5 +22,14 @@ COPY --from=source --chown=65532:65532 /var/lib /var/lib
 
 USER 65532:65532
 
+# These labels are required for OCP Certification
+LABEL description="Envoy Gateway is an implementation of the Kubernetes Gateway API using the envoy-gateway project"
```
nelljerram, line 26: Strictly speaking I'd say "part of" the implementation, rather than the whole thing.

Another approach would be "This image is a build of the github.com/envoyproxy/gateway code".  Factual if uninsightful!

third_party/envoy-gateway/Dockerfile
```diff
@@ -22,5 +22,14 @@ COPY --from=source --chown=65532:65532 /var/lib /var/lib
...
+LABEL maintainer="maintainers@tigera.io"
+LABEL name="Envoy Gateway"
+LABEL release="1"
+LABEL summary="Envoy Gateway is an implementation of the Kubernetes Gateway API using the envoy-gateway project"
```
nelljerram, line 30: Presume you'll do the same here as for description.

third_party/envoy-gateway/Dockerfile
```diff
@@ -22,5 +22,14 @@ COPY --from=source --chown=65532:65532 /var/lib /var/lib
...
+LABEL vendor="Project Calico"
```
nelljerram, line 31: Personally I wouldn't say that "Project Calico" is a vendor.  Feels like "Tigera" would be more honest.  But if we already have an established precedent for "Project Calico", fine.

third_party/envoy-proxy/Dockerfile
```diff
@@ -44,6 +44,15 @@ FROM ${CALICO_BASE}
 
 COPY --from=source / /
 
+# These labels are required for OCP Certification
+LABEL description="Envoy Proxy is WHATEVER ENVOY PROXY DOES"
```
nelljerram, line 48: Could be "This image is a build of the github.com/envoyproxy/envoy code"

third_party/envoy-ratelimit/Dockerfile
```diff
@@ -7,3 +7,12 @@ FROM ${CALICO_BASE}
 ARG TARGETARCH
 
 COPY bin/envoy-ratelimit-${TARGETARCH} /bin/ratelimit
+
+# These labels are required for OCP Certification
+LABEL description="Envoy Ratelimit is WHATEVER ENVOY RATELIMIT DOES"
```
nelljerram, line 12: Could be "This image is a build of the github.com/envoyproxy/ratelimit code"

third_party/envoy-gateway/Dockerfile
danudey: I agree that "Tigera" makes more sense, but we do already use "Project Calico" in all of our other images so I'll leaving it as-is for now.

radTuti (approved):
we should probably add OCI labels to the envoy images as well

PR #10380, opened 2025-05-08
CORE-11022: Implementing TLS Cipher configurable for API Server

lucastigera:
## Description

This PR configures the API Server to read TLS ciphers from an environment variable. If the variable is not set, a default list of PFS (Perfect Forward Secrecy) ciphers will be used.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TLS ciphers are now configurable. This allows greater flexibility for meeting compliance requirements or supporting specific client configurations. If no custom list is provided, a secure default set of PFS ciphers is used.
```

## Reminder for the reviewer

Other optional labels:


crypto/pkg/tls/tls.go
```diff
@@ -46,6 +47,63 @@ var tls13Ciphers = []uint16{
...
+	"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256":       tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+	"TLS_RSA_WITH_AES_256_GCM_SHA384":               tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
+	"TLS_RSA_WITH_AES_128_GCM_SHA256":               tls.TLS_RSA_WITH_AES_128_GCM_SHA256,
+	"TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA":          tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
+	"TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA":            tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
+	"TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA":            tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
+}
+
+// GetTLSCipherDefaults returns the ciphers supported by TLS 1.3 and the PFS ciphers supported by TLS 1.2.
+func GetTLSCipherDefaults() []uint16 {
```
caseydavenport, line 71: It's a bit unusual (though not strictly avoided) to see `GetX` function naming in Go - I'd probably just call this `DefaultCiphers`

caseydavenport, line 71: You might just want to do this?

```go
var DefaultCiphers = append(tls13Ciphers, tls12Ciphers...)
```

And avoid the need for a function altogether.

crypto/pkg/tls/tls.go
```diff
@@ -46,6 +47,63 @@ var tls13Ciphers = []uint16{
 	tls.TLS_AES_128_GCM_SHA256,
 }
 
+var supportedCiphers = map[string]uint16{
```
caseydavenport, line 50: I'd suggest getting rid of this map and building it on demand in `ParseTLSCiphers` from the above slices, as needed.

It means we don't need to keep multiple lists up-to-date, and also allows GC to cleanup the map when we're done with it.

caseydavenport (commented):
A few thoughts

crypto/pkg/tls/tls.go
lucastigera: I was planning to remove those two slices in the next iteration. =)
I'll propose a new approach in the next commit.

crypto/pkg/tls/tls.go
lucastigera: Good point!
What if we keep the map and build the array on demand in `DefaultCiphers()`?
I’m thinking on a map structure with two new fields: `default` and `order`.
This way, we can identify the default ciphers by iterating over the map and ensure the correct order of the default list - since we can’t rely on map ordering, and it's valuable to keep the strongest ciphers first in the list.

I'll put a suggestion in the next commit.
Please let me know if you think it's overengineering or making things more complex than they need to be.
If so, I’m happy to move forward with your suggestion as well. =)

caseydavenport: That feels needlessly complicated to me - we have the slices, which already implicitly define an ordering. So why not just use them?

caseydavenport: Yeah, we don't really have to worry about performance for this. The simpler the better.

rene-dekker:
I was wondering if we just want the tls.NewTLSConfig function to automatically check for env variables to be set, such that we don't need to change the function signature.

lucastigera:
> I was wondering if we just want the tls.NewTLSConfig function to automatically check for env variables to be set, such that we don't need to change the function signature.

Hey @rene-dekker ! That's a good point - I did consider it, but I think checking environment variables directly within a library function can be a code smell. In general, I believe libraries should be environment-agnostic and rely on explicit parameters instead of side effects like env vars. This approach makes the code more testable, predictable, and reusable.

However, I’m open to discussing it further if you think this case warrants an exception. =)

crypto/pkg/tls/tls.go
```diff
@@ -46,22 +45,91 @@ var tls13Ciphers = []uint16{
...
 	}
 }
 
-// NewMutualTLSConfig generates a tls.Config configured to enable mTLS with clients using the provided cert, key, and CA file paths.
-// If any of the files cannot be read, an error is returned.
-func NewMutualTLSConfig(cert, key, ca string) (*tls.Config, error) {
+// NewTLSConfigFromString returns a tls.Config with the recommended default settings for Calico components, using TLS cipher suite names provided as a comma-separated string.
+// The names are parsed into cipher suite IDs, and if parsing fails, an error is returned.
+// Based on build flags, boringCrypto may be used and fips strict mode may be enforced, which can override the parameters defined in this func.
+func NewTLSConfigFromString(cipherSuites string) (*tls.Config, error) {
```
caseydavenport, line 115:
```suggestion
func NewTLSConfigFromCipherString(cipherSuites string) (*tls.Config, error) {
```

We don't support setting other fields from string, so better to be explicit.

libcalico-go/lib/apiconfig/apiconfig.go
```diff
@@ -48,13 +48,14 @@ type CalicoAPIConfigSpec struct {
...
-	EtcdCertFile     string `json:"etcdCertFile" envconfig:"ETCD_CERT_FILE"`
-	EtcdCACertFile   string `json:"etcdCACertFile" envconfig:"ETCD_CA_CERT_FILE"`
+	EtcdEndpoints       string `json:"etcdEndpoints" envconfig:"ETCD_ENDPOINTS"`
+	EtcdDiscoverySrv    string `json:"etcdDiscoverySrv" envconfig:"ETCD_DISCOVERY_SRV"`
+	EtcdUsername        string `json:"etcdUsername" envconfig:"ETCD_USERNAME"`
+	EtcdPassword        string `json:"etcdPassword" envconfig:"ETCD_PASSWORD"`
+	EtcdKeyFile         string `json:"etcdKeyFile" envconfig:"ETCD_KEY_FILE"`
+	EtcdCertFile        string `json:"etcdCertFile" envconfig:"ETCD_CERT_FILE"`
+	EtcdCACertFile      string `json:"etcdCACertFile" envconfig:"ETCD_CA_CERT_FILE"`
+	EtcdTLSCipherSuites string `json:"etcdTLSCipherSuites" envconfig:"ETCD_TLS_CIPHER_SUITES"`
```
caseydavenport, line 58: libcalico is imported into all of our other components, so likely having this will mean we need both `TLS_CIPHER_SUITES` and `ETCD_TLS_CIPHER_SUITES` specified in each component.

Can we just drive this off the same environment variable - `TLS_CIPHER_SUITES` - for everything?

libcalico-go/lib/backend/etcdv3/etcdv3.go
```diff
@@ -115,7 +115,12 @@ func NewEtcdV3Client(config *apiconfig.EtcdConfig) (api.Client, error) {
 		return nil, fmt.Errorf("could not initialize etcdv3 client: %+v", err)
 	}
 
-	baseTLSConfig := calicotls.NewTLSConfig()
+	calicoTlsConfig, err := calicotls.NewTLSConfigFromString(config.EtcdTLSCipherSuites)
```
caseydavenport, line 118:
```suggestion
	calicoTLSConfig, err := calicotls.NewTLSConfigFromString(config.EtcdTLSCipherSuites)
```

libcalico-go/lib/backend/etcdv3/etcdv3.go
```diff
@@ -115,7 +115,12 @@ func NewEtcdV3Client(config *apiconfig.EtcdConfig) (api.Client, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create TLS config: %s", err)
+	}
+
+	baseTLSConfig := calicoTlsConfig
```
caseydavenport, line 123: Do we need this rename / copying of the variable?

typha/pkg/syncclient/sync_client.go
```diff
@@ -281,7 +283,10 @@ func (s *SyncerClient) connect(cxt context.Context, typhaAddr discovery.Typha) e
 			log.WithError(err).Error("Failed to load certificate and key")
 			return err
 		}
-		tlsConfig := calicotls.NewTLSConfig()
+		tlsConfig, err := calicotls.NewTLSConfigFromString(s.options.TLSCipherSuites)
+		if err != nil {
+			log.WithError(err).Panic("Fail to create TLS config: %w.", err)
```
caseydavenport, line 288: Looks like this function returns an error, so probably better to return an error instead of Panic

caseydavenport (commented):
I think it's fine to leave as individual config options rather than have the library read the env var if you'd like.

I do think we should use the same environment variable everywhere though.

libcalico-go/lib/backend/etcdv3/etcdv3.go
lucastigera: Oops! Definitely a code smell. Thanks for catching that.

typha/pkg/syncclient/sync_client.go
lucastigera: Nice catch!

lucastigera:
> I was wondering if we just want the tls.NewTLSConfig function to automatically check for env variables to be set, such that we don't need to change the function signature.

Hey @rene-dekker !
I just wanted to let you know that after refactoring the code to address Casey's comments and discussing it with him, reading the env var inside tls.NewTLSConfig() became a better option than changing the function signature. =)
I've updated the code to reflect this approach.
Thanks for your input on that!

caseydavenport (approved):

PR #10399, opened 2025-05-12
default policy name to not contain default tier prefix

MichalFupso:
## Description
Update the way we return policy names, if there is no annotation on the underlying CRD we return the name of policy in the default Tier without the default. prefix

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/k8s/resources/resources.go
+++ b/libcalico-go/lib/backend/k8s/resources/resources.go
@@ -327,8 +327,24 @@ func ConvertK8sResourceToCalicoResource(res Resource) error {
 	// so that they do not get overwritten.
 	if meta.Name == "" {
 		// We store the original v3 Name in our annotation when writing NetworkPolicy and GNP objects. If that's present, use it.
+		// If the kind is Policy and the annotation is not present, we remove the default. prefix to be compatible with previous Calico versions
 		// Otherwise, fill in the name from the underlying CRD.
-		meta.Name = rom.GetName()
+		switch res.GetObjectKind().GroupVersionKind().Kind {
+		case apiv3.KindGlobalNetworkPolicy:
+			policy := res.(*apiv3.GlobalNetworkPolicy)
+			meta.Name = determinePolicyName(rom.GetName(), policy.Spec.Tier)
+		case apiv3.KindNetworkPolicy:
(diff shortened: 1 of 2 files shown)


libcalico-go/lib/backend/k8s/resources/resources.go
```diff
@@ -365,3 +381,12 @@ func watchOptionsToK8sListOptions(wo api.WatchOptions) metav1.ListOptions {
 		AllowWatchBookmarks: wo.AllowWatchBookmarks,
 	}
 }
+
+func determinePolicyName(name, tier string) string {
```
MichalFupso, line 385: Can't add this to names/policy due to cycle import, maybe there is a better place to add it to not duplicate code

libcalico-go/test/mock-policies.yaml
```diff
@@ -0,0 +1,62 @@
+#These policies are used for testing if the correct policy name is returned when no name is part of the projectcalico.org/metadata annotation
```
caseydavenport, line 1: I'd really like to avoid checking in a YAML file as part of the tests - we should be aiming for UTs to not rely on external configuration ( I know this isn't already the case, but let's not make it any worse! )

We don't need to block this PR on the test - we can submit a follow-on to add it. I have ideas on how to do this without needing externally applied policies.

libcalico-go/lib/backend/model/keys.go
```diff
@@ -722,3 +730,12 @@ func SerializeValue(d *KVPair) ([]byte, error) {
 	}
 	return json.Marshal(d.Value)
 }
+
+func determinePolicyName(name, tier string) string {
```
caseydavenport, line 734: Should this be part of `parseMetadataAnnotation`? It seems like it must always be called at the same time as that function, so having them separately seems error prone?

libcalico-go/lib/backend/k8s/resources/resources.go
caseydavenport: We can just add a new package? Maybe `lib/backend/tiers` or something? Not the end of the world though.

Issue #8548, opened 2024-02-21
Over 30,000 leaked ipamhandles.crd.projectcalico.org objects not getting cleaned up by calico-kube-controller


bradbehle:
## Expected Behavior
calico-kube-controller should clean up leaked pod IPs / IPAMHandles

## Current Behavior
In at least this cluster we are looking at, calico-kube-controller is not cleaning these up

## Possible Solution
Figure out why calico-kube-controller doesn't seem to notice these 30,000 leaked IPAM handles, but `calicoctl ipam check` does (see details below

## Steps to Reproduce (for bugs)
1. Create an Openshift 4.6 cluster with Calico v3.17.2 (deployed via Operator v1.13.4) about 2.5 years ago
2. Use that cluster for those 2.5 years, upgrading Openshift and Calico
3. Get to today, running Openshift 4.12 and Calico 3.26.4, and notice 30,000+ leaked IPAM Handles
4. Wonder why calico-kube-controller isn't cleaning them up

## Context
We don't think this is causing any noticeable problems with the cluster at the moment, but the etcd performance can't be helped by all these leaked CRDs.  We could probably run `calicoctl ipam release` and clean these up.  But the concern is that this is probably also a problem on a bunch of old clusters we maintain, and eventually could become a problem, so we were hoping if we provide information about this cluster, someone could determine why calico-kube-controllers isn't cleaning these up, and fix it in an upcoming Calico release.

## Your Environment

* Calico version: Started at 3.17.2 (Operator v1.13.4), has been upgraded many times, now is at Calico 3.26.4
* Orchestrator version (e.g. kubernetes, mesos, rkt): Started at Openshift 4.6, now is at Openshift 4.12
* Operating System and version: Uses RHEL 8.9 worker nodes
* Link to your project (optional):

Here's the information that shows all the leaked IPAM handles (with IP addresses obscured):

```
+----------+----------------+-----------+------------+-------------+
| GROUPING |       CIDR     | IPS TOTAL | IPS IN USE |  IPS FREE   |
+----------+----------------+-----------+------------+-------------+
| IP Pool  | X.X.0.0/16     |     65536 | 537 (1%)   | 64999 (99%) |
| Block    | X.X.128.128/26 |        64 | 26 (41%)   | 38 (59%)    |
| Block    | X.X.128.64/26  |        64 | 1 (2%)     | 63 (98%)    |
| Block    | X.X.162.192/26 |        64 | 63 (98%)   | 1 (2%)      |
[...]

Issue #7508, opened 2023-03-28
IPAM GC not fast enough, leading to memory bloat


philnielsen:
## Expected Behavior

IPAM blocks for deleted nodes get cleaned up

## Current Behavior

we upgraded our cluster from 1.21->through to 1.24 (which included an upgrade to calico from 3.19.3 and now we are seeing that IPAM blocks are not getting cleaned up
ipamblock store gets larger and larger leading to OOM.
spam of the following in the logs:
```Calico Node referenced in IPAM data does not exist error=resource does not exist: Node(i-xxxxxxxxxxx) with error: nodes "i-xxxxxxxxxxx" not found```

## Possible Solution

I think our issue is related to #7433 and #7503 but it looks like the plan is only to cherry pick to 3.25?

## Steps to Reproduce (for bugs)

1. have a medium/large cluster with lots of autoscaling of nodes/pods
2. tail calico node controller logs and see a the spam above
3. eventually watch controller pod oom without any cleanup happening

## Context

Things never get cleaned up I think this will get worse and worse. Also API server instability because the reconciliation loop constantly hits it.

## Your Environment

* Calico version: v3.23.5 via kops
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes 1.24 via kops
* Operating System and version: Flatcar-stable-3374.2.4-hvm
* Link to your project (optional):

popsikle:
I think this comment is quite apt for this issue- https://github.com/projectcalico/calico/blob/f3a8ef66832a99b07cc39fd20f4fc2f44abec109/kube-controllers/pkg/controllers/node/ipam.go#L721

Having a MaxNodesPerCheckAllocations or something would help here because the number of allocations grows and grows, and there is no way to purge the allocation because the process oom before it gets through all the nodes.

coutinhop:
@philnielsen would you have any issues with upgrading to v3.25? Pinging @mgleung to see if it makes sense to cherry-pick https://github.com/projectcalico/calico/pull/7433 into other branches...

philnielsen:
Since we use kops (which for k8s 1.24 only goes up to [calico 3.23.5](https://kops.sigs.k8s.io/releases/1.24-notes/)) to manage our clusters as well as our calico install, this requires an upgrade of our cluster to k8s 1.25, which we are in the process of planning, but isn't something we can do quickly. 

We tried to manually update calico to 3.25 but ran into some rbac permission issues so we rolled that back.

[...]

PR #10428, opened 2025-05-18
feat: Add IPAM block cleanup for removed nodes

Kevinz857:
This adds a new CleanupBlocksForRemovedNodes method to the IPAM interface and implements a new CLI command 'calicoctl ipam clean orphaned-blocks' to help administrators clean up IPAM blocks for nodes that no longer exist.

The implementation includes:
- New interface method in libcalico-go/lib/ipam/interface.go
- Implementation in libcalico-go/lib/ipam/cleanup.go
- CLI command in calicoctl/calicoctl/commands/ipam/clean.go
- Test support in kube-controllers/pkg/controllers/node/fake_ipam_cleaner.go

This resolves the issue of slow IPAM block release when nodes are removed from the cluster, providing a built-in alternative to external cleanup scripts.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/calicoctl/calicoctl/commands/ipam.go
+++ b/calicoctl/calicoctl/commands/ipam.go
@@ -32,6 +32,7 @@ func IPAM(args []string) error {
   <BINARY_NAME> ipam <command> [<args>...]
 
     check            Check the integrity of the IPAM datastructures.
+    clean            Clean up orphaned or unused IPAM resources.
     release          Release a Calico assigned IP address.
     show             Show details of a Calico configuration,
                      assigned IP address, or of overall IP usage.
@@ -70,6 +71,8 @@ Description:
 	switch command {
 	case "check":
 		return ipam.Check(args, buildinfo.Version)
+	case "clean":
+		return ipam.Clean(args, buildinfo.Version)
 	case "release":
 		return ipam.Release(args, buildinfo.Version)
 	case "show":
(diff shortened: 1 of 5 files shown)


caseydavenport:
@Kevinz857 thanks for the PR! I have a few questions I'd like to talk about before getting too deep into the review:

- We do already a flow meant to do similar operations to this - an `ipam check` command that allows you to build a report on released resources, and then `ipam release` that allows you to release leaked objects from the report. Did you consider enhancing that to meet your use-case (if it doesn't already)?
- Release of IPAM blocks when nodes are deleted is meant to happen automatically by Calico - presumably you are not seeing this behavior? If not, I would like to get to the root cause of that rather than rely on manual intervention.

caseydavenport:
> This resolves the issue of slow IPAM block release when nodes are removed from the cluster

Is there a particular GitHub issue you are referring to?

Kevinz857:
> > This resolves the issue of slow IPAM block release when nodes are removed from the cluster
> 
> Is there a particular GitHub issue you are referring to?

@caseydavenport about this issue https://github.com/projectcalico/calico/issues/8548 

We are a scenario where K8s nodes frequently scale. Here, we will find that the release speed of ipamblocks is very slow, which leads to insufficient ipamblocks. At this time, manual cleaning is often required

caseydavenport:
@Kevinmmt the linked issue is primarily about releasing handles automatically, which the `calicoctl ipam release --from-report` option can already do. 

I just did some work in this area that should substantially improve IPAM garbage collection performance and speed: https://github.com/projectcalico/calico/issues/7508#issuecomment-2877796206

Might be worth seeing if those changes fix your situation. If not, then I think I'd also want to understand why the `ipam release --from-report` option is the direction I would point you. I'd like to funnel manual IPAM GC through a singular process to avoid proliferation of different commands and tools that do similar but not quite the same thing.

Issue #6412, opened 2022-07-21
Discussion: crd.projectcalico.org/v1 vs projectcalico.org/v3


caseydavenport:
This issue comes up frequently enough that I think it warrants its own parent issue to explain and discuss. I'll try to keep this up-to-date with the latest thinking and status should it change. 

The problem generally manifests itself as one of the following:

1. no matches for kind "X" in version "projectcalico.org/v3" when attempting to apply a resource.
2. Applying a resource with `apiVersion: crd.projectcalico.org/v1` and Calico not behaving as expected.

## **TL;DR**

**Don't touch `crd.projectcalico.org/v1` resources.** They are not currently supported for end-users and the entire API group is only used internally within Calico. Using any API within that group means you will bypass API validation and defaulting, which is bad and can result in symptoms like # 2 above. You should use `projectcalico.org/v3` instead. Note that `projectcalico.org/v3` requires that you [install the Calico API server in your cluster](https://projectcalico.docs.tigera.io/maintenance/install-apiserver), and will result in errors similar to # 1 above if the Calico API server is not running. 

## **Ok, but why do it that way?**

Well, it's partly because of limitations in CRDs, and partly due to historical reasons. CRDs provide some validation out of the box on their own, but can't do some of the more complex cross-field and cross-object API validation that the Calico API server can perform. For example, making sure that IP pools are consistent with the IPAM block resources within the cluster is a complex validation process that just can't be expressed in an OpenAPI schema. Same goes for some of the defaulting operations (e.g., conditional defaulting based on other fields).

As a result, Calico uses an [aggregation API server](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) to perform these complex tasks against `projectcalico.org/v3` APIs, and stores the resulting validated and defaulted resources in the "backend" as CRDs within the `crd.projectcalico.org/v1` group. Prior to the introduction of said API server, all of that validation and defaulting had to be performed client-side via `calicoctl`, but data was still stored in the "backend" as CRDs, for Calico itself to consume.

CRD validation has come a long way since Calico [initially started using them](https://github.com/projectcalico/libcalico-go/commit/444f3e9ad99f087cde69b270c7d5937ca311e485) way back in beta when they were actually called ThirdPartyResources. However, they still don't (and probably won't ever) support the types of validation that Calico currently enforces via its API server.

## **Pain points**

Yes, this model is not perfect and has a few known (non-trivial) pain points that I would love to resolve.

[...]

PR #10447, opened 2025-05-21
Ability to use projectcalico.org/v3 custom resource definitions 

caseydavenport:
## Description

This is the first step in removing the Calico API server and instead using v3 APIs directly as CRDs. 

Why?

- Calico API server constant source of bugs and burden.
- Differences in CRD vs v3 API
- Importing k8s libraries
- Massive amount of code needed for new APIs
- hostNetwork bugs on EKS, AKS.
- Must be running before APIs are accessible, HA, etc.

Scope / included in this PR:

This PR is scoped to adding the new CRDs and getting components and tests capable of using them instead of the v1.crd.projectcalico.org CRDs. Further work should be built incrementally on top of this with smaller PRs. 

- [X] Generate new v3 CRDs in addition to the existing crd.pc.org/v1 CRDs
- [X] Add switch to libcalico-go to read v3 CRDs instead of v1
- [X] Skip / disable translations in CRD backend.
- [X] Add IP Pool deletion controller for dependent resources
- [x] Add short names to CRDs
- [X] Add new v3 CRDs for ipamhandles / ipamblocks

Work coming in follow-on PRs:

- [ ] Add extensive kubebuilder validation + additional printer columns
- [ ] IP pool controller productized
- [ ] Add tier-based RBAC using webhook
- [ ] Update CI to run in both modes.

## Related issues/PRs

Tracking other issues I hit along the way!

- https://github.com/kubernetes/kubernetes/issues/97423

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Tech-preview support for using the projectcalico.org/v3 API directly as CRDs, without the need for an extension API server.
```

```release-note
The tigera-operator helm chart no longer includes crd.projectcalico.org CRDs. They must be created / updated prior to running the chart.
```

## Reminder for the reviewer

Other optional labels:


liuxu623:
Will etcd backend be completely removed?

nelljerram:
> Will etcd backend be completely removed?

I don't believe that is being considered at all, because the etcd datastore is needed for [Calico with OpenStack](https://docs.tigera.io/calico/latest/getting-started/openstack/overview).

caseydavenport:
Definitely not getting rid of etcdv3!

This is a prototype to see if we can get rid of the calico API server, and instead use CRDs directly for projectcalico.org/v3 APIs. 

Meant to address this: https://github.com/projectcalico/calico/issues/6412

halfcrazy:
Cel has been turned on by default since Kubernetes 1.25 and finally graduated to GA in Kubernetes [1.29.](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules) For high versions, you can use Cel to do cross-field validation from the apiserver without the need for a validation webhook. In addition, it supports some network validation [cidr/isCIDR/containsIP / containsCIDR / ip / masked / prefixLength](https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-cidr-libraryl).

Kubebuilder also supports the cel rule annotation [kubebuilder:validation:XValidation
](https://book.kubebuilder.io/reference/markers/crd-validation).

https://opensource.googleblog.com/2023/11/kubernetes-crd-validation-using-cel.html

caseydavenport:
Thanks @halfcrazy ! Agree CEL can / should be used for quite a bit of our validation.

Areas that I think may still require a validation webhook in some capacity would be:
- Cross resource validatoin (e.g., IPPools and IPAM blocks)
- Additional RBAC enforcement that we perform based on Tier access, that extends typical namespaced scoped RBAC.

api/pkg/apis/projectcalico/v3/ipreservation.go
```diff
@@ -53,9 +55,13 @@ type IPReservation struct {
 // IPReservationSpec contains the specification for an IPReservation resource.
 type IPReservationSpec struct {
 	// ReservedCIDRs is a list of CIDRs and/or IP addresses that Calico IPAM will exclude from new allocations.
-	ReservedCIDRs []string `json:"reservedCIDRs,omitempty" validate:"cidrs,omitempty"`
+	// +listType=set
+	ReservedCIDRs []CIDR `json:"reservedCIDRs,omitempty" validate:"cidrs,omitempty"`
 }
 
+// +kubebuilder:validation:Format=cidr
+type CIDR string
```
caseydavenport: This is just a type to allow inserting validation tags..

One consequence of this is that places that previously accepted an IP - e.g., 192.168.1.1 - will now need 192.168.1.1/32

Unclear if this is a problem or not, or if there is a better way to express "IP-or-CIDR" as a format (I have had trouble tracking down the docs for supported built-in formats so far)

james-callahan:
@caseydavenport is this something you expect to be merged/will it graduate from a PoC?

caseydavenport:
@james-callahan yes, I am actively working on this now. I know the PR looks out of date but that's because there are a few other intermediate PRs I need to work on first that are prerequisites for this before I can pick this one up again. 

I am hopeful that an alpha / tech-preview version of this can land in Calico v3.32.

libcalico-go/lib/testutils/resources.go
```diff
@@ -67,14 +67,39 @@ func (m *resourceMatcher) Match(actual interface{}) (success bool, err error) {
...
 	ma := res.(v1.ObjectMetaAccessor)
-	success = (ma.GetObjectMeta().GetNamespace() == m.namespace) &&
-		(ma.GetObjectMeta().GetName() == m.name) &&
-		(ma.GetObjectMeta().GetResourceVersion() != "") &&
-		(res.GetObjectKind().GroupVersionKind().Kind == m.kind) &&
-		(res.GetObjectKind().GroupVersionKind().Group == apiv3.Group) &&
-		(res.GetObjectKind().GroupVersionKind().Version == apiv3.VersionCurrent) &&
-		(m.spec == nil || reflect.DeepEqual(getSpec(res), m.spec)) &&
-		(m.status == nil || reflect.DeepEqual(getStatus(res), m.status))
+	success = true
```
caseydavenport, line 70: This is just a refactoring to get better error output for why the resources didn't match.

validatingwebhook/cmd/main.go
```diff
@@ -0,0 +1,386 @@
+// Copyright 2025 Tigera, Inc.
```
caseydavenport: Ignore this file / package for now - I'm planning to rip this out and submit it as a separate follow-on PR.

apiserver/pkg/registry/projectcalico/authorizer/authorizer.go
```diff
@@ -80,7 +80,11 @@ func (a *authorizer) AuthorizeTierOperation(
 
 		logrus.Trace("Checking authorization using tier resource type (user can get tier)")
 		logAuthorizerAttributes(attrs)
-		decisionGetTier, _, _ = a.Authorize(context.TODO(), attrs)
+		var reason string
+		decisionGetTier, reason, err = a.Authorize(context.TODO(), attrs)
```
Brian-McM: Since this logic is modified maybe you could just use the context that's passed in instead of the TODO (fine with leaving it though)?

api/pkg/apis/projectcalico/v3/register.go
```diff
@@ -88,6 +97,18 @@ func init() {
 	localSchemeBuilder.Register(addKnownTypes, addConversionFuncs)
 }
 
+func AddToGlobalScheme() error {
+	var err error
+	once.Do(func() {
```
Brian-McM, line 102: Not that I necessarily have any issue with this, but what case are you solving with this (i.e. where are we calling AddToGlobalScheme twice and why are we doing it).

caseydavenport: I am copying this logic from the other API registration we have - I think it came up in test code in the past IIRC.

api/config/crd/projectcalico.org_profiles.yaml
```diff
@@ -0,0 +1,392 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.18.0
+  name: profiles.projectcalico.org
+spec:
+  group: projectcalico.org
+  names:
+    kind: Profile
```
fasaxc, line 10: Should there be a profiles CRD?  Thought it was a synthesized resource?

api/pkg/apis/projectcalico/v3/policy_common.go
```diff
@@ -208,6 +229,7 @@ const (
 	Pass  Action = "Pass"
 )
 
+// +kubebuilder:validation:Enum=Set;Delete;Learn;Ignore
```
fasaxc, line 234: Might be hallucinating, but I think kubebuilder can auto-detect enum constants if they're defined like this.  Possible you don't need to write them out here too.

api/pkg/defaults/networkpolicy.go
```diff
@@ -0,0 +1,67 @@
...
+	changed = setTierLabel(p, p.Spec.Tier) || changed
+	return changed, nil
+}
+
+func setTierLabel(obj v1.Object, tier string) bool {
+	labels := obj.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	if labels[v3.LabelTier] != tier {
```
fasaxc, line 38: I think CRDs now support field selectors, so we could remove the reliance on a tier label.  If we require the tier label, we should use CEL to enforce it  Perhaps a MAP could set it automatically.

apiserver/test/integration/framework.go
```diff
@@ -112,10 +112,7 @@ func withConfigGetFreshApiserverServerAndClient(
 	return pcs, clientset, cfg, shutdownServer
 }
 
-func getFreshApiserverAndClient(
-	t *testing.T,
-	newEmptyObj func() runtime.Object,
-) (calicoclient.Interface, func()) {
+func getFreshApiserverAndClient(t *testing.T, newEmptyObj func() runtime.Object) (calicoclient.Interface, func()) {
```
fasaxc:
```suggestion
func getFreshAPIServerAndClient(t *testing.T, newEmptyObj func() runtime.Object) (calicoclient.Interface, func()) {
```

charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
```diff
@@ -31,6 +31,7 @@ rules:
       - gatewayapis.operator.tigera.io
       - imagesets.operator.tigera.io
       - installations.operator.tigera.io
+      - istios.operator.tigera.io
```
fasaxc: Should this be part of your PR?

cni-plugin/tests/calico_cni_ipam_test.go
```diff
@@ -101,9 +104,10 @@ var _ = Describe("Calico IPAM Tests", func() {
               "datastore_type": "%s",
               "ipam": {
                 "type": "%s"
-              }
-            }`, cniVersion, os.Getenv("ETCD_IP"), os.Getenv("DATASTORE_TYPE"), plugin)),
-			Entry("IPAM with IPv4 (explicit)", true, false, fmt.Sprintf(`
+              },
+							"calico_api_group": "%s"
```
fasaxc: Looks like the indentation is inconsistent with the other config lines

felix/fv/infrastructure/infra_etcd.go
```diff
@@ -130,6 +130,10 @@ func (eds *EtcdDatastoreInfra) GetCalicoClient() client.Interface {
 	return eds.client
 }
 
+func (eds *EtcdDatastoreInfra) UseV3API() bool {
```
fasaxc: I read this as "Use etcd v3 API" at first.

felix/fv/infrastructure/infra_k8s.go
```diff
@@ -467,49 +470,20 @@ func setupK8sDatastoreInfra(opts ...CreateOption) (kds *K8sDatastoreInfra, err e
...
-		// Sometimes the very first felix container that we start fails to
-		// mount this file.  Double check that it is there and we can read it.
-		if _, err := os.Stat(kds.CertFileName); err != nil {
-			log.WithError(err).Error("Failed to stat API server cert that we just copied?!")
-			if time.Since(start) > 120*time.Second {
-				return nil, err
-			}
-			time.Sleep(100 * time.Millisecond)
-			continue
+		if err == nil {
```
fasaxc: Looks like we lost some of my recent changes here?

felix/fv/infrastructure/topology.go
```diff
@@ -377,6 +378,11 @@ func StartNNodeTopology(
 			expectedIPs = append(expectedIPs, felix.ExpectedWireguardV6TunnelAddr)
 		}
 
+		// Deduplicate the expected IPs.
+		dedup := set.New[string]()
+		dedup.AddAll(expectedIPs)
+		expectedIPs = dedup.Slice()
```
fasaxc: Why do we get dupes?
```suggestion
		dedup := set.From(expectedIPs...)
		expectedIPs = dedup.Slice()
```

kube-controllers/pkg/controllers/ippool/pool_controller.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/projectcalico/calico/kube-controllers/pkg/controllers/controller"
+	"github.com/projectcalico/calico/libcalico-go/lib/ipam"
+	cnet "github.com/projectcalico/calico/libcalico-go/lib/net"
+)
+
+const (
+	IPPoolFinalizer = "projectcalico.org/ippool-finalizer"
+)
+
+// IPPoolController implements the Controller interface.  It is responsible for monitoring
```
fasaxc: Out of date comment?

kube-controllers/pkg/controllers/ippool/pool_controller.go
```diff
@@ -0,0 +1,229 @@
...
+	// Wait till k8s cache is synced
+	logrus.Debug("Waiting to sync with Kubernetes API")
+	if !cache.WaitForNamedCacheSync("pools", stopCh, c.poolInformer.HasSynced, c.blockInformer.HasSynced) {
+		logrus.Info("Failed to sync resources, received signal for controller to shut down.")
+		return
+	}
+
+	logrus.Debug("Finished syncing with Kubernetes API")
+
+	// We're in-sync. Start the sub-controllers.
```
fasaxc: Nothing being started?

kube-controllers/pkg/controllers/ippool/pool_controller.go
```diff
@@ -0,0 +1,229 @@
...
+func (c *IPPoolController) Reconcile(p *v3.IPPool) error {
+	ctx := context.TODO()
+	logCtx := logrus.WithFields(logrus.Fields{
+		"name":         p.Name,
+		"cidr":         p.Spec.CIDR,
+		"hasFinalizer": HasFinalizer(p),
+	})
+	logCtx.Debug("Reconciling IPPool")
+
+	if p.DeletionTimestamp != nil {
```
fasaxc, line 159: Does IPAM avoid allocating from delete-in-progress pools?

kube-controllers/pkg/controllers/networkpolicy/defaulter.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// We're in-sync. Start the sub-controllers.
+	<-stopCh
+	logrus.Info("Stopping NetworkPolicy defaulting controller")
+}
+
+func (c *policyDefaulter) Mux(obj any) {
+	switch t := obj.(type) {
+	case *v3.GlobalNetworkPolicy:
+		if err := c.defaultGlobalNetworkPolicy(t); err != nil {
```
fasaxc: Any concern with reconciliation of these objects. I'm thinking of Argo applying something without a types field and then fighting with us.  If we could (also) default with CEL then I'd expect that to be better tolerated since the returned value from your CREATE would show the default.  Not sure if I'm being paranoid!

api/pkg/apis/projectcalico/v3/networkpolicy.go
fasaxc, line 117: Could we default this at write time somehow (e.g. with a mutating admission policy)

libcalico-go/lib/backend/k8s/resources/customresource.go
```diff
@@ -271,13 +264,18 @@ func (c *customK8sResourceClient) Delete(ctx context.Context, k model.Key, revis
...
+			// The UID in the v3 resources is a translation of the UID in the CR. Translate it
+			// before passing as a precondition.
+			uid, err := conversion.ConvertUID(*uid)
+			if err != nil {
+				return nil, err
+			}
+			opts.Preconditions = &metav1.Preconditions{UID: &uid}
+		} else {
+			// If no transform is required, then we can pass the UID as-is.
+			opts.Preconditions = &metav1.Preconditions{UID: uid}
```
fasaxc, line 276: 🎉

libcalico-go/lib/backend/k8s/resources/customresource.go
```diff
@@ -440,18 +438,29 @@ func (c *customK8sResourceClient) listInterfaceToKey(l model.ListInterface) mode
...
+func (c *customResourceClient) nameToKey(name string) (model.Key, error) {
 	return model.ResourceKey{
 		Name: name,
-		Kind: c.resourceKind,
+		Kind: c.kind,
 	}, nil
 }
 
-func (c *customK8sResourceClient) convertResourceToKVPair(r Resource) (*model.KVPair, error) {
+// TODO: Don't instantiate a new TypeMeta for every call.
```
fasaxc: Not sure which TODOs you're meaning to do in this PR vs later.  could you do a TODO scan and either do them or track them in JIRA?

fasaxc: Not sure it does make sense to ever share TypeMetas

libcalico-go/lib/backend/k8s/resources/globalnetworkpolicies.go
```diff
@@ -18,29 +18,20 @@ import (
...
 
 const (
 	GlobalNetworkPolicyResourceName = "GlobalNetworkPolicies"
-	GlobalNetworkPolicyCRDName      = "globalnetworkpolicies.crd.projectcalico.org"
 )
 
-func NewGlobalNetworkPolicyClient(c kubernetes.Interface, r rest.Interface) K8sResourceClient {
-	return &customK8sResourceClient{
-		clientSet:       c,
+func NewGlobalNetworkPolicyClient(r rest.Interface, v3 bool) K8sResourceClient {
```
fasaxc: Might be nice to have an enum instead of a bool for the "CRD Source"

api/pkg/apis/projectcalico/v3/bgppeer.go
```diff
@@ -65,6 +65,7 @@ type BGPPeerSpec struct {
 	// If optional port number is not set, and this peer IP and ASNumber belongs to a calico/node
 	// with ListenPort set in BGPConfiguration, then we use that port to peer.
 	// +optional
+	// +kubebuilder:validation:Format=ip
```
fasaxc: The validation tag says "IP:port", do we allow a port here too?

libcalico-go/lib/backend/k8s/client_test.go
```diff
@@ -1794,7 +1841,7 @@ var _ = testutils.E2eDatastoreDescribe("Test Syncer API for Kubernetes backend",
 					Name: "aa-bb-cc",
 				},
 				Spec: apiv3.BGPPeerSpec{
-					PeerIP:   "aa:bb::cc",
+					PeerIP:   "aa:bb::cc/128",
```
fasaxc: Surprised to see a /128 here, shouldn't PeerIP be "just an IP"?

libcalico-go/lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
```diff
@@ -58,6 +59,15 @@ func ConvertNetworkPolicyV3ToV1Value(val any) (any, error) {
 		return nil, errors.New("Value is not a valid NetworkPolicy resource value")
 	}
 
+	// Run defaulting against the resource to ensure all fields are populated. We do this
+	// here because the resource may not have gone through the API server and so may
+	// not have had defaults applied. This logic is shared with the defaulting controller,
+	// to ensure consistency.
+	_, err := defaults.Default(v3res)
```
fasaxc: ISTR that we didn't do the same defaulting on read in order to preserve the semantics of old policies when we added the defaulting.  That was years ago, but I do think this could cause a breaking change because we've not previously done anything to force folks to migrate if they've been hacking CRDs.

Perhaps we just release note it, but Hyrum's law may apply!

node/tests/k8st/infra/build-operator.sh
```diff
@@ -5,7 +5,7 @@
 
 # Clone the repository if needed.
 set -e
-REPO=${REPO:-tigera/operator}
+REPO=${REPO:-caseydavenport/operator}
```
fasaxc: Dev change left in?

node/tests/k8st/infra/calicoctl.yaml
```diff
@@ -28,7 +28,7 @@ spec:
   serviceAccountName: calicoctl
   containers:
     - name: calicoctl
-      image: quay.io/calico/ctl:master
+      image: calico/ctl:test-build
```
fasaxc, line 31: Dev change?

node/tests/k8st/tests/test_bgp_filter.py
```diff
@@ -85,7 +85,11 @@ def setUp(self):
 """ % self.external_node_ip6)
 
     def tearDown(self):
-        super(TestBGPFilter, self).tearDown()
+        try:
+            super(TestBGPFilter, self).tearDown()
+        except Exception as e:
+            # Log the error, but don't block further teardown steps.
+            _log.error("Exception during tearDown: %s", e)
```
fasaxc, line 95: Should you re-raise at the end?  If there's a failure in teardown after a successful test then that'd be something we'd want to be alerted to.

node/tests/k8st/deploy_resources_on_kind_cluster.sh
```diff
@@ -88,10 +88,12 @@ if ! ( ${kubectl} wait --for=create --timeout=60s tigerastatus/calico &&
   ${kubectl} describe po -n calico-system
   exit 1
 fi
-if ! ${kubectl} wait --for=condition=Available --timeout=300s tigerastatus/apiserver; then
-  ${kubectl} get -o yaml tigerastatus/apiserver
-  exit 1
-fi
+
+# TODO
```
fasaxc: Think this TODO needs addressing.

metadata.mk
```diff
@@ -67,8 +67,8 @@ LIBBPF_VERSION=v1.6.2
 BPFTOOL_IMAGE=calico/bpftool:v7.5.0
 
 # The operator branch corresponding to this branch.
-OPERATOR_BRANCH ?= master
-OPERATOR_ORGANIZATION ?= tigera
+OPERATOR_BRANCH ?= casey-v3-crds
```
fasaxc: Dev change.

api/config/crd/projectcalico.org_profiles.yaml
caseydavenport: hmm, good catch. There shouldn't be really, although you can do `kubectl get profiles` with the apiserver and you won't be able to do that with the CRDs (which I think is fine).

api/pkg/apis/projectcalico/v3/policy_common.go
caseydavenport: I'll try it out - I also thought that, but then I also recall trying it out and it not working!

api/pkg/defaults/networkpolicy.go
caseydavenport: Hm, I thought I looked into using CEL here and didn't find a good option but I will check again. 

I was keeping the labels for backwards compat, but suppose it is probably niche enough that we could just update recommendations to use the field selector.

charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
caseydavenport: Nope!

node/tests/k8st/infra/calicoctl.yaml
caseydavenport: This one actually isn't a dev change - the `test-build` is the local tag our CI uses to force the use of the locally built images.

api/pkg/apis/projectcalico/v3/policy_common.go
caseydavenport: Yeah, I can't seem to find a way for it to auto-detect :confounded:

felix/fv/infrastructure/topology.go
caseydavenport: Looking at it closer, I think it might just be a bug in the code above that fills up `expectedIPs` - seems like at a minimum we're adding `felix.IPv6` in two different locations. 

I'll try removing that (and the dedupe logic) and see if that's it!

kube-controllers/pkg/controllers/ippool/pool_controller.go
caseydavenport: Yep! https://github.com/projectcalico/calico/blob/ace5c2c4f667398c91f0f1853dfd3d9ab3ddfab0/libcalico-go/lib/clientv3/client.go#L203-L206

libcalico-go/lib/backend/k8s/resources/customresource.go
caseydavenport: Hmm, I don't remember writing this TODO. I'll just remove it.

libcalico-go/lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
caseydavenport: I think this is probably one that we just change and release note, the behavior without this is pretty weird anyway (it will be missing a Types field, so won't get interpreted correctly)

kube-controllers/pkg/controllers/networkpolicy/defaulter.go
caseydavenport: OK it took a bit of messing around to get the right incantation, but I've added a MAP that manages to set the types field correctly, hooray!

Going to try the same for the label tomorrow.

api/pkg/apis/projectcalico/v3/networkpolicy.go
caseydavenport: Done! (after a bit of a hassle)

api/pkg/defaults/networkpolicy.go
caseydavenport: Added field selector support for the tier field! Got rid of this controller entirely.

libcalico-go/lib/backend/syncersv1/updateprocessors/networkpolicyprocessor.go
caseydavenport: Ended up removing this.

api/admission/networkpolicy.mutatingadmissionpolicybinding.yaml
```diff
@@ -0,0 +1,17 @@
...
+  matchResources:
+    resourceRules:
+      - apiGroups: ["projectcalico.org"]
+        apiVersions: ["v3"]
+        operations: ["CREATE", "UPDATE"]
+        resources:
+          - networkpolicies
+          - globalnetworkpolicies
+          - stagednetworkpolicies
+          - stagedglobalnetworkpolicies
```
fasaxc, line 17: Think you need to cover staged k8s netpols too.  Note that their defaulting rule is different[1]:

> policyTypes is a list of rule types that the NetworkPolicy relates to. Valid options are ["Ingress"], ["Egress"], or ["Ingress", "Egress"]. If this field is not specified, it will default based on the existence of ingress or egress rules; policies that contain an egress section are assumed to affect egress, and all policies (whether or not they contain an ingress section) are assumed to affect ingress. If you want to write an egress-only policy, you must explicitly specify policyTypes [ "Egress" ]. Likewise, if you want to write a policy that specifies that no egress is allowed, you must specify a policyTypes value that include "Egress" (since such a policy would not include an egress section and would otherwise default to just [ "Ingress" ]).

[1] https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#networkpolicyspec-v1-networking-k8s-io

api/admission/networkpolicy.mutatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,44 @@
...
+          - stagedglobalnetworkpolicies
+  failurePolicy: Fail
+  reinvocationPolicy: IfNeeded
+  variables:
+    # Determine the policy types based on the presence of ingress and egress rules.
+    - name: defaultTypes
+      expression: |
+        (has(object.spec.ingress) && has(object.spec.egress)) ? ["Ingress", "Egress"] :
+        has(object.spec.ingress) ? ["Ingress"] :
+        has(object.spec.egress) ? ["Egress"] : []
```
fasaxc: I don't think this matches our current libcalico-go behavior, which is 
```
if no egress rules: default to [Ingress] (for ancient back compat reasons, I think)
else if no ingress rules: default to [Egress]
else (have both types): default to [Ingress, Egress]
```

api/pkg/apis/projectcalico/v3/networkpolicy.go
fasaxc, line 117: With the mutating policy in place, can we make this required minLen=1 now

api/config/crd/projectcalico.org_profiles.yaml
fasaxc: Should we get rid of it then?  Still in the PR at present.

api/pkg/defaults/networkpolicy.go
fasaxc: > I was keeping the labels for backwards compat

Could still do that, but use a MAP to auto-set the label in order to keep it in sync with the field.

felix/fv/infrastructure/topology.go
fasaxc: Ah yep, I was wondering if it was just hiding a bug

api/admission/networkpolicy.mutatingadmissionpolicybinding.yaml
caseydavenport: Ahhh right.

That's... annoying! But should be fine. WDYT, try to handle it in a single MAP or write two? two seems simpler to reason about IMO.

api/pkg/apis/projectcalico/v3/networkpolicy.go
caseydavenport: Yeah, probably can ( this validation is still used for crd.projectcalico.org/v1 as well, but I think that is OK since it should always be going through our Go code before being created / updated )

api/admission/networkpolicy.mutatingadmissionpolicy.yaml
caseydavenport: ah yeah...

caseydavenport: I think it's pretty close though - the difference would be the final condition (i.e., no ingress or egress) should default to `["Ingress"]` instead of `[]`, IIUC.

api/admission/networkpolicy.mutatingadmissionpolicybinding.yaml
caseydavenport: Found a nice way to do it in one.

api/admission/networkpolicy.mutatingadmissionpolicy.yaml
```diff
@@ -21,15 +21,30 @@ spec:
...
   variables:
     # Determine the policy types based on the presence of ingress and egress rules.
+    # If both are present, set types to ["Ingress", "Egress"].
+    # If only ingress is present, set types to ["Ingress"].
+    # If only egress is present, set types to ["Egress"].
+    # If neither is present, default to ["Ingress"] for backward compatibility with policies from before types were required.
     - name: defaultTypes
       expression: |
         (has(object.spec.ingress) && has(object.spec.egress)) ? ["Ingress", "Egress"] :
         has(object.spec.ingress) ? ["Ingress"] :
```
fasaxc: Could just delete this line

api/admission/networkpolicy.mutatingadmissionpolicy.yaml
```diff
@@ -21,15 +21,30 @@ spec:
...
+        has(object.spec.egress) ? ["Egress"] : ["Ingress"]
+    # The logic for StagedKubernetesNetworkPolicy is slightly different from the above:
+    # - If both ingress and egress are present, set types to ["Ingress", "Egress"].
+    # - If only ingress is present, set types to ["Ingress"].
+    # - If only egress is present, set types to ["Ingress", "Egress"].
+    # - If neither is present, default to ["Ingress"] for backward compatibility.
+    - name: defaultKubeTypes
+      expression: |
+        (has(object.spec.ingress) && has(object.spec.egress)) ? ["Ingress", "Egress"] :
+        has(object.spec.ingress) ? ["Ingress"] :
```
fasaxc: Same here

PR #10451, opened 2025-05-22
[BPF] repurpose unused mark for CALI_SKB_MARK_TUNNEL_KEY_SET

tomastigera:
CALI_SKB_MARK_TO_NAT_IFACE_OUT is not used and we need a bit to tell tunnel (vxlan) that the tunnel has been already set. We could set the tunnel always at the vxlan device, however, for workloads, we have alredy done some lookups that we would need to duplicate at the tunnel device. At the same time, packets from host do no thave it set yet and we need to distinguish those. Especially when they are looped via the bpfin.cali device where the seen flag is already set.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/bpf.h
+++ b/felix/bpf-gpl/bpf.h
@@ -204,10 +204,10 @@ enum calico_skb_mark {
 	 * know that and we have resolved NAT etc.
 	 */
 	CALI_SKB_MARK_RELATED_RESOLVED        = 0x21000000,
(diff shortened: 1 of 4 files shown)


sridhartigera (approved):
LGTM

PR #10484, opened 2025-05-29
Optimize performance by querying affinity first.

gojoy:
## Description
For the Kubernetes API server, a `create` request consumes more resources and time than a `get` request. In the context of Calico, most resource creation requests trigger the `ErrorResourceAlreadyExists` error. Therefore, checking if an affinity exists before creating an affinity can improve performance to some extent.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/ipam/ipam_block_reader_writer.go
+++ b/libcalico-go/lib/ipam/ipam_block_reader_writer.go
@@ -177,7 +177,21 @@ func (rw blockReaderWriter) getPendingAffinity(ctx context.Context, affinityCfg 
 		Key:   model.BlockAffinityKey{Host: affinityCfg.Host, AffinityType: string(affinityCfg.AffinityType), CIDR: subnet},
 		Value: &model.BlockAffinity{State: model.StatePending},
 	}
-	aff, err := rw.client.Create(ctx, &obj)
+	aff, err := rw.queryAffinity(ctx, affinityCfg, subnet, "")
+	if err == nil {
+		logCtx.Info("Got existing affinity")
+		if aff.Value.(*model.BlockAffinity).State != model.StateConfirmed {
+			logCtx.Infof("Marking existing affinity with current state %s as pending", aff.Value.(*model.BlockAffinity).State)
+			aff.Value.(*model.BlockAffinity).State = model.StatePending
+			return rw.updateAffinity(ctx, aff)
+		}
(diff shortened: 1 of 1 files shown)


caseydavenport:
@gojoy thanks for raising this - I am trying to make sure I understand the motivation / scenarios in which this improves performance.

As I understand it, `getPendingAffinity` is called in the following scenarios:

1. Attempting to auto-assign an IP address, but there are no blocks assigned to the node with free allocation space.
2. Attempting to assign an explicit IP address (regardless of what blocks and affinities exist).
3. Explicit calls to ClaimAffinity

I assume that for your case, you are primarily concerned with (2)?

For (1), the affinity will almost certainly not exist and so attempting to create it first seems like the right priority. 
(3) is a rare operation - I believe it only happens in rare migration scenarios and not typically in normal cluster operations.

If all of this is correct, then I think we may want / need to distinguish between scenarios (1) and (2), as I think they represent very different reasons for querying the affinity. For (1), the affinity will most likely not exist, but for (2) it very likely will exist. 

WDYT?

gojoy:
> @gojoy thanks for raising this - I am trying to make sure I understand the motivation / scenarios in which this improves performance.
> 
> As I understand it, `getPendingAffinity` is called in the following scenarios:
> 
> 1. Attempting to auto-assign an IP address, but there are no blocks assigned to the node with free allocation space.
> 2. Attempting to assign an explicit IP address (regardless of what blocks and affinities exist).
> 3. Explicit calls to ClaimAffinity
> 
> I assume that for your case, you are primarily concerned with (2)?
> 
> For (1), the affinity will almost certainly not exist and so attempting to create it first seems like the right priority. (3) is a rare operation - I believe it only happens in rare migration scenarios and not typically in normal cluster operations.
> 
> If all of this is correct, then I think we may want / need to distinguish between scenarios (1) and (2), as I think they represent very different reasons for querying the affinity. For (1), the affinity will most likely not exist, but for (2) it very likely will exist.
> 
> WDYT?

Thank you for the detailed analysis. I mistakenly thought this was the optimization intended for (1), but upon reviewing your findings, it appears this optimization isn't necessary for (1). I'll close this pull request accordingly. Appreciate your clarification!

PR #10485, opened 2025-05-29
Fix up more ubuntu publishing issues [master]

danudey:
- **Add devscripts package for dch**
- **Fix up logic errors when uploading changes files**
- **Add UBUNTU_REPO_OVERRIDE env var to overide ubuntu repo**

### DIFF
--- a/.semaphore/push-images/packaging.yaml
+++ b/.semaphore/push-images/packaging.yaml
@@ -29,6 +29,9 @@ global_job_config:
       # Note that this override is for the docker containers we're building,
       # and not the VM that we're running on right now
       - export UBUNTU_REPO_OVERRIDE=ftp.uni-stuttgart.de
+      # install-package is a Semaphore-specific script that has a few benefits,
+      # notably being much faster because it uses Semaphore's cache to fetch
+      # files from.
       - install-package --no-install-recommends devscripts moreutils patchelf
 
 blocks:


.semaphore/push-images/packaging.yaml
```diff
@@ -26,8 +26,10 @@ global_job_config:
       - export SECRET_KEY=$HOME/secrets/marvin.txt
       - export GCLOUD_ARGS='--zone us-east1-c --project tigera-wp-tcp-redirect'
       - export HOST=ubuntu@binaries-projectcalico-org
-      - sudo apt-get update -y
-      - sudo apt-get install -y moreutils patchelf
+      # Note that this override is for the docker containers we're building,
+      # and not the VM that we're running on right now
+      - export UBUNTU_REPO_OVERRIDE=ftp.uni-stuttgart.de
+      - install-package --no-install-recommends devscripts moreutils patchelf
```
nelljerram, line 32: Is there a benefit from using `install-package`?  I've not heard of that, and there isn't a man page for it on my laptop.  Would rather stay with apt-get as I think more of us know that.

release/packaging/docker-build-images/install-ubuntu-build-deps
```diff
@@ -5,6 +5,12 @@ set -e # Exit immediately if a command returns non-zero
 
 export DEBIAN_FRONTEND=noninteractive
 
+mapfile -t sources_files <<< "$(find /etc/apt/sources.list* -type f)"
+
+test -n "${UBUNTU_REPO_OVERRIDE}" && (
+    sed -E -i "s/(archive|security).ubuntu.com/${UBUNTU_REPO_OVERRIDE}/" "${sources_files[@]}"
+)
```
nelljerram, line 12: Will this return non-zero and terminate the script if UBUNTU_REPO_OVERRIDE is not defined?

nelljerram: Also why execute the sed in a sub-shell?  (I think that's what the round brackets do.  I believe curly brackets will just give grouping, which is probably what you want.)

release/packaging/utils/publish-debs.sh
```diff
@@ -47,17 +36,18 @@ for series in focal jammy noble; do
     find "${outputdir}" -name "*-${series}_source.changes" -print0 | while read -r -d $'\0' changes_file; do
         filename=$(basename "${changes_file}")
         package_name="${filename%%_*}"
-        if test -f "${outputdir}/${existing_package}_source.ppa.upload"; then
+        base_ppa_file_path=$(basename ${changes_file} .changes)
+        if test -f "${outputdir}/${base_ppa_file_path}.upload"; then
             echo "Upload was already completed, skipping" | ts "[upload ${series} ${package_name}]"
-        elif test -f "${outputdir}/${existing_package}_source.ppa.previously-uploaded"; then
+        elif test -f "${outputdir}/${base_ppa_file_path}.previously-uploaded"; then
```
nelljerram, line 42: It looks like we're missing ".ppa" here - i.e. this will never match the
```
            touch "${outputdir}/${existing_package}_source.ppa.previously-uploaded"
```
line above.

release/packaging/utils/publish-debs.sh
```diff
@@ -47,17 +36,18 @@ for series in focal jammy noble; do
     find "${outputdir}" -name "*-${series}_source.changes" -print0 | while read -r -d $'\0' changes_file; do
         filename=$(basename "${changes_file}")
         package_name="${filename%%_*}"
-        if test -f "${outputdir}/${existing_package}_source.ppa.upload"; then
+        base_ppa_file_path=$(basename ${changes_file} .changes)
+        if test -f "${outputdir}/${base_ppa_file_path}.upload"; then
```
nelljerram, line 40: Do we touch this filename somewhere?

nelljerram (commented):
A few queries.

.semaphore/push-images/packaging.yaml
danudey: It's mostly a semaphore-specific tool (so we keep it in a semaphore-specific pipeline), but it's a faster and more reliable installation method because it draws on Semaphore's cache. Fine with removing it though if that makes more sense to you.

release/packaging/docker-build-images/install-ubuntu-build-deps
danudey: I'm not actually sure why I used a subshell here. I think I was going to do some other stuff here but didn't end up doing that. Will be fixing that soon.

> Will this return non-zero and terminate the script if UBUNTU_REPO_OVERRIDE is not defined?

In my testing this does work either way, but I'm going to double-check and make sure it works properly. I usually try to run shell scripts with `-u` to ensure we handle variables correctly, so I'll go back over this and make sure it's solid.

release/packaging/utils/publish-debs.sh
danudey: The `${base_ppa_file_path}` should contain the .ppa part of the file path, which is why it was removed from this line.

release/packaging/utils/publish-debs.sh
danudey: The tooling creates this file as part of the upload process, but I should add a comment to that effect to document it here.

release/packaging/docker-build-images/install-ubuntu-build-deps
danudey: Oh, I get it now. The reason why `UBUNTU_REPO_OVERRIDE` works regardless in this case is because the Dockerfile defines it regardless, but sets it to an empty string if we haven't specified a value. Will try to adjust the script to work even if that's not the case.

release/packaging/utils/publish-debs.sh
danudey: Edit: it does not.

.semaphore/push-images/packaging.yaml
nelljerram: Thanks for explaining, that makes sense.  Maybe a brief comment above that line to help future readers?

danudey: Done!

PR #10489, opened 2025-05-30
[CORE-11448] QoS controls: add packet burst configuration

coutinhop:
## Description

Add packet burst configuration support to QoS controls.

PR based on https://github.com/projectcalico/calico/pull/10458, diff will include those changes until that merges.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Added support for packet burst configuration to packet rate QoS controls.
```

## Reminder for the reviewer

Other optional labels:


felix/fv/qos_controls_test.go
```diff
@@ -329,22 +329,23 @@ var _ = infrastructure.DatastoreDescribe(
...
-					IngressPacketRate: 100,
+					IngressPacketRate:  100,
+					IngressPacketBurst: 4,
 				}
 				w[0].UpdateInInfra(infra)
 				Eventually(tc.Felixes[0].ExecOutputFn("ip", "r", "get", "10.65.0.2"), "10s").Should(ContainSubstring(w[0].InterfaceName))
 
 				By("Waiting for the config to appear in 'iptables-save/nft list ruleset' on workload 0")
 				if NFTMode() {
 					// ingress config should not be present
```
nelljerram: The following line looks like it's checking that something _does_ exist.

felix/fv/qos_controls_test.go
```diff
@@ -329,22 +329,23 @@ var _ = infrastructure.DatastoreDescribe(
...
-					Eventually(getRules(0), "10s", "1s").Should(MatchRegexp(`(?s)chain filter-cali-tw-` + w[0].InterfaceName + ` {[^}]*limit rate over \d+/second drop`))
+					Eventually(getRules(0), "10s", "1s").Should(MatchRegexp(`(?s)chain filter-cali-tw-` + w[0].InterfaceName + ` {[^}]*limit rate over \d+/second burst \d+ packets drop`))
```
nelljerram: Why do we allow any digits here, rather than checking for the exact values that we expect to see?

felix/fv/qos_controls_test.go
nelljerram: (IOW I think the comment might be wrong)

felix/fv/qos_controls_test.go
nelljerram: (if there's a markup here, please apply throughout to similar places in this file)

felix/fv/qos_controls_test.go
```diff
@@ -407,14 +409,14 @@ var _ = infrastructure.DatastoreDescribe(
 				By("Waiting for the config to disappear in 'iptables-save/nft list ruleset' on workload 1")
 				if NFTMode() {
 					// ingress config should be present
```
nelljerram: "should not be present", I think

libcalico-go/lib/backend/k8s/conversion/conversion_test.go
```diff
@@ -1378,6 +1386,33 @@ var _ = Describe("Test Pod conversion", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(wep.Value.(*libapiv3.WorkloadEndpoint).Spec.QoSControls).To(BeNil())
 	})
+
+	It("should ignore packet burst QoSControl annotation if packet rate is not present", func() {
```
nelljerram, line 1390: Is this actually an important test?  Based on the Felix coding, won't the packet burst will be ignored if the packet rate is not specified?

felix/fv/qos_controls_test.go
nelljerram, line 1: These changes do not include looking for anything in the iperf output that confirms the packet burst setting.  Would it be feasible to add that?

felix/fv/qos_controls_test.go
coutinhop: Fair point, I think there was some discussion way back around not being so strict on these checks here (as they're more of a way to wait for the config to be in effect), as there is already more thorough checking of the actual rule formatting. But both nftables and iptables should at least be consistent, I'll change it so we do check the values for both. Thanks!

felix/fv/qos_controls_test.go
coutinhop: yes, thanks for catching this!

felix/fv/qos_controls_test.go
coutinhop: again, thank you!

libcalico-go/lib/backend/k8s/conversion/conversion_test.go
coutinhop: you're right, it will be ignored, I guess I added it mirroring the test cases for bandwidth, do you think I should remove this?

felix/fv/qos_controls_test.go
coutinhop: It is feasible, but I did run into some issues: basically, using TCP for the packet rate control testing is far from ideal because of its own throttling for congestion avoidance, but I couldn't get it to work reliably using UDP (it lacks a retry mechanism for the initial connection test, so it more often than not just fails when using UDP).

I did manage to use iperf2 while developing but didn't add the required infra to use it on this test. I'll try to see if we can somewhat reliably test packet bursts on iperf3 and add to the test, otherwise I'll add iperf2 infra and improve the packet rate tests as a whole in a future PR.

nelljerram: FTR, we discussed this; I do think we should have an ongoing regression test for the packet burst, but I'm happy that that doesn't need to be in this PR - in that case, please let's just track that that is still needed.

libcalico-go/lib/backend/k8s/conversion/conversion_test.go
nelljerram: FTR, I withdraw my comment.  When I wrote this I think I was actually questioning whether the conversion code should be suppressing the packet burst setting at all.  But I'm happy now that that's a good thing to do.

nelljerram (approved):

felix/fv/qos_controls_test.go
coutinhop: I did manage to add iperf2+UDP testing with packet peakrate coverage

PR #10490, opened 2025-05-30
ebpf - Fix attaching ctlb for ipv6 and detaching programs in legacy mode.

sridhartigera:
## Description

This PR fixes
1. Attach cgroup for ipv6
2. While detaching cgroup-legacy, the cgroup root has to be opened, closed every time for each CGROUP attach type.
3. Close FD when opening link from a pin.
4. UTs for attach/detach for v4 and v6. UTs validate there are no ctlb programs lingering around after removing ctlb.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/ut/attach_test.go
+++ b/felix/bpf/ut/attach_test.go
@@ -916,78 +916,164 @@ func TestRepeatedAttach(t *testing.T) {
 
 func TestCTLBAttachLegacy(t *testing.T) {
 	RegisterTestingT(t)
-	err := nat.InstallConnectTimeLoadBalancerLegacy(true, false, "", "debug", 60*time.Second, false)
-	Expect(err).NotTo(HaveOccurred())
 
-	checkPinPath := func(pinPath string, mustExist bool) {
-		_, err := os.Stat(pinPath)
-		if mustExist {
-			Expect(err).NotTo(HaveOccurred())
-		} else {
-			Expect(err).To(HaveOccurred())
+	testCtlbAttachLegacy := func(v4, v6 bool) {
+		err := nat.InstallConnectTimeLoadBalancerLegacy(v4, v6, "", "debug", 60*time.Second, false)
+		Expect(err).NotTo(HaveOccurred())
+
+		checkPinPath := func(pinPath string, mustExist bool) {
+			_, err := os.Stat(pinPath)
+			if mustExist {
+				Expect(err).NotTo(HaveOccurred())
+			} else {
+				Expect(err).To(HaveOccurred())
+			}
 		}
+
+		checkPinPath("/sys/fs/bpf/ctlb/calico_connect_v4", false)
+		checkPinPath("/sys/fs/bpf/ctlb/calico_connect_v46", false)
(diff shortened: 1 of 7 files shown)


felix/bpf/ut/attach_test.go
```diff
@@ -916,78 +916,164 @@ func TestRepeatedAttach(t *testing.T) {
...
+		Expect(string(out)).ShouldNot(ContainSubstring("cali_ctlb_progs"))
+		cmd = exec.Command("bpftool", "prog", "show")
+		out, err = cmd.Output()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(out)).ShouldNot(ContainSubstring("calico_connect"))
+		Expect(string(out)).ShouldNot(ContainSubstring("calico_send"))
+		Expect(string(out)).ShouldNot(ContainSubstring("calico_recv"))
+	}
+	testCtlbAttachLegacy(true, false)
+	testCtlbAttachLegacy(false, true)
```
tomastigera, line 984: what about dual stack?

felix/bpf/ut/attach_test.go
```diff
@@ -916,78 +916,164 @@ func TestRepeatedAttach(t *testing.T) {
...
+		Expect(string(out)).ShouldNot(ContainSubstring("cali_ctlb_progs"))
+		cmd = exec.Command("bpftool", "prog", "show")
+		out, err = cmd.Output()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(out)).ShouldNot(ContainSubstring("calico_connect"))
+		Expect(string(out)).ShouldNot(ContainSubstring("calico_send"))
+		Expect(string(out)).ShouldNot(ContainSubstring("calico_recv"))
+	}
+	testCtlbAttach(true, false)
+	testCtlbAttach(false, true)
```
tomastigera, line 1068: dualstack?

felix/dataplane/linux/int_dataplane.go
```diff
@@ -982,7 +982,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 			log.Infof("Connect time load balancer enabled: %s", config.BPFConnTimeLB)
 		} else {
 			// Deactivate the connect-time load balancer.
-			err = nat.RemoveConnectTimeLoadBalancer(config.BPFCgroupV2)
+			err = nat.RemoveConnectTimeLoadBalancer(true, config.BPFCgroupV2)
```
tomastigera, line 985: why do we need the ipv4 arg when in practice it is always true?

tomastigera (commented):
some nits

felix/dataplane/linux/int_dataplane.go
sridhartigera: Just following the same convention as ```InstallConnectTimeLoadBalancer```. In future if we want to add `config.IPv4Enabled`, then we can replace this true.

PR #10503, opened 2025-06-03
Fix mounting cgroupv2 into calico-node

sridhartigera:
## Description

calico-node mounts the host's cgroupv2 to install CTLB. This is done by entering the cgroup namespace of the host which is at ```/proc/1/ns/cgroup``` and ```/proc/1/ns/mnt```. 

In the operator, we mount ```/proc``` into ```/nodeproc```  at https://github.com/tigera/operator/blob/master/pkg/render/node.go#L1055. In the mountns utility we were still using the wrong path to enter into the host ns which resulted in cgroupv2 not mounted.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
eBPF - Fixed mounting cgroupv2 for connect time load balancing.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/cmd/mountns/main.go
+++ b/node/cmd/mountns/main.go
@@ -49,11 +49,11 @@ import (
 __attribute__((constructor)) void set_namespaces(void) {
 	// open /initproc/ns/cgroup, which is equivalent to /proc/1/ns/cgroup on host.
 	// Then run setns syscall to change the cgroup namespace to this value.
-	setns(open("/initproc/ns/cgroup", O_RDONLY, 0), CLONE_NEWCGROUP);
+	setns(open("/nodeproc/1/ns/cgroup", O_RDONLY, 0), CLONE_NEWCGROUP);
 
-	// open /initproc/ns/mnt, which is equivalent to /proc/1/ns/mnt on host.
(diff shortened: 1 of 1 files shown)


node/cmd/mountns/main.go
```diff
@@ -49,11 +49,11 @@ import (
 __attribute__((constructor)) void set_namespaces(void) {
 	// open /initproc/ns/cgroup, which is equivalent to /proc/1/ns/cgroup on host.
```
matthewdupre: Comment is outdated

matthewdupre (approved):
Approved subject to fixing the comment (won't need to rereview).  I am curious how this escaped testing though: is the impact of not mounting this fairly small and not obvious?

Was there some other change (i.e. initproc worked before)?

sridhartigera:
> Approved subject to fixing the comment (won't need to rereview). I am curious how this escaped testing though: is the impact of not mounting this fairly small and not obvious?
> 
> Was there some other change (i.e. initproc worked before)?

initproc was the mountpoint for /proc/1 but was later changed to /nodeproc. This part was not changed though. I think we are lucky so far with the other platforms.

PR #10523, opened 2025-06-05
POC Maglev

aaaaaaaalex:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/conntrack.h
```diff
@@ -581,6 +581,15 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_lookup(struct cali_tc_c
...
 	struct ct_lookup_ctx *ct_ctx = &ct_lookup_ctx;
 
+	struct calico_maglev_key mg_lookup_key = {
+		.vip = STATE->ip_dst,
+		.port = STATE->dport,
+		.proto = STATE->ip_proto,
+		.ordinal = 0,
+	};
+	struct calico_nat_dest *has_maglev_entry;
+	has_maglev_entry = cali_mag_be_lookup_elem(&mg_lookup_key);
```
tomastigera: if I read it correctly, this is a NAT lookup rather than ct lookup. We do not want to resolve NAT everytime. Once there is conntrack, we should stick with what we have. Even if we do not want to do conntrack for magleved connections (stateless), we should first check if there is a conntrack and iff not then do NAT, either maglev or classic, right?

felix/bpf-gpl/tc.c
```diff
@@ -435,7 +435,8 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 	if (CALI_F_TO_HOST || (CALI_F_FROM_HOST && !skb_seen(ctx->skb) && !ctx->nat_dest /* no sport conflict */)) {
 		ctx->nat_dest = calico_nat_lookup_tc(ctx,
 						     &ctx->state->ip_src, &ctx->state->ip_dst,
-						     ctx->state->ip_proto, ctx->state->dport,
+						     ctx->state->ip_proto,
+							 ctx->state->sport, ctx->state->dport,
```
tomastigera, line 464: do we wen to have a single entry to NAT or are we going to ask for maglev and then for regular if maglev fails?

PR #10525, opened 2025-06-05
Use lib/std clock for time tracking in goldmane / whisker and use ControlledClock for fv tests

Brian-McM:
There's a tricky sporadic race condition a whisker integration test with goldmane that has to do with exact timing of push flows.

Use our lib/std clock for time allows us to swap out the time usage in both goldmane and whisker and control exactly what time these components see, and how the clock advances.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


whisker-backend/test/fv/goldmaneintegration_test.go
```diff
@@ -201,26 +214,25 @@ func TestGoldmaneIntegration_FilterHints(t *testing.T) {
 		app.Run(ctx, whiskerCfg)
 	}()
 
+	// We want to actually wait 5 seconds, not use our fake clock for this.
+	time.Sleep(clock.Second * 5)
```
caseydavenport: Why do we need to sleep here? I forget - might be worth adding to the comment.

Brian-McM: It's just to ensure that the flow is stored before the clock advances, otherwise the rollover could happen before it's actually pushed and this will be treated as a delayed flow.

PR #10500, opened 2025-06-03
OpenStack-level support for peakrate control

nelljerram:
Add the Calico-specific config parameters.

Change our current bandwidth rule coding so that:

- it configures burst size using the new Calico-specific parameters instead of max_burst_kbps

- it uses max_burst_kbps to configure the peakrate

- it configures the minburst size using new Calico-specific parameters.

## Release Note

```release-note
TBD
```

PR #10532, opened 2025-06-06
[release-v3.30] Auto pick #10192: OpenStack: Don't modify port dict when collecting additional #10500: OpenStack-level support for peakrate control #10521: OpenStack-level support for packet rate burst control

nelljerram:
Cherry pick of projectcalico/calico/pull/10192 projectcalico/calico/pull/10500 projectcalico/calico/pull/10521 on release-v3.30.

#10192: OpenStack: Don't modify port dict when collecting additional
#10500: OpenStack-level support for peakrate control
#10521: OpenStack-level support for packet rate burst control

# Original PR Body below

## Release Note

```release-note
Calico for OpenStack's QoS support has been enhanced to cover packet rate burst size configuration.  There are new driver settings (cluster-wide, set in `neutron.conf`) for configuring max and min packet burst sizes in each direction.

Calico for OpenStack's QoS support has been enhanced to cover peak rate configuration.  The "max_burst_kbps" field of bandwidth limit rules is honoured as the peak rate, and there are new driver settings (cluster-wide, set in `neutron.conf`) for configuring max and min burst sizes in each direction.
```

While working on #10160 I noticed that our UT code sometimes has to reset generated elements in the mock port dicts that it uses, e.g.

    del self.osdb_ports[0]['network_qos_policy_id']

This is because our driver code gets a port dict, in a create or update hook, and then has to gather various pieces of related information, from other tables in the Neutron DB, before being ready to map that to a corresponding Calico WorkloadEndpoint; and it stores that additional information by adding keys to the supplied port dict.

It's a bit fragile to modify the input port dict like that, and it would be better to collect the additional information in a separate object, as this change now does.

### DIFF
--- a/libcalico-go/lib/apis/v3/generated.openapi.go
+++ b/libcalico-go/lib/apis/v3/generated.openapi.go
@@ -3166,14 +3166,7 @@ func schema_libcalico_go_lib_apis_v3_QoSControls(ref common.ReferenceCallback) c
 					},
 					"ingressPacketRate": {
 						SchemaProps: spec.SchemaProps{
-							Description: "Ingress packet rate limit in packets per second.  Only applied if non-zero.  When non-zero, must be between 1 and 10^4 (10k).",
-							Type:        []string{"integer"},
-							Format:      "int64",
-						},
-					},
-					"egressPacketRate": {
-						SchemaProps: spec.SchemaProps{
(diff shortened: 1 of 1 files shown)


nelljerram:
CI is fully green, after a couple of reruns of the jobs that initially failed.

coutinhop (approved):

Issue #10469, opened 2025-05-25
Multihoming broken after upgrade to 3.30.0


sergeimonakhov:
Return packets dropped when arriving via different interface.

## Expected Behavior
In a multihomed environment, Calico should allow return packets to arrive through a different interface than the one used for the initial outbound traffic. This behavior worked as expected in Calico v3.29.4.

## Current Behavior
After upgrading to Calico v3.30.0, packets that return via a different interface than the one they were sent through are being dropped. This breaks multihomed setups that previously functioned without issue in v3.29.4.

## Possible Solution
This issue might be related to changes introduced in the following pull request:
https://github.com/projectcalico/calico/pull/10085

## Steps to Reproduce (for bugs)
1. Set up two Kubernetes nodes, each configured with multiple network interfaces (multihomed).
2. Deploy a pod running nginx on each node.
3. From one pod, send HTTP requests to the nginx pod on the other node.
4. Use source-based or policy-based routing to force traffic to return via a different interface than the one used for sending.
5. Observe that in Calico v3.30.0, return packets are dropped, whereas in v3.29.4, communication works correctly.

## Context
This issue is blocking our upgrade to Calico v3.30.0, as multihoming is a critical part of our network setup. Everything worked as expected in v3.29.4, but v3.30.0 breaks functionality that is essential to our operations.

## Your Environment
* Calico version: 3.30.0 (works in 3.29.4)
* Calico dataplane (iptables, windows etc.): ebpf,bgp
* Orchestrator version (e.g. kubernetes, mesos, rkt): Kubernetes v1.31
* Operating System and version: Ubuntu 24.04
* Link to your project (optional):

sridhartigera:
@sergeimonakhov Are any of the host interfaces not managed by calico? Can you please share the bpf debug logs? You can set `bpfLogLevel` to `Debug` and set `bpfLogFilters` accordingly to filter out logs that we are interested in.

tomastigera:
I am not sure what could have regressed in 3.30, there was quite a rework of how we forward packets. We are happy to work with you on fixing the problem if you can provide more debug info. We can also help to debug in slack https://calicousers.slack.com/ in #ebpf

tomastigera:
[...]

PR #10535, opened 2025-06-06
[release-v3.30] Auto pick #10511: fix regression for multihomed nodes

tomastigera:
Cherry pick of projectcalico/calico/pull/10511 on release-v3.30.

#10511: fix regression for multihomed nodes

# Original PR Body below

## Description

    When there is asymetric routing on a host with multiple NICs, we have
    packets returning from a different device then outbound packets are sent
    to. We should not use the ingress device indef for egress forwarding.
    
    This requires loose rpf check. We were previously switching remembered
    ifindex whenever we see packets arriving from a different NIC. This
    change will only remember the index if the returning packets satisfy
    strict RPF - that is when the return path reflects routing.

fixes https://github.com/projectcalico/calico/issues/10469

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: fix forwarding for asymetric routing https://github.com/projectcalico/calico/issues/10469
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/conntrack.h
+++ b/felix/bpf-gpl/conntrack.h
@@ -1007,27 +1007,40 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_lookup(struct cali_tc_c
 						src_to_dst->ifindex, ifindex);
 			}
 
-			bool rpf_passed = false;
+			int rpf_passed = RPF_RES_FAIL;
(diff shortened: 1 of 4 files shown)


sergeimonakhov:
@tomastigera https://github.com/projectcalico/calico/issues/10469#issuecomment-3126677600

PR #10539, opened 2025-06-09
Update default trusted bundle file path from "tigera-ca-bundle.crt" to "ca.crt"

Brian-McM:
## Description

The config map key has been changed in via the operator. The old value is still available but we want to remove that value after a release or two.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/test-tools/mocknode/cmd/mocknode/main.go
+++ b/test-tools/mocknode/cmd/mocknode/main.go
@@ -110,7 +110,7 @@ func (s *syncerCallbacks) LogStats() {
 const (
 	typhaNamespace      = "calico-system"
 	typhaK8sServiceName = "calico-typha"
-	typhaCAFile         = "/etc/pki/tls/certs/tigera-ca-bundle.crt"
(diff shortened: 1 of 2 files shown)


rene-dekker (approved):

hjiawei:
What is the rationale behind this change? On RHEL-based systems, the system-wide CA bundle is typically named ca-bundle.crt. From my perspective, it would be more appropriate to prepend the company name to our custom CA bundle (current naming).

Brian-McM:
>What is the rationale behind this change? On RHEL-based systems, the system-wide CA bundle is typically named ca-bundle.crt. From my perspective, it would be more appropriate to prepend the company name to our custom CA bundle (current naming).

This is to be inline with the naming that we updated in the operator so envoy gateway can pick up the changes, as the it's not configurable in envoy.

PR #10557, opened 2025-06-13
eBPF - Change ctmap to lru_hash

sridhartigera:
## Description

This PR has the following changes
1. Change the ctMap type to lru_hash. This also results in the ctmap version bumped to 4.
2. override the ```bpfConntrackCleanupMode``` to ```userspace``` always as BPF conntrack cleaner is incompatible the lru hash map as it would mess up the LRU algorithm.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
eBPF - Changed the conntrack map type to LRU Hash to facilitate conntrack cleaning even when felix is not running
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -635,8 +635,9 @@ type FelixConfigurationSpec struct {
 
 	// BPFConntrackCleanupMode controls how BPF conntrack entries are cleaned up.  `Auto` will use a BPF program if supported,
 	// falling back to userspace if not.  `Userspace` will always use the userspace cleanup code.  `BPFProgram` will
-	// always use the BPF program (failing if not supported).
-	// [Default: Auto]
+	// always use the BPF program (failing if not supported). To be deprecated soon, as we are changing the conntrack map type to
```
```suggestion
	// always use the BPF program (failing if not supported).
	//
	///To be deprecated in future versions as conntrack map type changed to
```

felix/bpf/ut/to_host_allowed_test.go
```diff
@@ -101,7 +92,7 @@ func TestToHostAllowedCTFull(t *testing.T) {
 	runBpfTest(t, "calico_from_host_ep", nil, func(bpfrun bpfProgRunFn) {
 		res, err := bpfrun(synPkt)
 		Expect(err).NotTo(HaveOccurred())
-		Expect(res.Retval).To(Equal(resTC_ACT_SHOT))
+		Expect(res.Retval).To(Equal(resTC_ACT_UNSPEC))
 	})
 
 	// Destination is a local workload - should not pass
```
tomastigera: These comments are not true anymore as the full ct does not block these cases anymore.

We should renames the test `TestToHostAllowedCTFull` as well.

It seems like testing whether the most recently created entries survive from case to case may have some value.

felix/fv/bpf_map_resize_test.go
```diff
@@ -163,110 +162,6 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ Felix bpf test configurable
 	})
 })
 
-var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ Felix bpf conntrack table dynamic resize", []apiconfig.DatastoreType{apiconfig.Kubernetes}, func(getInfra infrastructure.InfraFactory) {
```
tomastigera, line 166: This is still a valid test, isn't it? We still copy stuff over to a larger map. The only thing that changes is that LRU will be messed up after the resize. But that probably does not matter as there should be space and the cleaner will keep making room.

tomastigera:
@sridhartigera this PR https://github.com/projectcalico/calico/pull/9741 may be of interest after this change since the metrics provided by the bpfcleaner are not going to supersede the missing userspace metrics

felix/fv/bpf_map_resize_test.go
sridhartigera: As discussed, will add this test back once the automatic resizing support is added to the userspace cleaner.

tomastigera (approved):

PR #10563, opened 2025-06-16
defend against malformed notnets rules

dzacball:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/rules/policy.go
+++ b/felix/rules/policy.go
@@ -142,6 +142,50 @@ func filterNets(mixedCIDRs []string, ipVersion uint8) (filtered []string, filter
 	return
 }
 
+// isUniversalCIDR returns true if the CIDR represents "all addresses" for the given IP version.
+// This is used to detect problematic negated matches that would create logical contradictions.
+func isUniversalCIDR(cidr string, ipVersion uint8) bool {
+	if ipVersion == 4 {
+		return cidr == "0.0.0.0/0"
+	} else if ipVersion == 6 {
+		return cidr == "::/0"
+	}
+	return false
+}
+
+// filterNetsAndValidateNegation filters CIDRs by IP version and validates that negated universal
+// CIDRs are not present, which would create logical contradictions in iptables rules.
+func filterNetsAndValidateNegation(mixedCIDRs []string, ipVersion uint8, isNegated bool) (filtered []string, filteredAll bool) {
+	if len(mixedCIDRs) == 0 {
+		return nil, false
+	}
+	wantV6 := ipVersion == 6
+	filteredAll = true  // Start assuming all will be filtered out
+	
+	for _, net := range mixedCIDRs {
+		isV6 := strings.Contains(net, ":")
+		if isV6 != wantV6 {
+			continue
+		}
+		
+		// Check for universal CIDR in negated context, which creates logical contradictions
+		if isNegated && isUniversalCIDR(net, ipVersion) {
+			logrus.WithFields(logrus.Fields{
+				"cidr":      net,
+				"ipVersion": ipVersion,
+				"negated":   isNegated,
+			}).Warn("Filtering out rule with negated universal CIDR to prevent iptables logical contradiction")
+			// Return filteredAll=true to indicate the entire rule should be dropped
+			return nil, true
+		}
+		
+		filtered = append(filtered, net)
+		filteredAll = false  // Found at least one valid CIDR of the right IP version
+	}
+	
+	return
+}
+
 // FilterRuleToIPVersion: If the rule applies to the given IP version, returns a copy of the rule
 // excluding the CIDRs that are not of the given IP version. If the rule does not apply to the
 // given IP version at all, returns nil.
@@ -178,7 +222,7 @@ func FilterRuleToIPVersion(ipVersion uint8, pRule *proto.Rule) *proto.Rule {
 	if filteredAll {
 		return nil
 	}
-	ruleCopy.NotSrcNet, filteredAll = filterNets(pRule.NotSrcNet, ipVersion)
+	ruleCopy.NotSrcNet, filteredAll = filterNetsAndValidateNegation(pRule.NotSrcNet, ipVersion, true)
 	if filteredAll {
 		return nil
 	}
@@ -186,7 +230,7 @@ func FilterRuleToIPVersion(ipVersion uint8, pRule *proto.Rule) *proto.Rule {
 	if filteredAll {
 		return nil
 	}
-	ruleCopy.NotDstNet, filteredAll = filterNets(pRule.NotDstNet, ipVersion)
+	ruleCopy.NotDstNet, filteredAll = filterNetsAndValidateNegation(pRule.NotDstNet, ipVersion, true)
 	if filteredAll {
 		return nil
 	}
--- a/libcalico-go/lib/validator/v3/validator.go
+++ b/libcalico-go/lib/validator/v3/validator.go
@@ -1319,6 +1319,8 @@ func validateRule(structLevel validator.StructLevel) {
 
 	scanNets := func(nets []string, fieldName string) {
 		var v4, v6 bool
+		isNegatedField := strings.Contains(fieldName, "NotNets")
+		
 		for _, n := range nets {
 			_, cidr, err := cnet.ParseCIDR(n)
 			if err != nil {
@@ -1327,6 +1329,15 @@ func validateRule(structLevel validator.StructLevel) {
 			} else {
 				v4 = v4 || cidr.Version() == 4
 				v6 = v6 || cidr.Version() == 6
+				
+				// Check for universal CIDR in negated context, which creates logical contradictions
+				if isNegatedField {
+					if (cidr.Version() == 4 && n == "0.0.0.0/0") ||
+					   (cidr.Version() == 6 && n == "::/0") {
+						structLevel.ReportError(reflect.ValueOf(n), fieldName,
+							"", reason("universal CIDR in negation creates logical contradiction (matches no traffic)"), "")
+					}
+				}
 			}
 		}
 		if rule.IPVersion != nil && ((v4 && *rule.IPVersion != 4) || (v6 && *rule.IPVersion != 6)) {


felix/rules/policy.go
```diff
@@ -142,6 +142,50 @@ func filterNets(mixedCIDRs []string, ipVersion uint8) (filtered []string, filter
...
+// isUniversalCIDR returns true if the CIDR represents "all addresses" for the given IP version.
+// This is used to detect problematic negated matches that would create logical contradictions.
+func isUniversalCIDR(cidr string, ipVersion uint8) bool {
+	if ipVersion == 4 {
+		return cidr == "0.0.0.0/0"
+	} else if ipVersion == 6 {
+		return cidr == "::/0"
+	}
+	return false
+}
```
coutinhop: I haven't heard the term "universal CIDR", what do you think of using  "catch-all"? (We could also go with "wildcard CIDR" if you'd prefer).

Also, how about shortening this function?

```suggestion
// isCatchAllCIDR returns true if the CIDR represents "all addresses" for the given IP version.
// This is used to detect problematic negated matches that would create logical contradictions.
func isCatchAllCIDR(cidr string, ipVersion uint8) bool {
	return (ipVersion == 4 &&  cidr == "0.0.0.0/0") || (ipVersion == 6 && cidr == "::/0")
}
```

One thing we may need to worry about is catching all equivalent IPv6 representations of this CIDR (for example `0:0:0:0:0:0:0:0/0` also means the same, the issue is there's plenty of ways, really).

felix/rules/policy.go
```diff
@@ -142,6 +142,50 @@ func filterNets(mixedCIDRs []string, ipVersion uint8) (filtered []string, filter
...
+
+// filterNetsAndValidateNegation filters CIDRs by IP version and validates that negated universal
+// CIDRs are not present, which would create logical contradictions in iptables rules.
+func filterNetsAndValidateNegation(mixedCIDRs []string, ipVersion uint8, isNegated bool) (filtered []string, filteredAll bool) {
```
coutinhop: We added this new `filterNetsAndValidateNegation()` function with an `isNegated` argument but it's only being called for the negated rules. Did you mean to replace `filterNets()` with this and use `isNegated` to differentiate the behaviour?

We might just keep calling it `filterNets()` then, and replace it with this function. What do you think?

felix/rules/policy.go
```diff
@@ -142,6 +142,50 @@ func filterNets(mixedCIDRs []string, ipVersion uint8) (filtered []string, filter
...
+				"cidr":      net,
+				"ipVersion": ipVersion,
+				"negated":   isNegated,
+			}).Warn("Filtering out rule with negated universal CIDR to prevent iptables logical contradiction")
+			// Return filteredAll=true to indicate the entire rule should be dropped
+			return nil, true
+		}
+		
+		filtered = append(filtered, net)
+		filteredAll = false  // Found at least one valid CIDR of the right IP version
```
coutinhop: Thanks for this comment, please just have it above the line instead.

```suggestion
                // Found at least one valid CIDR of the right IP version
		filteredAll = false 
```

felix/rules/policy.go
```diff
@@ -142,6 +142,50 @@ func filterNets(mixedCIDRs []string, ipVersion uint8) (filtered []string, filter
...
+			continue
+		}
+		
+		// Check for universal CIDR in negated context, which creates logical contradictions
+		if isNegated && isUniversalCIDR(net, ipVersion) {
+			logrus.WithFields(logrus.Fields{
+				"cidr":      net,
+				"ipVersion": ipVersion,
+				"negated":   isNegated,
+			}).Warn("Filtering out rule with negated universal CIDR to prevent iptables logical contradiction")
```
coutinhop: I think the term "filtering" might be a little overloaded here. How about "ignoring"?

```suggestion
			}).Warn("Ignoring rule with negated catch-all CIDR to prevent iptables logical contradiction")
```

felix/rules/policy.go
```diff
@@ -142,6 +142,50 @@ func filterNets(mixedCIDRs []string, ipVersion uint8) (filtered []string, filter
...
+}
+
+// filterNetsAndValidateNegation filters CIDRs by IP version and validates that negated universal
+// CIDRs are not present, which would create logical contradictions in iptables rules.
+func filterNetsAndValidateNegation(mixedCIDRs []string, ipVersion uint8, isNegated bool) (filtered []string, filteredAll bool) {
+	if len(mixedCIDRs) == 0 {
+		return nil, false
+	}
+	wantV6 := ipVersion == 6
+	filteredAll = true  // Start assuming all will be filtered out
```
coutinhop: Thanks for this comment, please just have it above the line instead.

```suggestion
        // Start assuming all will be filtered out
	filteredAll = true
```

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -1319,6 +1319,8 @@ func validateRule(structLevel validator.StructLevel) {
 
 	scanNets := func(nets []string, fieldName string) {
 		var v4, v6 bool
+		isNegatedField := strings.Contains(fieldName, "NotNets")
```
coutinhop: The values of `fieldName` we are interested here are `Source.NotNets` and `Destination.NotNets`, right? How about testing for them directly instead of using `strings.Contains()`? I think that would be more robust against unforeseen behaviour if something changes in the future...

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -1327,6 +1329,15 @@ func validateRule(structLevel validator.StructLevel) {
 			} else {
 				v4 = v4 || cidr.Version() == 4
 				v6 = v6 || cidr.Version() == 6
+				
+				// Check for universal CIDR in negated context, which creates logical contradictions
+				if isNegatedField {
+					if (cidr.Version() == 4 && n == "0.0.0.0/0") ||
+					   (cidr.Version() == 6 && n == "::/0") {
+						structLevel.ReportError(reflect.ValueOf(n), fieldName,
+							"", reason("universal CIDR in negation creates logical contradiction (matches no traffic)"), "")
```
coutinhop:
```suggestion
							"", reason("catch-all CIDR in negation creates logical contradiction (matches no traffic)"), "")
```

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -1327,6 +1329,15 @@ func validateRule(structLevel validator.StructLevel) {
 			} else {
 				v4 = v4 || cidr.Version() == 4
 				v6 = v6 || cidr.Version() == 6
+				
+				// Check for universal CIDR in negated context, which creates logical contradictions
+				if isNegatedField {
+					if (cidr.Version() == 4 && n == "0.0.0.0/0") ||
+					   (cidr.Version() == 6 && n == "::/0") {
```
coutinhop: CI is not happy with this indentation

```suggestion
						   (cidr.Version() == 6 && n == "::/0") {
```

felix/rules/policy.go
dzacball: Thanks for the suggested code change. Yes, I agree, still not perfect (due to alternatives like `0:0:0:0:0:0:0:0/0`).

libcalico-go/lib/validator/v3/validator.go
dzacball: Ahh sorry & thanks

libcalico-go/lib/validator/v3/validator.go
dzacball: Thanks, fixed

felix/rules/policy.go
dzacball: Thanks, did just that

mazdakn:
@dzacball can you please add some description and possibly an issue related to this fix?
The intention might be clear at this point, but adding a description helps us to understand the motivation when referencing this in future.

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -1327,6 +1328,15 @@ func validateRule(structLevel validator.StructLevel) {
 			} else {
 				v4 = v4 || cidr.Version() == 4
 				v6 = v6 || cidr.Version() == 6
+
+				// Check for catch-all CIDR in negated context, which creates logical contradictions
+				if isNegatedField {
+					if (cidr.Version() == 4 && n == "0.0.0.0/0") ||
+						(cidr.Version() == 6 && n == "::/0") {
```
coutinhop: If we parse the CIDR we can guarantee we are covering all possible representations of the IPv6 catch-all:

```suggestion
						(cidr.Version() == 6 && cidr.Mask.String() == cnet.MustParseCIDR("::/0").Mask.String() && cidr.IP.Equal(cnet.MustParseCIDR("::/0").IP)) {
```

coutinhop (commented):
Thanks so much for considering my comments! This PR is looking great, we're only need to address a couple of points:
- please see my suggestion of parsing the v6 CIDR so that we can treat all ways of representing the v6 catch-all ("::/0", "0::0/0", "::0/0", etc)
- we need to add some testing, could you add some test cases to https://github.com/projectcalico/calico/blob/e9e30206a97f05b83fc5ce8cea3b0bb0b5f9039c/libcalico-go/lib/validator/v3/validator_test.go#L3605 (you can run the test locally with `make -C libcalico-go/ ut GINKGO_FOCUS=Validator`)
- could you also add a test case to https://github.com/projectcalico/calico/blob/e9e30206a97f05b83fc5ce8cea3b0bb0b5f9039c/felix/rules/policy_test.go#L106 ? I'm not 100% sure we should add IPv6 parsing here, we might be safe from the checks in the validator already...

coutinhop:
An example of test cases to add to `validator_test.go`:

```
		Entry("disallow a NotNets catch-all IPv4 CIDR match",
			&api.NetworkPolicy{
				ObjectMeta: v1.ObjectMeta{Name: "thing"},
				Spec: api.NetworkPolicySpec{
					Ingress: []api.Rule{
						{
							Action: "Allow",
							Destination: api.EntityRule{
								NotNets: []string{"0.0.0.0/0"},
							},
						},
					},
				},
			}, false,
		),
		Entry("disallow a NotNets catch-all IPv6 CIDR match",
			&api.NetworkPolicy{
				ObjectMeta: v1.ObjectMeta{Name: "thing"},
				Spec: api.NetworkPolicySpec{
					Ingress: []api.Rule{
						{
							Action: "Allow",
							Source: api.EntityRule{
								NotNets: []string{"::0/0"},
							},
						},
					},
				},
			}, false,
		),
		Entry("disallow a NotNets catch-all IPv6 CIDR match - non-standard representation",
			&api.NetworkPolicy{
				ObjectMeta: v1.ObjectMeta{Name: "thing"},
				Spec: api.NetworkPolicySpec{
					Ingress: []api.Rule{
						{
							Action: "Allow",
							Source: api.EntityRule{
								NotNets: []string{"0:0:0::/0"},
							},
						},
					},
				},
			}, false,
		),
```

Again, thanks so much for contributing and for bearing with me!

dzacball:
Thank you for reviewing the PR @coutinhop! I'll come back here to continue the work on it shortly.

coutinhop:
hey @dzacball don't mean to be a pain, just a nudge so this doesn't lose momentum (we're almost there!)

please let me know if you need help with any of the requested changes!

coutinhop:
@dzacball btw, the CI in this branch would benefit from a `rebase` to master (the failing unrelated windows tests have been fixed there)

libcalico-go/lib/validator/v3/validator.go
dzacball: Thanks, I did just this!

dzacball:
> Thanks so much for considering my comments! This PR is looking great, we're only need to address a couple of points:
> 
> * please see my suggestion of parsing the v6 CIDR so that we can treat all ways of representing the v6 catch-all ("::/0", "0::0/0", "::0/0", etc)
> * we need to add some testing, could you add some test cases to https://github.com/projectcalico/calico/blob/e9e30206a97f05b83fc5ce8cea3b0bb0b5f9039c/libcalico-go/lib/validator/v3/validator_test.go#L3605
>    (you can run the test locally with `make -C libcalico-go/ ut GINKGO_FOCUS=Validator`)
> * could you also add a test case to https://github.com/projectcalico/calico/blob/e9e30206a97f05b83fc5ce8cea3b0bb0b5f9039c/felix/rules/policy_test.go#L106
>    ? I'm not 100% sure we should add IPv6 parsing here, we might be safe from the checks in the validator already...

- I merged master into this branch
- I used your suggestion for parsing the v6 CIDR
- I added the test cases you suggested to validator_test.go. Thanks!
- You suggested adding tests to policy_test.go: https://github.com/projectcalico/calico/blob/e9e30206a97f05b83fc5ce8cea3b0bb0b5f9039c/felix/rules/policy_test.go#L106
   I created a separate test table for these tests. At first I wanted to add these cases to `ruleTestData` like this:
   ```
   Entry("Negated catch-all IPv4 source net (should be filtered out)", 4,
  &proto.Rule{NotSrcNet: []string{"0.0.0.0/0"}},
  "")
  ```
  but it seems problematic, because current tests that run on `ruleTestData` usually expect all rules to generate at least 2 iptables rules: 
  https://github.com/projectcalico/calico/blob/master/felix/rules/policy_test.go#L211
  but our filtered rules return 0 rules. I may very well be wrong, there might be a better approach, I'm not familiar with ginkgo, and how these tests are constructed.

coutinhop (approved):
Thank you so much for your contribution @dzacball! I hope you don't mind, I've added a release note.

dzacball:
thank you @coutinhop for your work on this!

PR #10573, opened 2025-06-17
ebpf - tcx support

sridhartigera:
## Description

This PR adds support for tcx support.
1. Compile tc_preamble.c into tcx_preamble_ingress and tcx_preamble_egress with section names set to tcx/ingress and tcx/egress.
2. A simple test program tcx_test.c to load and attach to a veth to check if the underlying kernel supports tcx.
3. A new felix config ```BPFAttachType``` to specify `tc` or `tcx`. This defaults to `tcx` and if the kernel doesn't support tcx, it falls back to using ```tc```.
4. AttachType(```tc``` or ```tcx```) is set in the attachPoint and depending on the configuration, support from kernel, the corresponding preamble programs are attached to the ingress and egress hooks.
5. If  ```tcx``` is used, we don't need a qdisc and its skipped.
6. Attaching ```tcx``` returns a ```bpf_link``` and it is pinned to ```/sys/fs/bpf/tc/globals/tcx``` in the bpffs so that programs stay attached during reloads. The pinPath is of the format ```<interface_name>_<ingress/egress>```. Any dot in the interface name is removed.
7. When attaching ```tcx```, any existing ```tc``` programs along with qdiscs will be removed. This is useful when users upgrade and want to use qos features.
8. when the user wants to switch to ```tc```, any existing ```tcx``` programs are removed.
9. when switching from ebpf to iptables dataplane, the entire ```tcx``` pin directory is removed to completely detach the programs.
10. FVs on 24.04 runs with TCX whereas the NFT runs on 22.04 runs with Tc.
11. UT to test attach of tcx programs.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
eBPF - Add support for tcx which is available from 6.6 kernel. Introduced a new config option to set the attach type to tcx or tc.
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -846,6 +854,11 @@ type FelixConfigurationSpec struct {
 	//+kubebuilder:validation:Enum=Enabled;Disabled;L2Only
 	BPFRedirectToPeer string `json:"bpfRedirectToPeer,omitempty"`
 
+	// BPFAttachType in BPF mode, controls the attach type for the BPF Programs.
+	// [Default: TCX]
```
coutinhop: maybe it'd be good to mention that TCX is the default if supported, and there is a fallback to TC if not?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -846,6 +854,11 @@ type FelixConfigurationSpec struct {
 	//+kubebuilder:validation:Enum=Enabled;Disabled;L2Only
 	BPFRedirectToPeer string `json:"bpfRedirectToPeer,omitempty"`
 
+	// BPFAttachType in BPF mode, controls the attach type for the BPF Programs.
```
```suggestion
	// BPFAttachType controls how are the BPF programs at the network interfaces attached. By default `tcx` is used where available to enable easier coexistence with 3rd party programs. `tc` can force the legacy method of attaching via a qdisc. `tcx` falls back to `tc` if `tcx` is not available.
```

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -846,6 +854,11 @@ type FelixConfigurationSpec struct {
...
+	// [Default: TCX]
+	//+kubebuilder:validation:Enum=TC;TCX
+	BPFAttachType *BPFAttachOption `json:"bpfAttachType,omitempty" validate:"omitempty,oneof=TC TCX"`
```
tomastigera: I think lowercase tc/tcx is a bit more natural as it stems from the `tc` utility and I always saw it in lowercase. But it is a nit, with the fall back, nobody is ever going to change it ;-)

felix/bpf-gpl/tc_preamble.c
```diff
@@ -21,7 +21,15 @@ const volatile struct cali_tc_preamble_globals __globals;
 #define JUMP(idx) globals->data.jumps[JUMP_IDX(idx)]
 #define JUMP_DEBUG(idx) globals->data.jumps[JUMP_IDX_DEBUG(idx)]
 
-SEC("tc")
+#if defined(CALI_TCX_INGRESS)
```
tomastigera: Since the preamble is always the same I wonder whether it would make things a easier or more complicated libbpf with the explicit type instead of letting it figure it out from the section name. BWe would have just a single object. But perhaps it would allow us to customize the preambles in the future

felix/bpf/bpfdefs/defs.go
```diff
@@ -23,6 +23,7 @@ const (
 	GlobalPinDir = DefaultBPFfsPath + "/tc/globals/"
 	ObjectDir    = "/usr/lib/calico/bpf"
 	CtlbPinDir   = "ctlb"
+	TcxPinDir    = GlobalPinDir + "/tcx"
```
tomastigera: Isn't it going to be `DefaultBPFfsPath + "/tc/globals/" + "/tcx"` ? Shouldn't we have `DefaultBPFfsPath + "/tcx/globals/"` instead?

felix/bpf/tc/attach.go
```diff
@@ -117,18 +166,56 @@ func (ap *AttachPoint) AttachProgram() error {
...
 	if err != nil {
 		return err
 	}
-
 	return ap.detachPrograms(progsToClean)
 }
 
+func (ap *AttachPoint) DetachProgram() error {
+	err := ap.detachTcxProgram()
+	if err != nil {
```
tomastigera, line 214: should we perhaps propagate the enoexists up to here and resolve it here so it is a bit more obvious that we are not just going to spam logs in tc mode?

felix/bpf/tc/attach.go
```diff
@@ -117,18 +166,56 @@ func (ap *AttachPoint) AttachProgram() error {
...
+		log.Warnf("error detaching tcx program from %s hook %s : %s", ap.Iface, ap.Hook, err)
+	}
+	err = ap.detachTcProgram()
```
tomastigera, line 217: if tcx was a success should we still do this? probably does not hurt

felix/bpf/ut/attach_test.go
```diff
@@ -1069,6 +1070,113 @@ func TestCTLBAttach(t *testing.T) {
 	testCtlbAttach(true, true)
 }
 
+func TestAttachTcx(t *testing.T) {
```
tomastigera, line 1073: should we also test switching from/to tc mode?

felix/bpf/tc/attach.go
sridhartigera: ```detachTcxProgram``` is the main API to remove tcx pins and it is called when using the `tc` mode. I feel its better this way.

felix/bpf/ut/attach_test.go
sridhartigera: This test already does that. This test attaches tcx, switches to tc and then back to tcx.

felix/bpf/tc/attach.go
sridhartigera: libbpf throws the below warning ```failed to open link at``` when the link we are trying to open does not exist. Hence I changed to test if the pin path exists and return nil without opening the link.

felix/bpf-gpl/tc_preamble.c
sridhartigera: Thanks for the suggestion. I was able to set the attach_type when loading and the ```SEC``` doesn't really matter.

felix/bpf/libbpf/libbpf.go
```diff
@@ -167,6 +167,21 @@ func DetachClassifier(ifindex, handle, pref int, ingress bool) error {
...
 
+func (o *Obj) SetAttachTypeTcx(progName string, ingress bool) error {
+	attachType := C.BPF_TCX_EGRESS
+	if ingress {
+		attachType = C.BPF_TCX_INGRESS
+	}
+	return o.setAttachType(progName, attachType)
+}
+
+func (o *Obj) setAttachType(progName string, attachType int) error {
```
tomastigera: I think this one should be exported, the API is useful for the future

felix/bpf/tc/attach.go
```diff
@@ -81,18 +82,19 @@ func (ap *AttachPoint) Log() *log.Entry {
...
 	return obj, nil
 }
 
-func (ap *AttachPoint) attachTCXProgram() error {
+func (ap *AttachPoint) attachTCXProgram(binaryToLoad string) error {
 	logCxt := log.WithField("attachPoint", ap)
-	binaryToLoad := path.Join(bpfdefs.ObjectDir, fmt.Sprintf("tcx_%s_preamble.o", ap.Hook))
-	obj, err := ap.loadObject(binaryToLoad)
+	obj, err := ap.loadObject(binaryToLoad, func(obj *libbpf.Obj) error {
+		return obj.SetAttachTypeTcx("cali_tc_preamble", ap.Hook == hook.Ingress)
```
tomastigera: imo the logic from SetAttachTypeTcx should be here

tomastigera (changes requested):
I think the libbpf api could be nicer, not focused on tcx, but otherwise lgtm. Great change!

tomastigera (approved):

PR #10574, opened 2025-06-17
Move API server component from calico-apiserver to calico-system name…

vara2504:
…space

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Component Migration: To support a minimal footprint and simplify resource management, the calico-apiserver component and its associated resources have been moved from the calico-apiserver namespace to the calico-system namespace.
```

## Reminder for the reviewer

Other optional labels:


manifests/apiserver.yaml
```diff
@@ -1,23 +1,13 @@
 # This is a tech-preview manifest which installs the Calico API server. Note that this manifest is liable to change
 # or be removed in future releases without further warning.
 #
-# Namespace and namespace-scoped resources.
```
vara2504: @caseydavenport The operator already had a logic to create the calico-apiserver namespace for the apiserver in OSS, as seen in this [code](https://github.com/tigera/operator/blob/master/pkg/render/apiserver.go#L330).
So, why is there still a need to define the namespace creation in the manifest as well?

vara2504: Do we have any use cases where the operator might :
- Not create the calico-system namespace?
- Not create the necessary resources for the apiserver component?
- Wait to create the calico-system namespace until after the apiserver component has been deployed

caseydavenport: @vara2504 this manifest is for installing the API server for users who don't use the tigera/operator. 

In that case, they need to create the namespace themselves before installing resources into it. We might just want to leave this one as it is for now? 

otherwise users who are using this manifest will need to manually delete the existing resources and switch namespaces, and we'll want to write upgrade instructions for how to do so.

node/tests/k8st/deploy_resources_on_kind_cluster.sh
```diff
@@ -77,9 +77,8 @@ ${kubectl} apply -f $TEST_DIR/infra/calicoctl.yaml
 echo
 
 echo "Wait for Calico to be ready..."
-wait_pod_ready -l k8s-app=calico-node -n calico-system
 wait_pod_ready -l k8s-app=calico-kube-controllers -n calico-system
```
caseydavenport: We don't need this line any more, since the one below waits for all in the calico-system namespace.

node/tests/k8st/deploy_resources_on_kind_cluster.sh
```diff
@@ -77,9 +77,7 @@ ${kubectl} apply -f $TEST_DIR/infra/calicoctl.yaml
 echo
 
 echo "Wait for Calico to be ready..."
-wait_pod_ready -l k8s-app=calico-node -n calico-system
-wait_pod_ready -l k8s-app=calico-kube-controllers -n calico-system
-wait_pod_ready -l k8s-app=calico-apiserver -n calico-apiserver
+wait_pod_ready -n calico-system --all
```
coutinhop, line 80: @caseydavenport @vara2504 this function doesn't only call `kubectl wait`, but it also calls `kubectl get pod` (in the main codepath), and `kubectl describe` and `kubectl logs` (when there's a timeout), which do not support the `--all` flag.

sig-conformance and kind-cluster have been failing to set up because of this: https://tigera.semaphoreci.com/workflows/09ac29ff-1e59-4d60-8436-43b824de33b5?pipeline_id=fc452dcf-71de-48ca-b734-c3ce9b54ac59

```
Wait for Calico to be ready...
error: no matching resources found
Waiting for pod -n calico-system --all to be ready...
error: unknown flag: --all
See 'kubectl get --help' for usage.
```

Perhaps we should go with copilot's suggestion of multiple labels?

PR #10583, opened 2025-06-20
[release tool] fix operator image list

radTuti:
## Description

This fixes an issue inadvertently introduced by #10524.

To allow generating operator using dev images while creating a hashrelease, the registry in operator gets updated and images passed have their names modified to remove the `calico/` namespace prefix.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/internal/pinnedversion/pinnedversion.go
+++ b/release/internal/pinnedversion/pinnedversion.go
@@ -309,6 +309,7 @@ func normalizeComponent(name string, c registry.Component) registry.Component {
 	if img == "" {
 		img = name
 	}
+	c.Image = img
 	if strings.HasPrefix(img, calicoImageNamespace) {
 		c.Image = strings.TrimPrefix(img, calicoImageNamespace)
 	}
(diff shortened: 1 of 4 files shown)


release/internal/pinnedversion/pinnedversion.go
```diff
@@ -300,3 +299,19 @@ func RetrieveVersions(outputDir string) (version.Versions, error) {
 
 	return version.NewHashreleaseVersions(version.New(pinnedVersion.Title), pinnedVersion.TigeraOperator.Version), nil
 }
+
+// normalizeComponent normalizes the component image name.
+// It checks if the image is in the registry.ImageMap and replaces it with the mapped value.
+// If the image is not found in the map, it sets it to the name of the component.
+// The image is also stripped of the calico namespace prefix if it exists.
+func normalizeComponent(name string, c registry.Component) registry.Component {
+	img := registry.ImageMap[c.Image]
```
vara2504, line 308: @radTuti  @coutinhop  The previous code on the left uses name as the key instead of c.Image,
img := registry.ImageMap[name] 
In the new code, we are fetch the img with c.Image key which results in an empty img value. As a result, the assignment ends up using the default name "flexvol" which is incorrect here.

PR #10588, opened 2025-06-23
Fix flakiness of win cni-plugin CI

song-jiang:
## Description

Azure may assign a different public IP to a VM even after it is marked as ready.
This PR add function which attempts to SSH into the VM multiple times to ensure it is accessible.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/process/testing/winfv-cni-plugin/aso/vmss.sh
+++ b/process/testing/winfv-cni-plugin/aso/vmss.sh
@@ -153,8 +153,13 @@ function confirm-nodes-ssh() {
   show_connections
   retry-ssh "${MASTER_CONNECT_COMMAND} echo"
   retry-ssh "${WINDOWS_CONNECT_COMMAND} -Command 'echo'"
+
+  # Azure may assigned another public ip to the VM.
+  # So even the first batch of SSHes works, the ip could be updated later.
+  # Sleep and retry.
   echo "sleep 30 seconds..."
   sleep 30
+  show_connections
   retry-ssh "${MASTER_CONNECT_COMMAND} echo"
   retry-ssh "${WINDOWS_CONNECT_COMMAND} -Command 'echo'"
   echo "VMs can be accessed by ssh.";echo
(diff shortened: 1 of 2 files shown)


process/testing/winfv-cni-plugin/aso/vmss.sh
```diff
@@ -112,6 +113,51 @@ EOF
...
+# This function attempts to SSH into the VM multiple times to ensure it is accessible.
+function confirm-nodes-ssh() {
+  echo;echo "confirm nodes can be accessed by ssh..."
+  show_connections
+  retry-ssh "${MASTER_CONNECT_COMMAND} echo"
+  retry-ssh "${WINDOWS_CONNECT_COMMAND} -Command 'echo'"
+  echo "sleep 30 seconds..."
+  sleep 30
+  retry-ssh "${MASTER_CONNECT_COMMAND} echo"
+  retry-ssh "${WINDOWS_CONNECT_COMMAND} -Command 'echo'"
```
coutinhop: We are essentially doubling the retry-ssh() timeout here, right? Any specific reason to call it twice instead of using a 600 timeout and calling it just once (per node)?

song-jiang: What I found is that:
-  Azure assigned a public ip a.a.a.a and marked the VM is ready.
-  After a while (I actually don't know how long), Azure assigned another public ip to the VM.
So even the first batch of SSHes works, the ip could be updated later. That's why I added a sleep and retry ssh again.

coutinhop: Oh I see, so the _CONNECT_COMMANDs may change between the calls? Gotcha, makes sense... Could you please add a comment stating that?

coutinhop (approved):

PR #10590, opened 2025-06-23
CORE-11535: Fix svlogd rotated log files permissions

skoryk-oleksandr:
## Description

📝 **Summary**
This PR addresses an issue where svlogd creates rotated log files with `0744` permissions, making them executable. Executable permissions on log files are unnecessary and potentially confusing or risky. The fix modifies `svlogd.c` to use `0644` instead, ensuring log files are readable but not executable.

🐞 **Initial Problem**
In the Calico node container, rotated log files under `/var/log/calico/*** were being created with permissions `-rwxr--r-- (0744)`. These permissions are overly permissive for logs, which should not be executable.

**Jira**
https://tigera.atlassian.net/browse/CI-1790

✅ **Testing**
 - Rebuilt `svlogd` with patched permissions using sed during the image build step.
 - Deployed and tested in a `local Kind` cluster.
 - Verified that rotated log files are now created with `0644` permissions.
 - Confirmed that `svlogd` continues to handle log rotation and writing correctly.

📦 Components Affected
 - `calico/node` image (Docker build step)
 - Logging subsystem (svlogd behavior)

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/Dockerfile.amd64
+++ b/node/Dockerfile.amd64
@@ -76,6 +76,7 @@ RUN rpm -i ${IPSET_SOURCERPM_URL} && \
 # get it from the debian repos as the official website doesn't support https
 RUN curl -sfL https://ftp.debian.org/debian/pool/main/r/runit/runit_${RUNIT_VER}.orig.tar.gz | tar xz -C /root && \
     cd /root/admin/runit-${RUNIT_VER} && \
+    sed -i '208s/0744/0644/' src/svlogd.c && sed -i '285s/0744/0644/' src/svlogd.c && sed -i '375s/0744/0644/' src/svlogd.c && \
     package/compile
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS ubi


node/Dockerfile.amd64
```diff
@@ -72,10 +72,15 @@ RUN rpm -i ${IPSET_SOURCERPM_URL} && \
     yum-builddep -y --spec /root/rpmbuild/SPECS/ipset.spec && \
     rpmbuild -bb /root/rpmbuild/SPECS/ipset.spec
 
+# Copy the patch that adjusts svlogd log file permissions from 0744 to 0644.
+# This file is used in the next step to apply the patch during the build process.
+COPY patches/svlogd_use_0644_permission_instead_of_0744.patch /svlogd_use_0644_permission_instead_of_0744.patch
+
```
stevegaossou, line 78: Are there any concerns around us applying this patch directly ourselves, in case in the future we update the version of runit that we download and use?

stevegaossou (commented):
Just have a single comment about thinking ahead to a scenario where we want to update the version of runit that we use. 

In that situation, how will applying direct patch for `svlogd` affect the update?

node/Dockerfile.amd64
skoryk-oleksandr: Good point.

If svlogd.c changes in a newer runit version, the patch might fail to apply — but that will be immediately caught during the build, since the patch step will fail right after we bump the version.

In that case, we’ll just need to update the patch. The change is minimal (just replacing 0744 with 0644), so adjusting it should be quick and straightforward.

coutinhop (approved):

PR #10600, opened 2025-06-25
Oleks local as number

skoryk-oleksandr:
## Description

✨ **Summary**
This PR introduces a new feature: support for specifying a **localASNumber** field in the **BGPPeer** resource. This allows operators to configure a distinct **local ASN** per **BGP peer**, enabling advanced BGP topologies such as eBGP between Calico nodes, eBGP with external routers, and route reflector setups that require custom local ASN configuration.

✅ **Why this should be merged**
Previously, Calico used the node's **global ASN** for all BGP peerings. This limited flexibility in network designs where **per-peer local ASN** is required. With this change, users can now specify a **localASNumber** in each **BGPPeer** to override the global setting on a per-peer basis.

🧪 **Testing done**
Manual testing in a Kind cluster with multiple FRR containers and different local ASNs per peer.

Verified that:

 - BIRD configuration files include the correct local as directives
 - Peers come up successfully using the specified local ASN
 - No regressions when localASNumber is omitted (defaults to global ASN)
 - Confirmed with both calico-node and calico-typha deployments.

📦 **Affected components**
 - confd
 - BIRD template rendering logic
 - BGPPeer controller logic in calico/node

## Todos

- [ ] Tests
**- [ ] Documentation (TBD!!!!!!!!!!!!!!!!!!!!!)
- [ ] Release note (TBD!!!!!!!!!!!!!!!!!!!)**

## Release Note **(TBD!!!!!!!!!!!!!!!!!)**

```release-note
TBD!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/bgppeer.go
```diff
@@ -71,6 +71,11 @@ type BGPPeerSpec struct {
 	// +optional
 	ASNumber numorstring.ASNumber `json:"asNumber,omitempty"`
 
+	// The optional Local AS Number to use when peering with this remote peer.
+	// If not specified, the node’s default ASN will be used.
```
mazdakn:
```suggestion
	// If not specified, the AS Number defined in default BGPConfiguration will be used.
```

confd/tests/mock_data/calicoctl/explicit_peering/local-as-global/delete.yaml
```diff
@@ -5,24 +5,28 @@ metadata:
...
+  sourceAddress: None
 
 ---
 kind: BGPPeer
 apiVersion: projectcalico.org/v3
 metadata:
   name: bgppeer-2
 spec:
-  peerIP: 10.192.0.4
+  peerIP: 10.192.0.1:166
```
mazdakn, line 16: Why this has changed? It's better to keep the changes to minimum if not needed.

confd/tests/mock_data/calicoctl/explicit_peering/local-as-global/delete.yaml
```diff
@@ -5,24 +5,28 @@ metadata:
 spec:
   peerIP: 10.192.0.3
   asNumber: 64567
+  sourceAddress: None
```
mazdakn, line 8: What's the reason for adding this?

confd/tests/mock_data/calicoctl/explicit_peering/local-as-global/delete.yaml
```diff
@@ -5,24 +5,28 @@ metadata:
...
 kind: BGPPeer
 apiVersion: projectcalico.org/v3
 metadata:
   name: bgppeer-3
 spec:
-  peerIP: 10.192.0.4
+  peerIP: 10.192.0.1
   asNumber: 64567
+  localASNumber: 65001
+  numAllowedLocalASNumbers: 1
```
mazdakn, line 29: Generally speaking, it's cleaner to keep the changes in this input file to minimum. I expect to see only `localASNumber` or `asNumber` changes. 
It's different for the new tests, where we need to provide complete resource. But for resources used by other tests, it's better to keep the changes to minimum.

confd/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -408,6 +411,9 @@ protocol bgp Node_{{$id}} from bgp_template {
 {{- else }}
   ttl security off;
   multihop;
+{{- end }}
+{{- if ne ($data.local_as_num) "0"}}
+  local as {{$data.local_as_num}};
```
mazdakn, line 419: We should also do the same for global and node specific local BGP peers, used for nested BGP. I can't think of any use case, but setting local ASN is a way to override the default Calico ASN. Given that we are allowing the override for other external peers, we normalise it for all external peers.

confd/tests/mock_data/calicoctl/explicit_peering/local-as/delete.yaml
```diff
@@ -4,7 +4,8 @@ metadata:
   name: bgppeer-1
 spec:
   peerIP: 10.192.0.3
-  asNumber: 64566
+  asNumber: 64512
```
mazdakn, line 7: Does this value have any effect? I wonder why the previous value, i.e. **64566** did not lead to test failure.

confd/etc/calico/confd/templates/bird.cfg.template
mazdakn: Please also update ones of the related tests, to check the result.

confd/pkg/backends/calico/client.go
```diff
@@ -933,6 +943,10 @@ func (c *client) nodeAsBGPPeers(nodeName string, v4 bool, v6 bool, v3Peer *apiv3
 				log.WithError(err).Warningf("Problem parsing global AS number %v for node %v", asNum, nodeName)
 			}
 		}
+
+		if v3Peer.Spec.LocalASNumber != nil {
+			peer.LocalASNum = *v3Peer.Spec.LocalASNumber
+		}
```
mazdakn: I don't think we need to add this here. This function is used for BGPPeers where `peerSelector` is set. These peers are basically iBGP peering: https://docs.tigera.io/calico/latest/reference/resources/bgppeer

Base on our doc, if `peerSelector` is set:
```
Selector for the remote nodes to peer with. When this is set, the peerIP and asNumber fields must be empty.
```

As such, it does not make sense to change local ASN.

confd/pkg/backends/calico/client.go
```diff
@@ -1946,6 +1960,9 @@ func (c *client) setPeerConfigFieldsFromV3Resource(peers []*bgpPeer, v3res *apiv
 
 	for _, peer := range peers {
 		peer.Password = password
+		if v3res.Spec.LocalASNumber != nil {
+			peer.LocalASNum = *v3res.Spec.LocalASNumber
+		}
```
mazdakn: This also needs to be removed due to previous comment. This function is called for iBGP peers as well.

confd/tests/mock_data/calicoctl/explicit_peering/local-as-global/delete.yaml
skoryk-oleksandr: I thought resources in `delete.yaml` should match to the resources we create in the `input.yaml`. In the `input.yaml` we have `peerIP: 10.192.0.1:166` https://github.com/projectcalico/calico/blob/89fbe7835b2cac37cd3a16e61277122141822e5e/confd/tests/mock_data/calicoctl/explicit_peering/local-as-global/input.yaml#L27

confd/tests/mock_data/calicoctl/explicit_peering/local-as-global/delete.yaml
skoryk-oleksandr: Shouldn't resources in `delete.yaml` match to the resources in the `input.yaml`? 
https://github.com/projectcalico/calico/blob/89fbe7835b2cac37cd3a16e61277122141822e5e/confd/tests/mock_data/calicoctl/explicit_peering/local-as-global/input.yaml#L19

confd/tests/mock_data/calicoctl/explicit_peering/local-as-global/delete.yaml
mazdakn: I see. there is a diff for that resource between those 2 files.
Make sense, thanks.

api/pkg/apis/projectcalico/v3/bgppeer.go
skoryk-oleksandr: Done

confd/etc/calico/confd/templates/bird.cfg.template
skoryk-oleksandr: Done

confd/pkg/backends/calico/client.go
skoryk-oleksandr: Done

confd/pkg/backends/calico/client.go
skoryk-oleksandr: Done

confd/tests/test_suite_common.sh
```diff
@@ -599,8 +599,10 @@ execute_tests_daemon() {
         run_individual_test 'explicit_peering/route_reflector'
         run_individual_test 'explicit_peering/keepnexthop'
         run_individual_test 'explicit_peering/keepnexthop-global'
-	run_individual_test 'explicit_peering/local-as'
-	run_individual_test 'explicit_peering/local-as-global'
+	      run_individual_test 'explicit_peering/local-as'
+	      run_individual_test 'explicit_peering/local-as-global'
```
mazdakn: nit: alignment in these 2 lines seems wrong.

confd/tests/mock_data/calicoctl/explicit_peering/local-as-global-ipv6/delete.yaml
```diff
@@ -0,0 +1,37 @@
...
+  peerIP: 10.192.0.4
+  asNumber: 64567
+
+---
+kind: IPPool
+apiVersion: projectcalico.org/v3
+metadata:
+  name: ippool-1
+spec:
+  cidr: 192.168.0.0/16
```
mazdakn, line 26: There is a mismatch between this file and the input one. However, it seems not important for the delete operation.

mazdakn (approved):
LGTM. There is just a couple of nits. Also add a release note for this.
Please also update the copyright statement for the changed file. In this PR you only need to update the statement in `confd/pkg/backends/calico/client.go` to:

```
Copyright (c) 2017-2025 Tigera, Inc. All rights reserved.
```

PR #10602, opened 2025-06-25
[BPF] scale trampoline stride in policy programs

tomastigera:
## Description

    When jit_harden is enabled, some long jump may get over the allowed
    distance of +/-15bit when rewritten by the kernel. When this happens, we
    may retry and recompile the policy programs with the trampolines that
    make the jumps shorter. If that still fails we may retry with even more
    dense trampolines.
    
    We remember the newly found stride for the given node so that we do not
    need to keep figuring out for other policy programs. It is likely that
    many policies would face the same issue.
    
    Note that the hardedning already increases the amount of executed
    instructions and adding some extra trampolines have probably much
    smaller effect than in a non-hardened program.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/asm/asm.go
+++ b/felix/bpf/asm/asm.go
@@ -695,6 +695,9 @@ func (b *Block) addInsnWithOffsetFixup(insn Insn, targetLabel string) {
 }
 
 func (b *Block) maybeWriteTrampoline(nextInsn Insn) {
+	if !b.trampolinesEnabled {
+		return
+	}
 	if len(b.insns)-b.lastTrampolineAddr < trampolineInterval {
 		return
 	}
--- a/felix/bpf/asm/asm.go
+++ b/felix/bpf/asm/asm.go
@@ -381,6 +381,7 @@ type Block struct {
 	lastTrampolineAddr int
 	deferredErr        error
 	NumJumps           int
+	trampolineStride   int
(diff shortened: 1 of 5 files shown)


sridhartigera:
LGTM to me. I feel it will be better to run the bpf UTs, FVs with jit_harden set to 2. WDYT?

felix/bpf/ut/pol_prog_test.go
```diff
@@ -2289,19 +2306,39 @@ func TestExpandedPolicyPrograms(t *testing.T) {
...
 }
 
+func execSysctl(param, value string) error {
+	cmd := exec.Command("sysctl", "-w", fmt.Sprintf("%s=%s", param, value))
+	_, err := cmd.CombinedOutput()
+	return err
+}
+
 var testExpanders = []testExpander{
+	/*
```
sridhartigera: Is this intended?

tomastigera: nope :man_facepalming:

sridhartigera (approved):
LGTM. Thanks

sridhartigera (approved):

PR #10607, opened 2025-06-26
CORE-11601: Initializing BPF networking in `mount-bpffs` init container

lucastigera:
## Description

**Context**

To drive adoption of Calico’s BPF dataplane, we aim to simplify its setup by automating current manual steps. The goal is to support default BPF installations on clusters that iinitially use kube-proxy (e.g., via kubeadm) without requiring user configuration.

This PR extends the `mount-bpffs` init container to handle the initial BPF setup required for network communication when BPF is enabled and kube-proxy is disabled.

**Motivation**
Running this logic in the init container avoids potential deadlocks during cluster or node restarts. Since BPF programs are lost on reboot and kube-proxy is disabled, Calico would be unable to reach the API server if this setup were done in calico-node.

**Changes**
The init container now:
- Parses injected environment variables to detect if the node uses IPv4, IPv6, or dual-stack.
- Creates and populates NAT maps:
  - Frontend: Kubernetes service IPs.
  - Backend: API server endpoints.
- Installs a connect-time load balancer (CTLB) using `bpfnat.InstallConnectTimeLoadBalancer`.

CTLB is chosen over TC programs because it works with host-networked processes and avoids additional workaround complexity.

The `node/Makefile` was updated to include BPF dependencies, since the `nodeinit` package now has some UTs that require them.

**Error Handling**
If any part of this setup fails (e.g., parsing, map creation, or CTLB installation), the init container logs the error and exits, blocking Calico startup as expected.

**Tests**
- FV tests will not be performed, as they basically cover Felix’s functionality.
- This change will be validated by the existing E2E tests when Banzai attempts to create a cluster that meets the automation requirements in BPF mode.

**Next step**
A follow-up PR will rename the `mount-bpffs` init container to `bpf-bootstrap` to better reflect its expanded responsibilities beyond mounting bpffs, as it now performs essential BPF initialization required for cluster networking.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
When installing Calico in BPF mode in a cluster using kube-proxy, the Operator now handles all required configuration automatically - no additional manual steps are required.
```

## Reminder for the reviewer

Other optional labels:


node/pkg/nodeinit/calico-init_linux.go
```diff
@@ -20,17 +20,30 @@ package nodeinit
...
 	"os/exec"
+	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
+	"github.com/google/gopacket/layers"
 	"github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/calico/felix/bpf/bpfdefs"
```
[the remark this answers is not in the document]
lucastigera: It is used in the file.
For example:
```
		if mountPoint == bpfdefs.DefaultBPFfsPath && fs == "bpf" {
			logrus.Info("BPF filesystem is mounted.")
			return nil
		}
```

```
	cgroupV2Path := bpfdefs.GetCgroupV2Path()
```

node/pkg/nodeinit/calico-init_linux.go
```diff
@@ -50,6 +62,156 @@ func Run(bestEffort bool) {
...
+	servicesIPPort, err := parseCommaSeparatedIPPorts(serviceAddr)
+	if err != nil {
+		return nil, err
+	}
+	endpointsIPPort, err := parseCommaSeparatedIPPorts(endpointAddrs)
+	if err != nil {
+		return nil, err
+	}
+
+	countIPv4, countIPv6 := countIPv4AndIPv6(append(endpointsIPPort, servicesIPPort...))
```
tomastigera: I am not sure this logic makes sense to me. You make the count over endpoints and services, but then you use the count as the endpoints count ...

node/pkg/nodeinit/calico-init_linux.go
```diff
@@ -50,6 +62,156 @@ func Run(bestEffort bool) {
...
+			err = bpfMaps.V4.FrontendMap.Update(
+				bpfnat.NewNATKey(service.IP, service.Port, uint8(layers.IPProtocolTCP)).AsBytes(),
+				bpfnat.NewNATValue(id, uint32(countIPv4), 0, 0).AsBytes(),
+			)
+			if err != nil {
+				logrus.WithError(err).Error("Failed to add IPv4 set entry in the frontend map.")
+				return nil, err
+			}
+			// Add the endpoints to the backend map.
+			for index, endpoint := range endpointsIPPort {
```
tomastigera: why not to flip this, first write the backends, count them and then write the front end.

node/pkg/nodeinit/calico-init_linux.go
```diff
@@ -50,6 +62,156 @@ func Run(bestEffort bool) {
...
+				bpfnat.NewNATKeyV6(service.IP, service.Port, uint8(layers.IPProtocolTCP)).AsBytes(),
+				bpfnat.NewNATValueV6(id, uint32(countIPv6), 0, 0).AsBytes(),
+			)
+			if err != nil {
+				logrus.WithError(err).Error("Failed to add IPv6 set entry in the frontend map.")
+				return nil, err
+			}
+			// Add the endpoints to the backend map.
+			for index, endpoint := range endpointsIPPort {
+				if !endpoint.IsIPv4 {
```
tomastigera: you may unify the logic, here you can just use `service.IsIPv4 == endpoint.IsIPv4` you just need to select the right map and the set of functions that go with it.

node/pkg/nodeinit/calico-init_linux.go
```diff
@@ -50,6 +62,156 @@ func Run(bestEffort bool) {
...
+		logrus.WithError(err).Error("Failed to create bpf maps.")
+		return nil, err
+	}
+
+	id := uint32(0)
+	for _, service := range servicesIPPort {
+		if service.IsIPv4 {
+			err = bpfMaps.V4.FrontendMap.Update(
+				bpfnat.NewNATKey(service.IP, service.Port, uint8(layers.IPProtocolTCP)).AsBytes(),
+				bpfnat.NewNATValue(id, uint32(countIPv4), 0, 0).AsBytes(),
```
tomastigera: we do not know whether any of the endpoints are local, but I am pretty sure it does not matter with ctlb :heavy_check_mark:

tomastigera (changes requested):

node/pkg/nodeinit/calico-init_linux.go
lucastigera: You are right. I changed the logic in the previous commit and forgot update the parameters. Thanks for catching it.

node/pkg/nodeinit/calico-init_linux.go
lucastigera: Yeah! Flipping looks to be a good idea.

node/pkg/nodeinit/calico-init_linux.go
lucastigera: I'm not sure if I understood this comment.
Does it still make sense after flipping the approach to first create the back end maps?

tomastigera: Just that the code for ipv4 and ipv6 is the same so perhaps a single function parametrized by the map and the key/value functions would be sufficient

node/pkg/nodeinit/calico-init_linux.go
```diff
@@ -50,6 +62,151 @@ func Run(bestEffort bool) {
 			os.Exit(3)
 		}
 	}
+	serviceAddr := os.Getenv("KUBERNETES_SERVICE_HOST_PORT")
```
sridhartigera: Should this be ```KUBERNETES_SERVICE_IPS_PORTS```, since we are talking about the actual service ip and port.

node/pkg/nodeinit/calico-init_linux.go
```diff
@@ -50,6 +62,151 @@ func Run(bestEffort bool) {
...
+	if err != nil {
+		return nil, err
+	}
+	endpointsIPPort, err := parseCommaSeparatedIPPorts(endpointAddrs)
+	if err != nil {
+		return nil, err
+	}
+
+	hasIPv4, hasIPv6 := hasIPv4AndIPv6(servicesIPPort)
+
```
sridhartigera, line 92: Should we validate service IPs with endpoints. I mean we shouldn't end up with a case where there is a ipv4 service ip but endpoints are only ipv6, which is a mismatch. We need not even create the maps if there is invalid data. WDYT? Is it an overkill?

sridhartigera: Better to do this in operator, I guess.

lucastigera: I think it's a good validation, and we can implement it on both sides - Operator and Calico.

One question: if a service doesn't have the related endpoints, should we just ignore the service, or should we panic and set the system to a degraded state?

node/pkg/nodeinit/calico-init_linux.go
lucastigera: Created two parameterized functions - one to update the frontend map and other to update the backend map.

node/pkg/nodeinit/calico-init_linux.go
lucastigera: Updated.

node/pkg/nodeinit/calico-init_linux.go
lucastigera: As discussed in person, this verification will be delegate to the Operator.

node/pkg/nodeinit/calico-init_test.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+		{
+			"192.168.0.1:6443,[2001:db8::1]:6443",
+			[]IPPort{
+				{net.ParseIP("192.168.0.1"), 6443, true},
+				{net.ParseIP("2001:db8::1"), 6443, false},
+			},
+			false,
+		},
+		{"1.1.1.1:port", nil, true},
```
tomastigera, line 50: what if there is no port?

lucastigera: I believe this would fall under the "made-up-addr" test case.
However, I added two new cases to explicitly cover:
-  an IP without a port, and 
- a port without an IP.

PR #10614, opened 2025-06-27
[CORE-11385] ebpf: add support for NATOutgoingExclusions

coutinhop:
## Description

Add support for the NATOutgoingExclusions field from FelixConfiguration to the ebpf dataplane. When set to IPPoolsAndHostIPs, the config is propagated through to the bpf endpoint manager into a CALI_GLOBALS_NATOUTGOING_EXCLUDE_HOSTS global flag, which when set exclude hosts from being SNAT'ed.

Also add a TestNATOutExcludeHosts() ut to verify this.

NATOutgoingExclusions was added, along with iptables/nftables dataplanes support, on
https://github.com/projectcalico/calico/pull/8961. This PR adds the ebpf dataplane equivalent support.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
ebpf: support for the new Felix config param natOutgoingExclusions, which allows for configuring which type of destinations are excluded from being masqueraded (SNAT).
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/tc.c
```diff
@@ -512,7 +512,13 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 			if (rt) {
 				flags = rt->flags;
 			}
-			if (!(flags & CALI_RT_IN_POOL) && !cali_rt_flags_local_host(flags)) {
+            // Don't perform SNAT if either:
+            // - packet is destined to an address in the same IP pool;
```
```suggestion
            // - packet is destined to an address in a IP pool;
```
doesn't have to be the same, we do not even have the info that it is the same or not

felix/bpf-gpl/tc.c
```diff
@@ -512,7 +512,13 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
...
+            // - packet is destined to local host; or
+            // - packet is destined to a host and the CALI_GLOBALS_NATOUTGOING_EXCLUDE_HOSTS global flag is set
+			if (!(flags & CALI_RT_IN_POOL) &&
+                    !cali_rt_flags_local_host(flags) &&
+                    !(cali_rt_flags_host(flags) && (GLOBAL_FLAGS & CALI_GLOBALS_NATOUTGOING_EXCLUDE_HOSTS))) {
```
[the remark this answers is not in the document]
tomastigera: In a way, I would side here with the copilots suggestion `should_perform_nat_outgoing()` which can then have a more readable condition like 

```
if in-pool return false
if is-host {
    if local return false // feel free to add cali_rt_flags_local(t) to routes.h
    if GLOBAL_FLAGS & CALI_GLOBALS_NATOUTGOING_EXCLUDE_HOSTS return false
}
return true
```

I was trying to figure out how to make the condition easier to read and/or faster. Perhaps it could be in the routes.h as `rt_should_perform_nat_outgoing()`

felix/bpf/ut/nat_test.go
```diff
@@ -3712,3 +3712,53 @@ func TestNATNodePortV6(t *testing.T) {
...
+	Expect(err).NotTo(HaveOccurred())
+	// Insert a route for the dest ip (flagged as remote host).
+	rtKey = routes.NewKey(dstV4CIDR).AsBytes()
+	rtVal = routes.NewValue(routes.FlagsRemoteHost).AsBytes()
+
+	err = rtMap.Update(rtKey, rtVal)
+	Expect(err).NotTo(HaveOccurred())
+	defer resetRTMap(rtMap)
+	_ = conntrack.NewKey(uint8(ipv4.Protocol),
+		ipv4.SrcIP, uint16(udp.SrcPort), ipv4.DstIP, uint16(udp.DstPort))
```
tomastigera: you never use this - you could use that key to make a lookup in ct map to see whether `v4.FlagNATOut` is set on the value. But I think testing the mark is good enough.

tomastigera (commented):
some nits but LGTM

felix/bpf-gpl/tc.c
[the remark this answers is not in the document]
coutinhop: good point! I added `rt_flags_should_perform_nat_outgoing()`, but I couldn't access `GLOBAL_FLAGS` trivially from it, so I passed a `bool` arg, not sure if that's the best way... let me know if there's a better way to do this
thanks!

felix/bpf-gpl/tc.c
coutinhop: updated, thanks

felix/bpf/ut/nat_test.go
coutinhop: thanks! removed

felix/bpf-gpl/tc.c
```diff
@@ -512,7 +512,8 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 			if (rt) {
 				flags = rt->flags;
 			}
-			if (!(flags & CALI_RT_IN_POOL) && !cali_rt_flags_local_host(flags)) {
+            bool exclude_hosts = (GLOBAL_FLAGS & CALI_GLOBALS_NATOUTGOING_EXCLUDE_HOSTS);
```
sridhartigera: Alignment issue?

sridhartigera (approved):
LGTM. Just one alignment to fix.

PR #10632, opened 2025-07-02
Fixup parent-release-branch script

danudey:


hack/find-parent-release-branch.sh
```diff
@@ -8,8 +8,45 @@ best=""
 : "${release_prefix:=release-v}"
 : "${git_repo_slug:=projectcalico/calico}"
 
-remote=$(git remote -v | grep "${git_repo_slug}.*fetch" | cut -f1 )
-echo "Git remote: ${git_repo_slug} -> ${remote}" >&2
+current_branch=$(git branch --show-current)
+
+# Some of our team members use "tigera" in their github usernames, which breaks
+# this script for specifically those members when `git_repo_slug` starts with `tigera`;
```
fasaxc: The comment makes it sound like the script is currently broken; but you've now fixed it. Think you could go a bit lighter ont he comment "Note that we need to handle the case where the user's github username has the repo slug as a suffix; we do that by matching on ... "

You could probably grep for ":" instead of "any non alpha numeric"?

danudey: Will clean this up, sure!

danudey: Updated this to include both `:` and `/`, since both could appear in a valid Github URL, but no other characters as there are potentially valid non-alphanumeric characters that could appear.

hack/find-parent-release-branch.sh
```diff
@@ -8,8 +8,53 @@ best=""
...
+# just include matches with one of [:/] in front of it. This covers both of the
+# delimiters used in these two common Git URL forms:
+#
+#                     ↓
+#   proto://github.com/foo/bar.git
+#       git@github.com:foo/bar.git
+#
+# Filtering to specifically these characters prevents us from misinterpreting valid
+# non-alphanumeric characters (e.g. test_tigera/somerepo.git).
+
```
fasaxc:
```suggestion
# Find the appropriate remote, handling common forms of git URL:
#
#  -  proto://github.com/foo/bar.git
#  -  git@github.com:foo/bar.git
#
#  And avoiding matching prefixes of the desired slug:
#
#   Match:  git@github.com:tigera/somerepository.git
#   But not: git@github.com:userbootigera/somerepository.git
```
Comment still seems wrong and overexplained/generally using too much vertical space for the value it provides.

fasaxc (changes requested):

PR #10658, opened 2025-07-10
Avoid IPAM handle leak when context expires during assignment.

fasaxc:
## Description

- Don't include time to acquire IPAM lock in the timeout. Under heavy load, IPAM allocation calls could get stacked up. As the lock acquisition time approaches 90s, it gets into an equilibrium where each IPAM call succeeds in creating a handle but times out while doing the actual assignment.  This leaks handles at a rapid rate.

  Speeding up IPAM requires a significant rework so this is a pragmatic fix for now.

- Belt and braces: use extended context timeouts for cleanup operations. If the context expires during IPAM allocation and hence we need to clean up a handle, use a fresh context so that the cleanup doesn't immediately time out.

- Clean up some unhandled errors and remove a duplicate conditional.
## Related issues/PRs

CORE-11632
CI-1807

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that IPAM allocation could leak handles when many workloads are scheduled to the same node at the same time, causing timeouts by "thundering herd".
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/pkg/ipamplugin/ipam_plugin.go
+++ b/cni-plugin/pkg/ipamplugin/ipam_plugin.go
@@ -166,8 +166,6 @@ func cmdAdd(args *skel.CmdArgs) error {
 	}
 
 	ctx := context.Background()
-	ctx, cancel := context.WithTimeout(ctx, 90*time.Second)
-	defer cancel()
 
 	r := &cniv1.Result{}
 	if ipamArgs.IP != nil {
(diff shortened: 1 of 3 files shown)


cni-plugin/pkg/ipamplugin/ipam_plugin.go
```diff
@@ -224,12 +228,14 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 		logger.Infof("Calico CNI IPAM request count IPv4=%d IPv6=%d", num4, num6)
 
-		v4pools, err := utils.ResolvePools(ctx, calicoClient, conf.IPAM.IPv4Pools, true)
+		rctx, cancel := context.WithTimeout(ctx, 90*time.Second)
```
song-jiang: I suggest moving the original code that creates a fresh context (ctx) here, instead of using rctx and relying on the ctx variable defined earlier.

```
ctx := context.Background()
ctx, cancel := context.WithTimeout(ctx, 90*time.Second)
defer cancel()
```

fasaxc: I removed the `ctx` variable and created contexts directly from `context.Background()`.  I split the two contexts here as suggested by the bot; 30s each.

caseydavenport (approved):

Issue #10640, opened 2025-07-07
Node with both external and internal IP may respond to external traffic with ICMP from the internal IP


wzrdtales:
## Expected Behavior

The traffic answers as expected and the connection establishes post SSH2_MSG_KEX_ECDH_REPLY all the time.

## Current Behavior

NodePort gets stuck (sometimes/ most of the times) on SSH2_MSG_KEX_ECDH_REPLY when eBPF and DSR are enabled and the workload runs on a different node.

When making sure the entrance point is the same node as the container receiving and answering it works fine. Which however is of course not the goal

## Steps to Reproduce (for bugs)

1. create nodeport with externalIP and listening port 22
2. point it to ssh in a container
3. make sure the container is on a different node than where the traffic enters the cluster
4. try to connect multiple times
5. at some random occasion it will get stuck

Here an example of a nodeport we configure

```yaml
apiVersion: v1
kind: Service
metadata:
  name: service-ssh
  namespace: some
spec:
  clusterIP: 10.43.1.12
  clusterIPs:
    - 10.43.1.12
  externalIPs:
    - publicipaddress
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
    - IPv4
  ipFamilyPolicy: SingleStack
  ports:
    - name: ssh
      nodePort: 32014
      port: 22
      protocol: TCP
      targetPort: 2222
  selector:
    app: your
    release: app
  sessionAffinity: ClientIP
  sessionAffinityConfig:
    clientIP:
      timeoutSeconds: 600
  type: NodePort
status:
  loadBalancer: {}
```

## Context

DSR

## Your Environment

* Calico version 3.30
* Calico dataplane (iptables, linux.)
* Orchestrator version (kubernetes (rke2, containerd)):
* Operating System and version: Linux k8s-sub1-2 5.15.0-127-generic #137-Ubuntu
* Link to your project (optional):

wzrdtales:
We checked multiple things and couldn't find an exact source of the problem yet.

One thing caught our eye though. In the debug log it shows

```
 mini_scheduler*-2178204 [006] ..s1. 163260.609559: bpf_trace_printk: calif30859068e4-E: CT: result: 0x6003
          <idle>-0       [005] .Ns.. 163260.609560: bpf_trace_printk: enp1s0----------I: UDP; ports: s=45985 d=4789
[...]

Issue #4439, opened 2021-03-03
Huge performance penalty when using eBPF


r3pek:
I'm running a 3 node cluster (1 master and 2 worker nodes) with all of them connected via a private network `192.168.10.0/24` with mtu set on `1450` (it's a VPS running on Hetzner). All 3 nodes have a public facing interface with a public IP address. all communication between is done via private lan.

The current test i'm doing is deploying a sftp-server on one worker node and upload a file to it (connecting via the public interface) and check the upload speed when the connection is made for each of the worker nodes, repeating the test on eBPF and non-eBPF modes.

## Expected Behavior

I expected to see some performance penalty when using eBPF but not as much as observed.

## Current Behavior

When connecting to the node that doesn't have the pod running, I see this:
eBPF speeds:
`Downloads/android-studio-ide-183.5692245-linux.tar.gz                                                                                                                            0% 2496KB  67.9KB/s 4:20:01 ETA`

non-eBPF speeds:
`Downloads/android-studio-ide-183.5692245-linux.tar.gz                                                                                                                            6%   66MB  16.3MB/s   00:59 ETA`

When connecting to the same node where the pod is running, speeds are almost the same on both modes (around 13/18 MB/s)

## Possible Solution

Don't really know :(

## Steps to Reproduce (for bugs)

I think that just setting up the same environment should do it

## Context

I was truly expecting some performance penalty, but not on this orders of magnitude. 16MB/s -> 60KB/s
The main reason for using eBPF was for maintaining the source IP when connection from external sources so that I can have services on the cluster that actually do something useful with that information (anti-spam for example), but since most of my services don't scale and only run one instance, i setup a DNS pointing to both nodes of the cluster and just let the cluster handle where the pod actually is, avoiding the dns-vudu and would have to be made to always have the correct (public) ip of the correct node running the pod.
This might not even be a bug, but a consequence of how eBPF works and in that case, i'll just have to drop the requirement of having the source ip available.

Heres my felixconfiguration:
```
# ./calicoctl get felixconfiguration default -o yaml
apiVersion: projectcalico.org/v3
kind: FelixConfiguration
metadata:
  creationTimestamp: "2021-02-16T00:46:06Z"
  name: default
  resourceVersion: "2665639"
  uid: a395f6ab-d0d5-4ba7-a54b-a797504970e7
spec:
  bpfEnabled: false
  bpfExternalServiceMode: Tunnel
  bpfLogLevel: ""
  logSeverityScreen: Info
[...]

PR #10660, opened 2025-07-10
[BPF] fix ICMP replies when host has multiple IPs

tomastigera:
Previously, we would only reply with the IP of the interface used as k8s internel IP. That may not be correct if that IP is private and the node has an external IP assigned and the client reaches out via that public IP.

For PMTU we respond before we do any NAT and so if the dest IP of the original packet was a host IP, use that same ip as the source for the icmp seponse.

fixes https://github.com/projectcalico/calico/issues/10640
fixes https://github.com/projectcalico/calico/issues/4439

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: fixes ICMP response source IP when nodes have multiple IPs assigned
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/icmp4.h
+++ b/felix/bpf-gpl/icmp4.h
@@ -84,18 +84,15 @@ static CALI_BPF_INLINE int icmp_v4_reply(struct cali_tc_ctx *ctx,
 
 	ctx->ipheader_len = 20;
 
-#ifdef CALI_PARANOID
-	/* XXX verify that ip_orig.daddr is always the node's IP
-	 *
-	 * we only call this function because of NodePort encap
-	 */
-	if (ip_orig.daddr != HOST_IP) {
-		CALI_DEBUG("ICMP v4 reply: ip_orig.daddr != HOST_IP 0x%x", ip_orig.daddr);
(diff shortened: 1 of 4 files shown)


wzrdtales:
Great!

`cali_rt_flags_local_host` under which condition is this flag set?

wzrdtales:
confirmed

image

tomastigera:
> Great!
> 
> `cali_rt_flags_local_host` under which condition is this flag set?

when the IP is assigned to one of the host devices we manage. It does not cover every base, but well, enough is covered :crossed_fingers:

wzrdtales:
> when the IP is assigned to one of the host devices we manage. It does not cover every base, but well, enough is covered 🤞

we have it assigned to lo for that one ip, but we also have instances were the IP is not assigned anywhere at all and only the nodeport makes the IP actually do something through the externalIP listing.
This just came to my mind, need to test once how this behaves in the case the IP is not assigned at all on the ingress host, which would be of course the case most of the time since you don't want anything to bind on these ips.

wzrdtales:
ok i just tested it. I can confirm, this wont work on ips not assigne at all.

PR #10668, opened 2025-07-11
[BPF] batch map userspace ops

tomastigera:
It is much faster to read many values/keys in a single syscall than to use a syscall to first get next key and then to get the value.

Also adds batch userspace update

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/maps/syscall.h
+++ b/felix/bpf/maps/syscall.h
@@ -18,6 +18,9 @@
 #include <errno.h>
 #include <unistd.h>
 #include <sys/syscall.h>
+#include <bpf.h>
+
+#include "libbpf.h"
 
 union bpf_attr *bpf_maps_attr_alloc() {
    union bpf_attr *attr = malloc(sizeof(union bpf_attr));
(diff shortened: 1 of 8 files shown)


felix/bpf/maps/syscall.h
```diff
@@ -130,3 +133,16 @@ int bpf_maps_map_load_multi(__u32 map_fd,
    }
    return count;
 }
+
+static void set_errno(int ret) {
+	errno = ret >= 0 ? ret : -ret;
+}
+
+void bpf_map_batch_lookup(int fd, void *in_batch, void *out_batch, void *keys,
```
sridhartigera, line 141: I feel its better to move this to libbpf_api.h, so that all batch APIs are in the same file.

felix/bpf/ut/bpf_ct_cleanup_test.go
```diff
@@ -28,6 +28,8 @@ import (
 )
 
 func TestBPFProgCleaner(t *testing.T) {
+	RegisterTestingT(t)
```
sridhartigera: There is a call to `RegisterTestingT(t)` below in `setUpConntrackScanTest` API

sridhartigera (approved):
Just 2 minor comments. Otherwise LGTM

felix/bpf/maps/syscall.h
tomastigera: iirc the reason for placing it here was some cgo conflicts when building

PR #10672, opened 2025-07-14
AutoHEPs - include all Node ips and interfaceName selector

MichalFupso:
## Description
Updates to Calico Node startup to record all interfaces that are attached to Node, containing their name and addresses. HostEndpoint controller is able to use this information from node.Spec and create HostEndpoints containing all node IPs or create HostEndpoint for with interfaceName based on updated HostEndpointTemplate

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Updates the HostEndpoint controller to be aware of all Node interfaces and IPs. Update the HostEndpoint template with interfaceSelector to be able to create automatic HostEndpoints containing interfaceName. Existing HostEndpoints created by the HostEndpoint controller might be updated after the upgrade to contain interfaces newly detected by Calico Node. 
```

## Reminder for the reviewer

Other optional labels:


lucastigera:
From the design doc, IIUC the `Release note` should mention possible policy impact due to AutoHostEndpoint updating expectedIPs field on Upgrades?

kube-controllers/pkg/controllers/node/auto_hep_fv_test.go
```diff
@@ -883,6 +931,128 @@ var _ = Describe("Auto Hostendpoint FV tests", func() {
...
+		expectedTemplateIPs := []string{"5.5.5.5"}
+		expectedTemplateHepLabels := map[string]string{
+			"projectcalico.org/created-by": "calico-kube-controllers",
+			"template-label":               "template-value",
+		}
+		expectedTemplateInterface := "eth0"
+		Eventually(func() error {
+			return testutils.ExpectHostendpoint(c, expectedTemplateHepName, expectedTemplateHepLabels, expectedTemplateIPs, autoHepProfiles, expectedTemplateInterface)
+		}, time.Second*15, 500*time.Millisecond).Should(BeNil())
+
```
lucastigera, line 972: Is it worth adding a check to verify if `eth1` is not present? Something like:

```
		// Expect a hostendpoint to NOT be created for eth1.
		expectedTemplateHepName := cn.Name + "-template-eth1-auto-hep"
		expectedTemplateIPs := []string{"6.6.6.6"}
		expectedTemplateHepLabels := map[string]string{
			"projectcalico.org/created-by": "calico-kube-controllers",
			"template-label":               "template-value",
		}
		expectedTemplateInterface := "eth1"
		Consistently(func() error {
			return testutils.ExpectHostendpoint(c, expectedTemplateHepName, expectedTemplateHepLabels, expectedTemplateIPs, autoHepProfiles, expectedTemplateInterface)
		}, time.Second*15, 500*time.Millisecond).ShouldNot(BeNil())
```

MichalFupso: Good idea, updated the test to check for HostEndpoint count instead to not have to create a dummy autohep

api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
```diff
@@ -132,6 +132,9 @@ type Template struct {
 	// InterfaceCIDRs contains a list of CIRDs used for matching nodeIPs to the AutoHostEndpoint
 	InterfaceCIDRs []string `json:"interfaceCIDRs,omitempty" validate:"cidrs"`
 
+	// InterfaceSelector contains regex string to match Node interfaceName to HostEndpoint
```
caseydavenport:
```suggestion
	// InterfaceSelector contains a regex string to match Node interface names. If specified, a HostEndpoint will be created for each matching interface on each selected node.
	// At least one of InterfaceCIDRs and InterfaceSelector must be specified.
```

libcalico-go/lib/apis/v3/node.go
```diff
@@ -67,6 +67,9 @@ type NodeSpec struct {
 
 	// Addresses list address that a client can reach the node at.
 	Addresses []NodeAddress `json:"addresses,omitempty" validate:"omitempty"`
+
+	// Interfaces list of interfaces on the node
```
caseydavenport:
```suggestion
	// Interfaces is a list of interfaces on the node.
```

kube-controllers/pkg/controllers/node/hostendpoints.go
```diff
@@ -462,8 +462,26 @@ func (c *autoHostEndpointController) validateName(name string) (string, error) {
 	return name, nil
 }
 
-// getExpectedIPsMatchingTemplate finds the matching node IPs to the CIDRs specified in the template
-func (c *autoHostEndpointController) getExpectedIPsMatchingTemplate(node *libapi.Node, template config.AutoHostEndpointTemplate) []string {
+// mergeExpectedIPsWithInterfaceIPs merges the provided expected IPs with the IP addresses
+// found on the specified interface of the given node.
+func (c *autoHostEndpointController) mergeExpectedIPsWithInterfaceIPs(node *libapi.Node, interfaceName string, expectedIPs []string) []string {
+	for _, iface := range node.Spec.Interfaces {
```
caseydavenport: AFAICT, the only caller of this function already has the correct `Interface` in hand - any reason it can't just pass that here instead?

e.g.,


```go
func mergeExpectedIPsWithInterfaceIPs(expectedIPs []string, ifce api.NodeInterface) []string 
```

?

kube-controllers/pkg/controllers/node/hostendpoints.go
```diff
@@ -415,31 +418,28 @@ func isAutoHostEndpoint(h *api.HostEndpoint) bool {
...
-// createAutoHostEndpoint creates the supplied HostEndpoint
-func (c *autoHostEndpointController) createAutoHostEndpoint(hostEndpoint *api.HostEndpoint) error {
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-	defer cancel()
-	hostEndpoint, err := c.client.HostEndpoints().Create(ctx, hostEndpoint, options.SetOptions{})
-	if err != nil {
-		logrus.Warnf("could not create hostendpoint for node: %v", err)
-		return err
+// generateAutoHostEndpointName generated the name based on the supplied parameters, creating unique name for each HostEndpoint
+func (c *autoHostEndpointController) generateAutoHostEndpointName(nodeName string, templateName *string, interfaceName *string) (string, error) {
```
caseydavenport: Do we need these to be pointers to strings? Seems like they could be strings and use `!= ""` to check?

That just prevents e.g., accidental or implied side-effects within this function. 

At least to me, making them pointers doesn't really imply the optional-ness that I believe you're going for here. I think it's more clear and less error prone to either use an `Options` pattern or just put a comment explaining which options are required.

caseydavenport: Side note - does this function need to be bound to the controller? AFAICT it could just be an unbound utility function? Not really a problem, but gave me pause when reading it.

api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
```diff
@@ -132,6 +132,9 @@ type Template struct {
 	// InterfaceCIDRs contains a list of CIRDs used for matching nodeIPs to the AutoHostEndpoint
```
caseydavenport:
```suggestion
	// InterfaceCIDRs contains a list of CIRDs used for matching nodeIPs to the AutoHostEndpoint.
	// If specified, only addresses within these CIDRs will be included in the expected IPs. 
	// At least one of InterfaceCIDRs and InterfaceSelector must be specified.  
```

kube-controllers/pkg/controllers/node/hostendpoints.go
MichalFupso: I think your suggestion with `!=""` will work well. I was trying to find a compromise between not having 3 function doing almost the same and having an easy way to determine what the resulting output will be.

api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
```diff
@@ -129,9 +129,15 @@ type Template struct {
 	// +kubebuilder:validation:MaxLength=253
 	GenerateName string `json:"generateName,omitempty" validate:"omitempty,name"`
 
-	// InterfaceCIDRs contains a list of CIRDs used for matching nodeIPs to the AutoHostEndpoint
+	// InterfaceCIDRs contains a list of CIRDs used for matching nodeIPs to the AutoHostEndpoint.
```
caseydavenport:
```suggestion
	// InterfaceCIDRs contains a list of CIDRs used for matching nodeIPs to the AutoHostEndpoint.
```

kube-controllers/pkg/controllers/node/hostendpoints.go
```diff
@@ -415,32 +418,29 @@ func isAutoHostEndpoint(h *api.HostEndpoint) bool {
...
-// generateTemplateAutoHostEndpointName returns the name for HostEndpoint created with a template
-func (c *autoHostEndpointController) generateTemplateAutoHostEndpointName(nodeName string, templateName string) (string, error) {
-	return c.validateName(fmt.Sprintf("%s-%s-%s", nodeName, templateName, hostEndpointNameSuffix))
+	return "", fmt.Errorf("insufficient HostEndpoint identifiers supplied")
 }
 
-func (c *autoHostEndpointController) validateName(name string) (string, error) {
+// validateName ensures the given name does not exceed the DNS1123 (253 characters) subdomain length limit.
+// If the name is too long, it generates a hash-based name that fits the limit.
+func validateName(name string) (string, error) {
```
caseydavenport:
```suggestion
func maybeHashName(name string) (string, error) {
```

Very much a nitpick, but `validateName` implies that this function checks the validity of an input string, whereas this is really "adjusting the input name so that it is valid"

node/pkg/lifecycle/startup/startup.go
```diff
@@ -573,6 +609,11 @@ func configureIPsAndSubnets(node *libapi.Node, k8sNode *v1.Node, getInterfaces f
 		return true, nil
 	}
 
+	if !reflect.DeepEqual(node.Spec.Interfaces, oldInterfaces) {
```
caseydavenport, line 618: How confident are you this list will remain sorted across multiple executions of calico/node? Might be worth explicitly sorting the returned interfaces so that we don't return `true` here arbitrarily.

caseydavenport (approved):
A few minor comments but otherwise this LGTM

PR #10694, opened 2025-07-17
BPF: Fix BW QoS controls

coutinhop:
## Description

In order for BW QoS controls to work properly on BPF mode, there should not be any redirection that bypasses the caliXXXX intf that has the tc qdisc attached.

Set a global flag to skip redirection if egress BW QoS is configured (CALI_GLOBALS_SKIP_EGRESS_REDIRECT), and set route and state level flags (CALI_RT_SKIP_INGRESS_REDIRECT and CALI_ST_SKIP_REDIR_PEER) to skip redirection when ingress BW QoS is configured (as the traffic originates in other BPF programs in the ingress case). Ingress redirection is also skipped in the case of VM workloads (e.g. kubevirt).

As such, a bit of the code that was VM workload-specific was generalized in order to cover both VM workloads and BW QoS (and possibly other use cases if there are any in the future). Notably, the WorkloadType field was removed from the WorkloadEndpoint protobuf, in favor of a WorkloadBpfSkipRedir (even though, currently only the ingress value is used).

Rename and extend the VM workload UT to cover skip ingress redirection, both from HEPs and from WEPs.

Extend the QoS FVs to work in BPF mode, and to cover different encapsulations (VXLAN, IPIP, no-encap). Add some skips to trim down the number of test cases (e.g. encapsulations on iptables/nftables mode, etcd datastore on BPF mode).

## Related issues/PRs

follow-up of https://github.com/projectcalico/calico/pull/10645

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -2327,6 +2327,9 @@ func (m *bpfEndpointManager) doApplyPolicy(ifaceName string) (bpfInterfaceState,
 
 	ap := m.calculateTCAttachPoint(ifaceName)
 	ap.IfIndex = ifindex
+	if wep != nil && wep.QosControls != nil && wep.QosControls.EgressBandwidth > 0 {
+		ap.HasEgressQoSBW = true
```
tomastigera: should it be just names skipEgressRedirect?

felix/bpf/ut/whitelist_test.go
```diff
@@ -134,7 +134,26 @@ func TestSkipRedirectPeerHostToVMWorkload(t *testing.T) {
...
 		// Approved by HEP
 		Expect(ctr.Data().A2B.Approved).To(BeTrue())
-		// NOt approved by WEP yet
+		// Not approved by WEP yet
+		Expect(ctr.Data().B2A.Approved).NotTo(BeTrue())
+		Expect(ctr.Flags() & v3.FlagNoRedirPeer).To(Equal(v3.FlagNoRedirPeer))
+	})
+
+	skbMark = 0
+	runBpfTest(t, "calico_from_workload_ep", nil, func(bpfrun bpfProgRunFn) {
```
tomastigera: :+1:

felix/calc/event_sequencer.go
```diff
@@ -428,9 +428,10 @@ func ModelWorkloadEndpointToProto(ep *model.WorkloadEndpoint, peerData *Endpoint
 		}
 	}
 
-	epType := proto.WorkloadType_REGULAR
-	if isVMWorkload(ep.Labels) {
-		epType = proto.WorkloadType_VM
+	// BPF ingress redirect should be skipped for VM workloads and workloads that have ingress BW QoS configured
+	skipRedir := &proto.WorkloadBpfSkipRedir{}
```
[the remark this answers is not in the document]
tomastigera: perhaps a good point, if we already have the option, fill it in and if anyone uses it in the future, it will avoid confusion that it is not set

felix/dataplane/linux/bpf_ep_mgr.go
coutinhop: good point, will do!

felix/bpf/ut/whitelist_test.go
sridhartigera: I think it will be better to reset the ct map here so that we can be sure ct entries are created with the skip_redirect flag

sridhartigera (approved):
LGTM

coutinhop:
All-green CI run: https://tigera.semaphoreci.com/workflows/8febd338-afce-4026-b341-9761a5b06847?pipeline_id=87b13821-51b9-4667-a866-6972d286da52

PR #10715, opened 2025-07-21
Bump to k8s 1.33.3 release train

MichalFupso:
## Description
Updates Calico k8s dependency to 1.33. Updates Calico to work with endpoinslices instead of deprecated endpoints. Removes deprecated hintsannotations.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico is now built with Go 1.24.6 against Kubernetes v1.33
```

## Reminder for the reviewer

Other optional labels:


hjiawei:
@marvin-tigera When updating Kuberentes dependencies, please consider including a release note. You can refer to our [release notes](https://docs.tigera.io/calico/latest/release-notes/#enhancements) for examples under the "Enhancements" section.

confd/pkg/backends/calico/routes.go
```diff
@@ -146,9 +147,19 @@ func (rg *routeGenerator) TriggerResync() {
...
-func (rg *routeGenerator) getServiceForEndpoints(ep *v1.Endpoints) (*v1.Service, string) {
+func (rg *routeGenerator) getServiceForEndpoints(ep *discoveryv1.EndpointSlice) (*v1.Service, string) {
 	// get key
-	key, err := cache.MetaNamespaceKeyFunc(ep)
+	var key string
+	var err error
+
+	svcName, ok := ep.Labels[discoveryv1.LabelServiceName]
+	if ok {
+		epCopy := ep.DeepCopy()
```
caseydavenport: I think we can simplify this by using the `cache.ObjectName` structure directly to construct the key?

But I'm not entirely sure what scenarios we expect the label to be present vs. not - is there one? Or is the else statement here just defensive?

confd/pkg/backends/calico/routes.go
```diff
@@ -166,43 +177,47 @@ func (rg *routeGenerator) getServiceForEndpoints(ep *v1.Endpoints) (*v1.Service,
...
 
 // getEndpointsForService retrieves the corresponding ep for the given svc
-func (rg *routeGenerator) getEndpointsForService(svc *v1.Service) (*v1.Endpoints, string) {
-	// get key
+func (rg *routeGenerator) getEndpointsForService(svc *v1.Service) ([]*discoveryv1.EndpointSlice, string) {
+	var eps []*discoveryv1.EndpointSlice
+	for _, obj := range rg.epIndexer.List() {
+		ep, ok := obj.(*discoveryv1.EndpointSlice)
+		if !ok {
+			continue
```
caseydavenport, line 185: Feels like this is at least worth a warning log?

confd/pkg/backends/calico/routes.go
```diff
@@ -166,43 +177,47 @@ func (rg *routeGenerator) getServiceForEndpoints(ep *v1.Endpoints) (*v1.Service,
 }
 
 // getEndpointsForService retrieves the corresponding ep for the given svc
-func (rg *routeGenerator) getEndpointsForService(svc *v1.Service) (*v1.Endpoints, string) {
-	// get key
+func (rg *routeGenerator) getEndpointsForService(svc *v1.Service) ([]*discoveryv1.EndpointSlice, string) {
+	var eps []*discoveryv1.EndpointSlice
+	for _, obj := range rg.epIndexer.List() {
```
caseydavenport, line 181: This is a complexity upgrade - going from O(1) to O(#eps) is probably OK, but the performance of this code now depends on roughly the number of services in the cluster.

We could keep a mapping of service key -> endpointslice keys (and vice-versa) if we wanted to address this. Probably not needed for this PR but might be worth looking into as a follow-up, if it's not too complex.

confd/tests/mock_data/calicoctl/mesh/static-routes/kubectl-input.yaml
```diff
@@ -123,18 +132,21 @@ spec:
       targetPort: 80
 
 ---
-kind: Endpoints
-apiVersion: v1
+kind: EndpointSlice
```
caseydavenport, line 154: I think it's probably worth augmenting one of these tests to deploy multiple EndpointSlices, to make sure our logic handles that properly.

typha/pkg/discovery/discovery.go
```diff
@@ -214,40 +215,55 @@ func (d *Discoverer) discoverTyphaAddrs() ([]Typha, error) {
 
 	// If we get here, we need to look up the Typha service endpoints using the k8s API.
 	logrus.Info("(Re)discovering Typha endpoints using the Kubernetes API...")
-	epClient := d.k8sClient.CoreV1().Endpoints(d.k8sNamespace)
-	eps, err := epClient.Get(context.Background(), d.k8sServiceName, v1.GetOptions{})
+	epClient := d.k8sClient.DiscoveryV1().EndpointSlices(d.k8sNamespace)
+	epsList, err := epClient.List(context.Background(), v1.ListOptions{})
```
caseydavenport: If the service name is always present in a label on the EPSlice, I think it would be good to use a LabelSelector for the query here instead of querying for all slices and filtering them out afterwards.

confd/pkg/backends/calico/routes.go
MichalFupso: We expect the label to be present on the endpointslice, that is how they are associated with the service. I think this should actually be updated to fail if we can't read the label.

caseydavenport: Yep! But part of my point was that it's a bit confusing to generate the service key by modifying the endpointslice. I think it's strictly valid, but is a bit confusing.

```go
// Build the corresponding service key to look it up in the indexer.
svc := v1.Service{ObjectMeta: metav1.ObjectMeta{Name: svcName, Namespace: ep.Namespace}}
key, err := cache.MetaNamespaceKeyFunc(svc)
```

^ I think we could do something like this and it would be fewer lines of code and also more obvious what is happening?

caseydavenport (approved):
One minor comment but LGTM

Issue #10711, opened 2025-07-21
allow announcing smaller k8s service ip routes


juliantaylor:
We are currently investigating announcing kubernetes service ips via calico-node (replacing metallb).

An issue we are encountering is that calico announces the full service range (e.g. bgpconfigurations.spec.serviceLoadBalancerIPs) for externalTrafficPolicy: Cluster services to the peers.

This conflicts with the anycast failover mechanism between datacenters we are using as routes of failed services are not withdrawn.

Would it be possible to add a configuration to calico that changes the behavior to only announce /32 routes based on existing service.status.LoadBalancer.Ingress.ip within the configured bgpconfiguration.serviceLoadBalancerIPs (and externalip equivalents) if any ready endpoints for externalTrafficPolicy Cluster service exists.

For example:

```
kind: BGPConfiguration
spec:
  singleIPAnnouncement: true  
  serviceLoadBalancerIPs:
  - cidr: 192.168.1.0/24
```

Would then instead of only announcing the single /24 route, would instead only announce multiple /32 routes per ready service.

The implementation appears to be rather simple as the core logic basically already exists for externalTrafficPolicy: Local services:
https://github.com/projectcalico/calico/blob/master/confd/pkg/backends/calico/routes.go#L485

caseydavenport:
I think this is a reasonable enhancement request, and probably fairly simple to implement.

caseydavenport:
Rather than a boolean, I'd probably use an enum type field:

```yaml
spec:
  serviceLoadBalancerAggregation: Enabled | Disabled
```

And probably want an equivalent field for other service types.

PR #10719, opened 2025-07-23
Allow announcing smaller k8s service ip routes

PersistentJZH:
## Description

allow announcing smaller k8s service ip routes

## Related issues/PRs

https://github.com/projectcalico/calico/issues/10711

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


PersistentJZH:
@caseydavenport Could you please take a look at this PR when you have time? Thanks!

caseydavenport:
@PersistentJZH will do, thanks for raising it! I've been out sick so just catching up.

api/pkg/apis/projectcalico/v3/bgpconfig.go
```diff
@@ -53,6 +53,17 @@ type BGPConfiguration struct {
...
 }
 
+// ServiceLoadBalancerAggregation defines how LoadBalancer service IPs should be aggregated for BGP advertisement.
+// +kubebuilder:validation:Enum=Enabled;Disabled
+type ServiceLoadBalancerAggregation string
+
+const (
+	// ServiceLoadBalancerAggregationEnabled means individual /32 routes will be advertised for each LoadBalancer service
+	ServiceLoadBalancerAggregationEnabled ServiceLoadBalancerAggregation = "Enabled"
+	// ServiceLoadBalancerAggregationDisabled means the full CIDR range will be advertised (default behavior)
```
caseydavenport: We shouldn't be changing the default behavior as part of this PR - that would be a breaking change on upgrade for any users currently relying on the aggregation behavior in Calico.

api/pkg/apis/projectcalico/v3/bgpconfig.go
```diff
@@ -53,6 +53,17 @@ type BGPConfiguration struct {
 	Spec BGPConfigurationSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
 }
 
+// ServiceLoadBalancerAggregation defines how LoadBalancer service IPs should be aggregated for BGP advertisement.
+// +kubebuilder:validation:Enum=Enabled;Disabled
+type ServiceLoadBalancerAggregation string
+
+const (
+	// ServiceLoadBalancerAggregationEnabled means individual /32 routes will be advertised for each LoadBalancer service
```
caseydavenport: I think the naming on these fields is backwards - when Aggregation is Enabled, that means Calico should be advertising the fully CIDR.

When Aggregation is Disabled, Calico shouldn't aggregate and instead will advertise the individual IP addresses.

confd/pkg/backends/calico/client.go
```diff
@@ -1554,7 +1570,17 @@ func (c *client) onLoadBalancerIPsUpdate(lbIPs []string) {
...
 	// which should instead be advertised from only a subset of nodes.
 	// So, we handle advertisement of any single-addresses found in the config on a per-service basis from within the routeGenerator.
-	globalLbIPs := filterNonSingleIPsFromCIDRs(lbIPs)
+
+	var globalLbIPs []string
+	if c.ServiceLoadBalancerAggregationEnabled() {
+		// When service load balancer aggregation is enabled, we don't advertise any CIDR ranges globally.
+		// Instead, individual /32 routes will be advertised per service based on their LoadBalancer.Ingress.IP.
+		globalLbIPs = []string{}
+		log.Info("Service load balancer aggregation enabled - not advertising LoadBalancer CIDR ranges globally")
```
caseydavenport: This can be a Debug log, otherwise it will be quite spammy.

confd/pkg/backends/calico/routes.go
```diff
@@ -465,12 +465,32 @@ func (rg *routeGenerator) advertiseThisService(svc *v1.Service, ep *v1.Endpoints
...
+			}
+		} else {
+			// Original logic for when single IP announcement is disabled
+			if svc.Spec.Type == v1.ServiceTypeLoadBalancer && rg.isSingleLoadBalancerIP(svc.Spec.LoadBalancerIP) {
+				logc.Debug("Advertising load balancer of type cluster because of single IP definition")
+				return true
+			}
 		}
 
+		// Handle ExternalIPs - this logic is unchanged and works for both aggregation modes
```
caseydavenport: No need to say that the logic is unchanged in a code comment, as that context is lost once the PR is merged :)

confd/pkg/backends/calico/routes.go
```diff
@@ -465,12 +465,32 @@ func (rg *routeGenerator) advertiseThisService(svc *v1.Service, ep *v1.Endpoints
 	// There are 2 cases inside this type:
 	// - LoadBalancer with a single IP.
 	// - Any one of externalIPs in service of type LoadBalancer or NodePort with a single IP.
```
caseydavenport:
```suggestion
	// There are 3 cases where we need to advertise this service explicitly, even though it is Cluster type:
	// - LoadBalancer belongs to a configured CIDR that encompasses a singular IP.
	// - LoadBalancer CIDR aggregation is Disabled.
	// - Any one of externalIPs in service of type LoadBalancer or NodePort with a single IP.
```

Think this comment could use some fixing up to be more clear (and account for the new logic).

confd/pkg/backends/calico/routes.go
```diff
@@ -465,12 +465,32 @@ func (rg *routeGenerator) advertiseThisService(svc *v1.Service, ep *v1.Endpoints
...
-	if svc.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeCluster {
-		if svc.Spec.Type == v1.ServiceTypeLoadBalancer && rg.isSingleLoadBalancerIP(svc.Spec.LoadBalancerIP) {
-			logc.Debug("Advertising load balancer of type cluster because of single IP definition")
-			return true
+	// Note: ExternalTrafficPolicy defaults to Cluster if not specified
+	if svc.Spec.ExternalTrafficPolicy == "" || svc.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeCluster {
```
caseydavenport: Do we really need to account for `ExternalTrafficPolicy: ""`? 

I believe that this field is populated by the Kubernetes API and so we will never actually receive an empty string here?

PersistentJZH: yes, I looked at the [implementation](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/v1/defaults.go#L137-L139) of API Server, the API server will sets a default value when the field is not specified in the service definition. so this field will never be empty, I will remove this judgment.

api/pkg/apis/projectcalico/v3/bgpconfig.go
PersistentJZH: thanks @caseydavenport, I will optimize this logic.

PersistentJZH:
Thanks @caseydavenport for the code review. I have resolved all the comments above. please review it again, thank you!

confd/pkg/backends/calico/routes.go
```diff
@@ -461,14 +461,31 @@ func (rg *routeGenerator) advertiseThisService(svc *v1.Service, ep *v1.Endpoints
...
 	// - Any one of externalIPs in service of type LoadBalancer or NodePort with a single IP.
 	if svc.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeCluster {
-		if svc.Spec.Type == v1.ServiceTypeLoadBalancer && rg.isSingleLoadBalancerIP(svc.Spec.LoadBalancerIP) {
-			logc.Debug("Advertising load balancer of type cluster because of single IP definition")
-			return true
+		// Check if service load balancer aggregation is disabled
+		if !rg.client.ShouldAggregateLoadBalancerServices() {
+			// When aggregation is disabled, we advertise LoadBalancer services
+			// based on their LoadBalancer.Ingress.IP if they have ready endpoints
+			if svc.Spec.Type == v1.ServiceTypeLoadBalancer && svc.Status.LoadBalancer.Ingress != nil {
```
juliantaylor: this appears to duplicate the announcement logic while this function just determines whether the a service ip should be announced or not

it may be cleaner to leave this part of the code as is and instead allow externalTrafficPolicy Cluster services to reach the logic used for Local services and adapt that to instead only check the aggregation flag or node local endpoints:

juliantaylor: this may be sufficient:

```patch
--- a/confd/pkg/backends/calico/routes.go
+++ b/confd/pkg/backends/calico/routes.go
@@ -482,7 +482,7 @@ func (rg *routeGenerator) advertiseThisService(svc *v1.Service, ep *v1.Endpoints
        }
 
        // we only need to advertise local services, since we advertise the entire cluster IP range.
-       if svc.Spec.ExternalTrafficPolicy != v1.ServiceExternalTrafficPolicyTypeLocal {
+       if svc.Spec.ExternalTrafficPolicy != v1.ServiceExternalTrafficPolicyTypeLocal && rg.client.ShouldAggregateLoadBalancerServices {
                logc.Debugf("Skipping service with non-local external traffic policy '%s'", svc.Spec.ExternalTrafficPolicy)
                return false
        }
@@ -495,7 +495,7 @@ func (rg *routeGenerator) advertiseThisService(svc *v1.Service, ep *v1.Endpoints
        for _, subset := range ep.Subsets {
                // not interested in subset.NotReadyAddresses
                for _, address := range subset.Addresses {
-                       if address.NodeName != nil && *address.NodeName == rg.nodeName && isIPv6(address.IP) == svcIsIPv6 {
+                       if ((svc.Spec.ExternalTrafficPolicy != v1.ServiceExternalTrafficPolicyTypeLocal) || (address.NodeName != nil && *address.NodeName == rg.nodeName)) && isIPv6(address.IP) == svcIsIPv6 {
                                logc.Debugf("Advertising local service")
                                return true
                        }

File: confd/pkg/backends/calico/client.go [hunk 8]
```diff
@@ -1228,6 +1234,16 @@ func (c *client) updateBGPConfigCache(resName string, v3res *apiv3.BGPConfigurat
 		c.getNodeMeshPasswordKVPair(v3res, model.GlobalBGPConfigKey{})
 		c.getIgnoredInterfacesKVPair(v3res, model.GlobalBGPConfigKey{})
 
+		// Update service load balancer aggregation setting
+		c.cacheLock.Lock()
```
juliantaylor: this is deadlocking for me, not familiar enough with golang to tell you more unfortunately

PersistentJZH: thanks @juliantaylor, this is a good suggestion. The [onUpdates function](https://github.com/projectcalico/calico/blob/master/confd/pkg/backends/calico/client.go#L987C1-L988C28) has already acquired the lock, so there is no need to acquire the lock again here, otherwise it will cause a deadlock.

confd/pkg/backends/calico/client.go
```diff
@@ -1582,6 +1608,13 @@ func (c *client) GetLoadBalancerIPs() []*net.IPNet {
 	return c.loadBalancerIPNets
 }
 
+// ServiceLoadBalancerAggregationEnabled returns whether service load balancer aggregation is enabled
+func (c *client) ShouldAggregateLoadBalancerServices() bool {
+	c.cacheLock.Lock()
```
juliantaylor, line 1613: this also deadlocks

PersistentJZH: fixed

confd/pkg/backends/calico/routes.go
PersistentJZH: Hi @caseydavenport, what's your opinion for this suggestion.

juliantaylor:
Is there any connection between the process that configures the bgp and the process that configures the local nodes network?
If a new endpoint becomes ready does calico wait for the endpoint to be completely configured on the local node before announcing the bgp route?

Using metallb for loadbalancerip annoucements we ran into the problem that a pod getting ready on a remote node does not imply that the network of the node announcing has also been programmed for traffic to flow to the pod.

Due to lack of a good feedback mechanism worked around this by putting a delay after the ready endpoint is seen and the first service ip annoucement. This was rather important because we also use kube-proxy in iptables mode which can take up to a minute to configure the network in larger clusters.

Though if calico is very fast in node programming its probably not a very big concern, how long does calico in ebpf mode typically take from endpoint readiness to node network configuration completion on larger clusters, say 30000 total pod and services with 1000 pods?

confd/pkg/backends/calico/routes.go
caseydavenport: Yeah, I like this suggestion. I think it keeps the code a bit cleaner. 

I think we'd want to keep the `if` statements separate for the second diff, though. For readability / logging purposes.

PersistentJZH: @caseydavenport @juliantaylor thanks for suggestion, I have optimize code, please help review again.

api/pkg/apis/projectcalico/v3/bgpconfig.go
```diff
@@ -76,6 +87,12 @@ type BGPConfigurationSpec struct {
 	// If specified, Calico will advertise these blocks, as well as any cluster IPs within them.
 	ServiceClusterIPs []ServiceClusterIPBlock `json:"serviceClusterIPs,omitempty" validate:"omitempty,dive" confignamev1:"svc_cluster_ips"`
 
+	// ServiceLoadBalancerAggregation controls how LoadBalancer service IPs are advertised.
+	// When set to "Disabled", individual /32 routes are advertised for each service instead of
+	// the full CIDR range. This is useful for anycast failover mechanisms where failed service
+	// routes need to be withdrawn. [Default: Enabled]
+	ServiceLoadBalancerAggregation ServiceLoadBalancerAggregation `json:"serviceLoadBalancerAggregation,omitempty" validate:"omitempty" confignamev1:"svc_loadbalancer_aggregation"`
```
caseydavenport:
```suggestion
        // +kubebuilder:default=Enabled
	ServiceLoadBalancerAggregation *ServiceLoadBalancerAggregation `json:"serviceLoadBalancerAggregation,omitempty" validate:"omitempty" confignamev1:"svc_loadbalancer_aggregation"`
```

A couple small things - this should be a pointer (it's an optional field, and optional fields in k8s APIs should be pointers) and we should add a kube-builder tag to automatically default when unset.

confd/pkg/backends/calico/routes.go
```diff
@@ -249,6 +249,15 @@ func (rg *routeGenerator) getAllRoutesForService(svc *v1.Service) []string {
 	}
 	svcID := fmt.Sprintf("%s/%s", svc.Namespace, svc.Name)
 
+	// For Cluster services, check aggregation setting
```
caseydavenport: I think this isn't quite right - the new config field is only for controlling LoadBalancer aggregation, not ExternalIPs.

So if `ExternalTrafficPolicy: Cluster` and `!ShouldAggregateLoadBalancerServices`, we should still aggregate `ExternalIPs` (and not advertise them explicitly) right?

caseydavenport:
> Is there any connection between the process that configures the bgp and the process that configures the local nodes network?
If a new endpoint becomes ready does calico wait for the endpoint to be completely configured on the local node before announcing the bgp route?

The endpoint won't show up in the `Addresses` of the Endpoints object until it is reporting ready by its probe, at which point Calico will announce the route (If it hasn't announced it already). However, there are other network factors that do not factor into endpoint readiness such as network policy, which means there is still a small race condition at play here. Calico generally programs this O(ms), so for the most part I wouldn't expect too much of an issue here. 

> Using metallb for loadbalancerip annoucements we ran into the problem that a pod getting ready on a remote node does not imply that the network of the node announcing has also been programmed for traffic to flow to the pod.

Was this using Calico for pod networking as well?

> how long does calico in ebpf mode typically take from endpoint readiness to node network configuration completion on larger clusters

Endpoint readiness is partially orthogonal to Calico's dataplane programming - a few things need to be set up (e.g., the pod network interface and network namespace) but then the amount of time it takes for a Pod to go ready depends on the readiness probe. Calico will continue programming network policy, etc., in parallel with that - but like I mentioned above it should be on the order of milliseconds in most steady-state scenarios.

@juliantaylor might be worth taking this conversation to a GH issue if you're seeing problems? I don't think it's necessarily related to this PR?

api/pkg/apis/projectcalico/v3/bgpconfig.go
PersistentJZH: fixed

confd/pkg/backends/calico/routes.go
PersistentJZH: optimized, @caseydavenport pls  review again. thank you

juliantaylor:
yes it is not directly related to this PR, more loadbalancer ip announcements in general, I'll report if I anything can be observed in another issues while testing this.

juliantaylor:
while testing I ran into an issue with the [ServiceLoopPrevention](https://docs.tigera.io/calico-enterprise/latest/reference/resources/felixconfig)

when announcing to an anycast range from only one datacenter traffic to the anycast ip in another cluster with the same  serviceLoadBalancerIPs which does not announce the ip is dropped due to the loop prevention.
Setting `FELIX_SERVICELOOPPREVENTION=disabled` made it work.

Maybe with aggregation disabled the loop prevention should be disabled by default as I don't think it is needed in this case?

caseydavenport:
@PersistentJZH I think you need to run `make generate` in the root of the repository to update manifests, code, etc with the new API field.

PersistentJZH:
> while testing I ran into an issue with the [ServiceLoopPrevention](https://docs.tigera.io/calico-enterprise/latest/reference/resources/felixconfig)
> 
> when announcing to an anycast range from only one datacenter traffic to the anycast ip in another cluster with the same serviceLoadBalancerIPs which does not announce the ip is dropped due to the loop prevention. Setting `FELIX_SERVICELOOPPREVENTION=disabled` made it work.
> 
> Maybe with aggregation disabled the loop prevention should be disabled by default as I don't think it is needed in this case?

Yes, but directly disable ServiceLoopPrevention will completely disable loop prevention for all services, including ClusterIP and ExternalIP. I am not sure whether this will bring some side effects. Perhaps we need a more refined control to make the disable of ServiceLoopPrevention only take effect on LoadBalancerIP?

PersistentJZH:
@caseydavenport  pls help to trigger test, thank you

PersistentJZH:
I see the test is failing again, but it doesn't seem to be caused by my changes, and I've seen the same error in other PR, Do you have any idea on this? @caseydavenport  thank you!

e.g.: https://tigera.semaphoreci.com/jobs/e04d8c79-287b-457d-91d8-53b609839552#L3023

caseydavenport:
Yep, that's a known flaky test and not related to your changes :+1:

PR #10721, opened 2025-07-23
BPF: Fix that stale-NAT cleanup didn't wait for kube-proxy sync

### LINKED ISSUE #10670: [bpf] Connections dropped during felix restart
## Expected Behavior

As usually observed in the iptables data plane, I'd expect a typha restart (as in a regular typha rolling redeployment) not having significant negative impact on the ebpf data plane. 

## Current Behavior

Using the ebpf data plane in a bigger cluster (280 nodes, 1900 services, 18k pods) causes hanging traffic after a typha restart. Seemingly, while a calico-node is reconnecting (despite having similar output as with the iptables data plane), the ebpf data plane is reconfigured and connot properly forward traffic until done. 
I'm specifically seeing timeouts in our ingress controllers / gateways (traefik, istio) are not able to forward traffic to the workload pods. 

In my last production test, I observed effects for up to 20m after the typha redeployment. Old typha pods took up to 5m to terminate (which is our termination grace period, but it seems all connections were handed off before termination). Typha metrics show that all connections were up at ~6m after the start of the shutdown - still I see application impact up to 20m after the restart. 

Noteworthy as well: I see significant drop of conntrack table size during the typha reconnects, which I would not expect. Maybe this even is the core of the problem, that connectiontable state is (partially?) dropped? 

1. `kubectl rollout restart -n kube-system deployment calico-typha`

## Your Environment

* Calico version 3.30.1
* Calico dataplane (iptables, windows etc.) ebpf
* Orchestrator version (e.g. kubernetes, mesos, rkt): k8s v1.31.8
* Operating System and version: FlatCar ContainerLinux 4152.2.3

fasaxc:
## Description

Conntrack cleanup could fire before kube-proxy had loaded the services, resulting in _every_ NAT entry looking stale.

## Related issues/PRs

Should fix #10670

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
eBPF: Fix race between loading kubernetes services and conntrack cleanup.  If conntrack cleanup ran before services were loaded, all service entries would look stale and get cleaned up.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/proxy/kube-proxy.go
+++ b/felix/bpf/proxy/kube-proxy.go
@@ -254,7 +254,7 @@ func (kp *KubeProxy) ConntrackFrontendHasBackend(ip net.IP, port uint16, backend
 	// Thanks to holding the lock since ConntrackScanStart, this condition holds for the
 	// whole iteration. So if we started without syncer, we will also finish without it.
 	// And if we had a syncer, we will have the same until the end.
-	if kp.syncer != nil {
+	if kp.syncer != nil && kp.syncer.HasSynced() {
 		return kp.syncer.ConntrackFrontendHasBackend(ip, port, backendIP, backendPort, proto)
 	}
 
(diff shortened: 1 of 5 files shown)


sridhartigera (approved):
LGTM

PR #10726, opened 2025-07-24
Add publishing OpenStack RPMs to Google Artifact Registry

danudey:
Currently we publish RedHat RPMs for Openstack to a Google Cloud VM, but for our use cases publishing to Google Cloud's Artifact Registry is probably sufficient. This would allow us to remove the VM entirely, reducing our maintenance overhead to host these RPMs.

In future:

1. Redirect binaries.projectcalico.org to Artifact Registry
2. Remove the VM once we know things are stable

Unfortunately since PackageCloud was acquired, there are no good "host my packages for me" services that I can find, so GAR is the most feasible option for now.

### DIFF
--- a/release/packaging/utils/create-update-packages.sh
+++ b/release/packaging/utils/create-update-packages.sh
@@ -74,6 +74,11 @@ function require_version {
         PATCH=${BASH_REMATCH[3]}
         PY2SUFFIX=${BASH_REMATCH[4]}
         : ${REPO_NAME:=calico-${MAJOR}.${MINOR}${PY2SUFFIX}}
+        # Set the GCLOUD_REPO_NAME; Google Artifact Registry doesn't allow for
+        # most characters, including periods, to be in registry names, meaning
+        # we can't use e.g. `calico-3.30`. For now, specify it manually. In
+        # future, after more testing, possibly we can just strip the period out
+        # of whatever REPO_NAME is set to.
         : ${GCLOUD_REPO_NAME:=calico-${MAJOR}${MINOR}}
     else
         echo "ERROR: Unhandled VERSION \"${VERSION}\""
@@ -185,13 +190,13 @@ function do_bld_images {
 }
 
 function do_net_cal {
-    # Build networking-calico packages.
+    # Build networking-calico packages, with RPM packages built first to be consistent
+    # with felix (which has a hard requirement on building rpm first at the moment)
     pushd "${rootdir}/networking-calico"
     PKG_NAME=networking-calico \
             NAME=networking-calico \
             DEB_EPOCH=3: \
-            # Build DEB packages first, followed by RPM packages, to maintain consistency with other build steps.
-            "${rootdir}/release/packaging/utils/make-packages.sh" deb rpm
+            "${rootdir}/release/packaging/utils/make-packages.sh" rpm deb
     # Packages are produced in rootDir/ - move them to the output dir.
     find ../ -type f -name 'networking-calico_*-*' -exec mv '{}' "$outputDir" \;
     # Revert the changes made to networking-calico as part of the package build.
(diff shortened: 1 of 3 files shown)


release/packaging/utils/lib.sh
```diff
@@ -146,14 +148,37 @@ function copy_rpms_to_host {
...
+    done
+}
+
+function copy_rpms_to_artifact_registry {
+    reponame=$1
+    rootdir=$(git_repo_root)
+    upload_errors=""
+    shopt -s nullglob
+    echo "Uploading RPMs to Google Artifact Registry"
+    for rpmfile in $(find ${rootdir}/release/packaging/output/dist/rpms-el7 -name "*.rpm" | sort); do
```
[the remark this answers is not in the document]
danudey: This is technically correct but I don't really care.

release/packaging/utils/create-update-packages.sh
```diff
@@ -58,28 +58,32 @@ function require_version {
...
-	MINOR=${BASH_REMATCH[2]}
-	PATCH=${BASH_REMATCH[3]}
-	PY2SUFFIX=${BASH_REMATCH[4]}
-	: ${REPO_NAME:=calico-${MAJOR}.${MINOR}${PY2SUFFIX}}
+        MAJOR=${BASH_REMATCH[1]}
+        MINOR=${BASH_REMATCH[2]}
+        PATCH=${BASH_REMATCH[3]}
+        PY2SUFFIX=${BASH_REMATCH[4]}
+        : ${REPO_NAME:=calico-${MAJOR}.${MINOR}${PY2SUFFIX}}
+        : ${GCLOUD_REPO_NAME:=calico-${MAJOR}${MINOR}}
```
nelljerram, line 82: Should there be a dot between MAJOR and MINOR?

And if the answer is yes, why can't GCLOUD_REPO_NAME be the same as REPO_NAME ?

nelljerram: (In all cases, I mean.  BTW, just to confirm, we don't need to worry about PY2SUFFIX any more.  Could rip that out now, although probably better in a separate PR.)

nelljerram: (IOW there are no remaining cases for anyone using a Python 2 version of our code.)

release/packaging/utils/create-update-packages.sh
```diff
@@ -184,9 +188,10 @@ function do_net_cal {
     # Build networking-calico packages.
     pushd "${rootdir}/networking-calico"
     PKG_NAME=networking-calico \
-	    NAME=networking-calico \
-	    DEB_EPOCH=3: \
-	    "${rootdir}/release/packaging/utils/make-packages.sh" deb rpm
+            NAME=networking-calico \
+            DEB_EPOCH=3: \
+            # Build DEB packages first, followed by RPM packages, to maintain consistency with other build steps.
```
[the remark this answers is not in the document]
nelljerram: Agree.  Probably better to put this comment before the `PKG_NAME=...` line.

release/packaging/utils/create-update-packages.sh
danudey: Unfortunately Google Artifact Registry doesn't let you put dots in repository names, so no, in this specific case we can't have the dot. This grotesquely means having a very slightly different repository name in Artifact Registry compared to what we use elsewhere.

release/packaging/utils/create-update-packages.sh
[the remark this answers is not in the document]
danudey: Agree. Also want to point out that it was Copilot's idea to put the comment there in the first place. Mind your business next time, Copilot!

release/packaging/utils/create-update-packages.sh
nelljerram: Does it allow anything else?  Otherwise, we are technically ambiguous and can confuse the future Calico v33.0 release with the current v3.30.

danudey: Basically no; for Yum repositories specifically (but not e.g. Docker ones), it's restricted to letters, numbers, and hyphens.

I propose we revisit this when we release Calico v31.0 and determine if GAR is still meeting our needs at that point.

nelljerram (approved):

PR #10736, opened 2025-07-28
eBPF: Hybrid conntrack cleaner

sridhartigera:
## Description

This PR implements the BPF conntrack cleaner. Earlier we had a separate BPF conntrack cleaner, which scans the entire conntrack map, runs the conntrack expiry logic. This was disabled when we moved to LRU conntrack map, as scanning the entire map from the kernel will disturb the LRU logic. 

This resulted in just the userspace cleaner and we are back to having races when cleaning up entries. This PR tries to solve this race as mentioned below
1. The liveness and stale NAT scanner scans the conntrack map and if the entry needs to be deleted, it returns `ScanVerdictDelete` along with the `last_seen` timestamp. If the conntrack entry is `NATForward`, the scanner returns the `last_seen` timestamp of the reverse entry.
2. This information is updated in the conntrack cleanup map. The conntrack cleanup map, has the following structure key = `conntrack_key` and value = `<conntrack_reverse_key>:<last_seen_ts><rev_last_seen_ts>`. 
3. After all the entries are scanned, the userspace cleaner runs the bpfCleaner.

### bpfCleaner

The `bpfCleaner` does the following
1. For each element in the cleanup map, we do a lookup of the key into the conntrack map. This lookup_key points to the reverse key stored in the cleanup map's value if there is a valid reverse key. 
2. The conntrack value's `last_seen` is compared with `last_seen` in the cleanupmap value or it is compared with the reverse_last_seen if the entry is of type `NATForward`.
3. If the timestamps match, the conntrack entry is deleted from the conntrack map as well as the cleanup map. In the case of NATForward entry, both the forward and reverse entries are deleted from the conntrack map.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
eBPF: conntrack cleanup now uses a hybrid approach with a scan being done from user space (to avoid disturbing LRU state) and then deletions being done from BPF program. 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/conntrack_cleanup.c
+++ b/felix/bpf-gpl/conntrack_cleanup.c
@@ -27,21 +27,6 @@ struct ct_iter_ctx {
 	__u64 num_cleaned;
 };
 
-static CALI_BPF_INLINE bool ct_key_equal(struct calico_ct_key *key1, struct calico_ct_key *key2) {
-	if (!key1 || !key2) {
-		return key1 == key2;
-	}
-	if ((key1->protocol == key2->protocol) &&
-			(key1->port_a == key2->port_a) &&
-			(key1->port_b == key2->port_b) &&
-			ip_equal(key1->addr_a, key2->addr_a) &&
-			ip_equal(key1->addr_b, key2->addr_b)) {
-		return true;
-	}
-	return false;
-
-}
-
 // process_ccq_entry processes an entry in the "cleanup queue" map. The map
 // is keyed with conntrack key which the userspace cleaner sees as expired.
 // The value has <rev_key>:<last_seen_ts>:<rev_last_seen_ts>
@@ -67,7 +52,7 @@ static long process_ccq_entry(void *map, struct calico_ct_key *key, struct cali_
 		// Check if the fwd key still points to the same reverse key.
 		if (nat_fwd_value) {
 			struct calico_ct_key *nat_rev_key = &nat_fwd_value->nat_rev_key;
-			if (!ct_key_equal(nat_rev_key, rev_key)) {
+			if (__builtin_memcmp(nat_rev_key, rev_key, sizeof(struct calico_ct_key))) {
 		       		goto delete;
 			}
 		}
--- a/felix/bpf-gpl/conntrack_cleanup.c
+++ b/felix/bpf-gpl/conntrack_cleanup.c
@@ -27,9 +27,9 @@ struct ct_iter_ctx {
 	__u64 num_cleaned;
 };
 
-static CALI_BPF_INLINE bool if_ct_key_equal(struct calico_ct_key *key1, struct calico_ct_key *key2) {
+static CALI_BPF_INLINE bool ct_key_equal(struct calico_ct_key *key1, struct calico_ct_key *key2) {
 	if (!key1 || !key2) {
-		return false;
+		return key1 == key2;
 	}
 	if ((key1->protocol == key2->protocol) &&
 			(key1->port_a == key2->port_a) &&
@@ -67,7 +67,7 @@ static long process_ccq_entry(void *map, struct calico_ct_key *key, struct cali_
 		// Check if the fwd key still points to the same reverse key.
 		if (nat_fwd_value) {
 			struct calico_ct_key *nat_rev_key = &nat_fwd_value->nat_rev_key;
-			if (!if_ct_key_equal(nat_rev_key, rev_key)) {
+			if (!ct_key_equal(nat_rev_key, rev_key)) {
 		       		goto delete;
 			}
 		}
--- a/felix/bpf-gpl/conntrack_cleanup.c
+++ b/felix/bpf-gpl/conntrack_cleanup.c
@@ -27,37 +27,38 @@ struct ct_iter_ctx {
 	__u64 num_cleaned;
 };
 
-// process_ccq_entry processes an entry in the "cleanup queue" map.  The map
-// is keyed on the reverse NAT entry's key, with the value storing the forward
-// entry's key (or a zero value if the forward entry is missing).
+// process_ccq_entry processes an entry in the "cleanup queue" map. The map
+// is keyed with conntrack key which the userspace cleaner sees as expired.
+// The value has <rev_key>:<last_seen_ts>:<rev_last_seen_ts>
+// The rev_key is dummy for Normal and reverse entries and is valid for Forward entry.
 static long process_ccq_entry(void *map, struct calico_ct_key *key, struct cali_ccq_value *value, void *ctx)
 {
-	bool isNatForward = false;
+	bool is_nat_forward = false;
 	struct calico_ct_key *lookup_key = key;
 	struct ct_iter_ctx *ictx = ctx;
 
 	// If NAT fwd entry, do a lookup of the rev_key.
 	if (value->rev_key.protocol != 0) {
-		isNatForward = true;
+		is_nat_forward = true;
 		lookup_key = &value->rev_key;
 	}
 
 	const struct calico_ct_value *actual_ct_value = cali_ct_lookup_elem(lookup_key);
 	if (actual_ct_value) {
-		if (isNatForward) {
+		if (is_nat_forward) {
 			// If the reverse key has expired, go to delete.
 			if (actual_ct_value->last_seen == value->rev_last_seen) {
 			       goto delete;
 			}
-			return 0;
+			goto cleanup;
 		} else {
 			// This is normal or NAT reverse entry.
 			if (actual_ct_value->last_seen == value->last_seen) {
 				goto delete;
 			}
-			return 0;
+			goto cleanup;
 		}
-	} else if (isNatForward) {
+	} else if (is_nat_forward) {
 		// Its a NAT forward entry but the reverse entry is not present in the CT table.
 		// Set the lookup_key to the forward entry and do a lookup of the fwd entry.
 		// If the timestamp's match, delete it.
@@ -67,31 +68,34 @@ static long process_ccq_entry(void *map, struct calico_ct_key *key, struct cali_
 		if (actual_ct_value && actual_ct_value->last_seen == value->last_seen) {
 			goto delete;
 		}
-		return 0;
+		goto cleanup;
 	}
 delete:
 	// If the entry is Normal or Reverse, lookup_key points to the entry.
 	// Delete the entry and if successful, increment the counter.
 	// If the entry is forward, we have 2 cases,
 	// Rev_entry present - lookup_key points to the reverse entry and key points to the
-	// forward entry. Delete both the entries togethere.
+	// forward entry. Delete both the entries together.
 	// Rev_entry not present - lookup_key points to the forward entry.
 	// Delete it.
 	if(!cali_ct_delete_elem(lookup_key)) {
 		ictx->num_cleaned++;
-		if (isNatForward && !cali_ct_delete_elem(key)) {
+		if (is_nat_forward && !cali_ct_delete_elem(key)) {
 			ictx->num_cleaned++;
 		}
-		cali_ccq_delete_elem(key);
 	}
+cleanup:
+	cali_ccq_delete_elem(key);
 	return 0;
 }
 
 // conntrack_cleanup is a BPF program that cleans up expired conntrack entries.
-// It does a single pass of the conntrack map, checking each entry for expiry.
-// Normal entries are deleted immediately if they are expired.  NAT entries are
-// queued for deletion in the cali_ccq map (so that forward and reverse entries
-// can be cleaned up together).
+// The expired entries are added to the cleanup_map from the userspace.
+// The cleaner does a single pass of the conntrack cleanup map.
+// It checks the following and deletes the entries.
+// * entry's last_seen is same as the one recorded by the userspace scanner.
+// * Does the NAT forward entry still point to the same reverse entry as seen by
+//   the userspace.
 __attribute__((section("tc"))) int conntrack_cleanup(struct __sk_buff *skb)
 {
 	struct ct_iter_ctx ictx = {};
--- a/felix/bpf-gpl/conntrack_cleanup.c
+++ b/felix/bpf-gpl/conntrack_cleanup.c
@@ -24,229 +24,66 @@ struct ct_iter_ctx {
 	__u64 now;
 	__u64 end_time;
 
-	__u64 num_kvs_seen_normal;
-	__u64 num_kvs_seen_nat_fwd;
-	__u64 num_kvs_seen_nat_rev;
-	__u64 num_kvs_deleted_normal;
-	__u64 num_kvs_deleted_nat_fwd;
-	__u64 num_kvs_deleted_nat_rev;
+	__u64 num_cleaned;
 };
-
-// sub_age calculates now-then assuming that the difference is less than
-// 1<<63.  Values larger than that are assumed to have wrapped (then>now) and
-// 0 is returned in that case.
-static __u64 sub_age(__u64 now, __u64 then)
-{
-	__u64 age = now - then;
-	if (age > (1ull<<63)) {
-		// Wrapped, assume that means then > now.
-		return 0;
-	}
-	return age;
-}
-
-#define CT_VALUE_FINS_SEEN_DSR(value) (value->a_to_b.fin_seen || value->b_to_a.fin_seen)
-#define CT_VALUE_FINS_SEEN_NON_DSR(value) (value->a_to_b.fin_seen && value->b_to_a.fin_seen)
-#define CT_VALUE_ESTABLISHED(value) (value->a_to_b.syn_seen && value->a_to_b.ack_seen && value->b_to_a.syn_seen && value->b_to_a.ack_seen)
-#define CT_VALUE_DSR(value) (value->flags & CALI_CT_FLAG_DSR_FWD)
-
-// max_age returns the maximum age for the given conntrack "tracking" entry.
-static __u64 calculate_max_age(const struct calico_ct_key *key, const struct calico_ct_value *value)
-{
-	__u64 max_age;
-	switch (key->protocol) {
-	case IPPROTO_TCP:
-		if (value->a_to_b.rst_seen || value->b_to_a.rst_seen) {
-			max_age = __globals.tcp_reset_seen;
-		} else if (
-		    (CT_VALUE_DSR(value) && CT_VALUE_FINS_SEEN_DSR(value)) ||
-		    CT_VALUE_FINS_SEEN_NON_DSR(value)
-		) {
-			max_age = __globals.tcp_fins_seen;
-		} else if (CT_VALUE_ESTABLISHED(value) || CT_VALUE_DSR(value)) {
-			if (value->rst_seen) {
-				/* We have seen RST in the past, but we have seen traffic
-				 * since then so we want to be cautious and not tear down
-				 * the conntrack too soon in case the RST was spurious,
-				 * but we are also not sure if the connection is still
-				 * established.
-				 */
-				max_age = __globals.tcp_fins_seen;
-			} else {
-				max_age = __globals.tcp_established;
-			}
-		} else {
-			max_age = __globals.tcp_syn_sent;
-		}
-		break;
-	case IPPROTO_UDP:
-		max_age = __globals.udp_timeout;
-		break;
-	case IPPROTO_ICMP_46:
-		max_age = __globals.icmp_timeout;
-		break;
-	default:
-		max_age = __globals.generic_timeout;
-		break;
-	}
-	CALI_DEBUG("max_age %d", max_age / 1000000000);
-	return max_age;
-}
-
-static bool entry_expired(const struct calico_ct_key *key, const struct calico_ct_value *value, struct ct_iter_ctx *ctx)
-{
-	__u64 age = sub_age(ctx->now, value->last_seen);
-	__u64 max_age = calculate_max_age(key, value);
-	return age > max_age;
-}
-
-// process_ct_entry callback function for the conntrack map iteration.  Checks
-// the entry for expiry.  Expired normal entries are deleted inline.  Expired
-// NAT entries are queued for deletion in the cali_ccq map.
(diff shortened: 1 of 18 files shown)


felix/bpf/conntrack/bpf_scanner.go
```diff
@@ -215,91 +138,28 @@ func (s *BPFProgLivenessScanner) RunBPFExpiryProgram(opts ...RunOpt) error {
...
-		return fmt.Errorf("failed to parse cleanup program result: %w", err)
+		return nil, fmt.Errorf("failed to parse cleanup program result: %w", err)
 	}
 	log.WithFields(log.Fields{
 		"timeTaken": result.Duration,
 		"stats":     cr,
 	}).Debug("Conntrack cleanup result.")
-
-	// Record stats...
-	summaryCleanerExecTime.Observe(result.Duration.Seconds())
```
fasaxc, line 237: Can we keep the still-useful stats?  Time taken in this case, and stats for number of entries actually deleted?

felix/bpf/conntrack/scanner.go
```diff
@@ -156,14 +167,29 @@ func (s *Scanner) Scan() {
...
 					log.Debug("Deleting conntrack entry.")
 				}
-				cleaned++
-				return maps.IterDelete
+				revKey := []byte{}
+				if ctVal.Type() == TypeNATForward {
+					revKey = ctVal.ReverseNATKey().AsBytes()
+				}
+
+				val = cleanupv1.NewValue(revKey, uint64(ctVal.LastSeen()), uint64(ts))
```
fasaxc: Better to use if/else instead of unconditionally setting the variable and then overwriting it.

felix/bpf/conntrack/scanner.go
```diff
@@ -156,14 +167,29 @@ func (s *Scanner) Scan() {
 				"entry": ctVal,
 			}).Debug("Examining conntrack entry")
 		}
+		val = cleanupv1.Value{}
```
fasaxc: Better to use if/else instead of unconditionally setting the variable and then overwriting it.

fasaxc: It's a bit of a smell to need to do this kind of switch inline too.  I'd be tempted to do this:
```
type Scanner[CleanupV cleanupv1.ValueInterface] struct ...
```
Or if there are too many things to parameterise, uses a factory:
```
s.ipVersionedFactory.Value(...)
```

felix/bpf/conntrack/scanner.go
```diff
@@ -156,14 +167,29 @@ func (s *Scanner) Scan() {
...
+					revKey = ctVal.ReverseNATKey().AsBytes()
+				}
+
+				val = cleanupv1.NewValue(revKey, uint64(ctVal.LastSeen()), uint64(ts))
+				if s.ipVersion == 6 {
+					val = cleanupv1.NewValueV6(revKey, uint64(ctVal.LastSeen()), uint64(ts))
+				}
+				err := s.ctCleanupMap.Update(ctKey.AsBytes(), val.AsBytes())
+				if err != nil {
+					log.Debugf("error updating cleanup map. Entry may not be cleaned up %v", err)
```
fasaxc: An obvious potential failure is if the cleaner map gets full.

Rather than waiting until the end of the iteration, what if we triggered the BPF cleaner whenever we've got at least N (say 1000) entries ready to go (and then once at the end)?

felix/bpf/conntrack/bpf_scanner.go
```diff
@@ -40,88 +35,45 @@ const (
...
-			gaugeVecConntrackEntries,
-			counterVecConntrackEntriesDeleted,
-			summaryCleanerExecTime,
-		)
-	})
-}
-
-// BPFProgLivenessScanner is a scanner that uses a BPF program to scan the
+// BPFProgCleaner uses a BPF program to scan the
 // conntrack table for expired entries.  The BPF program does the entry
```
fasaxc, line 57: Out of date: no longer scans the conntrack table, it scans the cleanup table only

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,229 +24,66 @@ struct ct_iter_ctx {
...
-
-	// Still expired, delete both entries.  The forward key might be a dummy
-	// all-zeros key but we know that key won't exist so we just let
-	// cali_ct_delete_elem handle that.
-	if (cali_ct_delete_elem(fwd_key) == 0) {
-		ictx->num_kvs_deleted_nat_fwd++;
-	}
-	if (cali_ct_delete_elem(rev_key) == 0) {
-		ictx->num_kvs_deleted_nat_rev++;
+		cali_ccq_delete_elem(key);
```
fasaxc: Looks like the element only gets deleted if we delete the conntrack entry, so we'll leak a cleanup entry every time we decide not to delete an entry?

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -264,21 +101,10 @@ __attribute__((section("tc"))) int conntrack_cleanup(struct __sk_buff *skb)
 		// time.
```
fasaxc, line 87: Comment explaining how the cleaner works (above here) is out of date.

felix/bpf/conntrack/scanner.go
```diff
@@ -156,14 +167,29 @@ func (s *Scanner) Scan() {
...
+		}
 
 		for _, scanner := range s.scanners {
-			if verdict := scanner.Check(ctKey, ctVal, s.get); verdict == ScanVerdictDelete {
+			if verdict, ts := scanner.Check(ctKey, ctVal, s.get); verdict == ScanVerdictDelete {
 				if debug {
 					log.Debug("Deleting conntrack entry.")
 				}
-				cleaned++
-				return maps.IterDelete
```
fasaxc, line 166: Have we lost the option to fall back to user-space only?  ISTR that BPF cleaner needs a new-ish kernel; we should make sure we know what kernel is needed and get sign off if we're effectively ending support for older versions.

Then we should release note it.

If user-space is no longer supported, we should deprecate the config.

felix/bpf/conntrack/scanner.go
```diff
@@ -156,14 +167,29 @@ func (s *Scanner) Scan() {
...
+				revKey := []byte{}
+				if ctVal.Type() == TypeNATForward {
+					revKey = ctVal.ReverseNATKey().AsBytes()
+				}
+
+				val = cleanupv1.NewValue(revKey, uint64(ctVal.LastSeen()), uint64(ts))
+				if s.ipVersion == 6 {
+					val = cleanupv1.NewValueV6(revKey, uint64(ctVal.LastSeen()), uint64(ts))
+				}
+				err := s.ctCleanupMap.Update(ctKey.AsBytes(), val.AsBytes())
```
fasaxc: Is this an ordinary map; should we use a caching map so that we can do batch writes (once we add that feature to caching map!)?

If a caching map, it needs to be flushed.

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,229 +24,66 @@ struct ct_iter_ctx {
...
-	__u64 max_age = calculate_max_age(key, value);
-	return age > max_age;
-}
-
-// process_ct_entry callback function for the conntrack map iteration.  Checks
-// the entry for expiry.  Expired normal entries are deleted inline.  Expired
-// NAT entries are queued for deletion in the cali_ccq map.
-static long process_ct_entry(void *map, const void *key, void *value, void *ctx)
+// process_ccq_entry processes an entry in the "cleanup queue" map.  The map
+// is keyed on the reverse NAT entry's key, with the value storing the forward
```
fasaxc: Think yo have normal entries in here too now?

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,229 +24,66 @@ struct ct_iter_ctx {
...
+// entry's key (or a zero value if the forward entry is missing).
+static long process_ccq_entry(void *map, struct calico_ct_key *key, struct cali_ccq_value *value, void *ctx)
 {
-	const struct calico_ct_key *ct_key = key;
-	struct calico_ct_value *ct_value = value;
-	struct calico_ct_value *rev_value;
+	bool isNatForward = false;
```
fasaxc: C naming scheme is `is_nat_forward`

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,229 +24,66 @@ struct ct_iter_ctx {
...
+	// If the entry is Normal or Reverse, lookup_key points to the entry.
+	// Delete the entry and if successful, increment the counter.
+	// If the entry is forward, we have 2 cases,
+	// Rev_entry present - lookup_key points to the reverse entry and key points to the
+	// forward entry. Delete both the entries togethere.
+	// Rev_entry not present - lookup_key points to the forward entry.
+	// Delete it.
+	if(!cali_ct_delete_elem(lookup_key)) {
+		ictx->num_cleaned++;
+		if (isNatForward && !cali_ct_delete_elem(key)) {
```
fasaxc: Not sure I grok why this is now nested; when there's a NAT pair, we should decide whether to delete both or neither and then either try to delete both or neither.

The first delete could fail due to LRU age out, for example.  (Pretty unlikely given we just looked it up but still...)

fasaxc (changes requested):

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,229 +24,66 @@ struct ct_iter_ctx {
...
 
-			// Entry is not fresh so it looks invalid.  Clean it up.
-			CALI_DEBUG("  INVALID: Forward NAT entry with no reverse entry, cleaning up.");
-			if (cali_ct_delete_elem(ct_key) == 0) {
-				ictx->num_kvs_deleted_nat_fwd++;
+	const struct calico_ct_value *actual_ct_value = cali_ct_lookup_elem(lookup_key);
+	if (actual_ct_value) {
+		if (isNatForward) {
+			// If the reverse key has expired, go to delete.
+			if (actual_ct_value->last_seen == value->rev_last_seen) {
```
fasaxc: Might be good to do more consistency checks; in case of NAT entry, lookup both fwd and reverse; check that the forward entry still relates to the reverse one (i.e. no new SYN has come in and triggered a new forward entry with a different reverse).

felix/bpf/conntrack/scanner.go
sridhartigera: ```BPFConntrackCleanupMode``` is already deprecated and defaulted to userspace.

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,237 +24,75 @@ struct ct_iter_ctx {
...
-// sub_age calculates now-then assuming that the difference is less than
-// 1<<63.  Values larger than that are assumed to have wrapped (then>now) and
-// 0 is returned in that case.
-static __u64 sub_age(__u64 now, __u64 then)
-{
-	__u64 age = now - then;
-	if (age > (1ull<<63)) {
-		// Wrapped, assume that means then > now.
-		return 0;
+static CALI_BPF_INLINE bool if_ct_key_equal(struct calico_ct_key *key1, struct calico_ct_key *key2) {
```
fasaxc:
```suggestion
static CALI_BPF_INLINE bool ct_key_equal(struct calico_ct_key *key1, struct calico_ct_key *key2) {
```

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,237 +24,75 @@ struct ct_iter_ctx {
...
+	if (!key1 || !key2) {
+		return false;
```
fasaxc:
```suggestion
		return key1 == key2;
```
Not sure if relevant, but sometimes we want NULLs to compare equal.

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,237 +24,75 @@ struct ct_iter_ctx {
...
-
-			// Entry is not fresh so it looks invalid.  Clean it up.
-			CALI_DEBUG("  INVALID: Forward NAT entry with no reverse entry, cleaning up.");
-			if (cali_ct_delete_elem(ct_key) == 0) {
-				ictx->num_kvs_deleted_nat_fwd++;
+	} else {
+		// Its a NAT forward entry with a valid reverse key.
+		struct calico_ct_value *nat_fwd_value = cali_ct_lookup_elem(key);
+		struct calico_ct_key *rev_key = &value->rev_key;
+		// Check if the fwd key still points to the same reverse key.
```
fasaxc, line 52: If it doesn't point to the same rev key then we should still delete the reverse entry listed in the cleanup queue?  It's no longer reachable from a forward entry.

felix/bpf/conntrack/scanner.go
```diff
@@ -127,6 +138,66 @@ func NewScanner(ctMap maps.Map, kfb func([]byte) KeyInterface, vfb func([]byte)
...
+		// same as that of entry's ts. Just go ahead with deletion.
+		if ts == rev_ts {
+			dummy := s.versionHelper.dummyKey()
+			s.updateCleanupMap(key, dummy, ts, rev_ts)
+			return
+		}
+		_, ok := s.keyTracker[revKey]
+		if !ok {
+			// Reverse entry not seen by the scanner. Don't queue it up for deletion.
+			// Wait to see if we the scanner sees the reverse entry.
```
fasaxc: Using the cleanup value as the value type for this map makes the case below very confusing.  
```
fwdKey := revVal.ReverseNATKey()
```
Suggest making a new struct that isn't tied to the cleanup map, just to avoid confusion.  Or, if that's awkward, you could change `ReverseNATKey()` to `OtherNATKey()` and comment that `keyTracker` stores the opposite direction of mapping to the cleanup queue.

felix/bpf/conntrack/scanner.go
```diff
@@ -127,6 +138,66 @@ func NewScanner(ctMap maps.Map, kfb func([]byte) KeyInterface, vfb func([]byte)
 		maxEntries:                   ctMap.Size(),
 		autoScale:                    strings.ToLower(autoScalingMode) == "doubleiffull",
 		configChangedRestartCallback: configChangedRestartCallback,
+		bpfCleaner:                   bpfCleaner,
+		keyTracker:                   make(map[KeyInterface]cleanupv1.ValueInterface),
```
fasaxc:
```suggestion
		revNATKeyToFwdNATInfo:                   make(map[KeyInterface]cleanupv1.ValueInterface),
```
Would make it unambiguous what you're storing in here.

felix/bpf/conntrack/scanner.go
```diff
@@ -158,22 +230,57 @@ func (s *Scanner) Scan() {
...
+	// We have scanned the entire map.
+	// Lets add it to the cleanup map.
+	if len(s.keyTracker) > 0 {
+		for k, v := range s.keyTracker {
+			// This is a forward entry and we haven't seen the rev entry.
+			// Maybe deleted by LRU
+			revKey := v.ReverseNATKey()
+			ts := v.Timestamp()
+			revTS := v.RevTimestamp()
+			if revKey.Proto() != 0 {
```
fasaxc: Might exceed map capacity?

felix/bpf/conntrack/scanner.go
```diff
@@ -208,6 +315,22 @@ func (s *Scanner) Scan() {
...
+	// Run the BPF cleanup program.
+	if s.bpfCleaner != nil {
+		if err := s.ctCleanupMap.ApplyAllChanges(); err != nil {
+			log.WithError(err).Warn("Failed to write updates to conntrack cleanup BPF map.")
+		}
+		cr, err := s.bpfCleaner.Run()
+		if err != nil {
+			log.WithError(err).Warn("Failed to run bpf conntrack cleaner.")
+		}
+		s.ctCleanupMap.Desired().DeleteAll()
```
fasaxc, line 344: We should do _something_ to tell the map that we know the dataplane values are now gone too; otherwise the map will drift out of sync (and its storage will grow and grow, I expect).

fasaxc (changes requested):

felix/bpf-gpl/conntrack_cleanup.c
sridhartigera: I feel its better for the rev key to be deleted in the next iteration of the scanner, WDYT?

felix/cachingmap/caching_map.go
```diff
@@ -80,13 +80,13 @@ func (c *CachingMap[K, V]) LoadCacheFromDataplane() error {
 type ReadOnlyMap[K comparable, V any] interface {
 	Get(k K) (v V, exists bool)
 	Iter(func(k K, v V))
+	DeleteAll()
```
fasaxc: The interface name is wrong now.  

Now that we've learned that it's useful  to allow the user of the API to tell us about changes to the dataplane side.  I suggest just exposing both as ReadWrite, with a comment explaining what it means to update the Dataplane side (it means that we know the map was updated out-of-band)

fasaxc (approved):
One minor issue

felix/bpf/conntrack/scanner.go
```diff
@@ -158,22 +236,65 @@ func (s *Scanner) Scan() {
...
+				// NAT entry has expired.
+				if ctVal.Type() != TypeNormal {
+					s.handleNATEntries(ctKey, ctVal, uint64(ts))
+					continue
+				}
+				dummy := s.versionHelper.dummyKey()
+				s.updateCleanupMap(ctKey, dummy, uint64(ts), uint64(ts))
 			}
 		}
+		if numExpired > 0 && numExpired%1000 == 0 {
```
tomastigera: should the 100 be a configurable? Or at least defined as a constant?

felix/bpf/conntrack/scanner.go
```diff
@@ -208,6 +329,23 @@ func (s *Scanner) Scan() {
...
+	// Run the BPF cleanup program.
+	if s.bpfCleaner != nil {
+		if err := s.ctCleanupMap.ApplyAllChanges(); err != nil {
+			log.WithError(err).Warn("Failed to write updates to conntrack cleanup BPF map.")
+		}
+		cr, err := s.bpfCleaner.Run()
+		if err != nil {
+			log.WithError(err).Warn("Failed to run bpf conntrack cleaner.")
+		}
+		s.ctCleanupMap.Desired().DeleteAll()
```
tomastigera, line 344: should you do the cleanup if you failed above?

felix/bpf/conntrack/scanner.go
```diff
@@ -127,6 +138,71 @@ func NewScanner(ctMap maps.Map, kfb func([]byte) KeyInterface, vfb func([]byte)
...
+}
+
+func (s *Scanner) handleNATEntries(key KeyInterface, val ValueInterface, rev_ts uint64) {
+	ts := uint64(val.LastSeen())
+	if val.Type() == TypeNATForward {
+		revKey := val.ReverseNATKey()
+		// If reverse key is not present in the conntrack map,
+		// timestamp returned from the scanner will match the
+		// same as that of entry's ts. Just go ahead with deletion.
+		if ts == rev_ts {
```
tomastigera, line 177: This feels a bit hacky :thinking:  took me quite a while to figure out why the ts would be the same

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,237 +24,75 @@ struct ct_iter_ctx {
...
-// 1<<63.  Values larger than that are assumed to have wrapped (then>now) and
-// 0 is returned in that case.
-static __u64 sub_age(__u64 now, __u64 then)
-{
-	__u64 age = now - then;
-	if (age > (1ull<<63)) {
-		// Wrapped, assume that means then > now.
-		return 0;
+static CALI_BPF_INLINE bool ct_key_equal(struct calico_ct_key *key1, struct calico_ct_key *key2) {
+	if (!key1 || !key2) {
```
tomastigera: I do not get this. The function is called from only a single place and neither key1 nor key2 are ever NULL. So this branch is never executed. Also the comparison within is pointer comparison. So they would would only equal if they are pointing to the same memory. But the keys are always taken from a value in a different map so they can never equal.

felix/bpf-gpl/conntrack_cleanup.c
```diff
@@ -24,237 +24,75 @@ struct ct_iter_ctx {
...
-	case IPPROTO_UDP:
-		max_age = __globals.udp_timeout;
-		break;
-	case IPPROTO_ICMP_46:
-		max_age = __globals.icmp_timeout;
-		break;
-	default:
-		max_age = __globals.generic_timeout;
-		break;
+	if ((key1->protocol == key2->protocol) &&
```
tomastigera: could use strcmp? I think there is a compiler built in as well https://docs.ebpf.io/linux/helper-function/bpf_strncmp/

felix/bpf/conntrack/scanner.go
sridhartigera: Yeah. Just wanted to avoid another lookup to check if reverse key is present or not.

felix/bpf-gpl/conntrack_cleanup.c
sridhartigera: ```bpf_strncmp``` requires the 3rd param to be readonly.

felix/bpf/conntrack/scanner.go
sridhartigera: I will define it as a constant.

felix/bpf/conntrack/scanner.go
sridhartigera: Every time we scan, we start with a clean slate. So I prefer cleaning up and leaving it for the subsequent scan. WDYT?

felix/bpf-gpl/conntrack_cleanup.c
sridhartigera: ```__builtin_memcmp``` works fine.

tomastigera (approved):

PR #10740, opened 2025-07-29
[CORE-11657] BPF: add packet rate QoS control

coutinhop:
## Description

Add the BPF dataplane implementation of the packet rate QoS control. Implement a token bucket filter (TBF) algorithm to limit ingress and/or egress packet rate(s). Use the ifstate map to keep the persistent parameters for the TBF. Add a UT and update QoS FVs for BPF coverage of packet rate controls.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Add support for packet rate QoS controls on the eBPF databplane.
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/globals.h
```diff
@@ -26,6 +26,10 @@ struct name {				\
 	__u8 iface_name[16];		\
 	__u32 log_filter_jmp;		\
 	__u32 jumps[40];		\
+	__s16 ingress_packet_rate; \
+	__s16 ingress_packet_burst; \
+	__s16 egress_packet_rate; \
+	__s16 egress_packet_burst; \
```
sridhartigera: nit: Alignment issue?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1031,6 +1040,10 @@ func (m *bpfEndpointManager) updateIfaceStateMap(name string, iface *bpfInterfac
 			iface.dpState.v6.policyIdx[hook.Egress],
 			iface.dpState.filterIdx[hook.Ingress],
 			iface.dpState.filterIdx[hook.Egress],
+			iface.dpState.qosInfo.packetRateTokens[hook.Ingress],
+			iface.dpState.qosInfo.packetRateTokens[hook.Egress],
+			iface.dpState.qosInfo.packetLastTime[hook.Ingress],
+			iface.dpState.qosInfo.packetLastTime[hook.Egress],
```
sridhartigera: I think we will overwrite the tokens and the timestamp, when felix restarts and programs are attached again. Better to validate this with UT. When we update, its better to read the timestamp and tokens from the map and write the same thing back.

sridhartigera (changes requested):

felix/dataplane/linux/bpf_ep_mgr.go
coutinhop: Right, I've added reading the QoS state (when on a workload) to the bpfEndpointManager and added coverage for it in the attach_test TestAttachWithMultipleWorkloadUpdate ut. Thanks for catching this!

sridhartigera (approved):
LGTM

PR #10763, opened 2025-08-04
Fine grained resource selection

### LINKED ISSUE #9857: allow fine-grained source/destination selection
We are currently re-working our network policies by introducing GlobalNetworkSets. We have labeled each network set with something like `networking.our.internal.stuff/networkset: <name>` to be able to select a single network set in our network policies. We are concerned that someone accidentally adds the same label to a pod or another resource and thus undermines our network policies.

## Expected Behavior

It would be cool if something similar to `projectcalico.org/orchestrator` exists that lets us specify the resource type we want to select. 

## Current Behavior

We can use `global()` to select across all global resources, but cannot limit this to just GlobalNetworkSets. We can use an external policy agent, like OPA or kyverno, to disallow setting our networkset labels on resources that are not network sets but this requires an additional system to be deployed.

## Possible Solution

Calico could automatically add a new label, e.g. `projectcalico.org/kind` to each calico resource so that we can select resources by kind and their labels within the same selector. In our case, we could write a selector like `networking.our.internal.stuff/networkset == "some-name" && projectcalico.org/kind == "GlobalNetworkSet"`

Another solution would be something like a `kind(GlobalNetworkSet)` expression that does the same, but does not require an extra label.

## Context

We want to be as strict as possible in our network policies and close any potential loophole that might exist in our setup.

knotseaborg:
I noticed that this issue hasn't been active for a month, so I'm giving it a try.

## Description

This draft addresses an enhancement suggestion for fine-grained selection of resources such as `GlobalNetworkSet`.

### Solution

I forced calico to set the label `projectcalico.org/kind` when a resource is created using the `clientv3` package. It seems to offer a consistent mechanism for resource selection.

### Tests

- Existing tests have been modified to verify the label after creation of a resource.
- Some tests in `calico/libcalico-go/lib/backend/syncersv1` have also been fixed by including the new label to prevent breakage.
---

@caseydavenport @sebhoss Let me know if this approach aligns with your intent, I can rework this draft.

## Related issues/PRs

fixes #9857

## Todos

- [x] Tests - Passed tests for `calico/libcalico-go`
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/etcdv3/conversion.go
+++ b/libcalico-go/lib/backend/etcdv3/conversion.go
@@ -82,6 +82,11 @@ func convertWatchEvent(e *clientv3.Event, l model.ListInterface) (*api.WatchEven
 		return nil, nil
 	}
 
+	// "projectcalico.org/kind" label is expected for certain resources
+	// for selection
+	model.AddKindLabel(oldKV)
+	model.AddKindLabel(newKV)
+
 	return &api.WatchEvent{
 		Old:  oldKV,
 		New:  newKV,
@@ -89,9 +94,7 @@ func convertWatchEvent(e *clientv3.Event, l model.ListInterface) (*api.WatchEven
 	}, nil
 }
 
-var (
-	ErrMissingValue = fmt.Errorf("missing etcd KV")
-)
+var ErrMissingValue = fmt.Errorf("missing etcd KV")
 
 // etcdToKVPair converts an etcd KeyValue into model.KVPair.
 func etcdToKVPair(key model.Key, ekv *mvccpb.KeyValue) (*model.KVPair, error) {
--- a/libcalico-go/lib/backend/etcdv3/etcdv3.go
+++ b/libcalico-go/lib/backend/etcdv3/etcdv3.go
@@ -197,6 +197,7 @@ func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPa
 		if len(getResp.Kvs) != 0 {
 			existing, _ = etcdToKVPair(d.Key, getResp.Kvs[0])
 		}
+		model.AddKindLabel(existing)
 		return existing, cerrors.ErrorResourceAlreadyExists{Identifier: keyCopy}
 	}
 
@@ -207,6 +208,8 @@ func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPa
 	d.Value = v
 	d.Revision = strconv.FormatInt(txnResp.Header.Revision, 10)
 
+	// "projectcalico.org/kind" label is required for certain resources when read by calico
+	model.AddKindLabel(d)
 	return d, nil
 }
 
@@ -214,6 +217,10 @@ func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPa
 // an ErrorResourceDoesNotExist error.  The ResourceVersion must be specified, and if
 // incorrect will return an ErrorResourceUpdateConflict error and the current entry.
 func (c *etcdV3Client) Update(ctx context.Context, d *model.KVPair) (*model.KVPair, error) {
+	// "projectcalico.org/kind" label should not be entered into the datastore. It is exclusively used by calico on reads only.
+	defer model.AddKindLabel(d)
+	model.RemoveKindLabel(d)
+
 	// Take a copy of the key before we default any policy names, we use this key for error returns to return the same name that user provided for update
 	keyCopy := d.Key
 
@@ -251,7 +258,6 @@ func (c *etcdV3Client) Update(ctx context.Context, d *model.KVPair) (*model.KVPa
 	).Else(
 		clientv3.OpGet(key),
 	).Commit()
-
 	if err != nil {
 		logCxt.WithError(err).Warning("Update failed")
 		return nil, cerrors.ErrorDatastoreError{Err: err}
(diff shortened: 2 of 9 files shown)


sebhoss:
I think it's great! I'm not quite sure how the selection of `clientv3` packages interacts with something like auto creation of host endpoints though. Does it use v3 as well or does it directly use v1 and therefore bypass the code in this PR?

CC @fasaxc

fasaxc:
Thanks for the PR.  IIUC, this will add the label on write to the datastore as long as you go through our client package.  TBH, I'm not sure if that's the best place to add the function because:

* It won't add the label to existing resources.  So, on upgrade, you'll have a mix of resources with and without it.
* It won't help folks who bypass our Go API and write CRDs directly.  Sadly there are quite a few folks who do this.
* It'll add the label to all our resources, not just the relevant ones.  This might cause datastore bloat or confuse controllers that try to reconcile labels.
* Someone malicious could change the label to bypass policy rules.

For the other labels that we auto-populate, we've tended to add them on read, when we convert datastore objects into our in-memory representation.  This has the downside that users can't see the labels with kubectl, but it does mean that we can easily retrofit a new special label and it prevents any sort of tampering.  I think I'd be inclined to do it that way (And then, if we later decide that making the labels "real" is important, we could do it for all the labels that we synthesize.)

It's not as generic, but perhaps the right place to change is the individual clients in libcalico-go/lib/backend/k8s/resources.

* WorkloadEndpoints are backed by pods, so there's a conversion function that's invoked that could add the label.
* The others are backed by a general-purpose CRD client, I might add an opt-in `applyKindLabel` flag to `customK8sResourceClient` that tells the generic client to add the extra label to those on read (need to cover the Get,List,Watch cases).

knotseaborg:
@fasaxc @sebhoss Thank you for the feedback. I am revising the draft PR and will push changes soon based on this suggestion.

> It's not as generic, but perhaps the right place to change is the individual clients in libcalico-go/lib/backend/k8s/resources.
> * WorkloadEndpoints are backed by pods, so there's a conversion function that's invoked that could add the label.
> * The others are backed by a general-purpose CRD client, I might add an opt-in applyKindLabel flag to customK8sResourceClient that tells the generic client to add the extra label to those on read (need to cover the Get,List,Watch cases).

knotseaborg:
Sorry, I didn't mean to close this draft (In my attempt to achieve a clean reset for the new approach, I accidentally closed it)

### Some updates

1. Commit 1 - `add kind label to customresource`
    * When `applyKindLabel` is enabled, `customk8sResourceClient` now adds the `projectcalico.org/kind` label to resources during Get, List, and Watch operations.

3. Commit 2 - `test kind label with globalnetworkset CRUD`
    * To validate this approach, I set `applyKindLabel=true` from the `NewGlobalNetworkSetClient` and updated related e2e tests.
    * Added label matching for `GlobalNetworkSet`-reads from the k8s datastore. I'm unsure if these test modifications fully align with expectations.

### Did the tests pass?

 No, but `libcalico-go` tests failed for both the `master` & this branch. The tests don't reach the part where updates can be validated. 

@fasaxc Could you please take a look at my commits to make sure that they align with your intent? Any feedback will be helpful.

fasaxc:
Yes, that's the kind of change I had in mind for the CRD client; but now I think about it, it'll be confusing if we don't also do this for etcd.  I still think we should add the labels on read, but we should make sure that both datastore drivers return the new label for that subset of resources.  Perhaps you could call a common function from both dataplanes that checks if the resource is one of the kinds that we should label and, if so, does the labeling.  You also need to tweak the workload endpoint client, which doesn't use the CRD client.

I think the resources that should get labels are:
* NetSets
* GlobalNetsets
* WorkloadEndpoints
* HostEndpoints

knotseaborg:
@fasaxc Thank you for the feedback! I will work on implementing this approach.
> Perhaps you could call a common function from both dataplanes that checks if the resource is one of the kinds that we should label and, if so, does the labeling. You also need to tweak the workload endpoint client, which doesn't use the CRD client.

knotseaborg:
Hello @fasaxc 

Since our intention is to let the label live when calico uses it but never have it stored, I have a few concerns.
1. I think `create` which is used to store a resource, acts as a read which returns the created resource. Should we add the label to the resource returned by `create` as well for consistency?
2. I suspect that the `update` operation could potentially write the label into storage. Should it be tweaked so as to strip the kind label before write?

caseydavenport:
@knotseaborg @fasaxc what's the status on this guy? Seems like a nice one to have if we can!

knotseaborg:
@caseydavenport @fasaxc @sebhoss Sorry for the stale draft, guys. 

The last time I looked, I was blocked by non-deterministic test failures. I will revisit it this weekend and report back with the status.

If you've got any additional context or direction, I can adjust to that.

caseydavenport:
Hey @knotseaborg @fasaxc, thanks for the work on this! I've been thinking about this feature some more and I think we should take a slightly different approach.

The label-on-read approach has a problem in v3 CRD mode, where users interact with the Kubernetes API directly and never go through our Go client. Policy selectors evaluate against what's stored in the datastore, so a synthesized label that only exists in-memory won't be visible to the selector engine in that mode.

Rather than trying to get the label stored reliably (which would need a mutating admission policy, migration logic for existing resources, etc), I think the cleaner solution is to add a new `kind` field directly on the rule's `source`/`destination` (the EntityRule type). Something like:

```yaml
kind: GlobalNetworkPolicy
spec:
  ingress:
  - action: Allow
    source:
      kind: GlobalNetworkSet
      selector: "networking.internal/networkset == 'dns-servers'"
```

This avoids the label machinery entirely - Felix would filter the endpoint sets by kind when evaluating the rule. When `kind` is omitted, behavior is unchanged (matches all resource types, same as today). Allowed values would be `GlobalNetworkSet`, `NetworkSet`, `HostEndpoint`, `WorkloadEndpoint`, and/or `Pod`.

It's a bigger change than the label approach since it touches the API types and Felix's policy evaluation. What do you guys think? I'd be happy to help implement this if you think it's suitable.

fasaxc:
@caseydavenport how would your implementation work differently? The easiest way that I can see to make the `kind:` thing work is to map it onto a label that we synthesize in the calculation graph (for example).  If we don't make a label then all the indexes/selectors will need to track type or you'll need one index per type, which would be a big change (possibly increasing RAM usage a lot).

caseydavenport:
@fasaxc yeah I suppose we could translate to a label internally within Felix still (and from the sounds of it, probably should!)

The main change in implementation would be that the API types themselves don't have the label, and the UX from the user's point of view wouldn't be done via label selection.

PR #10782, opened 2025-08-07
Update e2e hep to correctly check for hostEndpoints enabled

MichalFupso:
## Description
Updates the HostEndpoint tests to check if both AutoCreate and CreateDefaultHostEndpoint are enabled before running tests and returns the kcc into the previous state after the test have finished.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/e2e/pkg/tests/hostendpoints/auto_hostendpoints.go
+++ b/e2e/pkg/tests/hostendpoints/auto_hostendpoints.go
@@ -14,6 +14,7 @@ package hostendpoints
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


e2e/pkg/tests/hostendpoints/auto_hostendpoints.go
```diff
@@ -265,49 +271,42 @@ func getNodeHostname(node v1.Node) string {
...
-		toggle = "Disabled"
-	}
-
+// updateHostEndpointConfig updates the HostEndpointConfiguration to the desired state
+func updateHostEndpointConfig(client ctrlclient.Client, desiredKCC v3.KubeControllersConfiguration) {
 	// Get the kubecontrollersconfiguration and patch it to toggle the
 	// auto-creation of host endpoints.
-	var kcc v3.KubeControllersConfiguration
-	err := client.Get(context.Background(), types.NamespacedName{Name: "default"}, &kcc)
+	var currentKCC v3.KubeControllersConfiguration
```
caseydavenport, line 278: I'm not sure I understand why we query this yet again here. IIUC, we're:

- Querying the original KCC and holding on to it Iline 94)
- Making a copy, modifying it in memory. (line 101-106)
- Querying the same KCC again here
- Updating the new object in memory based on the copy
- Writing the second updated object back to the API.

caseydavenport: I guess we need to ensure the revision information is up to date when we revert to the original state?

I suppose that is reason enough, but it does seem a little bit clunky.

MichalFupso: We might loose the revision info if the kcc has originally heps disabled. Because of that we have to query it again. It is a bit clunky, but in my opinion better in comparison of just keeping track of autoCreate and createDefaultHostEndpoints.

PR #10820, opened 2025-08-14
Speed up prerequisites job

fasaxc:
## Description

Bump up the size of the Prerequisites job runner.  It's mainly CPU bound, and it's been getting quite slow (8+minutes in OS).  this brings it down to 5min

Also tried upping instance size and splitting go vet into its own job, that resulted in one 2 min job and one 4 min job.  Not sure if that's worth it.

## Related issues/PRs

Also tried running some bits in parallel, but he saving was only 10s: https://github.com/projectcalico/calico/pull/10818

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/lib.Makefile
+++ b/lib.Makefile
@@ -669,7 +669,7 @@ REPO_DIR=$(shell if [ -e hack/format-changed-files.sh ]; then echo '.'; else ech
 .PHONY: fix-changed go-fmt-changed goimports-changed
 # Format changed files only.
 fix-changed go-fmt-changed goimports-changed:
-	if [ -z "$(SKIP_FIX_CHANGED)" ]; then \
+	if [ "$(SKIP_FIX_CHANGED)" != "true" ]; then \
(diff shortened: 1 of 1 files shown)


stevegaossou:
> Also tried upping instance size and splitting go vet into its own job, that resulted in one 2 min job and one 4 min job. Not sure if that's worth it.

I guess the reduction from 5 min to 4 min doesn't feel as satisfying as going from 8 min to 5 min, but would there be any advantages for future maintenance of the Prerequisites from having the split?

fasaxc:
>I guess the reduction from 5 min to 4 min doesn't feel as satisfying as going from 8 min to 5 min, but would there be any advantages for future maintenance of the Prerequisites from having the split?

I was more thinking of the cost/impact on CI queue; I'm already doubling the size of the runner, so, on top of that, we'd be using two for almost as long.

PR #10824, opened 2025-08-14
Fix networking-calico problems reported by flake8

nelljerram:
In my editor setup (Emacs), the editor uses `pylsp` to identify and flag coding issues.  `pylsp` uses `flake8` to identify PEP8/lint violations.  flake8 can also be run directly with

    make -C networking-calico flake8

This commit reduces the total number of flake8 moans (`flake8 | wc -l`) from 472 to 14, with the fixed moans being as follows.

- Line too long.  We configure line length 88 in `.flake8` to match black's default, instead of 79. Some lines then still needed manual attention to get them within 88.

- Import ordering.  We now follow flake8's "cryptography" default, which means stdlib first in one group, then 3rd party packages each in their own group, then networking-calico itself.

- E203, E402 and W503 moans.  We suppress these because they explicitly disagree from what black does.

- A few trivial other issues.

There are two places where we need eventlet monkey-patching, and this causes moans because it breaks
import ordering and involves a non-import call before other imports.  Also we weren't doing this as
early as possible, which is what is advised.  Those are now addressed by:

1. moving the monkey-patch as early as possible
2. adding `noqa` to the following import line, to suppress the moan there.

### DIFF
--- a/networking-calico/.flake8
+++ b/networking-calico/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+max-line-length = 88
+ignore = E203,E402,W503
+application-import-names = networking_calico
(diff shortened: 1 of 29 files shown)


nelljerram:
Thanks @fasaxc .  Yes, I added the one true way in #10823 .  I haven't yet made CI enforce this, but that is coming very soon.

PR #10718, opened 2025-07-23
Set DSCP for workload's egress traffic leaving cluster

mazdakn:
## Description

Set DSCP for egress traffic leaving cluster when `qos.projectcalico.org/dscp` annotation is set on a workload endpoint, i.e. pod.

This PR includes:
- Adding a new type named `DSCP` that holds numeric values between 0-63 and common string values.
- Libcalico-go changes: Adding the new annotation to workload endpoint.
- Calc graph: 
  - Defining a new protobuf message called `QosPolicy`.
  - conversion from annotation value to `QosPolicy` for workload endpoints.
- Dataplane: 
  - new dataplane manager to watch for workload endpoints and potential assigned `QosPolicy`.
  - new iptables/nftables action for setting DSCP.
  - new static chain to jump to QoS policy rules in post routing chain of mangle table.
-  Felix FVs

The follow up PRs will cover:
- Adding DSCP for host endpoints: https://github.com/projectcalico/calico/pull/10825
- Optimization in nftables by using verdict map: https://github.com/projectcalico/calico/pull/10827
- Adding support in ebpf dataplane: 
  - https://github.com/projectcalico/calico/pull/10881
  - https://github.com/projectcalico/calico/pull/10934
  - https://github.com/projectcalico/calico/pull/10997
- Docs: https://github.com/tigera/docs/pull/2290

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Set the Differentiated Services Code Point (DSCP) on an endpoint's outgoing network traffic leaving cluster.
This is done by adding "qos.projectcalico.org/dscp" annotation to workloads and host endpoints.
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #10825, opened 2025-08-15
Set DSCP for host endpoints

mazdakn:
## Description

Setting DSCP for host endpoint. This is a follow up for the first PR: https://github.com/projectcalico/calico/pull/10718

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/rules/static.go
```diff
@@ -1067,10 +1067,11 @@ func (r *DefaultRuleRenderer) StaticManglePostroutingChain(ipVersion uint8) *gen
 
 	ipConf := r.ipSetConfig(ipVersion)
 	allIPsSetName := ipConf.NameForMainIPSet(IPSetIDAllPools)
+	allHostsSetName := ipConf.NameForMainIPSet(IPSetIDAllHostNets)
 	rules = append(rules, generictables.Rule{
 		Match: r.NewMatch().
-			SourceIPSet(allIPsSetName).
-			NotDestIPSet(allIPsSetName),
+			NotDestIPSet(allIPsSetName).
+			NotDestIPSet(allHostsSetName),
```
mazdakn: Had to make this change to be able to check for packets destined for outside of cluster, by checking that destination is not in any of IP pools, nor in any of hosts.

caseydavenport: Not sure if this is a valid scenario, but what about traffic that comes into a node and is re-routed to another destination outside of the cluster?

Should we still have a source match on allIPsSetName + allHostsSetName?

libcalico-go/lib/backend/syncersv1/updateprocessors/hostendpointprocessor.go
```diff
@@ -87,3 +98,26 @@ func convertHostEndpointV2ToV1(kvp *model.KVPair) (*model.KVPair, error) {
...
+		dscp := numorstring.DSCPFromString(str)
+		err := dscp.Validate()
+		if err != nil {
+			errs = append(errs, fmt.Errorf("error parsing DSCP annotation: %w", err))
+		} else {
+			qosControls.DSCP = &dscp
+		}
+	}
+
+	// return nil if no control is configured
```
caseydavenport: You could probably simplify this a bit by simply initializing a nil pointer to start with?

e.g.,

```go
var qosControls *model.QoSControls

if str, found := ... {
  qosControls = &model.QosControls{DSCP: &dscp}
}
return qosControls
```

no big deal though

mazdakn: Correct. I took this part from workload endpoint conversion, and didn't check much. Will fix it.

felix/rules/static.go
mazdakn: I think calico as gateway is a valid use case. I thought about this and I was not sure what would be the best way to  check source is either allIPsSetName + allHostsSetName but dest is not in one rule.
Maybe the right approach is to add two rules instead. WDYT?

caseydavenport: I guess the question is, do we want to apply DSCP to those packet paths? If we do, then what you have here is fine! If we don't, then maybe we need to tweak.

I suppose we could just maintain additional IP sets? Or we could it with multiple rules.

mazdakn: based on the current implementation we don't. This rule jump to the target chain, where we match based on source address, where source addresses are from either wep or hep. In gateway mode, packets in transit won't match any rule in that chain.

caseydavenport: ahh right. ok.

caseydavenport (approved):

PR #10843, opened 2025-08-18
bpf: use libbpf instead of bpftool to create maps

coutinhop:
## Description

Replace calls to exec.Command() of the bpftool binary with direct calls to libbpf in felix/bpf/maps/maps.go

Generate .o "map stub" files from .c files that include map headers and have a simple XDP_PASS program in order to load maps with BTF data.

Update includes in headers to fix errors/warnings.

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/parsing6.h
```diff
@@ -6,6 +6,7 @@
 #define __CALI_PARSING6_H__
 
 #include "bpf.h"
+#include "counters.h"
```
tomastigera, line 9: I wonder why all these includes were added. Specifically, this file does not use counters. If it was to make the stubs compilable, perhaps it should be added into the stub?

felix/bpf-gpl/Makefile
```diff
@@ -59,18 +59,65 @@ UT_C_FILES:=$(shell find ut -name '*.c')
...
+MAP_STUB_CS := $(MAP_OBJS:.o=.c)
+
+# Map stub ojbs cannot have '-emit-llvm' in flags
+MAP_CFLAGS := $(subst -emit-llvm,,$(CFLAGS))
+
+# Generate _map_stub.c
+define generate_stub
+	echo '#include "bpf.h"' > $@; \
+	for header in $^; do \
+		echo "#include \"$${header}\"" >> $@; \
```
tomastigera, line 78: guess all the includes are because of this unordered includes and so that creates wacky interdependencies between the .h files. Is there a value in generating this automatically and then manually "fixing" all the .h rather than creating the stubs manually once?

felix/bpf/libbpf/libbpf_api.h
```diff
@@ -481,3 +485,70 @@ int num_possible_cpu()
...
+		create_attr.btf_fd = 0;
+		create_attr.btf_key_type_id = 0;
+		create_attr.btf_value_type_id = 0;
+		break;
+	case BPF_MAP_TYPE_STRUCT_OPS:
+		create_attr.btf_value_type_id = 0;
+		break;
+	default:
+		break;
+	}
```
tomastigera: Afaict this only sets to zero what should be already zeroed https://github.com/libbpf/libbpf/blob/fc687b8ee934b7331fa6561311346ce8cf8086fa/src/libbpf_common.h#L60-L64

felix/bpf/libbpf/libbpf_api.h
```diff
@@ -481,3 +485,70 @@ int num_possible_cpu()
 {
     return libbpf_num_possible_cpus();
 }
+
+int create_bpf_map(enum bpf_map_type type, unsigned int key_size, unsigned int value_size,
+                   unsigned int max_entries, unsigned int flags, const char *name) {
```
tomastigera:
```suggestion
                   unsigned int max_entries, unsigned int flags, const char *name)
                   {
```

felix/bpf/libbpf/libbpf_api.h
```diff
@@ -481,3 +485,70 @@ int num_possible_cpu()
...
+		char *cp, errmsg[STRERR_BUFSIZE];
+
+		err = -errno;
+		cp = libbpf_strerror_r(err, errmsg, sizeof(errmsg));
+		printf("libbpf warn: Error in bpf_map_create(%s):%s(%d).\n", name, cp, err);
+	}
+	return fd;
+}
+
+int find_bpf_map_by_name( const char *name) {
```
tomastigera:
```suggestion
int find_bpf_map_by_name( const char *name)
{
```

felix/bpf/libbpf/libbpf_api.h
```diff
@@ -481,3 +485,70 @@ int num_possible_cpu()
...
+	while (!bpf_map_get_next_id(id, &id)) {
+		int fd = bpf_map_get_fd_by_id(id);
+		if (fd < 0)
+			continue;
+		struct bpf_map_info info = {};
+		__u32 info_len = sizeof(info);
+		if (!bpf_obj_get_info_by_fd(fd, &info, &info_len)) {
+			if (strcmp(info.name, name) == 0) {
+				// found
+				break;
```
tomastigera: This breaks the `while` and jumps to return -1, that should not be, right?

felix/bpf/libbpf/libbpf_api.h
```diff
@@ -481,3 +485,70 @@ int num_possible_cpu()
...
+			}
+		}
+		close(fd);
+		return fd;
```
tomastigera: this returns whenever you get fd and you do not have the name match, so if you do not have a match in your first iteration, this will never make second iteration.

Also, if you close the fd, it is no more valid, right? Don't you need to keep the fd open?

felix/bpf-gpl/parsing6.h
coutinhop: Initially, I had a stub .c (and .o) file for each .h, and these were added to address errors when including the .h files by themselves. I figured there's no harm in leaving the includes (since they are all guarded by #ifdef and #defines). Let me know if you'd like me to revert them (I'll probably need to leave a few for whatever errors we'd have with the stubs as they are now).

felix/bpf-gpl/Makefile
coutinhop: Again, I think this was because I was initially doing several stubs, each with a single .h include. I think the extra value in generating them like this is less pollution in the repo and a central place to manage it (the Makefile). What do you think?

felix/bpf/libbpf/libbpf_api.h
coutinhop: Good point! I copied this from a similar function in libbpf.c, but since I didn't use the setup prior to that, I also don't need this `switch`, will remove!

felix/bpf/libbpf/libbpf_api.h
```diff
@@ -489,3 +493,43 @@ int num_possible_cpu()
...
+		if (fd < 0)
+			continue;
+		struct bpf_map_info info = {};
+		__u32 info_len = sizeof(info);
+		if (!bpf_obj_get_info_by_fd(fd, &info, &info_len)) {
+			if (strcmp(info.name, name) == 0) {
+				// found
+				return fd;
+			}
+		}
```
tomastigera: you need to close(fd) of any map you did not return otherwise you leak fds

felix/Makefile
```diff
@@ -538,7 +541,7 @@ bin/bpf_debug.test: $(GENERATED_FILES) $(shell find bpf/ -name '*.go')
 	$(DOCKER_GO_BUILD_CGO) go test ./bpf/ut -c -gcflags="-N -l" -o $@
 
 .PHONY: ut-bpf
-ut-bpf:  $(LIBBPF_A) bin/bpf_ut.test bin/bpf.test build-bpf
+ut-bpf: $(LIBBPF_A) bin/bpf_ut.test bin/bpf.test build-bpf $(BPF_GPL_MAP_STUB_O_FILES)
```
tomastigera:
```suggestion
ut-bpf: $(LIBBPF_A) bin/bpf_ut.test bin/bpf.test build-bpf
```

The stub files are already part of build-bpf, right?

felix/bpf-gpl/qos.h
```diff
@@ -22,17 +22,14 @@ static CALI_BPF_INLINE int enforce_packet_rate_qos(struct cali_tc_ctx *ctx)
 		CALI_DEBUG("packet rate QoS: ifstate not found, accepting packet");
 		return TC_ACT_UNSPEC;
 	}
-
 #if CALI_F_INGRESS
 	if (INGRESS_PACKET_RATE == 0) {
 		ifstate->ingress_packet_rate_tokens = -1;
-		cali_iface_update_elem(&ifindex, ifstate, BPF_ANY);
```
tomastigera: I think changes to this file should be excluded from this change

felix/bpf/maps/maps.go
```diff
@@ -44,6 +41,49 @@ var (
...
+	"cali_v6_ct":      "ipv6_map_stub.o",
+	"cali_v6_ct_nats": "ipv6_map_stub.o",
+	"cali_v6_fsafes":  "ipv6_map_stub.o",
+	"cali_v6_ip_sets": "ipv6_map_stub.o",
+	"cali_v6_nat_aff": "ipv6_map_stub.o",
+	"cali_v6_nat_be":  "ipv6_map_stub.o",
+	"cali_v6_nat_fe":  "ipv6_map_stub.o",
+	"cali_v6_routes":  "ipv6_map_stub.o",
+	"cali_v6_srmsg":   "ipv6_map_stub.o",
+}
```
tomastigera: Almost feels like you can say

if cali_v4* then "ipv4_map_stub.o"
elif cali_v6* then ""ipv6_map_stub.o"
elif xdp* then "xdp_map_stub.o"
else "common_map_stub.o"


The benefit would be that adding a new map would not need to make a change here :thinking:

felix/bpf-gpl/parsing6.h
tomastigera: i think in the end of the day, the includes do not matter much

tomastigera (commented):
lgtm, some nits, would be nice if you excluded the qos.h changes, but I am fine with the rest. Made some Makefile changes myself.

felix/bpf/maps/maps.go
coutinhop: Right, excellent idea! It wouldn't have worked at first because I was relying on this map for maps that are not present in the stubs (via verifying the `ok` var), but now that the logic was reworked to fall back to non-BTF, this will work (and make things simpler)! Will do!

felix/Makefile
coutinhop: Right, thanks for catching this!

tomastigera (approved):

PR #10855, opened 2025-08-20
Speed up IPAM affinity lookup.

fasaxc:
## Description

Add labels to the Kubernetes encoding of IPAM block affinities so that they can be looked up efficiently with a label selector.

The KDD block affinity client automatically adds the labels on write.

To handle pre-existing bloacks, each node upgrades its own block affinities the first time the CNI IPAM plugin runs after upgrade.

For compatibility, add support for label selectors in the etcd backend (but these are implemented client-side).

## Related issues/PRs

CORE-11745
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/ipam/ipam.go
+++ b/libcalico-go/lib/ipam/ipam.go
@@ -26,6 +26,7 @@ import (
 	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/sync/semaphore"
+	"k8s.io/apimachinery/pkg/labels"
 
 	libapiv3 "github.com/projectcalico/calico/libcalico-go/lib/apis/v3"
 	bapi "github.com/projectcalico/calico/libcalico-go/lib/backend/api"
@@ -2349,4 +2350,76 @@ func (c ipamClient) getReservedIPs(ctx context.Context) (addrFilter, error) {
 		}
 	}
 	return cidrs, nil
+}
+
+func (c ipamClient) UpgradeHost(ctx context.Context, nodeName string) error {
+	delay := 100 * time.Millisecond
+	for {
+		err := c.attemptUpgradeHost(ctx, nodeName)
+		if err == nil {
+			return nil
+		}
+		log.WithError(err).Errorf("Failed to upgrade qIPAM blocks of this host.")
+		select {
+		case <-ctx.Done():
+			return err
+		case <-time.After(delay):
+			delay *= 2
+		}
+		log.WithError(err).Error("Retrying...")
+	}
+}
+
+func (c ipamClient) attemptUpgradeHost(ctx context.Context, nodeName string) error {
+	// Upgrade the block affinities that belong to this host.  Add labels that
+	// allow for efficient lookups in KDD.  Note that etcd still stores the
+	// old "v1" model.BlockAffinity objects directly so this List() will
+	// return no results in etcd.
+	log.Info("Upgrading any IPAM affinities from older versions....")
+
+	// We know that already-upgraded blocks will have the affinityType label,
+	// we can list only non-upgraded blocks.  We can't limit to any particular
+	// host here because that's the whole point of adding the new labels!
+	ls, err := labels.Parse("!affinityType")
+	if err != nil {
+		return err
+	}
+	kvs, err := c.client.List(ctx, model.ResourceListOptions{
+		Kind:          libapiv3.KindBlockAffinity,
+		LabelSelector: ls,
+	}, "")
+	if err != nil {
+		return err
+	}
+
+	var errs []error
+	numUpgraded := 0
+	for _, kv := range kvs.KVPairs {
+		if val, ok := kv.Value.(*libapiv3.BlockAffinity); ok {
+			if val.Spec.Node != nodeName {
+				// Only do _our_ affinities to avoid n x n conflicts with other
+				// nodes also doing this operation.
+				continue
+			}
+			libapiv3.EnsureBlockAffinityLabels(val)
+			_, err := c.client.Update(ctx, kv)
+			if err != nil {
+				errs = append(errs, err)
+				if errors.Is(err, context.DeadlineExceeded) {
+					break
+				}
+				continue
+			}
+			numUpgraded++
+		}
+	}
+	if numUpgraded == 0 && len(errs) == 0 {
+		log.Info("No affinities needed to be upgraded.")
+	}
+
+	log.Infof("Upgraded %d block affinities belonging to this host.", numUpgraded)
+	if len(errs) > 0 {
+		return fmt.Errorf("failed to upgrade some block affinities: %v", errs)
+	}
+	return nil
 }
--- a/cni-plugin/pkg/ipamplugin/ipam_plugin.go
+++ b/cni-plugin/pkg/ipamplugin/ipam_plugin.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path"
 	"path/filepath"
 	"runtime"
 	"time"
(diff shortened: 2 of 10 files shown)


kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -230,6 +231,15 @@ func (c *loadBalancerController) onUpdate(update bapi.Update) {
...
 	t := time.NewTicker(timer)
+	for {
+		if err := c.ensureDatastoreUpgraded(); err != nil {
+			log.WithError(err).Error("Failed to upgrade load balancer's IPAM block affinities.  Unable to sync load balancer services.  Will retry.")
+			time.Sleep(10 * time.Second)
+			continue
+		}
+		break
+	}
+	log.Info("Successfully upgraded load balancer's IPAM block affinities.  Continuing to sync load balancer services.")
```
caseydavenport: Can we long this only when we actually make a change to any data?

libcalico-go/lib/ipam/ipam_upgrade_kdd_test.go
```diff
@@ -0,0 +1,134 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(bc.Clean()).NotTo(HaveOccurred())
+		ic = NewIPAMClient(bc, ipPools, &fakeReservations{})
+
+		// Build a raw CRD REST client using the same kubeconfig as the backend client.
+		cfg, _, err := k8s.CreateKubernetesClientset(&config.Spec)
+		Expect(err).NotTo(HaveOccurred())
+		rc, err := k8s.RawCRDClientV1(*cfg)
+		Expect(err).NotTo(HaveOccurred())
+		restClient = rc
```
caseydavenport: Is there a reason to create the client as `rc` and then assign it to `restClient` as opposed to just doing `restClient, err = k8s.RawCRDClientV1()` ?

libcalico-go/lib/backend/k8s/k8s.go
```diff
@@ -644,8 +644,10 @@ func buildK8SAdminPolicyClient(cfg *rest.Config) (*adminpolicyclient.PolicyV1alp
 	return adminpolicyclient.NewForConfig(cfg)
 }
 
-// buildCRDClientV1 builds a RESTClient configured to interact with Calico CustomResourceDefinitions
-func buildCRDClientV1(cfg rest.Config) (*rest.RESTClient, error) {
+// RawCRDClientV1 builds a RESTClient configured to interact with Calico
+// CustomResourceDefinitions.  Exposed for use in tests to allow creation of
```
caseydavenport: I have a slight preference for not expanding our use of the rest client directly unless really needed for the tests, just because the rest client is a bit clunkier to use than something like the controller-runtime client.

IIUC, what these tests want is a way to create CRDs without going through the libcalico-go client logic, which is doable using a controller-runtime client that has the calischeme added to it (it can also be used to interact with other k8s objects if needed, so means we can write tests that don't need multiple different clients depending on the API group).

We already use a controller-runtime client in some tests to interact with CRDs directly. e.g., here: https://github.com/projectcalico/calico/blob/cd887ad7d12c61255e12d75371809e9e58b4103d/libcalico-go/lib/backend/k8s/k8s_test.go#L381

All of that said, it's only a slight preference really.

libcalico-go/lib/backend/model/block_affinity.go
```diff
@@ -15,13 +15,19 @@
...
 import (
+	"crypto/sha3"
+	"encoding/base32"
 	"fmt"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 
+	v4 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
```
caseydavenport: v4? 🤔

caseydavenport: I've been using `apiv3` (or just `v3`) and `libapiv3` to distinguish in my CRD rework PR, as part of making `projectcalico/api/pkg/apis/projectcalico/v3` the canonical API definition.

libcalico-go/lib/ipam/ipam.go
```diff
@@ -583,7 +585,7 @@ func (s *blockAssignState) findOrClaimBlock(ctx context.Context, minFreeIps int)
 				return nil, false, err
 			}
 			logCtx := log.WithFields(log.Fields{string(s.affinityCfg.AffinityType): s.affinityCfg.Host, "subnet": subnet})
-			logCtx.Info("Found unclaimed block")
+			logCtx.Infof("Found unclaimed block in %v", time.Since(findStart))
```
caseydavenport, line 588: If only this was in a daemon and we could lace it up with prometheus metrics 😭

node/pkg/lifecycle/startup/startup.go
```diff
@@ -112,6 +131,23 @@ func Run() {
 		}
 	}
 
+	// Make sure that this host's BlockAffinity resources are upgraded to add
+	// labels for efficient lookup.  CNI plugin also does this on the first
+	// allocation after upgrade.  Doing it here too handles some corner cases,
+	// such as calico-node being downgraded, doing some allocations and then,
+	// upgrading again.
+	upgradeCtx, cancel := context.WithTimeout(ctx, 90*time.Second)
```
caseydavenport, line 139: How confident are we that 90s will be enough to do this for large clusters? Probably pretty confident, given we only need to list the affinities once and we both shouldn't hit many conflicts (just with CNI potentially) and shouldn't have to make too many Update() calls, since each node handles its own?

caseydavenport (commented):
Surprisingly easier to review than I feared! Looking good, mostly minor comments / questions.

node/pkg/lifecycle/startup/startup.go
fasaxc: Pretty confident:
* List should take 2-3s at most
* We have a default blocks-per-host limit of 20 so we're talking about 20 writes.
Order of magnitude should be same as time to do pre-fix IPAM :-)

libcalico-go/lib/backend/model/block_affinity.go
fasaxc: :laughing: Not conscious of writing this

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
fasaxc: Library already logs completion; I'll remove this one.

libcalico-go/lib/backend/k8s/k8s.go
fasaxc: Added a new utility package that makes a ctrl client with our schemes.  I had to disable protobuf; presumably we're not generating the right stuff for all our CRDs right now.

fasaxc:
There's something wrong; libcalico-go CI is taking 2x as long.  Need to figure that out...

fasaxc:
> There's something wrong; libcalico-go CI is taking 2x as long. Need to figure that out...

Ah, it was test leakage; the new benchmarks left 1,000 nodes lying around.

libcalico-go/lib/backend/k8s/rawcrdclient/raw_crd_client.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	libapiv3 "github.com/projectcalico/calico/libcalico-go/lib/apis/v3"
+)
+
+// NewAPIClient returns a new controller-runtime client configured to use the projectcalico.org/v3 API group.
```
caseydavenport: Does this client hit the `crd.projectcalico.org` objects?

fasaxc: That's what I'm using it for (vis that libapiv3 registration), but now I see that I copied in the API server v3 API too. I'll drop that since I'm not using it.

fasaxc: Oh, I should use the ones in `lib/backend/k8s/scheme`; that has all the CRDs.  Not sure why we have a partial copy in libcalico-go; I forget the history!

PR #10856, opened 2025-08-20
Make OpenStack resync interval configurable and disable-able

nelljerram:
In principle the ongoing resync isn't needed at all, so let's make it possible to switch it off, and do our DevStack (FV) testing like that, to confirm the principle.

On the other hand, if it is wanted, let's make the period configurable.  For now we continue to default to the same as before, i.e. 60s.

Plus a small variable naming improvement.

### DIFF
--- a/networking-calico/devstack/plugin.sh
+++ b/networking-calico/devstack/plugin.sh
@@ -2,79 +2,79 @@
 # Devstack plugin code for Calico
 # ===============================
 
-mode=$1				# stack, unstack or clean
-phase=$2			# pre-install, install, post-config or extra
+mode=$1                         # stack, unstack or clean
+phase=$2                        # pre-install, install, post-config or extra
 
 if [ "${Q_AGENT}" = calico-felix ]; then
 
     case $mode in
 
-	stack)
-	    # Called by stack.sh four times for different phases of
-	    # its run.
-	    echo Calico plugin: stack
+        stack)
+            # Called by stack.sh four times for different phases of
+            # its run.
+            echo Calico plugin: stack
 
-	    case $phase in
+            case $phase in
 
-		pre-install)
-		    # Called after system (OS) setup is complete and
-		    # before project source is installed.
-		    echo Calico plugin: pre-install
+                pre-install)
+                    # Called after system (OS) setup is complete and
+                    # before project source is installed.
+                    echo Calico plugin: pre-install
 
-		    # Add Calico master PPA as a package source.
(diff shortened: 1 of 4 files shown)


networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -992,8 +997,13 @@ def periodic_resync_thread(self, expected_epoch):
...
+
+                    if cfg.CONF.calico.resync_interval_secs == 0:
+                        # No continuing periodic resync.
+                        break
+
                     # Reschedule ourselves.
-                    eventlet.sleep(RESYNC_INTERVAL_SECS)
+                    eventlet.sleep(cfg.CONF.calico.resync_interval_secs)
                 else:
                     # Shorter sleep interval before we check if we've become
```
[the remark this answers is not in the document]
nelljerram: No, because we always want to perform the resync once, at start of day.

nelljerram:
Note, I haven't prepared the docs PR yet because I'm not yet clear what our big picture will be on whether resync is still needed at all - and if so, what other options there might be for it.  Hence I've left the `docs-pr-required` label to make sure of coming back to this.

networking-calico/devstack/plugin.sh
```diff
@@ -110,13 +110,20 @@ EOF
...
 		    # defined, set that as the value of the
 		    # etcd_compaction_min_revisions setting.
 		    if test -n "$CALICO_ETCD_COMPACTION_MIN_REVISIONS"; then
 			iniset $NEUTRON_CONF calico etcd_compaction_min_revisions $CALICO_ETCD_COMPACTION_MIN_REVISIONS
 		    fi
 
+		    # If CALICO_RESYNC_INTERVAL_SECS is defined, set that as the value of the
+		    # resync_interval_secs setting.
+		    if test -n "$CALICO_RESYNC_INTERVAL_SECS"; then
+			iniset $NEUTRON_CONF calico resync_interval_secs $CALICO_RESYNC_INTERVAL_SECS
```
fasaxc: Mixed tabs and spaces here leading to wacky formatting

fasaxc (changes requested):
One bit of dodgy formatting

networking-calico/devstack/plugin.sh
nelljerram: Thanks, I've untabified both .sh files.  Not sure why Emacs doesn't do that automatically - will try to make it do so.

PR #10857, opened 2025-08-20
CORE-11715 Added benchmark test for diachronic_flow::addFlow function

skoryk-oleksandr:
## Description

CORE-11715 Added benchmark test for **diachronic_flow::addFlow function**

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/goldmane/pkg/storage/add_flow_bench_test.go
+++ b/goldmane/pkg/storage/add_flow_bench_test.go
@@ -2,15 +2,17 @@ package storage_test
 
 import (
 	"fmt"
-	"github.com/projectcalico/calico/goldmane/pkg/storage"
-	"github.com/projectcalico/calico/goldmane/pkg/types"
-	"github.com/projectcalico/calico/goldmane/proto"
-	"github.com/sirupsen/logrus"
 	"os"
 	"runtime"
 	"strings"
 	"testing"
 	"unique"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/calico/goldmane/pkg/storage"
+	"github.com/projectcalico/calico/goldmane/pkg/types"
+	"github.com/projectcalico/calico/goldmane/proto"
 )
 
 var (


goldmane/pkg/storage/add_flow_bench_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return labelMap
+}
+
+func generateRandomLabelMap(r *rand.Rand, prefix string, count int) map[string]string {
+	r.Shuffle(len(keys), func(i, j int) { keys[i], keys[j] = keys[j], keys[i] })
+	r.Shuffle(len(values), func(i, j int) { values[i], values[j] = values[j], values[i] })
+
+	return generateLabelMap(prefix, count)
```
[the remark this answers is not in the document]
caseydavenport: I think this one is real - we shouldn't be modifying the global slices to avoid potential unexpected interactions in the future.

goldmane/pkg/storage/add_flow_bench_test.go
```diff
@@ -0,0 +1,125 @@
...
+			encodeLabels(generateRandomLabelMap(r, "src", r.Intn(9)+2)),
+			encodeLabels(generateRandomLabelMap(r, "dst", r.Intn(9)+2)),
+		},
+	}
+
+	b.ResetTimer()
+	runAddFlowBenchmark(b, labelPairs)
+}
+
+func runAddFlowBenchmark(b *testing.B, labelPairs []labelPair) {
```
caseydavenport: Is there a reason we need to split this part out into a separate function?

goldmane/pkg/storage/add_flow_bench_test.go
```diff
@@ -0,0 +1,125 @@
...
+			PacketsOut:              200,
+			BytesIn:                 10000,
+			BytesOut:                20000,
+			NumConnectionsStarted:   1,
+			NumConnectionsCompleted: 1,
+			NumConnectionsLive:      1,
+			SourceLabels:            pair.src,
+			DestLabels:              pair.dst,
+		}
+		df.AddFlow(flow, 1000, 2000)
```
caseydavenport: We probably want to spread the flows across different buckets instead of adding each one to the same bucket?

goldmane/pkg/storage/add_flow_bench_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	runtime.GC()
+	var m2 runtime.MemStats
+	runtime.ReadMemStats(&m2)
+
+	heapDelta := m2.HeapAlloc - m1.HeapAlloc
+
+	fmt.Printf("\n\n\nGo Benchmark Output:\n")
+	fmt.Printf("\nHeap Alloc: %.3f MB\n", float64(heapDelta)/1024/1024)
+	fmt.Printf("\n%-20s %-23s %-17s %-5s\n", "[Iterations]", "[ns/op]", "[B/op]", "[allocs/op]")
```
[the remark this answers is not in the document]
caseydavenport: Yeah, we should be using the logger instead of Printf as well.


Do you foresee running this benchmark as part of CI? Should we have some way to mark as pass/fail? Seems risky to make CI rely on a performance test, so I'd probably recommend against that - but if this test is only being run manually then perhaps some pass/fail heuristic would be useful?

goldmane/pkg/storage/add_flow_bench_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+type labelPair struct {
+	src unique.Handle[string]
+	dst unique.Handle[string]
+}
+
+func BenchmarkDiachronicFlow_AddFlow(b *testing.B) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+
+	labelPairs := []labelPair{
```
caseydavenport: I know it's just test code, but we should still have some code comments here. It's not obvious to me why two sets of src/dst labels were chosen here (as opposed to 1, 3, or more), for example.

goldmane/pkg/storage/add_flow_bench_test.go
```diff
@@ -0,0 +1,125 @@
...
+		{
+			encodeLabels(generateLabelMap("src", 10)),
+			encodeLabels(generateLabelMap("dst", 10)),
+		},
+		{
+			encodeLabels(generateRandomLabelMap(r, "src", r.Intn(9)+2)),
```
caseydavenport: Same here - what does `r.Intn(9)+2` indicate? Why was this value chosen to be random?

Similarly, why does one set use `generateRandomLabelMap` and the other doesn't? Do we need both functions or can we consolidate to a single one?

caseydavenport (commented):
A few initial thoughts / questions!

goldmane/pkg/storage/add_flow_bench_test.go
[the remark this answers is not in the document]
skoryk-oleksandr: fixed

goldmane/pkg/storage/add_flow_bench_test.go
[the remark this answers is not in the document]
skoryk-oleksandr: Within the test code, we can only programmatically measure and validate heap allocation.
Other benchmark metrics such as ns/op or B/op are reported by the Go testing framework, but are not accessible within the test logic itself. 
At the end of the test, we print the expected threshold values for these metrics
to provide a reference baseline for others reviewing the results. I would not run this test as part of the CI.

goldmane/pkg/storage/add_flow_bench_test.go
skoryk-oleksandr: fixed

goldmane/pkg/storage/add_flow_bench_test.go
skoryk-oleksandr: Done

goldmane/pkg/storage/add_flow_bench_test.go
skoryk-oleksandr: Done

goldmane/pkg/storage/add_flow_bench_test.go
skoryk-oleksandr: Removed this

goldmane/pkg/storage/add_flow_bench_test.go
```diff
@@ -0,0 +1,145 @@
...
+	)
+
+	df := storage.NewDiachronicFlow(flowKey, 1)
+	return df
+}
+
+func BenchmarkDiachronicFlow_AddFlow(b *testing.B) {
+	loglevel := logrus.GetLevel()
+	logrus.SetLevel(logrus.WarnLevel)
+	defer logrus.SetLevel(loglevel)
```
caseydavenport: Worth also calling the utility to direct log output to logrus:

https://github.com/projectcalico/calico/blob/ceb00cfeb917e81dd563fada75a48e0fb5cb813d/goldmane/pkg/storage/diachronic_flow_test.go#L33

goldmane/pkg/storage/add_flow_bench_test.go
```diff
@@ -0,0 +1,145 @@
...
+		// Advance the time window every 100 iterations to simulate multiple non-overlapping time buckets.
+		// This ensures each 100 flows belong to the same (start, end) interval.
+		inc := int64((i / 100) * 100)
+		flow := flows[i%len(flows)]
+		df.AddFlow(flow, start+inc, end+inc)
+	}
+	b.StopTimer()
+
+	runtime.GC()
+	time.Sleep(time.Second)
```
caseydavenport: This seems like a magic sleep - what are we waiting for here?

goldmane/pkg/storage/add_flow_bench_test.go
```diff
@@ -0,0 +1,145 @@
...
+	defer logrus.SetLevel(loglevel)
+
+	df := getDiachronicFlow()
+	start := int64(1000)
+	end := int64(2000)
+
+	runtime.GC()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
```
caseydavenport: https://pkg.go.dev/testing#B.Loop

Looks like the latest recommendation is to use `b.Loop()` instead of `b.N`

This would avoid us needing to explicitly call `ResetTimer` and `StopTimer`

caseydavenport (commented):
Looking good - a few small comments / questions.

caseydavenport:
@skoryk-oleksandr also worth adding a `benchmarks` target to the Makefile to make it easy to run the benchmark suite.

goldmane/pkg/storage/add_flow_bench_test.go
skoryk-oleksandr: Sounds good. Used `b.Loop()`

goldmane/pkg/storage/add_flow_bench_test.go
skoryk-oleksandr: Leftover from the previous experiments. Removed it

goldmane/pkg/storage/add_flow_bench_test.go
skoryk-oleksandr: Done :+1:

PR #10876, opened 2025-08-25
Add slaac support for calico-dhcp-agent 

kristiangronas:
## Description

Currently calico-dhcp-agent only supports dhcp6_stateful, but this can introduce a failure mode where leases get stuck after for example an openstack rebuild. SLAAC allows the machine to assign it's own IP, based on the mac address, which is already decided by neutron and passed into the VM

For ipv4 it already works similarly, since dnsmasq can simply bind the leases to ipv4 mac addresses

From what i can tell you still need a normal subnet to make nova happy, but the main goal of this PR is having a SLAAC address with higher reliability

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/agent/dhcp_agent.py
+++ b/networking-calico/networking_calico/agent/dhcp_agent.py
@@ -829,6 +829,9 @@ class SubnetWatcher(etcdutils.EtcdWatcher):
         if ip_version == 6:
             subnet["ipv6_address_mode"] = constants.DHCPV6_STATEFUL
             subnet["ipv6_ra_mode"] = constants.DHCPV6_STATEFUL
+            if data.get("ipv6_address_mode") == constants.IPV6_SLAAC && data.get("ipv6_ra_mode") == constants.IPV6_SLAAC:
+                subnet["ipv6_address_mode"] = constants.IPV6_SLAAC
+                subnet["ipv6_ra_mode"] = constants.IPV6_SLAAC
 
         return dhcp.DictModel(subnet)
 
--- a/networking-calico/networking_calico/agent/linux/dhcp.py
+++ b/networking-calico/networking_calico/agent/linux/dhcp.py
@@ -99,6 +99,8 @@ class DnsmasqRouted(dhcp.Dnsmasq):
                     and not ra_mode
                 ):
                     mode = "static"
+                if (addr_mode == constants.IPV6_SLAAC && ra_mode == constants.IPV6_SLAAC):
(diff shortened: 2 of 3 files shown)


lwr20:
Tagging @nelljerram

nelljerram:
Also @kristiangronas would you like to suggest a brief release note?  I don't think we have docs for this kind of detail, but if we wanted to the place for that would be somewhere under https://docs.tigera.io/calico/latest/networking/openstack/, for which the source is at https://github.com/tigera/docs/tree/main/calico/networking/openstack.  WDYT - I think I'm happy to leave this undocumented, but also happy to consider a doc update if you'd like to propose it.

kristiangronas:
Added a test, i didn't totally understand what was happening with the order of the flags, but it shouldn't really matter

Maybe the doc could go in https://docs.tigera.io/calico/latest/networking/openstack/ipv6 ? It does mention dhcp6-stateful already

As for the release note maybe something like: `dhcp-agent: allow using SLAAC for more reliable ipv6 addresses`?

nelljerram (approved):
Thanks @kristiangronas , the test looks good to me.  For the release note we have to zoom out a bit - considering that Calico is for both Kubernetes and OpenStack, so I've put `OpenStack: support IPv6 subnets using SLAAC`, does that sound OK?

Your suggested doc location sounds good - would you like to propose that change?

nelljerram:
@kristiangronas We recently started enforcing 'black' formatting for the Calico/OpenStack Python code.  You need to `make -C networking-calico fmtpy` and commit the changes that that makes.  Probably also a good idea to check `make -C networking-calico flake8` and `make -C networking-calico tox-caracal`.

kristiangronas:
Ah, i thought it would go in an openstack section, but not sure if you do that, sounds good!

I was running tox -e black which was doing some weird changes, so i thought it wasn't being enforced yet, worked much better with the version in the makefile. I was already running the tests with makefile tox-caracel (and yoga), and it worked great. Thank you for the attention in this area lately, it made the experience a lot smoother

nelljerram:
@kristiangronas We're OK to merge this now, is that OK with you?

kristiangronas:
Sounds good!

nelljerram:
@kristiangronas Many thanks for this work.  If you decide to post a docs PR, please do tag me on that one also.

PR #10881, opened 2025-08-25
bpf: support for adding DSCP for egress traffic

mazdakn:
## Description

Add support for setting DSCP for egress traffic leaving cluster. 
This is the follow up on the main PRs:
- Workload endpoints: https://github.com/projectcalico/calico/pull/10718
- Host endpoints: https://github.com/projectcalico/calico/pull/10825

The follow up PR includes a change to only set DSCP for traffic leaving cluster.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/bpf/ut/qos_test.go
```diff
@@ -95,3 +95,134 @@ func TestQoSPacketRate(t *testing.T) {
...
+	rtVal := routes.NewValueWithIfIndex(routes.FlagsLocalWorkload|routes.FlagInIPAMPool, ifIndex).AsBytes()
+	err = rtMap.Update(rtKey, rtVal)
+	Expect(err).NotTo(HaveOccurred())
+	rtKey = routes.NewKey(dstV4CIDR).AsBytes()
+	rtVal = routes.NewValueWithIfIndex(routes.FlagsRemoteWorkload|routes.FlagInIPAMPool, ifIndex).AsBytes()
+	err = rtMap.Update(rtKey, rtVal)
+	Expect(err).NotTo(HaveOccurred())
+	defer resetRTMap(rtMap)
+
+	// Ingress, allow first packet, drop second (because of 1/sec limit)
```
tomastigera: copy-paste?

mazdakn: Removed.

felix/bpf-gpl/qos.h
```diff
@@ -111,4 +112,44 @@ static CALI_BPF_INLINE int enforce_packet_rate_qos(struct cali_tc_ctx *ctx)
...
+	// traffic class are stored in IPv6 priority field (4 bits), and the 4 least significant bits of it
+	// are stored in the 4 most significant bits of IPv6 flow_lbl[0] field. We must not change ECN bits here.
+	ip_hdr(ctx)->priority = (__u8) (dscp >> 2);
+	ip_hdr(ctx)->flow_lbl[0] = (__u8) (ip_hdr(ctx)->flow_lbl[0] & 0x3f) | (dscp << 6);
+#else
+	// In IPv4, DSCP (6bits) is located at the most significant bits of IPv4 TOS field.
+	// The 2 least significant bits are assigned to ECN and must not be touched.
+	ip_hdr(ctx)->tos = (__u8) ((ip_hdr(ctx)->tos & 0x03) | (dscp << 2));
+
+	__wsum ip_csum = bpf_csum_diff(0, 0, (__u32 *)ctx->ip_header, sizeof(struct iphdr), 0);
```
tomastigera, line 136: don't you need to set the csum field to 0 first if you want to compute csum of the entire header? `This includes the Header Checksum field itself, which is set to zero during computation. `

tomastigera: I think you should set csum to 0 if you want to compute the csum of the entire ip header `the checksum is computed as the 16-bit ones' complement of the ones' complement sum of all 16-bit words in the header. This includes the Header Checksum field itself, which is set to zero during computation.`

felix/bpf-gpl/qos.h
```diff
@@ -111,4 +112,44 @@ static CALI_BPF_INLINE int enforce_packet_rate_qos(struct cali_tc_ctx *ctx)
 	return TC_ACT_SHOT;
 }
 
+static CALI_BPF_INLINE int set_dscp(struct cali_tc_ctx *ctx)
+{
+#if (CALI_F_FROM_WEP || CALI_F_TO_HEP)
+	// TODO (mazdak): set DSCP only if traffic is leaving cluster
+	if (EGRESS_DSCP < 0) {
+		return TC_ACT_UNSPEC;
+	}
```
tomastigera: Would it make sense to take it out from this function? Could -1 be technically a valid value? (guess not since all valid values are 6bits only). Perhaps a flag that this field is set?

felix/bpf/ut/bpf_prog_test.go
```diff
@@ -1245,6 +1251,12 @@ func withEgressQoSPacketRate() testOption {
 	}
 }
 
+func withEgressDSCP(value int8) testOption {
+	return func(o *testOpts) {
+		o.dscp = &value
```
tomastigera: if -1 is invalid, you could have used -1 as invalid as well instead of using a pointer

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -2853,6 +2864,17 @@ func (d *bpfEndpointManagerDataplane) attachDataIfaceProgram(
 		return nil, err
 	}
 
+	ap.DSCP = -1
```
tomastigera, line 2867: ditto

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -2434,6 +2434,17 @@ func (m *bpfEndpointManager) doApplyPolicy(ifaceName string) (bpfInterfaceState,
 		}
 	}
 
+	ap.DSCP = -1
```
tomastigera, line 2437: you set the dscp to -1 but then you deny packets if it is set to -1 🤔

felix/bpf-gpl/qos.h
```diff
@@ -111,4 +112,44 @@ static CALI_BPF_INLINE int enforce_packet_rate_qos(struct cali_tc_ctx *ctx)
 	return TC_ACT_SHOT;
 }
 
+static CALI_BPF_INLINE int set_dscp(struct cali_tc_ctx *ctx)
+{
+#if (CALI_F_FROM_WEP || CALI_F_TO_HEP)
```
tomastigera: would it make sense to take this out? See my suggestion

felix/bpf-gpl/tc.c
```diff
@@ -1327,6 +1327,9 @@ int calico_tc_skb_accepted_entrypoint(struct __sk_buff *skb)
 		deny_reason(ctx, CALI_REASON_DROPPED_BY_QOS);
 		goto deny;
 	}
+	if (set_dscp(ctx) == TC_ACT_SHOT) {
```
```suggestion
	if ((CALI_F_FROM_WEP || CALI_F_TO_HEP) && EGRESS_DSCP >= 0 && !set_dscp(ctx)) {
```

you could make the set_dscp return bool. I'd suggest name it `qos_set_dscp` will suggest to rename the rate functions to qos_* too

tomastigera (commented):
nothing fundamental, mostly nits

felix/bpf-gpl/qos.h
mazdakn: Fixed.

felix/bpf/ut/bpf_prog_test.go
mazdakn: I added it this way only to consider the default value of type. Since default value of int, 0, is a valid DSCP value. 
I managed to find the initialization part, so now type is int.

felix/bpf-gpl/tc.c
mazdakn: Fixed.

felix/dataplane/linux/bpf_ep_mgr.go
mazdakn: It wasn't dropping packets, but clearly the return value was confusing. Changed the return val to bool.

felix/bpf-gpl/qos.h
mazdakn: Adding a flag to check if this value is set is probably an overkill. we can check `EGRESS_DSCP >= 0` in the caller part.

tomastigera (approved):
LGTM ... it would be great in the follow up PR to add a FV test that would set the DSCP. There are tests with external client (mainly for nodeports) and that could verify various scenarios. And we can use tcpdump to verify that the value is set when expected.

mazdakn:
@tomastigera we have some FV for this: https://github.com/projectcalico/calico/blob/bd405757167dd5bf1e5e14035b9b17ae4f46be02/felix/fv/dscp_test.go#L193
But I'm going to add more for a couple of more scenarios.

Issue #10502, opened 2025-06-03
Using ip pool annotations on namespaces and node selectors within the ip pools


bshekar2191:
When IP Pool annotation is used in namespaces using "[cni.projectcalico.org/ipv6pools"='["pool-amer-west", "pool-amer-east"]], and the pool-amer-west ippool has a nodeSelector set to match 'AMER-WEST' nodes and pool-amer-east ippool has a nodeSelector set to match 'AMER-EAST nodes', Calico/ K8s is allocating IPs from either pools in both regions. 
No vulnerability

## Expected Behavior
I was hoping that Calico understands that this namespace can use both IP Pools based on the annotation and then look at the node selector on each IP Pool and use that pool accordingly in Amer East or Amer West

## Current Behavior
It is using the same pool (pool-amer-west) in all the nodes for that namespace

## Possible Solution
Perhaps allow us to use both namespace annotations and ip pool node selectors together to finally choose which ippool to use

## Steps to Reproduce (for bugs)
This is perhaps not a bug

## Context
Here's what I'm trying to achieve: 
I have two regions  : AMER-EAST and AMER-WEST and they're using separate IP blocks for addressing (10.10.0.0/16 for AMER-EAST and 10.20.0.0/16 for AMER-WEST )
In addition, I want to do IP Pool per namespace for security.
Example: NS1 should use IP Pool 1 (CIDR 10.10.0.0/24 with node selector AMER-EAST ) in AMER-EAST nodes and IP Pool 2 (CIDR 10.20.0.0/24 with node selector AMER-WEST) in AMER-WEST nodes
Likewise NS2  should use IP Pool 3 (CIDR 10.10.1.0/24 with node selector AMER-EAST ) in AMER-EAST nodes and IP Pool 4 (CIDR 10.20.1.0/24 with node selector AMER-WEST) in AMER-WEST nodes
Is that possible with calico? Or do you have any other suggestions/ creative solutions?

## Your Environment

* Calico version : Calico Open Source - Latest
* Calico dataplane (iptables, windows etc.)
* Orchestrator version (e.g. kubernetes, mesos, rkt): Kubernetes 1.32
* Operating System and version: Linux
* Link to your project (optional):

stevegaossou:
Hi @bshekar2191, as you alluded to in the description of this issue, this behaviour is not a bug.  

[...]

PR #10888, opened 2025-08-27
support namespace selector field in ip pool

PersistentJZH:
## Description

introduce a `namespaceSelector` field to IP pools, enabling selective IP allocation based on namespace labels. 

This feature provides granular control over which namespaces can use specific IP pools, selector behavior: 

**Namespace Selector**: When configured, only namespaces matching the selector criteria can allocate IPs from the pool

**Combined Selectors**: When both namespaceSelector and nodeSelector are present, they operate as an "AND" condition - both namespace and node must match

**Priority Hierarchy**: When annotations and selectors coexist, annotations take precedence and selectors are disregarded

**Example Use Cases**
```
kind: IPPool
metadata:
  name: prod-pool
spec:
  cidr: 10.0.0.0/24
  namespaceSelector: "environment == 'production'"
  nodeSelector: "zone == 'us-west'"
```

```
kind: IPPool
metadata:
  name: dev-pool
spec:
  cidr: 10.1.0.0/24
  namespaceSelector: "environment == 'development'"
```

## Related issues/PRs
https://github.com/projectcalico/calico/issues/10502

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Enhanced IP Pool Allocation by introduce namespaceSelector Field
```

## Reminder for the reviewer

Other optional labels:


libcalico-go/lib/clientv3/namespace.go
```diff
@@ -0,0 +1,120 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/projectcalico/calico/libcalico-go/lib/backend/k8s"
+	"github.com/projectcalico/calico/libcalico-go/lib/options"
+)
+
+// NamespaceInterface provides methods to work with Kubernetes Namespace resources.
+// Used for namespaceSelector functionality in IPAM.
+type NamespaceInterface interface {
```
caseydavenport: I don't think we need to introduce a new `clientv3` interface to interact with Namespaces (they aren't part of the Calico v3 API!) 

We can simply use a Kubernetes corev1 clientset for this, or a controller-runtime client.

cni-plugin/pkg/ipamplugin/ipam_plugin.go
```diff
@@ -262,7 +283,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 			MaxBlocksPerHost: maxBlocks,
 			Attrs:            attrs,
 			IntendedUse:      v3.IPPoolAllowedUseWorkload,
+			Namespace:        namespace,
```
caseydavenport: Looks like we're passing both the Namespace name, as well as the Namespace labels here - might make sense to just pass the entire `corev1.Namespace{}` object for simplicity?

libcalico-go/lib/ipam/ipam.go
```diff
@@ -297,34 +297,50 @@ func (c ipamClient) determinePools(ctx context.Context, requestedPoolNets []net.
...
+		if pool.Spec.NamespaceSelector != "" {
+			var namespaceMatches bool
+			namespaceMatches, err = SelectsNamespace(pool, namespace, namespaceLabels)
+			if err != nil {
+				log.WithError(err).WithField("pool", pool).Error("failed to determine if namespace matches pool")
+				return
+			}
+			if !namespaceMatches {
+				// Do not consider pool enabled if the namespaceSelector doesn't match the namespace's labels.
+				log.Debugf("IP pool does not match this namespace: %s", pool.Name)
```
caseydavenport: Would be good to log out the Namespace as well here. 

We can use context logging:

```suggestion
				log.WithField("namespace", namespace).Debugf("IP pool does not match this namespace: %s", pool.Name)
```

libcalico-go/lib/ipam/ipam_block_reader_writer_test.go
```diff
@@ -1102,7 +1102,7 @@ var _ = testutils.E2eDatastoreDescribe("IPAM affine block allocation tests", tes
 				blockReaderWriter: rw,
 				reservations:      &fakeReservations{},
 			}
-			ia, err := ic.autoAssign(ctx, 1, nil, nil, nil, 4, host, 0, rsvdAttr, v3.IPPoolAllowedUseTunnel /* for variety */)
+			ia, err := ic.autoAssign(ctx, 1, nil, nil, nil, 4, host, 0, rsvdAttr, v3.IPPoolAllowedUseTunnel /* for variety */, "", map[string]string{})
```
caseydavenport:
```suggestion
			ia, err := ic.autoAssign(ctx, 1, nil, nil, nil, 4, host, 0, rsvdAttr, v3.IPPoolAllowedUseTunnel, "", map[string]string{})
```

libcalico-go/lib/ipam/pools.go
```diff
@@ -45,3 +45,21 @@ func SelectsNode(pool v3.IPPool, n libapiv3.Node) (bool, error) {
 	// Return whether or not the selector matches.
 	return sel.Evaluate(n.Labels), nil
 }
+
+// SelectsNamespace determines whether or not the IPPool's namespaceSelector
+// matches the labels on the given namespace.
+func SelectsNamespace(pool v3.IPPool, namespaceName string, namespaceLabels map[string]string) (bool, error) {
```
caseydavenport: The `namespaceName` argument seems to be unused here. Probably can remove it?

api/pkg/apis/projectcalico/v3/ippool.go
```diff
@@ -74,6 +74,10 @@ type IPPoolSpec struct {
 	// Allows IPPool to allocate for a specific node by label selector.
 	NodeSelector string `json:"nodeSelector,omitempty" validate:"omitempty,selector"`
 
+	// Allows IPPool to allocate for a specific namespace by label selector.
+	// If specified, both namespaceSelector and nodeSelector must match for the pool to be used.
+	NamespaceSelector string `json:"namespaceSelector,omitempty" validate:"omitempty,selector"`
```
caseydavenport, line 79: We should enhance the validation code to make sure this selector is valid.

https://github.com/projectcalico/calico/blob/291c8f5138429c8851bcbece1d5712fc1921567c/libcalico-go/lib/validator/v3/validator.go#L1127-L1247

It actually looks like we're missing validation logic for `NodeSelector` as well!

cni-plugin/pkg/ipamplugin/ipam_plugin.go
```diff
@@ -252,6 +253,26 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+				namespaceLabels = map[string]string{}
+			} else {
+				namespaceLabels = labels
+				logger.Debugf("Got namespace labels for %s: %v", namespace, namespaceLabels)
+			}
+		} else {
+			logger.Debugf("Using namespace %s without labels (no K8s config or empty namespace)", namespace)
+		}
+
 		assignArgs := ipam.AutoAssignArgs{
```
caseydavenport, line 274: We'll want to look for other places that call the IPAM assignment code and make sure they are given the Namespace information as well.

For example, the Service LoadBalancer controller likely wants this: https://github.com/projectcalico/calico/blob/291c8f5138429c8851bcbece1d5712fc1921567c/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go#L674-L681

CC @MichalFupso

caseydavenport: https://github.com/projectcalico/calico/blob/291c8f5138429c8851bcbece1d5712fc1921567c/node/pkg/allocateip/allocateip.go#L477-L482

The tunnel allocation code probably doesn't care about Namespace labels - the tunnel IPs don't belong to a Namespaced resource. 

It does mean that any IP pool with a NamespaceSelector specified fundamentally cannot be used to assign tunnel IPs. We probably want validation logic that rejects IP pools that have `allowedUse` of `Tunnel` and that also have a NamespaceSelector specified.

cni-plugin/pkg/ipamplugin/ipam_plugin.go
```diff
@@ -474,3 +498,22 @@ func cmdDel(args *skel.CmdArgs) error {
...
 }
+
+// getNamespaceLabels retrieves namespace labels using Calico client's Namespaces interface
+func getNamespaceLabels(calicoClient client.Interface, namespace string) (map[string]string, error) {
+	if namespace == "" {
+		return map[string]string{}, nil
+	}
+
+	// Use Calico client's Namespaces interface
+	ns, err := calicoClient.Namespaces().Get(context.Background(), namespace, options.GetOptions{})
```
caseydavenport: Yep - this should just be done with a standard k8s clientset. No need to plumb it through our v3 infra.

caseydavenport (changes requested):
Thanks @PersistentJZH ! 

I've taken a first look and added some thoughts. Let me know what you think!

PersistentJZH:
thanks @caseydavenport ‘s suggestions, I have resolved all the comments above. please help review it again, thank you

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -1244,6 +1244,100 @@ func validateIPPoolSpec(structLevel validator.StructLevel) {
...
+		// - k == 'v' - exact match
+		// - k != 'v' - not match
+		// - has(k) - existence check
+		// - k in { 'v1', 'v2' } - set match
+		// - k not in { 'v1', 'v2' } - set not match
+		// - k contains 's' - substring contains
+		// - k starts with 's' - starts with
+		// - k ends with 's' - ends with
+		// - Logical combinations with &&, ||, !, ()
+		if pool.NodeSelector != "all()" &&
```
caseydavenport: I think the `selector.Validate` call above should handle all of these cases, right?

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -1244,6 +1244,100 @@ func validateIPPoolSpec(structLevel validator.StructLevel) {
...
+		// - k == 'v' - exact match
+		// - k != 'v' - not match
+		// - has(k) - existence check
+		// - k in { 'v1', 'v2' } - set match
+		// - k not in { 'v1', 'v2' } - set not match
+		// - k contains 's' - substring contains
+		// - k starts with 's' - starts with
+		// - k ends with 's' - ends with
+		// - Logical combinations with &&, ||, !, ()
+		if !strings.Contains(pool.NamespaceSelector, "==") &&
```
caseydavenport: Same here

cni-plugin/pkg/ipamplugin/ipam_plugin.go
```diff
@@ -252,6 +255,26 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+		namespace := epIDs.Namespace
+		var namespaceObj *corev1.Namespace
+
+		// Only attempt to fetch namespace if we have Kubernetes configuration and a valid namespace
+		if (conf.Kubernetes.Kubeconfig != "" || conf.Policy.PolicyType == "k8s") && namespace != "" {
+			logger.Debugf("Getting namespace for: %s", namespace)
+
+			ns, err := getNamespace(conf, namespace, logger)
+			if err != nil {
+				logger.WithError(err).Warnf("Failed to get namespace for %s, using nil", namespace)
```
caseydavenport: Hm, any reason to not simply return an error here?

If we use a nil namespace, it is possible that we'll end up using the wrong IP pool to assign an IP address to this Pod. For example, if there is a pool with a namespace selector and one without?

cni-plugin/pkg/ipamplugin/ipam_plugin.go
```diff
@@ -252,6 +255,26 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 		}
+		// Get namespace information for namespaceSelector support
+		namespace := epIDs.Namespace
+		var namespaceObj *corev1.Namespace
+
+		// Only attempt to fetch namespace if we have Kubernetes configuration and a valid namespace
+		if (conf.Kubernetes.Kubeconfig != "" || conf.Policy.PolicyType == "k8s") && namespace != "" {
+			logger.Debugf("Getting namespace for: %s", namespace)
+
+			ns, err := getNamespace(conf, namespace, logger)
```
caseydavenport:
```suggestion
			namespaceObj, err = getNamespace(conf, namespace, logger)
```

You can assign directly to the `namespaceObj` variable and it will avoid the need for the `else` block below.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -671,13 +671,25 @@ func (c *loadBalancerController) assignIP(svc *v1.Service) ([]string, error) {
 		return nil, nil
 	}
 
+	// Get the namespace object for namespaceSelector support
+	var namespaceObj *v1.Namespace
+	if svc.Namespace != "" {
+		ns, err := c.clientSet.CoreV1().Namespaces().Get(context.Background(), svc.Namespace, metav1.GetOptions{})
```
caseydavenport:
```suggestion
		namespaceObj, err = c.clientSet.CoreV1().Namespaces().Get(context.Background(), svc.Namespace, metav1.GetOptions{})
```

Same here - we can assign directly.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -671,13 +671,25 @@ func (c *loadBalancerController) assignIP(svc *v1.Service) ([]string, error) {
 		return nil, nil
 	}
 
+	// Get the namespace object for namespaceSelector support
+	var namespaceObj *v1.Namespace
+	if svc.Namespace != "" {
```
caseydavenport: I think Services must always have a Namespace?

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -671,13 +671,25 @@ func (c *loadBalancerController) assignIP(svc *v1.Service) ([]string, error) {
...
+		ns, err := c.clientSet.CoreV1().Namespaces().Get(context.Background(), svc.Namespace, metav1.GetOptions{})
+		if err != nil {
+			log.WithError(err).WithField("namespace", svc.Namespace).Warn("Failed to get namespace for LoadBalancer IP assignment, proceeding without namespace selector support")
```
caseydavenport: I think we'll want to just return an error here - if the Namespace isn't accessible, we shouldn't try to assign an IP.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
caseydavenport: Ideally, we'd use an `Informer` to establish a watch here rather than performing a Get request for every Service .

I'm happy for this to be in a separate PR if you'd like, though.

libcalico-go/lib/ipam/ipam.go
```diff
@@ -297,34 +299,48 @@ func (c ipamClient) determinePools(ctx context.Context, requestedPoolNets []net.
...
 	for _, pool := range enabledPools {
 		if len(requestedPoolNets) == 0 && *pool.Spec.AssignmentMode != v3.Automatic {
 			continue
 		}
-		var matches bool
-		matches, err = SelectsNode(pool, node)
+
+		// Check node selector
+		var nodeMatches bool
+		nodeMatches, err = SelectsNode(pool, node)
```
caseydavenport:
```suggestion
		nodeMatches, err := SelectsNode(pool, node)
```

libcalico-go/lib/ipam/ipam.go
```diff
@@ -297,34 +299,48 @@ func (c ipamClient) determinePools(ctx context.Context, requestedPoolNets []net.
...
+		if !nodeMatches {
 			// Do not consider pool enabled if the nodeSelector doesn't match the node's labels.
 			log.Debugf("IP pool does not match this node: %s", pool.Name)
 			continue
 		}
-		log.Debugf("IP pool matches this node: %s", pool.Name)
+
+		// Check namespace selector
+		var namespaceMatches bool
+		namespaceMatches, err = SelectsNamespace(pool, namespace)
```
caseydavenport:
```suggestion
		namespaceMatches, err := SelectsNamespace(pool, namespace)
```

libcalico-go/lib/ipam/ipam.go
```diff
@@ -370,7 +386,35 @@ func (c ipamClient) prepareAffinityBlocksForHost(ctx context.Context, requestedP
 	}
 
 	// Determine the correct set of IP pools to use for this request.
-	poolsSelectingNode, allPools, err := c.determinePools(ctx, requestedPools, version, *v3n, maxPrefixLen)
+	// For LoadBalancer use, we don't have namespace context, so use empty values
```
caseydavenport:
```suggestion
	// For some IPs (e.g., tunnel addresses), we don't have namespace context, so use empty values
```

caseydavenport: I think it would be cleaner to simply pass `nil` and to have `determinePools` handle that?

Looks like `SelectsNamespace` already handles a nil input, so I think we can remove this whole section and simply use `namespace` instead of `namespaceObj` here?

caseydavenport (changes requested):
Thanks @PersistentJZH - I've got another round of thoughts, but overall looking pretty good.

cni-plugin/pkg/ipamplugin/ipam_plugin.go
PersistentJZH: yes, return an error directly is the best approach here, as "using the wrong IP pool" has a greater negative impact.

cni-plugin/pkg/ipamplugin/ipam_plugin.go
PersistentJZH: good suggestion

libcalico-go/lib/validator/v3/validator.go
PersistentJZH:
```
	// Allows IPPool to allocate for a specific node by label selector.
	NodeSelector string `json:"nodeSelector,omitempty" validate:"omitempty,selector"`

	// Allows IPPool to allocate for a specific namespace by label selector.
	// If specified, both namespaceSelector and nodeSelector must match for the pool to be used.
	NamespaceSelector string `json:"namespaceSelector,omitempty" validate:"omitempty,selector"`
```

Yes, one more question. The validate tag here will also trigger the selector.Validate check, right? So we don’t need another selector.Validate in validateIPPoolSpec()? In addition to validating global(), do we need to validate other cases?

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
PersistentJZH: Sure, I'm happy to use `informer` to  optimzie the logic here in another PR.

libcalico-go/lib/validator/v3/validator.go
caseydavenport: Good question! I think you're right - it should pick up on the tag and validate the field. 

A unit test should be able to verify it pretty quickly!

PersistentJZH:
thanks for @caseydavenport code review, if have no other concerns, could you please help trigger the e2e test? thank you!

caseydavenport (approved):

maksimsamt:
@caseydavenport, @PersistentJZH,

```yaml
kind: IPPool
metadata:
  name: some-pool
spec:
  cidr: 10.1.0.0/24
  namespaceSelector: "kubernetes.io/metadata.name == 'some-namespace'"
```

Could you please explain how works this example.
As far as I understand, IPs allocation from this IPPool `some-pool` only can get resources from namespace `some-namespace`, other namespaces are rejected?
In my case by this example IPs allocation can get from any namespace, no limitation.
Or is there some other logic here?

PR #10895, opened 2025-08-28
bpf: deprecate map repinning

coutinhop:
Remove map repinning code and deprecate the felix 'DebugBPFMapRepinEnabled' config option.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
The felix 'DebugBPFMapRepinEnabled' config option is now deprecated and maps are no longer repinned by name.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/bpf_test.go
+++ b/felix/fv/bpf_test.go
@@ -968,32 +968,32 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 						options.ExtraEnvVars["FELIX_DebugBPFMapRepinEnabled"] = "true"
 					})
 
-					It("should repin maps", func() {
+					It("should not repin maps (option is deprecated)", func() {
 						// Wait for the first felix to create its maps.
 						mapID := mustGetMapIDByPath(tc.Felixes[0], mapPath)
 
 						// Now, start a completely independent felix, which will get its own bpffs.  It should re-pin the
 						// maps, picking up the ones from the first felix.
-						tc, _ := infrastructure.StartSingleNodeTopology(options, infra)
-						defer tc.Stop()
+						tc2, _ := infrastructure.StartSingleNodeTopology(options, infra)
+						defer tc2.Stop()
 
-						secondMapID := mustGetMapIDByPath(tc.Felixes[0], mapPath)
+						secondMapID := mustGetMapIDByPath(tc2.Felixes[0], mapPath)
 						Expect(mapID).NotTo(BeNumerically("==", 0))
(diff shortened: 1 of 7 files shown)


felix/fv/bpf_test.go
```diff
@@ -968,32 +968,32 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 						options.ExtraEnvVars["FELIX_DebugBPFMapRepinEnabled"] = "true"
 					})
 
-					It("should repin maps", func() {
+					It("should not repin maps (option is deprecated)", func() {
```
tomastigera: We could completely remove `Describe("with map repinning enabled", func() {` right?

felix/fv/bpf_test.go
```diff
@@ -968,32 +968,32 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						secondMapID := mustGetMapIDByPath(tc2.Felixes[0], mapPath)
 						Expect(mapID).NotTo(BeNumerically("==", 0))
-						Expect(mapID).To(BeNumerically("==", secondMapID))
+						Expect(mapID).NotTo(BeNumerically("==", secondMapID))
 					})
 				})
 
 				Describe("with map repinning disabled", func() {
-					It("should repin maps", func() {
+					It("should not repin maps", func() {
```
tomastigera: there is no longer any point in having this test, right?

felix/fv/bpf_test.go
coutinhop: yes, I wasn't sure if we wanted to guard against possibly the function being brought back... also not sure if we could remove the felix config altogether, I've left it with a `[deprecated]` tag and it now does nothing. let me know if we should just remove it

felix/fv/bpf_test.go
```diff
@@ -963,40 +963,18 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
-						// Now, start a completely independent felix, which will get its own bpffs.  It should make its own
-						// maps.
-						tc2, _ := infrastructure.StartSingleNodeTopology(options, infra)
-						defer tc2.Stop()
-
-						secondMapID := mustGetMapIDByPath(tc2.Felixes[0], mapPath)
-						Expect(mapID).NotTo(BeNumerically("==", 0))
-						Expect(mapID).NotTo(BeNumerically("==", secondMapID))
-					})
+				It("should not repin maps", func() {
```
tomastigera: what is the value of this test. We do not do something and we test that it does not happen? :thinking: I think this test is real legacy

PR #10904, opened 2025-09-01
ConsistentHash/Maglev

aaaaaaaalex:
## Description

Implement the consistent hashing of 5-tuples for service traffic in eBPF dataplane, in order to alleviate downtime during loadbalancer failover.

## Todos

- [x] Hash func commented out
- [x] Mid-flow source port gets SNATted
- [ ] Exhausted uint16 flag
- [x] UT: #10913 
- [ ] FV: #11007
- [ ] E2E:
- Configuration:
  - [ ] Map size
  - [ ] Prime modulus 
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
BPF: Add Maglev consistent hash based load balancing for external traffic entering the cluster, controlled by Service annotation.
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/maglev.h
```diff
@@ -0,0 +1,52 @@
...
+static CALI_BPF_INLINE struct calico_nat_dest* maglev_select_backend(struct cali_tc_ctx *ctx)
+{
+	__u32 hash;
+	// ipv46_addr_t *ip_src = &ctx->state->ip_src;
+	ipv46_addr_t *ip_dst = &ctx->state->ip_dst;
+	// __u16 sport = ctx->state->sport;
+	__u16 dport = ctx->state->dport;
+	__u8 ip_proto = ctx->state->ip_proto;
+
+#if 0
```
fasaxc: The actual hash is commented out

felix/bpf-gpl/maglev.h
```diff
@@ -0,0 +1,52 @@
...
+	hash |= jhash_3words(ip_dst->c, ip_dst->d, (__u32)ip_proto, 0xC000);
+#else
+	hash = jhash_3words(*ip_src, sport, 0, 0xDEAD);
+	hash |= jhash_3words(*ip_dst, dport, (__u32)ip_proto, 0xBEEF);
+#endif /* IPVER6 */
+#endif
+	hash = 0;
+
+	struct calico_ch_key ch_key = {
+		.ordinal = (hash % 31),
```
fasaxc, line 34: 31?  Should this be 1009?

felix/bpf-gpl/nat_lookup.h
```diff
@@ -104,6 +104,15 @@ static CALI_BPF_INLINE struct calico_nat_dest* calico_nat_lookup(ipv46_addr_t *i
 		}
 		CALI_DEBUG("NAT: nodeport hit");
 	}
+
+# if HAS_MAGLEV_PROG
```
fasaxc: Should we move this above the block above so that we don't even try to do Maglev for nodeports (given that nodeports connections can't move between hosts?)

felix/bpf-gpl/nat_types.h
```diff
@@ -121,4 +123,23 @@ struct vxlanhdr {
...
+
+// 1009 * 1000 gives a large lookup table for up to 1000 services. -Alex
+#ifdef IPVER6
+CALI_MAP_NAMED(cali_v6_ch, cali_ch,,
+#else
+CALI_MAP_NAMED(cali_v4_ch, cali_ch,,
+#endif
+		BPF_MAP_TYPE_HASH,
+		struct calico_ch_key, struct calico_nat_dest,
+		1009*1000, BPF_F_NO_PREALLOC)
```
fasaxc, line 143: Let's define a constant for the maglev prime in maglev.h and use it here and in the actual lookup.  then we can replace that with config later.

felix/bpf-gpl/tc.c
```diff
@@ -2160,3 +2176,83 @@ int calico_tc_skb_ipv4_frag(struct __sk_buff *skb)
...
+			// Too late to try falling through?
+			CALI_DEBUG("Maglev: midflow miss, no backend found, drop.");
+			goto deny;
+		}
+
+		// deny_reason(ctx, CALI_REASON_NO_BACKEND);
+		CALI_DEBUG("Maglev: select backend failed");
+		goto deny;
+	}
+
```
fasaxc, line 2209: Commented?

felix/bpf-gpl/tc.c
```diff
@@ -2160,3 +2176,83 @@ int calico_tc_skb_ipv4_frag(struct __sk_buff *skb)
...
+	// CALI_CT_MID_FLOW_MISS implies traffic is TCP. If that changes,
+	// this condition should be adjusted.
+	if (ct_result_rc(ctx->state->ct_result.rc) == CALI_CT_MID_FLOW_MISS) {
+		/* treat it as if it was a new flow */
+		CALI_DEBUG("Maglev: mid-flow miss, treating as new flow");
+		/* remember that is was a mid-flow miss so that we can handle it in the new flow program */
+		ctx->state->ct_result.rc = CALI_CT_MAGLEV_MID_FLOW_MISS;
+	}
```
fasaxc, line 2235: TODO?

felix/bpf/consistenthash/test/types.go
```diff
@@ -0,0 +1,24 @@
+package types
```
fasaxc, line 1: New files should have apache headers/copyright

felix/bpf/nat/maps.go
```diff
@@ -363,6 +369,150 @@ func BackendValueFromBytes(b []byte) BackendValueInterface {
...
+
+var _ ConsistentHashBackendKeyInterface = ConsistentHashBackendKey{}
+
+const consistentHashBackendValueSize = backendValueSize
+
+var ConsistentHashMapParameters = maps.MapParameters{
+	Type:       "hash",
+	KeySize:    consistentHashBackendKeySize,
+	ValueSize:  consistentHashBackendValueSize,
+	MaxEntries: 1009 * 1000,
```
fasaxc, line 453: 1009 needs to be a constant

felix/bpf/nat/maps.go
```diff
@@ -363,6 +369,150 @@ func BackendValueFromBytes(b []byte) BackendValueInterface {
...
+		return false
+	}
+
+	for k, v := range m {
+		if v2, ok := cmp[k]; !ok || v != v2 {
+			return false
+		}
+	}
+
+	return true
```
fasaxc, line 476:
```suggestion
	return maps.Equal(m, cmp)
```

felix/bpf/nat/maps6.go
```diff
@@ -66,6 +67,74 @@ const backendKeyV6Size = 8
...
+//		__u32 ordinal; // should always be a value of [0..M], where M is a very large prime number. -Alex
+//	};
+const consistentHashBackendKeyV6Size = 24
+
+const consistentHashBackendValueV6Size = backendValueV6Size
+
+type ConsistentHashBackendKeyV6 [consistentHashBackendKeyV6Size]byte
+
+func NewConsistentHashBackendKeyV6(addr net.IP, port uint16, proto uint8, ordinal uint32) ConsistentHashBackendKeyV6 {
+	// TODO ADAPT TO V6
```
fasaxc, line 86: Out of date TODO now?

felix/bpf/nat/maps6.go
```diff
@@ -245,6 +314,72 @@ func BackendMapV6() maps.MapWithExistsCheck {
...
+		return false
+	}
+
+	for k, v := range m {
+		if v2, ok := cmp[k]; !ok || v != v2 {
+			return false
+		}
+	}
+
+	return true
```
fasaxc, line 345:
```suggestion
	return maps.Equal(m, cmp)
```

felix/bpf/proxy/kube-proxy.go
```diff
@@ -41,14 +41,15 @@ type KubeProxy struct {
...
-	backendMap  maps.MapWithExistsCheck
-	affinityMap maps.Map
-	ctMap       maps.Map
-	rt          *RTCache
-	opts        []Option
+	k8s               kubernetes.Interface
+	hostname          string
+	frontendMap       maps.MapWithExistsCheck
+	backendMap        maps.MapWithExistsCheck
+	ConsistentHashMap maps.MapWithExistsCheck
```
fasaxc: Does this need to be public?

felix/bpf/proxy/proxy.go
```diff
@@ -389,16 +389,23 @@ const (
 	ReapTerminatingUDPImmediatelly = "TerminatingImmediately"
 
 	ExcludeServiceAnnotation = "projectcalico.org/natExcludeService"
+	LoadBalancingAlgorithm   = "projectcalico.org/loadbalancingAlgorithm"
```
fasaxc: Out of date name for annotation?

fasaxc: Was expecting 
```
 lb.projectcalico.org/external-traffic-strategy
```

felix/bpf/proxy/proxy.go
```diff
@@ -389,16 +389,23 @@ const (
...
+)
+
+var (
+	LoadBalancingAlgorithmConsistentHash = "ConsistentHash"
```
fasaxc:
```suggestion
	LoadBalancingAlgorithmMaglev = "Maglev"
```

felix/bpf/proxy/syncer.go
```diff
@@ -814,6 +843,43 @@ func (s *Syncer) updateService(skey svcKey, sinfo Service, id uint32, eps []k8sp
...
+	return consistenthash.New(consistenthash.WithHash(fnv.New32(), fnv.New32()))
+}
+
+func (s *Syncer) writeConsistentHashBackends(skey svcKey, sinfo Service, eps []k8sp.Endpoint) {
+	if len(eps) == 0 || eps == nil {
+		log.WithFields(log.Fields{"service": skey, "name": sinfo.String()}).Info("Skipping ConsistentHash-enabled service update for empty endpoint list")
+		return
+	}
+
+	// ALEX TODO loadbalancer IP awareness.
```
fasaxc: TODO: load balancer IPs.  I think we do want to do those!

Why is the Maglev map indexed on dest IP instead of service index?  We'll need to write the Maglev backends once for LB IP and once for Cluster IP instead of sharing?

For v1, let's keep as is and add the load balancer IPs, doubling map sizze

felix/bpf/proxy/syncer.go
```diff
@@ -814,6 +843,43 @@ func (s *Syncer) updateService(skey svcKey, sinfo Service, id uint32, eps []k8sp
...
+			consistentHash.AddBackend(ep)
+		}
+	}
+	s.consistentHashMap[skey] = consistentHash
+
+	n := time.Now()
+	lut := consistentHash.Generate()
+	s.writeConsistentHashSvcBackends(vip, port, proto, lut)
+	dt := time.Since(n)
+	log.Infof("Wrote ConsistentHash-enabled service '%s' backends (%d) in %fs", skey.sname, len(eps), dt.Seconds())
```
fasaxc: Might be too verbose; do we write a log for every normal service?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+
+type backend struct {
+	permutation []int
+	endpoint    k8sp.Endpoint
+}
+
+// New returns a backend-hashing module.
+func New(o ...ConsistentHashOpt) *ConsistentHash {
+	c := &ConsistentHash{m: M}
+	c.backendNames = make([]string, 0)
```
fasaxc, line 59: Normally best to use a nil slice for 0-length; you can still append to it later.

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+type backend struct {
+	permutation []int
+	endpoint    k8sp.Endpoint
+}
+
+// New returns a backend-hashing module.
+func New(o ...ConsistentHashOpt) *ConsistentHash {
+	c := &ConsistentHash{m: M}
```
fasaxc, line 58: Set a default for h1/h2, require m as a parameter?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+			if n == ch.m {
+				break populate
+			}
+		}
+	}
+
+	return lut
+}
+
+// Permutation implements Permutator interface.
```
fasaxc, line 169: What Permutor interface?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+func (c *ConsistentHash) permutation(backendName string) ([]int, error) {
+	offset, skip, err := c.offsetAndSKip(backendName)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't generate permutation skip/offset for backend '%s': %w", backendName, err)
+	}
+
+	permutation := make([]int, c.m)
+	for j := range c.m {
+		permutation[j] = (offset + (j * skip)) % c.m
+	}
```
fasaxc, line 179: Do you need to actually build the permutation slice?  Looks like you could calculate `permutation[j]` on demand and avoid storing a large slice per backend. (I think that might be the point of using a prime and the clever offset/skip bit from the Maglev paper.)

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+	return permutation, nil
+}
+
+func (c *ConsistentHash) offsetAndSKip(s string) (int, int, error) {
+	offset, err := hashFromString(s, c.h1, []byte{0})
+	if err != nil {
+		return 0, 0, err
+	}
+
+	skip, err := hashFromString(s, c.h2, []byte{0xa})
```
fasaxc, line 190: Seems odd to have two hash functions but to also seed the second one.  Perhaps just remove the optionality of having 2 and inline use of FNV with appropriate seeds?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+
+	b.endpoint = kep
+	b.permutation = permutation
+
+	ch.backendsByName[name] = b
+	ch.backendNames = append(ch.backendNames, name)
+}
+
+func (ch *ConsistentHash) RemoveBackend(kep k8sp.Endpoint) {
+	name := kep.String()
```
fasaxc, line 103: Call to String() should be after nil check; do we actually expect calls with `nil`?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+	if kep == nil {
+		logrus.Warn("Ignoring RemoveBackend for nil endpoint")
+		return
+	}
+
+	b, exists := ch.backendsByName[name]
+	if !exists {
+		logrus.WithField("backend", name).Info("Aborting RemoveBackend for non-existent backend")
```
fasaxc:
```suggestion
		logrus.WithField("backend", name).Warn("Ignoring RemoveBackend for non-existent backend")
```

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+		return
+	}
+
+	b.endpoint = nil
+	delete(ch.backendsByName, name)
+
+	for i, b := range ch.backendNames {
+		if b == name {
+			slices.Delete(ch.backendNames, i, i)
```
fasaxc, line 120: Break out of loop once found?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// Generate sorts the list of backends and then generates a ConsistentHash LUT.
+func (ch *ConsistentHash) Generate() []k8sp.Endpoint {
+	if len(ch.backendNames) == 0 {
+		return nil
+	}
+
+	slices.Sort(ch.backendNames)
+	logrus.WithField("backends", ch.backendNames).Info("sorted backend names")
```
fasaxc, line 132: Drop to debug?  Could be large

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+				next[i]++
+				choice = prefs[next[i]]
+			}
+
+			lut[choice] = ch.backendsByName[backend].endpoint
+
+			// Its *next* preference (after this one). Not sure if necessary to remember this.
+			// Maybe it makes something easier later on if this info is to-hand?
+			// If its safe to discard, we can probably rewrite this whole func to be more Go-ful.
+			next[i]++
```
fasaxc, line 158: You need this because you don't want to start scanning the permutations slice from the start again; you know all the ones up to here have been used already.

felix/bpf/consistenthash/consistenthash_test.go
```diff
@@ -0,0 +1,78 @@
...
+			mag = newDefaultConsistentHash()
+		})
+
+		It("generate valid permutations", func() {
+			for _, b := range []string{"backend1", "backend2", "backend3", "sdfskjnksdnf", "sdfkjsdnfksjndf", "i am a backend with a very very very long nameeeeeeeeeeeeeeeeeeeeee"} {
+				p, err := mag.permutation(b)
+				Expect(err).NotTo(HaveOccurred())
+				dupMap := make(map[int]bool)
+				for _, idx := range p {
+					Expect(dupMap).NotTo(HaveKey(idx))
```
fasaxc, line 52: Worth asserting that the number is in range too

felix/bpf/consistenthash/consistenthash_test.go
```diff
@@ -0,0 +1,78 @@
...
+					dupMap[idx] = true
+				}
+			}
+		})
```
fasaxc, line 56: Could check that each permutation is unique too?

felix/bpf/consistenthash/consistenthash_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+		It("should create a lookup table", func() {
+			for _, b := range []string{"backend0", "backend1", "backend2", "backend3", "backend4", "backend5", "backend6", "backend7", "backend8", "backend9"} {
+				addBackend(mag, b, uint16(8080))
+			}
+
+			time.Sleep(4 * time.Second)
```
fasaxc: Sleep?

felix/bpf/consistenthash/consistenthash_test.go
```diff
@@ -0,0 +1,78 @@
...
+			lut := mag.Generate()
+			Expect(lut).NotTo(ContainElement(""))
+			Expect(lut).To(HaveLen(testM))
+		})
```
fasaxc, line 66: Other things to check:
* Each backend appears in the LUT 
* The max/min number of appearances of each LUT is as expected.  Since we backends in round robin fashion, you should get a skew of at most 1?

felix/bpf-gpl/nat_types.h
```diff
@@ -121,4 +123,23 @@ struct vxlanhdr {
 	__be32 flags;
 	__be32 vni;
 };
+
+struct calico_ch_key {
+	ipv46_addr_t vip;
+	__u16 port;
+	__u8 proto;
+	__u8 pad;
+	__u32 ordinal; // should always be a value of [0..M], where M is a very large prime number. -Alex
```
fasaxc: `[0-M)` ?  I'd expect M values in total, so, with 0-indexing, that's M-1 as the max value

fasaxc (changes requested):
Looking plausible so far; I didn't review the BPF C code in detail because:

* The hash itself was commented out 
* We're aware of a mid-flow bug where source IP is getting SNATted.

It passed the sniff test but I didn't dig in deeply.

felix/bpf-gpl/maglev.h
aaaaaaaalex: Yep - this is fixed in the WIP UT work :+1:

felix/bpf-gpl/maglev.h
aaaaaaaalex: Made it lower for the tests - must make it configurable

felix/bpf-gpl/tc.c
aaaaaaaalex: Must check with Tomas if he's happy with what the code is doing - originally his code.
Can cleanup the comments then

felix/bpf/consistenthash/consistenthash.go
aaaaaaaalex: But IIUC, if a backend with a lower index in the list disappears, it may free-up a higher-preference option for a subsequent backend  - so we would need to start over?

felix/bpf/consistenthash/consistenthash_test.go
aaaaaaaalex: No clue - past Alex is an enigma (fixed)

felix/bpf-gpl/maglev.h
fasaxc: Sorry, I think you should set this to a prod-ready value until you implement config.  It's too easy to miss this fix-up if you forget to do the config or got pressure to ship without.

felix/bpf/consistenthash/consistenthash.go
fasaxc: Yes, but this variable is local to the Generate function, we're only keeping track of this particular generation.

felix/bpf-gpl/nat_lookup.h
tomastigera: I think this is moved down here only because we know here that we have a value. Could be done higher up, but it wont take the branch that may do nodeport resolution in case the dest ip does not make a direct hit in NAT. That branch is essentially for nodeports being resolved for pods without CTLB (accessing nodeport of a different node).

This said, we should make sure that kube-proxy ignores maglev flag for nodeports

tomastigera:
deprecated by https://github.com/projectcalico/calico/pull/11159

PR #10905, opened 2025-09-01
Refine CI deps

fasaxc:
### Add new deps tool

```
go run ./hack/cmd/deps/ --help
CI Dependency helper tool.

Usage: 

  deps modules <package>          # Print go modules that package depends on
  deps local-dirs <package>       # Print in-repo go package dirs that 
                                  # package depends on.
  deps test-exclusions <package>  # Print glob patterns to match *_test.go 
                                  # files in dependency dirs outside the 
                                  # package itself.
  deps sem-change-in <package>[,<secondary package>...] # Print a SemaphoreCI
                                  # conditions DSL change_in() clause for 
                                  # <package>, including non-test deps from
                                  # any <secondary package> clauses.

The test-exclusions and sem-change-in sub-commands are intended to be used with
packages at the top-level of the repo.  Test exclusions are based on whether
a dependency is within the package.

The change_in() clause always depends on the whole package directory itself. 
Some non-Go dependencies are hard-coded in the tool.  For example, it knows
that node depends on felix/bpf-*.
```

### Refine change_in() clauses

Via the new tool:

- Use `go list` to find the real dependency packages.
- Filter that down to the packages within our repo.
- Write out:
  - change_in **inclusions** for:
     -  The "top level" package (such a `felix`) and all files within that tree.  "Felix should build if any aspect of felix or its tests change."
     - Dependency packages in other top-level directories.  "Felix depends on libcalico-go/lib/set, ...".
     - Some hard-coded files, such as `lib.Makefile` and `metadata.mk`.  Future work: try to figure out which variables in metadata.mk a particular build depends on and only rebuild if those change.
  - change_in **exclusions** for:
    - Various hard-coded files (README, LICENSE, etc)
    - Test files outside the top-level directory.  "No need to run felix's tests just because you changed the test files in libcalico-go/lib/set".
- Manually clean up non-Go `change_in()` expressions.  Most only depend on the library makefile, not go.mod etc.

Apply to most semaphore CI yaml blocks.  (But keep those for non-Go packages like whisker.)

### Avoid full rebuild on go.mod change

Not every package depends on every go module that the overall repo depends on.  Rather than triggering all builds when go.mod changes, use the deps tool to output a dependency summary file `*/deps.txt` to each directory that contains the subset of go.mod that that directory depends on. 

- Remove go.mod from the change_in clause.
- Include the new file.
- Enforce generation of the new files in CI.

There's no free lunch, if the version of go or k8s changes then most things will get rebuilt.  However, some key top-level dirs are winners: the release tooling should now be isolated from the main builds, for example.

## Related issues/PRs

CORE-11778

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -134,7 +134,7 @@ blocks:
             - make ci-preflight-checks
   - name: API
     run:
-      when: "false or change_in(['/*', '/api/'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
+      when: "false or change_in(['/metadata.mk', '/lib.Makefile', '/api/'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
     execution_time_limit:
       minutes: 30
     dependencies:
@@ -294,7 +294,7 @@ blocks:
             - ../.semaphore/run-and-monitor ci.log make ci
   - name: e2e script & pipeline validations
     run:
-      when: "false or change_in(['/*', '/.semaphore/end-to-end/'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
+      when: "false or change_in(['/.semaphore/end-to-end/'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
     execution_time_limit:
       minutes: 5
     dependencies:
@@ -320,7 +320,7 @@ blocks:
             - .semaphore/run-and-monitor e2e-test.log make e2e-test
   - name: envoy-gateway
     run:
-      when: "false or change_in(['/*', '/third_party/envoy-gateway'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
+      when: "false or change_in(['/metadata.mk', '/lib.Makefile', '/third_party/envoy-gateway'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
     execution_time_limit:
       minutes: 30
     dependencies:
@@ -344,7 +344,7 @@ blocks:
             - ../../.semaphore/run-and-monitor image-$ARCH.log make build ARCH=$ARCH
   - name: envoy-proxy
     run:
-      when: "false or change_in(['/*', '/third_party/envoy-proxy'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
+      when: "false or change_in(['/metadata.mk', '/lib.Makefile', '/third_party/envoy-proxy'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
     execution_time_limit:
       minutes: 30
     dependencies:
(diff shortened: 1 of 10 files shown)


.semaphore/semaphore.yml
```diff
@@ -149,7 +149,7 @@ blocks:
...
...st.go','libcalico-go/lib/apis/v3/*_test.go','libcalico-go/lib/backend/*_test.go','libcalico-go/lib/backend/api/*_test.go','libcalico-go/lib/backend/encap/*_test.go','libcalico-go/lib/backend/etcdv3/*_test.go','libcalico-go/lib/backend/k8s/*_test.go','libcalico-go/lib/backend/k8s/conversion/*_test.go','libcalico-go/lib/backend/k8s/resources/*_test.go','libcalico-go/lib/backend/k8s/scheme/*_test.go','libcalico-go/lib/backend/model/*_test.go','libcalico-go/lib/backend/watchersyncer/*_test.go','libcalico-go/lib/clientv3/*_test.go','libcalico-go/lib/errors/*_test.go','libcalico-go/lib/hash/*_test.go','libcalico-go/lib/ipam/*_test.go','libcalico-go/lib/json/*_test.go','libcalico-go/lib/logutils/*_test.go','libcalico-go/lib/names/*_test.go','libcalico-go/lib/namespace/*_test.go','libcalico-go/lib/net/*_test.go','libcalico-go/lib/options/*_test.go','libcalico-go/lib/resources/*_test.go','libcalico-go/lib/selector/*_test.go','libcalico-go/lib/selector/parser/*_test.go','libcalico-go/lib/selector/tokenizer/*_test.go','libcalico-go/lib/set/*_test.go','libcalico-go/lib/validator/v3/*_test.go','libcalico-go/lib/watch/*_test.go','libcalico-go/lib/winutils/*_test.go','pkg/buildinfo/*_test.go']})"
```
Brian-McM, line 152: Out of curiosity, in your testing did you run into any issues with having a large number of files in this change_in block?

hack/cmd/deps/deps.go
```diff
@@ -0,0 +1,311 @@
...
+		printUsageAndExit()
+	}
+
+	cmd := os.Args[1]
+	pkg := os.Args[2]
+
+	switch cmd {
+	case "modules":
+		printModules(pkg)
+	case "local-dirs":
```
Brian-McM, line 62: I'm likely just missing this (since github hides large diffs), but where is local-dirs used?

.semaphore/semaphore.yml
fasaxc: No, I was worried about that, but it "just worked", I haven't had to cull it.

hack/cmd/deps/deps.go
fasaxc: It's not used in CI, I was using it for local debugging (easier to read than the change_in clauses!).

.semaphore/semaphore.yml.d/blocks/20-api.yml
```diff
@@ -1,6 +1,6 @@
 - name: API
   run:
-    when: "${FORCE_RUN} or change_in(['/*', '/api/'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
+    when: "${FORCE_RUN} or change_in(['/metadata.mk', '/lib.Makefile', '/api/'], {exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
```
Brian-McM, line 3: Should this be change to use the new CHANGE_IN substitute?

fasaxc: I think the api dir is special, not allowed to depend on anything else in the monorepo so I left it as-is.

Brian-McM: Got it.

PR #10913, opened 2025-09-03
Maglev: Get UTs into basebranch, Fixes colliding tunnel-IP bug

aaaaaaaalex:
## Description

TC: handle midflow Maglev packets like new connections.
UT: TCP NodePort tests and midflow TCP failover handling test.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -474,9 +474,6 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 		CALI_DEBUG("NAT Lookup determined packet handled by maglev");
 		CALI_JUMP_TO(ctx, PROG_INDEX_MAGLEV);
 		CALI_DEBUG("Failed to jump to maglev program");
-		#if !HAS_MAGLEV
-		CALI_DEBUG("Maglev program missing");
-		#endif
 		goto deny;
 	}
 
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -2229,6 +2229,8 @@ int calico_tc_maglev(struct __sk_buff *skb)
 		ctx->state->flags |= CALI_ST_SKIP_REDIR_PEER;
 	}
 
+	// CALI_CT_MID_FLOW_MISS implies traffic is TCP. If that changes,
+	// this condition should be adjusted.
 	if (ct_result_rc(ctx->state->ct_result.rc) == CALI_CT_MID_FLOW_MISS) {
 		/* treat it as if it was a new flow */
 		CALI_DEBUG("Maglev: mid-flow miss, treating as new flow");
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -63,7 +63,6 @@
 #endif
 
 #define HAS_HOST_CONFLICT_PROG CALI_F_TO_HEP
-#define HAS_MAGLEV_PROG        (CALI_F_FROM_HEP && CALI_F_MAIN)
 
 /* calico_tc_main is the main function used in all of the tc programs.  It is specialised
  * for particular hook at build time based on the CALI_F build flags.
@@ -475,7 +474,7 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 		CALI_DEBUG("NAT Lookup determined packet handled by maglev");
 		CALI_JUMP_TO(ctx, PROG_INDEX_MAGLEV);
 		CALI_DEBUG("Failed to jump to maglev program");
-		#if !HAS_MAGLEV_PROG
+		#if !HAS_MAGLEV
 		CALI_DEBUG("Maglev program missing");
 		#endif
 		goto deny;
@@ -2181,7 +2180,7 @@ deny:
 }
 #endif /* !IPVER6 */
 
-#if HAS_MAGLEV_PROG
+#if HAS_MAGLEV
 SEC("tc")
 int calico_tc_maglev(struct __sk_buff *skb)
 {
@@ -2234,6 +2233,7 @@ int calico_tc_maglev(struct __sk_buff *skb)
 		/* treat it as if it was a new flow */
 		CALI_DEBUG("Maglev: mid-flow miss, treating as new flow");
 		/* remember that is was a mid-flow miss so that we can handle it in the new flow program */
+		//ctx->state->ct_result.rc = CALI_CT_MAGLEV_MID_FLOW_MISS;
 	}
 
 	CALI_DEBUG("Maglev: About to jump to policy program.");
@@ -2256,4 +2256,4 @@ int calico_tc_maglev(struct __sk_buff *skb)
 deny:
 	return TC_ACT_SHOT;
 }
-#endif /* HAS_MAGLEV_PROG */
+#endif /* HAS_MAGLEV */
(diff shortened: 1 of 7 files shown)


felix/bpf-gpl/tc.c
```diff
@@ -475,7 +474,7 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 		CALI_DEBUG("NAT Lookup determined packet handled by maglev");
 		CALI_JUMP_TO(ctx, PROG_INDEX_MAGLEV);
 		CALI_DEBUG("Failed to jump to maglev program");
-		#if !HAS_MAGLEV_PROG
+		#if !HAS_MAGLEV
 		CALI_DEBUG("Maglev program missing");
 		#endif
```
tomastigera: this is redundant, there is already CALI_DEBUG above that. With the change in nat_lookup we should not get here if `nat_res == NAT_MAGLEV` as such a program must `HAVE_MAGLEV` anything else is a serious bug in how we are setting up programs.

felix/bpf/ut/maglev_test.go
```diff
@@ -483,7 +483,876 @@ func TestMaglevNATPodPodXNodeUDP(t *testing.T) {
 	resetCTMap(ctMap)
 }
 
-func TestMaglevNATNodePort(t *testing.T) {
+func TestMaglevNATNodePortTCP(t *testing.T) {
```
tomastigera:
```suggestion
func TestMaglevTCPFailover(t *testing.T) {
```

felix/bpf/ut/maglev_test.go
```diff
@@ -483,7 +483,876 @@ func TestMaglevNATPodPodXNodeUDP(t *testing.T) {
...
+		nat.NewNATKey(ipv4.DstIP, uint16(tcp.DstPort), uint8(layers.IPProtocolTCP)).AsBytes(),
+		nat.NewNATValueWithFlags(0, 1, 0, 0, nat.NATFlgConsistentHash).AsBytes(),
+	)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = natBEMap.Update(
+		nat.NewNATBackendKey(0, 0).AsBytes(),
+		nat.NewNATBackendValue(natIP, natPort).AsBytes(),
+	)
+	Expect(err).NotTo(HaveOccurred())
```
tomastigera: Would it make sense not to program these backends to possibly trigger an error? It would deviate from what kube-proxy will eventually program, but we want to test a specific path and we want to error on the possibly unwanted fail over path.

felix/bpf/ut/maglev_test.go
```diff
@@ -483,7 +483,876 @@ func TestMaglevNATPodPodXNodeUDP(t *testing.T) {
...
+			ipv4R := ipv4L.(*layers.IPv4)
+			Expect(ipv4R.SrcIP.String()).To(Equal(node2ip.String()))
+			Expect(ipv4R.DstIP.String()).To(Equal(node1ip.String()))
+
+			checkVxlan(pktR)
+		}, withHostNetworked())
+	}
+}
+
+func TestMaglevNATNodePortUDP(t *testing.T) {
```
tomastigera:
```suggestion
func TestMaglevXNodeForwarding(t *testing.T) {
```

felix/bpf/ut/maglev_test.go
```diff
@@ -483,7 +483,876 @@ func TestMaglevNATPodPodXNodeUDP(t *testing.T) {
...
+		Expect(ctr.Data().A2B.Approved).To(BeTrue())
+		// Approved destination side as well
+		Expect(ctr.Data().B2A.Approved).To(BeTrue())
+	})
+
+	skbMark = 0
+
+	// Response leaving workload at node 2
+	runBpfTest(t, "calico_from_workload_ep", rulesDefaultAllow, func(bpfrun bpfProgRunFn) {
+		respPkt := tcpResponseRaw(recvPkt)
```
tomastigera, line 531: is it SYN+ACK ?

felix/bpf/ut/maglev_test.go
```diff
@@ -483,7 +483,876 @@ func TestMaglevNATPodPodXNodeUDP(t *testing.T) {
...
+		fmt.Printf("pktR = %+v\n", pktR)
+
+		ipv4L := pktR.Layer(layers.LayerTypeIPv4)
+		Expect(ipv4L).NotTo(BeNil())
+		ipv4R := ipv4L.(*layers.IPv4)
+		// check that the IP is fixed up
+		Expect(ipv4R.SrcIP.String()).To(Equal(node2ip.String()))
+		Expect(ipv4R.DstIP.String()).To(Equal(node1ip.String()))
+
+		checkVxlan(pktR)
```
tomastigera, line 585: we should probably consider DSR for this test. Also makes it simpler

felix/bpf/ut/maglev_test.go
```diff
@@ -483,7 +483,876 @@ func TestMaglevNATPodPodXNodeUDP(t *testing.T) {
...
+
+		tcpL := pktR.Layer(layers.LayerTypeTCP)
+		Expect(tcpL).NotTo(BeNil())
+		tcpR := tcpL.(*layers.TCP)
+
+		icmpMTUTooBig = makeICMPErrorFrom(net.ParseIP("69.69.69.69"), ipv4R, tcpR,
+			layers.ICMPv4TypeDestinationUnreachable, layers.ICMPv4CodeFragmentationNeeded)
+	})
+
+	dumpCTMap(ctMap)
```
tomastigera: this does not add any value to the test, this is already tested in the UDP test and nodeport tests.

felix/bpf/ut/maglev_test.go
```diff
@@ -483,7 +483,876 @@ func TestMaglevNATPodPodXNodeUDP(t *testing.T) {
...
+		icmpL := pktR.Layer(layers.LayerTypeICMPv4)
+		Expect(icmpL).NotTo(BeNil())
+
+		icmpEncaped = res.dataOut
+	})
+
+	skbMark = 0
+	// Another pkt with a different source port arriving at node 1.
+	// This time, not a SYN, i.e. a failed-over connection.
+	runBpfTest(t, "calico_from_host_ep", nil, func(bpfrun bpfProgRunFn) {
```
tomastigera: i suppose this would need to be on node3 (noed1 without CT restored would work the same, but perhaps node3 is better for clarity), we should also have the test case node2 (lands on the node with the backend as a failover)

Because of the tests above, all is set up on node1 and a non-SYN packet is not really a fail-over. It will hit CT, right?

felix/bpf/ut/maglev_test.go
aaaaaaaalex: Good catch - I don't think I modified the flags for the return patch.

aaaaaaaalex: Just checked the helper func - it sets ACK when SYN :ok_hand:

felix/bpf/ut/maglev_test.go
aaaaaaaalex: Would like to have both - can trigger a new, DSR flow after the non-DSR

felix/bpf/ut/maglev_test.go
aaaaaaaalex: Will make it node 3 and will clear out the CT map :+1:

felix/bpf/ut/maglev_test.go
aaaaaaaalex: Removed the ICMP networking from the test now, is that what you intended for me to remove?

felix/bpf/ut/maglev_test.go
aaaaaaaalex: Done

aaaaaaaalex:
Going to merge this into the base branch so that @fasaxc can get to work in parallel to the UT work.
Can come back and shave down the UTs later

PR #10930, opened 2025-09-04
Fix test flake due to env pollution

caseydavenport:
## Description

Spotted the flake here: https://tigera.semaphoreci.com/jobs/c459125c-606a-42c9-a636-a36520aa284d#L1339-L1367

Tests were leaving behind env changes that impacted other tests,
depending on test ordering.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/pkg/lifecycle/startup/startup_test.go
+++ b/node/pkg/lifecycle/startup/startup_test.go
@@ -89,11 +89,27 @@ func makeK8sNode(ipv4 string, ipv6 string) *v1.Node {
 	return node
 }
 
+// setEnv sets an environment variable and returns a function to restore the original value.
+func setEnv(k, v string) func() {
+	originalValue, hadOriginalValue := os.LookupEnv(k)
+	err := os.Setenv(k, v)
+	Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Error setting env var %s: %v", k, err))
+	return func() {
+		var err error
+		if hadOriginalValue {
+			err = os.Setenv(k, originalValue)
+		} else {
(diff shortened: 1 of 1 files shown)


node/pkg/lifecycle/startup/startup_test.go
```diff
@@ -89,11 +89,27 @@ func makeK8sNode(ipv4 string, ipv6 string) *v1.Node {
 	return node
 }
 
+// setEnv sets an environment variable and returns a function to restore the original value.
+func setEnv(k, v string) func() {
```
MichalFupso: A nit, maybe something like temporarySetEnv would be a better name?

node/pkg/lifecycle/startup/startup_test.go
```diff
@@ -89,11 +89,27 @@ func makeK8sNode(ipv4 string, ipv6 string) *v1.Node {
 	return node
 }
 
+// setEnv sets an environment variable and returns a function to restore the original value.
```
coutinhop: I think it'd be good to elaborate this comment with more details on how the function should be used, i.e., with `defer` to the function that is returned, etc. For example, if someone misses the last pair of `()` in  `defer setEnv("IP", "none")()`, they'd actually defer the setting of the env var, and restoring the original value would never happen, right? (sorry I'm just worried that the syntax may be a bit confusing, I can definitely see myself making that mistake in the future if I'm not careful 😅 )

PR #10931, opened 2025-09-04
OpenStack: Respond immediately to all QoS configuration changes

nelljerram:
In particular to these two classes of change for which we previously relied on eventual consistency via the resync process:

1. Adding/updating/removing `qos_policy_id` value to/in/from a Network with ports that don't have their own `qos_policy_id` field.

2. Adding/updating/removing a QoS rule to/in/from a QoS policy that is already in use.

Implementation details:

- Implement `update_network_postcommit` method in the mechanism driver, to detect when `qos_policy_id` value is changing and re-emit the WorkloadEndpoint data for each affected port. This covers case (1).

- Implement `update_policy` method in `CalicoQoSDriver` class, to call on to the mechanism driver and re-emit the WorkloadEndpoint data for each affected port.  This is called whenever the rules change for an existing QoS policy, and so covers case (2).

Testing details:

- Add a new step into our DevStack CI to run Calico-specific FV tests, before the existing Tempest tests.  The first FV tests, in this PR, test that WorkloadEndpoint QoSControls are correctly and promptly updated for all possible transitions that:

  - Modify the `qos_policy_id` value for a port, or for that port's network.

  - Modify the rules for an existing QoS policy.

Related fixes, discovered by the new testing:

- Calico plugin init should only add 'qos' to ML2 extension_drivers if not already present. Otherwise QoS handling code gets called twice per operation, which causes it to fail:

      2025-08-27 18:06:19,989 - ERROR - Test execution failed: ConflictException: 409: Client Error for url: http://192.168.122.233:9696/networking/v2.0/networks, Failed to create a duplicate QosPolicyNetworkBinding: for attribute(s) ['qos_network_policy_bindings.network_id'] with value(s) 9bf10a53-926e-4511-ab3c-756faca44c29

  Apparently DevStack setup includes 'qos', and then our plugin was adding it again:

      Aug 27 14:30:24 semaphore-vm neutron-server[71036]: DEBUG oslo_service.service [-] ml2.extension_drivers          = ['port_security', 'qos', 'qos'] {{(pid=71036) log_opt_values /opt/stack/data/venv/lib/python3.10/site-packages/oslo_config/cfg.py:2620}}

- For some reason it works better to get the QoS policy ID for a port's network as `port.get("qos_network_policy_id")` than it does as `network["qos_policy_id"]`, so I've modified the code to try both.  Example of the `qos_network_policy_id` attribute:

      Aug 28 22:29:51 semaphore-vm neutron-server[69834]: DEBUG networking_calico.plugins.ml2.drivers.calico.endpoints [None req-ecf3bc4e-e1c7-426c-862c-6611c3ad0c3e None None] port = {'id': '35d29ef7-38b6-4a99-b8cf-684a75fb4e71', 'name': 'test-port-network-qos-62416584', 'network_id': '22fca5e7-85c7-4267-8af1-8972b56b0184', 'tenant_id': '2503291a80764895813deda9dc8b082b', 'mac_address': 'fa:16:3e:e4:d0:e9', 'admin_state_up': True, 'status': 'DOWN', 'device_id': '', 'device_owner': 'compute:', 'standard_attr_id': 216, 'fixed_ips': [{'subnet_id': '02c83fed-e0c7-45f6-9aa9-a7711448f198', 'ip_address': '192.168.100.215'}], 'allowed_address_pairs': [], 'extra_dhcp_opts': [], 'security_groups': ['38323e51-1163-4c39-9904-979441de45a6'], 'description': '', 'binding:vnic_type': 'normal', 'binding:profile': {}, 'binding:host_id': '', 'binding:vif_type': 'unbound', 'binding:vif_details': {}, 'port_security_enabled': True, 'qos_policy_id': None, 'qos_network_policy_id': 'a8eb0869-a39e-480c-b003-24f8c1f762e7', 'resource_request': None, 'tags': [], 'created_at': '2025-08-28T22:29:32Z', 'updated_at': '2025-08-28T22:29:32Z', 'revision_number': 1, 'project_id': '2503291a80764895813deda9dc8b082b'} {{(pid=69834) get_extra_port_information /opt/stack/data/venv/lib/python3.10/site-packages/networking_calico/plugins/ml2/drivers/calico/endpoints.py:257}}

## Release Note

```release-note
OpenStack QoS handling has been improved to respond immediately to all kinds of QoS changes on the Neutron API.  In particular when rules are added, updated or deleted for a QoS policy that is already in use, and when a QoS policy is added, updated or deleted for a Neutron network that is already in use.  Calico's handling of these changes previously relied on a periodic resync, which could in practice take a long time; now those changes are handled immediately (within a few seconds).
```


networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -687,16 +685,101 @@ def get_mappings(self, agent):
...
+                filters={
+                    "network_id": [network_id],
+                    "qos_policy_id": ["", None],
+                },
+            )
+            self.update_existing_ports(
+                [p for p in ports if not p["qos_policy_id"]],
+                plugin_context,
+                "network changing qos_policy_id"
+            )
```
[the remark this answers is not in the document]
nelljerram: @copilot Yes, that's what I thought too.  But for some reason the preceding filter on `qos_policy_id` does not work.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -687,16 +685,101 @@ def get_mappings(self, agent):
...
+        with self._txn_from_context(plugin_context, tag="handle_qos_policy_update"):
+            # Update the existing ports that are directly using this qos_policy_id.
+            ports = self.db.get_ports(
+                plugin_context,
+                filters={
+                    "qos_policy_id": [policy_id],
+                },
+            )
+            self.update_existing_ports(
+                [p for p in ports if p["qos_policy_id"] == policy_id],
```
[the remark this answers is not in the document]
nelljerram: @copilot Same again.  For some reason the `qos_policy_id` filter does not reliably work.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -687,16 +685,101 @@ def get_mappings(self, agent):
...
+                plugin_context,
+                filters={
+                    "network_id": [
+                        n["id"] for n in networks if n["qos_policy_id"] == policy_id
+                    ],
+                    "qos_policy_id": ["", None],
+                },
+            )
+            self.update_existing_ports(
+                [p for p in ports if not p["qos_policy_id"]],
```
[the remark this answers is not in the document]
nelljerram: @copilot Same again.

networking-calico/devstack/qos_responsiveness_tests.py
```diff
@@ -0,0 +1,567 @@
...
+            "qos_policies": [],
+            "security_groups": [],
+        }
+
+        # Set up OpenStack connection
+        cls.conn = openstack.connection.Connection(
+            auth_url=os.environ.get("OS_AUTH_URL", "http://localhost/identity"),
+            project_name=os.environ.get("OS_PROJECT_NAME", "admin"),
+            username=os.environ.get("OS_USERNAME", "admin"),
+            password=os.environ.get("OS_PASSWORD", "015133ea2bdc46ed434c"),
```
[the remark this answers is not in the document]
nelljerram: @copilot This value is only used for this repo's own testing; it's not a real secret.

nelljerram:
For the sake of full disclosure and fair attribution...

- I kicked off this work in #10865 as a Copilot task.  Copilot did some initial steps but then I took over and added a very long string of sausage commits.  Because it was started by Copilot, the branch for that PR was in the projectcalico/calico repo, not my fork.
I wanted to tidy up the commits and messages, so I squashed everything locally, but then was blocked force-pushing to the branch on projectcalico/calico.  (I think I do normally have creds for pushing to projectcalico/calico, so this may be Copilot-specific, unless something else has recently changed.)

- Therefore I created this PR instead, using my own fork.  However this one was initially stuck on the CLA check, and it appears that was because I preserved Copilot as the original commit author, and/or because of the `Co-authored-by: Nell Jerram <nell@tigera.io>` that I put in the commit message.

- I've now changed commit authorship to myself, and removed the `Co-authored-by` line, and the CLA bot is happy.  So I'm leaving this comment to record Copilot's involvement.

nelljerram:
Raised https://tigera.atlassian.net/browse/RS-2757 to track the DHCP agent's problem with CIDR re-use.

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
```diff
@@ -336,7 +336,11 @@ def cap(setting, minmax):
                 setting = max
             return setting
 
-        qos_policy_id = port.get("qos_policy_id") or port_extra.network_qos_policy_id
+        qos_policy_id = (
+            port.get("qos_policy_id")
+            or port.get("qos_network_policy_id")
+            or port_extra.network_qos_policy_id
```
fasaxc: This is suspicious: the method that loads the port_extra loads a db model object, which doesn't seem to have this field.  I think that field is only on the object layer version.

https://github.com/nelljerram/calico/blob/6cdfa945379c977cb0876063740f32314c439b86/networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py#L248

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -697,16 +695,101 @@ def get_mappings(self, agent):
...
+        LOG.info(
+            "qos_policy_id for network %r changing from %r to %r",
+            network_id,
+            old_qos_policy_id,
+            new_qos_policy_id,
+        )
+
+        # Update the existing ports for this network and which don't have their own
+        # qos_policy_id.
+        plugin_context = context._plugin_context
```
fasaxc, line 726:
```suggestion
        plugin_context = context.plugin_context
```
Looks like there's an accessor for this field?

I'm not 100% sure that using the passed in context is right here.  We may inherit RBAC restrictions from whoever initiated the change.  Possible that we should be ignoring that and doing the db operations as the driver instead.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -697,16 +695,101 @@ def get_mappings(self, agent):
...
+    def handle_qos_policy_update(self, policy_id):
+        LOG.info("HANDLE_QOS_POLICY_UPDATE: %s" % policy_id)
+
+        plugin_context = ctx.get_admin_context()
+        with self._txn_from_context(plugin_context, tag="handle_qos_policy_update"):
+            # Update the existing ports that are directly using this qos_policy_id.
+            ports = self.db.get_ports(
+                plugin_context,
+                filters={
+                    "qos_policy_id": [policy_id],
```
fasaxc: Looks like this ultimately boils down to looking at fields on the db port object, and it doesn't have this field. 

Querying the backend ports would look something like this; cribbing from the object-layer code.
```
query = plugin_context.session.query(models_v2.Port).join(
        qos_models.QosPortPolicyBinding)
query = query.filter(
        qos_models.QosPortPolicyBinding.policy_id == policy_id
)
ports = query.all()
```
write_endpoint() eventually re-loads the dict port via db.get_ports anyway, so it looks like it wouldn't be too tough to make use of the db-layer ports for this.  Of course, it might make sense to rework all the code to use the db-layer for speed and precision but that's out of scope.

fasaxc (changes requested):
Few queries but we've been discussing these on slack already...

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
nelljerram: Agreed.  I've removed the use of `port_extra.network_qos_policy_id` now.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: +1 for the accessor.  I do want to take a look at the context RBAC question soon as well - ideally switch to the decorators that the ChatGPT response suggested - but shall we separate that to another PR?

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: I think you mean that it does have the field, but doesn't support filtering on it.

I think a reasonable thing to do for this PR is to remove the bogus filtering attempt, and keep the following code that filters manually.  Then in further DB work I'll look at adding the filtering back into the query.  WDYT?

nelljerram: Although in one case that would mean querying all ports.  I'll look for an alternative to that one.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: Turns out the accessor wasn't available in Yoga, so we should keep using `_plugin_context` for now.  But we can switch to the accessor when we discard support for Yoga and move on to the next OpenStack LTS release, which will be 'Gazpacho'.

nelljerram:
```
9a483f02bebfd0f1241e40fc98eb740496f78990
Author:     Sebastian Lohff <sebastian.lohff@sap.com>
AuthorDate: Wed Nov 16 16:16:22 2022 +0100
Commit:     Sebastian Lohff <sebastian.lohff@sap.com>
CommitDate: Thu Nov 17 11:03:44 2022 +0100

Parent:     bf44e70db6 OVN: Add support for DHCP option "domain-search" for IPv4
Merged:     master
Follows:    21.0.0.0rc1 (135)
Precedes:   22.0.0.0rc1 (349)

Make MechanismDriverContext plugin_context public

MechanismDriverContext has an attribute _plugin_context, which carries
the current context with it. This is used by many ml2 drivers, as it is
the only way for them to get the current context. We now make this a
public API by adding a property to MechanismDriverContext that returns
_plugin_context as a read-only attribute.

Change-Id: If9b05655286f42081cf26c90c563429ca2e63244
```

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: Have now changed `handle_qos_policy_update` to follow more modern practice.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -697,16 +698,91 @@ def get_mappings(self, agent):
...
+                self.endpoint_syncer.write_endpoint(p, plugin_context, must_update=True)
+
+    @requires_state
+    def handle_qos_policy_update(self, context, policy_id):
+        LOG.info("HANDLE_QOS_POLICY_UPDATE: %s %s", context, policy_id)
+
+        with db_api.CONTEXT_READER.using(context):
+            policy = policy_object.QosPolicy.get_policy_obj(context, policy_id)
+
+            # Find ports whose network use this QoS policy and that don't have a
```
fasaxc, line 754: Is it the case that only one policy ever matters, spec wise?  I mean, could you inherit packet rate from the network policy and max burst from the per-port policy?

fasaxc (approved):

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: I think so, but let me double check...

nelljerram: I haven't found a 100% definitive statement, but https://docs.openstack.org/neutron/latest/admin/config-qos.html says
> You can attach networks to a QoS policy. The meaning of this is that any compute port connected to the network will use the network policy by default unless the port has a specific policy attached to it.

and it's also implied by [this coding](https://github.com/openstack/neutron/blob/master/neutron/services/qos/qos_plugin.py#L330), which is what I used as inspiration for my rewrite of `handle_qos_policy_update`.

Issue #10872, opened 2025-08-24
Nftables rules not properly reloaded


CL0Pinette:
## Expected Behavior
When nftables rules are flushed, all the rules should be reapplied properly.

## Current Behavior
When flushing nftables rules, some of the calico rules are not reapplied properly.

We can see that the expectedHashes disappear while reloading the chain
```
2025-08-24 16:29:53.414 [WARNING][89] felix/table.go 592: Detected out-of-sync Calico chain, marking for resync chainName="nat-POSTROUTING" dpHashes=[]string(nil) expectedHashes=[]string{"z057H0Z77Fcko47w"} ipVersion=0x4 table="calico"
....
2025-08-24 16:29:53.474 [DEBUG][89] felix/table.go 941: Comparing chain hashes chainName="nat-POSTROUTING" current=[]string{} ipVersion=0x4 previous=[]string(nil) table="calico"
```

It looks like it only affects the rules that are appended to the chains (filter-FORWARD, nat-POSTROUTING)
```
2025-08-24 16:25:15.652 [DEBUG][89] felix/rules.go 117: Hashed rule action=Jump->nat-cali-POSTROUTING chain="nat-POSTROUTING*appends*" hash="z057H0Z77Fcko47w" position=0 ruleFragment="counter jump nat-cali-POSTROUTING"
```

```
2025-08-24 16:25:15.653 [DEBUG][89] felix/rules.go 117: Hashed rule action=Accept chain="filter-FORWARD*appends*" hash="yp7iNUYNEhDmKF6E" position=0 ruleFragment="meta mark & 0x10000 == 0x10000 counter accept"
```

## Possible Solution
When recomptuting the currentHashes
https://github.com/projectcalico/calico/blob/cf50b562271b7c2ad896af0488d48eddabbb74eb/felix/nftables/table.go#L921
It should also take into account the fact that there may be rules in a chain named `chain + "*appends*"`: https://github.com/projectcalico/calico/blob/cf50b562271b7c2ad896af0488d48eddabbb74eb/felix/nftables/table.go#L639
## Steps to Reproduce (for bugs)
1. Create an ippool with natOutgoing set to true
2. Rollout all the calico-node to have clean rules
3. Run `nft list chain calico nat-POSTROUTING`, there should be a `jump nat-cali-POSTROUTING`
4. Run nft flush ruleset
5. Wait for the rules to be reapplied
6. Run `nft list chain calico nat-POSTROUTING`, the chain is now empty
[...]

PR #10936, opened 2025-09-04
nftables: properly mark base chains as dirty after a flush

caseydavenport:
## Description

This fixes an issue where, in nftables mode, after an `nft flush ruleset`, Calico would fail to reprogram its top level jump rules into its nftables base chains.

This was occurring because the nftables Table structure uses two separate dirty trackers: `dirtyChains` and `dirtyBaseChains`.

The separate trackers are used because the logic within the dataplane `apply` is different. But since the base chains were being marked as dirty in `dirtyChains` and not `dirtyBaseChains`, the correct logic was not being executed for these chains.

## Related issues/PRs

Fixes https://github.com/projectcalico/calico/issues/10872#issuecomment-3241031834

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
nftables: fix reprogramming of base chain rules after out-of-band flush.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/nftables/table.go
+++ b/felix/nftables/table.go
@@ -489,7 +489,7 @@ func (t *NftablesTable) UpdateChain(chain *generictables.Chain) {
 	numRulesDelta := len(chain.Rules) - oldNumRules
 	t.gaugeNumRules.Add(float64(numRulesDelta))
 	if t.chainIsReferenced(chain.Name) {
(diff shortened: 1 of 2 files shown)


marvin-tigera:

PR #10941, opened 2025-09-05
Check if something else happened to break Tempest tests

nelljerram:
I've just started seeing Tempest failure in #10931.  This PR is to explore if it's equally broken in the status quo.


nelljerram:
No, there was no other cause of the breakage.  I think the problem was [this](https://github.com/projectcalico/calico/pull/10931/commits/a28157897b31c9c8a8d897c10d6ad220adfef500).

PR #10946, opened 2025-09-05
fix(build): Correctly pass GOFLAGS to docker run

rene-dekker:
## Description

The GOFLAGS environment variable was being passed to the docker run command in a way that caused the shell to misinterpret the arguments, specifically when GOFLAGS was empty and followed by another argument like -mod=readonly. This resulted in docker run receiving an invalid flag.

This change quotes the GOFLAGS assignment to ensure it's always treated as a single argument, preventing the build from failing.


stevegaossou (approved):

marvin-tigera:
OK, I will merge the pull request when it's ready, leave the commits as is when I merge it, and delete the branch after I've merged it.

PR #10954, opened 2025-09-08
Override with HOSTNAME env var if set when  hostname-override parameter specified in kubelet

DrAuYueng:
## Description
When kubelet specifies a hostname using the [hostname-override](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) parameter, for example: `--hostname-override=192.168.0.10`, the result will be different from the result obtained by the `hostname` command.   
This means that the endpointHostnameFilter will filter out node-related event processing. 
In this scenario, the hostname needs to be explicitly defined using the environment variable `HOSTNAME`. 
In the calico-node daemonset yaml file, `HOSTNAME` can be defined using the following configuration:
```
        - name: HOSTNAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
```

## Todos

- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/config/config_params.go
+++ b/felix/config/config_params.go
@@ -715,11 +715,17 @@ func (config *Config) applyDefaults() {
 	for _, param := range knownParams {
 		param.setDefault(config)
 	}
+	// Default hostname
 	hostname, err := names.Hostname()
 	if err != nil {
-		log.Warningf("Failed to get hostname from kernel, "+
-			"trying HOSTNAME variable: %v", err)
+		log.Warningf("Failed to get hostname from kernel: %v", err)
+	}
+	// Override with HOSTNAME env var if set.
+	if os.Getenv("HOSTNAME") != "" {
 		hostname = strings.ToLower(os.Getenv("HOSTNAME"))
+	}
+	if hostname == "" {
+		log.Warningf("Failed to get hostname from kernel or HOSTNAME env variable")
 	}
 	config.FelixHostname = hostname
 }


DrAuYueng:
The CI failure seems to be unrelated to the current PR. Is there something I missed?

mazdakn:
@DrAuYueng no, it's not related to this PR. it's a known issue, and we are working to fix it.
I'm going to review the PR tomorrow.

DrAuYueng:
The hostname can be specified through the environment variable `FELIX_FELIXHOSTNAME`, configured as follows:
```
- name: FELIX_FELIXHOSTNAME
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: spec.nodeName
```
, but I hope that automatic detection and processing can be done by default, after all, [cni-plugin](https://github.com/projectcalico/calico/blob/0f5ebb89ccc85c1f3e2d2f092511e79ffe12579a/cni-plugin/pkg/plugin/plugin.go#L211) and calico node [startup](https://github.com/projectcalico/calico/blob/0f5ebb89ccc85c1f3e2d2f092511e79ffe12579a/node/pkg/lifecycle/startup/startup.go#L111) both do similar processing.

mazdakn:
@DrAuYueng I do not understand the motivation here. Felix is a process inside Calico node, and it is node's responsibility to auto detect hostname and pass it to Felix (which seems it does).

PR #10969, opened 2025-09-09
[BPF] fix deltatracker batch iteration

tomastigera:
Wrong slice math in PendingUpdatesView [:batchSize-applied] vs [applied:]

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/deltatracker/delta_tracker.go
+++ b/felix/deltatracker/delta_tracker.go
@@ -457,9 +457,14 @@ func (c *PendingUpdatesView[K, V]) IterBatched(applyFn func(k []K, v []V) (int, 
 				applied++ // skip over the item that erred
 			}
 
-			ks = ks[applied:]
-			vs = vs[applied:]
 			count = batchSize - applied
+			if count == 0 {
+				ks = make([]K, 0, batchSize)
+				vs = make([]V, 0, batchSize)
+			} else {
+				ks = ks[applied:]
+				vs = vs[applied:]
+			}
 		}
 	}
 
@@ -474,8 +479,8 @@ func (c *PendingUpdatesView[K, V]) IterBatched(applyFn func(k []K, v []V) (int, 
 			applied++ // skip over the item that erred
 		}
 
-		ks = ks[:batchSize-applied]
-		vs = vs[:batchSize-applied]
+		ks = ks[applied:]
+		vs = vs[applied:]
 		count -= applied
 	}
 }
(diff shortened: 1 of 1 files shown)


felix/deltatracker/delta_tracker.go
```diff
@@ -536,8 +541,12 @@ func (c *PendingDeletionsView[K, V]) IterBatched(applyFn func(k []K) (int, error
 				applied++ // skip over the item that erred
 			}
 
-			ks = ks[applied:]
-			count = batchSize - applied
+			count -= applied
+			if count == 0 {
+				ks = make([]K, 0, batchSize)
```
fasaxc:
```suggestion
				ks = ks[:0]
```
Seems like you could re-use the slice?

felix/deltatracker/delta_tracker.go
```diff
@@ -457,9 +457,14 @@ func (c *PendingUpdatesView[K, V]) IterBatched(applyFn func(k []K, v []V) (int,
 				applied++ // skip over the item that erred
 			}
 
-			ks = ks[applied:]
-			vs = vs[applied:]
 			count = batchSize - applied
+			if count == 0 {
+				ks = make([]K, 0, batchSize)
```
fasaxc: Can you re-use the slices `ks = ks[:0]`?

felix/deltatracker/delta_tracker.go
```diff
@@ -536,8 +541,12 @@ func (c *PendingDeletionsView[K, V]) IterBatched(applyFn func(k []K) (int, error
 				applied++ // skip over the item that erred
```
fasaxc, line 553: This is a pretty surprising contract for `applyFn`; I would have expected that returning an error meant "batch failed" and the state was then unknown.

Worth documenting the precise expectation and the fact that the first return argument is used even if err != nil.

felix/deltatracker/delta_tracker.go
```diff
@@ -457,9 +457,14 @@ func (c *PendingUpdatesView[K, V]) IterBatched(applyFn func(k []K, v []V) (int,
...
 			}
 
-			ks = ks[applied:]
-			vs = vs[applied:]
 			count = batchSize - applied
```
fasaxc, line 467: I'd be tempted to get rid of count and use `len(ks)` and `if applied == len(ks)` for the condition.  One less thing to keep track of.

felix/deltatracker/delta_tracker.go
tomastigera: ok, do it now when we apply the whole batch, should be the common case. I don't think it pays off when we have many partial updates. We cold also just consume the batch and not add new elements.

PR #10974, opened 2025-09-10
iptables: Force-program chains for AssumeNeededOnEveryNode policies

fasaxc:
## Description

Previously, we'd thought that pre-programming iptables rules was a waste but it does seem to have some benefit.

- Pass the hints down to the dataplane.
- For the for filter table only (since it's the mainline and it's generally hard to tell in advance which table is needed); set a flag on marked policies.
- In iptables layer, incref policies with the flag so that they get programmed.

## Related issues/PRs

CORE-11815

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
When the AssumeNeededOnEveryNode performance hint is set on a policy, the iptables dataplane now renders the policy rules to the filter table, even if the chain is not yet referenced; this reduces the work needed when a new workload arrives needing the policy.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/iptables_force_programming_test.go
+++ b/felix/fv/iptables_force_programming_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2023 Tigera, Inc. All rights reserved.
+// Copyright (c) 2025 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/felix/iptables/table.go
+++ b/felix/iptables/table.go
@@ -540,14 +540,13 @@ func (t *Table) UpdateChain(chain *generictables.Chain) {
 	t.logCxt.WithField("chainName", chain.Name).Debug("Adding chain to available set.")
 	oldNumRules := 0
 
+	// Incref any newly-referenced chains, then decref the old ones.  By incrementing first we
+	// avoid marking a still-referenced chain as dirty.
 	if chain.ForceProgramming {
 		// A force-programmed chain refers to itself.
 		t.logCxt.WithField("chainName", chain.Name).Debug("Chain has force programming flag, incref.")
 		t.increfChain(chain.Name)
 	}
-
-	// Incref any newly-referenced chains, then decref the old ones.  By incrementing first we
-	// avoid marking a still-referenced chain as dirty.
 	t.maybeIncrefReferredChains(chain.Name, chain.Rules)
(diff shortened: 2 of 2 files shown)


felix/iptables/table.go
```diff
@@ -540,14 +540,25 @@ func (t *Table) UpdateChain(chain *generictables.Chain) {
 	t.logCxt.WithField("chainName", chain.Name).Debug("Adding chain to available set.")
 	oldNumRules := 0
 
+	if chain.ForceProgramming {
```
nelljerram, line 545: nit: slightly better for this block to come after the following comment "Incref any newly-..."

felix/fv/iptables_force_programming_test.go
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) 2023 Tigera, Inc. All rights reserved.
```
nelljerram: 2025

felix/fv/iptables_force_programming_test.go
```diff
@@ -0,0 +1,117 @@
...
+		tc     infrastructure.TopologyContainers
+		infra  infrastructure.DatastoreInfra
+		client client.Interface
+		w      [3]*workload.Workload
+		cc     *connectivity.Checker
+	)
+
+	BeforeEach(func() {
+		if NFTMode() {
+			Skip("This test is not yet supported in NFT mode")
```
nelljerram, line 45: I wondered if the hint might cause problems for NFT mode.  Do we guard higher up against passing the hint down in NFT mode, e.g. from the calc graph?

nelljerram: (In the non-test Felix code, I mean.)

nelljerram (approved):
I've noted a couple of nits, but this looks good, so long as we're happy to change the semantics of the existing hint.  You talked earlier about introducing a new hint instead - did you decide that that wasn't needed?

felix/fv/iptables_force_programming_test.go
fasaxc: No, it simply won't touch that field AFAICT.  Might make sense to implement for nft too but I thoght I'd start with iptables.

nelljerram: I was worried that the change in policy_mgr.go (which copies the chains with ForceProgramming added) might accidentally affect NFT as well.  But I see that it doesn't.

PR #10975, opened 2025-09-10
ServiceIP failover UTs (v4 & v6) and more.

aaaaaaaalex:
## Description

- ADDS "borrowed" JHASH file
- Set a Maglev flag on conntracks, use it to allow overwriting tunnel IPs on backing nodes.
- Rename test global consistentHashMap->maglevMap

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/conntrack.h
+++ b/felix/bpf-gpl/conntrack.h
@@ -794,11 +794,6 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_lookup(struct cali_tc_c
 			if (result.flags & CALI_CT_FLAG_MAGLEV) {
 				CALI_DEBUG("overwriting tunnel src for Maglev packet");
 				tracking_v->tun_ip = ctx->state->tun_ip;
-				int err = 0;
-				err = cali_ct_update_elem(&v->nat_rev_key, tracking_v, BPF_EXIST);
-				if (err) {
-					CALI_CT_DEBUG("Failed to update tunnel src to " IP_FMT, debug_ip(ctx->state->tun_ip));
-				}
 			}
 		}
 


felix/bpf-gpl/conntrack.h
```diff
@@ -787,8 +787,19 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_lookup(struct cali_tc_c
 		if (CALI_F_FROM_HEP && !ip_void(ctx->state->tun_ip) && !ip_void(result.tun_ip) &&
 				!ip_equal(result.tun_ip, ctx->state->tun_ip)) {
 			CALI_CT_DEBUG("tunnel src changed from " IP_FMT " to " IP_FMT "",
-					debug_ip(result.tun_ip), debug_ip(ctx->state->tun_ip));
+				debug_ip(result.tun_ip), debug_ip(ctx->state->tun_ip));
```
tomastigera, line 790: the .c and .h files use tabs

felix/bpf-gpl/conntrack.h
```diff
@@ -787,8 +787,19 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_lookup(struct cali_tc_c
...
 			ct_result_set_flag(result.rc, CT_RES_TUN_SRC_CHANGED);
+
+			if (result.flags & CALI_CT_FLAG_MAGLEV) {
+				CALI_DEBUG("overwriting tunnel src for Maglev packet");
+				tracking_v->tun_ip = ctx->state->tun_ip;
+				int err = 0;
+				err = cali_ct_update_elem(&v->nat_rev_key, tracking_v, BPF_EXIST);
+				if (err) {
+					CALI_CT_DEBUG("Failed to update tunnel src to " IP_FMT, debug_ip(ctx->state->tun_ip));
+				}
```
tomastigera: you do not need that, the value was already updated on line 796

felix/bpf-gpl/tc.c
```diff
@@ -358,10 +358,14 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
...
-		CALI_DEBUG("dropping tunnel pkt with changed source node");
-		goto deny;
+	if (!(ctx->state->ct_result.flags & CALI_CT_FLAG_MAGLEV)) {
+		/* Check if someone is trying to spoof a tunnel packet */
+		if (CALI_F_FROM_HEP && ct_result_tun_src_changed(ctx->state->ct_result.rc)) {
+			CALI_DEBUG("dropping tunnel pkt with changed source node");
+			goto deny;
+		}
+	} else {
+		CALI_DEBUG("Keeping Maglev packet with changed source node");
```
```suggestion
		CALI_DEBUG("Maglev: tunnel source node changed to " IP_FMT , ...);
```

felix/bpf/ut/maglev_test.go
```diff
@@ -103,136 +104,85 @@ func TestConsistentHashTable(t *testing.T) {
...
+	node2WlCIDR := net.IPNet{
+		IP:   podIP,
+		Mask: net.IPv4Mask(255, 255, 255, 0),
+	}
+	vni := uint32(0)
+	var encapedPkt []byte
+
+	// Prepare a test packet.
+	// SYN, Src IP 1.1.1.1, Src Port 1234, Dst IP 172.16.1.1, Dst Port 5678.
+	// Though we are using NodePorts, thus changing the 5-tuple of the request when failing-over,
```
tomastigera: The 5 tuple must remain the same for the fail over. You need to program the NAT table with an address of _a_ service. The IP should be different than the IP of the node. We need to emulate the real world scenario.

felix/bpf/ut/maglev_test.go
```diff
@@ -580,33 +513,80 @@ func TestMaglevTCPFailover(t *testing.T) {
...
-	hostIP = node1ip
+	// Now on node 1, handling a failing-over connection.
+	newNode(node1ip)
+	encapedPkt = nil
 
-	// change to routing again to a remote workload
-	resetRTMap(rtMap)
-	restoreRTMap(rtMap, rtNode1)
-	dumpRTMap(rtMap)
+	// SYN, Src IP 1.1.1.1, Src Port 1234, Dst IP 172.16.1.1, Dst Port 5678.
```
tomastigera:
```suggestion
	// Src IP 1.1.1.1, Src Port 1234, Dst IP 172.16.1.1, Dst Port 5678.
```

felix/bpf/ut/maglev_test.go
```diff
@@ -580,33 +513,80 @@ func TestMaglevTCPFailover(t *testing.T) {
...
-	skbMark = 0
+	// Node 1: Flagging frontend map item with the consistent-hash flag.
+	err = natMap.Update(
+		nat.NewNATKey(ipv4.DstIP, uint16(tcp.DstPort), uint8(layers.IPProtocolTCP)).AsBytes(),
+		nat.NewNATValueWithFlags(0, 1, 0, 0, nat.NATFlgConsistentHash).AsBytes(),
+	)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Node 1: Build a maglev LUT and program each item to the BPF map.
+	mglv = consistenthash.New(consistenthash.WithHash(fnv.New32(), fnv.New32()), consistenthash.WithPreferenceLength(31))
```
tomastigera, line 479: we usually can store and restore the maps so we do not need to regenerate them, not sure if all maps can do that

felix/bpf/ut/maglev_test.go
```diff
@@ -580,33 +513,80 @@ func TestMaglevTCPFailover(t *testing.T) {
...
+		routes.NewKey(ip.CIDRFromIPNet(&node1CIDR).(ip.V4CIDR)).AsBytes(),
+		routes.NewValue(routes.FlagsLocalHost).AsBytes(),
+	)
+	Expect(err).NotTo(HaveOccurred())
+	err = rtMap.Update(
+		routes.NewKey(ip.CIDRFromIPNet(&node2CIDR).(ip.V4CIDR)).AsBytes(),
+		routes.NewValue(routes.FlagsRemoteHost).AsBytes(),
+	)
+	Expect(err).NotTo(HaveOccurred())
+	dumpRTMap(rtMap)
```
tomastigera, line 516: This was a special case to test in NP, we do not need to repeat it here. I think you already test it on node3

felix/bpf/ut/maglev_test.go
```diff
@@ -615,268 +595,1063 @@ func TestMaglevTCPFailover(t *testing.T) {
...
+	dumpARPMap(arpMap)
+	arpNode2 = saveARPMap(arpMap)
+	Expect(arpNode2).To(HaveLen(2))
+	arpKey = arp.NewKey(node1ip, 1 /* ifindex is always 1 in UT */)
+	Expect(arpNode2).To(HaveKey(arpKey))
+	macDst = encapedPkt[0:6]
+	macSrc = encapedPkt[6:12]
+	Expect(arpNode2[arpKey]).To(Equal(arp.NewValue(macDst, macSrc)))
 
+	// try a spoofed tunnel packet, should be dropped and have no effect
```
tomastigera, line 684: that should not longer apply with maglev, right?

felix/bpf/ut/maglev_test.go
```diff
@@ -615,268 +595,1063 @@ func TestMaglevTCPFailover(t *testing.T) {
...
+
+	restoreCTMap(ctMap, ctNode2)
+	dumpCTMap(ctMap)
+
+	// now we are at the node with local workload
+	err = natMap.Update(
+		nat.NewNATKey(ipv4.DstIP, uint16(tcp.DstPort), uint8(ipv4.Protocol)).AsBytes(),
+		nat.NewNATValueWithFlags(0 /* id */, 1 /* count */, 1 /* local */, 0, nat.NATFlgConsistentHash).AsBytes(),
+	)
+	Expect(err).NotTo(HaveOccurred())
```
tomastigera: no need to test again

felix/bpf/ut/maglev_test.go
```diff
@@ -615,268 +595,1063 @@ func TestMaglevTCPFailover(t *testing.T) {
...
+
+		Expect(udpR.SrcPort).To(BeEquivalentTo(tcp.SrcPort ^ tcp.DstPort))
+		vxlanSrcPort = udpR.SrcPort
+
+		checkVxlanEncap(pktR, false, ipv4, tcp, payload)
+	})
+
+	skbMark = 0
+	// Another pkt with a different source port arriving at node 1.
+	// This time, not a SYN, i.e. a failed-over connection.
```
tomastigera: not true anymore, right?

felix/bpf/ut/maglev_test.go
```diff
@@ -615,268 +595,1063 @@ func TestMaglevTCPFailover(t *testing.T) {
...
+			ipv4R := ipv4L.(*layers.IPv4)
+			Expect(ipv4R.SrcIP.String()).To(Equal(node2ip.String()))
+			Expect(ipv4R.DstIP.String()).To(Equal(node1ip.String()))
+
+			checkVxlan(pktR)
+		}, withHostNetworked())
+	}
+}
+
+func TestMaglevNATServiceIPTCPV6(t *testing.T) {
```
tomastigera, line 911: I do not think we need to test this. It is the same code, we test np in v6 (primarily for dev) and it is going to be excercised in FVs

felix/bpf/ut/maglev_test.go
```diff
@@ -615,268 +595,1063 @@ func TestMaglevTCPFailover(t *testing.T) {
...
+			}))
+
+			ipv4L := pktR.Layer(layers.LayerTypeIPv4)
+			Expect(ipv4L).NotTo(BeNil())
+			ipv4R := ipv4L.(*layers.IPv4)
+			Expect(ipv4R.SrcIP.String()).To(Equal(node2ip.String()))
+			Expect(ipv4R.DstIP.String()).To(Equal(node1ip.String()))
+
+			checkVxlan(pktR)
+		}, withHostNetworked())
```
tomastigera: I don't think we need to replicate this. It is based on the NP code and should be potentially exercised by FVs

felix/bpf-gpl/conntrack_types.h
```diff
@@ -39,6 +39,7 @@ enum cali_ct_type {
 #define CALI_CT_FLAG_NP_REMOTE	0x1000 /* marks connections from local host to remote backend of a nodeport */
 #define CALI_CT_FLAG_NP_NO_DSR	0x2000 /* marks connections from a client which is excluded from DSR */
 #define CALI_CT_FLAG_SKIP_REDIR_PEER	0x4000 /* marks connections from a client which is excluded from redir */
+#define CALI_CT_FLAG_MAGLEV	0X8000 /* marks connections from a maglev LB. On a backing node, allows packets of an existing to arrive via a different tunnel for failovers.
```
tomastigera:
```suggestion
#define CALI_CT_FLAG_MAGLEV	0X8000 /* marks connections from a maglev LB. On a backing node, allows packets of an existing to arrive via a different tunnel for failovers. */
```

felix/bpf-gpl/conntrack.h
aaaaaaaalex: bah, VSCode keeps trying to be clever. I need to find out why it's doing that.

felix/bpf-gpl/conntrack.h
aaaaaaaalex: Hmm, not sure I understand.
The map update is not necessary?

felix/bpf/ut/maglev_test.go
aaaaaaaalex: Sorry - comment is stale. We're using a cluster IP now

felix/bpf/ut/maglev_test.go
aaaaaaaalex: To be fair, in the case of maglev I don't even think we need to mess with the map at all, since it should be the same on every node. This is probably a case of over-diligence on my behalf :D

felix/bpf/ut/maglev_test.go
aaaaaaaalex: I think the DENY here comes from a lack of a remote-host-route, rather than the route being different than before, so it makes sense to me.

felix/bpf/ut/maglev_test.go
aaaaaaaalex: Yep I think this block is redundant now, removing.

felix/bpf/ut/maglev_test.go
aaaaaaaalex: Removed one of these unknown VNI blocks - now testing it once for V4 and once for V6

felix/bpf/ut/maglev_test.go
aaaaaaaalex: This test models the failover case - maybe the name is misleading?

felix/bpf/ut/maglev_test.go
aaaaaaaalex: The whole highlighted block?
~~I found some good bugs by using this type of code to model failover - IMO it's valuable if not very similar (and already written :sweat_smile: )~~ OOPS, this is the hostnetwork test block - Ignore me!

felix/bpf/ut/maglev_test.go
aaaaaaaalex: Update: removed them all - they really are not very relevant!

felix/bpf-gpl/conntrack_types.h
aaaaaaaalex: HEAD should now compile - sorry!
Re: port collision - the original "Source collision" error I got went away when I migrated the test from NodePorts to ServiceIP. At the root, it seems to be because in the NP tests, the program wanted to call `conntrack_create`, but didn't in the ClusterIP version. My hypothesis was that the NP connection was considered NEW due to the differing NP IP's, but I didn't dig deeper than that.

What does seem to be the case now, though, is that once the tunnel-IP is updated on the conntrack (in the clusterIP-test), the conntracks are identical. I think this means that we can no longer detect a simultaneous connect from the same source IP without considering it a failover. I think that's acceptable given the nature of Maglev.

felix/bpf-gpl/jhash.h
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * # MSG FROM ALEX:
+ * > I took this function from linux/tools/include/linux/bitops.h
+ * > instead of including the whole dependency.
+ * > bitops.h has: SPDX-License-Identifier: GPL-2.0
+ */
+static CALI_BPF_INLINE __u32 rol32(__u32 word, unsigned int shift)
+{
+	return (word << shift) | (word >> ((-shift) & 31));
+}
```
fasaxc: Can we use the one from Bob's original:
```
#define rot(x,k) (((x)<<(k)) | ((x)>>(32-(k))))
```
Currently, the bpf programs are dual licensed apache/GPL so that we have no worries about GPL affecting felix.  We can do that because we wrote them.  If you add external GPL code, that changes.

felix/bpf-gpl/conntrack.h
fasaxc: Guessing Tomas means: when you load a value from a map, you get a pointer to the value, so, if you change things via that pointer, it's changed.  You don't need to write it back to the map as well.

aaaaaaaalex: oh! I've obviously been coddled by Go - I just expected it'd be a copied value

aaaaaaaalex: And sure enough - it passes without the call to update_elem!

aaaaaaaalex: Any risk of synchronization / concurrency problems when directly reaching into memory to modify?

PR #10980, opened 2025-09-10
align hashrelease with release artifacts

radTuti:
## Description

This updates hashrelease to include the calicoctl binaries and release tar (similar to releases).

Additionally, added wrapping for errors.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


release/pkg/manager/calico/manager.go
```diff
@@ -920,66 +936,91 @@ func (r *CalicoManager) uploadDir() string {
...
-		fmt.Sprintf("%s/cni:%s", registry, ver):                          filepath.Join(imgDir, "calico-cni.tar"),
-		fmt.Sprintf("%s/kube-controllers:%s", registry, ver):             filepath.Join(imgDir, "calico-kube-controllers.tar"),
-		fmt.Sprintf("%s/pod2daemon-flexvol:%s", registry, ver):           filepath.Join(imgDir, "calico-pod2daemon.tar"),
-		fmt.Sprintf("%s/dikastes:%s", registry, ver):                     filepath.Join(imgDir, "calico-dikastes.tar"),
-		fmt.Sprintf("%s/flannel-migration-controller:%s", registry, ver): filepath.Join(imgDir, "calico-flannel-migration-controller.tar"),
-	}
-	for img, out := range images {
-		err = r.archiveContainerImage(out, img)
+func (r *CalicoManager) buildReleaseTar() error {
+	ver := r.calicoVersion
```
caseydavenport: a nit, but better to just use the `r.calicoVersion` where needed instead of redefining a local variable (looks like you did this to keep the name aligned with the now-removed argument to this function). This just makes the code a tiny bit harder to follow end-to-end.

release/pkg/manager/calico/manager.go
```diff
@@ -920,66 +936,91 @@ func (r *CalicoManager) uploadDir() string {
...
 	// Add in release binaries that we ship.
 	binDir := filepath.Join(releaseBase, "bin")
-	err = os.MkdirAll(binDir, os.ModePerm)
-	if err != nil {
+	if err := os.MkdirAll(binDir, os.ModePerm); err != nil {
 		return fmt.Errorf("failed to create images dir: %s", err)
 	}
 
 	binaries := map[string]string{
-		// CNI plugin binaries are all placed in github dir.
```
caseydavenport, line 956: I don't quite follow why these are only included if `r.buildImages` is set now. Not sure it's a problem really, but is there a reason for it?

release/pkg/manager/calico/manager.go
```diff
@@ -920,66 +936,91 @@ func (r *CalicoManager) uploadDir() string {
...
 	// Add in manifests directory generated from the docs.
 	if _, err := r.runner.RunInDir(r.repoRoot, "cp", []string{"-r", "manifests", releaseBase}, nil); err != nil {
-		return err
+		return fmt.Errorf("failed to copy manifests: %w", err)
 	}
 
 	// tar up the whole thing, and copy it to the target directory
-	if _, err := r.runner.RunInDir(r.repoRoot, "tar", []string{"-czvf", fmt.Sprintf("release/_output/release-%s.tgz", ver), "-C", "release/_output", fmt.Sprintf("release-%s", ver)}, nil); err != nil {
-		return err
+	if _, err := r.runner.RunInDir(r.repoRoot, "tar", []string{"-czvf", releaseTarFilePath, "-C", r.tmpDir, fmt.Sprintf("release-%s", ver)}, nil); err != nil {
```
caseydavenport: Have you double checked that the output from this `tar` command matches the structure / naming of the existing one when un-tar'd?

caseydavenport: I am a little bit worried about using `tmpDir` for this - that directory up until now has been "where we put any temporary files that are side-effects of the build / release".

That seems like the wrong place to put actual, real release artifacts that we are packaging up to ship?

release/pkg/manager/calico/manager.go
```diff
@@ -920,66 +936,91 @@ func (r *CalicoManager) uploadDir() string {
...
+	env := append(os.Environ(),
+		fmt.Sprintf("VERSION=%s", r.calicoVersion),
+	)
+	dir := "calicoctl"
+	out, err := r.makeInDirectoryWithOutput(filepath.Join(r.repoRoot, dir), "build-all", env...)
+	if err != nil {
+		logrus.Error(out)
+		return fmt.Errorf("Failed to build %s: %w", dir, err)
 	}
+	logrus.Info(out)
```
caseydavenport: Not sure we need this Info level log on success - maybe debug is OK? Ideally we'd have debug level logging within `makeInDirectoryWithOutput` so individual callers wouldn't need to log this out.

release/pkg/manager/calico/manager.go
```diff
@@ -1188,20 +1233,33 @@ func (r *CalicoManager) assertManifestVersions(ver string) error {
...
 	}
-	return strings.TrimSpace(out)
+	return strings.TrimSpace(out), nil
 }
 
 // Uses docker to build a tgz archive of the specified container image.
 func (r *CalicoManager) archiveContainerImage(out, image string) error {
+	if !r.buildImages {
+		if _, err := r.runner.Run("docker", []string{"image", "inspect", image}, nil); err != nil {
+			logrus.WithError(err).WithField("image", image).Error("Image not found locally, will attempt to pull")
```
caseydavenport, line 1255: What use-case is this for? 

Call me paranoid, but I fear that having code to pull images might result in us accidentally ending up with the wrong code included in our archives if we're not careful.

radTuti: This is for hashreleases that do not build images in the hashrelease run.
I will update it to only attempt to pull images if it is a hashrelease with `BUILD_CONTAINER_IMAGES=false`

caseydavenport: I guess my point is why do we want to build this image archive for hashreleases with BUILD_CONTAINER_IMAGES=false?

release/pkg/manager/calico/manager.go
radTuti: yes it does match.
also I have switched it from using `tmpDir` to parent directory of the upload dir.

- hashrelease will use `release/_output/hashrelease`
- release will use `release/_output/upload`

release/pkg/manager/calico/manager.go
radTuti: I think it will be good to have content of hashreleases be the same regardless of how it is built.
At the same time, I am also happy to remove this and only have it include image archives when the release tool is set to builds images.

release/pkg/manager/calico/manager.go
radTuti: I have updated it to build all the release binaries.

caseydavenport: That's fine, although an answer to my question would have sufficed before making code changes :grin:

cni-plugin/Makefile
```diff
@@ -197,6 +197,7 @@ $(WINDOWS_BIN)/flannel.exe: $(FLANNEL_CNI_PLUGIN_CLONED)
 .PHONY: build-cni-bins build-win-cni-bins
 build-cni-bins: $(BIN)/flannel $(BIN)/loopback $(BIN)/host-local $(BIN)/portmap $(BIN)/tuning
 build-win-cni-bins: $(WINDOWS_BIN)/flannel.exe
+build-bins: build-cni-bins build-win-cni-bins
```
caseydavenport: There's already a build-all target that should do this. We have standardized "root" targets for building images and binaries so we shouldn't need to introduce a new one.

release/pkg/manager/calico/manager.go
```diff
@@ -963,22 +983,46 @@ func (r *CalicoManager) buildReleaseTar(ver string, targetDir string) error {
...
-		return err
+	if _, err := r.runner.RunInDir(r.repoRoot, "cp", []string{releaseTarFilePath, r.uploadDir()}, nil); err != nil {
+		return fmt.Errorf("failed to copy release tar: %w", err)
+	}
+	return nil
+}
+
+func (r *CalicoManager) buildBinaries() error {
+	// Skip building binaries if we are building images
+	// binaries are built as part of "release-build" target.
```
caseydavenport, line 1008: If I understand the main goal of this PR:

- We want to make the hashrelease artifacts identical to the real release artifacts.
- Real releases include calicoctl. HRs don't, and don't always build calicoctl because they can use pre-built images.
- So, add an explicit step to build the binaries.

The goal of this PR is to align hashrelease and release artifacts, but it should also be a goal that the way in which these artifacts are generated is identical between a release and a hashrelease.

I think this is... just about ok for now. But I'd really like to see this logic around using `if r.buildImages` to determine if we call these make targets go away. We want to get as close to a single path through this code as possible, rather than introduce different ways that hashreleases / releases build the same code.

The hashrelease should probably just call the release build target if they need the output from that target, and not try to be clever and call one of the "internal" targets that aren't meant to be called directly as part of a release.

release/pkg/manager/calico/manager.go
caseydavenport: > I think it will be good to have content of hashreleases be the same regardless of how it is built

I can see the logic in this. But I think in general part of the goal of this work was always to align not just the outputs, but the processes by which those outputs were generated.

I'm happy enough that we only do this during a hashrelease now, so let's leave it like you have it.

PR #10991, opened 2025-09-11
Retry connection attempts once in e2e

caseydavenport:
## Description

Have noticed some connection flakes. Our wget commands do inlucde a timeout that should retry, but it doesn't handle scenarios like this:

1. Create pods
1. Connectivity works
1. Create default deny policy 
1. Immediately check it is implemented.

Specifically, between step (3) and (4), there is a timing window where Felix may not have received the new policy update from the apiserver before we make the connection check. This PR uses a single retry after two seconds to handle this, which should be enough in most functioning clusters.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/e2e/pkg/utils/conncheck/conncheck.go
+++ b/e2e/pkg/utils/conncheck/conncheck.go
@@ -360,10 +360,7 @@ func (c *connectionTester) runConnection(exp *Expectation, results chan<- connec
 	// especially on CPU constrained environments such as CI, there can be a delay between making changes (e.g., applying a NetworkPolicy) and
 	// those changes taking effect. So a short retry loop is helpful.
 	timeout := time.After(10 * time.Second)
-	for {
-		if result == exp.ExpectedResult {
-			break
-		}
+	for result != exp.ExpectedResult {
 
 		select {
 		case <-timeout:


fasaxc:
In Felix FVs I think our default is to retry for up to 10s, BPF mode in particular can be a little slow to update policy on a CPU-constrained host due to waiting on the verifier.  You can have scenarios where something has just triggered a reprogram as you update the policy, then you have to wait for that to finish and then wait for your policy update to be applied.  When you run in kind, everything is sharing CPU resource, so only one program can be verified at a time per core.

fasaxc:
CI run timed out after 3 hours? I kicked off a rebuild.

e2e/pkg/utils/conncheck/conncheck.go
```diff
@@ -345,26 +345,49 @@ func (c *connectionTester) runConnection(exp *Expectation, results chan<- connec
...
+
+	// Retry up to 10 seconds if we didn't get the expected result. This helps to avoid flakes due to transient issues.
+	// We don't want to retry too many times, as that could mask real issues and slow down tests. However, we know that
+	// especially on CPU constrained environments such as CI, there can be a delay between making changes (e.g., applying a NetworkPolicy) and
+	// those changes taking effect. So a short retry loop is helpful.
+	timeout := time.After(10 * time.Second)
+	for {
+		if result == exp.ExpectedResult {
+			break
+		}
```
nelljerram: nit: you could write that more compactly as
```
for result != exp.ExpectedResult {
```

e2e/pkg/utils/conncheck/conncheck.go
```diff
@@ -386,19 +409,29 @@ func (c *connectionTester) Connect(client *Client, target Target) (string, error
...
+	out, _, err := c.execCommandInPod(c.clients[client.ID()].pod, cmd)
+	return out, err
+}
+
+func (c *connectionTester) execCommandInPod(pod *v1.Pod, cmd string) (string, ResultStatus, error) {
 	var out string
 	var err error
 
 	// Ensure the kubectl command executes in the remote cluster if required. Remote framework objects do not control
 	// how kubeconfigs are resolved by the kubectl command, so we must use this wrapper. See NewDefaultFrameworkForRemoteCluster.
```
nelljerram, line 418: I appreciate you haven't changed this, but FWIW I can't understand this comment.
- In what way is the cluster "remote", or IOW how is a "remote" cluster different from a regular one?
- What is a "remote framework object" in this area of code?
- What does it mean to "resolve" a kubeconfig?  e.g. Does it mean selecting the correct context from it?

e2e/pkg/utils/conncheck/conncheck.go
```diff
@@ -386,19 +409,29 @@ func (c *connectionTester) Connect(client *Client, target Target) (string, error
...
 	remotecluster.RemoteFrameworkAwareExec(c.f, func() {
 		if windows.ClusterIsWindows() {
```
nelljerram, line 420: Again, not your change, but `ClusterIsWindows` isn't really right, because Windows-ness should be scoped to nodes, not to the whole cluster.  Of course if an entire cluster was made of Windows nodes, this would be OK, but I'm pretty sure we have e2e setups that are mixed Linux / Windows.

nelljerram (approved):
Looks good, but a few things here that might be further improved.

e2e/pkg/utils/conncheck/conncheck.go
caseydavenport: To be honest, I am not sure. I wasn't really involved in the remote cluster e2e work and I've just left this intact!

e2e/pkg/utils/conncheck/conncheck.go
caseydavenport: Same here really, I agree this doesn't quite add up but it's how the logic has looked for some time!

PR #11007, opened 2025-09-15
Adds a Failover FV for Maglev

aaaaaaaalex:
## Description

- [x] Cluster IP Failover test
- [x] External IP  Failover test
- [x] "ConsistentHash" -> "Maglev"

## Related issues/PRs
https://github.com/projectcalico/calico/pull/10904 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


felix/fv/bpf_test.go
```diff
@@ -1916,6 +1916,297 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+							testMaglevMapKey = nat.NewMaglevBackendKey(net.ParseIP(clusterIP), port, proto, 0)
+						case "ipv6":
+							testMaglevMapKey = nat.NewMaglevBackendKeyV6(net.ParseIP(clusterIP), port, proto, 0)
+						default:
+							log.Panicf("Unexpected IP family %s", family)
+						}
+
+						log.Info("Waiting for Maglev map to converge...")
+						Eventually(maglevMapAnySearchFunc(testMaglevMapKey, family, tc.Felixes[1]), "10s").ShouldNot(BeNil(), "A maglev map entry never showed up")
+						Expect(maglevMapAnySearch(testMaglevMapKey, family, tc.Felixes[1]).Addr().String()).Should(Or(Equal(w[0][0].IP), Equal(w[0][1])))
```
tomastigera: what does this guarantee? Shoudn't you rather just count the expected number of entries? And perhaps make sure that they all point to the expected backends? (eventually?)

felix/fv/bpf_test.go
```diff
@@ -1916,6 +1916,297 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						// Configure routes on external client and Felix nodes.
+						// Use Felix[1] as a middlebox initially.
+						ipRoute := []string{"ip"}
+						if testOpts.ipv6 {
+							ipRoute = append(ipRoute, "-6")
+						}
+
+						cmdCleanRt := append(ipRoute, "route", "del", clusterIP)
+						_ = externalClient.ExecMayFail(strings.Join(cmdCleanRt, ""))
+						cmdCleanRt = append(ipRoute, "route", "del", externalIP)
```
tomastigera, line 2100: do we need to do both? Shouldn't we just rely on ut to make sure that both are programmed? Ultimately we want to test the failover mechanism and that is the same in both cases, isn't it?

felix/fv/bpf_test.go
```diff
@@ -1916,6 +1916,297 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						// }
+						switch family {
+						case "ipv4":
+							key = conntrack.NewKey(proto, ipA, portA, ipB, portB)
+						case "ipv6":
+							key = conntrack.NewKeyV6(proto, ipA, portA, ipB, portB)
+						}
+						return key
+					}
+					checkConntrackExists := func(ctK conntrack.KeyInterface, f *infrastructure.Felix) (conntrack.ValueInterface, bool) {
```
tomastigera:
```suggestion
					checkConntrackExists := func( f *infrastructure.Felix, ctK conntrack.KeyInterface) (conntrack.ValueInterface, bool) {
```

Easier to read where are you testing it. It would make sense to make it a Felix method

felix/fv/bpf_test.go
```diff
@@ -1916,6 +1916,297 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						ctVal, ctExists = checkConntrackExistsAnyDirection(family, clientIPAddr, uint16(pc.SourcePort), backingPodIPAddr, uint16(tgtPort), tc.Felixes[2])
+						Expect(ctExists).To(BeTrue(), "Conntrack didn't exist on Felix[2] for failover traffic. Did the failover actually occur?")
+
+						// Check the backing node updated conntrack tun_ip to the new loadbalancer node.
+						ctVal, ctExists = checkConntrackExistsAnyDirection(family, clientIPAddr, uint16(pc.SourcePort), backingPodIPAddr, uint16(tgtPort), tc.Felixes[0])
+						Expect(ctExists).To(BeTrue(), "Conntrack didn't exist on backing Felix[0].")
+						Expect(ctVal.Data().TunIP.String()).To(Equal(felixIP(2)), "Backing node did not update its conntrack tun_ip to the new loadbalancer IP")
+					})
+
+					It("should maintain connections to an external IP across loadbalancer failure using maglev", func() {
```
tomastigera: just a copy paste of the above. Would make some sense if it was parametrized, but I do not think testing both is important

felix/fv/bpf_test.go
aaaaaaaalex: I wanted to guarantee that the map was actually written by the time I started the test, with at least one relevant key - leaving the job of LUT-correctness to the UTs.

That being said, the `Or()` is no longer relevant to the test, since Im not doing anything with w[0][1] anymore

felix/fv/bpf_test.go
aaaaaaaalex: At the time of writing these, I hadn't a UT to check that kube-proxy programmed Ext IP's, so this test ended up finding a bug, doesn't seem worth it now to remove, though I'll add a UT in kube-proxy if possible to test the Ext IP handling

felix/fv/bpf_test.go
aaaaaaaalex: Would like to make it a Felix method but I'd have to also migrate all the `dump*` methods too, so I kept it here instead. Will adjust the signature, anyway!

felix/fv/bpf_test.go
aaaaaaaalex: It's parameterised now :+1:

PR #11009, opened 2025-09-15
Fixes for 32-bit architectures

twz123:
## Description

This addresses some issues when using Calico on 32-bit architectures. With these patches applied, Calico compiles again on ARMv7 and unit tests mostly pass (modulo some trivial test failures that want to stuff values bigger than two billions into a signed 32-bit integer).

I've tried to make small, self-describing commits.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


danudey:
Hi @twz123 , thanks for your suggestion!

Can I ask what environment you're looking to deploy Calico into which still supports armv7? Kubernetes has not supported 32-bit ARM [for quite some time](https://github.com/kubernetes/kubernetes/issues/115738), and wasn't well tested or stable for a while before then. This is the reason we removed armv7 support from our releases a while ago.

Are you still using an older version of Kubernetes which had published 32-bit ARM builds, or are you compiling kubernetes yourself?

twz123:
I'm doing this as part of [k0s's](https://github.com/k0sproject/k0s) maintenance work. We're building, testing and shipping releases for ARMv7.  While upstream Kubernetes won't provide pre-built binaries for ARMv7, its build system still supports it, and k0s's integration test suite still passes on ARMv7.

libcalico-go/lib/net/ipnet.go
```diff
@@ -74,6 +74,10 @@ func (i IPNet) Covers(n net.IPNet) bool {
 }
 
 func (i IPNet) NthIP(n int) IP {
+	return i.nthIP(int64(n))
```
caseydavenport, line 77: I don't quite follow this change - the existing function already cast `n` to an `int64` first thing, what is the indirection to another function doing for us here?

twz123: This is to avoid a signature change from NthIP itself. There's test cases which require `n` to be `> 0x7FFFFFFF`, which requires an int64 on 32-bit platforms. I figured delegating to a private method that accepts an int64 which can be used in tests would be the least invasive change.

Other options to tackle this:

- Make the test cases which require `int64` optional, based on the size of `int`.
- Change the signature of the exported method to `NthIP(int64) IP`. This requires an explicit cast to int64 at every call site.

caseydavenport:
@twz123 sorry for the delay on this one - it seems to be failing some tests, but I suspect they are unrelated to the changes. Would you be able to pull in the latest master branch to this PR to see if it resolves any of those test failures?

twz123:
>Would you be able to pull in the latest master branch to this PR to see if it resolves any of those test failures?

@caseydavenport Did a rebase.

caseydavenport:
I think this one finally appears to be passing tests, thank you!

twz123:
@caseydavenport Would it be possible to get this [backported](https://github.com/k0sproject/k0s/issues/6988#issuecomment-3778809051) to older releases?

PR #11012, opened 2025-09-15
Handle ENOLINK when updating an existing tcx program

sridhartigera:
## Description

This PR handles
1. ENOLINK when updating an existing link while attaching tcx program.
2. If the interface is deleted, remove the tcx pins.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -1090,9 +1090,7 @@ func cleanupTcxPins(iface string) {
 				Hook:  attachHook,
 			},
 		}
-		if err := ap.DetachTcxProgram(); err != nil {
-			log.WithError(err).Warnf("Failed to detach old tcx program from now gone device '%s' hook %s", iface, attachHook)
-		}
+		os.Remove(ap.ProgPinPath())
 	}
 }
 
--- a/felix/bpf/tc/attach.go
+++ b/felix/bpf/tc/attach.go
@@ -110,7 +110,7 @@ func (ap *AttachPoint) attachTCXProgram(binaryToLoad string) error {
 		return fmt.Errorf("object %w", err)
 	}
 	defer obj.Close()
-	progPinPath := ap.progPinPath()
+	progPinPath := ap.ProgPinPath()
 	if _, err := os.Stat(progPinPath); err == nil {
 		link, err := libbpf.OpenLink(progPinPath)
 		if err != nil {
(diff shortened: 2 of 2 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1147,6 +1147,16 @@ func (m *bpfEndpointManager) onInterfaceUpdate(update *ifaceStateUpdate) {
 		return
 	}
 
+	if update.State == ifacemonitor.StateNotPresent && m.bpfAttachType == apiv3.BPFAttachOptionTCX {
+		// Delete the tcx pins if the interface is gone.
+		// Check if the interface still exists, as we might get events out of order.
+		_, err := m.dp.getIfaceLink(update.Name)
```
tomastigera, line 1153: don't you need to remove it anyway? You got an event that it is not present, but the device is present, so there was a moment where it wasn't present, so it is in fact a different device even if with the same name. That would still break the pin, right? And then [tcx->dev](https://github.com/torvalds/linux/blob/master/kernel/bpf/tcx.c#L213-L215) is likely zero otherwise it would point to an invalid pointer.

So I think you should just delete it here. WDYT?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1147,6 +1147,16 @@ func (m *bpfEndpointManager) onInterfaceUpdate(update *ifaceStateUpdate) {
...
+		if err != nil {
+			if err := m.cleanupOldTcAttach(update.Name); err != nil {
```
tomastigera: shouldn't it only remove the pins? Isn't detaching fail and spit out bunch or errors which will be benign but possibly scare users?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1082,6 +1082,20 @@ func (m *bpfEndpointManager) cleanupOldXDPAttach(iface string) error {
...
 
+func cleanupTcxPins(iface string) {
+	for _, attachHook := range []hook.Hook{hook.Ingress, hook.Egress} {
+		ap := tc.AttachPoint{
+			AttachPoint: bpf.AttachPoint{
+				Iface: iface,
+				Hook:  attachHook,
+			},
+		}
+		if err := ap.DetachTcxProgram(); err != nil {
```
tomastigera: in the DetachTcxProgram, what if link.Open or link.Detach fail, you should still call `os.Remove(progPinPath)` right?

Is it worth reporting that the detach failed? The program is quite likely detached, we just need to remove the pin, don't we?

PR #11019, opened 2025-09-16
OpenStack task tracking

nelljerram:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


nelljerram:
> Seems reasonable; only concern would be if the thread local req ID is used somewhere else; for example, for doing OTel tracing.

I'm not currently aware of our users using their own code modifications alongside our Calico integration.  But it's certainly possible.  I think it makes sense to see if this gets reported, and then decide how to address the coexistence question.

Issue #10720, opened 2025-07-23
Shutting down due to fatal error error=failed to read from netlink (resync): device or resource busy


linecolumn:
We are intermittently encountering a fatal error in **calico-node** on random Kubernetes nodes on our. When this occurs, the affected node becomes completely unusable for workload, and the only reliable way to recover is to reboot the host.

```
[ERROR][392] felix/daemon.go 411: Shutting down due to fatal error error=failed to read from netlink (resync): device or resource busy
[WARNING][392] felix/daemon.go 777: Felix is shutting down reason="fatal error"
[INFO][392] felix/daemon.go 832: Sleeping to avoid tight restart loop. reason="fatal error"
[FATAL][392] felix/daemon.go 845: Exiting. reason="fatal error"
```

We see this issues on following on-prem clusters:

Kubernetes v1.32.4 + Calico v3.29.4 + RHEL v8
Kubernetes v1.30.12 + Calico v3.29.4 + RHEL v8

iptables-1.8.5-11.el8.x86_64 (RHEL v8 stock)
nftables-1.0.4-7.el8_10.x86_64 (RHEL v8 stock)

Calico settings:
- IPIP never
- VXLAN never

When observing host logs there's no events happening at same time which might affect this (no network issues), there are no interface changes, errors or drops happening.

When inspecting host-level logs (e.g., journalctl, dmesg) at the time of failure:
- No network disruptions, interface changes, or link flaps are seen
- No signs of packet drops, hardware errors, or system resource exhaustion are present
- Interfaces remain up and healthy

Please advise if you encountered similar issues or have fixes.

caseydavenport:
@linecolumn sorry for the delay. I haven't seen this particular issue with Calico before, but it sounds like there is some resource contention with netlink - either because Calico is overwhelming the netlink API or some other process is and it is impacting Calico.

Is there anything else notable running on these nodes? It may help to get a full `calico-node` log from one of the nodes that is encountering this issue to see if we can spot anything that might lead to this.

Is this on a new cluster? Or an upgraded one? i.e., any changes in the cluster that may have precipitated this occurrence?

AmerSelimovic:
We are encountering the same error as described by @linecolumn.

These are the component versions for environments in which the issues occur.

`First environment`
- Kubernetes v1.32.5
- Calico v3.29.5
- CoreDNS 1.11.3
- OS Rocky Linux 8.10
- Kernel 5.15.183

`Second environment`
[...]

PR #11021, opened 2025-09-16
[v3.29] Fix EBUSY errors from netlink operations.

caseydavenport:
## Cherry-pick history
- Pick onto **release-v3.29**: projectcalico/calico#9769
## Description

Hunch is that the change in https://github.com/vishvananda/netlink/pull/1018, will prevent EBUSY by draining the remaining messages from the kernel before returning.

Before https://github.com/vishvananda/netlink/pull/925, all messages would be read but the "interrupted" flag was ignored.  That PR made it so that the dump would return early, but that would leave some messages unconsumed.  There was logic elsewhere to discard these messages, but the kernel has a check for "dump in progress" at the start of a new dump and it returns EBUSY in that case:

```
int __netlink_dump_start(struct sock *ssk, struct sk_buff *skb,
			 const struct nlmsghdr *nlh,
			 struct netlink_dump_control *control)
{
...
	mutex_lock(&nlk->nl_cb_mutex);
	/* A dump is in progress... */
	if (nlk->cb_running) {
		ret = -EBUSY;
		goto error_unlock;
	}
```

I wasn't able to trace the exact path through the kernel code to verify when cb_running would be set, or that the error would be returned on the NLMSG_DONE as we see here but it seemed like a solid bet.

## Related issues/PRs

CORE-10999
#10720

## Release Note

```release-note
Fix that netlink list operations could fail with "device or resource busy" errors when under load resulting in Felix restarting.
```

### DIFF
--- a/felix/ifacemonitor/iface_monitor.go
+++ b/felix/ifacemonitor/iface_monitor.go
@@ -411,13 +411,13 @@ func (m *InterfaceMonitor) storeAndNotifyLinkInner(ifaceExists bool, ifaceName s
 
 func (m *InterfaceMonitor) resync() error {
 	log.Debug("Resyncing interface state.")
-	// we need to retry when the error returned is EINTR.
 	var links []netlink.Link
 	var err error
 	retries := 3
 	for {
 		links, err = m.netlinkStub.LinkList()
 		if err != nil {
+			// EINTR means the dump was inconsistent and we should retry.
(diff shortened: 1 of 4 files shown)


caseydavenport:
This doesn't appear to "just work" and I don't have cycles at the moment to diagnose, so for now this pick will need to remain on hold or be picked up by someone else!

PR #11023, opened 2025-09-16
WatcherCache handles missing APIs gracefully

caseydavenport:
## Description

This change allows Calico to run successfully even if one of its backing APIs is not installed in the cluster. 

Instead, Calico will treat the API as if it has successfully sync'd (with no resources) and retry to see if the API exists in 30 minutes.

Note - it is not expected that users will be adding / removing these APIs as a frequent occurrence. Rather, API should be set at install time and not modified afterwards. The implementation matches this expectation - we do not watch the CRDs themselves to see if they are added / removed. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Make Calico's backing CRDs optional. Calico will no longer fail to start if a CRD is missing. 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/watchersyncer/watchercache.go
+++ b/libcalico-go/lib/backend/watchersyncer/watchercache.go
@@ -52,6 +52,10 @@ type watcherCache struct {
 	resyncBlockedUntil     time.Time
 	lastSuccessfulConnTime time.Time
 	watchRetryTimeout      time.Duration
+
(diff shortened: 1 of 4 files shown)


mazdakn (approved):
LGTM.
Do we need any doc update to describe this behavior? Otherwise, we must make sure the release note is clear enough.

PR #11052, opened 2025-09-19
Maintain static ARP programming beyond endpoint start of day

nelljerram:
Before Calico v3.29 we refreshed the static ARP programming for a workload endpoint whenever the endpoint interface state changed, or when the dataplane performed a full route refresh - which by default means every 90 seconds (RouteRefreshInterval).  We accidentally lost that logic in https://github.com/projectcalico/calico/pull/8418.  After that change, Calico only programs static ARP when first configuring a workload endpoint, and does not refresh it after that point.

For most VMs (with Calico for OpenStack) and possibly all Kubernetes usage, that doesn't matter, because the host kernel sends a dynamic ARP request when it needs to, and most VMs and pods respond to that.  However it's possible for a VM to be configured such that it does not respond to such ARP requests - e.g. with `arp_ignore` set to 2.  Connectivity to VMs like that worked before Calico v3.29, but was broken from v3.29 onwards.

This PR reinstates the logic to refresh static ARP programming when the endpoint interface state changes or when the dataplane performs a full resync.

## Release Note

```release-note
Bugfix: reinstate support for VMs that are configured not to respond to ARP requests.
```

### DIFF
--- a/felix/routetable/route_table.go
+++ b/felix/routetable/route_table.go
@@ -142,6 +142,7 @@ type RouteTable struct {
 	// Interface update tracking.
 	fullResyncNeeded    bool
 	ifacesToRescan      set.Set[string]
+	ifacesToARP         set.Set[string]
 	makeARPEntries      bool
 	haveMultiPathRoutes bool
 
@@ -303,6 +304,7 @@ func New(
 
 		fullResyncNeeded: true,
 		ifacesToRescan:   set.New[string](),
+		ifacesToARP:      set.New[string](),
 		ownershipPolicy:  ownershipPolicy,
 
 		ifaceToRoutes: map[RouteClass]map[string]map[ip.CIDR]Target{},
@@ -915,6 +917,12 @@ func (r *RouteTable) maybeResyncWithDataplane() error {
 
 	if r.fullResyncNeeded {
 		return r.doFullResync(nl)
+
+		// Mark to reprogram static ARP for all interfaces.
+		for ifaceName := range r.pendingARPs {
+			r.ifacesToARP.Add(ifaceName)
+		}
+
 	}
 
 	// Do any partial per-interface resyncs.
@@ -1042,6 +1050,9 @@ func (r *RouteTable) resyncIface(nl netlinkshim.Interface, ifaceName string) err
 		r.logCxt.Debug("Ignoring rescan of unknown interface.")
 		return nil
 	}
+
+	// Mark to reprogram static ARP for this interface.
+	r.ifacesToARP.Add(ifaceName)
 
 	routeFilter := &netlink.Route{
 		Table:     r.tableIndex,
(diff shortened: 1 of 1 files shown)


felix/routetable/route_table.go
```diff
@@ -914,6 +916,11 @@ func (r *RouteTable) maybeResyncWithDataplane() error {
 	}
 
 	if r.fullResyncNeeded {
+		// Mark to reprogram static ARP for all interfaces.
+		for ifaceName := range r.pendingARPs {
+			r.ifacesToARP.Add(ifaceName)
+		}
+
```
tomastigera, line 923: Should this be the other way around? There should be nothing in pendingARPs if now new device appeared and we succeeded to sync before. But we need to say "now we need to also check ARPs for this device" But we do not keep the MACs for each device. So we cannot add them.

```suggestion
		// Mark to reprogram static ARP for all interfaces.
		if !r.makeARPEntries {
		    r.pendingARPs.Add(ifaceName)
		}
```

felix/routetable/route_table.go
```diff
@@ -142,6 +142,7 @@ type RouteTable struct {
 	// Interface update tracking.
 	fullResyncNeeded    bool
 	ifacesToRescan      set.Set[string]
+	ifacesToARP         set.Set[string]
```
tomastigera, line 145: this only track which devices need the permanent arp entries. But it does not track those ARPs. We can get the MACs from `ifaceToRoutes` where each route is Target and those target **may** have MAC set.

it used to be this https://github.com/projectcalico/calico/blob/release-v3.27/felix/routetable/route_table.go#L669-L679

felix/routetable/route_table.go
```diff
@@ -1470,18 +1480,33 @@ func (r *RouteTable) applyUpdates(attempt int) error {
 
 	arpErrs := map[string]error{}
 	for ifaceName, addrToMAC := range r.pendingARPs {
-		// Add static ARP entries (for workload endpoints).  This may have been
-		// needed at one point but it no longer seems to be required.  Leaving
-		// it here for two reasons: (1) there may be an obscure scenario where
-		// it is needed. (2) we have tests that monitor netlink, and they break
-		// if it is removed because they see the ARP traffic.
+		if !r.ifacesToARP.Contains(ifaceName) {
+			continue
```
tomastigera, line 1484: if you add the ARPs to the pending when the devices need resync, than there should not be any need for `ifacesToARP` unless we do not want to dig out the ARPs from the routes every time and cache them in `ifacesToARP`. Modifying routes would need to modify this cache as well.

felix/routetable/route_table.go
nelljerram: We no longer remove entries from `pendingARPs`, unless `removePendingARP` explicitly does that.  I should probably also rename `pendingARPs` to `staticARPs` or something like that.

nelljerram:
Thanks @tomastigera for looking at this.  I think all of your points are answered (or at least affected) by https://github.com/projectcalico/calico/pull/11052/files#diff-ccb6c09ae340f8ed75c8963880f16775105a44b0049a5a9e002c3cb73e423160L1513-L1514 - i.e. that we no longer remove entries from `pendingARPs`.

felix/routetable/route_table.go
```diff
@@ -1470,18 +1480,33 @@ func (r *RouteTable) applyUpdates(attempt int) error {
 
 	arpErrs := map[string]error{}
 	for ifaceName, addrToMAC := range r.pendingARPs {
-		// Add static ARP entries (for workload endpoints).  This may have been
-		// needed at one point but it no longer seems to be required.  Leaving
-		// it here for two reasons: (1) there may be an obscure scenario where
-		// it is needed. (2) we have tests that monitor netlink, and they break
-		// if it is removed because they see the ARP traffic.
+		if !r.ifacesToARP.Contains(ifaceName) {
```
tomastigera, line 1483: OK, right, ifacesToARP is the list of devices that needs to refresh ARP programming and pendingARPs is caching the ARPs per interfaces.

`for ifaceName, addrToMAC := range r.pendingARPs` could iterate over the ifacesToARP and then get the ARPs for that interface from pending.

nelljerram: Yes, but I don't think that would be a clear win, because I'd need to check if there's an entry for ifaceName in r.pendingARPs.  So I think the code would still be equally complex.

tomastigera: yes, but on a system with many ifaces, if only one flips you need to walk all. Probably does not matter to much.

PR #11099, opened 2025-09-30
[v3.30] Merge pull request #11052 from nelljerram/maintain-static-arp

nelljerram:
Maintain static ARP programming beyond endpoint start of day

(cherry-pick for v3.30 branch)

## Description

Before Calico v3.29 we refreshed the static ARP programming for a workload endpoint whenever the endpoint interface state changed, or when the dataplane performed a full route refresh - which by default means every 90 seconds (RouteRefreshInterval).  We accidentally lost that logic in https://github.com/projectcalico/calico/pull/8418.  After that change, Calico only programs static ARP when first configuring a workload endpoint, and does not refresh it after that point.

For most VMs (with Calico for OpenStack) and possibly all Kubernetes usage, that doesn't matter, because the host kernel sends a dynamic ARP request when it needs to, and most VMs and pods respond to that.  However it's possible for a VM to be configured such that it does not respond to such ARP requests - e.g. with `arp_ignore` set to 2.  Connectivity to VMs like that worked before Calico v3.29, but was broken from v3.29 onwards.

This PR reinstates the logic to refresh static ARP programming when the endpoint interface state changes or when the dataplane performs a full resync.

## Release Note

```release-note
Bugfix: reinstate support for VMs that are configured not to respond to ARP requests.
```

### DIFF
--- a/felix/netlinkshim/mocknetlink/netlink.go
+++ b/felix/netlinkshim/mocknetlink/netlink.go
@@ -955,6 +955,27 @@ func addNeighs(family int, neighMap map[NeighKey]*netlink.Neigh, neighs []netlin
 	}
 }
 
+// RemoveNeighs allows test code to remove neighbours from the mock dataplane
+// without going through the netlink API.
+func (d *MockNetlinkDataplane) RemoveNeighs(family int, neighs ...netlink.Neigh) {
+	err := d.checkNeighFamily(family)
+	if err != nil {
+		panic(err)
+	}
+	if d.NeighsByFamily[family] == nil {
+		return
+	}
+	removeNeighs(family, d.NeighsByFamily[family], neighs)
+}
+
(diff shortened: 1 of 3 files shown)


nelljerram:
There is a remaining CI issue here in "UT/FV tests on new kernel":
```
[batch=4 pid=11595] [Fail] _BPF_ _BPF-SAFE_ BPF tests (ipv4 tcp, ct=true, log=debug, tunnel=none, dsr=false, cleaner=Userspace) (kubernetes backend) with a single node and an allow-all policy [It] should cleanup after we disable eBPF 
```
However that same failure is occurring on the existing release-v3.30 branch, e.g. [here](https://tigera.semaphoreci.com/jobs/89092853-17d9-46d4-8875-a4b25eecc6f7).  So I think I'm OK to force-merge this PR now.

PR #11104, opened 2025-10-01
calicoctl: add optimize sub-command that can optimize GNPs for better efficiency

fasaxc:
## Description

Add a new `calicoctl optimize` sub-command that takes in one or more Calico resources and emits (possibly) optimized versions to stdout.

In this initial version, only global network policies are optimized, other resources are passed through unaltered.  The tool makes several optimizations to global network policies:

- Hoist selectors up from rules to the top-level subject selector.

  - When an **ingress** rule uses a **destination** selector, that results in rendering an expensive IP set to the dataplane (including every IP that the selector matches) even though Felix already knows which workload it is applying the policy to.
  - Similarly, **source** selectors in **egress** are inefficient.
  - When the tool spots such a selector, it splits the policy into smaller policies, each with a different selector.  Then it hoists the selector up to the top-level of the policy so that the policy only applies to that subset of workloads.  This removes the IP set, and, if the top-level selector was previously broad, it reduces the number of workloads that the sub policy even applies to.  Policies that don't apply to any endpoint on a given node are not even calculated and do not get rendered to iptables.

- Remove duplicate rules within a policy.
- Remove unreachable rules (rules after an unconditional Allow/Deny/Pass).
- Normalize policy selectors.

## Related issues/PRs

CORE-11913

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
calicoctl: add optimize sub-command that can optimize GNPs for better efficiency including splitting large policies that misuse source selectors on egress rules/dest selectors on ingress rules.
```

## Reminder for the reviewer

Other optional labels:


fasaxc:
Closing this one for now because one of the core assumptions that the optimization relies upon falls down in certain use cases.  In particular, the optimizer assumes that source/dest selector in rules will match workloads via their workload endpoint labels.  If that is violated  (because the user is matching workloads via IPs/CIDRs in network sets, then the optimization doesn't work).

PR #11111, opened 2025-10-01
Check for Whisker in cluster before running tests

caseydavenport:
## Description

This change produces a better error message when the e2e tests for Whisker are run against a cluster that does not have Whisker installed.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/e2e/pkg/tests/policy/staged_policy.go
+++ b/e2e/pkg/tests/policy/staged_policy.go
@@ -12,9 +12,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
-	appsv1 "k8s.io/api/apps/v1"
 	networkingv1 "k8s.io/api/networking/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


e2e/pkg/tests/policy/staged_policy.go
```diff
@@ -53,8 +53,12 @@ var _ = describe.CalicoDescribe(
 		BeforeEach(func() {
 			cli, err = client.New(f.ClientConfig())
 			Expect(err).NotTo(HaveOccurred())
-
 			checker = conncheck.NewConnectionTester(f)
+
+			// These tests rely on Whisker - if Whisker is not installed, short-circuit the tests.
+			installed, err := utils.WhiskerInstalled(cli)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(installed).To(BeTrue(), "Whisker is not installed in the cluster")
```
sridhartigera, line 61: shouldn't we skip these tests against the clusters where whisker isn't installed(for example the manifest based installs). IIUC, this will be flagged as a failure. Is that ok?

caseydavenport: I'm generally of the opinion tests shouldn't be self-skipping. It's too error prone, and makes it harder to use the tests to verify if a given cluster supports a feature. 

Whoever is running the tests should skip whatever tests are not applicable to that environment.

PR #11113, opened 2025-10-01
ci: Grab crane from github instead of via go-build

danudey:
Instead of using go-build for crane, which involves bind-mounting an ever-increasing number of directories and files into the go-build container and performing various copy operations on files in the container every time, simply:

1. Pull down crane from github (via github releases) into `bin/crane`
2. Use that

This should enable all the functionality we need now and going forward, as well as simplifying the process since we won't need to debug it every time things change.

We already depend on being able to pull go libraries from github in order to do our builds so this doesn't add additional external requirements to the build process, and we already had a makefile target to pull `bin/crane` anyway; now we're just using it consistently.

### DIFF
--- a/lib.Makefile
+++ b/lib.Makefile
@@ -1255,7 +1255,7 @@ bin/yq:
 # This setup is used to download and install the 'crane' binary into the local bin/ directory.
 # The binary will be placed at: ./bin/crane
 # Normalize architecture for go-containerregistry filenames
-CRANE_BUILDARCH := $(shell uname -m | sed 's/amd64/x86_64/;s/x86_64/x86_64/;s/aarch64/arm64/')
+CRANE_BUILDARCH := $(shell uname -m | sed 's/aarch64/arm64/')
 CRANE_OS := $(shell uname -s)
 ifeq ($(CRANE_BUILDARCH),)
   $(error Unsupported or unknown architecture: $(shell uname -m))
--- a/lib.Makefile
+++ b/lib.Makefile
@@ -1252,9 +1252,26 @@ bin/yq:
 	tar -zxvf $(TMP)/yq4.tar.gz -C $(TMP)
 	mv $(TMP)/yq_linux_$(BUILDARCH) bin/yq
 
-# Build and install crane binary into $(REPO_ROOT)/bin
+# This setup is used to download and install the 'crane' binary into the local bin/ directory.
+# The binary will be placed at: ./bin/crane
+# Normalize architecture for go-containerregistry filenames
+CRANE_BUILDARCH := $(shell uname -m | sed 's/amd64/x86_64/;s/x86_64/x86_64/;s/aarch64/arm64/')
+CRANE_OS := $(shell uname -s)
+ifeq ($(CRANE_BUILDARCH),)
+  $(error Unsupported or unknown architecture: $(shell uname -m))
+endif
+ifeq ($(CRANE_OS),)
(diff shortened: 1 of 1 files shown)


lib.Makefile
```diff
@@ -524,10 +523,10 @@ GIT           = $(GIT_CMD)
...
 RELEASE_PY3   = $(PYTHON3_CMD)
 else
-CRANE         = @echo [DRY RUN] $(CRANE_CMD)
-GIT           = @echo [DRY RUN] $(GIT_CMD)
-DOCKER        = @echo [DRY RUN] $(DOCKER_CMD)
-RELEASE_PY3   = @echo [DRY RUN] $(PYTHON3_CMD)
+CRANE         = echo [DRY RUN] $(CRANE_CMD)
+GIT           = echo [DRY RUN] $(GIT_CMD)
+DOCKER        = echo [DRY RUN] $(DOCKER_CMD)
+RELEASE_PY3   = echo [DRY RUN] $(PYTHON3_CMD)
```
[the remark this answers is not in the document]
danudey: We removed this because we use `$(CRANE)` in the Windows `release-image` rule, which is a giant shell script written directly into the Makefile and therefore can't run `@echo`. Will find a better option for this in future.

lib.Makefile
```diff
@@ -1253,23 +1252,9 @@ bin/yq:
...
-
-# Install crane binary into bin/
+# Build and install crane binary into $(REPO_ROOT)/bin
 bin/crane:
-	mkdir -p bin
-	$(eval CRANE_TMP := $(shell mktemp -d))
-	curl -sSfL --retry 5 -o $(CRANE_TMP)/crane.tar.gz $(CRANE_URL)
-	tar -xzf $(CRANE_TMP)/crane.tar.gz -C $(CRANE_TMP) crane
-	mv $(CRANE_TMP)/crane bin/crane
+	GOBIN=$(REPO_ROOT)/bin go install github.com/google/go-containerregistry/cmd/crane@$(CRANE_VERSION)
```
hjiawei: Probably better to install the pre-build binary instead like what we did [in go-build](https://github.com/projectcalico/go-build/blob/master/images/calico-go-build/Dockerfile#L172). There is no need to download dependencies, cached builds / rebuilds, and overwrite the `GOBIN`.

```
curl -sfL https://github.com/google/go-containerregistry/releases/download/${CONTAINERREGISTRY_VERSION}/go-containerregistry_Linux_x86_64.tar.gz | tar xz -C /usr/local/bin crane
```

danudey: Reverted this approach, with minor changes.

lib.Makefile
```diff
@@ -1252,9 +1252,26 @@ bin/yq:
 	tar -zxvf $(TMP)/yq4.tar.gz -C $(TMP)
 	mv $(TMP)/yq_linux_$(BUILDARCH) bin/yq
 
-# Build and install crane binary into $(REPO_ROOT)/bin
+# This setup is used to download and install the 'crane' binary into the local bin/ directory.
+# The binary will be placed at: ./bin/crane
+# Normalize architecture for go-containerregistry filenames
+CRANE_BUILDARCH := $(shell uname -m | sed 's/amd64/x86_64/;s/x86_64/x86_64/;s/aarch64/arm64/')
```
hjiawei: I think only `aarch64` to `arm64` is necessary. uname -m will always give `x86_64` for 64bit x84 architectures.

danudey: Fixed up.

PR #11121, opened 2025-10-03
surface stderr from hack/deps go mod cmd

aaaaaaaalex:
## Description

Title says it all

### DIFF
--- a/hack/cmd/deps/deps.go
+++ b/hack/cmd/deps/deps.go
@@ -430,9 +430,13 @@ func loadGoMods() ([]module, error) {
 }
 
 func loadGoToolJSON[Item any](args ...string) ([]Item, error) {
-	out, err := exec.Command("go", args...).Output()
+	cmd := exec.Command("go", args...)
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	out, err := cmd.Output()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


hack/cmd/deps/deps.go
```diff
@@ -430,9 +431,13 @@ func loadGoMods() ([]module, error) {
...
 func loadGoToolJSON[Item any](args ...string) ([]Item, error) {
-	out, err := exec.Command("go", args...).Output()
+	cmd := exec.Command("go", args...)
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	out, err := cmd.Output()
 	if err != nil {
-		return nil, err
+		errOut := stderr.Bytes()
+		return nil, errors.Wrap(err, string(errOut))
```
fasaxc:
```suggestion
		return nil, fmt.Errorf("failed to run go %v: %w, stderr: %s", args, err, string(stderr.Bytes()))
```
I think that errors package is the "old way"; fmt.Errorf is in the stdlib now.

aaaaaaaalex: Right you are. Fixed

aaaaaaaalex:
Oops, UI was lagging behind - ignore the review request

PR #11127, opened 2025-10-04
ci: Ensure we put crane in $(REPO_ROOT) and look there

danudey:


coutinhop (approved):

matthewdupre (approved):
Don't see how this could break a Felix FV, will force merge

PR #11129, opened 2025-10-06
Semaphore: depend on input YAML templates

fasaxc:
## Description

- Move semaphore.yml calculation to Go.
- Rework to allow for pre-calculation of Go dependencies while
  tracking which file we're working on.
- Add explciit deps on pre-amble and specific template files.

## Related issues/PRs

CORE-11916

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


hack/cmd/deps/deps.go
```diff
@@ -145,86 +163,93 @@ var defaultExclusions = []string{
...
-	changeInPattern := regexp.MustCompile(`\$\{CHANGE_IN\(([^)]+)\)}`)
-	replacements := map[string][]byte{}
-	for _, groups := range changeInPattern.FindAllSubmatch(fileContents, -1) {
-		pkg := string(groups[1])
-		if _, ok := replacements[pkg]; ok {
-			continue // already calculated.
-		}
-		replacements[pkg] = nil
-	}
+func calculateDeps(placeholders set.Set[string]) map[string]*Deps {
```
nelljerram: I'm wondering why the name "placeholders" here.  Looks like it's a list of packages.

hack/cmd/deps/deps.go
```diff
@@ -430,13 +447,9 @@ func loadGoMods() ([]module, error) {
...
 func loadGoToolJSON[Item any](args ...string) ([]Item, error) {
-	cmd := exec.Command("go", args...)
-	var stderr bytes.Buffer
-	cmd.Stderr = &stderr
-	out, err := cmd.Output()
+	out, err := exec.Command("go", args...).Output()
 	if err != nil {
-		errOut := stderr.Bytes()
-		return nil, fmt.Errorf("%w, %s", err, string(errOut))
+		return nil, err
```
nelljerram: Why removing the stderr here?  Is it known that it will never be useful?

hack/cmd/deps/deps.go
```diff
@@ -128,9 +137,18 @@ var nonGoDeps = map[string][]string{
...
+		"/third_party/snort3",
+	},
+	"elasticsearch": {
+		"/third_party/elasticsearch",
+	},
+	"fluentd": {
+		"/third_party/fluentd-base",
+	},
+	"kibana": {
+		"/third_party/kibana",
```
nelljerram: Wondering why these changes are here.

hack/cmd/deps/deps.go
```diff
@@ -450,3 +463,294 @@ func loadGoToolJSON[Item any](args ...string) ([]Item, error) {
...
+	if err != nil {
+		return err
+	}
+	if len(offending) > 0 {
+		return fmt.Errorf("All change_in clauses must include pipeline_file: 'ignore' (or 'track'). Offending lines:\n%s", strings.Join(offending, "\n"))
+	}
+	return nil
+}
+
+func calculateBranchStanza(semaphoreDir string) string {
```
nelljerram, line 698: It looks like this will work correctly and give "" for OSS.  But were you intending to rely on that, or did you mean to code another OSS/Enterprise switch somewhere?

hack/cmd/deps/deps.go
```diff
@@ -450,3 +463,294 @@ func loadGoToolJSON[Item any](args ...string) ([]Item, error) {
...
+	return os.WriteFile(file, data.Bytes(), 0644)
+}
+
+func indentBlocks(blocks []templateData) []templateData {
+	for i, block := range blocks {
+		lines := strings.Split(block.content, "\n")
+		indented := ""
+		for _, line := range lines {
+			if line == "" {
+				continue
```
nelljerram, line 609: You're intentionally omitting blank lines?  Worth adding a comment to say that.

hack/cmd/deps/deps.go
```diff
@@ -450,3 +463,294 @@ func loadGoToolJSON[Item any](args ...string) ([]Item, error) {
...
+
+	logrus.Info("Semaphore YAML generation complete")
+}
+
+func buildSemaphoreYAML(file string, templates []templateData, globalExtraDeps []string, deps map[string]*Deps, weekly bool, defaultBranchStanza string) error {
+	var data bytes.Buffer
+
+	data.WriteString(
+		"# !! WARNING, DO NOT EDIT !! This file is generated from semaphore.yml.d.\n" +
+			"# To update, modify the template and then run 'make gen-semaphore-yaml'.\n",
```
nelljerram: "the template" isn't quite right.  "the relevant template"?

hack/cmd/deps/deps.go
```diff
@@ -450,3 +463,294 @@ func loadGoToolJSON[Item any](args ...string) ([]Item, error) {
...
+	)
+	globalExtraDeps = globalExtraDeps[:len(globalExtraDeps):len(globalExtraDeps)]
```
nelljerram, line 566: What does this line do?  It looks like it's worth a comment.

.semaphore/semaphore.yml.d/09-blocks.yml
nelljerram, line 1: Sorry, another q.  Is this file intended?  I wasn't sure what it would be for in OSS - maybe to do with weekly run in Enterprise?

hack/cmd/deps/deps.go
fasaxc: Good spot, this got clobbered when I upstreamed the changes from ent.

hack/cmd/deps/deps.go
fasaxc: I upstreamed the enterprise version, which had diverged.  I'll take it out since it's confusing and likely to diverge in future.

hack/cmd/deps/deps.go
fasaxc: No, Copilot wrote this bit, translating the upstreamed version; I didn't catch the ent-specific release name.

.semaphore/semaphore.yml.d/09-blocks.yml
fasaxc: Yes, it's intended.  The old script used to output the blocks line itself in between the pre-amble and the block files.  I removed that special case so I need to output that line somehow.

hack/cmd/deps/deps.go
fasaxc: Just the result of a refactoring; fixed.

hack/cmd/deps/deps.go
fasaxc: Have changed to use a regex that should detect OS/ent versions.

PR #11130, opened 2025-10-06
Add more convenient method for swapping the time out for tests using the testing cleanup

Brian-McM:
## Description

It's a bit tedious requiring the closure for every test that needs this, instead, we can just rely on a tests cleanup function to return the clock back to the original state.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/lib/std/time/time.go
+++ b/lib/std/time/time.go
@@ -65,6 +65,21 @@ func DoWithClock(shim Clock, fn func() error) error {
 	return fn()
 }
 
+type CleanUpRegisterable interface {
+	Cleanup(func())
+}
+
+// ShimClockForTestingT temporarily sets the shim as the time and runs the given function for a test and when the test
+// is done, the time is returned to its original state.
+func ShimClockForTestingT(t CleanUpRegisterable, shim Clock) {
+	original := activeClock
+	activeClock = shim
+
+	t.Cleanup(func() {
+		activeClock = original
+	})
+}
+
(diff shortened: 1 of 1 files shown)


tomastigera (approved):

lib/std/time/time.go
```diff
@@ -65,6 +65,21 @@ func DoWithClock(shim Clock, fn func() error) error {
 	return fn()
 }
 
+type CleanUpRegisterable interface {
+	Cleanup(func())
+}
+
+// DoWithClockForTest temporarily sets the shim as the time and runs the given function for a test and when the test
+// is done, the time is returned to its original state.
+func DoWithClockForTest(shim Clock, t CleanUpRegisterable) {
```
fasaxc: `Do` prefix usually hints at taking a closure.  I'd probably call it `ShimClockForTestingT` or similar.  (Quite a few "ForTestingT" functions around already.)

I'd probably put the testing.T as first arg.

Brian-McM: Changed the name and made the cleanup function the first argument, I didn't want to make the argument testing.T since all we need is something that implements the Cleanup function (no need to add testing.T here).

caseydavenport (approved):
Approving to unblock, but spoke about addressing any of @fasaxc's comments if he's got more!

PR #11144, opened 2025-10-07
initial commit

mazdakn:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
```diff
@@ -67,8 +67,6 @@ rules:
       - delete
     resourceNames:
       - clusternetworkpolicies.policy.networking.k8s.io
-      - adminnetworkpolicies.policy.networking.k8s.io
```
fasaxc, line 70: Is it right to remove these; won't the operator need to be able to clean them up after upgrade?  CC @caseydavenport

mazdakn: Right. Probably it's better to keep it.

caseydavenport: Our operator doesn't (AFAIK) ever delete CRDs from the cluster. 

This delete permission is needed specifically because on OpenShift, there is a permissions check that prevents owner references being set unless the we have delete permissions for the object we are applying the owner references to (which makes sense).

mazdakn: @caseydavenport if I understand your comment, then we should not need this permission, since we won't need to set owner references on ANP/BANP CRDs anymore. No?

caseydavenport: yep!

caseydavenport (approved):

PR #11147, opened 2025-10-08
Introduces istio images to be used by operator

radixo:
## Description

Pushes images that will be used by tigera-operator to deploy Istio.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/third_party/istio/Dockerfile.ztunnel
+++ b/third_party/istio/Dockerfile.ztunnel
@@ -12,6 +12,14 @@ FROM scratch AS source
 
 COPY --from=istio-ztunnel --chown=0:0 --chmod=755 \
     /usr/local/bin/ztunnel /usr/local/bin/ztunnel
+# Copying the glibc from ztunnel container because the one present in
+# CALICO_BASE is minor version and not compatible.
+COPY --from=istio-ztunnel --chown=0:0 --chmod=755 \
+    /usr/lib/x86_64-linux-gnu/libc.so.6 /lib64/libc.so.6
+COPY --from=istio-ztunnel --chown=0:0 --chmod=755 \
+    /usr/lib/x86_64-linux-gnu/libm.so.6 /lib64/libm.so.6
+COPY --from=istio-ztunnel --chown=0:0 --chmod=755 \
+    /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2 /lib64/ld-linux-x86-64.so.2
 
 # dependencies
 COPY --from=ubi /bin/sh /bin/sh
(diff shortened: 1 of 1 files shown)


.semaphore/semaphore.yml.d/blocks/20-istio.yml
```diff
@@ -0,0 +1,24 @@
+- name: istio
+  run:
+    when: "${FORCE_RUN} or change_in(['/metadata.mk', '/lib.Makefile', '/third_party/istio'], {pipeline_file: 'ignore', exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
```
hjiawei, line 3: nit: Consider using the new deps tool format introduced by Shaun.

.semaphore/semaphore.yml.d/blocks/20-istio.yml
hjiawei, line 1: You can move the third-party image build pipeline to `50-`. `20-` is reserved for Calico OSS components.

radixo: Why our compilation of Envoy is on `20-`?

radixo:
>If there's a strong need to use the upstream vendor binaries, it would be more appropriate to consume their official image directly rather than copying those binaries into our base image.

I believe we will need to rebuild it from our image instead of Istio base images, because I don't believe to use the official images is a good alternative for clusters that has white lists of repositories. I will discuss with @electricjesus today the best alternative anyway.

third_party/istio/Dockerfile.ztunnel
```diff
@@ -0,0 +1,77 @@
...
+FROM ${UBI_IMAGE} AS ubi
+
+FROM scratch AS source
+
+COPY --from=istio-ztunnel --chown=0:0 --chmod=755 \
+    /usr/local/bin/ztunnel /usr/local/bin/ztunnel
+# Copying the glibc from ztunnel container because the one present in
+# CALICO_BASE is minor version and not compatible.
+COPY --from=istio-ztunnel --chown=0:0 --chmod=755 \
+    /usr/lib/x86_64-linux-gnu/libc.so.6 /lib64/libc.so.6
```
hjiawei, line 18: Copying `libc.so.6` from Debian-based Istio ztunnel image effectively changes the base image to Debian. This means we're no longer using RHEL UBI-based `calico/base` image.

third_party/istio/Dockerfile.install-cni
```diff
@@ -0,0 +1,68 @@
...
+ARG UBI_IMAGE
+ARG ISTIO_CNI_IMAGE
+
+FROM ${ISTIO_CNI_IMAGE} AS istio-cni
+
+FROM ${UBI_IMAGE} AS ubi
+
+FROM scratch AS source
+
+COPY --from=istio-cni --chown=0:0 --chmod=755 \
```
hjiawei, line 13: `--chown` to root and `--chmod` to 755 are generally unnecessary. These are the defaults for binaries and no `USER` directive is specified before `COPY`. You'd need them if the source image uses different ownership or permissions, in which cases it's important to verify why.

.semaphore/semaphore.yml.d/blocks/20-istio.yml
radixo: I tried to look for what you meant here, I believe it is the CHANGE_IN function that @fasaxc introduced on f3dadfed7.
But it seems to me that it doesn't apply to istio third_party, the same way it didn't apply to envoy-gateway.
Am I missing anything?

PR #11149, opened 2025-10-08
IPAM: try to reclaim empty IPAM blocks if pool is empty

fasaxc:
## Description

Fixes a corner case for special-purpose block-size-/32 pools where all the blocks are "stranded" on nodes that don't need them.
## Related issues/PRs

CORE-11223

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
IPAM: when a pool is full, try to reclaim empty blocks from other nodes before giving up. Helps the most with special-purpose block-size-/32 pools; prevents blocks from getting stranded on nodes that can't use them. 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/ipam/ipam_block_reader_writer.go
+++ b/libcalico-go/lib/ipam/ipam_block_reader_writer.go
@@ -108,7 +108,14 @@ func findContainingPool(pools []v3.IPPool, addr net.IP) (*v3.IPPool, error) {
 //
 // Note that the block may become claimed between receiving the CIDR from this function and attempting to claim the corresponding
 // block as this function does not reserve the returned IPNet.
-func (rw blockReaderWriter) findUsableBlock(ctx context.Context, affinityCfg AffinityConfig, version int, pools []v3.IPPool, reservations addrFilter, config IPAMConfig) (*cnet.IPNet, error) {
+func (rw blockReaderWriter) findUsableBlock(
+	ctx context.Context,
+	affinityCfg AffinityConfig,
+	version int,
+	pools []v3.IPPool,
+	reservations addrFilter,
+	config IPAMConfig,
+) (*cnet.IPNet, error) {
 	// If there are no pools, we cannot assign addresses.
 	if len(pools) == 0 {
 		return nil, fmt.Errorf("no configured Calico pools for %s:%s", affinityCfg.AffinityType, affinityCfg.Host)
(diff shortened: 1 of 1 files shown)


libcalico-go/lib/ipam/ipam_block_reader_writer.go
```diff
@@ -158,13 +186,58 @@ func (rw blockReaderWriter) findUsableBlock(ctx context.Context, affinityCfg Aff
...
 		}
 	}
+
+	if len(emptyBlocks) > 0 {
+		// This step is very important for small special-purpose IPAM pools
+		// with small blocks (and especially /32 blocks). Without it, we can
+		// be unable to allocate an IP even though there are fewer workloads
+		// using the special pool than there are total blocks (because the
+		// blocks get stranded, affine to other nodes).
+		log.Info("Ran out of empty blocks, trying to reclaim an empty one.")
```
caseydavenport:
```suggestion
		log.Info("Ran out of unclaimed blocks, trying to reclaim an empty one.")
```

libcalico-go/lib/ipam/ipam_block_reader_writer.go
```diff
@@ -327,6 +412,10 @@ func (rw blockReaderWriter) releaseBlockAffinity(ctx context.Context, affinityCf
 	}
 	b := allocationBlock{obj.Value.(*model.AllocationBlock)}
 
+	if opts.ExpectedSequenceNumber != nil && *opts.ExpectedSequenceNumber != b.SequenceNumber {
```
caseydavenport: I suppose it's possible for the block to be updated after we make this check here, but before we delete / update the block below.  

We do seem to round-trip the resource version information though: https://github.com/projectcalico/calico/blob/master/libcalico-go/lib/backend/k8s/resources/ipam_affinity.go#L150

So, I suppose this check isn't harmful but also should already be handled by the revision round-tripping?

libcalico-go/lib/ipam/ipam_block_reader_writer.go
```diff
@@ -158,13 +186,58 @@ func (rw blockReaderWriter) findUsableBlock(ctx context.Context, affinityCfg Aff
...
+					block.cidr.String(), age.Seconds())
+				continue
+			}
+			err := rw.releaseBlockAffinity(ctx, block.affinityCfg, block.cidr, releaseAffinityOpts{
+				RequireEmpty:           true,
+				ExpectedSequenceNumber: &block.seqNo,
+			})
+			if err != nil {
+				log.Warnf("Failed to release affinity while trying to reclaim block. %v left to try: %s", len(emptyBlocks)-i, err)
+				if ctx.Err() != nil {
```
caseydavenport, line 227: If we didn't do this error check here, presumably we'd just iterate and the next `releaseBlockAffinity` would fail immediately with a context error since the context is passed all the way down to the API client?

libcalico-go/lib/backend/model/block.go
```diff
@@ -113,6 +114,9 @@ type AllocationBlock struct {
 	// Affinity of the block, if this block has one. If set, it will be of the form
 	// "host:<hostname>". If not set, this block is not affine to a host.
 	Affinity *string `json:"affinity"`
+	// Time the affinity was claimed; may be zero for old blocks that predate
```
caseydavenport, line 118:
```suggestion

	// Time the affinity was claimed; may be zero for old blocks that predate
```

libcalico-go/lib/ipam/ipam_block_reader_writer.go
fasaxc: Yes, I just wanted to make sure the loop ended early since we're otherwise trying each block and ignoring errors.

libcalico-go/lib/ipam/ipam_block_reader_writer.go
fasaxc: This extra check ensures that the block wasn't modified since it was read by `findUsableBlock`.  Since `releaseBlockAffinity` re-reads the block, it avoids a race where multiple hosts try to reclaim the block:

* A sees stale empty block
* B see stale empty block
* A and B both try to reclaim
* A wins, releasing IPAM affinity triggers deletion of block
* A creates new version block, it's empty
* B's releaseBlockAffinity  see's A's new block, it's empty so it deletes A's new block
* A's IPAM allocation fails

`findUsableBlock` can take a few seconds in a large full pool so the window is quite large for this kind of thing to happen.  I will add comments to explain.

fasaxc:
Seeing failures in kube-controllers.  Pushed some extra diags to see if it's anything to do with my change.

fasaxc:
Passed on a rerun, I moved the extra diags to #11184, which is now merged.

PR #11151, opened 2025-10-08
Confd Bird template refactoring 

song-jiang:
## Description

This PR implements a major refactoring of the confd Bird templating, moving most of the logic from the Bird configuration template into Go. The current approach has become increasingly difficult to maintain and is hindering future BGP feature development.

It introduces a new Go backend (bgp_processor.go) that processes BGP configuration data and provides it to the templates via the `getBGPConfig` function. This simplifies the template logic by moving complex data processing from the templates to Go code.

Templates Updated:
`bird.cfg.template` - IPv4 BGP configuration
`bird6.cfg.template` - IPv6 BGP configuration
These templates now use getBGPConfig to retrieve pre-processed BGP peer information, community rules, and node configuration, significantly reducing template complexity. The remaining templates continue to work as before with no changes to their logic or behavior.

The refactored templates produce identical BIRD configuration output, with the only difference being the removal of decorative section header comments (e.g., # ------------- Global peers -------------) that separated peer types. These comments have no effect on BIRD's behavior.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/confd/etc/calico/confd/templates/bird.cfg.template
+++ b/confd/etc/calico/confd/templates/bird.cfg.template
@@ -22,12 +22,6 @@ router id {{$config.RouterID}};{{if $config.RouterIDComment}}  {{$config.RouterI
 listen bgp{{if $config.ListenAddress}} address {{$config.ListenAddress}}{{end}}{{if $config.ListenPort}} port {{$config.ListenPort}}{{end}};
 {{- end}}
 
-{{- define "LOGGING"}}
-{{- if .DebugMode}}
-  debug {{.DebugMode}};
-{{- end}}
-{{- end}}
-
 # Configure synchronization between routing tables and kernel.
 protocol kernel {
   learn;             # Learn all alien routes from the kernel
@@ -46,12 +40,16 @@ protocol kernel {
 
 # Watch interface up/down events.
 protocol device {
-{{- template "LOGGING" $config}}
+{{- if $config.DebugMode}}
+  debug {{$config.DebugMode}};
+{{- end}}
   scan time 2;    # Scan interfaces every 2 seconds
 }
 
 protocol direct {
-{{- template "LOGGING" $config}}
+{{- if $config.DebugMode}}
+  debug {{$config.DebugMode}};
+{{- end}}
   interface {{$config.DirectInterfaces}};{{if eq $config.DirectInterfaces `-"cali*", -"kube-ipvs*", "*"`}} # Exclude cali* and kube-ipvs* but
                                           # include everything else.  In
                                           # IPVS-mode, kube-proxy creates a
@@ -67,9 +65,11 @@ protocol direct {
 {{else}}
 # Template for all BGP clients
 template bgp bgp_template {
-{{- template "LOGGING" $config}}
+{{- if $config.DebugMode}}
+  debug {{$config.DebugMode}};
+{{- end}}
   description "Connection to BGP peer";
-  local as {{$config.AsNumber}};
+  local as {{$config.ASNumber}};
   gateway recursive; # This should be the default, but just in case.
   add paths on;
   graceful restart;  # See comment in kernel section about graceful restart.
@@ -93,10 +93,10 @@ protocol bgp {{.Name}} from bgp_template {
 {{- else if .TTLSecurity}}
   ttl security {{.TTLSecurity}};
 {{- end}}
-{{- if .LocalAsNumber}}
-  local as {{.LocalAsNumber}};
+{{- if .LocalASNumber}}
+  local as {{.LocalASNumber}};
 {{- end}}
-  neighbor {{.IP}}{{if .Port}} port {{.Port}}{{end}} as {{.AsNumber}};
+  neighbor {{.IP}}{{if .Port}} port {{.Port}}{{end}} as {{.ASNumber}};
 {{- if .SourceAddr}}
   source address {{.SourceAddr}};  # The local address we use for the TCP connection
 {{- end}}
--- a/confd/pkg/backends/calico/bgp_processor.go
+++ b/confd/pkg/backends/calico/bgp_processor.go
@@ -20,6 +20,7 @@ import (
 	"sort"
 	"strings"
 
+	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/calico/confd/pkg/backends/types"
@@ -35,30 +36,25 @@ type bgpConfigCache struct {
 	revision uint64
 }
 
-var (
-	configCache   *bgpConfigCache
-	configCacheV6 *bgpConfigCache
-)
+// configCache is indexed by IP version (4 or 6)
+var configCache map[int]*bgpConfigCache
+
+func init() {
+	configCache = make(map[int]*bgpConfigCache)
+}
 
 // GetBirdBGPConfig processes raw datastore data into a clean BGP configuration structure
 // ipVersion should be 4 for IPv4 or 6 for IPv6
 func (c *client) GetBirdBGPConfig(ipVersion int) (*types.BirdBGPConfig, error) {
-	// Use separate caches for IPv4 and IPv6
+	logc := log.WithField("ipVersion", ipVersion)
 	currentRevision := c.GetCurrentRevision()
 
-	if ipVersion == 4 && configCache != nil &&
-		configCache.revision == currentRevision {
-		log.Debug("BGP config cache hit (IPv4), returning cached configuration")
-		return configCache.config, nil
+	if cached, ok := configCache[ipVersion]; ok && cached.revision == currentRevision {
+		logc.Debug("BGP config cache hit, returning cached configuration")
+		return cached.config, nil
 	}
 
-	if ipVersion == 6 && configCacheV6 != nil &&
-		configCacheV6.revision == currentRevision {
-		log.Debug("BGP config cache hit (IPv6), returning cached configuration")
-		return configCacheV6.config, nil
-	}
-
-	log.Debugf("BGP config cache miss or expired, processing new configuration for IPv%d", ipVersion)
+	logc.Debug("BGP config cache miss or expired, processing new configuration")
 
 	config := &types.BirdBGPConfig{
 		NodeName:    NodeName,
@@ -69,15 +65,17 @@ func (c *client) GetBirdBGPConfig(ipVersion int) (*types.BirdBGPConfig, error) {
 
 	// Get basic node configuration
 	if err := c.populateNodeConfig(config, ipVersion); err != nil {
-		log.WithError(err).Warn("Failed to populate node configuration")
+		logc.WithError(err).Warn("Failed to populate node configuration")
+		return nil, err
 	}
-	log.Debugf("Populated node configuration (IPv%d): node=%s, ip=%s, ipv6=%s, as=%s", ipVersion, config.NodeName, config.NodeIP, config.NodeIPv6, config.AsNumber)
+	logc.Debugf("Populated node configuration: node=%s, ip=%s, ipv6=%s, as=%s", config.NodeName, config.NodeIP, config.NodeIPv6, config.ASNumber)
 
 	// Process all peer types
 	if err := c.processPeers(config, ipVersion); err != nil {
-		log.WithError(err).Warnf("Failed to process BGP peers for IPv%d", ipVersion)
+		logc.WithError(err).Warn("Failed to process BGP peers")
+		return nil, err
 	}
-	log.Debugf("Processed BGP peers (IPv%d): found %d peers", ipVersion, len(config.Peers))
+	logc.Debugf("Processed BGP peers: found %d peers", len(config.Peers))
 
 	// Sort peers by name for consistent output ordering
 	sort.Slice(config.Peers, func(i, j int) bool {
@@ -86,24 +84,17 @@ func (c *client) GetBirdBGPConfig(ipVersion int) (*types.BirdBGPConfig, error) {
 
 	// Process community rules
 	if err := c.processCommunityRules(config, ipVersion); err != nil {
-		log.WithError(err).Warnf("Failed to process community rules for IPv%d", ipVersion)
+		logc.WithError(err).Warn("Failed to process community rules")
+		return nil, err
 	}
-	log.Debugf("Processed community rules (IPv%d): found %d rules", ipVersion, len(config.Communities))
+	logc.Debugf("Processed community rules: found %d rules", len(config.Communities))
 
-	// Update appropriate cache
-	if ipVersion == 4 {
-		configCache = &bgpConfigCache{
-			config:   config,
-			revision: currentRevision,
-		}
-		log.Debug("Updated BGP config cache for IPv4")
-	} else {
-		configCacheV6 = &bgpConfigCache{
-			config:   config,
-			revision: currentRevision,
-		}
-		log.Debug("Updated BGP config cache for IPv6")
+	// Update cache
+	configCache[ipVersion] = &bgpConfigCache{
+		config:   config,
+		revision: currentRevision,
 	}
+	logc.Debug("Updated BGP config cache")
 
 	return config, nil
 }
@@ -114,54 +105,48 @@ func (c *client) populateNodeConfig(config *types.BirdBGPConfig, ipVersion int) 
 	nodeIPv4Key := fmt.Sprintf("/calico/bgp/v1/host/%s/ip_addr_v4", NodeName)
 	if nodeIP, err := c.GetValue(nodeIPv4Key); err == nil {
 		config.NodeIP = nodeIP
-		config.RouterID = nodeIP // Default router ID to IPv4 address
 	} else {
-		log.WithError(err).Warnf("Failed to get node IPv4 address from %s", nodeIPv4Key)
+		return fmt.Errorf("failed to get node IPv4 address from %s: %w", nodeIPv4Key, err)
 	}
 
-	// Get node IPv6 address
+	// Get node IPv6 address (optional - not all nodes have IPv6)
 	nodeIPv6Key := fmt.Sprintf("/calico/bgp/v1/host/%s/ip_addr_v6", NodeName)
 	if nodeIPv6, err := c.GetValue(nodeIPv6Key); err == nil {
 		config.NodeIPv6 = nodeIPv6
-	} else {
-		log.WithError(err).Debugf("Failed to get node IPv6 address from %s", nodeIPv6Key)
 	}
 
-	// Get AS number (try node-specific first, then global)
-	nodeASKey := fmt.Sprintf("/calico/bgp/v1/host/%s/as_num", NodeName)
-	if asNum, err := c.GetValue(nodeASKey); err == nil {
-		config.AsNumber = asNum
-	} else if globalAS, err := c.GetValue("/calico/bgp/v1/global/as_num"); err == nil {
-		config.AsNumber = globalAS
-	} else {
-		log.Warnf("Failed to get AS number from node-specific (%s) or global key", nodeASKey)
+	// Get AS number (try node-specific first, then global). Return error if both fail.
+	asNum, err := c.getNodeOrGlobalValue(NodeName, "as_num")
+	if err != nil {
+		return fmt.Errorf("failed to get AS number: %w", err)
+	}
+	config.ASNumber = asNum
+
+	// Get logging configuration. If not found, logLevel will be empty string (uses default).
+	logLevel, err := c.getNodeOrGlobalValue(NodeName, "loglevel")
+	if err == nil {
+		config.LogLevel = logLevel
 	}
 
-	// Get logging configuration
-	var logLevel string
-	nodeLogKey := fmt.Sprintf("/calico/bgp/v1/host/%s/loglevel", NodeName)
-	if level, err := c.GetValue(nodeLogKey); err == nil {
-		logLevel = level
-	} else if globalLog, err := c.GetValue("/calico/bgp/v1/global/loglevel"); err == nil {
-		logLevel = globalLog
-	}
-
-	config.LogLevel = logLevel
-
-	// Compute debug mode based on log level. If logLevel == "none", DebugMode stays empty (no debug output).
-	if logLevel == "debug" {
+	// Compute debug mode based on log level.
+	switch logLevel {
+	case "none":
+		// DebugMode stays empty (no debug output)
+	case "debug":
 		config.DebugMode = "all"
-	} else if logLevel != "none" && logLevel != "" {
-		config.DebugMode = "{ states }"
-	} else if logLevel == "" {
-		// Default behavior when no log level is set
+	default:
+		// Default behavior for empty string or any other log level
 		config.DebugMode = "{ states }"
 	}
 
 	// Handle router ID logic
 	routerID := os.Getenv("CALICO_ROUTER_ID")
 	if routerID == "hash" {
-		config.RouterID = template.HashToIPv4(config.NodeName)
+		hashedID, err := template.HashToIPv4(config.NodeName)
+		if err != nil {
+			return fmt.Errorf("failed to hash node name to IPv4: %w", err)
+		}
+		config.RouterID = hashedID
 		if ipVersion == 6 {
 			config.RouterIDComment = "# Use IP address generated by nodename's hash"
 		}
@@ -169,24 +154,17 @@ func (c *client) populateNodeConfig(config *types.BirdBGPConfig, ipVersion int) 
 		if routerID != "" {
 			config.RouterID = routerID
 		} else if config.NodeIP != "" {
-			config.RouterID = config.NodeIP
+			config.RouterID = config.NodeIP // Default router ID to IPv4 address
 		}
 		if ipVersion == 6 {
 			config.RouterIDComment = "# Use IPv4 address since router id is 4 octets, even in MP-BGP"
 		}
 	}
 
-	// Process bind mode and listen address (matching song-original.cfg.template)
-	var bindMode string
-	bindModeKey := fmt.Sprintf("/calico/bgp/v1/host/%s/bind_mode", NodeName)
-	if mode, err := c.GetValue(bindModeKey); err == nil {
-		bindMode = mode
-	} else if globalMode, err := c.GetValue("/calico/bgp/v1/global/bind_mode"); err == nil {
-		bindMode = globalMode
-	}
-
+	// Process bind mode and listen address
+	bindMode, err := c.getNodeOrGlobalValue(NodeName, "bind_mode")
 	// Set listen address if bind mode is NodeIP and we have a node IP
-	if bindMode == "NodeIP" {
+	if err == nil && bindMode == "NodeIP" {
 		if ipVersion == 6 && config.NodeIPv6 != "" {
 			config.ListenAddress = config.NodeIPv6
 		} else if ipVersion == 4 && config.NodeIP != "" {
@@ -195,24 +173,16 @@ func (c *client) populateNodeConfig(config *types.BirdBGPConfig, ipVersion int) 
 	}
 
 	// Process listen port (node-specific takes precedence over global)
-	listenPortKey := fmt.Sprintf("/calico/bgp/v1/host/%s/listen_port", NodeName)
-	if port, err := c.GetValue(listenPortKey); err == nil {
+	port, err := c.getNodeOrGlobalValue(NodeName, "listen_port")
+	if err == nil {
 		config.ListenPort = port
-	} else if globalPort, err := c.GetValue("/calico/bgp/v1/global/listen_port"); err == nil {
-		config.ListenPort = globalPort
 	}
 
 	// Process ignored interfaces and build complete interface string
-	var ignoredInterfaces string
-	nodeIgnoredKey := fmt.Sprintf("/calico/bgp/v1/host/%s/ignored_interfaces", NodeName)
-	if ifaces, err := c.GetValue(nodeIgnoredKey); err == nil {
-		ignoredInterfaces = ifaces
-	} else if globalIfaces, err := c.GetValue("/calico/bgp/v1/global/ignored_interfaces"); err == nil {
-		ignoredInterfaces = globalIfaces
-	}
+	ignoredInterfaces, err := c.getNodeOrGlobalValue(NodeName, "ignored_interfaces")
 
 	// Build the complete interface pattern string
-	if ignoredInterfaces != "" {
+	if err == nil && ignoredInterfaces != "" {
 		// Parse comma-separated list and build pattern
 		ifaceList := strings.Split(ignoredInterfaces, ",")
 		var patterns []string
@@ -237,17 +207,17 @@ func (c *client) processPeers(config *types.BirdBGPConfig, ipVersion int) error 
 
 	// Process node-to-node mesh peers
 	if err := c.processMeshPeers(config, nodeClusterID, ipVersion); err != nil {
-		log.WithError(err).Warnf("Failed to process mesh peers for IPv%d", ipVersion)
+		return fmt.Errorf("failed to process mesh peers: %w", err)
 	}
 
-	// Process global peers (both regular and local BGP peers)
+	// Process global peers (remote and local BGP peers)
 	if err := c.processGlobalPeers(config, nodeClusterID, ipVersion); err != nil {
-		log.WithError(err).Warnf("Failed to process global peers for IPv%d", ipVersion)
+		return fmt.Errorf("failed to process global peers: %w", err)
 	}
 
-	// Process node-specific peers (both regular and local BGP peers)
+	// Process node-specific peers (remote and local BGP peers)
 	if err := c.processNodePeers(config, nodeClusterID, ipVersion); err != nil {
-		log.WithError(err).Warnf("Failed to process node-specific peers for IPv%d", ipVersion)
+		return fmt.Errorf("failed to process node-specific peers: %w", err)
 	}
 
 	return nil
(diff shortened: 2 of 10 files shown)


confd/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -1,91 +1,30 @@
...
-        {{- if eq $length 2}}
-          bgp_community.add(({{index $i 0}}, {{index $i 1}}));
-        {{- else}}
-          bgp_large_community.add(({{index $i 0}}, {{index $i 1}}, {{index $i 2}}));
-        {{- end}}
-        {{- end}}
-      }
-      {{- end}}
+{{- range $config.Communities}}
+  if (net ~ {{.CIDR}}) then {
```
nelljerram: I'm not sure, but it feels strange to me to keep the `if (net ~ ` and `) then {` here in the template.  Suspect it would be neater to generate those in Go too.

nelljerram: That said, it's already fine as it is, so no need to go further at this stage.

confd/pkg/resource/template/template_funcs.go
```diff
@@ -51,6 +54,24 @@ func addFuncs(out, in map[string]interface{}) {
 	}
 }
 
+// addCalicoFuncs adds Calico-specific template functions
+func addCalicoFuncs(funcMap map[string]interface{}, storeClient interface{}) {
+	// Add getBGPConfig function that takes the client as parameter
+	funcMap["getBGPConfig"] = func(client interface{}) (interface{}, error) {
+		if calicoClient, ok := client.(interface {
+			GetBirdBGPConfig() (*types.BirdBGPConfig, error)
```
nelljerram: Better to define this interface externally.

confd/pkg/resource/template/template_funcs.go
```diff
@@ -51,6 +54,24 @@ func addFuncs(out, in map[string]interface{}) {
...
+		}); ok {
+			config, err := calicoClient.GetBirdBGPConfig()
+			if err != nil {
+				// Return error to fail template execution and prevent broken config
+				return nil, err
+			}
+			return config, nil
+		}
+		return nil, errors.New("client does not support GetBirdBGPConfig")
```
nelljerram, line 64: Is it possible for this to be the case?

confd/pkg/resource/template/resource.go
```diff
@@ -86,6 +86,9 @@ func NewTemplateResource(path string, config Config) (*TemplateResource, error)
 	tr.syncOnly = config.SyncOnly
 	addFuncs(tr.funcMap, tr.store.FuncMap)
 
+	// Add Calico-specific functions if using Calico backend
```
nelljerram: We are always using the Calico backend, so why say "if" here?

confd/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -107,22 +46,13 @@ protocol kernel {
 
 # Watch interface up/down events.
 protocol device {
-{{- template "LOGGING"}}
+{{- template "LOGGING" $config}}
```
nelljerram: I can't find any public doc about passing a variable to a `{{ template ... }}` call.  Presumably this populates the value of `.` inside the template definition - but do you know where that is documented?

song-jiang: Yes. Links here https://pkg.go.dev/text/template#hdr-Actions 
```
{{template "name" pipeline}}
	The template with the specified name is executed with dot set
	to the value of the pipeline.
```

confd/pkg/resource/template/template_funcs.go
song-jiang: No. But I think there is no harm to add an extra error log if we ever make a mistake.

caseydavenport:
Just wanted to pop in and note that the DinD tests we use to verify a lot of this stuff are going to start failing soon as Semaphore VM images are upgraded. So, we'll either need to get this in before that happens or wait until we have successfully re-enabled that coverage in some other way!

confd/tests/test_suite_common.sh
```diff
@@ -822,7 +822,8 @@ compare_templates() {
         fi
         expected=/tests/compiled_templates/${testdir}/${f}
         actual=/etc/calico/confd/config/${f}
-        if ! diff --ignore-blank-lines -q ${expected} ${actual} 1>/dev/null 2>&1; then
+
+        if ! diff_files_ignoring_comments_and_blank_lines ${expected} ${actual} ${output}; then
```
nelljerram: We shouldn't handle `${output}` - in terms of displaying the remaining diffs - both inside `diff_files_ignoring_comments_and_blank_lines` and just below.

Would the correct resolution be to remove the `diff ${expected} ${actual}` line below?

It would be nice to see what a failure case now looks like - could you provide an example of that?

confd/tests/test_suite_common.sh
```diff
@@ -863,6 +864,40 @@ compare_templates() {
     return $rc
 }
 
+# Compares two files ignoring comments (lines starting with #) and blank lines
```
nelljerram, line 867: This begs some questions:
1. Do we want to have comments, at all, in our generated BIRD configs?
2. If we do think some comments are useful, why would we **not** want to verify those?

I think the answer to (1) is No.  IMO we have put comments here before because these files were effectively acting as code, and coding comments are helpful for developers.  Now, the corresponding comments will be (or should be) in bgp_processor.go.  I don't think comments need to exist at all (e.g. for our customers) in the generated BIRD configs.  If customers want to understand details of our configs, they can read bgp_processor.go to see that.

However, I know that I also said "it will be much easier to review **this** PR if the configs are not changing at all".  Therefore, if you agree with the above, I suggest:
- let's review and merge this PR with the comments still in place, and with `diff_files_ignoring_comments_and_blank_lines` removing the comments
- then, let's make a following PR that removes all the comments and simplifies `diff_files_ignoring_comments_and_blank_lines`

WDYT?

confd/pkg/resource/template/util.go
```diff
@@ -87,3 +89,51 @@ func recursiveFindFiles(root string, pattern string) ([]string, error) {
...
+		return "", err
+	}
+	truncationLen := maxLen - hashStrSize
+	hashStr := fmt.Sprintf("%X", hash.Sum(nil))
+	truncatedName := fmt.Sprintf("%s_%s", name[:truncationLen], hashStr[:hashCharsToUse])
+	return truncatedName, nil
+}
+
+// HashToIPv4 hashes the given string and formats the resulting 4 bytes as an IPv4 address.
+func HashToIPv4(nodeName string) string {
```
nelljerram: This should return an error in the error case below, instead of an empty string.

(I realise you've only moved code that already existed.  But we perhaps couldn't do that before, and now we can.)

confd/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -1,91 +1,30 @@
...
-{{- else if exists "/bgp/v1/global/loglevel"}}
-{{- $logging := getv "/bgp/v1/global/loglevel"}}
-{{- if eq $logging "debug"}}
-  debug all;
-{{- else if ne $logging "none"}}
-  debug { states };
-{{- end}}
-{{- else}}
-  debug { states };
+{{- if .DebugMode}}
```
nelljerram: Should this be `$config.DebugMode`?  If not, where does `.DebugMode` come from?

nelljerram: Oh, I see now.  But I think it would be clearer to inline this in the 3 places where `template "LOGGING"` occurs, using `$config.DebugMode`.

confd/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -107,42 +46,30 @@ protocol kernel {
...
+{{else}}
 # Template for all BGP clients
 template bgp bgp_template {
-{{- $as_key := or (and (exists $node_as_key) $node_as_key) "/bgp/v1/global/as_num"}}
-{{- $node_as_num := getv $as_key}}
-{{- template "LOGGING"}}
+{{- template "LOGGING" $config}}
   description "Connection to BGP peer";
-  local as {{$node_as_num}};
+  local as {{$config.AsNumber}};
```
nelljerram: nit, I think this should be `ASNumber`, because that looks better to me, and for consistency with `NodeIP`

confd/etc/calico/confd/templates/bird.cfg.template
```diff
@@ -156,442 +83,68 @@ template bgp bgp_template {
...
-{{/* Skipping global bgp peer */}}
-{{- else}}
-{{- if $first}}
-# ------------- Global local bgp peers -------------
-{{$first = false}} {{/* Update the flag so it is printed only once */}}
-{{- end}}
-{{- if $data.port}}
-{{- $id = printf "%s_port_%.0f" $id $data.port}}
+{{- if .LocalAsNumber}}
+  local as {{.LocalAsNumber}};
```
nelljerram: Same nit as above.

confd/pkg/resource/template/template_funcs.go
```diff
@@ -31,6 +35,7 @@ func newFuncMap() map[string]interface{} {
 	m["datetime"] = time.Now
 	m["toUpper"] = strings.ToUpper
 	m["toLower"] = strings.ToLower
+	m["title"] = title.String
```
nelljerram: I don't think `title` is used in the templates, so let's revert this.  (Guessing you used it during development but then changed that?)

confd/pkg/resource/template/template_funcs.go
```diff
@@ -39,7 +44,6 @@ func newFuncMap() map[string]interface{} {
 	m["fileExists"] = isFileExist
 	m["base64Encode"] = Base64Encode
 	m["base64Decode"] = Base64Decode
-	m["hashToIPv4"] = hashToIPv4
 	m["bgpFilterFunctionName"] = BGPFilterFunctionName
 	m["bgpFilterBIRDFuncs"] = BGPFilterBIRDFuncs
```
nelljerram, line 42: Do we still need these "bgp..." ones?

confd/pkg/resource/template/template_funcs.go
```diff
@@ -51,6 +55,22 @@ func addFuncs(out, in map[string]interface{}) {
 	}
 }
 
+// addCalicoFuncs adds Calico-specific template functions
+func addCalicoFuncs(funcMap map[string]interface{}, storeClient interface{}) {
```
nelljerram: `storeClient` seems to be unused here.  So I'm not sure how this works.  How does the template provide the `client` arg?

confd/pkg/resource/template/resource.go
```diff
@@ -178,7 +181,7 @@ func (t *TemplateResource) createStageFile() error {
 		return err
 	}
 
-	if err = tmpl.Execute(temp, nil); err != nil {
+	if err = tmpl.Execute(temp, t.storeClient); err != nil {
```
nelljerram, line 184: Does this mean that `t.storeClient` becomes an extra arg for every template function call?

confd/pkg/backends/client.go
```diff
@@ -1,10 +1,15 @@
...
+)
+
 // The StoreClient interface is implemented by objects that can retrieve
 // key/value pairs from a backend store.
 type StoreClient interface {
 	SetPrefixes(keys []string) error
 	GetValues(keys []string) (map[string]string, error)
 	WatchPrefix(prefix string, keys []string, waitIndex uint64, stopChan chan bool) (string, error)
 	GetCurrentRevision() uint64
+	GetBirdBGPConfig(ipVersion int) (*types.BirdBGPConfig, error)
```
nelljerram, line 14: Why have you added this here, given that you also have the test `if calicoClient, ok := client.(types.BirdBGPConfigProvider); ok` below?  I think we need the added method only in `type BirdBGPConfigProvider`

confd/pkg/backends/types/bird_bgp_config.go
```diff
@@ -0,0 +1,68 @@
...
+package types
+
+// BirdBGPConfigProvider is an interface for clients that can provide BIRD BGP configuration
+type BirdBGPConfigProvider interface {
+	GetBirdBGPConfig(ipVersion int) (*BirdBGPConfig, error)
+}
+
+// BirdBGPConfig represents the processed BGP configuration for templates
+type BirdBGPConfig struct {
+	NodeName         string            `json:"node_name"`
```
nelljerram: Do we need the JSON tags, given that the templates do not use them?

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+	config := &types.BirdBGPConfig{
+		NodeName:    NodeName,
+		Peers:       make([]types.BirdBGPPeer, 0),
+		Filters:     make(map[string]string),
+		Communities: make([]types.CommunityRule, 0),
+	}
+
+	// Get basic node configuration
+	if err := c.populateNodeConfig(config, ipVersion); err != nil {
+		log.WithError(err).Warn("Failed to populate node configuration")
```
nelljerram: return err here?

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+	}
+	log.Debugf("Populated node configuration (IPv%d): node=%s, ip=%s, ipv6=%s, as=%s", ipVersion, config.NodeName, config.NodeIP, config.NodeIPv6, config.AsNumber)
+
+	// Process all peer types
+	if err := c.processPeers(config, ipVersion); err != nil {
+		log.WithError(err).Warnf("Failed to process BGP peers for IPv%d", ipVersion)
```
nelljerram: return err here?

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+	log.Debugf("Processed BGP peers (IPv%d): found %d peers", ipVersion, len(config.Peers))
+
+	// Sort peers by name for consistent output ordering
+	sort.Slice(config.Peers, func(i, j int) bool {
+		return config.Peers[i].Name < config.Peers[j].Name
+	})
+
+	// Process community rules
+	if err := c.processCommunityRules(config, ipVersion); err != nil {
+		log.WithError(err).Warnf("Failed to process community rules for IPv%d", ipVersion)
```
nelljerram: return err here?

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+
+// BGP configuration cache
+type bgpConfigCache struct {
+	config   *types.BirdBGPConfig
+	revision uint64
+}
+
+var (
+	configCache   *bgpConfigCache
+	configCacheV6 *bgpConfigCache
```
nelljerram: nit: could use a map indexed by IP version.  That would simplify some of the if elses below.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+
+// populateNodeConfig fills in basic node configuration
+func (c *client) populateNodeConfig(config *types.BirdBGPConfig, ipVersion int) error {
+	// Get node IPv4 address
+	nodeIPv4Key := fmt.Sprintf("/calico/bgp/v1/host/%s/ip_addr_v4", NodeName)
+	if nodeIP, err := c.GetValue(nodeIPv4Key); err == nil {
+		config.NodeIP = nodeIP
+		config.RouterID = nodeIP // Default router ID to IPv4 address
+	} else {
+		log.WithError(err).Warnf("Failed to get node IPv4 address from %s", nodeIPv4Key)
```
nelljerram: Return error.  I'll stop commenting this, but review and apply everywhere.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+	return config, nil
+}
+
+// populateNodeConfig fills in basic node configuration
+func (c *client) populateNodeConfig(config *types.BirdBGPConfig, ipVersion int) error {
+	// Get node IPv4 address
+	nodeIPv4Key := fmt.Sprintf("/calico/bgp/v1/host/%s/ip_addr_v4", NodeName)
+	if nodeIP, err := c.GetValue(nodeIPv4Key); err == nil {
+		config.NodeIP = nodeIP
+		config.RouterID = nodeIP // Default router ID to IPv4 address
```
nelljerram: Don't do this here because RouterID is handled below, including this default again.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+
+	// Compute debug mode based on log level. If logLevel == "none", DebugMode stays empty (no debug output).
+	if logLevel == "debug" {
+		config.DebugMode = "all"
+	} else if logLevel != "none" && logLevel != "" {
+		config.DebugMode = "{ states }"
+	} else if logLevel == "" {
+		// Default behavior when no log level is set
+		config.DebugMode = "{ states }"
+	}
```
nelljerram, line 140: A bit confusing.  Reorder as: if "none", else if "debug", else ...

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+		}
+	}
+
+	// Process listen port (node-specific takes precedence over global)
+	listenPortKey := fmt.Sprintf("/calico/bgp/v1/host/%s/listen_port", NodeName)
+	if port, err := c.GetValue(listenPortKey); err == nil {
+		config.ListenPort = port
+	} else if globalPort, err := c.GetValue("/calico/bgp/v1/global/listen_port"); err == nil {
+		config.ListenPort = globalPort
+	}
```
nelljerram, line 179: Probably worth creating a utility function for looking at both a node-specific and a global key.  But feel free to leave that to later if you prefer.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+func (c *client) processPeers(config *types.BirdBGPConfig, ipVersion int) error {
+	// Get node's route reflector cluster ID
+	nodeClusterID, _ := c.GetValue(fmt.Sprintf("/calico/bgp/v1/host/%s/rr_cluster_id", NodeName))
+
+	// Process node-to-node mesh peers
+	if err := c.processMeshPeers(config, nodeClusterID, ipVersion); err != nil {
+		log.WithError(err).Warnf("Failed to process mesh peers for IPv%d", ipVersion)
+	}
+
+	// Process global peers (both regular and local BGP peers)
```
nelljerram: What do "regular" and "local" mean?

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+	// If this node is a route reflector, skip mesh processing
+	if nodeClusterID != "" {
+		log.Infof("Node %s is a route reflector with cluster ID %s, skipping mesh", NodeName, nodeClusterID)
+		return nil
+	}
+
+	// Check if mesh is enabled
+	meshConfigValue, err := c.GetValue("/calico/bgp/v1/global/node_mesh")
+	if err != nil {
+		return nil // No mesh configuration
```
nelljerram, line 239: I think this should return the error, no?

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+	}
+
+	var meshConfig map[string]interface{}
```
nelljerram: Probably worth defining the required struct.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+		return err
+	}
+
+	// Extract unique hosts from the keys
+	hostsMap := make(map[string]bool)
+	for key := range hostIPsMap {
+		// Keys are like /calico/bgp/v1/host/<hostname>/ip_addr_v4
+		parts := strings.Split(key, "/")
+		if len(parts) >= 6 {
+			hostsMap[parts[5]] = true
```
nelljerram: You can store the actual value here instead of true, to save using GetValue again just below.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+
+		// Create mesh peer name based on IP version
+		var peerName string
+		if ipVersion == 4 {
+			peerName = fmt.Sprintf("Mesh_%s", strings.ReplaceAll(peerIP, ".", "_"))
+		} else {
+			peerName = fmt.Sprintf("Mesh_%s", strings.ReplaceAll(peerIP, ":", "_"))
+		}
+
+		// Mesh peers have same AS, so use true for calico_export_to_bgp_peers
```
nelljerram: What does this "true" value mean?

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+		if peerClusterID != "" {
+			log.Debugf("Skipping %s as it is a route reflector", peerIP)
+			continue
+		}
+
+		// Get peer's AS number
+		peerASKey := fmt.Sprintf("/calico/bgp/v1/host/%s/as_num", host)
+		peerAS, err := c.GetValue(peerASKey)
+		if err != nil {
+			peerAS = config.AsNumber // Use global AS
```
nelljerram: Most fields are not defaulted at this stage, so it would be more consistent to leave this until later.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+	if err != nil {
+		log.WithError(err).Debugf("No global IPv%d peers found or error retrieving them", ipVersion)
+		return nil
+	}
+
+	log.Debugf("Found %d global IPv%d peer entries", len(kvPairs), ipVersion)
+
+	for key, value := range kvPairs {
+		log.Debugf("Processing global peer key: %s", key)
+		var peerData map[string]interface{}
```
nelljerram: Worth defining a proper struct.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+			log.Debugf("Added global peer: %s", peer.Name)
+		} else {
+			log.Debugf("buildPeerFromData returned nil for key %s", key)
+		}
+	}
+
+	// Process global local BGP peers
+	for _, value := range kvPairs {
+		var peerData map[string]interface{}
+		if err := json.Unmarshal([]byte(value), &peerData); err != nil {
```
nelljerram: Inelegant to unmarshal for a second time.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+			if isLocal, _ := peerData["local_bgp_peer"].(bool); !isLocal {
+				continue
+			}
+
+			peer := c.buildPeerFromData(peerData, "Local_Workload", config, nodeClusterID, ipVersion)
+			if peer != nil {
+				config.Peers = append(config.Peers, *peer)
+			}
+		}
+	}
```
nelljerram, line 426: Lots of common code between here and processGlobalPeers - can you easily deduplicate that?

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+
+	peer := &types.BirdBGPPeer{
+		Name: peerName,
+		IP:   peerIP,
+		Type: strings.ToLower(prefix),
+	}
+
+	// Basic fields
+	if port, ok := raw["port"].(float64); ok && port > 0 {
```
nelljerram: Yes, definitely want a proper struct here.  In fact I think we already have one: `bgpPeer` in client.go.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+	if ipVersion == 4 {
+		peerName = fmt.Sprintf("%s_%s", prefix, strings.ReplaceAll(peerIP, ".", "_"))
+	} else {
+		peerName = fmt.Sprintf("%s_%s", prefix, strings.ReplaceAll(peerIP, ":", "_"))
+	}
+	if port, ok := raw["port"].(float64); ok && port > 0 {
+		peerName = fmt.Sprintf("%s_port_%.0f", peerName, port)
+	}
+
+	peer := &types.BirdBGPPeer{
```
nelljerram: I wonder if we can embed `bgpPeer` here, instead of defining a new set of equivalent fields.  But happy to leave that to later work.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+	}
+	// Legacy keep_next_hop field - only apply for eBGP peers
+	if keepNH, ok := raw["keep_next_hop"].(bool); ok {
+		if keepNH && peer.AsNumber != effectiveNodeAS {
+			peer.NextHopKeep = true
+		}
+	}
+
+	// Route reflector handling
+	// If the peer is eBGP and has a different cluster ID than the current node,
```
nelljerram:
```suggestion
	// If the peer is iBGP and has a different cluster ID than the current node,
```

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+		filterSuffix = "V6"
+	}
+
+	// Process BGP filters
+	if filters, ok := raw["filters"].([]interface{}); ok {
+		for _, filter := range filters {
+			if filterName, ok := filter.(string); ok {
+				filterKey := fmt.Sprintf("/calico/resources/v3/projectcalico.org/bgpfilters/%s", filterName)
+				if filterValue, err := c.GetValue(filterKey); err == nil {
+					var filterSpec map[string]interface{}
```
nelljerram: Let's define a proper struct, or re-use one that we might already have.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -0,0 +1,811 @@
...
+		return nil
+	}
+
+	kvPairs, err := c.GetValues([]string{communitiesKey})
+	if err != nil {
+		return err
+	}
+
+	for _, value := range kvPairs {
+		var advertisements []map[string]interface{}
```
nelljerram: Same again.

confd/pkg/resource/template/util.go
song-jiang: That's a very good point. Another benefit of having the business logic being processed in go.

confd/tests/test_suite_common.sh
song-jiang: Agree what you have suggested. The `Two PRs` approach should work well in this case.

confd/pkg/resource/template/template_funcs.go
song-jiang: `bgpFilterFuncs` is still needed in the template to populate BGP filters. I'll remove the registration for `bgpFilterFunctionName`. With the new template, this function is for internal use only.

confd/pkg/resource/template/template_funcs.go
song-jiang: Yes, the `storeClient` is not used and should be removed. 
- `tmpl.Execute(temp, t.storeClient)` makes t.storeClient available as . in the template.
- Template calls {{- $config := getBGPConfig 4 . -}} to passing . explicitly.
- `getBGPConfig` receives it as the client parameter.

confd/pkg/resource/template/resource.go
song-jiang: No. The data parameter passed to tmpl.Execute(writer, data) becomes the "dot" context, only accessible where the template explicitly references `.`.
So basically 
{{- $config := getBGPConfig 4 . -}} means the template calls `t.storeClient.GetBGPConfig(4)` and saves the return value to $config.

confd/pkg/backends/client.go
song-jiang: We do need to add `GetBirdBGPConfig` method here as the function also calls other interface methods (e.g. GetValues). I think it is a mistake that we have a separate `types.BirdBGPConfigProvider` type. We don't need it at all. The code should be updated to 
```
   		if storeClient, ok := client.(backends.StoreClient); ok {
			config, err := storeClient.GetBirdBGPConfig(ipVersion)
```

confd/pkg/backends/types/bird_bgp_config.go
song-jiang: No we don't.

confd/pkg/backends/calico/bgp_processor.go
song-jiang: Good point. Utility function added.

confd/pkg/backends/calico/bgp_processor.go
song-jiang: Updated to ""remote and local BGP peers"

confd/pkg/backends/calico/bgp_processor.go
song-jiang: It should return nil because it's not an err to skip if mesh is not enabled. I've updated the comment to make it clear.
```
	// Skip mesh processing if not enabled
	meshConfigValue, err := c.GetValue("/calico/bgp/v1/global/node_mesh")
	if err != nil {
		return nil // No mesh configuration
	}
```

confd/pkg/backends/calico/bgp_processor.go
song-jiang: Good point.

confd/tests/test_suite_common.sh
song-jiang: Here is an [_example_ ](https://tigera.semaphoreci.com/jobs/6ae64d1f-dee5-4fff-9ea7-4342b568e682)
```
Populating calico with test data using calicoctl:  explicit_peering/route_reflector02:03
Successfully applied 8 resource(s)02:03
Starting Typha02:03
Comparing with templates in explicit_peering/route_reflector02:03
comparing templates attempt 102:03
comparing templates attempt 202:05
comparing templates attempt 302:06
comparing templates attempt 402:07
comparing templates attempt 502:08
comparing templates attempt 602:09
comparing templates attempt 702:10
comparing templates attempt 802:11
comparing templates attempt 902:12
comparing templates attempt 1002:13
47,48d4602:14
<   rr client;02:14
<   rr cluster id 10.0.0.1;02:14
62,63d5902:14
<   rr client;02:14
<   rr cluster id 10.0.0.1;02:14
Failed: bird.cfg templates do not match, showing diff of expected vs received02:14
62,77c6202:14
< 02:14
< # ------------- Node-to-node mesh -------------02:14
< # This node (kube-master) is configured as a route reflector with cluster ID 10.0.0.1;02:14
< # ignore node-to-node mesh setting.02:14
< 02:14
< 02:14
< # ------------- Global peers -------------02:14
< # No global peers configured.02:14
< 02:14
< 02:14
< # ------------- Node-specific peers -------------02:14
< 02:14
< 02:14
< 02:14
< 02:14
< # For peer /bgp/v1/host/kube-master/peer_v4/10.192.0.302:14
---02:14
> # BGP Protocol Configurations02:14
90,91d7402:14
<   rr client;02:14
<   rr cluster id 10.0.0.1;02:14
93,95d7502:14
< 02:14
< 02:14
< # For peer /bgp/v1/host/kube-master/peer_v4/10.192.0.402:14
108,109d8702:14
<   rr client;02:14
<   rr cluster id 10.0.0.1;02:14
111,113d8802:14
< 02:14
< 02:14
< # For peer /bgp/v1/host/kube-master/peer_v4/172.19.4.8702:14
127d10102:14
< 02:14
Copying confd rendered output to /tests/logs/kdd/rendered/bird.cfg02:14
60,61d5902:14
<   rr client;02:14
<   rr cluster id 10.0.0.1;02:14
75,76d7202:14
<   rr client;02:14
<   rr cluster id 10.0.0.1;02:14
Failed: bird6.cfg templates do not match, showing diff of expected vs received02:14
46d4502:14
< 02:14
62,77c6102:14
< 02:14
< # ------------- Node-to-node mesh -------------02:14
< # This node (kube-master) is configured as a route reflector with cluster ID 10.0.0.1;02:14
< # ignore node-to-node mesh setting.02:14
< 02:14
< 02:14
< # ------------- Global peers -------------02:14
< # No global peers configured.02:14
< 02:14
< 02:14
< # ------------- Node-specific peers -------------02:14
< 02:14
< 02:14
< 02:14
< 02:14
< # For peer /bgp/v1/host/kube-master/peer_v6/ac13::57-5002:14
---02:14
> # BGP Protocol Configurations02:14
91,93d7402:14
< 02:14
< 02:14
< # For peer /bgp/v1/host/kube-master/peer_v6/fe0a::302:14
106,107d8602:14
<   rr client;02:14
<   rr cluster id 10.0.0.1;02:14
109,111d8702:14
< 02:14
< 02:14
< # For peer /bgp/v1/host/kube-master/peer_v6/fe0a::402:14
124,125d9902:14
<   rr client;02:14
<   rr cluster id 10.0.0.1;02:14
127d10002:14
< 02:14
Copying confd rendered output to /tests/logs/kdd/rendered/bird6.cfg02:14
Recording failed testcase directory to /tests/logs/kdd/testcase_directory.txt
```

confd/pkg/backends/calico/bgp_processor.go
song-jiang: I'm not sure I understand what you mean by `leave this until later`. I don't see much benefit in moving this code to after getting `peerListenPortKey` & `peerName`.

confd/tests/test_suite_common.sh
song-jiang: Yes, it's a mistake. The code outputted same diff twice. I've fixed it.

confd/pkg/resource/template/resource.go
nelljerram: Thanks for explaining that.

confd/pkg/backends/calico/bgp_processor.go
nelljerram: I think I understand now that "local" is for nested BGP, and means peering to a nested node inside a local endpoint.

confd/pkg/backends/calico/bgp_processor.go
nelljerram: What I meant is that the previous template code assumed that `/calico/bgp/v1/global/node_mesh` always exists.  So this code can assume that too.

confd/pkg/backends/calico/bgp_processor.go
nelljerram: This comment still needs a response, please.

confd/pkg/backends/calico/bgp_processor.go
nelljerram: I'm sorry, I think I misunderstood.  This is fine as is.

confd/pkg/backends/calico/bgp_processor.go
nelljerram: I see you made a different (and better) change here.

confd/pkg/backends/calico/bgp_processor.go
nelljerram: (but definitely not a blocking concern!)

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -303,20 +276,21 @@ func (c *client) processMeshPeers(config *types.BirdBGPConfig, nodeClusterID str
...
+			}
 		}
 	}
 
-	for host := range hostsMap {
-		peerIPKey := fmt.Sprintf("/calico/bgp/v1/host/%s/%s", host, ipAddrSuffix)
-		peerIP, err := c.GetValue(peerIPKey)
-		if err != nil || peerIP == "" {
+	for host, peerIP := range hostsMap {
+		if peerIP == "" {
```
nelljerram, line 293: nit: is it possible now for `peerIP` to be empty here?

confd/pkg/backends/calico/bgp_processor.go
nelljerram: Sorry, I see you've covered this in the markups.

confd/pkg/backends/calico/bgp_processor_test.go
```diff
@@ -1444,7 +1440,7 @@ func TestProcessGlobalPeers_BasicPeer(t *testing.T) {
 
 	peerData := map[string]interface{}{
```
nelljerram, line 1441: nit: maybe we should use bgpPeer here too.  But that might need it to become public.

nelljerram (commented):
LGTM.  A few remaining nits, but I'm happy to leave those to your judgement, either in this PR or in follow on work.

nelljerram (approved):
D'oh, forgot to click "Approve" 😆

confd/pkg/backends/calico/bgp_processor.go
song-jiang: Maybe not in reality but from this function point of view, the value (peerIP) of the KV could be empty.

PR #11159, opened 2025-10-09
[BPF] Maglev load balancing for external traffic

aaaaaaaalex:
## Description

Maglev loadbalancing makes sure that a service's backend selection is the same across all nodes, so long as the incoming packet's 5-tuple is the same, and the set of backends does not churn, and almost consistent when the churn is small. 

This allows a connection originally handled by one LB'ing node to be handed off to another node if the original node crashed or became unavailable. This ensures continuous service in case of node downtime without breaking connections and minimizing disruption. To make a service use Maglev LB, the service needs to be annotated. This is intended only for services exposed outside of a cluster like ExternlIP, LoadBalancer type or ClusterIP as long as they are reachable.

NodePort type services are excluded as the client is tied to a specific node and would not benefit from a failover.

The work is based on [Maglev: A Fast and Reliable Software Network Load Balancer](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/44824.pdf) research paper by Google.

This PR makes the necessary changes to existing TC programs to allow:
 - packets to maglev services to have their backends selected differently
 - mid-flow packets without corresponding CT entries to be checked as Maglev failovers
 - recheck policy and recreate conntrack entries on failover nodes 

FelixConfiguration allows users to declare how many Maglev endpoints and services they will have, so that Felix can provision space for the lookup tables.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
BPF: Adds support for Maglev-style service loadbalancer, which enables service traffic failover to another loadbalancer in case of a failure.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/bpf_test.go
+++ b/felix/fv/bpf_test.go
@@ -2140,15 +2140,21 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 						externalClient.Exec(ipRouteReplace...)
 
 						lastPongCount := pc.PongCount()
-						Eventually(pc.PongCount, "5s", "100ms").Should(BeNumerically(">", lastPongCount), "Connection is no longer ponging after route failover")
 
-						ctVal, ctExists = checkConntrackExistsAnyDirection(tc.Felixes[2], clientIPAddr, uint16(pc.SourcePort), backingPodIPAddr, uint16(tgtPort), family)
-						Expect(ctExists).To(BeTrue(), "Conntrack didn't exist on Felix[2] for failover traffic. Did the failover actually occur?")
+						checkCTExistsFn := func() bool {
+							_, ctExists = checkConntrackExistsAnyDirection(tc.Felixes[2], clientIPAddr, uint16(pc.SourcePort), backingPodIPAddr, uint16(tgtPort), family)
+							return ctExists
+						}
+						Eventually(checkCTExistsFn, "10s").Should(BeTrue(), "Conntrack didn't exist on Felix[2] for failover traffic. Did the failover actually occur?")
 
 						// Check the backing node updated conntrack tun_ip to the new loadbalancer node.
 						ctVal, ctExists = checkConntrackExistsAnyDirection(tc.Felixes[0], clientIPAddr, uint16(pc.SourcePort), backingPodIPAddr, uint16(tgtPort), family)
 						Expect(ctExists).To(BeTrue(), "Conntrack didn't exist on backing Felix[0].")
 						Expect(ctVal.Data().TunIP.String()).To(Equal(felixIP(2)), "Backing node did not update its conntrack tun_ip to the new loadbalancer IP")
+
+						// Connection should persist after the changeover.
+						Eventually(pc.PongCount, "5s", "100ms").Should(BeNumerically(">", lastPongCount), "Connection is no longer ponging after route failover")
+
 					}
 
 					It("should maintain connections to a cluster IP across loadbalancer failure using maglev", func() { testFailover(clusterIP) })


libcalico-go/lib/consistenthash/primes_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+var _ = DescribeTable("RulesAPIToBackend",
+	func(input int, expected int) {
+		Expect(consistenthash.NearestPrimeUint16(input)).To(ConsistOf(expected))
+	},
+	Entry("Negative number should return 2", -1, 2),
+	Entry("0 should return 2", 0, 2),
+	Entry("Numbers higher than 65521 should return 65521", 65525, 65521),
+	Entry("Numbers should get the closest prime (1/2)", 31742, 31741),
+	Entry("Numbers shoudl get the closest prime (2/2)", 31740, 31741),
```
tomastigera: this essentially tests whether sort.Search works correctly ;-)

libcalico-go/lib/consistenthash/primes_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/projectcalico/calico/libcalico-go/lib/consistenthash"
+)
+
+var _ = DescribeTable("RulesAPIToBackend",
+	func(input int, expected int) {
+		Expect(consistenthash.NearestPrimeUint16(input)).To(ConsistOf(expected))
+	},
+	Entry("Negative number should return 2", -1, 2),
+	Entry("0 should return 2", 0, 2),
+	Entry("Numbers higher than 65521 should return 65521", 65525, 65521),
```
tomastigera: I think these 3 should return an error. Perhaps the 0 case is ok.

felix/bpf-gpl/conntrack.h
```diff
@@ -787,8 +787,14 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_lookup(struct cali_tc_c
 		if (CALI_F_FROM_HEP && !ip_void(ctx->state->tun_ip) && !ip_void(result.tun_ip) &&
 				!ip_equal(result.tun_ip, ctx->state->tun_ip)) {
 			CALI_CT_DEBUG("tunnel src changed from " IP_FMT " to " IP_FMT "",
-					debug_ip(result.tun_ip), debug_ip(ctx->state->tun_ip));
+				debug_ip(result.tun_ip), debug_ip(ctx->state->tun_ip));
```
tomastigera: Your IDE is mangling indentation. It is tabs

felix/bpf-gpl/conntrack_types.h
```diff
@@ -40,6 +40,7 @@ enum cali_ct_type {
 #define CALI_CT_FLAG_NP_NO_DSR	0x2000 /* marks connections from a client which is excluded from DSR */
 #define CALI_CT_FLAG_SKIP_REDIR_PEER	0x4000 /* marks connections from a client which is excluded from redir */
 #define CALI_CT_FLAG_SET_DSCP	0x8000 /* marks connections that needs to set DSCP */
+#define CALI_CT_FLAG_MAGLEV	0X10000 /* marks connections from a maglev LB. On a backing node, allows packets of an existing to arrive via a different tunnel for failovers. */
```
tomastigera:
```suggestion
#define CALI_CT_FLAG_MAGLEV	0X10000 /* marks maglev connections. Allows packets of an existing existing to arrive via a different tunnel after failover. */
```

felix/bpf-gpl/conntrack_types.h
```diff
@@ -154,9 +155,9 @@ struct ct_create_ctx {
 	ipv46_addr_t tun_ip; /* is set when the packet arrive through the NP tunnel.
 			* It is also set on the first node when we create the
 			* initial CT entry for the tunneled traffic. */
-	__u16 flags;
+	__u32 flags;
 	__u8 proto;
-	__u8 __pad;
+	__u8 __pad[3];
```
tomastigera, line 160: Wasn't this supposed to pick ut the changes from master without making a breaking change that would require an upgrade?

felix/bpf-gpl/counters.h
```diff
@@ -8,7 +8,7 @@
 #include "bpf.h"
 #include "types.h"
 
-#define MAX_COUNTERS_SIZE 25
+#define MAX_COUNTERS_SIZE 27
```
tomastigera: does this PR add only one counter?

felix/bpf-gpl/nat_lookup.h
```diff
@@ -104,6 +104,13 @@ static CALI_BPF_INLINE struct calico_nat_dest* calico_nat_lookup(ipv46_addr_t *i
 		}
 		CALI_DEBUG("NAT: nodeport hit");
 	}
+
+	if (HAS_MAGLEV && (nat_lv1_val->flags & NAT_FLG_MAGLEV)) {
+		CALI_DEBUG("NAT: 1st level hit; id=%d, NAT maglev", nat_lv1_val->id);
+		*res = NAT_MAGLEV;
+		return NULL;
```
tomastigera:
```suggestion
		return NULL; /* Make the lookup fail, but signal, that another lookup should be done for maglev */
```

felix/bpf-gpl/nat_types.h
```diff
@@ -121,4 +123,23 @@ struct vxlanhdr {
...
+	ipv46_addr_t vip;
+	__u16 port;
+	__u8 proto;
+	__u8 pad;
+	__u32 ordinal; // should always be a value of [0..M-1], where M is a very large prime number. -Alex
+};
+
+// 1009 * 1000 gives a large lookup table for up to 1000 services. -Alex
+#ifdef IPVER6
+CALI_MAP_NAMED(cali_v6_ch, cali_ch,,
```
tomastigera:
```suggestion
CALI_MAP_NAMED(cali_v6_mglv, cali_maglev,,
```

felix/bpf-gpl/nat_types.h
```diff
@@ -121,4 +123,23 @@ struct vxlanhdr {
 	__be32 flags;
 	__be32 vni;
 };
+
+struct calico_ch_key {
```
tomastigera: Imo we do not really need to prepend with cali/calico we use cali_ mostly. But I would just call it maglev instead of ch so it is obvious what it is.

```suggestion
struct cali_maglev_key {
```

felix/bpf-gpl/tc.c
```diff
@@ -353,11 +354,14 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
...
-	if (CALI_F_FROM_HEP && ct_result_tun_src_changed(ctx->state->ct_result.rc)) {
-		CALI_DEBUG("dropping tunnel pkt with changed source node");
+	if (CALI_F_FROM_HEP
+			&& ct_result_tun_src_changed(ctx->state->ct_result.rc)
+			&& !(ctx->state->ct_result.flags & CALI_CT_FLAG_MAGLEV)) {
+		CALI_DEBUG("dropping non-maglev tunnel pkt with changed source node");
 		goto deny;
 	}
 
+
```
```suggestion
```

felix/bpf-gpl/tc.c
```diff
@@ -460,6 +454,28 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 						     !ip_void(ctx->state->tun_ip), &nat_res);
 	}
 
+	if (nat_res == NAT_MAGLEV) {
```
tomastigera: I think this should be all wrapped in `if (CALI_F_FROMHEP) {` so that it is not compiled in other places. All other places should not see handling maglev.

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+func WithHash(hash1, hash2 hash.Hash) func(*ConsistentHash) {
+	return func(c *ConsistentHash) {
+		c.h1 = hash1
+		c.h2 = hash2
+	}
+}
+
+// WithLUTLength configures the LUT-size and subsequently,
+// the preference-list length for each backend.
+// Should be a prime number.
```
tomastigera:
```suggestion
// Must be a prime number.
```

felix/bpf/nat/maps.go
```diff
@@ -363,6 +369,150 @@ func BackendValueFromBytes(b []byte) BackendValueInterface {
...
+
+var _ MaglevBackendKeyInterface = MaglevBackendKey{}
+
+const maglevBackendValueSize = backendValueSize
+
+var MaglevMapParameters = maps.MapParameters{
+	Type:       "hash",
+	KeySize:    MaglevBackendKeySize,
+	ValueSize:  maglevBackendValueSize,
+	MaxEntries: 1009 * 1000,
```
tomastigera:
```suggestion
	MaxEntries: consistenthash.M * 1000,
```

felix/bpf/proxy/options.go
```diff
@@ -87,6 +87,32 @@ func WithIPFamily(ipFamily int) Option {
...
+		p, ok := P.(*KubeProxy)
+		if !ok {
+			return nil
+		}
+
+		p.maglevLUTSize = MaglevLUTSizeDefault
+		return nil
+	}
+}
+
```
tomastigera: why this?

felix/bpf/proxy/proxy.go
```diff
@@ -394,16 +394,23 @@ const (
 	ReapTerminatingUDPImmediatelly = "TerminatingImmediately"
 
 	ExcludeServiceAnnotation = "projectcalico.org/natExcludeService"
+	ExternalTrafficStrategy  = "lb.projectcalico.org/external-traffic-strategy"
```
tomastigera:
```suggestion
	ExternalTrafficStrategyAnnotation  = "lb.projectcalico.org/external-traffic-strategy"
```

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+func WithPreferenceLength(m int) func(*ConsistentHash) {
```
tomastigera: why preference?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,215 @@
...
+	c.backendsByName = make(map[string]backend)
+
+	for _, option := range o {
+		option(c)
+	}
+
+	if c.h1 == nil || c.h2 == nil {
+		panic("nil hashing function for ConsistentHash")
+	}
+
```
tomastigera, line 45: if both must be specified, it is not really an option and should be part of the New args

libcalico-go/lib/consistenthash/primes_test.go
aaaaaaaalex: Yup :laughing: but, might be helpful if we ever mess with the code under test in the future (tests as documentation I guess?)

libcalico-go/lib/consistenthash/primes_test.go
aaaaaaaalex: The validation layer above in the Felixconfig definitions can catch those values :+1:

felix/bpf-gpl/conntrack.h
aaaaaaaalex: They're still tabs here, just a different number of them. I think that was probably me playing with the line, will correct it

felix/bpf-gpl/conntrack_types.h
aaaaaaaalex: This is in the create_ctx struct, right? I didn't know that struct had migration code/requirements. We might need a sync-up on this?

felix/bpf-gpl/counters.h
aaaaaaaalex: Hmm it seems like this should be 26 alright - I added `CALI_RESON_MAGLEV_NO_BACKEND`.
Not sure if I'm meant to count `RESERVED1` in the count - counting that would bring the number to 27

felix/bpf-gpl/tc.c
aaaaaaaalex: I've amended this line now - LMK if it's OK

felix/bpf/nat/maps.go
aaaaaaaalex: Question on this - can we dynamically change this number later on, i.e. when we read Felix config params?
If this is a harder limit that we can't change, I think I should set it to the highest size we can ever expect

felix/bpf/proxy/options.go
aaaaaaaalex: Obsolete - deleted

felix/bpf/consistenthash/consistenthash.go
aaaaaaaalex: The LUT size and the preference lists generated for each backend are the same size, so the two numbers are the same, though I've changed it to be called LUTSize now, since it's more consistent

felix/bpf-gpl/conntrack_types.h
tomastigera: ohhh yeah that is private, ignore me :+1:

felix/bpf-gpl/counters.h
tomastigera: I count 26. You added one so 25+1. RESERVED1 is included

felix/bpf/nat/maps.go
tomastigera: yes, can be done dynamically

felix/bpf/proxy/syncer.go
```diff
@@ -550,6 +549,7 @@ func (s *Syncer) applyDerived(
 	sname k8sp.ServicePortName,
 	t svcType,
 	sinfo Service,
+	maglevEPs []k8sp.Endpoint,
```
tomastigera, line 552: derived should be pointing to the same backends as the primary service. Therefore we do not pass in the endpoints. All frontends (clusterIP, nodeport, etc.) are pointing to the same set of backends. Is there a reason for maglev to be different?

libcalico-go/lib/consistenthash/constants.go
```diff
@@ -0,0 +1,7 @@
+package consistenthash
+
+const (
+	MaglevEndpointLUTFactor      = 5
+	MaglevMaxEndpointsPerService = 3000
```
tomastigera: Isn't that a bit excessive default? I'd expect it somewhere below 100 :thinking:

felix/bpf/proxy/syncer.go
```diff
@@ -747,9 +799,8 @@ func (s *Syncer) Apply(state DPSyncerState) error {
 	return s.cleanupSticky()
 }
 
-func (s *Syncer) updateService(skey svcKey, sinfo Service, id uint32, eps []k8sp.Endpoint) (int, int, error) {
+func (s *Syncer) updateService(skey svcKey, sinfo Service, id uint32, eps []k8sp.Endpoint, maglevEPs []k8sp.Endpoint) (int, int, error) {
```
tomastigera, line 802: feels like you can generate the maglev eps inside this function

felix/bpf/proxy/syncer.go
```diff
@@ -797,6 +848,11 @@ func (s *Syncer) updateService(skey svcKey, sinfo Service, id uint32, eps []k8sp
 		flags |= nat.NATFlgInternalLocal
 	}
 
+	if sinfo.UseMaglev() && maglevEPs != nil {
```
tomastigera, line 851: When generating the backends, you should probably filter out those that are not ready `if ep.IsReady() {`

felix/bpf/proxy/syncer.go
```diff
@@ -797,6 +848,11 @@ func (s *Syncer) updateService(skey svcKey, sinfo Service, id uint32, eps []k8sp
 		flags |= nat.NATFlgInternalLocal
```
tomastigera, line 848: Does the external traffic policy also apply to maglev? Probably not, because that feels mutually exclusive, but it should be called out in docs

felix/bpf/proxy/syncer.go
```diff
@@ -816,6 +872,29 @@ func (s *Syncer) updateService(skey svcKey, sinfo Service, id uint32, eps []k8sp
...
 
+func (s *Syncer) newConsistentHash() *consistenthash.ConsistentHash {
+	return consistenthash.New(
+		s.maglevLUTSize,
+		fnv.New32(), fnv.New32(),
+	)
+}
+
+// writeMaglevBackends takes frontend info, and a pre-populated CH module (backends already programmed).
+func (s *Syncer) writeMaglevSvcBackends(skey svcKey, sinfo Service, maglevEPs []k8sp.Endpoint) {
```
tomastigera, line 883: you may probably pass all the valid endpoints all the way here and generate and write the set here, right?

felix/bpf/proxy/syncer.go
```diff
@@ -816,6 +872,29 @@ func (s *Syncer) updateService(skey svcKey, sinfo Service, id uint32, eps []k8sp
...
+	port := uint16(sinfo.Port())
+	proto := ProtoV1ToIntPanic(sinfo.Protocol())
+
+	n := time.Now()
+	for i, b := range maglevEPs {
+		mKey := s.newMaglevKey(vip, port, proto, uint32(i))
+		mVal := s.newBackendValue(net.ParseIP(b.IP()), uint16(b.Port()))
+		s.bpfMaglevEps.Desired().Set(mKey, mVal)
+	}
+	dt := time.Since(n)
```
tomastigera: note that this does not write them into dataplane, this is all in memory

felix/bpf/proxy/syncer.go
```diff
@@ -620,7 +657,18 @@ func (s *Syncer) apply(state DPSyncerState) error {
 			}
 		}
 
-		err := s.applySvc(skey, svc, eps)
+		var maglevEPs []k8sp.Endpoint
+		if svc.UseMaglev() {
+			ch := s.newConsistentHash()
+			for _, ep := range eps {
+				if ep.IsReady() {
```
tomastigera, line 664: ohhh you are filtering them here

felix/bpf/proxy/syncer.go
```diff
@@ -659,7 +708,7 @@ func (s *Syncer) apply(state DPSyncerState) error {
 					// separately
 					continue
 				}
-				err := s.applyDerived(sname, svcTypeNodePort, npInfo)
+				err := s.applyDerived(sname, svcTypeNodePort, npInfo, nil)
```
tomastigera, line 711: would it make more sense, if a nodeport is annotated with maglev, just not set the maglev flag on nodeport type when writing the frontend?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,191 @@
...
+	b.endpoint = nil
+	delete(ch.backendsByName, name)
+
+	for i, b := range ch.backendNames {
+		if b == name {
+			slices.Delete(ch.backendNames, i, i)
+		}
+	}
+}
```
tomastigera: do you ever use this?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,191 @@
...
+				break populate
+			}
+		}
+	}
+
+	return lut
+}
+
+// Permutation implements Permutator interface.
+func (c *ConsistentHash) permutation(backendName string) ([]int, error) {
```
tomastigera:
```suggestion
func (ch *ConsistentHash) permutation(s string) ([]int, error) {
```

nit: I am surprised that linter does not complain about you are using sometimes c and sometimes ch

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,191 @@
...
+	// so we're counting manually.
+	n := 0
+populate:
+	for {
+		for i, backend := range ch.backendNames {
+			prefs := ch.backendsByName[backend].permutation
+			choice := prefs[next[i]]
+			for lut[choice] != nil {
+				next[i]++
+				choice = prefs[next[i]]
```
tomastigera, line 102: may you get out of bounds here? can next[i] get larger than len(prefs) ?

felix/bpf/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,191 @@
...
+
+			lut[choice] = ch.backendsByName[backend].endpoint
+
+			// Its *next* preference (after this one). Not sure if necessary to remember this.
+			// Maybe it makes something easier later on if this info is to-hand?
+			// If its safe to discard, we can probably rewrite this whole func to be more Go-ful.
+			next[i]++
+			n++
+			if n == ch.m {
+				break populate
```
tomastigera:
```suggestion
				return lut
```

felix/bpf/consistenthash/consistenthash.go
aaaaaaaalex: I dont think so now, since I never calculate deltas of backends. Just reprogramming the whole lut every time which probably isn't great for performance...

libcalico-go/lib/consistenthash/constants.go
aaaaaaaalex: Stale variable - deleted.
Right now I have the Felixconfig defaulting to 100/100 endpoints/services

felix/bpf/proxy/syncer.go
aaaaaaaalex: Since we incorporate the dst-IP of the packet into the maglev backend keys for that map, we need a LUT for each possible IP the user can hit, or, we need to translate all the possible IP's a user can hit into a canonical IP so that we come up with the same key each time. I'd like to go back and do the latter later.

felix/bpf/proxy/syncer.go
aaaaaaaalex: Since I need to pass them to  `applyDerived` too, I'm generating them at the top level
(the reason I need them for `applyDerived` is above!)

felix/bpf/proxy/syncer.go
aaaaaaaalex: @ctauchen FYI - let's chat about this

felix/bpf/proxy/syncer.go
aaaaaaaalex: thx - originally I was generating the LUT between those timings, and wanted to see how long that took, but was only a debug/test anyways

felix/bpf/proxy/syncer.go
aaaaaaaalex: It didn't feel convenient to write it that way, and I was a little paranoid there may be edge cases.
Maybe we could come back and clean up the rough edges on a followup PR?

felix/bpf/consistenthash/consistenthash.go
aaaaaaaalex: As long as the `prefs` list is truly a scrambled version of the list 0...M with no duplicate values, then it should never go out of bounds.

felix/bpf/proxy/syncer.go
tomastigera: ohhh I see :+1:

felix/bpf/consistenthash/consistenthash.go
tomastigera: `next[i]++` may technically get over. Isn't that some sort of a corner case a break-out-of-the-loop condition?

libcalico-go/lib/consistenthash/constants.go
```diff
@@ -0,0 +1,5 @@
+package consistenthash
+
+const (
+	MaglevEndpointLUTFactor = 5
```
tomastigera, line 4: is that constant used anywhere?

libcalico-go/lib/consistenthash/constants.go
```diff
@@ -0,0 +1,5 @@
+package consistenthash
```
tomastigera, line 1: why do we have this package and why in libcalico-go?

libcalico-go/lib/consistenthash/constants.go
aaaaaaaalex: Yep this is multiplied by the number of endpoints as part of finding a large prime number

libcalico-go/lib/consistenthash/constants.go
aaaaaaaalex: This and the primes code was placed outside of Felix because it's depended on by api pkg, and importing low-level felix pkgs (the CH module etc) felt messy

felix/bpf/consistenthash/consistenthash.go
aaaaaaaalex: It would imply we need to pick an index for a backend after all LUT spots are already taken. This would mean a pretty major bug and I think it'd be best if we panicked to be honest :thinking:

felix/fv/bpf_test.go
```diff
@@ -1912,6 +1914,253 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+							},
+						}
+
+						allowIngressFromExtClientSelector := "all()"
+						allowIngressFromExtClient.Spec.Selector = allowIngressFromExtClientSelector
+						allowIngressFromExtClient = createPolicy(allowIngressFromExtClient)
+
+						// cmd = append(ipRoute, "route", "add", "local", extIP, "dev", "eth0")
+						// tc.Felixes[1].Exec(cmd...)
+						// tc.Felixes[2].Exec(cmd...)
```
tomastigera: debug leftover?

felix/fv/bpf_test.go
```diff
@@ -1912,6 +1914,253 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						// Traffic is flowing over LB 1. Change ExtClient's serviceIP route to go via LB 2.
+						ipRoute := []string{"ip"}
+						if testOpts.ipv6 {
+							ipRoute = append(ipRoute, "-6")
+						}
+						ipRouteReplace := append(ipRoute, "route", "replace", serviceIP, "via", felixIP(2))
+						externalClient.Exec(ipRouteReplace...)
+
+						lastPongCount := pc.PongCount()
+						Eventually(pc.PongCount, "5s", "100ms").Should(BeNumerically(">", lastPongCount), "Connection is no longer ponging after route failover")
```
tomastigera: I think this should get after you check that the connection failed over to make a proof that the connection continues.

Also perhaps there should be a tcpdump probe that makes sure that the connection is not going through the original node. Perhaps watching vxlan packets and making sure their source changes. The conntrack test is a good indication.

felix/fv/bpf_test.go
aaaaaaaalex: Yes, thx.
There is enhanced route programming happening a few lines down

felix/fv/bpf_test.go
aaaaaaaalex: The "failover" for this test is to replace the ext-node's route to the backend which is above this line (`ipRouteReplace...`)

IMO the next line down confirms the failover happened - a conntrack shows up on another node. WDYT?

tomastigera:
unrelated bpf fv failures

Issue #11175, opened 2025-10-11
IPAMBlocks Resource Leak After Upgrading  v3.30.0


gojoy:
We found that after upgrading the cluster's Calico to version v3.30.0, many IPAMBlock resources were not properly cleaned up. Through investigation, we identified that the issue is related to the newly added .spec.type field in BlockAffinities.
## Expected Behavior

1. delete Node from k8s cluster
2. delete the corresponding IPPools (using v3 API)
3.  calico-apiserver receives the IPPool deletion request, it calls ReleasePoolAffinities to check `affinityMatches` and clean up IPAMBlock resources under the deleted IPPool.
## Current Behavior

1. delete Node from k8s cluster
2. delete the corresponding IPPools (using v3 API)
3. calico-apiserver receives the IPPool deletion request, it calls ReleasePoolAffinities to check `affinityMatches` ,Unfortunately, the ``affinityMatches`` check failed, preventing the subsequent IPAMBlock cleanup workflow from being executed.
## Possible Solution

Since the .spec.type field in BlockAffinities resources was introduced in version 3.30, previously created BlockAffinities resources had empty .spec.type fields, which led to the validation check failure and subsequent cleanup steps could not be executed.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Calico version v3.30.0
* Calico dataplane  iptables
* Orchestrator version kubernetes
* Operating System and version: Anolis OS 8.9
* Link to your project (optional):

caseydavenport:
@gojoy thanks for raising this and good catch! I'm working on a fix now.

caseydavenport:
I think this will do it: https://github.com/projectcalico/calico/pull/11179

PR #11179, opened 2025-10-13
Fix block cleanup for blocks with no affinity type

caseydavenport:
## Description

Fixes https://github.com/projectcalico/calico/issues/11175

We need to default the AffinityType field on read, as older blocks don't have this field set 
and the newer IPAM code expects it to be. 

etcd doesn't need a corresponding fix, as the affinity type was already part of the key even 
before the AffnityType field was added to our API.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix IPAM block leak of older blocks when deleting IP pools. 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
+++ b/libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
@@ -35,7 +35,6 @@ import (
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/model"
 	cerrors "github.com/projectcalico/calico/libcalico-go/lib/errors"
-	"github.com/projectcalico/calico/libcalico-go/lib/ipam"
 	"github.com/projectcalico/calico/libcalico-go/lib/names"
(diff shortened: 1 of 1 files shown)


libcalico-go/lib/backend/k8s/resources/ipam_affinity.go
```diff
@@ -94,10 +94,17 @@ func (c *blockAffinityClient) toV1(kvpv3 *model.KVPair) (*model.KVPair, error) {
 		}
 	}
 
+	// Default affinity type to "host" if not set. Older versions of Calico's CRD backend
+	// did not set this field, assuming "host" as the default.
+	affinityType := "host"
```
MichalFupso: Could we use the AffinityTypeHost here? https://github.com/projectcalico/calico/blob/f3c4ecbe18a6ac3349488e21f13253ef40e8cc8e/libcalico-go/lib/ipam/ipam_types.go#L203

caseydavenport: The answer is "not really, because of an import cycle"

PR #11180, opened 2025-10-13
Start removing old Python STs

caseydavenport:
## Description

The impetus here is that SemaphoreCI is upgrading to a new kernel that doesn't support cgroupv1, required for the version of docker in docker used in these tests to function.

We've wanted to clean these up for a while anyway, so this is good motivation. Some of them are already redundant with other UTs/FVs - these can just be deleted.

Some others should be moved to other test suites (e.g., Felix FVs or e2e)

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/e2e/pkg/tests/bgp/export.go
+++ b/e2e/pkg/tests/bgp/export.go
@@ -91,8 +91,9 @@ var _ = describe.CalicoDescribe(
 			// - Create a server pod and corresponding service in the main namespace for the test.
 			// - Create a client pod and assert that it can connect to the service.
 			By(fmt.Sprintf("Creating server pod in namespace %s", f.Namespace.Name))
-			// Use cuztomizers to ensure pods use the test IP pool and avoid landing on the same node.
-			cusomtizer := conncheck.CombineCustomizers(
+
+			// Use customizers to ensure pods use the test IP pool and avoid landing on the same node.
+			customtizer := conncheck.CombineCustomizers(
 				conncheck.UseV4IPPool(pool.Name),
 				conncheck.AvoidEachOther,
 			)
@@ -100,12 +101,12 @@ var _ = describe.CalicoDescribe(
 				"server",
 				f.Namespace,
 				conncheck.WithServerLabels(map[string]string{"role": "server"}),
-				conncheck.WithServerPodCustomizer(cusomtizer),
+				conncheck.WithServerPodCustomizer(customtizer),
 			)
 			client1 = conncheck.NewClient(
 				"client",
 				f.Namespace,
-				conncheck.WithClientCustomizer(cusomtizer),
+				conncheck.WithClientCustomizer(customtizer),
 			)
 			checker.AddServer(server1)
 			checker.AddClient(client1)
--- a/e2e/pkg/tests/bgp/route_reflector.go
+++ b/e2e/pkg/tests/bgp/route_reflector.go
@@ -21,6 +21,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+	"github.com/projectcalico/api/pkg/lib/numorstring"
 	v1 "github.com/tigera/operator/api/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -118,10 +119,13 @@ var _ = describe.CalicoDescribe(
 			// Disable full mesh BGP.
 			disableFullMesh(cli)
 
-			// Set the AS number for the cluster to a non-default value.
-			setASNumber(cli, 64514)
+			// Set the AS number for the cluster to a non-default value. We do this to get
+			// coverage of changing the AS number, but it is not strictly required for this test.
+			asn, err := numorstring.ASNumberFromString("4294.566")
+			Expect(err).NotTo(HaveOccurred(), "Error converting AS number from string")
+			setASNumber(cli, asn)
 
-			// Verify connectivity is lost.
+			// Verify connectivity is lost because the full mesh is disabled.
 			checker.ResetExpectations()
 			checker.ExpectFailure(client1, server1.ClusterIPs()...)
 			checker.ExpectFailure(client2, server1.ClusterIPs()...)
@@ -162,7 +166,8 @@ var _ = describe.CalicoDescribe(
 			// Disable full mesh BGP.
 			disableFullMesh(cli)
 
-			// Set the AS number for the cluster to a non-default value.
+			// Set the AS number for the cluster to a non-default value. We do this to get
+			// coverage of changing the AS number, but it is not strictly required for this test.
 			setASNumber(cli, 64514)
 
 			// Verify connectivity is lost.
@@ -201,11 +206,21 @@ var _ = describe.CalicoDescribe(
 			checker.ExpectSuccess(client2, server1.ClusterIPs()...)
 			checker.Execute()
 
+			// By bringing down one of the nodes acting as a route reflector, we can verify that
+			// route reflection is functioning correctly.
+			deleteCalicoNode(cli, &nodes.Items[1])
+
+			// Verify connectivity is maintained.
+			checker.ResetExpectations()
+			checker.ExpectSuccess(client1, server1.ClusterIPs()...)
+			checker.ExpectSuccess(client2, server1.ClusterIPs()...)
+			checker.Execute()
+
 			// Now, remove one of the RRs and verify connectivity is maintained.
 			By("Removing one of the route reflectors")
 			setNodeAsNotRouteReflector(cli, &nodes.Items[1])
 
-			// Verify c nnectivity is maintained.
+			// Verify connectivity is maintained.
 			checker.ResetExpectations()
 			checker.ExpectSuccess(client1, server1.ClusterIPs()...)
 			checker.ExpectSuccess(client2, server1.ClusterIPs()...)
--- a/e2e/pkg/tests/bgp/utils.go
+++ b/e2e/pkg/tests/bgp/utils.go
@@ -18,7 +18,7 @@ import (
 )
 
 // ensureInitialBGPConfig checks for an existing BGPConfiguration resource and ensures that full mesh BGP is enabled.
-// It returns the initial configuration (if any) and a cleanup function to restore the original state after the test.
+// It returns a cleanup function to restore the original state after the test.
 func ensureInitialBGPConfig(cli ctrlclient.Client) func() {
 	// Ensure full mesh BGP is functioning before each test.
 	initialConfig := &v3.BGPConfiguration{}
@@ -40,12 +40,12 @@ func ensureInitialBGPConfig(cli ctrlclient.Client) func() {
 			err := cli.Delete(context.Background(), &v3.BGPConfiguration{ObjectMeta: metav1.ObjectMeta{Name: "default"}})
 			ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error deleting BGPConfiguration resource")
 		}
-	} else {
-		By("Ensuring full mesh BGP is enabled in existing BGPConfiguration")
-		ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error querying BGPConfiguration resource")
-		ExpectWithOffset(1, initialConfig.Spec.NodeToNodeMeshEnabled).NotTo(BeNil(), "nodeToNodeMeshEnabled is not configured in BGPConfiguration")
-		ExpectWithOffset(1, *initialConfig.Spec.NodeToNodeMeshEnabled).To(BeTrue(), "nodeToNodeMeshEnabled is not enabled in BGPConfiguration")
 	}
+
+	By("Ensuring full mesh BGP is enabled in existing BGPConfiguration")
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error querying BGPConfiguration resource")
+	ExpectWithOffset(1, initialConfig.Spec.NodeToNodeMeshEnabled).NotTo(BeNil(), "nodeToNodeMeshEnabled is not configured in BGPConfiguration")
+	ExpectWithOffset(1, *initialConfig.Spec.NodeToNodeMeshEnabled).To(BeTrue(), "nodeToNodeMeshEnabled is not enabled in BGPConfiguration")
 
 	return func() {
 		By("Restoring initial BGPConfiguration")
@@ -69,7 +69,7 @@ func disableFullMesh(cli ctrlclient.Client) {
 	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error updating BGPConfiguration resource")
 }
 
-func setASNumber(cli ctrlclient.Client, asn int32) {
+func setASNumber(cli ctrlclient.Client, asn numorstring.ASNumber) {
 	By("Setting AS number in BGPConfiguration")
 	config := &v3.BGPConfiguration{}
 	err := cli.Get(context.Background(), ctrlclient.ObjectKey{Name: "default"}, config)
@@ -82,18 +82,24 @@ func setASNumber(cli ctrlclient.Client, asn int32) {
 func setNodeAsRouteReflector(cli ctrlclient.Client, rrNode *corev1.Node, rrClusterID string) {
 	By(fmt.Sprintf("Using node %s as a route reflector", rrNode.Name))
 	prePatch := ctrlclient.MergeFrom(rrNode.DeepCopy())
+
+	// Adding the cluster ID as an annotation tells Calico to configure the node as a route reflector.
+	if rrNode.Annotations == nil {
+		rrNode.Annotations = map[string]string{}
+	}
+
+	// Adding it as a label allows us to select the node in a BGPPeer.
+	if rrNode.Labels == nil {
+		rrNode.Labels = map[string]string{}
+	}
+
 	rrNode.Labels[resources.RouteReflectorClusterIDAnnotation] = rrClusterID
 	rrNode.Annotations[resources.RouteReflectorClusterIDAnnotation] = rrClusterID
 	err := cli.Patch(context.Background(), rrNode, prePatch)
 	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error marking node as route reflector")
 
 	DeferCleanup(func() {
-		// Remove the label and annotation from the node.
-		prePatch := ctrlclient.MergeFrom(rrNode.DeepCopy())
-		delete(rrNode.Labels, resources.RouteReflectorClusterIDAnnotation)
-		delete(rrNode.Annotations, resources.RouteReflectorClusterIDAnnotation)
-		err = cli.Patch(context.Background(), rrNode, prePatch)
-		ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error removing route reflector label from node during cleanup")
+		setNodeAsNotRouteReflector(cli, rrNode)
 	})
 }
 
@@ -104,4 +110,18 @@ func setNodeAsNotRouteReflector(cli ctrlclient.Client, rrNode *corev1.Node) {
 	delete(rrNode.Annotations, resources.RouteReflectorClusterIDAnnotation)
 	err := cli.Patch(context.Background(), rrNode, prePatch)
 	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error removing route reflector label from node")
+}
+
+func deleteCalicoNode(cli ctrlclient.Client, node *corev1.Node) {
+	By(fmt.Sprintf("Simulating failure of node %s by deleting calico-node Pod", node.Name))
+
+	// Get the calico/node Pod on the node.
+	podList := &corev1.PodList{}
+	err := cli.List(context.Background(), podList, ctrlclient.MatchingLabels{"k8s-app": "calico-node"}, ctrlclient.MatchingFields{"spec.nodeName": node.Name})
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error listing calico-node Pods on node")
+	ExpectWithOffset(1, len(podList.Items)).To(BeNumerically(">", 0), "No calico-node Pod found on node")
+
+	// Delete the calico/node Pod to simulate failure.
+	err = cli.Delete(context.Background(), &podList.Items[0])
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error deleting calico-node Pod to simulate node failure")
 }
--- a/e2e/pkg/tests/networking/ipip.go
+++ b/e2e/pkg/tests/networking/ipip.go
@@ -100,8 +100,9 @@ var _ = describe.CalicoDescribe(
 			// - Create a server pod and corresponding service in the main namespace for the test.
 			// - Create a client pod and assert that it can connect to the service.
 			By(fmt.Sprintf("Creating server pod in namespace %s", f.Namespace.Name))
-			// Use cuztomizers to ensure pods use the test IP pool and avoid landing on the same node.
-			cusomtizer := conncheck.CombineCustomizers(
+
+			// Use customizers to ensure pods use the test IP pool and avoid landing on the same node.
+			customizer := conncheck.CombineCustomizers(
 				conncheck.UseV4IPPool(pool.Name),
 				conncheck.AvoidEachOther,
 			)
@@ -109,12 +110,12 @@ var _ = describe.CalicoDescribe(
 				"server",
 				f.Namespace,
 				conncheck.WithServerLabels(map[string]string{"role": "server"}),
-				conncheck.WithServerPodCustomizer(cusomtizer),
+				conncheck.WithServerPodCustomizer(customizer),
 			)
 			client1 = conncheck.NewClient(
 				"client",
 				f.Namespace,
-				conncheck.WithClientCustomizer(cusomtizer),
+				conncheck.WithClientCustomizer(customizer),
 			)
 			checker.AddServer(server1)
 			checker.AddClient(client1)
@@ -222,7 +223,7 @@ var _ = describe.CalicoDescribe(
 
 				out, err := conncheck.ExecInPod(&p, "sh", "-c", "ip addr show tunl0")
 				Expect(err).NotTo(HaveOccurred(), "Error querying tunl0 interface in pod %s", p.Name)
-				Expect(out).To(ContainSubstring(expectedIP), "tunl0 interface in pod %s does not have an IPIP address", p.Name)
+				Expect(out).To(ContainSubstring(expectedIP), "tunl0 interface in pod %s does not have the expected IPIP address", p.Name)
 			}
 		})
 	})
--- a/node/tests/k8st/tests/test_health.py
+++ b/node/tests/k8st/tests/test_health.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2018-2019 Tigera, Inc. All rights reserved.
+# Copyright (c) 2018-2025 Tigera, Inc. All rights reserved.
 #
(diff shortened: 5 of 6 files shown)


Makefile
```diff
@@ -164,15 +164,15 @@ image:
 # Run local e2e smoke test against the checked-out code
 # using a local kind cluster.
 ###############################################################################
-E2E_FOCUS ?= "sig-network.*Conformance|sig-calico.*Conformance"
+E2E_FOCUS ?= "sig-network.*Conformance|sig-calico.*Conformance|BGP"
```
stevegaossou, line 167: Any reason to have the new BGP tests be in the default focus? Were the old python tests run as part of conformance?

node/tests/k8st/tests/test_bgp_advert.py
```diff
@@ -280,45 +280,6 @@ def test_cluster_ip_advertisement(self):
...
-            # # Connectivity to nginx-cluster will rarely succeed because it is load-balanced across all nodes.
-            # # When the traffic hits a node that doesn't host one of the service's pod, it will be re-routed
-            # #  to another node and SNAT will cause the policy to drop the traffic.
-            # # Try to curl 10 times.
-            # try:
-            #   for i in range(attempts):
-            #     curl(cluster_svc_ip)
-            #   self.fail("external node should not be able to consistently access the cluster svc")
-            # except subprocess.CalledProcessError:
-            #   pass
```
stevegaossou, line 321: Yay to removing commented out lines. 🎉

stevegaossou:
> Copilot reviewed 21 out of 23 changed files in this pull request and generated no comments.

Hmm weird, that's not even the whole set of file changes. I guess Copilot is capped at a certain number of lines of changes?

stevegaossou:
> Some others should be moved to other test suites (e.g., Felix FVs or e2e)

@caseydavenport could you confirm that you basically removed all of the problematic tests in this one PR (i.e. the DinD dependent tests). I'm assuming that is true judging by the volume of Python test code removed. 

And you have replaced some of them here. But there are a few remaining tests that got removed here that will need to be replaced with a Go-equivalent test in the near future. 

Is that correct?

caseydavenport:
@stevegaossou this PR doesn't get rid of every single DinD test - there are still a few more that I was going to do in a follow-on. 

The tests that I have removed fall into one of a few buckets (in order of prevalence in this PR):
- Replaced by a Go test in this PR.
- Already covered by another UT or e2e, so we don't need to replace.
- Replaced by a Python k8st in this PR.
- Low value test that doesn't need immediate replacement (but I have tracked in my doc)

libcalico-go/lib/backend/k8s/resources/node.go
```diff
@@ -49,13 +49,14 @@ const (
...
 	nodeBgpAsnAnnotation                  = "projectcalico.org/ASNumber"
-	nodeBgpCIDAnnotation                  = "projectcalico.org/RouteReflectorClusterID"
 	nodeK8sLabelAnnotation                = "projectcalico.org/kube-labels"
 	nodeWireguardIpv4IfaceAddrAnnotation  = "projectcalico.org/IPv4WireguardInterfaceAddr"
 	nodeWireguardIpv6IfaceAddrAnnotation  = "projectcalico.org/IPv6WireguardInterfaceAddr"
 	nodeWireguardPublicKeyAnnotation      = "projectcalico.org/WireguardPublicKey"
 	nodeWireguardPublicKeyV6Annotation    = "projectcalico.org/WireguardPublicKeyV6"
 	nodeInterfacesAnnotation              = "projectcalico.org/Interfaces"
+
+	RouteReflectorClusterIDAnnotation = "projectcalico.org/RouteReflectorClusterID"
```
stevegaossou, line 59: Does this constant need to be exported? I noticed it used to be private to the package, and it seems like (at least for now) it still is used only within the same package (`libcalico-go/lib/backend/k8s/resources/`).

node/tests/k8st/tests/test_health.py
```diff
@@ -0,0 +1,99 @@
...
+from subprocess import CalledProcessError
+
+from tests.k8st.test_base import TestBase
+from tests.k8st.utils.utils import DiagsCollector, calicoctl, kubectl, exec_in_calico_node, run, retry_until_success, node_info, start_external_node_with_bgp, update_ds_env
+
+_log = logging.getLogger(__name__)
+_log.setLevel(logging.DEBUG)
+
+
+class TestReadiness(TestBase):
```
stevegaossou, line 25: @caseydavenport is the expectation that this new test logic written in Python will be able to live with the remaining Python tests for as long as we want to keep them around? 

Meaning, there isn't any reason in the near future to move these over to Golang? 

Was the reason to leave them in Python form simply a matter of being pragmatic? I.e. it would have been more overhead to convert them over because of surrounding utility functions etc (or that it makes more sense to be grouped with other existing Python tests).

Makefile
caseydavenport: This isn't really meant to define what Conformance is, it's just meant to define what tests we run as part of CI.

I should probably rename the Semaphore Job.

libcalico-go/lib/backend/k8s/resources/node.go
caseydavenport: Yeah, I use it in some of the new BGP tests.

node/tests/k8st/tests/test_health.py
caseydavenport: I moved it here since I don't think these are appropriate for e2e tests because they rely on specific implementation details of calico/node, and we don't have a suitable Golang FV test suite that can do these types of things yet.

e2e/pkg/tests/bgp/utils.go
```diff
@@ -0,0 +1,107 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/projectcalico/calico/libcalico-go/lib/backend/k8s/resources"
+)
+
+// ensureInitialBGPConfig checks for an existing BGPConfiguration resource and ensures that full mesh BGP is enabled.
+// It returns the initial configuration (if any) and a cleanup function to restore the original state after the test.
```
nelljerram: Function only has one return value.  Doesn't appear to return the initial config.

e2e/pkg/tests/bgp/utils.go
```diff
@@ -0,0 +1,107 @@
...
+			},
+		})
+		ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error creating BGPConfiguration resource")
+
+		return func() {
+			By("Deleting BGPConfiguration created for tests")
+			err := cli.Delete(context.Background(), &v3.BGPConfiguration{ObjectMeta: metav1.ObjectMeta{Name: "default"}})
+			ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error deleting BGPConfiguration resource")
+		}
+	} else {
```
nelljerram: I think it would look more natural to promote this code out of the else block.  (And then remove the else.)

e2e/pkg/tests/bgp/utils.go
```diff
@@ -0,0 +1,107 @@
...
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error querying BGPConfiguration resource")
+	config.Spec.ASNumber = ptr.To(numorstring.ASNumber(asn))
+	err = cli.Update(context.Background(), config)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error updating BGPConfiguration resource")
+}
+
+func setNodeAsRouteReflector(cli ctrlclient.Client, rrNode *corev1.Node, rrClusterID string) {
+	By(fmt.Sprintf("Using node %s as a route reflector", rrNode.Name))
+	prePatch := ctrlclient.MergeFrom(rrNode.DeepCopy())
+	rrNode.Labels[resources.RouteReflectorClusterIDAnnotation] = rrClusterID
```
nelljerram, line 96: Do we need to set this as a label, as well as as an annotation?

e2e/pkg/tests/bgp/utils.go
```diff
@@ -0,0 +1,107 @@
...
+	err := cli.Patch(context.Background(), rrNode, prePatch)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error marking node as route reflector")
+
+	DeferCleanup(func() {
+		// Remove the label and annotation from the node.
+		prePatch := ctrlclient.MergeFrom(rrNode.DeepCopy())
+		delete(rrNode.Labels, resources.RouteReflectorClusterIDAnnotation)
+		delete(rrNode.Annotations, resources.RouteReflectorClusterIDAnnotation)
+		err = cli.Patch(context.Background(), rrNode, prePatch)
+		ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Error removing route reflector label from node during cleanup")
```
nelljerram: Could use setNodeAsNotRouteReflector here instead of duplicating.

e2e/pkg/tests/bgp/export.go
```diff
@@ -0,0 +1,157 @@
...
+			DeferCleanup(func() {
+				err = cli.Delete(context.Background(), pool)
+				Expect(err).NotTo(HaveOccurred(), "Error deleting IP pool")
+			})
+
+			// Before each test, perform the following steps:
+			// - Create a server pod and corresponding service in the main namespace for the test.
+			// - Create a client pod and assert that it can connect to the service.
+			By(fmt.Sprintf("Creating server pod in namespace %s", f.Namespace.Name))
+			// Use cuztomizers to ensure pods use the test IP pool and avoid landing on the same node.
```
nelljerram: sp: customisers

e2e/pkg/tests/bgp/export.go
```diff
@@ -0,0 +1,157 @@
...
+			cusomtizer := conncheck.CombineCustomizers(
```
nelljerram: sp

e2e/pkg/tests/bgp/route_reflector.go
```diff
@@ -0,0 +1,224 @@
...
+		})
+
+		It("should support in-cluster route reflectors", func() {
+			// Disable full mesh BGP.
+			disableFullMesh(cli)
+
+			// Set the AS number for the cluster to a non-default value.
+			setASNumber(cli, 64514)
+
+			// Verify connectivity is lost.
```
nelljerram: Presumably this is because of the mesh being disabled, and not because of the AS number?  Worth a comment to say that.

e2e/pkg/tests/bgp/route_reflector.go
```diff
@@ -0,0 +1,224 @@
...
+			Expect(nodes.Items).NotTo(BeEmpty(), "No nodes found in the cluster")
+			setNodeAsRouteReflector(cli, &nodes.Items[0], "225.0.0.4")
+
+			// Create BGP peer that causes all non-RR nodes to peer with the RR.
+			By("Creating a BGPPeer to re-enable peers via the RR")
+			peer := &v3.BGPPeer{
+				ObjectMeta: metav1.ObjectMeta{Name: "peer-to-rrs"},
+				Spec: v3.BGPPeerSpec{
+					NodeSelector: fmt.Sprintf("!has(%s)", resources.RouteReflectorClusterIDAnnotation),
+					PeerSelector: fmt.Sprintf("has(%s)", resources.RouteReflectorClusterIDAnnotation),
```
nelljerram, line 148: OK, I guess that's why we need an RR label.

e2e/pkg/tests/bgp/route_reflector.go
```diff
@@ -0,0 +1,224 @@
...
+			checker.ExpectSuccess(client1, server1.ClusterIPs()...)
+			checker.ExpectSuccess(client2, server1.ClusterIPs()...)
+			checker.Execute()
+		})
+
+		It("should support clustered route reflectors", func() {
+			// Disable full mesh BGP.
+			disableFullMesh(cli)
+
+			// Set the AS number for the cluster to a non-default value.
```
nelljerram: Can you comment why this is needed?

e2e/pkg/tests/bgp/route_reflector.go
```diff
@@ -0,0 +1,224 @@
...
+			checker.ResetExpectations()
+			checker.ExpectSuccess(client1, server1.ClusterIPs()...)
+			checker.ExpectSuccess(client2, server1.ClusterIPs()...)
+			checker.Execute()
+
+			// Now, remove one of the RRs and verify connectivity is maintained.
+			By("Removing one of the route reflectors")
+			setNodeAsNotRouteReflector(cli, &nodes.Items[1])
+
+			// Verify c nnectivity is maintained.
```
nelljerram: sp

e2e/pkg/tests/networking/ipip.go
```diff
@@ -0,0 +1,250 @@
...
+			DeferCleanup(func() {
+				err = cli.Delete(context.Background(), pool)
+				Expect(err).NotTo(HaveOccurred(), "Error deleting IP pool")
+			})
+
+			// Before each test, perform the following steps:
+			// - Create a server pod and corresponding service in the main namespace for the test.
+			// - Create a client pod and assert that it can connect to the service.
+			By(fmt.Sprintf("Creating server pod in namespace %s", f.Namespace.Name))
+			// Use cuztomizers to ensure pods use the test IP pool and avoid landing on the same node.
```
nelljerram:
```suggestion
			// Use customizers to ensure pods use the test IP pool and avoid landing on the same node.
```

e2e/pkg/tests/networking/ipip.go
```diff
@@ -0,0 +1,250 @@
...
+			cusomtizer := conncheck.CombineCustomizers(
```
nelljerram:
```suggestion
			customizer := conncheck.CombineCustomizers(
```

nelljerram: Please check and fix the same typos throughout the current PR.

e2e/pkg/tests/networking/ipip.go
```diff
@@ -0,0 +1,250 @@
...
+				// Query the node hosting this pod to get its IP address.
+				node := corev1.Node{}
+				err = cli.Get(context.Background(), ctrlclient.ObjectKey{Name: p.Spec.NodeName}, &node)
+				Expect(err).NotTo(HaveOccurred(), "Error querying node %s", p.Spec.NodeName)
+				expectedIP, ok := node.Annotations["projectcalico.org/IPv4IPIPTunnelAddr"]
+				Expect(ok).To(BeTrue(), "Node %s does not have an IPv4Address annotation", node.Name)
+
+				out, err := conncheck.ExecInPod(&p, "sh", "-c", "ip addr show tunl0")
+				Expect(err).NotTo(HaveOccurred(), "Error querying tunl0 interface in pod %s", p.Name)
+				Expect(out).To(ContainSubstring(expectedIP), "tunl0 interface in pod %s does not have an IPIP address", p.Name)
```
nelljerram: "does not have expected IPIP address"

node/tests/k8st/tests/test_health.py
```diff
@@ -0,0 +1,99 @@
+# Copyright (c) 2018-2019 Tigera, Inc. All rights reserved.
```
nelljerram: 2025 ?

node/tests/st/bgp/test_global_peers.py
```diff
@@ -1,80 +0,0 @@
...
-            # Allow network to converge
-            self.assert_true(workload_host1.check_can_ping(DEFAULT_IPV4_ADDR_2, retries=10))
-
-            # Turn the node-to-node mesh off and wait for connectivity to drop.
-            update_bgp_config(host1, nodeMesh=False)
-            self.assert_true(workload_host1.check_cant_ping(DEFAULT_IPV4_ADDR_2, retries=10))
-
-            # Configure global peers to explicitly set up a mesh.
-            create_bgp_peer(host1, 'global', host2.ip, LARGE_AS_NUM)
-            create_bgp_peer(host2, 'global', host1.ip, LARGE_AS_NUM)
```
nelljerram, line 60: Do your replacement tests cover `LARGE_AS_NUM` - which I assume means 32 bit?

node/tests/st/bgp/test_ipip.py
```diff
@@ -1,626 +0,0 @@
...
-            # we remove the original IP, and allocate a new one from the new pool
-            new_ipv4_pool = IPNetwork("192.168.0.0/16")
-            self.pool_action(host, "create", new_ipv4_pool, ipip_mode="Always", pool_name="pool-b")
-            self.pool_action(host, "delete", ipv4_pool)
-            self.assert_tunl_ip(host, new_ipv4_pool)
-
-    @parameterized.expand([
-        ('bird',),
-    ])
-    def test_issue_1584(self, backend):
```
nelljerram, line 196: I don't think you've created a replacement for this one.  Do we need it?

node/tests/st/bgp/test_ipip.py
```diff
@@ -1,626 +0,0 @@
...
-
-        match = re.search(r'RX packets:(\d+) ',
-                          output)
-        return int(match.group(1))
-
-    @parameterized.expand([
-        (False,),
-        (True,),
-    ])
-    def test_gce(self, with_ipip, backend='bird'):
```
nelljerram, line 466: WDYT about this one?  I don't think it's mentioned in your analysis doc.

node/tests/st/bgp/test_ipip.py
```diff
@@ -1,626 +0,0 @@
...
-                        # during the following monitoring, as it might be a
-                        # little slower than what is needed for the workload
-                        # connectivity tested above.
-                        expected_route_string = "/26 via " + rr_ip
-                        def check():
-                            self.assertIn(expected_route_string,
-                                          host.execute("ip r"))
-                        retry_until_success(check, retries=10)
-
-                    # Check that routes are not flapping, on either host, by
```
nelljerram, line 576: This is something I remember as having been important once - i.e. we had a bug where the symptom was a persistent flapping with tunnel and non-tunnel routes.

node/tests/st/bgp/test_node_peers.py
```diff
@@ -1,80 +0,0 @@
...
-                                                       workload_host2],
-                                        ip_pass_list=[DEFAULT_IPV4_ADDR_1,
-                                                      DEFAULT_IPV4_ADDR_2])
-
-            # Check the BGP status on each host.
-            check_bird_status(host1, [("node specific", host2.ip, "Established")])
-            check_bird_status(host2, [("node specific", host1.ip, "Established")])
-
-    @attr('slow')
-    def test_bird_node_peers(self):
```
nelljerram, line 77: Still to migrate?  Your doc indicates that it should move to e2e, but I don't think you've coded this yet.

node/tests/st/bgp/test_route_reflector_cluster.py
```diff
@@ -1,132 +0,0 @@
...
-from tests.st.utils.route_reflector import RouteReflectorCluster
-
-from .peer import create_bgp_peer
-from tests.st.utils.utils import update_bgp_config
-
-class TestRouteReflectorCluster(TestBase):
-
-    def test_route_reflector_cluster_resilience(self):
-        """
-        Runs a cluster of route reflectors, brings one node down, and ensures that traffic still flows
```
nelljerram, line 28: Still to do?  Don't think your e2e test brings one node down.

node/tests/st/bgp/test_route_reflector_cluster.py
```diff
@@ -1,132 +0,0 @@
...
-            # Allow network to converge (which it won't)
-            self.assert_false(workload_host1.check_can_ping(workload_host2.ip, retries=5))
-            self.assert_true(workload_host1.check_cant_ping(workload_host3.ip))
-            self.assert_true(workload_host2.check_cant_ping(workload_host3.ip))
-
-            # Set distributed peerings between the hosts, each host peering
-            # with a different set of redundant route reflectors.
-            for host in [host1, host2, host3]:
-                for rr in rrc.get_redundancy_group():
-                    create_bgp_peer(host, "node", rr.ip, 64513, metadata={'name': host.name + rr.name.lower()})
```
nelljerram, line 113: Do we have a replacement for this?  IIRC the e2e test does not use multiple RR cluster IDs.

node/tests/st/bgp/test_switch_route_reflector.py
```diff
@@ -1,145 +0,0 @@
...
-
-logger = logging.getLogger(__name__)
-
-class TestSwitchRouteReflector(TestBase):
-
-    @attr('slow')
-    def _test_switch_route_reflector(self, backend='bird', bgpconfig_as_num=64514, peer_as_num=64514):
-        """
-        Test that switching from node-to-node full mesh to route reflectors doesn't disrupt dataplane traffic if done as per
-        https://projectcalico.docs.tigera.io/networking/bgp#change-from-node-to-node-mesh-to-route-reflectors-without-any-traffic-disruption
```
nelljerram, line 40: Need replacement for this one?  Seems to require more than the current e2e connchecker can do.

node/tests/st/bgp/test_update_ip_addr.py
```diff
@@ -1,68 +0,0 @@
...
-from tests.st.test_base import TestBase
-from tests.st.utils.docker_host import DockerHost, CLUSTER_STORE_DOCKER_OPTIONS
-
-class TestUpdateIPAddress(TestBase):
-
-    @attr('slow')
-    def test_update_ip_address(self):
-        """
-        Test updating the IP address automatically updates and fixes the
-        Bird BGP config.
```
nelljerram, line 26: Replacement for this one?

nelljerram (approved):
Have approved so as not to block, but...
1. Have noted a few nits and concerns - happy to leave those to your judgement.
2. I feel like a lot of testing is being deleted that is still useful and not yet replaced in any other form - which is fine for this PR, so long as we have good tracking for recreating that in further work.  Please could you consider all the cases that I've pointed out and create detailed tracking for recreating those, if it doesn't already exist?  (I think we need something a bit more precise than the existing analysis doc; perhaps JIRA tickets.)

e2e/pkg/tests/bgp/route_reflector.go
caseydavenport: Yeah, added a comment to the code - but this was actually just taken from the previous test (I believe it to just be for extra coverage of non-default AS numbers, and not strictly required for the test)

e2e/pkg/tests/bgp/route_reflector.go
caseydavenport: that's a strange one

node/tests/st/bgp/test_global_peers.py
caseydavenport: Nope, good catch. I'll update the alternative ASN in one of the tests.

node/tests/st/bgp/test_ipip.py
caseydavenport: Yeah, I haven't recreated this one entirely, although I did cover some of it with a test that swaps IPIP mode and checks for address assignment and routing updates. 

I think the main thing that is missing from this test was the "cold start" aspect, which I felt wasn't really necessary to test explicitly in an e2e any more.

node/tests/st/bgp/test_ipip.py
caseydavenport: I left this one out because we now have e2e suites that actually run regularly in e2e in various configurations, so I didn't think this was necessary any more. I'll add it to the doc, not sure how I missed it there!

node/tests/st/bgp/test_ipip.py
caseydavenport: Yeah, I think a more detailed ip monitor test would be good to have. I've added a note to the doc to re-implement this, likely in the k8st suite.

node/tests/st/bgp/test_node_peers.py
caseydavenport: Ah yeah. I think I had considered this covered by the BGPPeer e2e I wrote that uses selectors, which admittedly is subtly different. Although, I do think it's not really worth writing e2e tests for all of the different ways to select a node - especially since we have confd tests that cover this to ensure our config generation is correct: https://github.com/caseydavenport/calico/blob/ae5d06492d3921a6c224366784e218a39f776c75/confd/tests/mock_data/calicoctl/explicit_peering/specific_node/input.yaml#L9-L16

I'll add a new test for this regardless.

node/tests/st/bgp/test_update_ip_addr.py
caseydavenport: I believe this to be suitably covered by existing unit tests in libcalico and confd.

node/tests/st/bgp/test_switch_route_reflector.py
caseydavenport: Yeah, the continuous traffic checking is beyond what it can do - I was intending to add an enhancement for continuous checking in a follow-on (maybe using iperf, or just a looping ping of some sort).

I'll see about just adding it to this batch, though.

node/tests/st/bgp/test_route_reflector_cluster.py
caseydavenport: I don't have a test that configures one node to talk to two different route reflector nodes with explicit BGP peers, but also don't think that's really providing much value over having one node talk to a single route reflector? Or having one node explicitly peered with other nodes?

node/tests/st/bgp/test_route_reflector_cluster.py
caseydavenport: It doesn't - I don't think there's a great way to do this in e2e, but I'll see what I can do. Might be enough to just delete the calico/node pod on one of the nodes and verify consistent connectivity.

caseydavenport: https://tigera.atlassian.net/browse/CORE-12002

node/tests/st/bgp/test_switch_route_reflector.py
caseydavenport: https://tigera.atlassian.net/browse/CORE-12003

I also have a follow-on PR in the works but think it is large enough to warrant its own PR.

node/tests/st/bgp/test_ipip.py
nelljerram: Fair enough.  Maybe there was some subtle difference between cold start and dynamic behaviour - unfortunately there isn't an issue URL to follow to review that - but this feels small enough to risk losing.

node/tests/st/bgp/test_ipip.py
nelljerram: That makes sense.  I think we originally needed this because the Docker modelling would still work if a test was intending to test IPIP but accidentally configured no-encap instead, or if it configured IPIP but BIRD programmed no-encap.

node/tests/st/bgp/test_ipip.py
nelljerram: Following some archaeology, I think this test was for https://github.com/projectcalico/bird/pull/38.  IOW it was for a bug in our patching of BIRD for IP-IP.  I guess that means we wouldn't need this test if we could be sure about switching soon to Mazdak's code for IP-IP.  (And if there is no risk of that code having the same bug!)

node/tests/st/bgp/test_route_reflector_cluster.py
nelljerram: IIRC peering to multiple RRs with the same cluster ID is a fairly normal redundancy measure.  I don't remember the use case for using RRs (or RR groups) with different cluster IDs - @fasaxc any idea?  It's possible there isn't a compelling use case but we coded this to cover the different BGP export behaviour between RRs - i.e. that an RR should not re-export routes to an RR with the same cluster ID as itself, but should re-export routes to an RR with a different cluster ID.

PR #11182, opened 2025-10-14
CNI plugin: check IPv6 LL address is correct

fasaxc:
## Description

- Defensive: toggle interface down/up if we detect incorrect IP.
  Should fix flake in CNI plugin tests.
- Clean up lint (such as unexported return type).
 
## Related issues/PRs

CORE-11954

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
CNI plugin: double-check the IPv6 LL address on the host side of the veth and refresh it if it seems wrong.  Sometimes the kernel uses a stale MAC to calculate it.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/net/ipv6_ll_test.go
+++ b/libcalico-go/lib/net/ipv6_ll_test.go
@@ -30,10 +30,13 @@ func TestMACToIPv6LinkLocal(t *testing.T) {
 			expected: "fe80::0200:00ff:fe00:0000",
 		},
 		{
+			// Normal 48-bit MAC address, should get expanded to 64 bits by
+			// inserting ff:ee.
 			in:       "11:22:33:44:55:66",
 			expected: "fe80::1322:33ff:fe44:5566",
 		},
 		{
+			// EUI-64; no expansion needed.
 			in:       "11:22:33:44:55:66:77:88",
 			expected: "fe80::1322:3344:5566:7788",
 		},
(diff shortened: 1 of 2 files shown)


libcalico-go/lib/net/ipv6_ll_test.go
```diff
@@ -0,0 +1,75 @@
...
+		{
+			in:       "00:00:00:00:00:00",
+			expected: "fe80::0200:00ff:fe00:0000",
+		},
+		{
+			in:       "11:22:33:44:55:66",
+			expected: "fe80::1322:33ff:fe44:5566",
+		},
+		{
+			in:       "11:22:33:44:55:66:77:88",
```
nelljerram, line 40: Can you add a comment to explain what's going on in this case?

nelljerram: (IOW, what's the meaning of an 8 byte MAC address?)

fasaxc:
Have tried to improve the commenting / loop structure to make it clear that it's a retry loop. PTAL

libcalico-go/lib/net/ipv6_ll_test.go
fasaxc: It's a newer standard used on some types of network (such as Infiniband).  It can be plonked straight into an IPv6 without the extra bytes in the middle.

PR #11202, opened 2025-10-17
feat: implement node selector for nodeport svc

### LINKED ISSUE #5495: select nodes to expose NodePort service in eBPF mode
Is it possible to select only specific nodes to expose NodePort service when we use the eBPF mode?

## Expected Behavior

Calico could allow the cluster administrator to select specific nodes to expose `NodePort` service. In a multi-tenant environment, suppose that each tenant has its own set of dedicated worker nodes in the cluster. It should be useful to open ports only in tenant's owned nodes instead of all the cluster nodes, where also other tenants have their own nodes.

tanujd11:
## Description

Adds a `projectcalico.org/nodeSelector` annotation to selectively expose NodePort services on specific nodes based on node labels. Designed for multi-tenant clusters where tenants have dedicated nodes.

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/5495
## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: expose annotated nodeports on selected nodes only
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/proxy/kube-proxy.go
+++ b/felix/bpf/proxy/kube-proxy.go
@@ -178,7 +178,6 @@ func (kp *KubeProxy) run(hostIPs []net.IP) error {
 }
 
 func (kp *KubeProxy) start() error {
-	kp.nodeLabels = kp.fetchNodeLabels()
 	var withLocalNP []net.IP
 	if kp.ipFamily == 4 {
 		withLocalNP = append(withLocalNP, podNPIP)
@@ -186,7 +185,8 @@ func (kp *KubeProxy) start() error {
 		withLocalNP = append(withLocalNP, podNPIPV6)
 	}
 
-	syncer, err := NewSyncer(kp.ipFamily, withLocalNP, kp.frontendMap, kp.backendMap, kp.affinityMap, kp.rt, kp.excludedCIDRs, kp.nodeLabels)
+	// Node labels will be fetched in run() when we have the actual host IPs
+	syncer, err := NewSyncer(kp.ipFamily, withLocalNP, kp.frontendMap, kp.backendMap, kp.affinityMap, kp.rt, kp.excludedCIDRs, nil)
 	if err != nil {
 		return errors.WithMessage(err, "new bpf syncer")
 	}
--- a/felix/bpf/proxy/kube-proxy.go
+++ b/felix/bpf/proxy/kube-proxy.go
@@ -15,11 +15,14 @@
 package proxy
 
 import (
+	"context"
 	"net"
 	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/projectcalico/calico/felix/bpf/bpfmap"
@@ -43,6 +46,8 @@ type KubeProxy struct {
 
 	k8s         kubernetes.Interface
 	hostname    string
+	nodeLabels  map[string]string
+	lastHostIPs []net.IP
 	frontendMap maps.MapWithExistsCheck
 	backendMap  maps.MapWithExistsCheck
 	affinityMap maps.Map
@@ -107,6 +112,29 @@ func (kp *KubeProxy) Stop() {
 	})
 }
 
+func (kp *KubeProxy) fetchNodeLabels() map[string]string {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	node, err := kp.k8s.CoreV1().Nodes().Get(ctx, kp.hostname, metav1.GetOptions{})
+	if err != nil {
+		log.WithError(err).WithField("hostname", kp.hostname).Warn("Failed to fetch node labels, using empty labels")
+		return make(map[string]string)
+	}
+
+	labels := make(map[string]string, len(node.Labels))
+	for k, v := range node.Labels {
+		labels[k] = v
+	}
+
+	log.WithFields(log.Fields{
+		"hostname": kp.hostname,
+		"labels":   labels,
+	}).Debug("Fetched node labels")
+
+	return labels
+}
+
 func (kp *KubeProxy) run(hostIPs []net.IP) error {
 
 	ips := make([]net.IP, 0, len(hostIPs))
@@ -123,6 +151,9 @@ func (kp *KubeProxy) run(hostIPs []net.IP) error {
 	kp.lock.Lock()
 	defer kp.lock.Unlock()
 
+	kp.lastHostIPs = hostIPs
+	kp.nodeLabels = kp.fetchNodeLabels()
+
 	withLocalNP := make([]net.IP, len(hostIPs), len(hostIPs)+1)
 	copy(withLocalNP, hostIPs)
 	if kp.ipFamily == 4 {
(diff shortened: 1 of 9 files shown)


felix/bpf/proxy/kube-proxy.go
```diff
@@ -107,6 +112,29 @@ func (kp *KubeProxy) Stop() {
 	})
 }
 
+func (kp *KubeProxy) fetchNodeLabels() map[string]string {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	node, err := kp.k8s.CoreV1().Nodes().Get(ctx, kp.hostname, metav1.GetOptions{})
```
caseydavenport, line 119: I think we shouldn't be performing any Get requests from within the Felix code. Felix runs as a DaemonSet on every node in the cluster, and for very large clusters this can result in problems with scale.

We will need to feed this via the Syncer which is a Watch API fed by Typha for scale purposes. I believe Felix is already receiving node labels over that API (or at least has the ability to without much fuss)

felix/bpf/proxy/syncer.go
```diff
@@ -1558,3 +1591,23 @@ func K8sSvcWithReapTerminatingUDP() K8sServicePortOption {
 		s.(*servicePort).reapTerminatingUDP = true
 	}
 }
+
+// NewK8sServicePortWithSelector creates a new k8s ServicePort with a node selector
+func NewK8sServicePortWithSelector(clusterIP net.IP, port int, proto v1.Protocol,
```
caseydavenport, line 1596: This looks like it is only used in test code, so should probably live in the `_test.go` file

felix/bpf/proxy/kube-proxy.go
```diff
@@ -219,6 +251,45 @@ func (kp *KubeProxy) OnHostIPsUpdate(IPs []net.IP) {
...
+
+	log.WithFields(log.Fields{
+		"oldLabels": oldLabels,
+		"newLabels": labels,
+	}).Info("Node labels changed, triggering kube-proxy resync")
+
+	// Trigger a resync with the current host IPs
+	// This will recreate the syncer with updated node labels
+	if len(hostIPs) > 0 {
+		kp.OnHostIPsUpdate(hostIPs)
```
caseydavenport, line 287: Looks like we are already getting node label updates streamed to us, so I think instead of calling `OnHostIPsUpdate` here we could introduce a new method for updating the labels that triggers a re-evaluation of Service programming, getting rid of the need for the Get() call mentioned above, right?

felix/bpf/proxy/kube-proxy.go
```diff
@@ -219,6 +251,45 @@ func (kp *KubeProxy) OnHostIPsUpdate(IPs []net.IP) {
...
+// This will trigger a resync to update NodePort service programming based on the new labels.
+func (kp *KubeProxy) OnNodeLabelsUpdate(labels map[string]string) {
+	kp.lock.Lock()
+	oldLabels := kp.nodeLabels
+	kp.nodeLabels = labels
+	hostIPs := kp.lastHostIPs
+	kp.lock.Unlock()
+
+	// Check if labels actually changed
+	labelsChanged := len(oldLabels) != len(labels)
```
caseydavenport, line 264: You should be able to use `maps.Equal` for this to simplify a bit

caseydavenport (changes requested):
Just doing a first pass here - I think the main thing is that we need to get rid of the explicit Get() calls to the API server and instead rely on syncer data.

felix/bpf/proxy/kube-proxy.go
tomastigera: yes, I was going to say that we should perhaps use the same/similar approach as when updating the host IPs. Both is expected to be rather infrequent event.

felix/bpf/proxy/kube-proxy.go
```diff
@@ -123,6 +151,9 @@ func (kp *KubeProxy) run(hostIPs []net.IP) error {
 	kp.lock.Lock()
 	defer kp.lock.Unlock()
 
+	kp.lastHostIPs = hostIPs
+	kp.nodeLabels = kp.fetchNodeLabels()
```
tomastigera, line 155: you do not need to fetch the labels, you already got them in `OnNodeLabelsUpdate()`

felix/bpf/proxy/kube-proxy.go
tomastigera: Follow the `OnHostIPsUpdate` example. To avoid possibly restarting kube-proxy 2x, it might be the best to move https://github.com/projectcalico/calico/blob/master/felix/dataplane/linux/bpf_route_mgr.go#L590 to https://github.com/projectcalico/calico/blob/master/felix/dataplane/linux/bpf_route_mgr.go#L242 if there was any change and send labels and hostIPs as a single update.

tomastigera (commented):
Note that this would only exclude external traffic from being handled by the nodeport. It would not exclude a pod from reaching the nodeport via the excluded host's IP. But that should not really matter as there is no forwarding from that IP's node to the node with the backend. Nodeport is always resolved straight to the backend pod IP on the client pod's node.

felix/bpf/proxy/kube-proxy.go
tanujd11: yes, agreed. I am updating it. I got a bit confused there. Thanks.

tomastigera:
Also note that bumping k8s libs in current master https://github.com/kubernetes/kubernetes/pull/133059 we no need to incorporate node manager https://github.com/kubernetes/kubernetes/pull/133059 so you may get all you need in a much simpler way if you hold off a little bit.

tomastigera:
Turns out that the node selector is not that helpful after all :disappointed: so we would need to send the labels as I proposed. We would like to decouple the callback and the proxy execution so that reprogramming kube-proxy would not block the rest of felix from running.

tomastigera:
@tanujd11 just to follow up @aaaaaaaalex is now working on some of the base work for this - it did not get completely off our radar.

tomastigera:
@tanujd11 what is the exact scenario? Excluding exposing a port on a certain node is sort of like placing a deny policy that would block access to the port on selected nodes. Does you LB already understand which nodes not to use to target the service? I guess you do not want to run the backend pods on these nodes and you have a way how to do that. I guess your LB would only target the nodes with the pods.

So is this change just to not implemented the nodeport on these nodes to make things clean? Or to leave the nodeport for the use of other services on these nodes?

lwr20:
> @tanujd11 what is the exact scenario? Excluding exposing a port on a certain node is sort of like placing a deny policy that would block access to the port on selected nodes. Does you LB already understand which nodes not to use to target the service? I guess you do not want to run the backend pods on these nodes and you have a way how to do that. I guess your LB would only target the nodes with the pods.
> 
> So is this change just to not implemented the nodeport on these nodes to make things clean? Or to leave the nodeport for the use of other services on these nodes?

We even have this use-case in a policy tutorial in the docs: https://docs.tigera.io/calico/latest/network-policy/services/kubernetes-node-ports#allow-ingress-traffic-to-specific-node-ports

caseydavenport:
@tomastigera @tanujd11 @aaaaaaaalex what's the latest on this guy?

aaaaaaaalex:
If we have a real use-case for this feature which isn't already solved with policy features, my linked PR (merged) plumbed cluster host metadata down into BPF KP, which might help facilitate the feature for the eBPF dataplane.

That being said, the feature would probably increase our divergence from K8s KP. It would also increase divergence between our eBPF dataplane features and the other dataplanes, unless we found a way to get this feature into other dataplanes too.

PR #11076, opened 2025-09-25
updates to support custom image path in operator

radTuti:
## Description

- update template for pkg/components/image.go to allow setting image path
  - get image path from product registry and operator image for calico and tigera operator respectively
- update pinned component list generated for operator
  - remove setting default image as all components should have default image defined in operator
  - remove flannel as it is not used by operator
- update code generation check to use the operator org, repo & branch for the (hash)release in progress
  - to allows testing with a fork of operator
- update manifest generation to replace static manifests even if version is master as registry may have changed.
- update tests where relevant
- update operator components file used by Node KinD tests to new version

## Related issues/PRs

- https://github.com/tigera/operator/pull/4163
- https://github.com/projectcalico/calico/pull/11191 (NodeST fix)

## Todos

- [x] Tests

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #11207, opened 2025-10-17
[v3.31] fix(release-tool): operator built to properly support private registry

radTuti:
## Description

This fixes private registy installation of Calico v3.31 hashrelease that are currently resolving images to `<registry>/<component>` instead of `<registry>/calico/<component>`. For operator to properly resolve private registries, it expects `calico/` be part of the image name.

**NOTE**: this fix is only for v3.31, master is already fixed differently with the combination of https://github.com/tigera/operator/pull/4163 and https://github.com/projectcalico/calico/pull/11076.

## Related issues/PRs

- error introduced in #10524 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/release/internal/pinnedversion/pinnedversion.go
+++ b/release/internal/pinnedversion/pinnedversion.go
@@ -223,11 +223,9 @@ func GenerateOperatorComponents(srcDir, outputDir string) (registry.OperatorComp
 	}
 	// Remove components that are not needed in the operator components file.
 	// These either do not produce images or are not used by the operator.
-	components := pinnedVersion.Components
 	for _, c := range operatorIgnoreComponents {
-		delete(components, c)
+		delete(pinnedVersion.Components, c)
 	}
-	pinnedVersion.Components = components
 
 	logrus.Info("Generating operator components file")
 	operatorComponentsFilePath := filepath.Join(srcDir, operatorComponentsFileName)
@@ -304,10 +302,10 @@ func LoadHashrelease(repoRootDir, outputDir, hashreleaseSrcBaseDir string, lates
 // RetrieveImageComponents retrieves the images from Calico components in the pinned version file that produce images.
 // It also adds the Tigera operator and its init image to the returned map.
 //
-// Images are expected to be in the format "<registry>/<image-name>" where <registry> includes the registry and image path.
+// Images returned from this function are expected to be in the format "<registry>/<image-name>" where <registry> includes the registry and image path. However, the versions.yml file represents images differently, including the image path as the first part of the image name instead of as part of the registry.
 // As a result, the image path is stripped from the image name to only return the image name so that images are properly formatted.
 //
-// For example, with "calico/node", this allows the fully qualified image to be "quay.io/calico/node" when registry is included.
+// For example, if the image name is "calico/node", this allows the fully qualified image to be "quay.io/calico/node" when a registry of `quay.io/calico` is prepended.
 // and prevents duplication of the image path(i.e. "quay.io/calico/calico/node").
 func RetrieveImageComponents(outputDir string) (map[string]registry.Component, error) {
 	pinnedVersion, err := retrievePinnedVersion(outputDir)
--- a/release/internal/pinnedversion/release.go
+++ b/release/internal/pinnedversion/release.go
@@ -57,12 +57,13 @@ func (p *CalicoReleaseVersions) GenerateFile() (version.Versions, error) {
 	return nil, nil
 }
 
+// ImageList return a list of Calico images built by this repo for release validation.
+// It excludes flannel and the Tigera operator images as those are validated separately.
 func (p *CalicoReleaseVersions) ImageList() ([]string, error) {
 	components, err := RetrieveImageComponents(p.Dir)
 	if err != nil {
 		return nil, err
 	}
-	// Exclude flannel (not built by Calico) and the Tigera operator images.
 	componentNames := make([]string, 0, len(components))
 	for name, component := range components {
 		if strings.HasPrefix(component.Image, p.OperatorCfg.Image) || name == flannelComponentName {
--- a/release/pkg/manager/operator/manager.go
+++ b/release/pkg/manager/operator/manager.go
@@ -155,7 +155,7 @@ func (o *OperatorManager) modifyComponentsImagesFile() error {
 		"Registry": o.registry,
 		// Trim "/calico" from the product registry since operator expects the registry only.
 		// The image path "calico" is part of image name and this template adds "/" after the registry.
-		"ProductRegistry": strings.TrimPrefix(o.productRegistry, "/calico"),
+		"ProductRegistry": strings.TrimSuffix(o.productRegistry, "/calico"),
 		"Year":            time.Now().Format("2006"),
 	}); err != nil {
 		logrus.WithError(err).Errorf("Failed to write to file %s", destFilePath)
@@ -171,9 +171,6 @@ func (o *OperatorManager) Build() error {
 	if o.validate {
 		if err := o.PreBuildValidation(); err != nil {
 			return err
-		}
-		if !strings.HasSuffix(o.productRegistry, "/calico") {
-			return fmt.Errorf("operator does not support product registry %q, it must end with %q", o.productRegistry, "/calico")
 		}
 	}
 	component, componentsVersionPath, err := pinnedversion.GenerateOperatorComponents(o.tmpDir, o.outputDir)


release/pkg/manager/operator/manager.go
```diff
@@ -153,7 +153,7 @@ func (o *OperatorManager) modifyComponentsImagesFile() error {
 
 	if err := tmpl.Execute(dest, map[string]string{
 		"Registry":        o.registry,
-		"ProductRegistry": o.productRegistry,
+		"ProductRegistry": strings.TrimPrefix(o.productRegistry, registry.CalicoNamespace),
```
stevegaossou: Why would we need to trim "calico" off the front of the product registry value?

stevegaossou: Also, what is the difference in values between the `registry` field and the `productRegistry`? 

The comments for them seem to imply they are the same thing: 
- https://github.com/projectcalico/calico/blob/master/release/pkg/manager/operator/manager.go#L81-L85

caseydavenport: Yeah, this line needs a lot of explanation with code comments. It feels pretty "magic" to me and it's very much not obvious why it's here.

release/internal/registry/registry.go
```diff
@@ -22,3 +22,5 @@ var DefaultCalicoRegistries = []string{
 	"asia.gcr.io/projectcalico-org",
 	"us.gcr.io/projectcalico-org",
 }
+
+const CalicoNamespace = "calico"
```
caseydavenport: What is Namespace in this context? This isn't a term we use elsewhere when constructing image names.

Does it belong in the registry package? It isn't a registry?

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -293,12 +308,20 @@ func RetrieveImageComponents(outputDir string) (map[string]registry.Component, e
...
+		if slices.Contains(noImageComponents, name) || name == flannelComponentName {
 			delete(components, name)
 			continue
 		}
-		components[name] = normalizeComponent(name, component)
+		img := registry.ImageMap[name]
+		if img == "" {
+			img = name
+		}
+		// Strip "calico/" prefix from image name since registry is <registry>/calico
```
caseydavenport: This is pretty confusing. Why are we including "calico/" in the registry at all? Sounds like we're including `calico/` in multiple parts of the image construction? (registry + image name)?

I think we need to be more precise about the syntactical components that make up an fully qualified image, and what belongs in each.

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -293,12 +308,20 @@ func RetrieveImageComponents(outputDir string) (map[string]registry.Component, e
 	}
 	components := pinnedVersion.Components
 	for name, component := range components {
-		// Remove components that do not produce images.
-		if utils.Contains(excludedComponents, name) {
+		// Remove components that should be excluded.
+		// Either because they do not have an image,
+		// or not built by Calico.
```
caseydavenport:
```suggestion
		// Remove components that should be excluded. Either because they do not have an image, or not built by Calico.
```

Nit, but these comment lines are way too short.

radTuti:
> * Does this change need to be propagated to `master` (moving forward)?

No, there is an update in master (https://github.com/projectcalico/calico/pull/11076) that resolved this in a better way.

release/pkg/manager/operator/manager.go
radTuti: I have fixed the comment (comment fix for master is in https://github.com/projectcalico/calico/pull/11210).

`registry` is the registry for operator images
`productRegistry` is the registry for calico components images

There is a _long-running_ discrepancy between what is a registry for operator and calico. 
Currently Calico considers a registry as `quay.io/calico` which is the registry + imagepath/namespace; meanwhile, operator considers registry as `quay.io` with the imagepath/namespace as part of the image name (i.e. `tigera/operator`).

I do think resolving this discrepancy is beyond the scope of this fix

release/internal/registry/registry.go
stevegaossou: Yes, when I see "namespace" mentioned I cannot help but immediately associated it with Kubernetes namespaces. 

I'm not sure if there is an official term for the part of the image path that is after the image registry domain (i.e. the "calico/" part). But semantically they way we're using it means it represents the product type (calico being OSS). 

I believe the lingo for the part of the image path that is "calico/<component>" like "calico/node" is the image repository. Basically whatever is between the registry domain and the ":" that separate the image path from the image tag is contained a qualifying name for the image repository. 

Anyways, I guess this comment was more for myself (to refresh my own memory). You can ignore this. 🤣

stevegaossou:
@caseydavenport I'll let you be the approver for this PR, since your comments are more in-depth than mine. 

I think Tuti has already address my feedback.

caseydavenport:
@stevegaossou I can do that, but tbh I am still piecing together what the actual issue and context are here! I don't fully understand what has changed that has caused this to be a problem, or why.

@radTuti is there a concise summary somewhere of what the overall strategy is for building "fully qualified" image names?

radTuti:
> is there a concise summary somewhere of what the overall strategy is for building "fully qualified" image names?

There is not a concise summary in one place. That [section of code](https://github.com/projectcalico/calico/blob/release-v3.31/release/internal/pinnedversion/pinnedversion.go#L289-L308) is used to construct fully qualified images in calico [here](https://github.com/projectcalico/calico/blob/release-v3.31/release/internal/registry/component.go#L34) for [sending images to ISS](https://github.com/projectcalico/calico/blob/release-v3.31/release/pkg/manager/calico/manager.go#L569) and [here](https://github.com/projectcalico/calico/blob/release-v3.31/release/pkg/postrelease/images_test.go#L34) for post-release testing.

@caseydavenport I did also add a comment to the `RetrieveImageComponents` fn.

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -211,9 +221,14 @@ func GenerateOperatorComponents(srcDir, outputDir string) (registry.OperatorComp
...
 	}
-	for name, component := range pinnedVersion.Components {
-		pinnedVersion.Components[name] = normalizeComponent(name, component)
+	// Remove components that are not needed in the operator components file.
+	// These either do not produce images or are not used by the operator.
+	components := pinnedVersion.Components
+	for _, c := range operatorIgnoreComponents {
+		delete(components, c)
 	}
+	pinnedVersion.Components = components
```
caseydavenport:
```suggestion
	for _, c := range operatorIgnoreComponents {
		delete(pinnedVersion.Components, c)
}
```

Do we need the intermediate variable here?

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -286,19 +301,37 @@ func LoadHashrelease(repoRootDir, outputDir, hashreleaseSrcBaseDir string, lates
 	}, nil
 }
 
+// RetrieveImageComponents retrieves the images from Calico components in the pinned version file that produce images.
+// It also adds the Tigera operator and its init image to the returned map.
+//
+// Images are expected to be in the format "<registry>/<image-name>" where <registry> includes the registry and image path.
+// As a result, the image path is stripped from the image name to only return the image name so that images are properly formatted.
```
caseydavenport: I think we need to do a follow-on to fix this properly. It's really confusing if sometimes the "image" part of the fully qualified name includes "calico/" and sometimes it's the registry. 

It seems to me like we have a few pieces that we should be standardizing on across both operator and here:

- **Registry** - e.g., quay.io, docker.io, gcr.io
- **Org** - e.g., calico, tigera, projectcalico-org
- **Path** - typically unset, but could be a sub-path after the organization.
- **Name** - e.g., node, typha, cni, etc. 
- **Version** - i.e., the tag. 

Then, a fully qualified image would look like this:

```
<registry>/<org>/<path>/<name>:<version>
```

There's shouldn't ever be any additional modification of the inputs needed to build the fully qualified representation, and the logic that generates the fully-qualified version should be centralized and common.

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -286,19 +301,37 @@ func LoadHashrelease(repoRootDir, outputDir, hashreleaseSrcBaseDir string, lates
 	}, nil
 }
 
+// RetrieveImageComponents retrieves the images from Calico components in the pinned version file that produce images.
+// It also adds the Tigera operator and its init image to the returned map.
+//
+// Images are expected to be in the format "<registry>/<image-name>" where <registry> includes the registry and image path.
```
caseydavenport: > Images are expected to be in the format "<registry>/<image-name>"

Does this mean images in the versions file are expected to be of this format? Or the images returned from this function are expected to be in this format? 

> registry> includes the registry and image path.

This is self-referential. Can we define registry without using the word registry?

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -286,19 +301,37 @@ func LoadHashrelease(repoRootDir, outputDir, hashreleaseSrcBaseDir string, lates
...
+// As a result, the image path is stripped from the image name to only return the image name so that images are properly formatted.
+//
+// For example, with "calico/node", this allows the fully qualified image to be "quay.io/calico/node" when registry is included.
```
caseydavenport:
```suggestion
// For example, if the image name is "calico/node", this allows the fully qualified image to be "quay.io/calico/node" when a registry of `quay.io/calico` is prepended.
```

release/internal/pinnedversion/pinnedversion.go
caseydavenport:
```suggestion
// Images returned from this function are expected to be in the format "<registry>/<image-name>" where <registry> includes the registry and image path. However, the versions.yml file represents images differently, including the image path as the first part of the image name instead of as part of the registry.
```

I think this is correct?

release/internal/pinnedversion/release.go
```diff
@@ -62,12 +62,13 @@ func (p *CalicoReleaseVersions) ImageList() ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Exclude flannel (not built by Calico) and the Tigera operator images.
```
caseydavenport: This is obvious from reading the code. The interesting thing would be to call out _why_ we exclude these images :)

release/pkg/manager/operator/manager.go
```diff
@@ -75,10 +75,10 @@ type OperatorManager struct {
 	// outputDir is the absolute path to the output directory
 	outputDir string
 
-	// productRegistry is the registry to use for product images
+	// registry is the registry to use for operator (e.g. quay.io)
 	registry string
 
-	// productRegistry is the registry to use for product images
+	// productRegistry is the registry to use for product images (e.g. quay.io/calico)
```
caseydavenport, line 81: "product images" is really vague. Like, the operator is part of our "product".

I think this should just be `operatorRegistry` and `calicoRegistry`?

I'd suggest tackling this as a separate PR though - you didn't introduce these variables here.

release/pkg/manager/operator/manager.go
```diff
@@ -152,8 +152,10 @@ func (o *OperatorManager) modifyComponentsImagesFile() error {
 	}
 
 	if err := tmpl.Execute(dest, map[string]string{
-		"Registry":        o.registry,
-		"ProductRegistry": o.productRegistry,
+		"Registry": o.registry,
+		// Trim "/calico" from the product registry since operator expects the registry only.
+		// The image path "calico" is part of image name and this template adds "/" after the registry.
+		"ProductRegistry": strings.TrimPrefix(o.productRegistry, "/calico"),
```
caseydavenport: Isn't `productRegistry` something like `quay.io/calico` ? Shouldn't this be `TrimSuffix`?

release/pkg/manager/operator/manager.go
```diff
@@ -167,9 +169,12 @@ func (o *OperatorManager) Build() error {
 		return fmt.Errorf("operator manager builds only for hash releases")
 	}
 	if o.validate {
-		if err := o.PreBuildValidation(o.tmpDir); err != nil {
+		if err := o.PreBuildValidation(); err != nil {
 			return err
 		}
+		if !strings.HasSuffix(o.productRegistry, "/calico") {
```
caseydavenport: I don't think we should be baking in assumptions like this - it's going to be awkward the moment we have an exception to this rule. 

Does the operator code really not allow for us to specify images like `quay.io/tigera/node`?

release/pkg/manager/operator/manager.go
```diff
@@ -75,10 +75,10 @@ type OperatorManager struct {
 	// outputDir is the absolute path to the output directory
 	outputDir string
 
-	// productRegistry is the registry to use for product images
+	// registry is the registry to use for operator (e.g. quay.io)
```
caseydavenport, line 78: It's pretty confusing to have two variables with the word "registry" in them that have different syntaxes (this one claims to be `quay.io` and the productRegistry claims to be `quay.io/calico`) - Is this correct?

caseydavenport (changes requested):

release/pkg/manager/operator/manager.go
radTuti: yes. this is due to the current discrepancy b/w operator & monorepo

release/pkg/manager/operator/manager.go
radTuti: I do not think Calico will have components like `quay.io/tigera/<image-name>` but will remove since not pertinent

release/internal/pinnedversion/pinnedversion.go
```diff
@@ -286,19 +299,37 @@ func LoadHashrelease(repoRootDir, outputDir, hashreleaseSrcBaseDir string, lates
...
+		if slices.Contains(noImageComponents, name) || name == flannelComponentName {
 			delete(components, name)
 			continue
 		}
-		components[name] = normalizeComponent(name, component)
+		img := registry.ImageMap[name]
+		if img == "" {
+			img = name
+		}
+		component.Image = img
```
caseydavenport, line 326: Just noticed, but this phrasing is a bit awkward. 

We assign a local variable `img`, and halfway through processing the name we switch to operating on `component.Image`

I'd suggest either:
- doing all the processing on `img` and assigning the value to `component.Image` only at the very end.
- getting rid of the `img` intermediate, and assigning to `component.Image` throughout.

caseydavenport (approved):
One more comment re: code clarity, but otherwise LGTM.

We should definitely revisit this to clean this up - there are a few too many different meanings for the words "image" and "registry" throughout.

PR #11212, opened 2025-10-19
Fix nil pointer panic in LoadBalancer poolContains

oyvindh:
## Description

Fixes a nil pointer panic in the LoadBalancer controller's `poolContains()` function.

When invalid IP addresses are present in the IPAM allocation tracker (e.g., due to data corruption, migration issues, or race conditions), `net.ParseIP()` returns `nil`, but the code attempted to use this nil pointer in `cidr.Contains(ip)`, causing the controller to crash.

## Changes

- Add nil check after `net.ParseIP()` in `poolContains()` 
- Add warning log when invalid IPs are encountered to aid diagnosis
- Add unit test to verify invalid IPs don't cause panics

## Testing

Added `TestSyncServiceWithInvalidIP` which:
1. Places an invalid IP address in the allocation tracker
2. Calls `syncService()` which triggers the code path
3. Verifies it doesn't panic


caseydavenport:
Thanks @oyvindh ! 

this LGTM - did you encounter this or did you just spot it in a code read? 

If you encountered this, how did you end up triggering it? Was something writing invalid IPAM data?

oyvindh:
> Thanks @oyvindh !
> 
> this LGTM - did you encounter this or did you just spot it in a code read?
> 
> If you encountered this, how did you end up triggering it? Was something writing invalid IPAM data?

I did trigger it, I tried a long time to figure out where the bad data came from, but could not find it. Probably something in my crazy setup.

PR #11232, opened 2025-10-21
Remove tier prefixing restriction for policy names

caseydavenport:
## Description

Since we introduced the concept of tiers, we have used validation to ensure that NetworkPolicy names were prefixed with the tier they belonged to (with the exception of the default tier). This was done since we wanted the Tier to be an identifier of the policy, but leads to an unintuitive UX and additional code complexity behind the scenes. 

Additionally, as part of a larger effort to remove the need for the Calico API server, we want to remove this validation. This PR changes all policy types such that the TIer is no longer an ID field, instead identifying a policy entirely by its (kind, namespace, name).

It impacts a few internal and external APIs, including:

- The projectcalico v3 API validation. 
- Internal model API used by Felix and Typha
- Felix dataplane protobuf API
- Flow log API

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Naming restrictions for NetworkPolicy, GlobalNetworkPolicy, and all staged policy variants have been relaxed to no longer require the policy tier prefix for policies in non-default tiers. Calico will perform an automatic upgrade of existing policies to be compatible with this change on upgrade. 
```

## Reminder for the reviewer

Other optional labels:


libcalico-go/lib/clientv3/globalnetworkpolicy.go
```diff
@@ -142,50 +141,21 @@ func (r globalNetworkPolicies) Get(ctx context.Context, name string, opts option
 	if out != nil {
 		// Add the tier labels if necessary
 		out.GetObjectMeta().SetLabels(defaultTierLabelIfMissing(out.GetObjectMeta().GetLabels()))
-		// Fill in the tier information from the policy name if we find it missing.
```
caseydavenport: Should update this to use the spec field instead of simply removing it.

felix/bpf/polprog/pol_prog_builder.go
```diff
@@ -172,8 +173,10 @@ type Rule struct {
 }
 
 type Policy struct {
-	Name  string
-	Rules []Rule
+	Name      string
+	Namespace string
+	Kind      string
```
fasaxc: Nit: I'd order from most general to least. Kind, Namespace, Name?

felix/calc/states_for_test.go
```diff
@@ -267,23 +271,36 @@ var localEp2WithPolicyAndTier = withPolicyAndTier.withKVUpdates(
...
-	[3]string{"pol-1", "pol-2", "pol-3"},
+	[3]types.PolicyID{
+		{Name: "pol-1", Kind: v3.KindGlobalNetworkPolicy},
+		{Name: "pol-2", Kind: v3.KindGlobalNetworkPolicy},
+		{Name: "pol-3", Kind: v3.KindGlobalNetworkPolicy},
+	},
 )
 
-func commercialPolicyOrderState(policyOrders [3]float64, expectedOrder [3]string) State {
+func commercialPolicyOrderState(policyOrders [3]float64, expectedOrder [3]types.PolicyID) State {
```
fasaxc: Misnamed now :-)

felix/calc/calc_graph_fv_test.go
```diff
@@ -685,7 +685,6 @@ var _ = Describe("Async calculation graph state sequencing tests:", func() {
 		for _, expander := range testExpanders() {
```
fasaxc, line 691: Now that the policy's tier can change without changing its name, it'd be good to have a test that goes from say `localEp1WithPolicy` to `localEp1WithPolicyAndTier`.

felix/calc/policy_lookup_cache.go
```diff
@@ -186,19 +190,28 @@ func (pc *PolicyLookupsCache) updatePolicyRulesNFLOGPrefixes(key model.PolicyKey
...
+	tier, ok := pc.keyToTier[key]
+	if !ok {
+		// We have never seen this policy. Nothing to do.
+		log.Warnf("Attempted to remove unknown policy %v from PolicyLookupsCache", key)
+		return
+	}
+	delete(pc.keyToTier, key)
+
 	// If this is the last entry for the tier, remove the default action entries for the tier.
 	// Increment the reference count so that we don't keep adding tiers.
```
fasaxc: (Old) comment looks wrong, this is the decrement path

felix/calc/policy_lookup_cache.go
```diff
@@ -370,15 +388,15 @@ const (
...
-	Tier string
 	// The policy or profile name. This has the tier removed from the name. If this is blank, this represents
 	// a "no match" rule. For k8s policies, this will be the full v3 name (knp.default.<k8s name>) - this avoids
 	// name conflicts with Calico policies.
 	Name string
 	// The namespace. This is only non-blank for a NetworkPolicy type. For Tiers, GlobalNetworkPolicies and the
 	// no match rules this will be blank.
 	Namespace string
+	// The kind of policy.
+	Kind string
```
fasaxc: Nit: I think Kind, Namespace, Name makes more sense from an ordering point of view (but beware any places where these get initialized without field names if you make that tweak!)

felix/calc/active_rules_calculator.go
```diff
@@ -353,20 +353,24 @@ func (arc *ActiveRulesCalculator) onMatchStarted(selID, labelId interface{}) {
...
 		if !known {
 			log.WithField("policy", polKey).Panic("Policy active but missing from allPolicies.")
 		}
 		arc.sendPolicyUpdate(polKey, policy)
 	}
+
 	if labelId, ok := labelId.(model.EndpointKey); ok {
 		for _, l := range arc.PolicyMatchListeners {
-			l.OnPolicyMatch(polKey, labelId)
+			l.OnPolicyMatch(polKey, labelId, policy.Tier)
```
fasaxc: I don't think you should be passing tier from here.  Since tier is no longer an ID field, this method won't fire when it changes and the policy resolver will get out of sync.

felix/calc/policy_resolver.go
```diff
@@ -145,11 +145,18 @@ func (pr *PolicyResolver) markEndpointsMatchingPolicyDirty(polKey model.PolicyKe
 	})
 }
 
-func (pr *PolicyResolver) OnPolicyMatch(policyKey model.PolicyKey, endpointKey model.EndpointKey) {
+func (pr *PolicyResolver) OnPolicyMatch(policyKey model.PolicyKey, endpointKey model.EndpointKey, tier string) {
```
fasaxc: Passing tier in here feels a bit hacky, could you refactor this object to store the matches as-is and then do the resolution in the Flush method (which is only called after all nodes have seen the update, and may be batched up even more than that to improve throughput).

Note: Flush was added because this object turned out to be a bottleneck when many policies match an endpoint.  Previously we were doing a sort for every match update, which gave us O(n^2) behavior!

felix/calc/policy_sorter.go
```diff
@@ -318,7 +355,15 @@ func PolKVLess(i, j PolKV) bool {
 	// We map the default order to +Inf, which compares equal to itself so,
 	// this "just works".
 	if i.Value.Order == j.Value.Order {
-		// Order is equal, use name as tie-break.
+		// Order is equal, use name to break ties.
+		if i.Key.Name == j.Key.Name {
```
fasaxc: Wonder if this is consistent with the ordering that we used to use, based on the name prefixes?  IIRC, before we had `namespace/knp.foo`, should we mimic that sorting here?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4464,10 +4466,21 @@ func (m *bpfEndpointManager) delRoute(cidr ip.CIDR) {
...
+
 // updatePolicyCache modifies entries in the cache, adding new entries and marking old entries dirty.
-func (m *bpfEndpointManager) updatePolicyCache(name string, owner string, inboundRules, outboundRules []*proto.Rule) {
+func (m *bpfEndpointManager) updatePolicyCache(id types.PolicyID, inboundRules, outboundRules []*proto.Rule) {
+	// Build a unique string name for the policy
+	name := id.String()
+	m.updateCache(id, name, "Policy", inboundRules, outboundRules)
+}
+
+// TODO: Do we need to consider the owner here for identical names between Profile and Policy?
```
fasaxc: TODO ?

felix/dataplane/linux/endpoint_mgr.go
```diff
@@ -1584,34 +1578,28 @@ func (m *endpointManager) GetRawHostEndpoints() map[types.HostEndpointID]*proto.
 	return m.rawHostEndpoints
 }
 
-func (m *endpointManager) groupPolicies(tierName string, names []string, direction rules.PolicyDirection) []*rules.PolicyGroup {
-	if len(names) == 0 {
+func (m *endpointManager) groupPolicies(tierName string, policies []*proto.PolicyID, direction rules.PolicyDirection) []*rules.PolicyGroup {
```
fasaxc: Looks like `tierName` can be removed now.  I think the ref counting takes care of moving a group between tiers (group chain should just get retained, which is a win).

typha/pkg/syncproto/sync_proto.go
```diff
@@ -286,11 +284,12 @@ type MsgDecoderRestart struct {
 
 // MsgACK is a general-purpose ACK message, currently used during the initial handshake to acknowledge the
 // switch to compressed mode.
-type MsgACK struct {
-}
-type MsgSyncStatus struct {
-	SyncStatus api.SyncStatus
-}
+type (
```
fasaxc: Not sure about this combination for types; better to put in a whitespace line?

felix/collector/goldmane/client.go
```diff
@@ -272,6 +272,10 @@ func toFlowPolicySet(policies []*proto.PolicyHit) flowlog.FlowPolicySet {
 		if s, err := pol.ToString(); err != nil {
 			logrus.WithError(err).WithField("policy", pol).Error("Failed to convert policy hit to string")
 		} else {
+			logrus.WithFields(logrus.Fields{
+				"policyHit": pol,
+				"policy":    s,
+			}).Debug("Converted policy hit to string")
```
fasaxc: Not sure if you're on the hot path here, but if so you might want to make this conditional on debug being enabled.  WithFields is awful for garbage.

felix/collector/collector_test.go
```diff
@@ -244,73 +272,85 @@ var (
...
+	return p
+}
+
 // Nflog prefix test parameters
 var (
-	defTierAllowIngressNFLOGPrefix   = [64]byte{'A', 'P', 'I', '0', '|', 'd', 'e', 'f', 'a', 'u', 'l', 't', '.', 'p', 'o', 'l', 'i', 'c', 'y', '1'}
-	defTierAllowEgressNFLOGPrefix    = [64]byte{'A', 'P', 'E', '0', '|', 'd', 'e', 'f', 'a', 'u', 'l', 't', '.', 'p', 'o', 'l', 'i', 'c', 'y', '1'}
-	defTierDenyIngressNFLOGPrefix    = [64]byte{'D', 'P', 'I', '0', '|', 'd', 'e', 'f', 'a', 'u', 'l', 't', '.', 'p', 'o', 'l', 'i', 'c', 'y', '2'}
-	defTierDenyEgressNFLOGPrefix     = [64]byte{'D', 'P', 'E', '0', '|', 'd', 'e', 'f', 'a', 'u', 'l', 't', '.', 'p', 'o', 'l', 'i', 'c', 'y', '2'}
+	defTierAllowIngressNFLOGPrefix   = toprefix("API0|GlobalNetworkPolicy/policy1")
```
fasaxc: 👍

felix/fv/flow_logs_goldmane_staged_test.go
```diff
@@ -325,8 +325,8 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ goldmane flow log with stag
 
 		if !bpfEnabled {
 			// Wait for felix to see and program some expected nflog entries, and for the cluster IP to appear.
-			Eventually(getRuleFunc(tc.Felixes[0], "APE0|default.ep1-1-allow-all"), "10s", "1s").ShouldNot(HaveOccurred())
-			Eventually(getRuleFunc(tc.Felixes[1], "APE0|default.ep1-1-allow-all"), "10s", "1s").Should(HaveOccurred())
+			Eventually(getRuleFunc(tc.Felixes[0], "APE0|GlobalNetworkPolicy/default.ep1-1-allow-all"), "10s", "1s").ShouldNot(HaveOccurred())
```
fasaxc: First time I've clocked what the rules look like; is it OK to bloat the prefix with the full type name? We've been pretty terse with the other prefixes, maybe that has a slight perf impact (or maybe we're just trying to leave as much room as possible for the policy name)?

Could compress it to a single char `G=GNP` `N=NP` `K=K8s NP`, ...?

libcalico-go/lib/backend/model/keys.go
```diff
@@ -354,6 +350,16 @@ func keyFromDefaultPathInner(path string, parts []string) Key {
 				case "labels":
 					return ProfileLabelsKey{ProfileKey: ProfileKey{pk}}
 				}
+			default:
+				// Policy key of form /calico/v1/policy/<Kind>/<Namespace>/<Name>
```
fasaxc, line 360: This is a breaking change on the typha/felix API, during upgrade backlevel felix talking to uplevel typha will see no policies and vice versa.

felix/proto/felixbackend.proto
```diff
@@ -236,8 +236,13 @@ message ActivePolicyRemove {
 }
 
 message PolicyID {
-  string tier = 1;
+  // The "type" field has been deprecated and removed in favor of "types".
```
fasaxc: Comment copy/pasta

felix/rules/endpoints_test.go
```diff
@@ -23,21 +23,45 @@ import (
...
+var (
+	// Expected ID suffixes for policies used in tests.
+	gnpAI = "_ZMdvlDxVmjIQZiLZp4U"
+	gnpBI = "_8W91OfQf0rhQJtk1nWi"
+	gnpAE = "_BOgFuOi8tbV0awqZFjg"
+	gnpBE = "_TPOgx2WKFeXH7YpWFda"
+
+	// These don't exceed the length limit for iptables chain names, so
+	// they do not get hashed.
+	npAI  = "NetworkPolicy/ai"
```
fasaxc: Can we shorten the kind prefixes; very useful to have names that don't get hashed as much as possible.  Some customers use short names to make it easier to grok the rules.

libcalico-go/lib/clientv3/globalnetworkpolicy.go
fasaxc, line 145: Any upgrade concerns here, could there be policies stored without tier?

fasaxc (changes requested):
🎉 

Might have been nice to have a preliminary PR to do the migration cleanup just to take all of that out of the review, but thank god for the new Github PR large PR mode(!)

felix/calc/policy_resolver.go
caseydavenport: Yeah, correctly spotted the hack :grin: 

I had meant to come back to this and then missed it in my last passes. I'll take a look at refactoring.

felix/calc/policy_sorter.go
caseydavenport: It's not consistent, and I think we can't make it exactly consistent either because the old sorting used to include the tier in the name value. I might be able to make it _closer_ but not exactly the same - not sure if that is worthwhile since we don't guarantee ordering of two policies with the same order?

felix/dataplane/linux/bpf_ep_mgr.go
caseydavenport: Whoops, yeah I fixed this but left the TODO in.

typha/pkg/syncproto/sync_proto.go
caseydavenport: yeah, that was my editor getting aggressive about types next to each other.

felix/rules/endpoints_test.go
caseydavenport: Yeah, I considered this (and can still do this) although nftables has a substantially larger name limit which means we can be more explicit and I think writing the full kind is much easier for users to parse than abbreviations of them.

So, looking forward I think this is a much nicer user experience (although it will hash more things in iptables mode, so can understand if we still want to prioritize readability there instead).

libcalico-go/lib/clientv3/globalnetworkpolicy.go
caseydavenport: We are going to (unfortunately) need a migration step as part of this work regardless, which will give us an opportunity to ensure there are no blank tiers (I think if it's blank, we know it is the default tier anyway though). 

We need the migration regardless because the v1 storage always uses `tier.name` and the outcome we want from this work is for the v1 storage to use the same name as the v3 API.

libcalico-go/lib/backend/model/keys.go
caseydavenport: Yeah, good catch. I'm not entirely sure how we can handle this, though.

I suppose we could massage the old-style keys into new style ones (since they do contain enough information to extract the kind, namespace, and name).

If an old Felix connects to a new Typha, I suppose the correct fix would be to have a handshake field to agree on the key format that needs to be used? Old felixes wouldn't set the field, and so new Typhas would know to send the old-style keys?

felix/fv/flow_logs_goldmane_staged_test.go
caseydavenport: not sure - technically, we're no longer including the tier in this value for every policy any more so the names on average will be shorter, which should account for some of the extra characters used on the kind?

We could swap this to an abbreviation though, and probably worth doing regardless (although I do like using a consistent formatting of these ID fields where possible).

felix/calc/calc_graph_fv_test.go
caseydavenport: yep, good call. My plan was also to use validation to make the tier field immutable for the first iteration, since especially on the enterprise side I am worried about assumptions that may exist in this area, and then relaxing that validation over time.

felix/calc/policy_sorter.go
caseydavenport: Hm, we do have the tier in-hand here so we could theoretically generate the same strings to sort on... e.g., `namespace/knp.default.foo` vs `namespace/tier1.foo` vs `default.gnp1` - we could do a switch statement to generate the right "old style" name for each kind, but I do wonder if it's worth it.

100% guarantee this will break someone, but if it does that policy is already strung together with tape and glue and I'm not sure it makes sense to cater to that? Maybe we can just release note that policies with the same order may sort differently, and if you are relying on implicit ordering with deny rules you should add an order field?

app-policy/checker/check_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2025 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2025 Tigera, Inc. All rights reserved.
```
mazdakn: nit: change by mistake?

app-policy/policystore/process.go
```diff
@@ -198,7 +198,7 @@ func (store *PolicyStore) processWorkloadEndpointUpdate(subscriptionType string,
 	case "per-host-policies":
 		store.Endpoints[types.ProtoToWorkloadEndpointID(update.Id)] = update.Endpoint
 		log.Debugf("%d endpoints received so far", len(store.Endpoints))
-		//store.wepUpdates.onWorkloadEndpointUpdate(update, store.IPToIndexes)
+		// store.wepUpdates.onWorkloadEndpointUpdate(update, store.IPToIndexes)
```
mazdakn: nit: probably remove this and the one below?

app-policy/policystore/process_test.go
```diff
@@ -123,15 +133,17 @@ func TestIPSetUpdateDispatch(t *testing.T) {
 	RegisterTestingT(t)
 	id := "test_id"
 	store := NewPolicyStore()
-	//inSync := make(chan struct{})
+	// inSync := make(chan struct{})
```
mazdakn: nit: worth removing.

felix/bpf/polprog/pol_prog_builder.go
```diff
@@ -172,8 +173,10 @@ type Rule struct {
 }
 
 type Policy struct {
-	Name  string
-	Rules []Rule
+	Kind      string
+	Namespace string
```
mazdakn, line 177: What's the purpose of `Kind` and `namespace`. Seems not used anywhere.

felix/collector/flowlog/types_test.go
```diff
@@ -74,18 +75,18 @@ func setEgressTraceAndMetrics(mu metric.Update, egress, pendingEgress []*calc.Ru
...
+	egress4 := calc.NewRuleID(v3.KindNetworkPolicy, "tier4", "policy4", "namespace4", 1, rules.RuleDirEgress, rules.RuleActionAllow)
 
 	BeforeEach(func() {
 		ca = NewAggregator()
 	})
 
 	DescribeTable("splits up FlowStore into multiple FlowLogs for multiple items in the FlowPolicySets",
 		func(mus []*metric.Update, aggregation AggregationKind, expected TraceAndMetrics) {
-			//ca.AggregateOver(aggregation)
+			// ca.AggregateOver(aggregation)
```
mazdakn: nit: please remove.

felix/collector/stats_test.go
```diff
@@ -295,7 +306,7 @@ var _ = Describe("Rule Trace", func() {
 			It("should have not have action set", func() {
 				Expect(data.IngressAction()).NotTo(Equal(rules.RuleActionAllow))
 				Expect(data.IngressAction()).NotTo(Equal(rules.RuleActionDeny))
-				//Expect(data.IngressAction()).NotTo(Equal(rules.RuleActionPass))
+				// Expect(data.IngressAction()).NotTo(Equal(rules.RuleActionPass))
```
mazdakn: nit: remove pls.

felix/dataplane/linux/xdp_state_test.go
```diff
@@ -1461,7 +1462,7 @@ var _ = Describe("XDP state", func() {
 				//	eligiblePolicies: map[string][][]string{
 				//		"policy": [][]string{[]string{"ipset"}},
 				//	},
-				//	endpoints: map[string][]string{
+				//	endpoints: map[string][]*proto.PolicyID{
```
mazdakn, line 1465: Is there any value in keeping this test?

felix/proto/felixbackend.proto
```diff
@@ -236,8 +236,13 @@ message ActivePolicyRemove {
 }
 
 message PolicyID {
-  string tier = 1;
+  // The "tier" field has been deprecated and removed.
+  reserved 1;
+  reserved "tier";
```
mazdakn, line 241: Do we need to reserve these, given that we do not support version skew in Felix?
I'm working on removing duplicate `HostMetadata` messages, and this is a question that applies also there.

felix/rules/endpoints_test.go
```diff
@@ -520,42 +578,62 @@ func endpointRulesTests(flowLogsEnabled bool, dropActionOverride string) func()
...
-							EgressPolicies:  []string{"ae", "be"},
+							Name:          "default",
+							DefaultAction: "Pass",
+							IngressPolicies: []*proto.PolicyID{
+								{Name: "ai", Kind: v3.KindNetworkPolicy},
+								{Name: "bi", Kind: v3.KindNetworkPolicy},
+							},
+							EgressPolicies: []*proto.PolicyID{
+								{Name: "ae", Kind: v3.KindNetworkPolicy},
+								{Name: "be", Kind: v3.KindNetworkPolicy},
```
mazdakn: IIUC all these policies and some of the following ones should be `KindGlobalNetworkPolicy`, right?
Might not matter in these tests, but worth using the correct kind anyway.

goldmane/proto/methods_test.go
```diff
@@ -101,14 +101,10 @@ func TestInvalidStrings(t *testing.T) {
 		"0|kube-admin|kcnp.kube-admin.name|badaction|1",
 
 		// Tier fields do not match.
-		"0|tier|namespace/knp.default.name|allow|2",
-		"0|kube-admin|kcnp.foobar.policy1|allow|-1",
-		"1|tier|tier2.name|allow|0",
```
mazdakn, line 106: I think the next test should also be removed, beside these 3 ones. Since policy names are not prefixed with tier, these 4 test cases, are now invalid.

libcalico-go/lib/names/policy_test.go
```diff
@@ -13,83 +13,3 @@
 // limitations under the License.
 
 package names_test
-
```
mazdakn, line 16: Probably remove the whole file?

felix/calc/policy_sorter.go
fasaxc: The old name included tier but we're sorting within a tier here so the names would all have the same tier.  I don't think you need to be heroic, but putting the components in the same order that they used to be in should break the fewest people?

libcalico-go/lib/backend/model/keys.go
fasaxc: Yeah, the handshake behaves like JSON parsing (i.e ignoring fields that it doesn't know about) so you can add new feature flags.  Only trouble is that typha's only option if the client doesn't support something it needs is to kill the connection.

typha/fv-tests/server_test.go
```diff
@@ -1682,7 +1680,7 @@ var _ = Describe("with server requiring TLS", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			// Client sends a few valid bytes of a Client hello but then stops...
-			_, err = tcpConn.Write([]byte{16, 3, 01})
```
fasaxc, line 1685: Looks like this was unintentional octal

typha/fv-tests/server_test.go
```diff
@@ -779,9 +781,10 @@ var _ = Describe("With an in-process Server with short ping timeout", func() {
 			BeforeEach(func() {
 				err := w.Encode(syncproto.Envelope{
 					Message: syncproto.MsgClientHello{
-						Hostname: "me",
-						Version:  "test",
-						Info:     "test info",
+						Hostname:                 "me",
+						Version:                  "test",
+						Info:                     "test info",
+						SupportsModernPolicyKeys: true,
```
fasaxc, line 807: Should probably have a test that verifies that clients that don't set this are rejected.

libcalico-go/lib/backend/model/policy.go
```diff
@@ -127,3 +127,59 @@ func (p Policy) String() string {
 	}
 	return strings.Join(parts, ",")
 }
+
+// buildLegacyPolicyKey builds a legacy policy key from the given name.
+// Legacy policy key names included the namespace and name of the policy, with
+// hints of the kind that can be used to reconstruct the full key.
+//
+// Legacy policy keys will be sent by Typha versions prior to v3.32.0.
+func buildLegacyPolicyKey(name string) Key {
```
fasaxc:
```suggestion
func parseLegacyPolicyName(name string) PolicyKey {
```

libcalico-go/lib/backend/model/policy.go
```diff
@@ -127,3 +127,59 @@ func (p Policy) String() string {
...
+	// First, split based on "/" to see if we have a namespace.
+	parts := strings.SplitN(name, "/", 2)
+	if len(parts) == 2 {
+		// We have a namespace. This can be one of a few different kinds:
+		// - knp.default.<name>          -> KubernetesNetworkPolicy
+		// - staged:<name>               -> StagedNetworkPolicy
+		// - staged:knp.default.<name>   -> StagedKubernetesNetworkPolicy
+		// - <name>                      -> NetworkPolicy
+		namespace := parts[0]
```
fasaxc, line 149: Do we need to worry about openstack security groups here?  Looks like they have prefix `ossg.default.`

felix/fv/flow_logs_goldmane_staged_test.go
```diff
@@ -325,8 +325,8 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ goldmane flow log with stag
 
 		if !bpfEnabled {
 			// Wait for felix to see and program some expected nflog entries, and for the cluster IP to appear.
-			Eventually(getRuleFunc(tc.Felixes[0], "APE0|GlobalNetworkPolicy/default.ep1-1-allow-all"), "10s", "1s").ShouldNot(HaveOccurred())
-			Eventually(getRuleFunc(tc.Felixes[1], "APE0|GlobalNetworkPolicy/default.ep1-1-allow-all"), "10s", "1s").Should(HaveOccurred())
+			Eventually(getRuleFunc(tc.Felixes[0], "APE0|gnp/default.ep1-1-allow-all"), "10s", "1s").ShouldNot(HaveOccurred())
```
fasaxc, line 328: First time I've clocked what the rules look like; is it OK to bloat the prefix with the full type name? We've been pretty terse with the other prefixes, maybe that has a slight perf impact (or maybe we're just trying to leave as much room as possible for the policy name)?

Could compress it to a single char `G=GNP` `N=NP` `K=K8s NP`, ...?

fasaxc: Short names LGTM

felix/calc/policy_resolver.go
```diff
@@ -182,6 +179,19 @@ func (pr *PolicyResolver) Flush() {
 		log.Debugf("Not in sync, skipping flush")
 		return
 	}
+	// Resolve any pending policy updates, and clear the set.
+	pr.pendingPolicyUpdates.Iter(func(polKey model.PolicyKey) error {
+		policy, ok := pr.allPolicies[polKey]
+		if !ok {
+			log.Warnf("PolicyResolver missing policy metadata for %s during flush", polKey)
```
fasaxc, line 187: Might be worth toggling this to panic when running the calc graph FVs.  Might find a bug.

felix/calc/policy_sorter.go
```diff
@@ -160,18 +160,22 @@ func TierLess(i, j tierInfoKey) bool {
...
 }
 
 var polMetaDefaultOrder = math.Inf(1)
 
 func ExtractPolicyMetadata(policy *model.Policy) policyMetadata {
-	m := policyMetadata{}
+	if policy.Tier == "" {
+		// This shouldn't happen - all policies should have a tier assigned by now.
+		logrus.WithField("policy", policy).Warn("Policy has no tier assigned")
+	}
```
mazdakn, line 182: Is it safe to continue if tier is empty? Maybe we should just return a safe policyMetadata explicitly.

libcalico-go/lib/backend/etcdv3/etcdv3.go
```diff
@@ -663,14 +656,6 @@ func defaultPolicyName(d *model.KVPair) error {
 		}
```
MichalFupso, line 656: I think we can remove the whole logic around storing the Tiered name in an annotation. The crd name should be what is returned. We should be able to remove storePolicyName and defaultPolicyName all together

libcalico-go/lib/backend/k8s/resources/resources.go
```diff
@@ -189,27 +188,6 @@ func ConvertCalicoResourceToK8sResource(resIn Resource) (Resource, error) {
 	meta.Namespace = rom.GetNamespace()
```
MichalFupso, line 188: re line 140 and 149: we should be able to remove the logic storing tiered name in the annotation

libcalico-go/lib/backend/model/keys.go
```diff
@@ -528,8 +541,9 @@ func OldKeyFromDefaultPath(path string) Key {
...
 		log.Debugf("Path is a policy: %v", path)
 		return PolicyKey{
-			Tier: unescapeName(m[1]),
-			Name: unescapeName(m[2]),
+			Kind:      unescapeName(m[1]),
+			Namespace: unescapeName(m[2]),
+			Name:      unescapeName(m[3]),
 		}
 	} else if m := matchProfile.FindStringSubmatch(path); m != nil {
 		log.Debugf("Path is a profile: %v (%v)", path, m[2])
```
MichalFupso, line 548: line 667 - 700: another place where we use the annotation to determine the policy name

libcalico-go/lib/clientv3/globalnetworkpolicy.go
```diff
@@ -142,50 +141,21 @@ func (r globalNetworkPolicies) Get(ctx context.Context, name string, opts option
 	if out != nil {
```
MichalFupso, line 133: Lines 103 - 106: You removed the policyNameValidation, I think we are not validation the policy name. We could add the name validation into the validator (same for all policy kinds)

libcalico-go/lib/clientv3/globalnetworkpolicy_e2e_test.go
```diff
@@ -386,54 +378,21 @@ var _ = testutils.E2eDatastoreDescribe("GlobalNetworkPolicy tests", testutils.Da
...
-				Expect(err).NotTo(HaveOccurred())
-
-				// We should be able to get it without the default. prefix
-				_, err = c.GlobalNetworkPolicies().Get(ctx, "prefix-test-policy", options.GetOptions{})
-				Expect(err).ToNot(HaveOccurred())
-			}
-		})
-	})
-
 	DescribeTable("GlobalNetworkPolicy name validation tests",
```
MichalFupso: This test name seems to be confusing now, especially the entry descriptions. Is it worth keeping this test, or would it better to move to the validator? (same for all policy kinds)

typha/fv-tests/server_test.go
caseydavenport: indeed 😬

app-policy/checker/check_test.go
caseydavenport: yeah that's an odd one

libcalico-go/lib/backend/model/policy.go
caseydavenport: Hm, I think in the absence of any special logic they will still be handled correctly by this particular function (we just won't trim the prefix on the name when sending to Felix, which I think is actually desirable for this code since it's attempting to use the "legacy" name anyway?)

I'm not too familiar with the openstack SG integration though - maybe we need to represent those as a distinct `Kind` value moving forward?

felix/bpf/polprog/pol_prog_builder.go
caseydavenport: hm, seems like not (although it feels like they _should_ be...)

caseydavenport: I think the comments that are added should include the kind + namespace at a minimum - WDYT? 

That will match what we do for iptables/nftables as well (since you can now have two policies with the same name but different kinds)

felix/dataplane/linux/xdp_state_test.go
caseydavenport: I'm not sure, but didn't feel like this PR was the place to make that decision!

felix/proto/felixbackend.proto
caseydavenport: Strictly speaking, there are potentially other external dataplane implementations that might care about the version skew?

libcalico-go/lib/backend/model/keys.go
caseydavenport: I think I will still want the logic to unpack the policy name - I'm going to be submitting a follow on PR that updates policies in storage so that any existing policies actually use their correct name, so will need to be able to read back the value from the annotation

libcalico-go/lib/clientv3/globalnetworkpolicy.go
caseydavenport: I think we don't need any form of name validation any more, so we can probably just remove this entirely.

felix/calc/policy_resolver.go
caseydavenport: Good call, spotted a bug where we weren't removing items from the pending set when the policy was deleted :grimacing: 

Do you know if we have an existing mechanism for detecting when we're running as part of UT/FV in Felix? Or do I just need to plumb a new option through the calc graph for this?

felix/bpf/polprog/pol_prog_builder.go
```diff
@@ -550,11 +550,18 @@ func (p *Builder) writePolicyRules(policy Policy, actionLabels map[string]string
 }
 
 func (p *Builder) writePolicy(policy Policy, actionLabels map[string]string, destLeg matchLeg) {
-	p.b.AddCommentF("Start of policy %s", policy.Name)
-	log.Debugf("Start of policy %q %d", policy.Name, p.policyID)
+	// Identifyin comment at the start and end of the policy.
```
caseydavenport: whoops, typo

felix/proto/felixbackend.proto
mazdakn: Make sense. Better safe than sorry!

felix/bpf/polprog/pol_prog_builder.go
mazdakn: yes, make sense. Rule comment needs to include Kind + Namespace to differentiate policies similar to iptables/nftables.

I think updating the comments is enough. `calico-bpf` policy dump reads those comments, so you most likely need to update some FVs as well. @sridhartigera is this statement correct?

sridhartigera: Yes that's right.

libcalico-go/lib/backend/model/policy.go
caseydavenport: I've got this tracked as a separate task now - I think this is OK but the thing we might hit issues with is if users create NetworkPolicies named "ossg.default.foo" in addition to a security group named "foo"

fasaxc: I think a different kind makes sense. Worth discussing with @nelljerram I think.

nelljerram: Hm, are you thinking about a different kind to disambiguate the case that @caseydavenport just described?  Otherwise I don't see a strong case for it.  Perhaps we could also rethink how we map from security groups to policies, if there's another way of doing it that would be unambiguous.  Need to review any migration issues, but I don't think that would be too bad.

MichalFupso (approved):

felix/calc/policy_resolver.go
fasaxc: I think there's a Go way to do it, can you check the build tags for `test` or something?

libcalico-go/lib/backend/model/policy.go
fasaxc: My 2p: I think having a separate kind would be consistent; it's what we're dong for the other derived kinds.

Issue #11160, opened 2025-10-09
packet drop due to unexpected size in tcp gso with ebpf vxlan encap - nodeports


juliantaylor:
## Current Behavior
When running calico with the ebpf dataplane with vxlan in kubernetes and enabled generic receive offload there are occasional packet drops for specific sized packets that have been recombined from segments when they are segmented again for transmit.

The drops occur tcp_gso_segment function in the check that the skb size is smaller than the gso_size 
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/tcp_offload.c#n163
```
	mss = skb_shinfo(skb)->gso_size;
	if (unlikely(skb->len <= mss))
		goto out;
```

This path returns an error back to validate_xmit_skb which drops it
```
        if (netif_needs_gso(skb, features)) {
                struct sk_buff *segs;

                segs = skb_gso_segment(skb, features);
                if (IS_ERR(segs)) {
                        goto out_kfree_skb;

```

This can be tracked via following bpftrace program:

```
kprobe:tcp_gso_segment {
    $skb = (struct sk_buff*)arg0;
    $shinfo =(struct skb_shared_info*)($skb->head + $skb->end);
    if ($skb->len <= $shinfo->gso_size) {
        printf("%ld %ld %d %d\n", $skb->len, $shinfo->gso_size,
              $shinfo->gso_type, ((struct napi_gro_cb*)$skb->cb)->count);
    }   
}
```

following simple test script from outside an cluster against a http echo service /e.g. httpbin/get will timeout for several packetsizes
```
#!/bin/bash

dst=mydestination
while true; do
    for i in {1400..1600}; do
        echo $i
        for j in {1..5}; do
            hdr=$(LANG=C tr -dc A-Za-z0-9 </dev/urandom | head -c $i)
            curl -m 0.2 -sS -H "test: $hdr" $dst >/dev/null
        done
    done
done
```

the bpftrace returns:

```
skb->len skb->gso_size skb->gso_type gro->count
1211 1228 1027 2
1212 1228 1027 2
1215 1228 1027 2
1216 1228 1027 2
1217 1228 1027 2
1218 1228 1027 2
1226 1228 1027 2
1227 1228 1027 2
1228 1228 1027 2
```
the difference which is dropped appears to be 17 bytes (1228-1211), possibly the 18 byte ethernet header?
(the mtu in the cluster interface is 1500 and the clients 1280 for this output)

[...]

PR #11252, opened 2025-10-23
Set BPF_F_ADJ_ROOM_FIXED_GSO during vxlan decap correctly

juliantaylor:
The flag has been added to the mode flag where it is a noop, the GSO size needs to be fixed as otherwise with receive offload/gro enabled and the packets are close to the mtu the gso size will be increased and then dropped by the tcp_gso_segment check skb->len <= gso_size.
The packet is then retransmitted which causes delays.

This can be reproduced with following bpftraces and traffic with appropriately sized packets but the exact conditions are not well understood yet.

```
kprobe:tcp_gso_segment {
    $skb = (struct sk_buff*)arg0;
    $shinfo =(struct skb_shared_info*)($skb->head + $skb->end);
    if ($skb->len <= $shinfo->gso_size) {
        printf("DROP skb->len %ld gso_size %ld %d %d\n", $skb->len, $shinfo->gso_size,
              $shinfo->gso_type, ((struct napi_gro_cb*)$skb->cb)->count);
    }
}
kprobe:bpf_skb_adjust_room {
        $skb = (struct sk_buff*)arg0;
        if (((struct napi_gro_cb*)$skb->cb)->count > 1) {
                $shinfo =(struct skb_shared_info*)($skb->head + $skb->end);
                printf("adj skb->len %ld gso_size %ld len_diff %ld mode %u flags %lu\n", $skb->len, $shinfo->gso_size, (int32) arg1, (uint32) arg2, (uint32) arg3);
        }
}

> adj  skb->len 1466 gso_size 1348 len_diff -50 mode 1 flags 0
> DROP skb->len 1382 gso_size 1398 1027 2

```

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/11160

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: Do not adjust gso_size after nodeport tunnel vxlan decap. There is no guarantee that there would be enough data after removing tunnel headers. The packet is shrunk by 50 bytes while the gso_size would grow. Kernel would drop the packet if the original gso packet is too small.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/nat4.h
+++ b/felix/bpf-gpl/nat4.h
@@ -95,7 +95,7 @@ static CALI_BPF_INLINE int vxlan_decap(struct __sk_buff *skb)
 	extra_hdrsz = sizeof(struct ethhdr) + sizeof(struct iphdr) +
 		sizeof(struct udphdr) + sizeof(struct vxlanhdr);
 
-	ret = bpf_skb_adjust_room(skb, -extra_hdrsz, BPF_ADJ_ROOM_MAC | BPF_F_ADJ_ROOM_FIXED_GSO, 0);
+	ret = bpf_skb_adjust_room(skb, -extra_hdrsz, BPF_ADJ_ROOM_MAC, BPF_F_ADJ_ROOM_FIXED_GSO);
 
 	return ret;
 }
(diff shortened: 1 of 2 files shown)


tomastigera:
failed only in an unrelated test with a known issue

PR #11276, opened 2025-10-28
[EV-6037] Skip policy trace evaluation for deleted eps

dimitri-nicolo:
## Description

**Summary**
This PR implements functionality to skip RuleHits processing in flow logs when endpoints are marked for deletion, preventing unnecessary policy trace evaluation for endpoints that are being cleaned up. As a consequence, the change also ensures that the flow log action is not updated when one of the endpoints is deleted.

**Problem**
A race condition occurs when a pod is deleted while processing packet information for traffic with a pre-DNAT destination IP directed at a service. This leads to a 'deny' action for incoming packets, even when an existing policy would permit the traffic.

**Solution**
Added logic to detect when source or destination endpoints are marked for deletion and skip policy trace processing.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fixed a race in flow log generation that could mis-report service traffic as denied when a backing pod was deleted while the packet was being processed.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/calc/endpoint_lookup_cache.go
+++ b/felix/calc/endpoint_lookup_cache.go
@@ -98,8 +98,6 @@ type EndpointData interface {
 type endpointData interface {
 	EndpointData
 	allIPs() [][16]byte
-	isMarkedToBeDeleted() bool
-	setMarkedToBeDeleted(b bool)
 }
 
 type MatchData struct {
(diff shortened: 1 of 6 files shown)


sridhartigera (approved):
LGTM

PR #11279, opened 2025-10-28
Redirect to workload from l3

sridhartigera:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Deprecated the "L2Only" value for BPFRedirectToPeer. BPFRedirectToPeer="Enabled" now explicitly allows direct redirection to the peer side from both L2 and L3 host devices (e.g. IPIP/WireGuard). 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/list-objs
+++ b/felix/bpf-gpl/list-objs
@@ -36,7 +36,7 @@ for log_level in debug no_log; do
       if [ "${host_drop}" = "host_drop_" ]; then
         # The workload-to-host drop setting only applies to the from-workload hook.
         ep_types="wep"
-        directons="from"
+        directions="from"
       else
         ep_types="wep hep ipip l3 nat lo vxlan"
       fi


felix/bpf-gpl/fib_co_re.h
```diff
@@ -9,22 +9,48 @@
 
 #include <linux/if_packet.h>
 
+static CALI_BPF_INLINE int make_room_for_l2_header(struct cali_tc_ctx *ctx)
+{
+	if (!CALI_F_L3_DEV) {
+		return 0;
+	}
```
tomastigera: I think I would take this out and put it around the call to `make_room_for_l2_header` as it would be a bit more obvious when it is needed.

Or, you can replace all the bpf_redirect_* with a wrapper that has this inside it. Just call it `pkt/skb_redirect_` or just `redirect_` or something like that and make it decide what it needs to do. Then you do not need to change the flow in the main function that much.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -897,8 +897,8 @@ type FelixConfigurationSpec struct {
 	// it on ingress. Value of Enabled also allows redirection from L3 host devices like
 	// IPIP tunnel or Wireguard directly to the peer side of the workload's device. This
 	// makes redirection faster, however, it breaks tools like tcpdump on the peer side.
-	// Use Enabled with caution. [Default: L2Only]
-	//+kubebuilder:validation:Enum=Enabled;Disabled;L2Only
+	// Use Enabled with caution. [Default: Enabled]
```
tomastigera: If we default to enabled, than it is weird to say to use Enabled with caution :thinking:  Does it make sense to deprecate L2Only  and treat it like Enabled?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -891,14 +891,12 @@ type FelixConfigurationSpec struct {
...
-	// BPFRedirectToPeer controls which whether it is allowed to forward straight to the
-	// peer side of the workload devices. It is allowed for any host L2 devices by default
-	// (L2Only), but it breaks TCP dump on the host side of workload device as it bypasses
-	// it on ingress. Value of Enabled also allows redirection from L3 host devices like
-	// IPIP tunnel or Wireguard directly to the peer side of the workload's device. This
-	// makes redirection faster, however, it breaks tools like tcpdump on the peer side.
-	// Use Enabled with caution. [Default: L2Only]
-	//+kubebuilder:validation:Enum=Enabled;Disabled;L2Only
+	// BPFRedirectToPeer controls whether traffic may be forwarded directly to the peer side of a workload’s device.
+	// Note that the legacy "L2Only" option is now deprecated — it used to permit direct redirection only for host L2 devices.
```
```suggestion
	// Note that the legacy "L2Only" option is now deprecated and if set it is treated like "Enabled"
```

felix/daemon/daemon.go
```diff
@@ -380,6 +380,13 @@ configRetry:
 			if err != nil {
 				log.WithError(err).Panic("Bug: failed to override config parameter BPFConntrackCleanupMode")
 			}
+			if configParams.BPFRedirectToPeer == "L2Only" {
+				log.Warn("BPFRedirectToPeer 'L2Only' is deprecated. Falling back to 'Enabled'.")
```
tomastigera:
```suggestion
				log.Warn("BPFRedirectToPeer 'L2Only' is deprecated and equals 'Enabled' now.")
```

PR #11284, opened 2025-10-28
Remove duplicate host metadata updates

mazdakn:
## Description

Remove redundant `HostMetadata*` messages. Atm, there are 3 variants of this message, named `HostMetadataUpdate`, `HostMetadataV6Update` and `HostMetadataV4V6Update`. This PR removes the first two, and uses the last one in all places.

There will be follow up changes:
- Rename `HostMetadatV4V6*` messages to just `HostMetadata*` (basically removing V4V6 part) for sake of simplicity.
- Moving Host and Node updates to a separate component out of data plane passthru component in calc graph.  

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Deprecating HostMetadataUpdate, HostMetadataRemove, HostMetadataV6Update, and HostMetadataV6Remove internal protobuf messages in favor of HostMetadataV4V6Update and HostMetadataV4V6 messages.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/calc/dataplane_passthru.go
+++ b/felix/calc/dataplane_passthru.go
@@ -15,8 +15,6 @@
 package calc
 
 import (
-	"maps"
-
 	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	"github.com/sirupsen/logrus"
 	kapiv1 "k8s.io/api/core/v1"
@@ -121,6 +119,7 @@ func (h *DataplanePassthru) OnUpdate(update api.Update) (filterOut bool) {
 func (h *DataplanePassthru) processModelHostIP(key model.HostIPKey, update api.Update) {
 	hostname := key.Hostname
 	before := h.merge(hostname)
+
 	if update.Value == nil {
 		logrus.WithField("update", update).Debug("Passing-through HostIP deletion")
 		delete(h.hostIPv4, hostname)
@@ -163,14 +162,14 @@ func (h *DataplanePassthru) processKindNode(key model.ResourceKey, update api.Up
 		// CRD validation (`cidrv4`/`cidrv6` tags). Bare IPs are dropped silently
 		// here so the dataplane never sees an ambiguous address.
 		if addr := bgpSpec.IPv4Address; addr != "" {
-			if s, ok := parseBGPCIDR(addr); ok {
+			if s, ok := parseCIDR(addr); ok {
 				info.ip4Addr = s
 			} else {
 				logrus.WithField("addr", addr).Warn("Ignoring Node BGP IPv4Address: not a CIDR")
 			}
 		}
 		if addr := bgpSpec.IPv6Address; addr != "" {
-			if s, ok := parseBGPCIDR(addr); ok {
+			if s, ok := parseCIDR(addr); ok {
 				info.ip6Addr = s
 			} else {
 				logrus.WithField("addr", addr).Warn("Ignoring Node BGP IPv6Address: not a CIDR")
@@ -179,7 +178,8 @@ func (h *DataplanePassthru) processKindNode(key model.ResourceKey, update api.Up
 		if node.Spec.BGP.ASNumber != nil {
 			info.asnumber = node.Spec.BGP.ASNumber.String()
 		}
-	} else { // node.Spec.BGP is nil
+	} else {
+		// node.Spec.BGP is nil
 		if h.ipv6Support {
 			// BGP is turned off; fall back to the Node's address list for IPv6.
 			// Mirrors how HostIPKey is generated for IPv4 in
@@ -206,13 +206,15 @@ func (h *DataplanePassthru) processKindNode(key model.ResourceKey, update api.Up
 // states.
 func (h *DataplanePassthru) emitTransition(hostname string, before *HostInfo) {
 	after := h.merge(hostname)
+
 	if after == nil {
 		if before != nil {
 			h.callbacks.OnHostMetadataRemove(hostname)
 		}
 		return
 	}
-	if before != nil && hostInfoEqual(before, after) {
+
+	if before != nil && before.equals(after) {
 		return
 	}
 	h.callbacks.OnHostMetadataUpdate(hostname, after)
@@ -223,14 +225,14 @@ func (h *DataplanePassthru) emitTransition(hostname string, before *HostInfo) {
 // are present; the HostIPKey-derived /32 only fills in IPv4 when there is no
 // Node-derived value for it.
 func (h *DataplanePassthru) merge(hostname string) *HostInfo {
-	node, hasNode := h.nodeInfo[hostname]
+	node := h.nodeInfo[hostname]
 	hostIP, hasHostIP := h.hostIPv4[hostname]
 
-	if !hasNode && !hasHostIP {
+	if node == nil && !hasHostIP {
 		return nil
 	}
 
-	if hasNode {
+	if node != nil {
 		merged := *node
 		if merged.ip4Addr == "" && hasHostIP {
 			merged.ip4Addr = hostIP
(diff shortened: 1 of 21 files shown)


felix/dataplane/linux/wireguard_mgr.go
```diff
@@ -63,34 +63,20 @@ func (m *wireguardManager) OnUpdate(protoBufMsg any) {
 	logCtx := log.WithField("ipVersion", m.ipVersion)
 	logCtx.WithField("msg", protoBufMsg).Debug("Received message")
 	switch msg := protoBufMsg.(type) {
-	case *proto.HostMetadataUpdate:
-		logCtx.WithField("msg", msg).Debug("HostMetadataUpdate update")
+	case *proto.HostMetadataV4V6Update:
+		logCtx.WithField("msg", msg).Debug("HostMetadataV4V6Update update")
 		if m.ipVersion != 4 {
```
mazdakn: need to adjust it based on ip family.

felix/proto/felixbackend.proto
```diff
@@ -84,16 +84,10 @@ message ToDataplane {
...
-    HostMetadataUpdate host_metadata_update = 14;
-    // HostIPRemove is sent when a host IP is removed.
-    HostMetadataRemove host_metadata_remove = 18;
-
     // HostMetadataV4V6Update is sent when a host is added or updated.
-    HostMetadataV4V6Update host_metadata_v4v6_update = 37;
+    HostMetadataV4V6Update host_metadata_v4v6_update = 14;
     // HostIPRemove is sent when a host is removed.
-    HostMetadataV4V6Remove host_metadata_v4v6_remove = 38;
+    HostMetadataV4V6Remove host_metadata_v4v6_remove = 18;
```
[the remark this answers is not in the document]
mazdakn: I think eventually we should use `HostMetadataUpdate` and `HostMetadataRemove` message names. I left the renaming for the next PR, for sake of easier review of this PR.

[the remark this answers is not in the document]
mazdakn: hmm well, I think using `HostMetadatUpdate` and `HostMetadataRemove`, makes this PR  even cleaner, so I'm going to include it here.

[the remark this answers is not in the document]
mazdakn: This generated more code changes, so I'll cover it in next PR.

felix/calc/dataplane_passthru.go
```diff
@@ -58,137 +73,190 @@ func (h *DataplanePassthru) RegisterWith(dispatcher *dispatcher.Dispatcher) {
...
+			if s, ok := parseBGPCIDR(addr); ok {
+				info.ip6Addr = s
+			} else {
+				logrus.WithField("addr", addr).Warn("Ignoring Node BGP IPv6Address: not a CIDR")
+			}
+		}
+		if node.Spec.BGP.ASNumber != nil {
+			info.asnumber = node.Spec.BGP.ASNumber.String()
+		}
+	} else { // node.Spec.BGP is nil
```
caseydavenport:
```suggestion
	} else { 
	  // node.Spec.BGP is nil
```

felix/calc/dataplane_passthru.go
```diff
@@ -58,137 +73,190 @@ func (h *DataplanePassthru) RegisterWith(dispatcher *dispatcher.Dispatcher) {
...
+		merged := *node
+		if merged.ip4Addr == "" && hasHostIP {
+			merged.ip4Addr = hostIP
+		}
+		return &merged
+	}
+	return &HostInfo{ip4Addr: hostIP}
+}
+
+func hostInfoEqual(a, b *HostInfo) bool {
```
caseydavenport: very much a nit, but I might have done a bound function on HostInfo?

```go
info.Equals(other)
```

felix/calc/dataplane_passthru.go
```diff
@@ -58,137 +73,190 @@ func (h *DataplanePassthru) RegisterWith(dispatcher *dispatcher.Dispatcher) {
...
+// Node-derived value for it.
+func (h *DataplanePassthru) merge(hostname string) *HostInfo {
+	node, hasNode := h.nodeInfo[hostname]
+	hostIP, hasHostIP := h.hostIPv4[hostname]
+
+	if !hasNode && !hasHostIP {
+		return nil
+	}
+
+	if hasNode {
```
caseydavenport: Also a nit, but I think in general it's a little bit safer to do this?

```go
if node != nil {
  merged := *node
```

Otherwise we're susceptible to subtle bugs where these values get out of sync somehow (clearly not possible in the current code, so feel free to ignore).

felix/calc/dataplane_passthru.go
```diff
@@ -58,137 +73,190 @@ func (h *DataplanePassthru) RegisterWith(dispatcher *dispatcher.Dispatcher) {
...
+		merged := *node
+		if merged.ip4Addr == "" && hasHostIP {
```
caseydavenport, line 237: Curious, I wonder how / when we hit this branch. Seems harmless to include it, but I wonder if it's real?

felix/calc/dataplane_passthru.go
```diff
@@ -58,137 +73,190 @@ func (h *DataplanePassthru) RegisterWith(dispatcher *dispatcher.Dispatcher) {
...
+	return a.ip4Addr == b.ip4Addr &&
+		a.ip6Addr == b.ip6Addr &&
+		a.asnumber == b.asnumber &&
+		maps.Equal(a.labels, b.labels)
+}
+
+// parseBGPCIDR strictly parses a CIDR string and returns it normalized to "host
+// IP / mask" form (e.g. "192.168.0.2/24" rather than "192.168.0.0/24"). Returns
+// false if the string is not a valid CIDR — bare IPs are rejected.
+func parseBGPCIDR(s string) (string, bool) {
```
caseydavenport:
```suggestion
func parseCIDR(s string) (string, bool) {
```

Don't think anything about this is BGP aware?

felix/dataplane/linux/noencap_mgr.go
```diff
@@ -91,24 +91,28 @@ func newNoEncapManagerWithSims(
...
+		if addrStr == "" {
+			m.logCtx.WithFields(logrus.Fields{
+				"hostname":  msg.Hostname,
+				"ipVersion": m.ipVersion,
+			}).Debug("Ignoring HostMetadataV4V6Update with no address for this IP version")
+			return
 		}
-	case *proto.HostMetadataV6Remove:
+		m.routesNeedUpdate(addrStr)
+	case *proto.HostMetadataV4V6Remove:
```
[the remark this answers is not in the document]
caseydavenport: It might be worth thinking about this one @mazdakn ? I think there are niche cases where our CIDR detection can be wrong (or manually configured), which might trigger this - unless I'm misunderstanding the scenario which could also be true!

felix/proto/felixbackend.proto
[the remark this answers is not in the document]
caseydavenport: I suppose this is a breaking change to the Felix dataplane API regardless, since we'll no longer emit the old message types - we'll need to release note this and alert the VPP team in case this affects them. I think we should leave the `host_metadata_v4v6_update` at it's current numbering though - is there a reason not to do that? 

CC @sknat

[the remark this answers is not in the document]
mazdakn: @caseydavenport I'm just thinking having `v4v6` in the `host_metadata_v4v6_*` messages is unnecessary, and we should try to use `host_metadata_*` names instead. I'm thinking of removing `v4v6` in the next PR, so kept the message number of original `host_metadata_*` messages. 

I believe `host_metadata_v4v6_*` were added by VPP team (@sknat) to get both IPv4 and IPv6 address in one message.

felix/dataplane/linux/noencap_mgr.go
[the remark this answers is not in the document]
mazdakn: hmm looks sensible. The calc graph part is updated to send CIDR rather just an address. So need to remove it before updating the parent interface address.

felix/calc/dataplane_passthru.go
mazdakn: Fixed.

felix/calc/dataplane_passthru.go
mazdakn: Done.

felix/calc/dataplane_passthru.go
mazdakn: Done.

felix/calc/dataplane_passthru.go
mazdakn: Nope. Fixed.

felix/dataplane/linux/noencap_mgr.go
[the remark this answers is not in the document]
mazdakn: Addressed in the new commit.

felix/calc/dataplane_passthru.go
mazdakn: No exactly sure. Maybe there is no real scenario, but from just pure coding standpoint, worth keeping it.

But now wondering if there is no real case for this, is there any value for still processing or keeping `model.HostIPKey`?

caseydavenport: I believe we originally kept it because older Typhas wouldn't send the Node object so we needed to keep it for version skew. 

I suspect we could get rid of it now!

mazdakn: interesting. I'll try to remove it in a separate PR.

caseydavenport (approved):

PR #11293, opened 2025-10-29
Run e2e tests with felix programming cluster routes

mazdakn:
## Description

Run e2e tests with Felix programming cluster routes.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


metadata.mk
```diff
@@ -67,7 +67,7 @@ LIBBPF_VERSION=v1.4.6
 BPFTOOL_IMAGE=calico/bpftool:v7.5.0
 
 # The operator branch corresponding to this branch.
-OPERATOR_BRANCH       ?= master
+OPERATOR_BRANCH       ?= felix-routing
```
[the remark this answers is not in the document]
mazdakn: This branch is used to remove a validation of installation resource: https://github.com/tigera/operator/pull/4243

node/tests/k8st/deploy_resources_on_kind_cluster.sh
```diff
@@ -69,8 +69,14 @@ echo "Install additional permissions for BGP password"
 ${kubectl} apply -f $TEST_DIR/infra/additional-rbac.yaml
 echo
 
-echo "Install Calico using the helm chart"
-$HELM install calico $CHART -f $TEST_DIR/infra/values.yaml -n tigera-operator --create-namespace
+BGP_CONFIG=${BGP_CONFIG:-"Enabled"}
+if [[ "$BGP_CONFIG" == "Disabled" ]]; then
```
caseydavenport: I think a `VALUES_FILE` env var would be a bit nicer here. I'm imagining we might want additional alternative configurations in the future and that makes it more flexible. 

It would be extra nice if we could somehow configure the felixconfiguation patch via the operator, but we can leave that for a follow-on later.

caseydavenport: e.g.,

```
$HELM install calico $CHART -f $TEST_DIR/infra/$VALUES_FILE`
```

caseydavenport (approved):
I think the copilot suggestions seem correct, but otherwise this LGTM. I'm assuming we don't have a way to configure this in the operator API, do we need / want one? 

Another thought is that we create a default FelixConfiguration from within the product, but we don't for BGPConfiguration... should we?

PR #11305, opened 2025-11-03
Fix maglev e2e test

song-jiang:
## Description

Maglev e2e tests are now passing against the latest master hash release.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/e2e/pkg/tests/networking/maglev.go
+++ b/e2e/pkg/tests/networking/maglev.go
@@ -77,7 +77,7 @@ var _ = describe.CalicoDescribe(
 			}
 
 			nodesInfo := utils.GetNodesInfo(f, nodes, false)
-			nodeNames := nodesInfo.GetNames()
+			nodeNames = nodesInfo.GetNames()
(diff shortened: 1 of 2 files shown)


e2e/pkg/tests/networking/maglev.go
```diff
@@ -77,7 +77,7 @@ var _ = describe.CalicoDescribe(
 			}
 
 			nodesInfo := utils.GetNodesInfo(f, nodes, false)
-			nodeNames := nodesInfo.GetNames()
+			nodeNames = nodesInfo.GetNames()
```
aaaaaaaalex, line 80: Assume you're assigning to a more-global var here, is that var still scoped to just this file? (Just curious of how widespread the change will be)

song-jiang: Yes. There's a global var declared but it was masked by this line here.

PR #11308, opened 2025-11-04
[eBPF] - Fix host->host wireguard traffic

sridhartigera:
## Description

For vxlan and ipip, host->host traffic does not go through the tunnel interface. Whereas for wireguard, we do support host to host encryption which requires traffic to go through the wireguard interface. We had
```        } else if (CALI_F_TUNNEL && CALI_F_TO_HEP) {
                if (!(ctx->skb->mark & CALI_SKB_MARK_SEEN) ||
                        !skb_mark_equals(ctx->skb, CALI_SKB_MARK_TUNNEL_KEY_SET, CALI_SKB_MARK_TUNNEL_KEY_SET)) {
                        /* packet to vxlan from the host, needs to set tunnel key. Either
                         * it wasn't seen or it was routed via the bpfnat device because
                         * its destination was a service and CTLB is disabled
                         */
                        struct cali_rt *dest_rt = cali_rt_lookup(&ctx->state->ip_dst);
                        if (dest_rt == NULL) {
                                CALI_DEBUG("No route for " IP_FMT " at tunnel device", &ctx->state->ip_dst);
                                goto deny;
                        }
                        if (!cali_rt_is_tunneled(dest_rt)) {
                                CALI_DEBUG("Not a tunnel route for " IP_FMT " at tunnel device", &ctx->state->ip_dst);
                                goto deny;
                        }
```
This will result in host to host traffic getting dropped at the wireguard interface. This PR skips fib in case the dst route is not tunnelled.  This PR also fixes allowing known wireguard traffic in HEPs.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/fib_co_re.h
+++ b/felix/bpf-gpl/fib_co_re.h
@@ -235,16 +235,9 @@ skip_redir_ifindex:
 			}
 			if (!cali_rt_is_tunneled(dest_rt)) {
 				CALI_DEBUG("Not a tunnel route for " IP_FMT " at tunnel device", &ctx->state->ip_dst);
-#if CALI_F_L3_DEV
-				// In the case of wireguard, we support host->host encryption.
-				// Remote host route will not be tunneled and we should not be
-				// denying such packets. Hence let them pass to IP stack.
-				// CALI_F_L3_DEV is set for ipip as well but we should not
-				// have host->host ipip traffic hitting the tunnel device.
+				// We dont have a tunnel route, so nothing to do here.
+				// Better to leave it to the tunnel device to handle it.
 				goto skip_fib;
-#else
-				goto deny;
-#endif
 			}
 
 			struct bpf_tunnel_key key = {


felix/bpf-gpl/fib_co_re.h
```diff
@@ -235,7 +235,16 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
 			}
 			if (!cali_rt_is_tunneled(dest_rt)) {
 				CALI_DEBUG("Not a tunnel route for " IP_FMT " at tunnel device", &ctx->state->ip_dst);
+#if CALI_F_L3_DEV
+				// In the case of wireguard, we support host->host encryption.
+				// Remote host route will not be tunneled and we should not be
+				// denying such packets. Hence let them pass to IP stack.
```
tomastigera: This is towards HEP `if (CALI_F_TUNNEL && CALI_F_TO_HEP) {` so the packet won't go to IP stack, it is just processed by the device.

felix/bpf-gpl/fib_co_re.h
```diff
@@ -235,7 +235,16 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
 			}
 			if (!cali_rt_is_tunneled(dest_rt)) {
 				CALI_DEBUG("Not a tunnel route for " IP_FMT " at tunnel device", &ctx->state->ip_dst);
```
tomastigera, line 237: Should we perhaps do the skip for everything? It got routed here. vxlan would drop it, ipip perhaps as well, wg is fine. Perhaps this was just a bit to much of a precaution. I'd keep the log though.

felix/fv/bpf_test.go
```diff
@@ -409,15 +409,6 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 			options.ExtraEnvVars["FELIX_BPFExtToServiceConnmark"] = "0x80"
 			options.ExtraEnvVars["FELIX_HEALTHENABLED"] = "true"
 			options.ExtraEnvVars["FELIX_BPFDSROptoutCIDRs"] = "245.245.0.0/16,beaf::dead/64"
-			if testOpts.tunnel == "wireguard" {
-				options.ExtraEnvVars["FELIX_BPFREDIRECTTOPEERFROML3DEVICE"] = "true"
```
tomastigera, line 413: should this option get deprecated?

sridhartigera: We don't have such a config.

felix/bpf-gpl/fib_co_re.h
sridhartigera: Host to workload will still take this path and tunnel key needs to be set right?

sridhartigera: you mean skip fib for all tunnels?

felix/bpf-gpl/tc.c
```diff
@@ -256,7 +256,8 @@ static CALI_BPF_INLINE int pre_policy_processing(struct cali_tc_ctx *ctx)
 	case PARSING_ERROR:
 		goto deny;
 	case PARSING_ALLOW_WITHOUT_ENFORCING_POLICY:
-		if (CALI_F_FROM_HEP && ctx->state->ip_proto == IPPROTO_IPIP) {
+		if (CALI_F_FROM_HEP && (ctx->state->ip_proto == IPPROTO_IPIP ||
+					ctx->state->dport == WG_PORT)) {
```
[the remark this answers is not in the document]
sridhartigera: I already return `PARSING_ALLOW_WITHOUT_ENFORCING_POLICY` only if UDP and WG_PORT. Hence I feel it is enough if I check for WG_PORT.

felix/fv/bpf_test.go
tomastigera: huh ok!

tomastigera (approved):

PR #11322, opened 2025-11-04
chore: Clarify when the create permission on custom resources is nece…

rene-dekker:
…ssary

### DIFF
--- a/charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
+++ b/charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
@@ -6,8 +6,10 @@ metadata:
   labels:
     {{- include "tigera-operator.labels" (dict "context" .) | nindent 4 }}
 rules:
-  # The tigera/operator installs CustomResourceDefinitions necessary for itself
+  # The tigera/operator needs the following permissions when configured with the --bootstrapCRDs=true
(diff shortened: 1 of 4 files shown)


charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
```diff
@@ -16,6 +16,8 @@ rules:
       - get
       - list
       - watch
+      # The create permission can be omitted if you are not using the --bootstrapCRDs=true
```
caseydavenport: I suppose this is fine enough, although I doubt many users are going to know what to do with that.

Feel like this might be a more clear but still minimalist wording: 

> The tigera/operator needs 'create' permissions when configured with the `--bootstrapCRDs=true` 
> argument. You can remove this permission if your operator deployment does not use this option.

WDYT?

rene-dekker: It is better.

It looks actually like the entire permission block was added when the feature flag was introduced. I believe that it also means that the whole block can be marked with this comment. Let me put that up shortly. This is the commit that introduced it: 
https://github.com/projectcalico/calico/commit/6617031c7e4bba4b1c3aa8d423f4aa6ce4588ec0

caseydavenport (approved):

PR #11330, opened 2025-11-05
Replace Windows 1809 with LTSC2019 as the supported build base

hjiawei:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico Windows container images have been updated to use Nano Server LTSC 2019 instead of 1809.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/metadata.mk
+++ b/metadata.mk
@@ -53,7 +53,7 @@ WINDOWS_DIST = dist/windows
 # The Windows HPC container version used as base for Calico Windows images
 WINDOWS_HPC_VERSION ?= v1.0.0
 # The Windows versions used as base for Calico Windows images
-WINDOWS_VERSIONS ?= 1809 ltsc2022
+WINDOWS_VERSIONS ?= ltsc2019 ltsc2022
 
 # The CNI plugin and flannel code that will be cloned and rebuilt with this repo's go-build image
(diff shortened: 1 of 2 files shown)


cni-plugin/Dockerfile-windows
```diff
@@ -12,11 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# FIXME: Use WINDOWS_HPC_VERSION and image instead of nanoserver and WINDOWS_VERSIONS when containerd v1.6 is EOL'd
```
hjiawei, line 15: @coutinhop Is this WINDOWS_HPC_VERSION comment still relevant?

coutinhop: yes, there's some changes that would simplify windows image builds significantly, but they require containerd 1.7+ if I recall correctly

coutinhop: but I mean, we could remove the comment, ended up not using it because we need to support containerd 1.6

hjiawei: Good to know. I have reverted these changes.

Issue #11051, opened 2025-09-19
Calico-Node ipset version vs Debian 13 (6.12.43-1)


mfigueroa7:
I have the following scenario:

Nodes: 

Linux k3s-worker 6.12.43+deb13-amd64 
Linux k3s-worker2 6.12.43+deb13-amd64 
Linux k3s-master  6.12.43+deb13-amd64 

Calico version: 3.30

Inside calico pods: ipset v7.11, protocol version: 7
Nodes: ipset v7.22, protocol version: 7

## Expected Behavior

Calico-node pods running with ipset

## Current Behavior

2025-09-19 06:44:15.163 [ERROR][2796843] felix/ipsets.go 671: Bad return code from 'ipset list cali40this-host'. error=exit status 1 family="inet" stderr="ipset v7.11: Kernel and userspace incompatible: settype hash:ip with revision 6 not supported by userspace.\n"

## Possible Solution

Upload ipset version on calico-node pods to 7.22

## Steps to Reproduce (for bugs)

1. Install new kernel from debian 13 (6.12.43) on nodes
2. Install latest version from calico (v3.30.3)

## Context

I have updated from debian12 to debian13.

## Your Environment

* Calico version: 3.30.3
* Calico dataplane (bpf, nftables, iptables, windows etc.): iptables
* Orchestrator version (e.g. kubernetes, openshift, etc.): k3s
* Operating System and version: Debian13 - 6.12.43
* Link to your project (optional):

mazdakn:
@mfigueroa7 this should not be an issue since the ipset binary bundled in Calico node image is used for any operation (creation, update, ...) performed. Now I wonder how your cluster ended up in this situation.

Is it possible that another component beside Calico has touched the mentioned ipset?

frbimo:
@mazdakn i have identical problem like above using k3s in a single node. Ive tried version v3.30.1, v3.30.3, v3.29.6 but all the same. Not sure if i can answer your question.
```
025-09-24 09:05:03.551 [INFO][14917] felix/wireguard.go 1763: Trying to connect to linkClient ipVersion=0x4
2025-09-24 09:05:03.552 [INFO][14917] felix/wireguard.go 648: Public key out of sync or updated ipVersion=0x4 ourPublicKey=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
[...]

PR #11340, opened 2025-11-07
IPSets: re-create ipsets failed during re-sync

mazdakn:
## Description

Re-create (by swapping) an ipset that was failed during re-syncing with dataplane. It includes:
- Introduce a new field to `dataplaneMetadata` called `ListFailed` to store if listing an ipsets is successful or not.
- If listing an ipset fails, set `ListFailure` to `true`.
- If listing an ipset fails several times, try to re-create that ipset by swapping it out with one created from desired state.

## Related issues/PRs
https://github.com/projectcalico/calico/issues/11051

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Re-create and swap out Calico ipsets that are not possible to list due to different failures like user-space/kernel incompatibility.  
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/ipsets/ipsets.go
+++ b/felix/ipsets/ipsets.go
@@ -478,13 +478,11 @@ func (s *IPSets) tryResync() (err error) {
 				failedIPSets = append(failedIPSets, name)
 			}
 			if desired {
-				s.logCxt.WithError(err).WithFields(log.Fields{
-					"name":    name,
-				}).Warn("Failed to parse required Calico-owned ipset that is needed, will try recreating it.")
+				s.logCxt.WithError(err).WithField("name", name).
+					Warn("Failed to parse required Calico-owned ipset that is needed, will try recreating it.")
 			} else {
-				s.logCxt.WithError(err).WithFields(log.Fields{
-					"name":    name,
-				}).Warn("Failed to parse Calico-owned ipset that is no longer needed, will queue it for deletion.")
+				s.logCxt.WithError(err).WithField("name", name).
+					Warn("Failed to parse Calico-owned ipset that is no longer needed, will queue it for deletion.")
 			}
 		} else {
 			// Successful resync of this IP set, clear any pending partial resync.


felix/ipsets/ipsets.go
```diff
@@ -681,6 +692,9 @@ func (s *IPSets) resyncIPSet(ipSetName string) error {
 		return scanner.Err()
 	})
 	if err != nil {
+		s.setNameToProgrammedMetadata.Dataplane().Set(ipSetName, dataplaneMetadata{
+			ListFailed: true,
+		})
```
mazdakn: It's possible to use an empty meta struct here. But thought to add a field to be more explicit.

felix/ipsets/ipsets.go
```diff
@@ -551,8 +561,9 @@ func (s *IPSets) resyncIPSet(ipSetName string) error {
 	//
 	// As we stream through the data, we extract the name of the IP set and its members. We
 	// use the IP set's metadata to convert each member to its canonical form for comparison.
+	meta := dataplaneMetadata{}
```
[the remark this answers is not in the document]
fasaxc: Think Copilot is right here, I don't see where you parse that so you'll always go through the swap path?

felix/ipsets/ipsets.go
```diff
@@ -365,8 +366,16 @@ func (s *IPSets) ApplyUpdates(listener UpdateListener) {
...
 				s.logCxt.WithError(err).Warning("Failed to resync with dataplane")
-				backOff()
-				continue
+
+				// After a few attempts, most likely, we are dealing with a persistent failure.
+				// This could be due to different failures, like userspace and kernel incompatibility.
+				// The incompatibility failure can be fixed by swapping the one Felix understand (created from
+				// desired state) and the one (with higher revision) in dataplane. As such, we should stop re-trying
+				// to re-sync, and instead continue with the next steps.
+				if attempt < MaxRetryAttempt/2 {
```
fasaxc: This is only doing backoff/continue if we're on an early attempt; was that intentional?  (Seems unrelated to the main change, which should just be to parse the revision out.)

felix/ipsets/ipsets_test.go
```diff
@@ -464,6 +466,42 @@ var _ = Describe("IP sets dataplane", func() {
 		})
 	})
 
+	It("Calico IP sets with unsupported revision and in desired sets should be re-created", func() {
```
fasaxc, line 467: Please add the counterpoint test: if everything is correct then it should not do a swap.  Doing a swap will always result in the right answer but it's a lot slower than re-using what's there; it invalidates a major optimisation.

fasaxc:
An easy mistake to make will be when we do rev ipset to the next version, we'll forget to update the ipset revision constant and all the tests will fail.  Suggest adding a canary test somewhere that just checks which revision ipset generates so that we get a clear failure instead of a timeout that needs untangling.

felix/ipsets/ipsets_test.go
mazdakn: Added one in the same test.

felix/ipsets/ipsets.go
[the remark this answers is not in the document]
mazdakn: The idea is not about parsing revision number. Revision field is added to the dataplane mock just to simulate the bug. We can parse revision for sure, but it's tricky since the supported revision number is different for each type. For example, for `hash:net` type, revision number is 7 in both ipset version 7.11 (available in node image), and ipset version 7.22 (which users are hitting the bug with). But for `hash:ip` type the supported revision number is different (6 and 7 respectively).
It's possible to hard code supported revision numbers or parse the output of `ipset help` to get the revision number for each type, but IMO that adds unnecessary complexity.

Additionally, in this bug, we can't even parse metadata for an unsupported revision number. We simply receive an error message like
```
ipset v7.11: Kernel and userspace incompatible: settype hash:ip with revision 6 not supported by userspace.
```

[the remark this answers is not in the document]
mazdakn: The idea here is to try to parse metadata as before. But instead always update Dataplane state regardless of failure or success.
- If reading ipset metadata from dataplane was a success, then we continue as before (the outcome would be different based on the desired state)
- If it was a failure for any reason (like incompatibility), just add a zero dataplane metadata. If after some retry reading from dataplane still fails, then we continue, but ipset will be re-created since the dataplane metadata (since it's zero) differs from the desired state.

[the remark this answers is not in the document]
mazdakn: Instead of using an empty dataplane metadata. We can add a bool field like `ListFailed` and set to `true` to explicitly mark that reading from dataplane was not successful. It might be more verbose than just using a zero value. WDYT?

felix/ipsets/ipsets.go
mazdakn: The idea here is to:
- Retry a new iteration, if dealing with a transient error. 
- Continue with the next steps of iteration if dealing with a permanent error like incompatibility. By continuing we might be able to re-create faulty ipsets (by swapping).

Whether we are dealing with a transient or permanent error is identified based on the iteration number. The same logic is used in the next section of the for loop.

felix/ipsets/ipsets.go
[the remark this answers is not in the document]
fasaxc: Ah, I understand now;  yes, would be good to have something explicit with a comment explaining what's going on?  Perhaps `dataplaneMetadata.ListError = "ipset v7.11: Kernel and userspace incompatible: settype hash:ip with revision 6 not supported by userspace."`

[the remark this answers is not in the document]
mazdakn: OK, I added `ListError` and it's set directly from ipset listing returned error: `dataplaneMetadata.ListError = err`.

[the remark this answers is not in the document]
mazdakn: I tried with `ListError` field. But I don't think it provides any more insight than a single bool field like `ListFailed` mainly because we already have the error message in other logs. That's why I switched back to using a bool field. 

This is an example of logs. It clearly spit out the error message.
```
2025-11-13 22:40:30.698 [ERROR][748] test/ipsets.go 741: Bad return code from 'ipset list cali40t:qMt7iLlGDhvLnCjM0l9nzxb'. error=revision 6 not supported family="inet" stderr=""                                                                             
2025-11-13 22:40:30.698 [DEBUG][748] test/ipsets.go 697: Parsed metadata from dataplane {Type: MaxSize:0 RangeMin:0 RangeMax:0 DeleteFailed:false ListFailed:true} family="inet" setName="cali40t:qMt7iLlGDhvLnCjM0l9nzxb"                                     
2025-11-13 22:40:30.698 [ERROR][748] test/ipsets.go 475: Failed to parse ipset cali40t:qMt7iLlGDhvLnCjM0l9nzxb error=revision 6 not supported family="inet"                         
```
WDYT?

felix/ipsets/ipsets.go
```diff
@@ -44,6 +45,7 @@ type dataplaneMetadata struct {
 	RangeMin     int
 	RangeMax     int
 	DeleteFailed bool
+	listFailed   bool
```
fasaxc:
```suggestion
	ListFailed   bool
```
Just to be consistent with the other fields

felix/ipsets/ipsets.go
```diff
@@ -353,10 +355,11 @@ func (s *IPSets) ApplyUpdates(listener UpdateListener) {
 		retryDelay *= 2
 	}
 
-	for attempt := 0; attempt < 10; attempt++ {
+	for attempt := 0; attempt < MaxRetryAttempt; attempt++ {
 		if attempt > 0 {
 			s.logCxt.Info("Retrying after an ipsets update failure...")
 		}
+		treatFailureAsTransient := attempt < MaxRetryAttempt/2
```
fasaxc, line 360: Nice use of a variable to name the condition

felix/ipsets/ipsets.go
```diff
@@ -378,15 +389,15 @@ func (s *IPSets) ApplyUpdates(listener UpdateListener) {
 
 		dirtyIPSets := s.dirtyIPSetsForUpdate()
 		if err := s.tryUpdates(dirtyIPSets, listener); err != nil {
-			if attempt >= 5 {
+			if treatFailureAsTransient {
+				// More than one failure, resync the IP sets that we failed to update.
```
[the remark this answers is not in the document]
fasaxc: Agree with copilot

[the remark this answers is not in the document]
fasaxc: Maybe I'm getting lost but can you get here with treatFailureAsTransient==true?  Wouldn't you go through the continue above

felix/ipsets/ipsets.go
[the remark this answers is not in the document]
fasaxc: Happy with the bool

felix/ipsets/ipsets.go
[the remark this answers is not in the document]
mazdakn: Well, branches related to `treatFailuresAsTransient == true` are only executed when local err variables are not nil. if `tryResync` (the first section) is a success, but `tryUpdates` (this section) fails these two branches are executed based on `treatFailuersAsTransient` values. 

I agree the logic in this loop is a bit hard to read, and I was going to re-factor it. But I think since this bug fix is tricky by itself, it's probably better to leave any refactoring outside of this one. WDYT?

felix/ipsets/ipsets.go
mazdakn: Applied.

felix/ipsets/ipsets.go
```diff
@@ -365,8 +368,16 @@ func (s *IPSets) ApplyUpdates(listener UpdateListener) {
...
+				// This could be due to different failures, like userspace and kernel incompatibility.
+				// The incompatibility failure can be fixed by swapping the one Felix understand (created from
+				// desired state) and the one (with higher revision) in dataplane. As such, we should stop re-trying
+				// to re-sync, and instead continue with the next steps.
+				if treatFailureAsTransient {
+					backOff()
+					continue
+				}
 			}
 			s.fullResyncRequired = false
```
fasaxc: Still looks wrong to me that, if tryResync fails and we're on attempt >5, we fall through and set `s.fullResyncRequired = false`.  That's meant to be the "success, I'm in sync" flag and we didn't have success

felix/ipsets/ipsets_test.go
```diff
@@ -464,6 +464,57 @@ var _ = Describe("IP sets dataplane", func() {
...
+
+		/*dataplane.ExpectMembers(map[string][]string{
+			v4MainIPSetName: []string{"10.0.0.1", "10.0.0.2"}, // New IPSet from the desired state.
+			//v4TempIPSetName0: []string{"10.0.0.1", "10.0.0.3"}, // Temp IPSet from dataplane state.
+
+			// v4MainIPSetName2 should be destroyed since it's not in the desired state.
+
+			v4MainIPSetName3: []string{"10.0.0.5", "10.0.0.6"}, // This IPSet should not be touched.
+		})*/
+		/*Expect(dataplane.IPSetMembers).To(HaveLen(3))
```
mazdakn: temp change. Will be updated.

felix/ipsets/ipsets.go
```diff
@@ -353,46 +355,62 @@ func (s *IPSets) ApplyUpdates(listener UpdateListener) {
...
-			s.fullResyncRequired = false
 		}
 
 		// Opportunistically delete some temporary IP sets.  It's possible
 		// that ApplyDeletions doesn't get called if there's another failure
 		// and deleting some temp sets might free up some room.
 		s.tryTempIPSetDeletions()
+		if resyncErr != nil {
+			_ = s.ApplyDeletions()
+		}
```
mazdakn: @fasaxc I need you input here. Based on these changes, if we can't list an ipset that's in the desired stated, we re-create and swap it. But, if listing an ipset that's not in desired state fails, we won't re-create it, and listing it persistently fails.

The only option to recover from this state is destroying it. The new UT covers this case as well.
This is the reason `ApplyDeletion` is called here. However, I'm not sure if this is the best approach (Also the comment above kind of implies that we might not want to call `AllyDeletion` here). Maybe , similar to `ApplyDeletion` we should return a re-schedule flag to dataplane manager, so they re-try `ApplyUpdate`. WDYT?

felix/ipsets/ipsets.go
mazdakn: Right. Please take a look at the new changes.

felix/ipsets/ipsets.go
fasaxc: You can't delete an IP set that's in use.  That's why the ApplyUpdates and ApplyDeletions methods are separate.  It allows us to call ApplyUpdates before we update iptables (to set up the IP sets that the iptables rules need); Then to call ApplyDeletions after the iptables update, when we're no longer referencing them.

It's absolutely mainline for the IP set to still be referenced when we try to do updates.

You can _try_ to delete it, but it'll probably fail.

fasaxc: Question is, do we actually know we can hit this case in practice, or is it just a "what if?" scenario

felix/ipsets/ipsets.go
```diff
@@ -681,9 +698,13 @@ func (s *IPSets) resyncIPSet(ipSetName string) error {
 		return scanner.Err()
 	})
 	if err != nil {
-		return err
+		meta.ListFailed = true
```
fasaxc, line 707:
```suggestion
		// This can occur if we have version skew with the version of IP set
		// used to create the IP set. Mark the metadata as invalid in order 
		// to trigger the IP set to be recreated.
		meta.ListFailed = true
```

mazdakn: Applied.

felix/ipsets/ipsets.go
dailai: We usually run into this kind of problem when we just install k3s, so I think these incompatible ipsets should be able to be removed. However, it is not clear to me why this incompatible ipset appears just after installing k3s.

felix/ipsets/ipsets.go
```diff
@@ -462,16 +475,28 @@ func (s *IPSets) tryResync() (err error) {
...
+			_, desired := s.setNameToProgrammedMetadata.Desired().Get(name)
+			if desired {
+				failedIPSets = append(failedIPSets, name)
+			}
+			if debug {
+				s.logCxt.WithError(err).WithFields(log.Fields{
+					"name":    name,
+					"desired": desired,
+				}).Error("Failed to parse an ipset")
+			}
```
fasaxc: I think it's worth logging at warning level here (and worth tailoring the message to avoid worrying users immediately.
```suggestion
			if desired {
				s.logCxt.WithError(err).WithFields(log.Fields{
					"name":    name,
				}).Warn("Failed to parse required Calico-owned ipset that is needed, will try recreating it.")
			} else {
				s.logCxt.WithError(err).WithFields(log.Fields{
					"name":    name,
				}).Warn("Failed to parse Calico-owned ipset that is no longer needed, will queue it for deletion.")
			}
```

felix/ipsets/ipsets.go
```diff
@@ -353,22 +355,31 @@ func (s *IPSets) ApplyUpdates(listener UpdateListener) {
...
-				backOff()
-				continue
+			if resyncErr = s.tryResync(); resyncErr != nil {
+				s.logCxt.WithError(resyncErr).Warning("Failed to resync with dataplane")
+
+				// After a few attempts, most likely, we are dealing with a persistent failure.
+				// This could be due to different failures, like userspace and kernel incompatibility.
+				// The incompatibility failure can be fixed by swapping the one Felix understand (created from
+				// desired state) and the one (with higher revision) in dataplane. As such, we should stop re-trying
+				// and instead fall through to the next steps.
```
fasaxc:
```suggestion
				// The incompatibility failure can be fixed by swapping the one Felix understand (created from
				// desired state) and the one (with higher revision) in dataplane. As such, we should stop re-trying
				// and instead fall through to tryUpdates() below, which will do the swap.
```

PR #11342, opened 2025-11-07
Refactor JSON parsing to go through Key/Resource type.

fasaxc:
## Description

Previously, parsing was handled by one central function that handled all special cases.

Move parsing to a method on the key type:

- Main advantage is that it makes stack traces for parsing unique so pprofs attribute the usage to the particular type.
- Uses generics to instantiate objects instead of reflection,.
- Move special cases to the individual key types.
- Make resourceInfo generic instead of using reflection; cleaner and fits more nicely with the new method approach.
- Remove some old special case logic that seemed to be defunct.

image

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Adjust JSON parsing logic to give unique stack trace per resource type.  Makes memory profiles a lot more descriptive.
```

## Reminder for the reviewer

Other optional labels:


libcalico-go/lib/backend/model/bgp_node.go
```diff
@@ -50,6 +50,10 @@ func (key BGPNodeKey) valueType() (reflect.Type, error) {
 	return typeBGPNode, nil
 }
 
+func (key BGPNodeKey) parseValue(data []byte) (interface{}, error) {
```
caseydavenport:
```suggestion
func (key BGPNodeKey) parseValue(data []byte) (any, error) {
```

Not really a big deal, but I get linter errors now when using `interface{}` instead of `any

libcalico-go/lib/backend/model/resource.go
```diff
@@ -244,9 +160,44 @@ func (key ResourceKey) defaultDeleteParentPaths() ([]string, error) {
...
-	return ri.typeOf, nil
+	return ri.TypeOf(), nil
+}
+
+func (key ResourceKey) parseValue(rawData []byte) (any, error) {
+	ri, ok := resourceInfoByKindLower[strings.ToLower(key.Kind)]
+	if !ok {
+		return nil, fmt.Errorf("unknown resource kind: %s", key.Kind)
+	}
+	v, err := ri.ParseValue(key, rawData)
```
caseydavenport, line 171: missing an err handling branch here

PR #11357, opened 2025-11-10
[BPF][Maglev] include FROM_WEP traffic in the midflow-miss check

aaaaaaaalex:
## Description

Currently, we open a pinhole in the midflow-miss handling in order to allow potential Maglev packets further into the TC program (before ultimately forwarding/dropping depending on the Maglev decision).

The pinhole is currently too big, allowing FROM_WEP, TO_HOST packets through to the Maglev check. This is unnecessary because Maglev failover only applies to HEP packets. So, this PR narrows the pinhole to exclude WEP pkts.

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -384,44 +384,60 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 		goto deny;
 	}
 
-	// We avoid this block when TO_HOST, since we must perform a NAT lookup
-	// to determine if such packets belong to a Maglev failover connection.
-	if (ct_result_rc(ctx->state->ct_result.rc) == CALI_CT_MID_FLOW_MISS && !CALI_F_TO_HOST) {
-		if (CALI_F_HEP) {
-			// HEP egress for a mid-flow packet with no BPF or Linux CT state.
-			// This happens, for example, with asymmetric untracked policy,
-			// where we want the return path packet to be dropped if there is a
-			// HEP present (regardless of the policy configured on it, for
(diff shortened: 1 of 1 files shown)


felix/bpf-gpl/tc.c
```diff
@@ -384,9 +384,12 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
...
 
-	// We avoid this block when TO_HOST, since we must perform a NAT lookup
-	// to determine if such packets belong to a Maglev failover connection.
-	if (ct_result_rc(ctx->state->ct_result.rc) == CALI_CT_MID_FLOW_MISS && !CALI_F_TO_HOST) {
+	// Handle all non-Maglev midflow misses here.
+	// That's all misses:
+	// - from WEPs to/from host
+	// - from HEPs *from* from host
+	// That leaves misses from HEPs *to* host to be checked as Maglev failover.
+	if (ct_result_rc(ctx->state->ct_result.rc) == CALI_CT_MID_FLOW_MISS && (!CALI_F_TO_HOST || CALI_F_FROM_WEP)) {
```
fasaxc:
```suggestion
	// Handle non-Maglev mid-flow packets.  Maglev only applies on the "from HEP" path.
	if (ct_result_rc(ctx->state->ct_result.rc) == CALI_CT_MID_FLOW_MISS && !CALI_F_FROM_HEP) {
```
Does `CALI_F_FROM_HEP` exist?  If so, feels cleaner to use that instead of a combination of other flags.  If not, we could add it.

PR #11360, opened 2025-11-11
Fix UT/FV test reports and logs collection.

fasaxc:
## Description

We weren't getting logs from the felix FV runs.  When I pulled on the thread to figure out why, it all unravelled:

* We had multiple copies of scripts with inconsistent expectations about the directory they were called from.
* Scripts were being called from different directories, resulting in not agreeing on where logs and test reports should go.
* After fixing _that_ I found that we were missing a lot of test reports because the ginkgo suite files weren't correct.
* After getting test reports into semaphore, I found they were inconsistently named and generally a bit of a mess.
* We were also missing key tests, such as the BPF UTs, which don't use Ginkgo.
* The `test-results` script is finicky and `test-results publish` can only be run once per job (so some jobs failed to push some diags)

After several days of iteration, I think this is in a much better place, with 16k+ well-organised tests reporting in a well(ish) organised way:

* Duplicate scripts axed.
* All Felix FV runs reporting test results (including remote ones); uses custom name so that they don't clobber each other in the main report.
* All Felix UT runs reporting (including BPF ones)
* All Felix FVs reporting all their logs.
* Node STs reporting test results and logs.
* CI job names normalised so that every job has a full name (so that the per-job web page actually tells you what you're looking at).

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


typha/pkg/tlsutils/tlsutils_suite_test.go
```diff
@@ -30,6 +30,6 @@ func init() {
 
 func TestTLSUtils(t *testing.T) {
 	RegisterFailHandler(Fail)
-	junitReporter := reporters.NewJUnitReporter("../report/tls_utils_suite.xml")
+	junitReporter := reporters.NewJUnitReporter("../../report/tls_utils_suite.xml")
```
aaaaaaaalex, line 33: Not asking for it in this PR, but it'd be great to ban `../`  IMO. I find it to be a mental burden

fasaxc: Oh yeah, I thought about doing `testutils.AutoConfigureGinkgo()` or something but I need to bank something!!

fasaxc: Also ginkgo v2 is out, so it'd be good to switch to that; it has a completely different way of doing test reports.

Again, wanted to bank something before this grew and grew.

PR #11380, opened 2025-11-13
calico-dhcp-agent: don't include mtu in router advertisements

kristiangronas:
## Description

The calico plugin does not have the network mtu in it's data model, so i think it's better to just leave it out, and then let the instance set it to the same value as the interface
```
$ cat /proc/sys/net/ipv6/conf/eth0/mtu 
1500
```

The overall goal is to allow instances to use jumbo frames over ipv6, like they can for ipv4

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
OpenStack: don't force the MTU to 1500 in IPv6 router advertisements
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/agent/linux/dhcp.py
+++ b/networking-calico/networking_calico/agent/linux/dhcp.py
@@ -219,6 +219,8 @@ class DnsmasqRouted(dhcp.Dnsmasq):
 
         # Add '--enable-ra'.
         cmd.append("--enable-ra")
+        # Don't advertise the MTU, since we don't have the value here
+        cmd.append("--ra-param=ns-*,mtu:off,0")
 
         # Enumerate precisely the TAP interfaces to listen on.
         cmd.remove("--interface=tap*")
--- a/networking-calico/networking_calico/tests/test_dhcp_agent.py
+++ b/networking-calico/networking_calico/tests/test_dhcp_agent.py
@@ -485,6 +485,7 @@ class TestDnsmasqRouted(base.BaseTestCase):
                 "--dhcp-range=set:subnet-v6subnet-1,2001:db8:1::"
                 + ",static,off-link,80,86400s",
                 "--enable-ra",
+                "--ra-param=ns-*,mtu:off,0",
                 "--interface=tap1",
                 "--interface=tap2",
                 "--interface=tap3",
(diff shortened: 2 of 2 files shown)


mazdakn:
@nelljerram can you please take a look at this one.

nelljerram:
@kristiangronas Looks good, thanks.  For the record, please can you just add a few more words here about how the correct behaviour - i.e. _not_ setting `ra-params` - results in the MTU being forced to 1500?  Also am I right that:
1. we're talking about the TAP interfaces to guest VMs?
2. the MTU gets set on the VM side, as a result of the VM receiving the RA?

nelljerram:
@kristiangronas You need a `make -C networking-calico fmtpy`, to fix this:
```
-        cmd.append('--ra-param=ns-*,mtu:off,0')
+        cmd.append("--ra-param=ns-*,mtu:off,0")
```

kristiangronas:
When you don't set the param, it auto-configures the mtu to the same value as the ns interface (which gets set to the placeholder 1500 value) which will look like `mtu option (5), length 8 (1):  1500`, then when an instance receives this in an RA, it will add it to e.g. /proc/sys/net/ipv6/conf/eth0/mtu, so you can't use jumbo frames for ipv6 anymore. By turning off mtu advertisements, dnsmasq does not include the MTU option at all

When the instance boots, the mtu of the interface inside the VM (e.g. eth0) is instead the same as the tap interface on the hypervisor, which is a little less ideal than dynamic as per-neutron port dhcp-configuration like with ipv4. Sadly I didn't find a good way to replicate that, since the ra-param flags points to the ns interface, so i couldn't say that requests to tap-2.. should get mtu 9000 for example

The result is that i can send large packets without fragmentation: 
`13:36:24.789096 eth0  In  IP6 wc-worker-001 > wc-worker-005: ICMP6, echo request, id 14, seq 329, length 8948`
`13:36:24.789137 eth0  Out IP6 wc-worker-005 > wc-worker-001: ICMP6, echo reply, id 14, seq 329, length 8948`

nelljerram:
Many thanks @kristiangronas

nelljerram (approved):

Issue #11311, opened 2025-11-04
Unable to delete or modify ipv6 ippool


Mmx233:
Unable to modify or delete the ipv6 ippool once created.

```shell
root@mmx-cluster-03:~# kubectl delete ippools ipv6-ippool 
Error from server: error with field IPPool.Spec.CIDR = '2001:cafe:42::/56' (IPPool CIDR cannot be modified)
```

ippool content:

```shell
root@mmx-cluster-03:~# kubectl get ippools ipv6-ippool -o yaml
apiVersion: projectcalico.org/v3
kind: IPPool
metadata:
  creationTimestamp: "2025-02-22T19:26:10Z"
  generation: 1
  labels:
    app.kubernetes.io/managed-by: tigera-operator
  name: ipv6-ippool
  resourceVersion: "535353192"
  uid: a96ed48f-1e2c-4688-8532-b368ce0e8e42
spec:
  allowedUses:
  - Workload
  - Tunnel
  assignmentMode: Automatic
  blockSize: 122
  cidr: 2001:cafe:42::00/56
  ipipMode: Never
  natOutgoing: true
  nodeSelector: all()
  vxlanMode: CrossSubnet
```

## Expected Behavior

Able to delete or modify the ippool

## Current Behavior

Unable to delete with error `error with field IPPool.Spec.CIDR = '2001:cafe:42::/56' (IPPool CIDR cannot be modified)`

## Possible Solution

N/A

## Steps to Reproduce (for bugs)

1. Install tigera operator with value like:

```yaml
whisker:
  enabled: true
goldmane:
  enabled: true
kubernetesServiceEndpoint:
  host: "10.0.0.1"
  port: "6443"
installation:
  controlPlaneTolerations:
    - key: node.kubernetes.io/network-unavailable
      operator: Exists
  cni:
    type: Calico
  calicoNetwork:
    bgp: Disabled
    containerIPForwarding: Enabled
    linuxDataplane: BPF
    nodeAddressAutodetectionV4:
      interface: enp.*s.*
    nodeAddressAutodetectionV6:
      interface: enp.*s.*
    ipPools:
      - blockSize: 26
        cidr: 10.42.0.0/16
        encapsulation: VXLANCrossSubnet
        name: ipv4-ippool
        natOutgoing: Enabled
        nodeSelector: all()
      - blockSize: 122
        cidr: 2001:cafe:42::/56
        encapsulation: VXLANCrossSubnet
        name: ipv6-ippool
[...]

PR #11385, opened 2025-11-14
Fixed ippool cidr string validation

skoryk-oleksandr:
## Fix: IPPool CIDR Validation Fails on Semantically-Identical IPv6 CIDRs

**User Issue:** https://github.com/projectcalico/calico/issues/11311

 - **Problem**

Updating or deleting an IPPool fails when the CIDR is written using a different—but semantically identical—IPv6 textual form.
Example that currently fails validation:

 - Existing pool: `2001:cafe:42::00/56`
 - Updated manifest: `2001:cafe:42::/56`

Although both represent the same IPv6 network, the validator rejects the update with:

`IPPool CIDR cannot be modified`

This happens because the CIDR immutability check performs a string equality comparison, not a semantic comparison of IP network and mask.

As a result, simple textual differences caused by IPv6 zero-compression or omitted leading zeros trigger false-positive validation errors.

 - **Root Cause**

The IPPool validator uses:

`old.Spec.CIDR != new.Spec.CIDR`

This compares raw strings, ignoring the fact that network-equivalent IPv6 representations normalize differently depending on formatting (e.g., `::00` vs `::`).

Meanwhile, IPPool normalization already happens inside the library - causing the incoming object to be normalized but not matching the stored string.

 - **Solution**

Introduce a helper that compares IPv6 CIDRs semantically, not textually.

New helper: `equalCIDRStrings()`

Parses both CIDRs using `cnet.ParseCIDR()`

Compares IP network addresses and masks

Falls back to string comparison if parsing fails (preserves existing handling of malformed CIDRs)

Updated immutability check

**Old:**
`
if old.Spec.CIDR != new.Spec.CIDR {
    ...
}
`
**New:**
`
if !equalCIDRStrings(old.Spec.CIDR, new.Spec.CIDR) {
    ...
}
`

This ensures that formatting differences like:

`::00/56` **vs** `::/56`

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix: IPPool CIDR Validation Fails on Semantically-Identical IPv6 CIDRs
```

## Reminder for the reviewer

Other optional labels:


fasaxc:
I'd flip the logic:

* if strings equal, return true
* Otherwise, parse and compare and only accept if the new CIDR is semantically equal *and* in canonical form.  So that it's not valid to go from cafe::/100  to cafe:0000::00/100; you must be correcting the non-canonical address, not introducing it!

k8s is moving towards validating that pod IPs etc are in canonical form.

skoryk-oleksandr:
> I'd flip the logic:
> 
> * if strings equal, return true
> * Otherwise, parse and compare and only accept if the new CIDR is semantically equal _and_ in canonical form.  So that it's not valid to go from cafe::/100  to cafe:0000::00/100; you must be correcting the non-canonical address, not introducing it!
> 
> k8s is moving towards validating that pod IPs etc are in canonical form.

@fasaxc Fixed :+1:

libcalico-go/lib/clientv3/ippool.go
```diff
@@ -333,9 +365,9 @@ func (r ipPools) validateAndSetDefaults(ctx context.Context, new, old *apiv3.IPP
 	}
 
 	// If there was a previous pool then this must be an Update, validate that the
-	// CIDR has not changed.  Since we are using normalized CIDRs we can just do a
```
caseydavenport, line 336: I am a little bit surprised here - this comment seems to suggest we thought we had handled this already via normalizing the CIDRs on create? Do we not?

skoryk-oleksandr: You're right — when an IPPool is created on a running Calico cluster, the CIDR is normalized during creation.

However, the issue appears when IPPools are provided as part of a new Calico installation (for example, through operator-managed manifests or Helm values). In that flow, normalization does not occur before the validation step. As a result, it's possible to end up with an IPPool whose CIDR is stored in a non-normalized form, and later attempts to update or delete it fail because the validation logic compares the raw strings.

caseydavenport: ick. I see. That's still effectively a bug in those codepaths, rather than a bug here, but I suppose now that we have those types of pools floating around we need to handle them. I do worry there might be other places in our code where we rely on this normalization being done, though.

skoryk-oleksandr: @caseydavenport I fixed few more places where we compare CIDRs.  67502ecdf284213eebf1b0f4a1b9aa33c1ed2b03
Operator part will be fixed in separate ticket.

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -117,7 +117,7 @@ Examples:
 		}
 
 		for _, pool := range poolList.Items {
-			if pool.Spec.CIDR == oldPoolCIDR {
+			if equalCIDRStrings(oldPoolCIDR, pool.Spec.CIDR) {
```
caseydavenport: I think in this case, we do actually expect exact CIDR matching? Since we're searching for a specific IP pool based on its CIDR.

felix/rules/policy.go
```diff
@@ -157,7 +158,14 @@ func filterNets(mixedCIDRs []string, ipVersion uint8, isNegated bool) (filtered
 // isCatchAllCIDR returns true if the CIDR represents "all addresses" for the given IP version.
 // This is used to detect problematic negated matches that would create logical contradictions.
 func isCatchAllCIDR(cidr string, ipVersion uint8) bool {
-	return (ipVersion == 4 && cidr == "0.0.0.0/0") || (ipVersion == 6 && cidr == "::/0")
```
caseydavenport, line 160: Similarly here this isn't based on the IP pool CIDR, so I think we can trust that it is normalized. I am not sure that adding this ParseCIDR into the mainline path here will be worthwhile given potential performance impacts.

libcalico-go/lib/clientv3/ippool.go
```diff
@@ -296,6 +297,37 @@ func (r ipPools) Watch(ctx context.Context, opts options.ListOptions) (watch.Int
 	return r.client.resources.Watch(ctx, opts, apiv3.KindIPPool, nil)
 }
 
+func equalCIDRStrings(oldCIDR, newCIDR string) bool {
```
caseydavenport:
```suggestion
// cidrChangeOK returns true if the new CIDR is a valid and equivalent version of the old. We allow identical CIDRs. We also allow the new CIDR to be more canonical than the old (but not vice-versa). 
// This is needed because some clients may bypass our validation code that ensures a canonicalized CIDR, and so we must account for unconventional representations being present in our API.
func cidrChangeOK(oldCIDR, newCIDR string) bool {
```

A bit of a nit, but I think the name change is worthwhile since this isn't really an equivalence check - it's checking if thew new CIDR value is allowable based on the current CIDR value.

Also added a comment since this need for this type of logic isn't super intuitive.

caseydavenport (commented):
This looks good to me! A few minor comments.

calicoctl/calicoctl/commands/ipam/split.go
skoryk-oleksandr: This change simply extends the behavior to also handle cases where a user searches for a pool using a non-normalized CIDR while the stored one is normalized (or vice-versa). Exact string matches will continue to work as before — this just makes the lookup more tolerant to equivalent CIDR representations.
That said, I agree this isn’t a critical improvement, and it might be reasonable to keep the strict exact-match behaviour if we prefer consistency.

PR #11398, opened 2025-11-17
Speed up Felix FVs

fasaxc:
## Description

* Avoid rebuilding felix/FV test binaries in every remote test runner VM.

  * We were already building the needed artifacts in the "build" job; store them in a GCS bucket instead of the Semaphore cache so that they become available (with very high download speed) in the test VMs.
  * The bucket is set to auto-expire files after a short time to avoid bloat.
  * Adjust remote VM setup script to pull the artifacts and run a new `fv-bpf-no-prereqs` target, which works like the mainline one.
 Overall, this reduces the time from starting tests to first test by about 9 minutes, bringing it down to 4-5s!

* Increase the number of remote runners to bring the total time down (aiming for <40 minutes).  The reduced overhead means that this shouldn't increase GCP instance hour cost.

  * Since increasing the number of runners dramatically means that set-up flakes become more common, make a bunch of robustness fixes to set-up: add retries, use bulk VM creation; reduce number of ssh commands that have to succeed.
  * Stop using `gcloud compute ssh` and manually configure the VMs to allow the semaphore instance's ssh key.  `gcloud compute ssh` becomes unreliable when running many instances in parallel (due to all the commands trying to update project metadata etc).  Get the VM's public ssh key from the gcloud API to avoid known host warnings.
  * Make log monitoring resumable by using nohup and tailing the log in an ssh session.
  * Fix a bunch of bash bugs, including making sure that we kill process groups, not just the "lead" process.
  * Set a time limit on each VM instance.

* Move "main" FVs to run on remote VMs, this reduces the number of semaphore runners needed from 6  `f1-standard-4` runners for over an hour down to a single `f1-standard-2` for ~35 minutes.  This _does_ add GCP instance hour cost (since we weren't doing it before) but it should dramatically reduce pressure on semaphore resources, which are currently causing us pain.

* Refactor FV jobs to call a standard prologue/epilogue script.  We would normally put these FV jobs in a single block so all this logic would be shared, but we want to make sure that "Rebuild pipeline" doesn't rerun the whole block when only one has flaked.

* For now, `node` has a stub implementation of the bucket cache; it just pushes the working copy.  In future, we should push the node images etc and re-use them.
  
## Related issues/PRs

CORE-12073

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/.semaphore/semaphore.yml.d/blocks/20-felix.yml
+++ b/.semaphore/semaphore.yml.d/blocks/20-felix.yml
@@ -210,7 +210,7 @@
           minutes: 180
         env_vars:
           - name: FELIX_TEST_GROUP
-            value: "bpf-22.04-ipt-with-ut"
+            value: "bpf-22.04-nft-with-ut"
           - name: FELIX_FV_BPFATTACHTYPE
             value: "tc"
         commands:
@@ -238,7 +238,7 @@
           minutes: 60
         env_vars:
           - name: FELIX_TEST_GROUP
-            value: "bpf-25.04-ipt-no-fv-with-ut"
+            value: "bpf-25.04-nft-no-fv-with-ut"
           - name: FELIX_FV_BPFATTACHTYPE
             value: "tc"
         commands:
@@ -265,7 +265,7 @@
           minutes: 180
         env_vars:
           - name: FELIX_TEST_GROUP
-            value: "bpf-25.04-ipt-with-ut-jitharden"
+            value: "bpf-25.04-nft-with-ut-jitharden"
           - name: FELIX_FV_BPFATTACHTYPE
             value: "tc"
         commands:
--- a/.semaphore/semaphore-scheduled-builds.yml
+++ b/.semaphore/semaphore-scheduled-builds.yml
@@ -635,7 +635,7 @@ blocks:
             minutes: 180
           env_vars:
             - name: FELIX_TEST_GROUP
-              value: "bpf-22.04-ipt-with-ut"
+              value: "bpf-22.04-nft-with-ut"
             - name: FELIX_FV_BPFATTACHTYPE
               value: "tc"
           commands:
@@ -663,7 +663,7 @@ blocks:
             minutes: 60
           env_vars:
             - name: FELIX_TEST_GROUP
-              value: "bpf-25.04-ipt-no-fv-with-ut"
+              value: "bpf-25.04-nft-no-fv-with-ut"
             - name: FELIX_FV_BPFATTACHTYPE
               value: "tc"
           commands:
@@ -690,7 +690,7 @@ blocks:
             minutes: 180
           env_vars:
             - name: FELIX_TEST_GROUP
-              value: "bpf-25.04-ipt-with-ut-jitharden"
+              value: "bpf-25.04-nft-with-ut-jitharden"
             - name: FELIX_FV_BPFATTACHTYPE
               value: "tc"
           commands:
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -164,7 +164,7 @@ blocks:
       jobs:
         - name: Pre-flight checks
           commands:
-            - true || make ci-preflight-checks
+            - make ci-preflight-checks
   - name: API
     run:
       when: "false or change_in(['/metadata.mk', '/lib.Makefile', '/api/'], {pipeline_file: 'ignore', exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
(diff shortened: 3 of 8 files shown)


.semaphore/vms/create-test-vms
```diff
@@ -16,27 +16,54 @@
...
+       --zone=${zone} \
+       --machine-type=n4-standard-4 \
+       --image-family=${IMAGE_FAMILY} \
+       --image-project=ubuntu-os-cloud \
+       --boot-disk-size=$disk_size \
+       --boot-disk-type=hyperdisk-balanced \
+       --max-run-duration="${MAX_RUN_DURATION}" \
+       --instance-termination-action=DELETE \
+       --metadata-from-file startup-script="$my_dir/vm-bootstrap.sh" \
+       --metadata block-project-ssh-keys=TRUE,ssh-keys="ubuntu:$(ssh-keygen -y -f $HOME/.ssh/id_rsa)",enable-guest-attributes=TRUE
```
sridhartigera, line 55: Do we have a quota on the number of VMs that can be created at a time or is it just the API quota? Is this bulk create `all or nothing` way to create VMs?

felix/.semaphore/cache-test-artifacts
```diff
@@ -0,0 +1,37 @@
...
+cd ../..
+
+# Future work: Split UT and FV artifacts so we can be even more selective.
+# Excluding Typha binaries because we only need the image.
+tar -czf /tmp/working-copy.tgz \
+  --exclude=.go-pkg-cache \
+  --exclude=calico/typha/bin \
+  --exclude=calico/typha/docker-image/bin \
+  --exclude=artifacts \
+  ${CALICO_DIR_NAME}
```
sridhartigera, line 35: Rather than copying the source from semaphore vm to every gcp vm, you push it to gcs and pull it from there. Nice!!!

.semaphore/semaphore.yml.d/blocks/20-felix.yml
```diff
@@ -129,238 +126,159 @@
...
-        - ../.semaphore/vms/create-test-vms ${VM_PREFIX}
     jobs:
       - name: "Felix: BPF tests on Ubuntu 22.04 (nftables)"
+        execution_time_limit:
+          minutes: 180
         env_vars:
-          - name: FELIX_FV_NFTABLES
-            value: "Enabled"
+          - name: FELIX_TEST_GROUP
+            value: "bpf-22.04-ipt-with-ut"
```
sridhartigera: Should be `bpf-22.04-nft-with-ut`?

.semaphore/semaphore.yml.d/blocks/20-felix.yml
```diff
@@ -129,238 +126,159 @@
...
     jobs:
-      - name: "Felix: BPF tests on Ubuntu 25.04, load/verify programs"
+      - name: "Felix: BPF program-loading check on 25.04"
+        execution_time_limit:
+          minutes: 60
         env_vars:
-          - name: FELIX_FV_NFTABLES
-            value: "Enabled"
+          - name: FELIX_TEST_GROUP
+            value: "bpf-25.04-ipt-no-fv-with-ut"
```
sridhartigera: should be `nft`

.semaphore/semaphore.yml.d/blocks/20-felix.yml
```diff
@@ -129,238 +126,159 @@
...
     jobs:
-      - name: "Felix: BPF tests with bpf_jit_harden=2"
+      - name: "Felix: BPF tests on Ubuntu 25.04 with jitharden=2"
+        execution_time_limit:
+          minutes: 180
         env_vars:
-          - name: FELIX_FV_NFTABLES
-            value: "Enabled"
+          - name: FELIX_TEST_GROUP
+            value: "bpf-25.04-ipt-with-ut-jitharden"
```
sridhartigera: Same here. `nft`

.semaphore/vms/create-test-vms
fasaxc: Using a bulk creation saves API quota and makes it more robust vs trying to hit the API 60 times in parallel.

The default behavior is all-or-nothing :+1: 

> --min-count=MIN_COUNT
> The minimum number of Compute Engine virtual machines that must be successfully created for the operation to be considered a success. If the operation successfully creates as many virtual machines as specified here they will be persisted, otherwise the operation rolls back and deletes all created virtual machines. If not specified, this value is equal to --count.

We could play with using min-count and making use of whatever number we got but that's a bigger rework.  So many ideas to try but I feel like I need to bank something!

felix/.semaphore/cache-test-artifacts
fasaxc: Yep, instead of 100MiB/s split between all VMs, it's 100MiB/s to GCS and then even faster for each VM to download it.

.semaphore/semaphore.yml.d/blocks/20-felix.yml
fasaxc: :facepalm:

felix/.semaphore/cache-test-artifacts
```diff
@@ -0,0 +1,37 @@
...
+gcloud storage cp /tmp/working-copy.tgz /tmp/calico-felix-test.tar /tmp/felixtest-typha.tar "${GCS_WORKFLOW_DIR}/felix/fv-artifacts/"
+wait
```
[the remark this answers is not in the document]
fasaxc: Best-effort is fine in this case

PR #11415, opened 2025-11-19
Add more time constants and global functions

Brian-McM:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/lib/std/time/time.go
+++ b/lib/std/time/time.go
@@ -13,27 +13,38 @@ type (
 
 	Duration = time.Duration
 
+	Weekday = time.Weekday
+
 	Month = time.Month
 
 	Location = time.Location
 )
 
 const (
-	January Month = 1 + iota
-	February
-	March
-	April
-	May
-	June
-	July
-	August
-	September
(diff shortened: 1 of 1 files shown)


lib/std/time/time.go
```diff
@@ -13,27 +13,38 @@ type (
...
-	December
+	Sunday    Weekday = time.Sunday
+	Monday    Weekday = time.Monday
+	Tuesday   Weekday = time.Tuesday
+	Wednesday Weekday = time.Wednesday
+	Thursday  Weekday = time.Thursday
+	Friday    Weekday = time.Friday
+	Saturday  Weekday = time.Saturday
+
+	January   Month = time.January
```
Brian-McM, line 32: I updated this to be consistent with how I did the new weekday stuff. I realised it was better to do the explicit assigned just incase the ordering changed in the time package (unlikely, but might as well).

PR #10810, opened 2025-08-12
Implement Cluster Network Policy

mazdakn:
## Description

This PR implements [ClusterNetwokrPolicy](https://github.com/kubernetes-sigs/network-policy-api/blob/375c8a75a50a2e030258f8ee7a0c76fd485cf5bc/apis/v1alpha2/clusternetworkpolicy_types.go#L38). It includes:
- Adding two new tier defined at the start of cluster, named `kube-admin`, and `kube-baseline`.
- Convert `ClusterNetworkPolicy` resources to `GlobalNetworkPolicy`.
  - ClusterNetworkPolicies with [Tier](https://github.com/kubernetes-sigs/network-policy-api/blob/375c8a75a50a2e030258f8ee7a0c76fd485cf5bc/apis/v1alpha2/clusternetworkpolicy_types.go#L81) set to `Admin` is placed into Calico `kube-admin` tier.
  - ClusterNetworkPolicies with [Tier](https://github.com/kubernetes-sigs/network-policy-api/blob/375c8a75a50a2e030258f8ee7a0c76fd485cf5bc/apis/v1alpha2/clusternetworkpolicy_types.go#L81) set to `Baseline` is placed into Calico `kube-baseline` tier.

This PR tracks the main branch of upstream repo: https://github.com/kubernetes-sigs/network-policy-api/tree/main
A follow up PR will pin to a specific tag when it's created upstream.

## Related issues/PRs
Operator: https://github.com/tigera/operator/pull/4155
Removing ANP/BANP: https://github.com/projectcalico/calico/pull/11144
Goldmane support: https://github.com/projectcalico/calico/pull/11417
Docs: https://github.com/tigera/docs/pull/2416

Pick to private: https://github.com/tigera/calico-private/pull/10621

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add support for the standard and experimental (networks peers and named ports) functionalities of the new k8s ClusterNetworkPolicy API.
Two new static tiers named kube-admin(order 1,000) and kube-baseline(order 10,000,000) are created at startup where ClusterNetworkPolicy resources in Admin tier, and ClusterNetworkPolicy in Baseline resources in Baseline tier are enforced respectively. 
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
[...]

PR #11417, opened 2025-11-19
Goldmane: Support ClusterNetworkPolicy

mazdakn:
## Description

This PR includes the following changes in Goldmane/Whisker:
- Adding support for `ClusterNetworkPolicy` as it's now supported in Calico: https://github.com/projectcalico/calico/pull/10810
- Remove support of `AdminNetworkPolicy` and `BaselineAdminNetworkPolicy` since these are going to be removed soon: https://github.com/projectcalico/calico/pull/11144

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/whisker-backend/pkg/apis/v1/flows.go
+++ b/whisker-backend/pkg/apis/v1/flows.go
@@ -166,9 +166,8 @@ const (
 	PolicyKindStagedGlobalNetworkPolicy     = PolicyKind(proto.PolicyKind_StagedGlobalNetworkPolicy)
 	PolicyKindStagedKubernetesNetworkPolicy = PolicyKind(proto.PolicyKind_StagedKubernetesNetworkPolicy)
 
-	PolicyKindNetworkPolicy                    = PolicyKind(proto.PolicyKind_NetworkPolicy)
-	PolicyKindClusterNetworkPolicyAdminTier    = PolicyKind(proto.PolicyKind_ClusterNetworkPolicyAdminTier)
-	PolicyKindClusterNetworkPolicyBaselineTier = PolicyKind(proto.PolicyKind_ClusterNetworkPolicyBaselineTier)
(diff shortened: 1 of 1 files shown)


goldmane/proto/api.proto
```diff
@@ -247,8 +247,8 @@ enum PolicyKind {
 
   // Native Kubernetes types.
   NetworkPolicy = 6;
-  AdminNetworkPolicy = 7;
-  BaselineAdminNetworkPolicy = 8;
+  ClusterNetworkPolicyAdminTier = 7;
+  ClusterNetworkPolicyBaselineTier = 8;
```
mazdakn: How about `AdminClusterNetworkPolicy` and `BaselineClusterNeworkPolicy`?

caseydavenport: Do we need different types for the different tiers? that seems wrong to me. These should map directly to `Kind` and not anything else.

mazdakn: I did'nt find a way to restructure the trace string from hit trace, but then realised it store all information to do so. Fixed.

goldmane/proto/methods.go
```diff
@@ -21,6 +21,8 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/kubernetes/pkg/apis/core/validation"
+
+	"github.com/projectcalico/calico/libcalico-go/lib/names"
```
mazdakn, line 25: Did we intentionally hard coded prefixes in this implementation instead of using the const defined in the `names` package? Using it add dependency between goldmane and libcalico-go. Wondering if this was by purpose.

caseydavenport: I think it was unintentional, but also largely because this code was / is considered a bit throwaway. Fine to introduce the libcalico import.

Issue #10124, opened 2025-04-02
calico-kube-controller watch or Decode memory leaks


xiaoertong:
[pprof_heap.pdf](https://github.com/user-attachments/files/19565473/pprof_heap.pdf)

## Expected Behavior
If the cluster size and network resource consumption remain the same, the memory does not continue to increase.

## Current Behavior
The memory usage increases over time. Some functions have been found to use a lot of memory.(pprof:  Specifically as follows)

kube-controllers
Type: inuse_space
Time: Apr 2, 2025 at 2:34pm (CST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 61362.12MB, 99.44% of 61707.29MB total
Dropped 222 nodes (cum <= 308.54MB)
Showing top 10 nodes out of 42
      flat  flat%   sum%        cum   cum%
20969.57MB 33.98% 33.98% 20969.57MB 33.98%  reflect.unsafe_NewArray
11715.03MB 18.98% 52.97% 13553.15MB 21.96%  github.com/json-iterator/go.(*Iterator).ReadString
10701.31MB 17.34% 70.31% 10701.31MB 17.34%  k8s.io/apimachinery/pkg/apis/meta/v1.(*FieldsV1).UnmarshalJSON
 6873.71MB 11.14% 81.45%  6873.71MB 11.14%  reflect.mapassign
 4853.71MB  7.87% 89.31%  4853.71MB  7.87%  github.com/modern-go/reflect2.(*unsafeType).UnsafeNew
 3731.14MB  6.05% 95.36%  3731.14MB  6.05%  reflect.New
 1838.12MB  2.98% 98.34%  1838.12MB  2.98%  github.com/json-iterator/go.(*Iterator).readStringSlowPath
  679.53MB  1.10% 99.44%   679.53MB  1.10%  reflect.makemap
         0     0% 99.44%    57589MB 93.33%  github.com/json-iterator/go.(*Iterator).ReadVal
         0     0% 99.44% 20434.57MB 33.12%  github.com/json-iterator/go.(*OptionalDecoder).Decode

## Steps to Reproduce (for bugs)

1. Create a k8s cluster. 
2. After use for a period of time, the memory is found to be increasing.

## Your Environment

* Calico version : 3.22.1
* Calico dataplane (iptables, windows etc.)
* Orchestrator version (e.g. kubernetes, mesos, rkt): k8s
[...]

PR #11433, opened 2025-11-22
fix watcher cleanup in kube-controllers runconfig

fusidic:
## Description

In my Kubernetes environment (Kubernetes 1.28, Calico 3.27.4), I observed a slow, persistent increase in memory usage within the calico-kube-controllers. This appeared to indicate a potential memory leak. 
After investigation, I identified the cause of the memory leak: defer() inside a loop. Verification confirmed this effectively resolves the calico-kube-controllers memory leak (due to certain security considerations, I cannot include my environment's verification details in this PR, but the bug is quite obvious).

Therefore, the summary of this PR is as follows:
- fix kube-controllers runconfig watch loop so each recreated watcher is stopped immediately and the active watcher is cleaned up on shutdown
- prevents deferred Stop calls from piling up and leaking watcher resources when the config watch is recreated

## Related issues/PRs

This issue <https://github.com/projectcalico/calico/issues/10124> is most closely matching the observed problem. 
The heap data characteristics we obtained in our environment are highly similar to those described in this issue.

## Todos

- [ ] Tests  (please let me know if you want specific tests run)
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Fix kube-controllers watch handling to avoid leaking watchers when the configuration watch is recreated.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/kube-controllers/pkg/config/runconfig.go
+++ b/kube-controllers/pkg/config/runconfig.go
@@ -218,7 +218,6 @@ MAINLOOP:
 			time.Sleep(datastoreBackoff)
 			continue MAINLOOP
 		}
-		
 		for e := range w.ResultChan() {
 			switch e.Type {
 			case watch.Error:


fusidic:
seems like something's wrong with CI. Should I do something?

caseydavenport:
@fusidic yeah, looks like a static-checks failure. You should be able to fix it by running `make fix-changed` and committing the results.

marvin-tigera:

fusidic:
Hi @caseydavenport 

I’ve fixed the preflight `check-dirty` issue.  
Everything from my side should now be ready.

Could you please approve the pending CI workflows so the checks can run?  
Thanks a lot!

fusidic:
Hi @caseydavenport 👋,

I’ve fixed the static check issues and successfully passed the pre-flight checks. However, the following jobs failed after that:
E2E tests: The BGP peer connection failed with a timeout.
CNI Plugin tests: Windows Containerd plugin tests are failing.

Since my changes were focused on runconfig.go, I don’t think they are related to these failures. Could you help investigate or let me know if I should take further steps?

Thanks for your help!

caseydavenport:
Yeah, those are almost certainly unrelated to this. Thanks!

PR #11443, opened 2025-11-25
ebpf - Split policy jump map

sridhartigera:
## Description

This PR splits the jump maps used for policy in ebpf dataplane into 2, similar to program maps.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/bpf.h
+++ b/felix/bpf-gpl/bpf.h
@@ -108,7 +108,7 @@
 #if CALI_F_INGRESS
 #define CALI_HOOK_INGRESS
 #else
-#define CALI_F_HOOK_EGRESS
+#define CALI_HOOK_EGRESS
 #endif
 #else
 // For WEPs, the policy direction is opposite to the tc hook that the
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -319,7 +319,6 @@ type bpfEndpointManager struct {
 	jumpMapAllocIngress *jumpMapAlloc
 	jumpMapAllocEgress  *jumpMapAlloc
 	xdpJumpMapAlloc     *jumpMapAlloc
-	policyDefaultObj    *libbpf.Obj
 	policyTcAllowFDs    [2]bpf.ProgFD
 	policyTcDenyFDs     [2]bpf.ProgFD
 
@@ -1619,6 +1618,8 @@ func (m *bpfEndpointManager) syncIfaceProperties() error {
 	return nil
 }
 
+// loadDefaultPolicies loads the default allow and deny policy programs for the given hook
+// and not policy direction.
 func (m *bpfEndpointManager) loadDefaultPolicies(hk hook.Hook) error {
 	file := path.Join(bpfdefs.ObjectDir, fmt.Sprintf("policy_default_%s.o", hk))
 	obj, err := libbpf.OpenObject(file)
@@ -1656,8 +1657,6 @@ func (m *bpfEndpointManager) loadDefaultPolicies(hk hook.Hook) error {
 	if err := obj.Load(); err != nil {
 		return fmt.Errorf("default policies: %w", err)
 	}
-
-	m.policyDefaultObj = obj
 
 	fd, err := obj.ProgramFD("calico_tc_deny")
 	if err != nil {
--- a/felix/dataplane/linux/bpf_ep_mgr_test.go
+++ b/felix/dataplane/linux/bpf_ep_mgr_test.go
@@ -1987,17 +1987,17 @@ var _ = Describe("BPF Endpoint Manager", func() {
 			val123 := ifstate.ValueFromBytes([]byte(ifStateMap.Contents[string(key123)]))
 			val124 := ifstate.ValueFromBytes([]byte(ifStateMap.Contents[string(key124)]))
 
-			tcIDsSeenFH := set.New[int]()
-			tcIDsSeenTH := set.New[int]()
+			tcIDsSeenIngress := set.New[int]()
+			tcIDsSeenEgress := set.New[int]()
 			for _, v := range []ifstate.Value{val123, val124} {
 				Expect(v.XDPPolicyV4()).To(Equal(-1), "WEPs shouldn't get XDP IDs")
 				Expect(v.IngressPolicyV4()).NotTo(Equal(-1), "WEPs should have ingress pol")
-				Expect(tcIDsSeenFH.Contains(v.IngressPolicyV4())).To(BeFalse(), "Saw same jump map ID more than once")
-				tcIDsSeenFH.Add(v.IngressPolicyV4())
+				Expect(tcIDsSeenIngress.Contains(v.IngressPolicyV4())).To(BeFalse(), "Saw same jump map ID more than once")
+				tcIDsSeenIngress.Add(v.IngressPolicyV4())
 
 				Expect(v.EgressPolicyV4()).NotTo(Equal(-1), "WEPs should have egress pol")
-				Expect(tcIDsSeenTH.Contains(v.EgressPolicyV4())).To(BeFalse(), "Saw same jump map ID more than once")
-				tcIDsSeenTH.Add(v.EgressPolicyV4())
+				Expect(tcIDsSeenEgress.Contains(v.EgressPolicyV4())).To(BeFalse(), "Saw same jump map ID more than once")
+				tcIDsSeenEgress.Add(v.EgressPolicyV4())
 
 				Expect(v.XDPPolicyV6()).To(Equal(-1), "WEPs shouldn't get XDP IPv6 ID")
 				Expect(v.IngressPolicyV6()).To(Equal(-1), "WEPs shouldn't get IPv6 ingress pol")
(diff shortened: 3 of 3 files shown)


felix/bpf-gpl/jump.h
```diff
@@ -120,7 +120,19 @@ CALI_MAP_V1(cali_jump_prog_map, BPF_MAP_TYPE_PROG_ARRAY, __u32, __u32, 2400, 0)
 	bpf_tail_call((ctx)->xdp, &cali_jump_prog_map, (ctx)->xdp_globals->jumps[PROG_INDEX_POLICY])
 #else /* CALI_F_XDP */
 
-#define cali_jump_prog_map map_symbol(cali_jump, 3)
+#if CALI_F_HEP || CALI_F_PREAMBLE || CALI_F_DEF_POLICY
+#if CALI_F_INGRESS
+#define cali_jump_prog_map map_symbol(cali_jump_fh, 2)
```
fasaxc: Seems backwards: if it's a HEP and it's an ingress program (policy and tc direction are both aligned) then that's the "to host" direction, but looks like you'll use `fh`

fasaxc: CALI_F_INGRESS is defined to be the policy direction.  Perhaps all of this logic would be clearer if we split these flags so that we no longer have any ambiguity:
```
CALI_F_POLICY_INGRESS / EGRESS
CALI_F_PROG_INGRESS / EGRESS
```

felix/bpf/bpf_syscall.h
```diff
@@ -48,55 +51,22 @@ void bpf_attr_setup_obj_pin(union bpf_attr *attr, char *path, __u32 fd, __u32 fl
...
-	   }
-   }
+	DECLARE_LIBBPF_OPTS(bpf_prog_load_opts, opts,
+		.log_level = log_level,
+		.log_size = log_size,
+		.log_buf = log_buf,
+		.kern_version = 0,
+		.expected_attach_type = attach_type,
+	);
+	int fd = bpf_prog_load(prog_type, name, license, insns, insn_count, &opts);
```
fasaxc, line 77: does that function do the sanitization that we've lost?

felix/bpf-gpl/jump.h
```diff
@@ -120,7 +120,19 @@ CALI_MAP_V1(cali_jump_prog_map, BPF_MAP_TYPE_PROG_ARRAY, __u32, __u32, 2400, 0)
 	bpf_tail_call((ctx)->xdp, &cali_jump_prog_map, (ctx)->xdp_globals->jumps[PROG_INDEX_POLICY])
 #else /* CALI_F_XDP */
 
-#define cali_jump_prog_map map_symbol(cali_jump, 3)
+#if CALI_F_HEP || CALI_F_PREAMBLE || CALI_F_DEF_POLICY
+#if CALI_F_INGRESS
```
fasaxc: How about moving the nested conditional to bpf.h and using it to #define some new constants:

```
#if CALI_F_HEP || CALI_F_PREAMBLE || CALI_F_DEF_POLICY
// For HEPs policy direction (CALI_F_INGRESS) matches attachment direction.
#if CALI_F_INGRESS
#define CALI_HOOK_INGRESS
#else
#define CALI_HOOK_EGRESS
#endif
#else
// For WEPs, the policy direction is opposite to the tc hook that the
// program is attached to.
#if CALI_F_INGRESS
#define CALI_HOOK_EGRESS
#else
#define CALI_HOOK_INGRESS
#endif
#endif
```
Then here, you'd just need this, which might be less prone to misunderstanding:
```
#ifdef CALI_HOOK_INGRESS
#define cali_jump_prog_map map_symbol(cali_jump_ing, 2)
#else
#define cali_jump_prog_map map_symbol(cali_jump_egr, 2)
#endif
```

felix/bpf/maps/maps.go
```diff
@@ -729,7 +729,7 @@ func (b *PinnedMap) EnsureExists() error {
 	case strings.HasPrefix(b.Name, "cali_"):
 		objName = "common_map_stub.o"
 	case strings.HasPrefix(b.Name, "cali_progs_map_fh"):
-		objName = "common_map_stub_ing.o"
+		objName = "common_map_stub_fh.o"
```
fasaxc: _fh naming still used here?  Is Copilot right about needing something for the other direction?

felix/bpf/ut/attach_test.go
```diff
@@ -271,13 +271,14 @@ func runAttachTest(t *testing.T, ipv6Enabled bool) {
 
 		}
 
-		pm := jumpMapDump(commonMaps.JumpMap)
-		Expect(pm).To(HaveKey(hostep1State.IngressPolicyV4()))
-		Expect(pm).To(HaveKey(hostep1State.EgressPolicyV4()))
+		pmFH := jumpMapDump(commonMaps.JumpMaps[hook.Ingress])
```
fasaxc: FH/TH naming, please can you do a pass and make it all consistent?

felix/fv/infrastructure/felix.go
```diff
@@ -697,7 +697,7 @@ func (f *Felix) BPFNumPolProgramsByEntryPoint(entryPointIdx int) (contiguous, to
 		k := polprog.SubProgramJumpIdx(entryPointIdx, i, jump.TCMaxEntryPoints)
 		out, err := f.ExecOutput(
 			"bpftool", "map", "lookup",
-			"pinned", "/sys/fs/bpf/tc/globals/cali_jump3",
+			"pinned", "/sys/fs/bpf/tc/globals/cali_jump_egr2",
```
fasaxc: Don't you need to count the ingress ones too?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -315,11 +316,12 @@ type bpfEndpointManager struct {
 	legacyCleanUp           bool
 	hostIfaceTrees          bpfIfaceTrees
 
-	jumpMapAlloc     *jumpMapAlloc
-	xdpJumpMapAlloc  *jumpMapAlloc
-	policyDefaultObj *libbpf.Obj
-	policyTcAllowFD  bpf.ProgFD
-	policyTcDenyFD   bpf.ProgFD
+	jumpMapAllocFromHEP *jumpMapAlloc
```
fasaxc: Let's use the consistent ingress/egress naming

felix/dataplane/linux/bpf_ep_mgr_test.go
```diff
@@ -444,8 +446,10 @@ var _ = Describe("BPF Endpoint Manager", func() {
 		commonMaps.ProgramsMaps = append(commonMaps.ProgramsMaps, mock.NewMockMap(progsParamsIng))
 		commonMaps.ProgramsMaps = append(commonMaps.ProgramsMaps, mock.NewMockMap(progsParamsEg))
 		commonMaps.XDPProgramsMap = mock.NewMockMap(progsParamsIng)
-		jumpMap = mock.NewMockMap(progsParamsIng)
-		commonMaps.JumpMap = jumpMap
+		jumpMapFH = mock.NewMockMap(progsParamsIng)
```
fasaxc: FH/TH

felix/bpf/maps/maps.go
sridhartigera: The one for the other direction in common_map_stub.o, which we already have.

felix/fv/infrastructure/felix.go
sridhartigera: Fixed this.

felix/bpf-gpl/bpf.h
```diff
@@ -101,6 +103,23 @@
 #define CALI_F_CGROUP	(((CALI_COMPILE_FLAGS) & CALI_CGROUP) != 0)
 #define CALI_F_DSR	((CALI_COMPILE_FLAGS & CALI_TC_DSR) != 0)
 
+#if CALI_F_HEP || CALI_F_PREAMBLE || CALI_F_DEF_POLICY
+// For HEPs policy direction (CALI_F_INGRESS) matches attachment direction.
+#if CALI_F_INGRESS
+#define CALI_HOOK_INGRESS
+#else
+#define CALI_F_HOOK_EGRESS
```
fasaxc: Inconsistent F in this one

felix/dataplane/linux/bpf_ep_mgr_test.go
```diff
@@ -1976,16 +1987,17 @@ var _ = Describe("BPF Endpoint Manager", func() {
 			val123 := ifstate.ValueFromBytes([]byte(ifStateMap.Contents[string(key123)]))
 			val124 := ifstate.ValueFromBytes([]byte(ifStateMap.Contents[string(key124)]))
 
-			tcIDsSeen := set.New[int]()
+			tcIDsSeenFH := set.New[int]()
```
fasaxc: Old naming in this file

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -315,11 +316,12 @@ type bpfEndpointManager struct {
...
 
-	jumpMapAlloc     *jumpMapAlloc
-	xdpJumpMapAlloc  *jumpMapAlloc
-	policyDefaultObj *libbpf.Obj
-	policyTcAllowFD  bpf.ProgFD
-	policyTcDenyFD   bpf.ProgFD
+	jumpMapAllocIngress *jumpMapAlloc
+	jumpMapAllocEgress  *jumpMapAlloc
+	xdpJumpMapAlloc     *jumpMapAlloc
+	policyDefaultObj    *libbpf.Obj
```
fasaxc: Looks like this field gets clobbered each time we load one of the programs

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1591,8 +1619,8 @@ func (m *bpfEndpointManager) syncIfaceProperties() error {
 	return nil
 }
 
-func (m *bpfEndpointManager) loadDefaultPolicies() error {
-	file := path.Join(bpfdefs.ObjectDir, "policy_default.o")
+func (m *bpfEndpointManager) loadDefaultPolicies(hk hook.Hook) error {
```
fasaxc, line 1623: The name of the default policy obj file is based on the policy direction, but this function takes a hook direction. I've forgotten exactly what these objects are for, are they only used for HEPs or are we deliberately only naming them by hook name?

If we only want them named by hook then you should comment that in the BPF code and here since the way the flags is calculated goes via the policy direction, which adds another layer to the confusion!

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: Looks like this field isn't used. Removing it.

felix/dataplane/linux/bpf_ep_mgr.go
sridhartigera: These are named as per hook direction similar to the preamble programs.

fasaxc (approved):

PR #11454, opened 2025-11-27
Verify that WrapPayloadWithEnvelope is exhaustive

fasaxc:
## Description

The `WrapPayloadWithEnvelope` function is supposed to cover all `ToDataplane` payload types, but it's easy to forget to update it when adding one (especially when merging code to enterprise, where there are more types).

Add a reflection-based UT that checks that all possible types are covered.  I also experimented with replacing the function with a reflection-based one, but that was 6x slower and this is close to the critical path in some cases.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/external/ext_dataplane_test.go
+++ b/felix/dataplane/external/ext_dataplane_test.go
@@ -1,0 +1,97 @@
+// Copyright (c) 2025 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package extdataplane
+
+import (
+	"reflect"
+	"testing"
+
+	"google.golang.org/protobuf/reflect/protoregistry"
+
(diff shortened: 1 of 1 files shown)


felix/dataplane/external/ext_dataplane_test.go
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) 2025 Tigera, Inc. All rights reserved.
```
nelljerram, line 1: 2026

felix/dataplane/external/ext_dataplane_test.go
```diff
@@ -0,0 +1,97 @@
...
+	t.Log("Number of payload types:", payload.Fields().Len())
+	for i := 0; i < payload.Fields().Len(); i++ {
+		field := payload.Fields().Get(i)
+		msgDesc := field.Message()
+		msgType, err := protoregistry.GlobalTypes.FindMessageByName(msgDesc.FullName())
+		if err != nil {
+			t.Fatalf("Failed to find message type for %v: %v", msgDesc.FullName(), err)
+		}
+		msg := msgType.New().Interface()
+		enveloped, err := WrapPayloadWithEnvelope(msg, 42)
```
aaaaaaaalex, line 61: Are the hardcoded numbers here meaningful or foobar outside of the scope of the test?

fasaxc: They're out of scope.  The prod calling code puts a unique seq number on each message, but here I just need to pass something.

felix/dataplane/external/ext_dataplane_test.go
fasaxc: This file hasn't been touched since 2025!

PR #11463, opened 2025-11-28
CI: cache Go's build cache

fasaxc:
## Description

Previously, we tried to save/restore the go build cache to  Semaphore's cache.  however, this never worked due to go's  use of mtime comparison.

On branch builds, cache the working copy using tar so that we save mtimes.  Then, for the "build" jobs, save off the go build cache for each group of jobs as a second tarfile. Different components build with different go options (CGO, race detector, etc) so using a shared single cache is not optimal.

On all builds, restore the working copy cache and then do a git fetch/checkout to get the intended commit. This is fast and it maximizes the effectiveness of the build cache.

On PR builds only, restore the build cache from the target branch, this makes PR builds faster, saving 1.5-6 minutes per job. Skip restoring the cache for jobs that don't build anything (such as felix FV jobs).

Split up felix's build/UT jobs.  The overhead of doing that is now much smaller and it shortens the time before we start running felix fv tests (and hence the overall time to do a CI run).
  
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/.semaphore/semaphore-scheduled-builds.yml
+++ b/.semaphore/semaphore-scheduled-builds.yml
@@ -25,8 +25,8 @@ global_job_config:
       value: calico-transient-build-artifacts-europe-west3
   prologue:
     commands:
-      - export GCS_BUILD_CACHE_BUCKET=calico-transient-build-artifacts-europe-west3
-      - export GCS_CACHE_DIR=gs://$GCS_BUILD_CACHE_BUCKET/cache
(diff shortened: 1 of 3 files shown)


aaaaaaaalex (approved):
LGTM 🚀

PR #11470, opened 2025-12-02
fix url parse

gojoy:
## Description
after go 1.24.8. `url.Parse` enforce stricter parsing of bracketed IPv6 hostnames, When the URL like https://[10.250.0.1]:443, the operation fails. Therefore, in the kubeconfig file generated by install-cni, the square brackets around KUBERNETES_SERVICE_HOST should be removed.

> 2025-12-02 17:43:33.194 [ERROR][279844] cni-plugin/plugin.go 593: Final result of CNI DEL was an
error. error=host must be a URL or a host:port pair: "https://[10.250.0.1]:443"

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix parsing of URLs in CNI configuration. 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/cni-plugin/pkg/install/install.go
+++ b/cni-plugin/pkg/install/install.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"runtime"
 	"strconv"
@@ -478,7 +479,7 @@ kind: Config
 clusters:
 - name: local
   cluster:
-    server: __KUBERNETES_SERVICE_PROTOCOL__://[__KUBERNETES_SERVICE_HOST__]:__KUBERNETES_SERVICE_PORT__
+    server: __KUBERNETES_SERVICE_PROTOCOL__://__KUBERNETES_HOST_PORT__
     __TLS_CFG__
 users:
 - name: calico
@@ -502,8 +503,7 @@ current-context: calico-context`
 	}
 	data = strings.Replace(data, "TOKEN", tu.Token, 1)
 	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_PROTOCOL__", getEnv("KUBERNETES_SERVICE_PROTOCOL", "https"))
-	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_HOST__", getEnv("KUBERNETES_SERVICE_HOST", ""))
-	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_PORT__", getEnv("KUBERNETES_SERVICE_PORT", ""))
+	data = strings.ReplaceAll(data, "__KUBERNETES_HOST_PORT__", net.JoinHostPort(getEnv("KUBERNETES_SERVICE_HOST", ""), getEnv("KUBERNETES_SERVICE_PORT", "")))
 
 	skipTLSVerify := os.Getenv("SKIP_TLS_VERIFY")
 	if skipTLSVerify == "true" {


caseydavenport:
@gojoy does this mean we need to re-insert the brackets for v6 addresses?

gojoy:
> @gojoy does this mean we need to re-insert the brackets for v6 addresses?

Thank you for the reminder. I believe specifying the Host in the REST configuration is a good practice, as the client-go automatically handles IPv6 addresses through URL parsing. Refer to [writeKubeconfig](https://github.com/projectcalico/calico/blob/master/node/pkg/cni/token_watch.go#L288)

cni-plugin/pkg/install/install.go
```diff
@@ -501,9 +501,7 @@ current-context: calico-context`
 		logrus.WithError(err).Fatal("Unable to create token for CNI kubeconfig")
 	}
 	data = strings.Replace(data, "TOKEN", tu.Token, 1)
-	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_PROTOCOL__", getEnv("KUBERNETES_SERVICE_PROTOCOL", "https"))
-	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_HOST__", getEnv("KUBERNETES_SERVICE_HOST", ""))
```
caseydavenport, line 505: I think we do still want the replace lines here - it's possible that users have their own CNI configurations relying on this replacement (even if we don't use them in our default config)

gojoy: use net.JoinHostPort to handle both  IPv6 and IPv4 addresses

gojoy:
Would you mind reviewing it once more? @caseydavenport

cni-plugin/pkg/install/install.go
```diff
@@ -502,8 +503,7 @@ current-context: calico-context`
 	}
 	data = strings.Replace(data, "TOKEN", tu.Token, 1)
 	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_PROTOCOL__", getEnv("KUBERNETES_SERVICE_PROTOCOL", "https"))
-	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_HOST__", getEnv("KUBERNETES_SERVICE_HOST", ""))
-	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_PORT__", getEnv("KUBERNETES_SERVICE_PORT", ""))
+	data = strings.ReplaceAll(data, "__KUBERNETES_HOST_PORT__", net.JoinHostPort(getEnv("KUBERNETES_SERVICE_HOST", ""), getEnv("KUBERNETES_SERVICE_PORT", "")))
```
caseydavenport, line 506:
```suggestion
	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_HOST__", getEnv("KUBERNETES_SERVICE_HOST", ""))
	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE_PORT__", getEnv("KUBERNETES_SERVICE_PORT", ""))
	data = strings.ReplaceAll(data, "__KUBERNETES_SERVICE__", net.JoinHostPort(getEnv("KUBERNETES_SERVICE_HOST", ""), getEnv("KUBERNETES_SERVICE_PORT", "")))
```

I think we need to leave the existing replace statements - some users provide their own CNI configuration templates and may be relying on this replacement. Just because they aren't used by the default config template doesn't mean they might not be used elsewhere. 

Let's also call it `KUBERNETES_SERVICE` since HOST_PORT has a meaning already in k8s networking.

caseydavenport:
@gojoy are you still working on this one? I just had one outstanding comment I think (and some conflicts to resolve)

gojoy:
> @gojoy are you still working on this one? I just had one outstanding comment I think (and some conflicts to resolve)

Yes, I’ve already rolled out this feature on our internal Kubernetes (K8s) cluster with thousands of nodes last month, and it’s working very well.

caseydavenport:
This one has actually be fixed it looks like: https://github.com/projectcalico/calico/pull/11713

Thanks as always for the contribution though, @gojoy !

PR #11491, opened 2025-12-04
[v3.30] Fix PATH env var in calico-node Windows image

coutinhop:
## Cherry-pick history
- Pick onto **release-v3.30**: projectcalico/calico#11489
The PATH env var in Windows images doesn't seem to be populated with sane values (powershell, system32, etc). Manually add those to the calico-node Windows image.

## Release Note

```release-note
TBD
```


song-jiang (approved):
LGTM

PR #11493, opened 2025-12-04
Add Policy name migration controller

caseydavenport:
## Description

Migrate stored policies so underlying storage identifies them using the same name as the v3 API. 

This is a follow-on to recent work that removes the tier prefixing requirement, and uses the real v3 API object names throughout Calico.

As a result of that work, any new policies created will be stored using their v3 object name. So, we must also write translation code for any pre-existing policies to ensure we are using a consistent naming scheme for the underlying data across new and old policies. 

Notes:
- This does perform a create / delete, and so object UIDs will change.
- This controller waits for calico/node to be upgraded before making the switch. This is important so that older calico/nodes don't receive policy updates they don't understand. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add controller to migrate underlying policy storage to use v3 API names.
```

## Reminder for the reviewer

Other optional labels:


kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator.go
```diff
@@ -0,0 +1,358 @@
...
+		return true // Default to true if Tier field is not found.
+	}
+
+	tier, ok := tierField.Interface().(string)
+	if !ok {
+		logCtx.Warn("Tier field is not a string")
+		return true // Default to true if Tier field is not a string.
+	}
+	return tier == "" || tier == "default"
+}
```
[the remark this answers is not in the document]
caseydavenport: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/projectcalico/calico/pull/11493#discussion_r2590319453)

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator.go
```diff
@@ -0,0 +1,381 @@
...
+			"namespace": p.GetNamespace(),
+			"v3name":    p.GetName(),
+			"kind":      p.GetObjectKind().GroupVersionKind().Kind,
+		})
+		logCtx.Debug("Processing policy for potential migration")
+
+		// The name in the Key is the actual v1 ID used in the datastore, whereas the ObjectMeta.Name is the v3
+		// object name. If they differ, we need to correct the underlying datastore entry to align with the v3 naming.
+		if !needsMigration(p, k) {
+			logCtx.Debug("No migration needed")
```
MichalFupso, line 235: If migration is not needed, should the key be removed from pendingWork?

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator.go
```diff
@@ -0,0 +1,381 @@
...
+			return
+		case status := <-c.statusUpdates:
+			c.status = status
+			logrus.Infof("Syncer status updated: %s", status.String())
+			c.kick()
+		case updates := <-c.updates:
+			logEntry := logrus.WithFields(logrus.Fields{"controller": "PolicyMigrator"})
+			utils.ProcessBatch(c.updates, updates, c.processUpdates, logEntry)
+			c.kick()
+		case <-c.doWork:
```
MichalFupso, line 165: I think it's possible to have items "stuck" in pending work. The channel is kicked on update, if during the processPendingWork we fail to migrate policy name and we receive no more updates after that we never try to migrate them again. We could add a timer (in other controllers we use 5 minutes) and trigger doWork periodically.

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator.go
```diff
@@ -0,0 +1,381 @@
...
+// the naming mismatch.
+//
+// The migration process involves:
+//  1. Listing all network policies in the datastore.
+//  2. Identifying policies where the v1 name (in the datastore Key) differs from the v3 name (in the ObjectMeta).
+//  3. For each mismatched policy:
+//     a. Create a new policy entry in the datastore with the correct v3 name.
+//     b. Delete the old policy entry with the v1 name.
+//
+// This controller runs until all existing policies have been successfully migrated.
```
MichalFupso: nit: Please update the comment to make it more clear the controller does not exit after migration, but runs continuously

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator.go
```diff
@@ -0,0 +1,381 @@
...
+func NewMigratorController(ctx context.Context, cs kubernetes.Interface, cli clientv3.Interface) controller.Controller {
+	// Kinds to register with on the syncer API.
+	resourceTypes := []watchersyncer.ResourceType{
```
MichalFupso: Other controllers share a syncer among them, I am assuming you're not using the shared syncer as only this controller needs to access policies?

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator.go
```diff
@@ -0,0 +1,381 @@
...
+			ListInterface: model.ResourceListOptions{Kind: v3.KindStagedGlobalNetworkPolicy},
+		},
+	}
+
+	type accessor interface {
+		Backend() bapi.Client
+	}
+
+	// Read the namespace from the service account file to determine the namespace we're running in.
+	namespace, err := os.ReadFile(winutils.GetHostPath("/var/run/secrets/kubernetes.io/serviceaccount/namespace"))
```
MichalFupso, line 48: Is it possible with manifest install to have kube-controllers and calico-node running in different namespaces?

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator_test.go
```diff
@@ -0,0 +1,559 @@
...
+				if kerrors.IsNotFound(err) {
+					return nil
+				}
+				return err
+			}
+			return fmt.Errorf("expected not to find old key for mismatched policy")
+		}, 5*time.Second, 1*time.Second).Should(BeNil(), "expected not to find old key for mismatched policy")
+	})
+
+	It("sholuld update a GlobalNetworkPolicy name correctly", func() {
```
MichalFupso:
```suggestion
	It("should update a GlobalNetworkPolicy name correctly", func() {
```

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator_test.go
```diff
@@ -0,0 +1,559 @@
...
+
+		// Check that the other two policies are still present under their original names.
+		kvp = expectFound(bcli, matchingWithPrefix.Name, matchingWithPrefix.Namespace, v3.KindNetworkPolicy)
+		Expect(kvp.Value.(*v3.NetworkPolicy).Name).To(Equal(matchingWithPrefix.Name))
+
+		kvp = expectFound(bcli, matchingNoPrefix.Name, matchingNoPrefix.Namespace, v3.KindNetworkPolicy)
+		Expect(kvp.Value.(*v3.NetworkPolicy).Name).To(Equal(matchingNoPrefix.Name))
+	})
+
+	It("sholuld update a GlobalNetworkPolicy name correctly", func() {
```
MichalFupso:
```suggestion
	It("should update a GlobalNetworkPolicy name correctly", func() {
```

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator_test.go
```diff
@@ -0,0 +1,559 @@
+// Copyright (c) 2025 Tigera, Inc. All rights reserved.
```
MichalFupso, line 1: The behavior is the same for StagedGlobalNetworkPolicy as for other policies, but might still be worth it adding a small test for the kind

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator.go
caseydavenport: Good catch!

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator.go
caseydavenport: Yeah, the other syncer is entirely for IPAM / node data and is scoped within. Might still make sense for me to share it here, let me see how easy it would be to do.

kube-controllers/pkg/controllers/networkpolicy/policy_name_migrator.go
caseydavenport: Theoretically yes, but I don't think it's something we support and I don't think we need to account for it here.

PR #11496, opened 2025-12-04
Include policy information in Log prefix

mazdakn:
## Description

Allow users to set the following specifier in `LogPrefix` in `FelixConfiguration` to include extra policy information in system logs:

```
%t: Tier name where policy is placed.
%k: Policy or profile kind. Short names are used. For example, gnp for GlobalNetworkPolicy.
%n: Policy name.
%p: Namespaced policy name. Either namespace/name for namespaced kinds or just name for non namespaced ones.
```

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Define specifiers to allow users include policy details in system logs generated from rules with LOG action.
%t: Tier name where policy is placed.
%k: Policy or profile kind. Short names are used. For example, gnp for GlobalNetworkPolicy.
%n: Policy name.
%p: Namespaced policy name. Either namespace/name for namespaced kinds or just name for non namespaced ones.
```

## Reminder for the reviewer

Other optional labels:


felix/rules/policy.go
```diff
@@ -711,6 +708,19 @@ func (r *DefaultRuleRenderer) CombineMatchAndActionsForProtoRule(
 	return finalRules
 }
 
+func (r *DefaultRuleRenderer) generateLogPrefix(id types.IDMaker) string {
+	logPrefix := ""
+	if len(r.LogPrefix) != 0 {
+		logPrefix = r.LogPrefix
+	}
```
nelljerram, line 733: Aren't the 4 previous lines equivalent to `logPrefix := r.LogPrefix` ?

nelljerram: Also is it intentional to lose the hardcoded default value "calico-packet" ?

mazdakn: Right! Basically, the first line should be `logPrefix := "calico-packet"` :-D

felix/rules/policy.go
```diff
@@ -711,6 +716,47 @@ func (r *DefaultRuleRenderer) CombineMatchAndActionsForProtoRule(
...
+
+	var kind, name, namespace string
+	switch v := id.(type) {
+	case *types.PolicyID:
+		kind = v.KindShortName()
+		name = v.Name
+		namespace = v.Namespace
+	case *types.ProfileID:
+		kind = "profile"
```
caseydavenport: I think we can still use a shortname for this if you'd like - e.g., "pro"

felix/rules/policy.go
```diff
@@ -711,6 +716,47 @@ func (r *DefaultRuleRenderer) CombineMatchAndActionsForProtoRule(
...
+		name = v.Name
+	default:
+		logrus.Warnf("Unrecognized resource type %T when generating log preix", id)
```
caseydavenport: should we return here (maybe return "unknown" or something recognizable?)

PR #11514, opened 2025-12-08
Rate limit LOG Action

mazdakn:
## Description

Rules with `LOG` action can generate overwhelming number of system logs. To limit the number of hits, this PR introduces two options to `FelixConfiguration` to allow users change the behavior.  The new configs are:
- `LogActionRateLimit`: the rate of hitting a log action. 
- `LogActionRateLimitBurst`: which is the burst of log generation when `LogActionRateLimit` is enabled.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Introduce logActionRateLimit and logActionRateLimitBurst in FelixConfiguration to rate limit the number of hitting a rule with LOG action.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/api/pkg/apis/projectcalico/v3/felixconfig.go
+++ b/api/pkg/apis/projectcalico/v3/felixconfig.go
@@ -332,13 +332,11 @@ type FelixConfigurationSpec struct {
 
 	// LogActionRateLimit sets the rate of hitting a Log action. The value must be in the format "N/unit",
 	// where N is a number and unit is one of: second, minute, hour, or day. For example: "10/second" or "100/hour".
-	// [Default: ""]
 	// +optional
 	// +kubebuilder:validation:Pattern=`^\d+/(?:second|minute|hour|day)$`
-	LogActionRateLimit string `json:"logActionRateLimit,omitempty"`
+	LogActionRateLimit *string `json:"logActionRateLimit,omitempty"`
 
 	// LogActionRateLimitBurst sets the rate limit burst of hitting a Log action when LogActionRateLimit is enabled.
-	// [Default: 5]
 	// +optional
 	// +kubebuilder:validation:Minimum=0
 	LogActionRateLimitBurst *int `json:"logActionRateLimitBurst,omitempty"`
--- a/felix/config/config_params.go
+++ b/felix/config/config_params.go
@@ -55,7 +55,7 @@ var (
 	IfaceParamRegexp         = regexp.MustCompile(`^[a-zA-Z0-9:._+-]{1,15}$`)
 	// Hostname  have to be valid ipv4, ipv6 or strings up to 64 characters.
 	HostAddressRegexp   = regexp.MustCompile(`^[a-zA-Z0-9:._+-]{1,64}$`)
-	LogActionRateRegexp = regexp.MustCompile(`^(\d+/(second|minute|hour|day))?$`)
+	LogActionRateRegexp = regexp.MustCompile(`^(\d+/(?:second|minute|hour|day))?$`)
 )
 
 // Source of a config value.  Values from higher-numbered sources override
--- a/api/pkg/apis/projectcalico/v3/zz_generated.deepcopy.go
+++ b/api/pkg/apis/projectcalico/v3/zz_generated.deepcopy.go
@@ -1234,6 +1234,11 @@ func (in *FelixConfigurationSpec) DeepCopyInto(out *FelixConfigurationSpec) {
 		*out = new(int)
 		**out = **in
 	}
+	if in.LogActionRateLimit != nil {
+		in, out := &in.LogActionRateLimit, &out.LogActionRateLimit
+		*out = new(string)
+		**out = **in
+	}
 	if in.LogActionRateLimitBurst != nil {
 		in, out := &in.LogActionRateLimitBurst, &out.LogActionRateLimitBurst
 		*out = new(int)
--- a/api/pkg/openapi/generated.openapi.go
+++ b/api/pkg/openapi/generated.openapi.go
@@ -2658,14 +2658,14 @@ func schema_pkg_apis_projectcalico_v3_FelixConfigurationSpec(ref common.Referenc
 					},
 					"logActionRateLimit": {
 						SchemaProps: spec.SchemaProps{
-							Description: "LogActionRateLimit sets the rate of hitting a Log action. The value must be in the format \"N/unit\", where N is a number and unit is one of: second, minute, hour, or day. For example: \"10/second\" or \"100/hour\". [Default: \"\"]",
+							Description: "LogActionRateLimit sets the rate of hitting a Log action. The value must be in the format \"N/unit\", where N is a number and unit is one of: second, minute, hour, or day. For example: \"10/second\" or \"100/hour\".",
 							Type:        []string{"string"},
 							Format:      "",
 						},
 					},
 					"logActionRateLimitBurst": {
 						SchemaProps: spec.SchemaProps{
-							Description: "LogActionRateLimitBurst sets the rate limit burst of hitting a Log action when LogActionRateLimit is enabled. [Default: 5]",
+							Description: "LogActionRateLimitBurst sets the rate limit burst of hitting a Log action when LogActionRateLimit is enabled.",
 							Type:        []string{"integer"},
 							Format:      "int32",
 						},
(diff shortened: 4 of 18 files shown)


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -330,6 +330,16 @@ type FelixConfigurationSpec struct {
...
 
+	// LogActionRate sets the rate of hitting a Log action. The value must be in the format "N/unit",
+	// where N is a number and unit is one of: second, minute, hour, or day. For example: "10/second" or "100/hour".
+	// [Default: ""]
+	// +kubebuilder:validation:Pattern=`^(([0-9]+/(second|minute|hour|day)))?$`
+	LogActionRate string `json:"logActionRate,omitempty"`
+
+	// LogActionBurst sets the burst of hitting a Log action when LogActionRate is enabled. [Default: 5]
+	// +kubebuilder:validation:XValidation:rule="self >= 0", message="logActionBurst must be a non-negative integer"
+	LogActionBurst *int `json:"logActionBurst,omitempty"`
```
fasaxc: I think it's important to add Limit into the names:

- LogActionRateLimit
- LogActionRateLimitBurst

Avoid using CEL (XValidation) if there's a basic validation option because CRDs get "scored" on how much CEL they use and there's a limit on the score.  should be able to use `+kubebuilder:validation:Minimum` here.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -330,6 +330,16 @@ type FelixConfigurationSpec struct {
 	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9%: /_-])*$`
 	LogPrefix string `json:"logPrefix,omitempty"`
 
+	// LogActionRate sets the rate of hitting a Log action. The value must be in the format "N/unit",
+	// where N is a number and unit is one of: second, minute, hour, or day. For example: "10/second" or "100/hour".
+	// [Default: ""]
+	// +kubebuilder:validation:Pattern=`^(([0-9]+/(second|minute|hour|day)))?$`
```
fasaxc: Bit clearer to use "optional" instead of allow empty string in the regex (if it works!)?
```suggestion
	// +optional    
	// +kubebuilder:validation:Pattern=`^\d+/(?:second|minute|hour|day)$`
```

felix/config/config_params.go
```diff
@@ -329,6 +329,8 @@ type Config struct {
 	IptablesMangleAllowAction   string `config:"oneof(ACCEPT,RETURN);ACCEPT;non-zero,die-on-fail"`
 	IptablesFilterDenyAction    string `config:"oneof(DROP,REJECT);DROP;non-zero,die-on-fail"`
 	LogPrefix                   string `config:"string;calico-packet"`
+	LogActionRate               string `config:"string;"`
```
fasaxc: You should add a custom validation type here too to guard against bad env var/config file settings
```suggestion
	LogActionRate               string `config:"log-rate;"`
```

Then below in `loadParams`, add
```
		case "log-rate":
			param = &RegexpParam{
				Regexp: LogRateRegexp,
				Msg:    "invalid log rate limit",
			}
```
(see other uses of RegexpParam for more examples)

api/pkg/apis/projectcalico/v3/felixconfig.go
mazdakn: fixed.

api/pkg/apis/projectcalico/v3/felixconfig.go
mazdakn: fixed.

felix/config/config_params.go
mazdakn: Thanks for the suggestion. Fixed.

mazdakn:
@fasaxc the test in `felix/fv/policy_test.go` is updated to check that limit matches are programmed correctly. I added this test for log prefix, and I'm just expanding it here.

Regarding bpf support, I think I should address it in another PR. It probably needs more work.

I addressed your other comments. PTAL.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -330,6 +330,19 @@ type FelixConfigurationSpec struct {
...
 
+	// LogActionRateLimit sets the rate of hitting a Log action. The value must be in the format "N/unit",
+	// where N is a number and unit is one of: second, minute, hour, or day. For example: "10/second" or "100/hour".
+	// [Default: ""]
+	// +optional
+	// +kubebuilder:validation:Pattern=`^\d+/(?:second|minute|hour|day)$`
+	LogActionRateLimit string `json:"logActionRateLimit,omitempty"`
+
+	// LogActionRateLimitBurst sets the rate limit burst of hitting a Log action when LogActionRateLimit is enabled.
+	// [Default: 5]
```
caseydavenport: do we need a kubebuilder default for this? Or do we actually not default this and should just remove this comment?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -330,6 +330,19 @@ type FelixConfigurationSpec struct {
 	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9%: /_-])*$`
 	LogPrefix string `json:"logPrefix,omitempty"`
 
+	// LogActionRateLimit sets the rate of hitting a Log action. The value must be in the format "N/unit",
+	// where N is a number and unit is one of: second, minute, hour, or day. For example: "10/second" or "100/hour".
+	// [Default: ""]
+	// +optional
+	// +kubebuilder:validation:Pattern=`^\d+/(?:second|minute|hour|day)$`
+	LogActionRateLimit string `json:"logActionRateLimit,omitempty"`
```
caseydavenport: Optional fields should conventionally be pointers, so this should probably be `*string`

caseydavenport: We haven't always followed this convention but I think it's a good idea to do so.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -330,6 +330,19 @@ type FelixConfigurationSpec struct {
 	// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9%: /_-])*$`
 	LogPrefix string `json:"logPrefix,omitempty"`
 
+	// LogActionRateLimit sets the rate of hitting a Log action. The value must be in the format "N/unit",
+	// where N is a number and unit is one of: second, minute, hour, or day. For example: "10/second" or "100/hour".
+	// [Default: ""]
+	// +optional
+	// +kubebuilder:validation:Pattern=`^\d+/(?:second|minute|hour|day)$`
```
caseydavenport, line 336: Is there a reason not to use a v1.Duration field / format here? I think that is what we use (and k8s in general uses) for similar fields.

caseydavenport (commented):
A couple of minor thoughts

api/pkg/apis/projectcalico/v3/felixconfig.go
mazdakn: It's default to 5 in Felix. This can be removed. 
Fixed.

api/pkg/apis/projectcalico/v3/felixconfig.go
mazdakn: Right. Updated.

api/pkg/apis/projectcalico/v3/felixconfig.go
mazdakn: Cause they are a bit different. `v1.Duration` accepts millisecond, second, minute and hour duration, while limit action supports only second, minute, hour, and day periods.

caseydavenport: Yeah, I suppose I was thinking we could use validation to limit and then translate between the API representation and the dataplane representation. If that's not too challenging, I think it would be a nicer experience to keep the API time fields using a consistent format.

mazdakn: I'll give it a shot!

mazdakn: @caseydavenport there is a major difference between the rate limit and duration type. A 5s value for a duration literally means 5 seconds, but in case of rate limit 5/second means 5 actions are allowed per one second. 

I tried to use Duration for this new field, but it's inherently a different type. I don't think we should use metav1.Duration type here.

caseydavenport: ohhh right yeah, I totally misread this. :facepalm:

caseydavenport (approved):

PR #11541, opened 2025-12-12
Fix construction of chain names

### LINKED ISSUE #10811: [Bug] Network policy host endpoint chain name collision with cali* interface grouping
## Expected Behavior
Calico should ensure that chain names created for Network Policy host endpoint grouping do not conflict with chain names created by other internal mechanisms, such as the per-interface-prefix grouping used for `cali* `workload interfaces.

## Current Behavior
We have identified an issue in Calico where Global Network Policy (GNP) host endpoint grouping chains can conflict with interface prefix-based chains used by Calico’s default `cali*` interface grouping. When host endpoint interfaces have names starting with a hex character [a-f] and two or more share the same first character, Calico’s GNP chain creation logic generates a chain like: `cali-set-endpoint-mark-[a-f]`. This chain name collides with chains generated by the per-interface-prefix grouping mechanism for `cali*` interfaces, which also uses the same `cali-set-endpoint-mark-[a-f]` naming convention.

**Impact**
1. Incorrect packet marking → DNAT skipped before policy enforcement
2. Traffic may bypass intended policy processing paths.
3. Current workaround requires renaming host endpoints to avoid [a-f] start, breaking backward compatibility.

## Possible Solution
Current workaround requires renaming host endpoints to avoid [a-f] start but that breakes backward compatibility.

1. Create two dummy interfaces on the Node
```shell
sudo ip link add name adummy0 type dummy
sudo ip link add name adummy1 type dummy
```

2. Create a HostEndpoints and Service object
```shell
kubectl apply -f - <<EOF
apiVersion: crd.projectcalico.org/v1
kind: HostEndpoint
metadata:
  name: control-plane-dummy0
  labels:
    interface: adummy0
    node: foo
spec:
  interfaceName: adummy0
  node: foo
---
apiVersion: crd.projectcalico.org/v1
kind: HostEndpoint
metadata:
  name: control-plane-dummy1
  labels:
    interface: adummy1
    node: foo
spec:
  interfaceName: adummy1
  node: foo
---
apiVersion: v1
kind: Service
metadata:
  name: nginx
  namespace: test1
spec:
  selector:
    app: nginx
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
EOF
```

3. Create some workload Pods so that Calico creates chains
```shell
kubectl apply -f - <<EOF
apiVersion: apps/v1
kind: Deployment
metadata:
  name: troubleshooting-pod
  namespace: test1
spec:
  replicas: 10
  selector:
    matchLabels:
      fw: block
  template:
    metadata:
      labels:
        fw: block
    spec:
      nodeSelector:
        kubernetes.io/hostname: foo
      containers:
        - name: sleeper
          image: busybox:1.28
          command: ['sh', '-c', 'echo "Hello, Kubernetes!" && sleep 3600']
---
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: nginx
    run: nginx
  name: nginx
  namespace: test1
spec:
  nodeSelector:
    kubernetes.io/hostname: foo
  containers:
    - image: nginx:1.14.2
      imagePullPolicy: Always
      name: nginx
  restartPolicy: Always
EOF
```

4. Create a network policy and observer the chains being overwritten 
```shell
kubectl apply -f - <<EOF
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: testpol1
  namespace: test1
spec:
  egress:
    - ports:
        - port: 80
          protocol: TCP
      to:
        - podSelector:
            matchLabels:
              fw: block
  policyTypes:
    - Egress
EOF
```

## Context
- Occurs only when host endpoint interface names start with [a-f] and share the same first letter
- Does not occur when interface names start with non-hex letters (e.g., i, g, x), since those prefixes are not used by the per-interface grouping mechanism?

## Your Environment

* Calico version
v3.30.2
* Calico dataplane (bpf, nftables, iptables, windows etc.)
iptables
* Orchestrator version (e.g. kubernetes, openshift, etc.):
Kubernetes (kubeadm)
* Operating System and version:
Ubuntu 24.04

terror96:
Add support to inject an infix into the constructed chain name in `buildSingleDispatchChainTree` when creating chains in multiple patches.

## Description

There is an issue when running Calico in IPVS mode which can cause a harmful overlap with constructed chain names. This occurs when there are two or more interfaces starting with the prefix, say 'calia', and native interfaces starting with the letter 'a'.

The current implementation creates a chain called 'cali-set-endpoint-mark-a' for both groups of interfaces, which then leads to the situation that rules created for the latter group of interfaces will overwrite the rules created for the former group of interfaces. This is due to the fact that constructing chains is done in two passes.

## Related issues/PRs

Fixes #10811

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix potential HEP / WEP chain name conflicts in IPVS mode.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/rules/dispatch.go
+++ b/felix/rules/dispatch.go
@@ -103,6 +103,7 @@ func (r *DefaultRuleRenderer) WorkloadInterfaceAllowChains(
 			return r.Allow()
 		},
 		endRules,
+		"",
 	)
 	chains = append(chains, toChildChains...)
 	chains = append(chains, toRootChain)
@@ -315,6 +316,7 @@ func (r *DefaultRuleRenderer) interfaceNameDispatchChains(
 				return r.GoTo(EndpointChainName(pfx, name, r.maxNameLength))
 			},
 			fromEndRules,
+			"",
 		)
 		chains = append(chains, fromChildChains...)
 		chains = append(chains, fromRootChain)
@@ -333,6 +335,7 @@ func (r *DefaultRuleRenderer) interfaceNameDispatchChains(
 				return r.GoTo(EndpointChainName(pfx, name, r.maxNameLength))
 			},
 			toEndRules,
+			"",
 		)
 		chains = append(chains, toChildChains...)
 		chains = append(chains, toRootChain)
@@ -363,9 +366,18 @@ func (r *DefaultRuleRenderer) endpointMarkDispatchChains(
 
 	// The workload and host endpoint share the same root chain. We also need to put an non-cali mark rules at the end.
 	// Work out child chains and root rules for workload and host endpoint separately and merge them back together.
-	for _, names := range [][]string{wlNames, hepNames} {
+	for index, names := range [][]string{wlNames, hepNames} {
 		if len(names) > 0 {
 			commonPrefix, prefixes, prefixToNames := r.sortAndDivideEndpointNamesToPrefixTree(names)
+			var infix string
+			switch index {
+			case 0:
+				infix = "wep"
+			case 1:
+				infix = "hep"
+			default:
+				log.Panicf("no mapping for index %d to infix", index)
+			}
 
 			childChains, _, rootRules := r.buildSingleDispatchChains(
 				dispatchSetMarkEndpointChainName,
@@ -378,6 +390,7 @@ func (r *DefaultRuleRenderer) endpointMarkDispatchChains(
 					return r.GoTo(EndpointChainName(pfx, name, r.maxNameLength))
 				},
 				nil,
+				"-"+infix,
 			)
 
 			chains = append(chains, childChains...)
@@ -468,6 +481,7 @@ func (r *DefaultRuleRenderer) buildSingleDispatchChains(
 	getMatchForEndpoint func(name string) generictables.MatchCriteria,
 	getActionForEndpoint func(pfx, name string) generictables.Action,
 	endRules []generictables.Rule,
+	infix string,
 ) ([]*generictables.Chain, *generictables.Chain, []generictables.Rule) {
 	if r.NFTables && (endpointPfx == WorkloadFromEndpointPfx || endpointPfx == WorkloadToEndpointPfx) {
 		// Currently only supported for nftables workload endpoint dispatch.
@@ -486,6 +500,7 @@ func (r *DefaultRuleRenderer) buildSingleDispatchChains(
 		getMatchForEndpoint,
 		getActionForEndpoint,
 		endRules,
+		infix,
 	)
 }
 
@@ -500,6 +515,7 @@ func (r *DefaultRuleRenderer) buildSingleDispatchChainTree(
 	getMatchForEndpoint func(name string) generictables.MatchCriteria,
 	getActionForEndpoint func(pfx, name string) generictables.Action,
 	endRules []generictables.Rule,
+	infix string,
 ) ([]*generictables.Chain, *generictables.Chain, []generictables.Rule) {
 	childChains := make([]*generictables.Chain, 0)
 	rootRules := make([]generictables.Rule, 0)
@@ -518,7 +534,14 @@ func (r *DefaultRuleRenderer) buildSingleDispatchChainTree(
 			// More than one name, render a prefix match in the root chain...
 			nextChar := prefix[len(commonPrefix):]
 			ifaceMatch := prefix + r.wildcard
-			childChainName := chainName + "-" + nextChar
+			// Inject an infix into the child chain name. This fixes the chain
+			// naming collision that can occur when chains are constructed in
+			// more than a single pass. Without this, interfaces like "caliX..."
+			// (Calico workload interfaces) can end up in the same chain as
+			// interfaces exposed as host endpoints that start with the same
+			// character "X". In this case rules generated in the previous round
+			// could be overwritten by the new ones.
+			childChainName := fmt.Sprintf("%s%s-%s", chainName, infix, nextChar)
 			logCxt := logCxt.WithFields(log.Fields{
 				"childChainName": childChainName,
 				"ifaceMatch":     ifaceMatch,
--- a/felix/rules/dispatch_test.go
+++ b/felix/rules/dispatch_test.go
@@ -238,7 +238,7 @@ var _ = Describe("Dispatch chains", func() {
 						},
 					},
 					{
-						Name: "cali-set-endpoint-mark-2",
+						Name: "cali-set-endpoint-mark-wep-2",
 						Rules: []generictables.Rule{
 							inboundGotoRule("cali2333", "cali-sm-cali2333"),
 							inboundGotoRule("cali2444", "cali-sm-cali2444"),
@@ -248,7 +248,7 @@ var _ = Describe("Dispatch chains", func() {
 						Name: "cali-set-endpoint-mark",
 						Rules: []generictables.Rule{
 							inboundGotoRule("cali1234", "cali-sm-cali1234"),
-							inboundGotoRule("cali2+", "cali-set-endpoint-mark-2"),
+							inboundGotoRule("cali2+", "cali-set-endpoint-mark-wep-2"),
 							smUnknownEndpointDropRule("cali"),
 							smUnknownEndpointDropRule("tap"),
 							smNonCaliSetMarkRule,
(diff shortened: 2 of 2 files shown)


felix/rules/dispatch.go
```diff
@@ -468,6 +472,7 @@ func (r *DefaultRuleRenderer) buildSingleDispatchChains(
 	getMatchForEndpoint func(name string) generictables.MatchCriteria,
 	getActionForEndpoint func(pfx, name string) generictables.Action,
 	endRules []generictables.Rule,
+	passNumber int,
```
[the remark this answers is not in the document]
terror96: passNumber -> indexNumber

felix/rules/dispatch.go
```diff
@@ -519,6 +526,15 @@ func (r *DefaultRuleRenderer) buildSingleDispatchChainTree(
 			nextChar := prefix[len(commonPrefix):]
 			ifaceMatch := prefix + r.wildcard
 			childChainName := chainName + "-" + nextChar
+			// Extend the chain name with the pass number if greater than zero. This
+			// fixes the chain naming collision in case chains are constructed in
+			// more than a single pass. In that case, interfaces like caliX... and
+			// named like caliX... end up in the same chain than interfaces that start
+			// with the same character X. In this case rules generated in the previous
+			// round are overwritten by the new ones.
```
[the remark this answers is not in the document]
terror96: Valid comments. Modified the text.

felix/rules/dispatch.go
```diff
@@ -519,6 +526,17 @@ func (r *DefaultRuleRenderer) buildSingleDispatchChainTree(
...
+			// Extend the chain name with the index number if greater than zero.
+			// This fixes the chain naming collision that can occur when chains
+			// are constructed in more than a single pass. Without this,
+			// interfaces like "caliX..." (Calico workload interfaces) can end
+			// up in the same chain than interfaces that start with the same
+			// character "X.." if they share the same prefix. In this case rules
+			// generated in the previous round could be overwritten by the new
+			// ones.
+			if indexNumber > 0 {
+				childChainName += fmt.Sprintf("%d", indexNumber)
```
caseydavenport: I think this can still hit an issue if the conflicting prefix is of the form `nextChar` followed by `index` ? 

Probably make sense to put the index indicator before the hyphen to avoid that:

```
fmt.Sprintf("%s%s-%s", chainName, index, nextChar)
```

caseydavenport: Also, instead of passing an index integer it would probably be worthwhile to make this a string so that we can pass more human-readable "infix" values (e.g., "hep" and "wep") so that the resulting chain names have a bit more context. WDYT?

terror96: Thanks @caseydavenport for the feedback. I think your suggestions make a lot of sense, and it is better to be safe than sorry.

Also, it is indeed good to change `indexNumber` to `infix` (string). `buildSingleDispatchChains()` is called from a few other places within the file, and we can use an empty infix value (and keep the original `childChainName` in that case) for others calls than in `endpointMarkDispatchChains()` because other calls use a unique `chainName` and render rules as a single pass.

terror96: Pushed a new version. I also added an extra hyphen to infix to make the chain name a bit more readable e.g. `cali-set-endpoint-mark-wep-e` and `cali-set-endpoint-mark-hep-a`.

caseydavenport:
@terror96 looks like UTs are failing - you can run them with `make ut` locally in the Felix directory. 

```
Summarizing 2 Failures:03:56
[Fail] Dispatch chains workload rendering tests [It] interfaces sharing prefix 03:56
/go/src/github.com/projectcalico/calico/felix/rules/dispatch_test.go:11803:56
[Fail] Dispatch chains workload rendering tests [It] Multiple interfaces sharing multiple prefixes 
```

terror96:
> @terror96 looks like UTs are failing - you can run them with `make ut` locally in the Felix directory.
> 
> ```
> Summarizing 2 Failures:03:56
> [Fail] Dispatch chains workload rendering tests [It] interfaces sharing prefix 03:56
> /go/src/github.com/projectcalico/calico/felix/rules/dispatch_test.go:11803:56
> [Fail] Dispatch chains workload rendering tests [It] Multiple interfaces sharing multiple prefixes 
> ```

Thanks for the info. I was expecting that some of the tests might fail, and that would have been my next step to ask for some pointers and guidance. But you were already there... :)

terror96:
> > @terror96 looks like UTs are failing - you can run them with `make ut` locally in the Felix directory.
> > ```
> > Summarizing 2 Failures:03:56
> > [Fail] Dispatch chains workload rendering tests [It] interfaces sharing prefix 03:56
> > /go/src/github.com/projectcalico/calico/felix/rules/dispatch_test.go:11803:56
> > [Fail] Dispatch chains workload rendering tests [It] Multiple interfaces sharing multiple prefixes 
> > ```
> 
> Thanks for the info. I was expecting that some of the tests might fail, and that would have been my next step to ask for some pointers and guidance. But you were already there... :)

These test failures should now be fixed.

terror96:
These changes seem to trigger a failure in `staged_policy.go:128` due to "number of flow items does not match, expected 2, got 1". I'm wondering is there a dependency to a certain naming conversion of iptables chains that just got broken, or what is going on. Unfortunately, I do not understand well enough the test and what is going wrong here in order to fix the problem.

terror96:
> These changes seem to trigger a failure in `staged_policy.go:128` due to "number of flow items does not match, expected 2, got 1". I'm wondering is there a dependency to a certain naming conversion of iptables chains that just got broken, or what is going on. Unfortunately, I do not understand well enough the test and what is going wrong here in order to fix the problem.

I have been trying to replicate the test failure locally by running:

```
make e2e-test E2E_FOCUS="'Validate name, tier, action'"
```
but I get with and without my modifications the same end result:
```
...
Ran 3 of 802 Specs in 77.815 seconds
SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 799 Skipped
PASS
```
However, the next test "e2e-run-cnp-test" fails even on unmodified version:

```
    helper.go:127: Namespaces and Pods in network-policy-conformance-gryffindor, network-policy-conformance-slytherin, network-policy-conformance-hufflepuff, network-policy-conformance-ravenclaw, network-policy-conformance-forbidden-forrest namespaces are ready
    --- FAIL: TestConformance/CNPBaselineTierEgressInlineCIDRPeers (34.26s)
        suite.go:189: Applying base/baseline_tier/standard-egress-inline-cidr-rules.yaml
        apply.go:124: Creating default ClusterNetworkPolicy
        --- FAIL: TestConformance/CNPBaselineTierEgressInlineCIDRPeers/Should_support_a_'deny-egress'_rule_policy_for_egress-cidr-peer (15.56s)
            baseline-admin-network-policy-standard-egress-inline-cidr-rules.go:61: FAILED Command was [/agnhost connect --timeout=3s --protocol=tcp 192.168.110.134:80]
            baseline-admin-network-policy-standard-egress-inline-cidr-rules.go:61: Expected connection to fail from network-policy-conformance-gryffindor/harry-potter-1 to 192.168.110.134, but instead it successfully connected.
            baseline-admin-network-policy-standard-egress-inline-cidr-rules.go:63: 
                	Error Trace:	/go/pkg/mod/sigs.k8s.io/network-policy-api@v0.1.8-0.20251017092043-375c8a75a50a/conformance/tests/baseline-admin-network-policy-standard-egress-inline-cidr-rules.go:63
                	Error:      	Should be true
                	Test:       	TestConformance/CNPBaselineTierEgressInlineCIDRPeers/Should_support_a_'deny-egress'_rule_policy_for_egress-cidr-peer
        apply.go:132: Deleting default ClusterNetworkPolicy
```


The only difference to this PR is that I applied my changes on top of 970146e9c19a603789b09578fcdb64fd8758abbf (which was the newest version a few hours ago). I also tried to run the test on the version in this PR (i.e. not rebased yet), but I could not get e2e tests running:

```
make[1]: Entering directory '/home/ubuntu/tmp/calico'
bin/helm package ./charts/tigera-operator \
--destination ./bin/ \
--version origin-132-g2a4c875ac0f1 \
--app-version origin-132-g2a4c875ac0f1
walk.go:74: found symbolic link in path: /home/ubuntu/tmp/calico/charts/tigera-operator/crds/calico resolves to /home/ubuntu/tmp/calico/libcalico-go/config/crd. Contents of linked file included and used
Error: Invalid Semantic Version
```

EDIT: I dunno if there is a smarter way but I now also applied my changes on top of 354bb6284ab236b6c148fd605d41a38f3f94e3af (matches with this PR's base) and run 'make e2e-test' again. I got the same result: 'e2e-run-test' phase passes and 'e2e-run-cnp-test' fails.

terror96:
Rebased.

caseydavenport (approved):

caseydavenport:
e2e looks unrelated, thanks @terror96

terror96:
Thank you @caseydavenport and @fasaxc for your support with this PR.

Issue #6887, opened 2022-10-21
Helm "enable metrics" boolean


jcpunk:
## Expected Behavior
A number of common helm charts have a value you can set to enable metrics reporting.  Since calico has a metrics interface, having a clean way to enable it via helm would be nice.

## Current Behavior
You must manually patch the instance to enable metrics

## Possible Solution
Add a boolean to `values.yml` and the relevant templates to enable the metrics service on request.

## Steps to Reproduce (for bugs)

1. Install via helm
2. look for way to enable metrics as a parameter
3.
4.

## Context

A consistent way to enable metrics across a wide array of helm charts makes things easier.

## Your Environment

* Calico version: 3.24.2
* Orchestrator version (e.g. kubernetes, mesos, rkt): 1.25.2
* Operating System and version: CentOS Stream9
* Link to your project (optional):

jcpunk:
Neat, I didn't see those in there.  Can it be added to the documentation?

caseydavenport:
Yeah, the documentation isn't very clear at the moment. We need to do better.

The `installation` section contains all of the fields specified in the InstallationSpec here: https://projectcalico.docs.tigera.io/reference/installation/api#operator.tigera.io/v1.InstallationSpec

Maybe there's a way we can auto-generate the values.yaml doc to include all of those fields, but for now that link is effectively the documentation.

mgleung:
@jcpunk out of curiosity, which pages of the documentation do you normally look at for this? We have the fields in https://projectcalico.docs.tigera.io/reference/installation/api but it seems that it's not a great UX. Any feedback on how we can better arrange this information is welcome.

mgleung:
OK, sounds like the right path forward is to beef up the documentation in the README for now. If there are any other suggestions you have on how to make this easier to read going forward, please let us know!

lou-lan:
> OK, sounds like the right path forward is to beef up the documentation in the README for now. If there are any other suggestions you have on how to make this easier to read going forward, please let us know!

@mgleung   @caseydavenport  Automatically generating the README.md document for the Chart could make the project easier to deploy. If there are enhancement plans, I think I could do this work and the following are the planned steps.

* Select a tool for automatically generating the README.md document.
  * https://github.com/norwoodj/helm-docs
  * https://github.com/bitnami-labs/readme-generator-for-helm
* Add comments to the values.yaml.
* Add a step to the Makefile, for example `make helm-chart-docs`.

PR #11546, opened 2025-12-12
docs(tigera-operator): Update values.yaml & the readme

### LINKED ISSUE #11479: CSI Driver gets Installed contradicting Documentation
## Expected Behavior
The default helm values, list `kubeletVolumePluginPath` under `installation.kubeletVolumePluginPath`.

If I don't touch the helm values I'd expect the default values of `None`.

I expect that after the installation `kubeletVolumePluginPath`(in the crd) will be `None` as well.

Even if it is set to `None` (default value) it will install the csi.

## Current Behavior
The csi driver get installed if you set installation.kubeletVolumePluginPath` to "None".
The CSI driver gets installed even if you don't touch the default values.

If you set `kubeletVolumePluginPath: "None"` it will not install: i.e. <s>installation.</s>kubeletVolumePluginPath

## Possible Solution

Adjust documentation and helm default values or source code.

Install the operator via helm without touching the values.yaml

## Your Environment
Helm chart version: 3.31.2

## References
Default values.yaml:
```yaml
  # Disable CSI driver by default.
  kubeletVolumePluginPath: "None"
```

https://github.com/projectcalico/calico/pull/10400/files
https://github.com/projectcalico/calico/issues/7569
https://github.com/projectcalico/calico/issues/8335

https://github.com/tigera/operator/blob/d124be09e418c7e4d8a1ff918e89d6f60eb9f13a/api/v1/installation_types.go#L136

# TL;DR
1. The source code says CSI will be enabled by default. Documentation says otherwise.
2. `kubeletVolumePluginPath` is NOT under `installation`

tom-ludwig:
## Description

Update the documentation for the tigera operator helm chart.
 
- Update readme of the helm chart
    - Remove old `kubeletVolumePluginPath` and move it to `installation.kubeletVolumePluginPath`
    - Update URL of the installation spec
- Update `values.yaml`
    - Add URL to installationSpec
    - Add most important default values to allow the user to quickly adapt the values to their use case

When I start with a poc of a new operator, I'll usually try to deploy it to our poc landscape as soon as possible.
For that I just copy the values.yaml and tweak them to my use case.
I think these changes will make it easier and more clearer where to start.

## Related issues/PRs

* closes #11479 

## Todos

- [x] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
Improve tigera-operator helm chart default values.yaml so it is easier to discover available configuration options. 
```

## Reminder for the reviewer

Other optional labels:


charts/tigera-operator/README.md
```diff
@@ -118,6 +118,10 @@ installation:
   # Example: --set installation.imagePullSecrets[0].name=my-existing-secret
   imagePullSecrets: []
 
+  # Disable CSI driver. Default: /var/lib/kubelet
+  kubeletVolumePluginPath: "None"
+
```
caseydavenport, line 123:
```suggestion
```

charts/tigera-operator/values.yaml
```diff
@@ -5,18 +5,82 @@
...
+
+    # Custom MTU (default: auto-detected based on node interface (1450))
+    # mtu: 1450
+
+    ipPools:
+      - cidr: "192.168.0.0/16"
+        encapsulation: "IPIP" # IPIP, VXLAN, IPIPCrossSubnet, VXLANCrossSubnet, None
+        natOutgoing: "Enabled"
+        nodeSelector: "all()"
+        blockSize: 26
```
caseydavenport: I'm not sure about putting hard-coded IP pools into values.yaml - this is often auto-detected by the operator for the platform (it can detect on some platforms what CIDR the user as chosen automatically) so I worry this might break users on upgrade. Maybe we can leave it commented out?

charts/tigera-operator/values.yaml
```diff
@@ -5,18 +5,82 @@
...
+    # "Calico" (default), "NonPrivileged", "AmazonVPC", "GKE", "AzureVNET", "HostLocal"
+    type: "Calico"
+    ipam:
+      # "Calico" (default), "HostLocal", "AmazonVPC", "AzureVNET"
+      type: "Calico"
+
+  # --- Networking & Dataplane ---
+  calicoNetwork:
+    # "Enabled" (default) or "Disabled"
+    bgp: "Enabled"
```
caseydavenport: I think the same goes here so far as forcing a default that otherwise is auto-detected. Probably should leave `bgp`, `linuxDataplane`, `hostPorts` present but commented out (like you have for MTU) - WDYT?

charts/tigera-operator/values.yaml
```diff
@@ -5,18 +5,82 @@
...
-  kubeletVolumePluginPath: "None"
+  # Valid options: EKS, GKE, AKS, RKE2, OpenShift, DockerEnterprise, TKG, Kind.
+  # If empty the operator will attempt to automatically determine the current provider.
+  kubernetesProvider: ""
+  # Calico variant: "Calico" (default) or "TigeraSecureEnterprise"
+  variant: "Calico"
+
+  # --- CNI Configuration ---
+  cni:
+    # "Calico" (default), "NonPrivileged", "AmazonVPC", "GKE", "AzureVNET", "HostLocal"
```
caseydavenport:
```suggestion
    # "Calico" (default), "AmazonVPC", "GKE", "AzureVNET", "HostLocal"
```

charts/tigera-operator/values.yaml
```diff
@@ -5,18 +5,82 @@
...
   # Example: --set installation.imagePullSecrets[0].name=my-existing-secret
   imagePullSecrets: []
 
-  # Disable CSI driver by default.
-  kubeletVolumePluginPath: "None"
+  # Valid options: EKS, GKE, AKS, RKE2, OpenShift, DockerEnterprise, TKG, Kind.
+  # If empty the operator will attempt to automatically determine the current provider.
+  kubernetesProvider: ""
+  # Calico variant: "Calico" (default) or "TigeraSecureEnterprise"
+  variant: "Calico"
```
caseydavenport: I don't think this field is really supported by the helm chart here actually - the operator understands it, but we don't install all the other CRDs, etc. that would be needed. let's just leave it out for now. 

```suggestion
```

charts/tigera-operator/values.yaml
```diff
@@ -5,18 +5,82 @@
...
+      - cidr: "192.168.0.0/16"
+        encapsulation: "IPIP" # IPIP, VXLAN, IPIPCrossSubnet, VXLANCrossSubnet, None
+        natOutgoing: "Enabled"
+        nodeSelector: "all()"
+        blockSize: 26
+
+  controlPlaneReplicas: 2
+
+  controlPlaneNodeSelector: {}
+
```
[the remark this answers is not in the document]
tom-ludwig: Resolving bc those comments wouldn't add any value.

charts/tigera-operator/values.yaml
```diff
@@ -5,18 +5,82 @@
...
   # imagePullSecrets are a slice of LocalObjectReferences, which is the same format they appear as on deployments.
   #
   # Example: --set installation.imagePullSecrets[0].name=my-existing-secret
   imagePullSecrets: []
 
-  # Disable CSI driver by default.
-  kubeletVolumePluginPath: "None"
+  # Valid options: EKS, GKE, AKS, RKE2, OpenShift, DockerEnterprise, TKG, Kind.
+  # If empty the operator will attempt to automatically determine the current provider.
+  kubernetesProvider: ""
```
caseydavenport: Have you tested this works for upgrades? Will helm revert this to an empty string after the operator defaults it? 

I think we should leave fields commented out if we are intending to rely on operator defaults, and allow users to uncomment them if they want to set them.

tom-ludwig: Oh most probably not 😅. Since its empty by default it shouldn't make a difference ihmo.
I also was in the helm chart before (as an empty string as well), I just added the documentation.

caseydavenport: Ah right! yeah it's probably fine then.

charts/tigera-operator/values.yaml
tom-ludwig: Nice, I wasn't aware of that! I guess that happens on cloud provider environments, not on-premise deployments, so that's why I missed it 👍

charts/tigera-operator/values.yaml
tom-ludwig: Totally! If I understood correctly, the auto-detect logic is in the tigera-operator, right? Is that why it's missing from the `installationSpec` Calico docs?

tom-ludwig:
@caseydavenport I didn't think of that, good catch! I haven't been in the operator code, so I'll take your word for it 😅. Leaving them as comments sounds perfect and would already help me as a user.

tom-ludwig:
@caseydavenport can you please take another look at this? Appreciate it!

charts/tigera-operator/values.yaml
```diff
@@ -5,18 +5,79 @@
...
+
+  cni:
```
caseydavenport: This is another instance where we actually do "magic" defaulting based on the platform, so I think this section should be left commented as well.

For example, the operator code detects the CNI plugin based on the platform provider here: https://github.com/tigera/operator/blob/076feeb163410f4bbb700270f802b6e31c989312/pkg/controller/installation/core_controller.go#L487-L502

charts/tigera-operator/values.yaml
```diff
@@ -5,18 +5,79 @@
...
+
+  # Run Calico in non-privileged mode (Rootless)
+  nonPrivileged: "Disabled"
+
+  # Enable FIPS compliance mode (Requires FIPS compliant cluster/images)
+  # Only supported for Variant=Calico.
+  fipsMode: "Disabled"
+
+  # Configure log severity (Info, Debug, Warning, Error, Fatal)
+  logging:
```
caseydavenport: This one as well - this configuration is only valid for CNI type of `Calico`, so shouldn't be set unconditionally. 

https://github.com/tigera/operator/blob/076feeb163410f4bbb700270f802b6e31c989312/pkg/controller/installation/validation.go#L390-L395

caseydavenport:
Thanks @tom-ludwig - sorry for the delay, I've been out for the holidays. 

I added two more comments on "magic" fields that probably can't be defaulted in the values.yaml file. This is looking pretty close, though!

charts/tigera-operator/values.yaml
```diff
@@ -5,18 +5,79 @@
 # Example: --set-file imagePullSecrets.gcr=./pull-secret.json
 imagePullSecrets: {}
 
+# Configures general installation parameters for Calico. Schema is based
+# on the operator.tigera.io/Installation API documented
+# here: https://docs.tigera.io/calico/latest/reference/installation/api#installationspec
 installation:
   enabled: true
-  kubernetesProvider: ""
+
```
caseydavenport:
```suggestion
  # Uncomment to enable prometheus metrics reporting for node and typha. 
  # nodeMetricsPort: 9090
  # typhaMetricsPort: 9091
```

^ This will address https://github.com/projectcalico/calico/issues/6887

tom-ludwig:
@caseydavenport thanks for you time and the review.

I commented the marked parameters out 👍🏻

caseydavenport:
whoops @tom-ludwig I think you'll need to run `make generate` to update manifests ( we use the chart to generate them ). Although it looks like the generation might be deleting some things it shouldn't, so possibly a bug?

tom-ludwig:
@caseydavenport I likely won't have the bandwidth to work on this until next week. Could you please try running the generation to see if it resolves the issue?

caseydavenport:
@tom-ludwig do you have a moment to run the generation? (Sorry I haven't gotten to this myself)

tom-ludwig:
@caseydavenport I gave this another go on a Linux machine after wasting quite a bit of time trying to get it working on Mac. 

Unfortunately, `make generate` is still giving me a massive diff and I'm not sure if thats supposed to happen. The docs are pretty thin on this, and I’ve already sunk more time into just getting the environment ready than I can justify. I’m going to have to step away from this one for now, sorry.

caseydavenport:
@tom-ludwig sorry about that, I will try my best to get it over the line.

caseydavenport:
Merged here https://github.com/projectcalico/calico/pull/11907

Apologies for the difficulty, and thank you for the PR!

PR #11556, opened 2025-12-15
Goldmane: Add additional filters

skoryk-oleksandr:
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Whisker:  Added new filtering capabilities including "Reporter" filter, "pending/staged actions" filter
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/goldmane/proto/api.pb.go
+++ b/goldmane/proto/api.pb.go
@@ -1161,7 +1161,7 @@ type Filter struct {
 	Actions []Action `protobuf:"varint,7,rep,packed,name=actions,proto3,enum=goldmane.Action" json:"actions,omitempty"`
 	// Policies matches on policy fields. Combined using logical OR.
 	Policies []*PolicyMatch `protobuf:"bytes,8,rep,name=policies,proto3" json:"policies,omitempty"`
-	// Reporter filters on the reporter field. Combined using logical OR.
+	// Reporter filters on the reporter field.
 	Reporter Reporter `protobuf:"varint,9,opt,name=reporter,proto3,enum=goldmane.Reporter" json:"reporter,omitempty"`
 	// Pending/Staged Actions filters on the action field. Combined using logical OR.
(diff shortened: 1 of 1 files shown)


goldmane/pkg/storage/bucket_ring.go
```diff
@@ -190,6 +190,10 @@ func extractPolicyFieldsFromFlowKey(getField func(*proto.PolicyHit) string) func
 		policyTrace := types.FlowLogPolicyToProto(key.Policies())
 		for _, policyList := range [][]*proto.PolicyHit{policyTrace.EnforcedPolicies, policyTrace.PendingPolicies} {
 			for _, p := range policyList {
+				if p.Kind == proto.PolicyKind_Profile {
```
caseydavenport, line 195:
```suggestion
				if p.Kind == proto.PolicyKind_Profile {
				  // Skip Profiles in hints, as these aren't a real kind in Kubernetes clusters - these are 
				  // logically equivalent to "default allow" or "no policies matched". 
```

caseydavenport: Worth commenting on why we're skipping these for any future readers.

goldmane/proto/api.proto
```diff
@@ -200,6 +202,12 @@ message Filter {
 
   // Policies matches on policy fields. Combined using logical OR.
   repeated PolicyMatch policies = 8;
+
+  // Reporter filters on the reporter field. Combined using logical OR.
```
caseydavenport:
```suggestion
  // Reporter filters on the reporter field.
```

^ This isn't a repeated field, so there's nothing to combine. That said, perhaps it _should_ be a repeated field, in case we add more reporters in the future and we want to select on multiple at once?

caseydavenport (changes requested):
Looking pretty good, only minor comments

goldmane/pkg/storage/bucket_ring.go
skoryk-oleksandr: Done :+1:

goldmane/proto/api.proto
skoryk-oleksandr: Done :+1:

PR #11567, opened 2025-12-16
Carry the Tier from the Key when reading from old Typhas

caseydavenport:
## Description

Old versions of Typha don't set the Tier field in the `model.Policy`
object when sending to Felix. This is a problem because new versions of
Felix expect it!

Old versions of Typha DO include the tier in the Key. So, use a private
field to carry that information through and add it to the model.Policy
client-side in order to remain compatible. 

This is a bit of a kludge, but I'm not sure how else to do it!

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/typha/pkg/syncproto/sync_proto_test.go
+++ b/typha/pkg/syncproto/sync_proto_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/model"
 )


libcalico-go/lib/backend/model/policy.go
```diff
@@ -182,8 +185,62 @@ func parseLegacyPolicyName(name string) PolicyKey {
...
+	p, err := l.PolicyKey.parseValue(data)
+	if err != nil {
+		return nil, err
+	}
+	p.(*Policy).Tier = l.Tier
+	return p, nil
+}
+
+func (l LegacyPolicyKey) String() string {
+	panic("LegacyPolicyKey should not be used except to generate a modern PolicyKey")
```
fasaxc: Best not to panic on String(), it can explode a log call unexpectedly

typha/pkg/syncproto/sync_proto.go
```diff
@@ -416,6 +416,12 @@ func (s SerializedUpdate) ToUpdate() (api.Update, error) {
 			}
 		}
 	}
+
+	// If the key supports upgrade, do it now that we've used the key to parse the value.
+	if k, ok := parsedKey.(model.LegacyKey); ok {
+		parsedKey = k.Upgrade()
```
fasaxc, line 422: Debug log?

Maybe add a test to catch if this code got removed/broken?

fasaxc (dismissed):
Couple of minor points

PR #11573, opened 2025-12-17
CI: Rework dependency default branch calculation

fasaxc:
## Description

- Use more precise regex to avoid matching auto-pick branches.
- Improve error handling and logging.
## Related issues/PRs

CORE-12171
## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,7 @@ get-operator-crds: var-require-all-OPERATOR_ORGANIZATION-OPERATOR_GIT_REPO-OPERA
 gen-semaphore-yaml:
 	$(DOCKER_GO_BUILD) sh -c "DEFAULT_BRANCH_OVERRIDE=$(DEFAULT_BRANCH_OVERRIDE) \
 	                          SEMAPHORE_GIT_BRANCH=$(SEMAPHORE_GIT_BRANCH) \
+	                          RELEASE_BRANCH_PREFIX=$(RELEASE_BRANCH_PREFIX) \
 	                          go run ./hack/cmd/deps $(DEPS_ARGS) generate-semaphore-yamls"
 
 GO_DIRS=$(shell find -name '*.go' | grep -v -e './lib/' -e './pkg/' | grep -o --perl '^./\K[^/]+' | sort -u)
--- a/hack/cmd/deps/deps.go
+++ b/hack/cmd/deps/deps.go
@@ -61,6 +61,8 @@ that node depends on felix/bpf-*.
 	os.Exit(1)
 }
 
+const mainBranchName = "master"
+
 var (
 	pretty   = flag.Bool("pretty", false, "Pretty-print the output (only applies to sem-change-in).")
 	logLevel = flag.String("loglevel", "warn", "Logrus log level (debug, info, warn, error, fatal, panic).")
@@ -771,7 +773,7 @@ func calculateBranchStanza(semaphoreDir string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if branch == "master" {
+	if branch == mainBranchName {
 		// Default, so no need to specify.
 		return "", nil
 	}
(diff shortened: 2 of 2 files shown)


hack/cmd/deps/deps.go
```diff
@@ -763,42 +766,63 @@ func validateChangeInClauses(semaphoreDir string) error {
...
+	// Fallback to git.
+	out, err := exec.Command("git", "branch", "--show-current").Output()
+	if err != nil {
+		return "", fmt.Errorf("git branch --show-current failed: %w", err)
+	}
+	branch := strings.TrimSpace(string(out))
+
+	if branch == "naster" {
+		logrus.Info("On master branch, using that for default branch.")
+		return branch, nil
```
radTuti: i think this should just return empty since a default branch is not needed on master. Also avoid having to check for master again at the top.
```suggestion
	if branch == "master" {
		logrus.Info("On master branch, no default branch needed.")
		return "", nil
```

hack/cmd/deps/deps.go
```diff
@@ -763,42 +766,63 @@ func validateChangeInClauses(semaphoreDir string) error {
...
+	}
+
+	// Check for release branch.
+	releaseBranchRegexp := regexp.MustCompile(`release(-calient)?-v[\d.-]+$`)
```
radTuti: might be better to have this use `RELEASE_BRANCH_PREFIX` from env. This is specified in `metadata.mk`

hack/cmd/deps/deps.go
```diff
@@ -763,42 +766,63 @@ func validateChangeInClauses(semaphoreDir string) error {
...
+	if err != nil {
+		return "", err
 	}
-	if current == "" {
-		// Fallback to git.
-		out, err := exec.Command("git", "branch", "--show-current").Output()
-		if err == nil {
-			current = strings.TrimSpace(string(out))
-		}
+	if branch == "master" {
```
radTuti: the default branch fn should return empty for "master" branch since no branch stanza will be needed
```suggestion
	if branch == "" {
```

hack/cmd/deps/deps.go
```diff
@@ -763,42 +766,63 @@ func validateChangeInClauses(semaphoreDir string) error {
...
 	if detected != "" {
-		logrus.Warnf("Currently on a non-master, non-release branch. This branch appears to be a branch of %s; using that as default branch.", detected)
-		return fmt.Sprintf(", default_branch: '%s'", detected)
+		logrus.Infof("Found default branch %s in semaphore.yml, using it for default branch.", detected)
+		return detected, nil
 	}
-	logrus.Warn("Currently on a non-master, non-release branch. Appears to be a branch of master; not specifying default branch.")
-	return ""
+	logrus.Info("Found no default branch in semaphore.yml, assuming master.")
+	return "master", nil
```
radTuti:
```suggestion
	logrus.Info("Found no default branch in semaphore.yml, assuming not needed.")
	return "", nil
```

Makefile
```diff
@@ -108,7 +108,9 @@ get-operator-crds: var-require-all-OPERATOR_ORGANIZATION-OPERATOR_GIT_REPO-OPERA
 	$(MAKE) fix-changed
 
 gen-semaphore-yaml:
-	$(DOCKER_GO_BUILD) sh -c "go run ./hack/cmd/deps generate-semaphore-yamls"
+	$(DOCKER_GO_BUILD) sh -c "DEFAULT_BRANCH_OVERRIDE=$(DEFAULT_BRANCH_OVERRIDE) \
```
radTuti, line 111: have this also take in the `RELEASE_BRANCH_PREFIX` for the release branch regex

fasaxc: Does make the tool a bit harder to run "by hand", but makes sense

PR #11577, opened 2025-12-17
Add configurable max time between successful resyncs

LeoBJenkins:
## Description

This PR
1. Creates a thread that monitors the activity of the `periodic_resync` thread and logs an error if the interval between successful resyncs is too long.
2. Allows users to configure the maximum interval between successful resyncs.

These features are necessary because there is a lack of visibility into whether the resync process is completing regularly. The monitoring thread will grant networking calico users peace of mind as they know the system will alert them via logged errors when the resync process is no longer meeting their expectations.

This PR has both been tested automatically -- see attached tests -- and manually in a development cloud where we verified that errors were only logged when the resync lasted longer than the configured maximum.

Other than increased logging (and a slight increase in threads to be scheduled), there should be no change to the functionality of networking-calico as a result of this PR.

## Related issues/PRs

https://github.com/projectcalico/calico/issues/11579

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Adds configurable field `resync_max_interval_secs` that defines the max time allowed between successful periodic resyncs. Logs an error message if the interval exceeds the threshold.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
+++ b/networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
@@ -459,6 +459,7 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
                 # We deliberately do this last, to ensure that all of the setup
                 # above is complete before we start running.
                 self._epoch += 1
+                eventlet.spawn(self.resync_monitor_thread, self._epoch)
                 eventlet.spawn(self.periodic_resync_thread, self._epoch)
                 if cfg.CONF.calico.etcd_compaction_period_mins > 0:
                     eventlet.spawn(self.periodic_compaction_thread, self._epoch)
@@ -474,19 +475,6 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
 
             self._my_pid = current_pid
 
-            # Start our resynchronization process and status updating. Just in
-            # case we ever get two same threads running, use an epoch counter
-            # to tell the old thread to die.
-            # We deliberately do this last, to ensure that all of the setup
-            # above is complete before we start running.
-            self._epoch += 1
-            eventlet.spawn(self.resync_monitor_thread, self._epoch)
-            eventlet.spawn(self.periodic_resync_thread, self._epoch)
-            if cfg.CONF.calico.etcd_compaction_period_mins > 0:
-                eventlet.spawn(self.periodic_compaction_thread, self._epoch)
-            eventlet.spawn(self._status_updating_thread, self._epoch)
-            for _ in range(cfg.CONF.calico.num_port_status_threads):
-                eventlet.spawn(self._loop_writing_port_statuses, self._epoch)
             LOG.info(
                 "Calico mechanism driver initialisation done in process %s", current_pid
             )


networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -1068,6 +1084,39 @@ def _update_port(self, plugin_context, port):
...
+                    LOG.info("I am master: monitoring periodic resync")
+
+                    time_delta = datetime.now() - self.last_resync_time
+                    if time_delta.seconds > cfg.CONF.calico.resync_max_interval_secs:
+                        LOG.error(
+                            "The time since the last resync completion has surpassed"
+                            f" {cfg.CONF.calico.resync_max_interval_secs} seconds"
+                        )
+
+                    eventlet.sleep(cfg.CONF.calico.resync_max_interval_secs / 5)
```
nelljerram: It would be more precise to sleep until `self.last_resync_time + cfg.CONF.calico.resync_max_interval_secs`, which means to sleep **for** a period of `self.last_resync_time + cfg.CONF.calico.resync_max_interval_secs - now()`.

Then if `self.last_resync_time` is still unchanged, we will log the ERROR at the earliest appropriate time.  OTOH, if `cfg.CONF.calico.resync_max_interval_secs / 5` is less than the period above, then the next check will inevitably not see a problem and will just sleep again.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
```diff
@@ -0,0 +1,161 @@
...
+        self.compaction_patcher = mock.patch(
+            "networking_calico.plugins.ml2.drivers"
+            ".calico.mech_calico.check_request_etcd_compaction"
+        )
+
+    def tearDown(self):
+        self.sleep_patcher.stop()
+        super(TestResyncMonitorThread, self).tearDown()
+
+    def increment_epoch(self, curr_epoch):
```
nelljerram: This feels a bit unintuitive.  Could it be named `set_epoch`, remove `+ 1` below, and add `+ 1` to each of the callers?

nelljerram: (The key problem is that it **doesn't** get and increment the current `self._epoch`.)

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
```diff
@@ -0,0 +1,161 @@
...
+        return lambda _: setattr(self.driver, "_epoch", curr_epoch + 1)
+
+    def test_monitor_does_nothing_when_not_master(self):
+        """Test that a driver that is not master does not monitor."""
+        self.driver.elector.master.return_value = False
+        self.mock_sleep.side_effect = self.increment_epoch(INITIAL_EPOCH)
```
nelljerram: I'm not following this.  Is it important to set the epoch to INITIAL_EPOCH+1 ?

nelljerram: Oh, I guess it's a way to avoid ending up with multiple resync and monitor threads running?

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
```diff
@@ -0,0 +1,161 @@
...
+        """If interval is below max, no error should be logged."""
+        lib.m_compat.cfg.CONF.calico.resync_max_interval_secs = TEST_MAX_INTERVAL
+        self.driver.elector.master.return_value = True
+        self.mock_sleep.side_effect = self.increment_epoch(INITIAL_EPOCH)
+
+        self.driver.resync_monitor_thread(INITIAL_EPOCH)
+
+        self.log_error.assert_not_called()
+
+    def test_monitor_exception_stops_elector(self):
```
nelljerram, line 102: This one feels like it would belong better in test_election.py - WDYT?  (It doesn't have any resync-related detail.)

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # resync mocks
+        self.driver.subnet_syncer = mock.Mock()
+        self.driver.policy_syncer = mock.Mock()
+        self.driver.endpoint_syncer = mock.Mock()
+        self.driver.provide_felix_config = mock.Mock()
+        self.compaction_patcher = mock.patch(
+            "networking_calico.plugins.ml2.drivers"
+            ".calico.mech_calico.check_request_etcd_compaction"
+        )
```
nelljerram: I don't think you'll need this, as our master code has moved compaction to a separate thread.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
```diff
@@ -0,0 +1,161 @@
...
+            self.driver.periodic_resync_thread(curr_epoch)
+        curr_epoch += 1
+
+        self.mock_sleep.side_effect = self.increment_epoch(curr_epoch)
+        self.driver.resync_monitor_thread(curr_epoch)
+
+        self.log_error.assert_called_once()
+
+    def test_errors_continue_to_log(self):
+        """Test that errors contine logging if resync does not occur."""
```
nelljerram:
```suggestion
        """Test that errors continue logging if resync does not occur."""
```

nelljerram: Also I'm wondering why this is a useful behaviour.  E.g. if the resync thread has locked up, why is it helpful to get N ERRORs, instead of just 1?

nelljerram (changes requested):
Many thanks!  I've made some specific comments, but I have an overall point as well: my medium term is to eliminate periodic resync completely, because we don't actually have any need for it.  So I'd just like to check that you won't mind if all of this is ripped out again if/when that happens?

To be clear, we **will** still have a start of day resync, each time that the Neutron server starts or restarts, but we **won't** have further resyncs periodically after the initial one.

nelljerram:
@LeoBJenkins WDYT about a release note and documentation for this?  I think not the latter, as it's a small point, and one that is likely to be reversed in the nearish future.  But perhaps a release note?  If you agree, please fill in that section of the PR description.

nelljerram:
Also needs `make -C networking-calico fmtpy`.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
nelljerram: And we've also done that on the 3.30 and 3.31 branches, so it also won't be needed for backports to those branches.

tj90241:
> Many thanks! I've made some specific comments, but I have an overall point as well: my medium term plan is to eliminate periodic resync completely, because we don't actually have any need for it. So I'd just like to check that you won't mind if all of this is ripped out again if/when that happens?
> 
> To be clear, we **will** still have a start of day resync, each time that the Neutron server starts or restarts, but we **won't** have further resyncs periodically after the initial one.

I will say that at scale, we still do face infrequent problems with e.g. live-migration that could make this change visible. It is likely sufficient to just note it as part of release notes so that operators are aware and can adapt accordingly, but at least an FYI for thought:

IMHO, a harsh reality is that Nova's live-migration process is kind of fragile - in essence, it hinges on an RPC request initiated from the control plane (with RabbitMQ as the substrate) towards a source `nova-compute` service in a manner that is essentially unsupervised in the sense that if the source or destination `nova-compute` ever crashes mid-request, the state generated for the live-migration is never rectified or followed up on automatically. This can be problematic for Neutron, as one of the first steps of live-migration is unfortunately creating an intentionally duplicated port-binding at the destination for post-migration operations...

Upstream has floated the idea of [this script](https://opendev.org/openstack/neutron/commit/d033ab6eb6dadeb1770eb259d4d59b469d9e3bc0?style=unified&whitespace=ignore-change&show-outdated=) to remove the duplicate bindings, but it does direct DB calls IIUC, which leaves the ML2 driver no opportunity to observe changes. In most cases, you'd expect the port-binding to remain ACTIVE at the source (i.e. in sync with `etcd` state?), but I'd be lying if I told you we were never in the situation where we had to help ourselves to the database to push the binding towards the destination with similar database updates. In such cases, we try to make it habit of live-migrating the instance (again, a second time) after a failed live-migration to allow the ML2 driver to observe any changes and flush them to `etcd`, but reality is direct database updates still do happen in breakfix scenarios.

nelljerram:
> > Many thanks! I've made some specific comments, but I have an overall point as well: my medium term plan is to eliminate periodic resync completely, because we don't actually have any need for it. So I'd just like to check that you won't mind if all of this is ripped out again if/when that happens?
> > To be clear, we **will** still have a start of day resync, each time that the Neutron server starts or restarts, but we **won't** have further resyncs periodically after the initial one.
> 
> I will say that at scale, we still do face infrequent problems with e.g. live-migration that could make this change visible. It is likely sufficient to just note it as part of release notes so that operators are aware and can adapt accordingly, but at least an FYI for thought:
> 
> IMHO, a harsh reality is that Nova's live-migration process is kind of fragile - in essence, it hinges on an RPC request initiated from the control plane (with RabbitMQ as the substrate) towards a source `nova-compute` service in a manner that is essentially unsupervised in the sense that if the source or destination `nova-compute` ever crashes mid-request, the state generated for the live-migration is never rectified or followed up on automatically. This can be problematic for Neutron, as one of the first steps of live-migration is unfortunately creating an intentionally duplicated port-binding at the destination for post-migration operations...
> 
> Upstream has floated the idea of [this script](https://opendev.org/openstack/neutron/commit/d033ab6eb6dadeb1770eb259d4d59b469d9e3bc0?style=unified&whitespace=ignore-change&show-outdated=) to remove the duplicate bindings, but it does direct DB calls IIUC, which leaves the ML2 driver no opportunity to observe changes. In most cases, you'd expect the port-binding to remain ACTIVE at the source (i.e. in sync with `etcd` state?), but I'd be lying if I told you we were never in the situation where we had to help ourselves to the database to push the binding towards the destination with similar database updates. In such cases, we try to make it habit of live-migrating the instance (again, a second time) after a failed live-migration to allow the ML2 driver to observe any changes and flush them to `etcd`, but reality is direct database updates still do happen in breakfix scenarios.

Many thanks Tyler, this is great to be aware of.  It still seems to me that a full periodic resync is the wrong hammer for this nail.  I was already thinking that we might create a way to explicitly trigger a resync for a particular port or network, for specific cases where that is needed, and it sounds like this could be one of those cases.  Or, even more specifically, perhaps this is something we should design as part of our live migration support: something like schedule a recheck for the relevant port at a time after the live migration should have completed.

Net, even if we remove the periodic resync, we'll still have these other solutions in mind:
1. Restart the Neutron server.
2. Trigger an explicit resync for the relevant port or network.
3. Something more automatic, but focused on the particular case of concern.

WDYT?

tj90241:
> I was already thinking that we might create a way to explicitly trigger a resync for a particular port or network, for specific cases where that is needed, and it sounds like this could be one of those cases

This would be super! It sounds general enough that e.g. you could still trigger a resync via cron or something in a pinch if needed in addition leveraging it for more-specific breakfix issues.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
LeoBJenkins: I understand the current polling approach is a bit inefficient but I think it's the best solution available and here are a few reasons why.

I think the calculation `self.last_resync_time` + `cfg.CONF.calico.resync_max_interval_secs` - `now()` would lead to negative sleep times if the max interval is exceeded.

A similar approach that we considered would be to sleep for `cfg.CONF.calico.resync_max_interval_secs` every time. However, this approach leads to an extremely high difference between the desired and enforced "max interval" in the worst case. For example, if `cfg.CONF.calico.resync_max_interval_secs` is 5 hours, then it is possible we would not actually log an error until the duration reaches ~10 hours. This would happen if the monitor thread checks to see if the interval has exceeded the threshold right before it does so. The monitor thread would then sleep another 5 hours before checking again and observing the issue. Or, put simply, our worst case error bound on the interval is equal to the time the monitor thread sleeps.

The current solutions balances the worst case error and the inefficiency of polling by sleeping for `cfg.CONF.calico.resync_max_interval_secs / 5` but that was a fairly arbitrary choice.

As you mentioned, this will cause the error to be logged multiple times if the resync still has not executed. I think this a feature as it allows us to distinguish between one-off and persistent failures. It also allows us to setup alarms that increase in severity if the issue is unresolved.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
LeoBJenkins: Yeah so the `mock_sleep` object and `increment_epoch` logic are to turn a long-running thread like `resync_monitor_thread` into a function that runs once and returns. If you look at the `resync_monitor_thread` it loops forever while `self._epoch == launch_epoch`. So by incrementing `self._epoch` instead of sleeping, the thread will die after a single run.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
LeoBJenkins: I think it belongs where it is because it calls and tests the monitor thread. Plus, it does not test any of the actual election logic (the actual elector is entirely mocked out), it just checks that the function is called.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
LeoBJenkins: I didn't see this question previously so I answered it here
https://github.com/projectcalico/calico/pull/11577#discussion_r2651306469

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
LeoBJenkins: I renamed it to `simulate_epoch_progression` to better describe how it's being used in the tests. Please let me know what you think

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
LeoBJenkins: Yes! Removed!

nelljerram:
Formatting is good now, but there are a few UT issues, please look for "FAIL:" here: 
[caracal-ut.log](https://github.com/user-attachments/files/24381680/caracal-ut.log)

nelljerram:
> > @LeoBJenkins WDYT about a release note and documentation for this? I think not the latter, as it's a small point, and one that is likely to be reversed in the nearish future. But perhaps a release note? If you agree, please fill in that section of the PR description.
> 
> I'm happy to do whatever you think is best! What goes into a release note/is there any documentation I can follow when writing one? Thank you!

I don't think we have any formal doc about this, but it's just a sentence or two summarising the change.  Like one of the bullet points here: https://docs.tigera.io/calico/latest/release-notes/#enhancements

nelljerram:
> > I was already thinking that we might create a way to explicitly trigger a resync for a particular port or network, for specific cases where that is needed, and it sounds like this could be one of those cases
> 
> This would be super! It sounds general enough that e.g. you could still trigger a resync via cron or something in a pinch if needed in addition leveraging it for more-specific breakfix issues.

Thanks; so I think that will be the plan in principle.  But it will probably take a good while to land because we have a lot of higher priority things to do first :smile:  So it definitely makes sense to land this current PR for the interim.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: I see that this interacts with the point about repeating the error if the problem persists.  With that in mind, what about the best of both worlds:
- `self.last_resync_time + cfg.CONF.calico.resync_max_interval_secs - now()` if this is >= 0
- else `cfg.CONF.calico.resync_max_interval_secs / 5`

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
nelljerram: I don't understand why this function takes a `curr_epoch` or `init_epoch` arg?  Why would you not use `getattr` to get the current `_epoch` value, then increment, then `setattr` to set that?

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
nelljerram: I'm sorry, you're right.  I must have misread before.

nelljerram (commented):
Just two remaining questions now:
- about the epoch increment coding
- about the ideal sleep time.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
LeoBJenkins: I like that -- it's the best of both worlds. I implemented this logic and added two tests for it.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_monitor_thread.py
LeoBJenkins: Oh yeah that's not necessary. I removed the `curr_epoch` argument. I had to replace it with `expected_sleep_time`  in order to test the new sleep time logic, however, I think it is much simpler/easy to understand now.

nelljerram:
> @nelljerram Hello, sorry to bother you. Just wanted to see if the most recent commit resolved your remaining questions. Thank you!

@LeoBJenkins I'm sorry, have been busy with other things.  I think probably yes; just need to give this some focus.  I'm not sure if I have time for that today, but definitely on Monday.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -459,6 +474,19 @@ def _post_fork_init(self, voting=False):
 
             self._my_pid = current_pid
 
+            # Start our resynchronization process and status updating. Just in
```
nelljerram: I'm sorry, but Git is placing this code in the wrong place, following the merging of #11580.  I think you need to add the `resync_monitor_thread` line in the similar block above, instead, and delete this block.

LeoBJenkins: Oof yeah I think I messed that up when I rebased with the `master` branch. I think I just fixed it.

nelljerram:
@LeoBJenkins All good to go except the CLA approval is not quite hooked up yet.  Would it be easy for you to add the email you've used in your commits to your GitHub account, in order to satisfy this?
image

nelljerram:
Thanks @LeoBJenkins , all merged now, and I will prepare cherry picks for the v3.30 and v3.31 branches.

PR #11581, opened 2025-12-17
Add Gateway API conformance tests to CI pipeline

electricjesus:
## Description

This PR adds Gateway API conformance testing to the Calico CI pipeline.

Automated conformance testing to ensure Calico's Gateway API implementation meets the standard specification and to catch regressions early.

Installs latest calicoreates an `operator.tigera.io/GatewayAPI` resource with matching `gateway.networking.k8s.io/Gateway` then Gateway Conformance tests are run.

## Related issues/PRs

N/A

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```


.semaphore/semaphore.yml.d/blocks/20-api.yml
```diff
@@ -1,15 +0,0 @@
-- name: API
```
caseydavenport, line 1: @electricjesus was it intentional to remove all of these blocks? Seems wrong?

electricjesus: hey casey - just got back to this. intentional but only temporary so i can "focus" semaphore only on the conformance tests that i want

third_party/envoy-gateway/Makefile
```diff
@@ -2,7 +2,7 @@ include ../../metadata.mk
 
 PACKAGE_NAME ?= github.com/projectcalico/calico/third_party/envoy-gateway
 
-ENVOY_GATEWAY_IMAGE ?= envoy-gateway
+ENVOY_GATEWAY_IMAGE ?= calico/envoy-gateway
```
nelljerram: It looks like this is against our convention, i.e. other places don't have "calico/" prefix.  Is there a reason for adding it here?

Same for other Envoy images.

lib.Makefile
```diff
@@ -1611,6 +1617,18 @@ $(REPO_ROOT)/whisker/.image.created-$(ARCH):
...
+# names (.envoy-<comp>.created-$(ARCH)). The sub-make handles fetching
+# upstream sources and building the image as calico/envoy-<comp>:latest-$(ARCH).
+$(REPO_ROOT)/third_party/envoy-gateway/.envoy-gateway.created-$(ARCH):
+	$(MAKE) -C $(REPO_ROOT)/third_party/envoy-gateway image
+
+$(REPO_ROOT)/third_party/envoy-proxy/.envoy-proxy.created-$(ARCH):
+	$(MAKE) -C $(REPO_ROOT)/third_party/envoy-proxy image
+
+$(REPO_ROOT)/third_party/envoy-ratelimit/.envoy-ratelimit.created-$(ARCH):
+	$(MAKE) -C $(REPO_ROOT)/third_party/envoy-ratelimit image
```
nelljerram, line 1730: Hmm; probably not an issue for this PR specifically, but looks like we're starting to develop a parallel release image process here, which doesn't feel right.

Makefile
```diff
@@ -275,6 +275,140 @@ e2e-run-cnp:
...
+# Gateway API conformance
+#
+# Runs the upstream sigs.k8s.io/gateway-api conformance suite against
+# Calico's Envoy-Gateway-based implementation on the cluster at $KUBECONFIG,
+# and emits a ConformanceReport YAML for upstream submission.
+#
+# Caller must:
+#   1. set KUBECONFIG (e.g. $(KIND_KUBECONFIG))
+#   2. set GATEWAY_CONFORMANCE_VERSION to a Calico release tag (e.g. v3.31.5)
+#      for upstream submissions; required by the gateway-api report schema.
```
nelljerram: This detail feels wrong.  Surely this Makefile should be able to infer the current tag?

Shouldn't the overall logic be:
1. Always run the conformance test, as part of our CI.
2. If it passes, AND if we're on a release tag, make an upstream submission to report that.

Makefile
```diff
@@ -275,6 +275,140 @@ e2e-run-cnp:
...
+###############################################################################
+GATEWAY_CONFORMANCE_VERSION ?=
+GATEWAY_CLASS_NAME ?= tigera-gateway-class
+GATEWAY_CONFORMANCE_PROFILES ?= GATEWAY-HTTP
+GATEWAY_CONFORMANCE_MODE ?= default
+GATEWAY_CONFORMANCE_REPORT ?= $(REPO_ROOT)/$(E2E_OUTPUT_DIR)/gateway-conformance-report.yaml
+GATEWAY_CONFORMANCE_ORG ?= projectcalico
+GATEWAY_CONFORMANCE_PROJECT ?= calico
+GATEWAY_CONFORMANCE_URL ?= https://github.com/projectcalico/calico
+GATEWAY_CONFORMANCE_CONTACT ?= https://github.com/projectcalico/calico/blob/master/GOVERNANCE.md
```
nelljerram: `GOVERNANCE.md` file does not exist.  What is the intent here?

e2e/cmd/gateway/manifests/gatewayapi.yaml
```diff
@@ -0,0 +1,29 @@
...
+# speaker node is required. See envoyproxy.yaml for the full rationale.
+#
+# Bootstrap order matters: the gatewayapi controller resolves
+# envoyProxyRef before it renders the `tigera-gateway` namespace, so
+# applying this CR alone makes the operator requeue forever with
+# "EnvoyProxy ... not found". `make e2e-gateway-setup` works around
+# this by applying this manifest, waiting for the EnvoyProxy CRD,
+# pre-creating the `tigera-gateway` namespace itself, and then
+# applying envoyproxy.yaml so the next reconcile finds the EnvoyProxy
+# and proceeds.
```
nelljerram: A 3rd party EnvoyProxy (for `envoyProxyRef`) should not be configured in the tigera-gateway namespace; and it sound like it would avoid most of the above issue if it wasn't.  Can you tweak to configure the EnvoyProxy in a different namespace instead?

e2e/cmd/gateway/manifests/envoyproxy.yaml
```diff
@@ -0,0 +1,34 @@
...
+# share a node. Calico's kind setup is multi-node (1 control + 3
+# workers), so any test where the speaker and the backend pod land on
+# different nodes is dropped by IPVS — visible as i/o timeouts on
+# BackendTLSPolicy*, HTTPRouteCrossNamespace, etc.
+#
+# Override to "Cluster" so traffic from any speaker node can route to
+# any envoy pod via IPVS. The Calico tigera-operator picks this up
+# through the GatewayAPI CR's envoyProxyRef (see gatewayapi.yaml).
+#
+# This resource is created in the `tigera-gateway` namespace.
```
nelljerram: As commented elsewhere, I think it should be provisioned in a different namespace.

e2e/cmd/gateway/manifests/metallb-pool.yaml
```diff
@@ -0,0 +1,49 @@
+# Gateway API conformance metallb IPAddressPool.
+#
+# Calico's default metallb pool (hack/test/kind/infra/metallb-config.yaml)
+# uses BGP advertisement with public-IP addresses (80.15.0.0/24). Those IPs
+# get programmed as routes inside the cluster via Calico's BGP fan-out, so
+# in-cluster pods can reach them, but the Gateway API conformance test
+# binary runs from the host (Semaphore VM or developer laptop) where there
+# is no BGP session and the IPs don't resolve.
```
nelljerram:
```suggestion
# is no BGP session and so the IPs aren't reachable.
```

e2e/cmd/gateway/manifests/metallb-pool.yaml
```diff
@@ -0,0 +1,49 @@
...
+# `make e2e-gateway-setup`, which queries `docker network inspect kind`
+# for the kind bridge IPv4 subnet and picks .200-.250 of the same prefix
+# (matches what envoyproxy/gateway's tools/hack/create-cluster.sh does).
+#
+# serviceAllocation.priority is higher than the default BGP pool so the
+# conformance Services pin here deterministically; namespaces are the
+# three the upstream conformance suite's base manifests create
+# (gateway-conformance-{infra,app-backend,web-backend}) plus
+# tigera-gateway in case the operator places EG-managed LB Services in
+# the controller's namespace rather than the Gateway's.
```
nelljerram: I believe our Gateway impl always provisions Gateway Services in the tigera-gateway namespace, no?  Because that's what EG does.  I've heard of EG being enhanced to follow the namespace of the Gateway resource, but has that landed yet?

Makefile
```diff
@@ -275,6 +275,140 @@ e2e-run-cnp:
...
+GATEWAY_API_CR ?= $(REPO_ROOT)/e2e/cmd/gateway/manifests/gatewayapi.yaml
+GATEWAY_ENVOY_PROXY ?= $(REPO_ROOT)/e2e/cmd/gateway/manifests/envoyproxy.yaml
+GATEWAY_METALLB_POOL ?= $(REPO_ROOT)/e2e/cmd/gateway/manifests/metallb-pool.yaml
+# Name of the docker network kind binds to. The kind default is "kind".
+GATEWAY_KIND_DOCKER_NETWORK ?= kind
+# Calico's tigera-operator-provisioned GatewayClass doesn't populate
+# .status.supportedFeatures, so the conformance suite's auto-inference returns
```
nelljerram: Is that a bug that we should just fix in operator, instead of working around?

Makefile
```diff
@@ -275,6 +275,140 @@ e2e-run-cnp:
...
+## installs the Envoy Gateway controller, and only then creates the default
+## GatewayClass. `kubectl wait --for=condition` against a resource type that
+## doesn't yet exist on the server errors immediately rather than retrying,
+## so we poll for the CRD first, then wait for the GatewayClass.
+GATEWAY_SETUP_CRD_TIMEOUT ?= 300
+GATEWAY_SETUP_GWC_TIMEOUT ?= 5m
+
+.PHONY: e2e-gateway-setup
+e2e-gateway-setup:
+	@if [ -z "$(KUBECONFIG)" ]; then echo "e2e-gateway-setup: KUBECONFIG must be set"; exit 1; fi
```
nelljerram: I would prefer to delete all the leading `@`s here (and throughout).  IMO more output is helpful.

Makefile
```diff
@@ -275,6 +275,140 @@ e2e-run-cnp:
...
+	kubectl --kubeconfig=$(KUBECONFIG) apply -f $(GATEWAY_API_CR)
+	@echo "Waiting up to $(GATEWAY_SETUP_CRD_TIMEOUT)s for tigera-operator to install Gateway API CRDs..."
+	@# The gatewayapi controller renders CRDs early (gatewayapi_controller.go:221)
+	@# but only renders the tigera-gateway namespace + non-CRD resources AFTER
+	@# resolving every envoyProxyRef (line 419, after the read at line 350). With
+	@# envoyProxyRef set on first reconcile the operator returns early at line 351
+	@# with "EnvoyProxy ... not found", so the namespace never lands. We unblock
+	@# this by pre-creating the namespace ourselves once the EnvoyProxy CRD is
+	@# in, then dropping the EnvoyProxy CR in. The next reconcile finds it and
+	@# the operator owns the namespace from then on.
```
nelljerram: If the EnvoyProxy is provisioned in another namespace, we still need the CRD.  So I think the sequence becomes:
1. provision GatewayAPI CR
2. wait for CRDs
3. provision EnvoyProxy
4. wait for GatewayClass

so pretty much the same except don't need to pre-provision tigera-gateway namespace.

Makefile
```diff
@@ -275,6 +275,140 @@ e2e-run-cnp:
...
+	@end=$$(( $$(date +%s) + $(GATEWAY_SETUP_CRD_TIMEOUT) )); \
+	until kubectl --kubeconfig=$(KUBECONFIG) get crd gatewayclasses.gateway.networking.k8s.io >/dev/null 2>&1 \
```
nelljerram: Don't need to wait for GatewayClass CRD here.

Makefile
```diff
@@ -275,6 +275,140 @@ e2e-run-cnp:
...
+	    echo "Timed out waiting for Gateway API + EnvoyProxy CRDs"; \
+	    kubectl --kubeconfig=$(KUBECONFIG) get gatewayapi default -o yaml || true; \
+	    kubectl --kubeconfig=$(KUBECONFIG) get tigerastatus || true; \
+	    kubectl --kubeconfig=$(KUBECONFIG) -n tigera-operator logs deploy/tigera-operator --tail=200 || true; \
+	    exit 1; \
+	  fi; \
+	  sleep 5; \
+	done
+	@echo "Pre-creating tigera-gateway namespace so the EnvoyProxy CR can be applied before the operator reconciles past the EnvoyProxyRef check"
+	kubectl --kubeconfig=$(KUBECONFIG) create namespace tigera-gateway --dry-run=client -o yaml | kubectl --kubeconfig=$(KUBECONFIG) apply -f -
```
nelljerram: Delete 2 previous lines

Makefile
nelljerram: Also this is a lot of shell to have directly in Makefile.  Worth moving to a script somewhere?

Makefile
```diff
@@ -275,6 +275,140 @@ e2e-run-cnp:
...
+	    echo "Timed out waiting for gatewayclass/$(GATEWAY_CLASS_NAME) to exist"; \
+	    kubectl --kubeconfig=$(KUBECONFIG) get gatewayclass || true; \
+	    exit 1; \
+	  fi; \
+	  sleep 5; \
+	done
+	kubectl --kubeconfig=$(KUBECONFIG) wait --for=condition=Accepted=true --timeout=$(GATEWAY_SETUP_GWC_TIMEOUT) gatewayclass/$(GATEWAY_CLASS_NAME)
+
+## Run the Gateway API conformance suite. Requires e2e-gateway-setup to have completed.
+e2e-run-gateway-conformance:
```
nelljerram: Depend on e2e-gateway-setup?

.semaphore/semaphore.yml.d/blocks/20-gateway-conformance.yml
```diff
@@ -0,0 +1,42 @@
...
+          # Build the conformance runner.
+          - make -C e2e bin/gateway/e2e.test
+          # Bring up a kind cluster and deploy Calico (via Helm + tigera-operator).
+          - make kind-up
+          # Apply the GatewayAPI operator CR and wait for the default
+          # GatewayClass to reach Accepted=true.
+          - make e2e-gateway-setup KUBECONFIG=$(pwd)/hack/test/kind/kind-kubeconfig.yaml
+          # Run the upstream Gateway API conformance suite against the cluster
+          # and emit the ConformanceReport YAML to ./report/.
+          - .semaphore/run-and-monitor gateway-conformance.log make e2e-run-gateway-conformance KUBECONFIG=$(pwd)/hack/test/kind/kind-kubeconfig.yaml GATEWAY_CONFORMANCE_VERSION=${SEMAPHORE_GIT_TAG_NAME:-${SEMAPHORE_GIT_BRANCH}}
```
nelljerram: Can we replace all the above with `make e2e-test-gateway-conformance` ?  It feels wrong to duplicate the internal steps both here and in Makefile.

nelljerram (changes requested):
A few detailed comments here, but overall this is looking great.

I'd appreciate another set of eyes on aspects I don't feel confident about:
1. metallb update - does that have effects elsewhere across our Calico product?
2. Adding gateway images to KIND_CALICO_IMAGES - does that have effects elsewhere across our Calico product?

nelljerram:
Looking at the gateway-conformance-report.yaml artifact in CI:

> `  organization: projectcalico`
Should it be Tigera?  Or "Project Calico"?  Do you know how this field gets consumed and presented on the Gateway API website?

> `  project: calico`
Ditto; perhaps should be Calico, to look better on the website.

nelljerram:
One more nit: I noticed this traceback in the full log:
```
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
Detected at:
	>  goroutine 880 [running]:
	>  runtime/debug.Stack()
	>  	/usr/local/go/src/runtime/debug/stack.go:26 +0x5e
	>  sigs.k8s.io/controller-runtime/pkg/log.eventuallyFulfillRoot()
	>  	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/log/log.go:60 +0xcd
	>  sigs.k8s.io/controller-runtime/pkg/log.(*delegatingLogSink).WithValues(0x1730f2c02540, {0x0, 0x0, 0x0})
	>  	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/log/deleg.go:168 +0x49
...
```
Would it be easy to clean that up by adding a call to `log.SetLogger`?

electricjesus:
thanks @nelljerram ! pinging @caseydavenport for a re-review

electricjesus:
Thanks for the careful read Nell. On the two points you flagged not feeling confident about — both are scoped strictly to the kind CI flow, nothing changes for shipped Calico:

The metallb upgrade only touches files under `hack/test/kind/`. Production Calico doesn't bundle metallb at all (it's purely a kind-cluster convenience for assigning LoadBalancer IPs in CI). I had to upgrade because metallb 0.14 dropped the legacy ConfigMap install path — the new metallb-config.yaml is a straight 1:1 conversion of the old ConfigMap (same `80.15.0.0/24` + `fdff::/64` ranges, same BGPAdvertisement). I checked deploy_resources.sh and the kind infra layout to make sure nothing else was leaning on the ConfigMap shape. So no spillover.

The KIND_CALICO_IMAGES extension is similarly local. The three `calico/envoy-{gateway,proxy,ratelimit}` images already get built per-PR by the existing Envoy Gateway / Envoy Proxy / Envoy Ratelimit blocks; this change just teaches kind-build to load them so the conformance block has them on the kind nodes. They aren't released, packaged, or tagged differently. Happy to walk through the call graph if helpful.

The reason both pieces of plumbing existed at all is that Calico's kind topology diverges from Envoy Gateway's own conformance CI, which is also why a chunk of the per-test workarounds were needed. Posting a side-by-side because I think it'll help future-me too:

Calico kind: 1 control + 3 workers. Envoy Gateway kind: `NUM_WORKERS=0`, single node. That single difference is why we needed the L2 metallb pool from the kind subnet (single-node EG kind never has cross-node speaker/backend mismatch) and why we override `externalTrafficPolicy` to `Cluster` on the Envoy data-plane Service (`Local` works fine for them because speaker = backend node always; on multi-node we drop packets in IPVS).

Calico install path: tigera-operator GatewayAPI CR. Envoy Gateway CI: their own Helm chart, direct. That's why the bootstrap order is more involved on our side. After your review I moved the EnvoyProxy CR out of `tigera-gateway` and into a test-owned namespace — the operator's gatewayapi controller renders namespaces after resolving every `envoyProxyRef`, so referencing `tigera-gateway` on a fresh install was a chicken-and-egg. With the EnvoyProxy in our own namespace, no pre-create dance is needed and the comments calling out an operator quirk are gone. (Still worth filing a separate operator issue: the controller could just requeue on missing EnvoyProxy instead of going Degraded, but that's not in scope for this PR.)

Outside those two roots, everything else is a thin wrapper around the same upstream `sigs.k8s.io/gateway-api/conformance/utils/suite` that EG itself uses. The curated feature/test/profile set is mirrored verbatim from envoyproxy/gateway v1.7.0's `internal/gatewayapi/conformance/suite.go` — so what we claim is what they claim, minus what's not relevant for us (mesh).

third_party/envoy-gateway/Makefile
electricjesus: The `calico/` prefix is what KIND_CALICO_IMAGES expects (kind-build's loader greps for it to decide what's a Calico-built image vs an upstream pull). I added it on the three Envoy targets so the conformance block can reuse the existing kind-build path without duplicating image-loading logic. Happy to revert and inline a separate loader instead if there's a reason the convention is intentional — I couldn't find one in the surrounding history but I may be missing something. Which way would you prefer?

lib.Makefile
electricjesus: Yeah I see why it reads that way. To clarify: this is purely the kind-cluster image-load path (KIND_IMAGE_MARKERS files that kind-build uses to decide whether an image needs to be reloaded). It doesn't touch release packaging, tagging, or the published images. The `third_party/envoy-*` directories already produce these images for their own per-PR CI; I'm only adding markers so kind-build knows about them. Open to factoring it out if it bothers you, but I think it's the same shape as the existing KIND_CALICO_IMAGES entries.

Makefile
electricjesus: Resolved a different way after looking at it again. Moved the EnvoyProxy CR out of `tigera-gateway` (which the operator owns) into `e2e-gateway-conformance` (which we own). The deadlock was that the operator's gatewayapi controller resolves every `envoyProxyRef` before it renders the `tigera-gateway` namespace, so referencing it on a fresh install was a chicken-and-egg. With the EnvoyProxy in our own namespace, no workaround is needed. Pushed the change in 095040d9 / 385a7a60 — the namespace pre-create code and the explanatory comments about operator internals are gone.

The underlying operator behaviour is still arguably wrong (a missing EnvoyProxyRef could just requeue instead of going Degraded and returning early), but that's a separate operator-side fix; not in scope here.

electricjesus:
@nelljerram Re: logger, fixed in 41ae1a91 — the runner now installs a zap logger up front, mirroring upstream `conformance.DefaultOptions`. The traceback is gone from subsequent runs.

electricjesus:
@nelljerram good catch on this one, the answer surprised me. I went and looked at the existing submissions in `kubernetes-sigs/gateway-api/conformance/reports/v1.3.0/` and the convention is actually lowercase, hyphenated. Examples:

- envoyproxy / envoy-gateway
- cilium / cilium
- aws / aws-load-balancer-controller
- GKE / gke-gateway (sole TitleCase outlier)

These values become the directory key under `conformance/reports/<gw-api-version>/<org>-<project>/`. The implementations index on gateway-api.sigs.k8s.io is hand-curated, so the display niceness doesn't depend on the YAML.

So the original `projectcalico` / `calico` was already in line — matches the convention, matches the GitHub org name. Reverted in 5081bbb6.

e2e/cmd/gateway/manifests/metallb-pool.yaml
electricjesus: Checked here and you're right `internal/infrastructure/kubernetes/proxy_infra.go::GetResourceNamespace` returns the EG `ControllerNamespace` (so `tigera-gateway` for us) unless `GatewayNamespaceMode` is set. The "follow the namespace of the Gateway resource" enhancement landed (it's `KubernetesDeployModeTypeGatewayNamespace` in v1alpha1, exposed via `Provider.Kubernetes.Deploy.Type`), but it's opt-in. Our tigera-operator's gateway-api render doesn't set it (`Provider.Kubernetes.Deploy` is left at its zero value), so we're firmly in default mode and Services always end up in `tigera-gateway`.

Trimmed the `serviceAllocation.namespaces` list down to just `tigera-gateway` and dropped the hedging comment in 5c98d08a. Verified the conformance run is still green with the trimmed pool.

caseydavenport:
Just noting that I am planning to get to this one, but it might be a bit later in the week! Thanks @electricjesus !

e2e/cmd/gateway/manifests/envoyproxy.yaml
```diff
@@ -0,0 +1,32 @@
...
+kind: EnvoyProxy
+metadata:
+  name: gateway-conformance-base
+  namespace: e2e-gateway-conformance
+spec:
+  provider:
+    type: Kubernetes
+    kubernetes:
+      envoyService:
+        externalTrafficPolicy: Cluster
```
nelljerram, line 32: This is fine, but FWIW we now also have the option of using the new Patch field.  (Although that would make this harder to backport.)

Makefile
nelljerram: What I meant was: should we fix operator so that our GatewayClasses have status.supportedFeatures populated?  It sounds like we are talking at cross-purposes here.

nelljerram (approved):
I'm happy with this now.  Couple more comments below, but non-blocking.

electricjesus:
thanks @nelljerram ! 


RE:

> What I meant was: should we fix operator so that our GatewayClasses have status.supportedFeatures populated? It sounds like we are talking at cross-purposes here.

looks like the empty `status.supportedFeatures` actually isn't tigera-operator's omission but envoyproxy/gateway deliberately leaves it unpopulated:
- 1.7.2 https://github.com/envoyproxy/gateway/blob/release/v1.7/internal/gatewayapi/status/gatewayclass.go#L36-L41
- also in main https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/status/gatewayclass.go#L36 

And they cite the field's experimental/unstable datatype as the reason

So the `-supported-features` override in the Makefile is the same workaround upstream's own conformance CI uses, not Calico-specific.

I'll update the comment in the Makefile to make that clearer. If/when upstream re-enables population once the datatype graduates, we'll inherit it.

nelljerram:
> thanks @nelljerram !
> 
> RE:
> 
> > What I meant was: should we fix operator so that our GatewayClasses have status.supportedFeatures populated? It sounds like we are talking at cross-purposes here.
> 
> looks like the empty `status.supportedFeatures` actually isn't tigera-operator's omission but envoyproxy/gateway deliberately leaves it unpopulated:
> 
> * 1.7.2 https://github.com/envoyproxy/gateway/blob/release/v1.7/internal/gatewayapi/status/gatewayclass.go#L36-L41
> * also in main https://github.com/envoyproxy/gateway/blob/main/internal/gatewayapi/status/gatewayclass.go#L36
> 
> And they cite the field's experimental/unstable datatype as the reason
> 
> So the `-supported-features` override in the Makefile is the same workaround upstream's own conformance CI uses, not a Calico-specific.
> 
> I'll update the comment in the Makefile to make that clearer. If/when upstream re-enables population once the datatype graduates, we'll inherit it.

Ah right.  I was imagining that tigera/operator could populate status.supportedFeatures at the time that it provisions the GatewayClass - but I guess not, because we should only populate `spec`, and it makes sense that upstream is responsible for `status`

e2e/cmd/gateway/manifests/envoyproxy.yaml
```diff
@@ -0,0 +1,32 @@
+# Base EnvoyProxy used by the Gateway API conformance flow.
+#
+# Envoy Gateway upstream defaults the data-plane Service's
+# `externalTrafficPolicy` to "Local" (api/v1alpha1/shared_types.go in
+# envoyproxy/gateway, +kubebuilder:default:="Local"). Their own
+# conformance CI gets away with this because it uses a single-node kind
+# cluster (tools/hack/create-cluster.sh defaults NUM_WORKERS to 0), so
+# the metallb speaker that wins the IP and the envoy backend pod always
+# share a node. Calico's kind setup is multi-node (1 control + 3
```
caseydavenport, line 9: We do support a single-node kind setup that might be helpful here? libcalico-go spins it up - there's a kind config env var that controls it. 

Multi-node is fine too, just a bit more resource / time intensive.

hack/test/kind/deploy_resources.sh
```diff
@@ -187,9 +187,27 @@ ${kubectl} apply -f ${INFRA_DIR}/calicoctl.yaml
 echo
 
 echo "Install MetalLB controller for allocating LoadBalancer IPs"
+# metallb-native.yaml ships its own Namespace; create-ns is for legacy
+# layouts and stays here as a no-op safeguard for older clones.
 ${kubectl} create ns metallb-system || true
 ${kubectl} apply -f ${INFRA_DIR}/metallb.yaml
-${kubectl} apply -f ${INFRA_DIR}/metallb-config.yaml
+# metallb-native installs CRDs and a ValidatingWebhookConfiguration; wait
```
caseydavenport: I do wonder if we should just be moving away from metal LB altogether in this test suite - IIUC we have our own LoadBalancer IP allocation implementation now that is available without needing to install any extra addons.

Probably a cleaner approach?

node/tests/k8st/tests/test_bgp_advert.py
```diff
@@ -764,8 +764,8 @@ def test_bgp_filter_ip_advertisement(self):
     action: Reject
 EOF
 """)
-        kubectl("patch --type=merge bgppeer node-extra.peer --patch '{\"spec\": {\"filters\": [\"test-filter-export-1\"]}}'")
-        self.add_cleanup(lambda: kubectl("patch --type=merge bgppeer node-extra.peer --patch '{\"spec\": {\"filters\": []}}'"))
+        kubectl("patch --type=merge bgppeer.crd.projectcalico.org node-extra.peer --patch '{\"spec\": {\"filters\": [\"test-filter-export-1\"]}}'")
```
caseydavenport: This is wrong - we shouldn't be referring to this API group here. Why did you need this change?

lib.Makefile
caseydavenport: Yeah, this isn't really the same as the release-image flow - this is a dev flow builit specifically for fast iteration against kind (and other clusters too!)

@electricjesus if you rebase on master I think you'll find that the logic for these stamps has changed a little bit to detect when e.g., the stamp exists but the image has been cleaned up under its feet.

lib.Makefile
```diff
@@ -1711,6 +1717,26 @@ $(REPO_ROOT)/key-cert-provisioner/.image.created-$(ARCH): \
...
+
+# Operator is built from a separate repo/branch. It only needs
+# calico_versions.yml (a static file with version strings), not the
+# actual built images, so it can run in parallel with component builds.
+# Track the operator source selection in a generated inputs file so
+# changes to repo/branch invalidate the operator stamp.
+.PHONY: FORCE
```
caseydavenport: What's going on here? Looks like dead code?

caseydavenport (commented):
A few thoughts / questions.

node/tests/k8st/tests/test_bgp_advert.py
electricjesus: Fixed — pushed `.projectcalico.org` (v3 group) everywhere. Qualification is still needed because metallb v0.14 ships `bgppeers.metallb.io`, so `bgppeer` shortname resolves ambiguously now. Stripping that CRD from the manifest crashes metallb's controller at startup (it pre-resolves every type at boot).

Same push tackles your "we have our own LB allocator now" point: 80.15.0.0/24 and fdff::/64 are now Calico-native IPPools (`allowedUses: [LoadBalancer]`); IPv6 single-IP LB test converges on the IPv4 sibling's pattern. metallb stays installed only for Gateway API conformance L2 reachability — confd is BGP-only and kube-controllers' LB controller is IPAM-only, so metallb's L2 speaker remains the only LB-IP L2 announcer in our stack.

Validated on a build VM: node k8st BGP 24/24, Gateway API conformance 430/430.

lib.Makefile
electricjesus: done removed

hack/test/kind/deploy_resources.sh
electricjesus: Done — Calico's kube-controllers loadbalancer controller is now the IPAM authority (Calico-native IPPools in `hack/test/kind/infra/calico-lb-pools.yaml`, `allowedUses: [LoadBalancer]`). metallb is still installed, but only to serve the Gateway API conformance L2 pool — Calico has no L2 announcer (confd is BGP-only, the LB controller is IPAM-only), so the conformance host runner still needs metallb's L2 speaker to ARP for LB IPs on the kind docker bridge. Validated on a build VM: node k8st BGP 24/24, Gateway API conformance 430/430.

node/tests/k8st/tests/test_bgp_advert.py
caseydavenport: Must we install the bgppeers CRD from metallb?

caseydavenport: Yeah, seems like we must.

caseydavenport (approved):

PR #11592, opened 2025-12-19
CI: Tear down test VMs as soon as they're finished.

fasaxc:
## Description

Previously, we ran several global steps in sequence, waiting for each to complete before starting the next:

* Set up the test VMs
* (For the JIT-hardened tests) Do an in-line tweak to the VMs in the semaphore command list
* Run tests
* Collect artifacts
* Tear down VMs

This is inefficient for two reasons:

* Some test batched finish sooner and the VM ends up living longer than needed
* Collecting artifacts contends for bandwidth, slowing it down.

This PR runs a backgrounded subshell for each VM and then does the steps in sequence in the per-VM subshell, without waiting for any other VM to reach a particular stage.  This results in each VM being torn down as soon as it has finished.  It also staggers diags collection, making better use of bandwidth.

## Related issues/PRs

CORE-12175
Follows on from #11398

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/.semaphore/vms/configure-test-vm
+++ b/.semaphore/vms/configure-test-vm
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright (c) 2019-2025 Tigera, Inc. All rights reserved.
+# Copyright (c) 2019-2026 Tigera, Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.semaphore/vms/run-tests-on-vms
+++ b/.semaphore/vms/run-tests-on-vms
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright (c) 2023-2025 Tigera, Inc. All rights reserved.
+# Copyright (c) 2023-2026 Tigera, Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


.semaphore/vms/vm-ip
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+# Copyright (c) 2023-2025 Tigera, Inc. All rights reserved.
```
nelljerram: 2026 now

.semaphore/vms/configure-test-vm
```diff
@@ -90,3 +85,7 @@ set -x
   tar -xzf /tmp/working-copy.tgz && \
   ${CALICO_DIR_NAME}/${COMPONENT}/.semaphore/load-test-artifacts"
 
+if [ "$ENABLE_JIT_HARDENING" = "true" ]; then
```
nelljerram, line 88: Is this a separate change, or intentionally part of this PR?  I don't think it's mentioned in the PR description.

.semaphore/vms/run-tests-on-vms
```diff
@@ -51,58 +117,93 @@ for r in "${log_monitor_regexps[@]}"; do
...
-echo
-
-pids=()
+test_pid=()
 monitor_pids=()
 log_files=()
-for batch in "${batches[@]}"; do
+
+for i in "${!batches[@]}"; do
+  batch=${batches[$i]}
```
nelljerram: I don't think you've used `i` anywhere below, so how is this better than `for batch in "${batches[@]}"; do` ?  It does look a little more obscure, I think.

nelljerram (commented):
Couple of queries and a nit.

.semaphore/vms/configure-test-vm
fasaxc: Yes, it used to be done in the semaphore yaml between the "config" step and the "run tests" step.  In order to inline the steps I had to make it part of the config script.

.semaphore/vms/run-tests-on-vms
fasaxc: I think it must have come from an intermediate state of the code where I needed the index as well as the name of the batch.  Reverted :+1:

PR #11598, opened 2025-12-29
CORE-11758: Added benchmark test for Goldmane.Receive function

skoryk-oleksandr:
## Description

**Add Benchmarks for Goldmane.Receive Performance**

This PR adds benchmark tests for the `Goldmane.Receive` ingestion path to track CPU and memory behavior in single-threaded and parallel execution.

**What’s included:**
- BenchmarkGoldmaneReceive — serial execution
- BenchmarkGoldmaneReceiveParallel — concurrent execution
- Heap usage reporting + custom metrics

| Metric            | Single | Parallel |
|-------------------|--------|---------|
| Max HeapAlloc MB  | 5.0    | 8.0     |
| Max ns/op (ref)   | 13,000 | 35,000  |

Tests fail if heap allocation exceeds limits to catch memory regressions early.

**How to run benchmarks**
-  `cd goldmane`
-  `make benchmark`

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/goldmane/pkg/goldmane/goldmane_benchmark_test.go
+++ b/goldmane/pkg/goldmane/goldmane_benchmark_test.go
@@ -15,7 +15,6 @@
 package goldmane_test
 
 import (
-	"github.com/projectcalico/calico/goldmane/pkg/testutils"
 	"math/rand/v2"
 	"os"
 	"runtime"
@@ -25,6 +24,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/calico/goldmane/pkg/goldmane"
+	"github.com/projectcalico/calico/goldmane/pkg/testutils"
 	"github.com/projectcalico/calico/goldmane/pkg/types"
 	"github.com/projectcalico/calico/lib/std/time"
 )


goldmane/pkg/goldmane/goldmane_benchmark_test.go
```diff
@@ -0,0 +1,275 @@
...
+	defer cleanup()
+
+	// Setup Goldmane
+	gm := goldmane.NewGoldmane()
+	now := time.Now().Unix()
+	<-gm.Run(now)
+	defer gm.Stop()
+
+	// Create a test flow
+	flow := types.ProtoToFlow(newRandomFlow(now))
```
caseydavenport: This is going to inject the same flow over and over - is that sufficient? We might want to generate new flows for each iteration (or at least some number of distinct flows greater than 1)

skoryk-oleksandr: @caseydavenport  Good catch :+1: . Fixed.

goldmane/pkg/goldmane/goldmane_benchmark_test.go
```diff
@@ -0,0 +1,307 @@
...
+	b.Logf("\t\tMax allowed B/op (Bytes/op):\t%.1f", maxBytesPerOp)
+	b.Logf("\t\tNote: ns/op is measured by Go's benchmark framework (shown in output above)")
+
+	// Fail the test if memory thresholds are exceeded
+	if heapAllocMB > maxHeapAllocMBParallel {
+		b.Fatalf("HeapAllocMB too high: got %.2f MB, max %.2f MB", heapAllocMB, maxHeapAllocMBParallel)
+	}
+}
+
+func newRandomFlow(start int64) *proto.Flow {
```
caseydavenport: We seem to have this logic replicated in a few places now - any reason not to import form testutils? 

https://github.com/projectcalico/calico/blob/d2e0cf9ee617b07f9419d2b7364e74d51355460c/goldmane/pkg/testutils/flows.go#L9-L114

skoryk-oleksandr: @caseydavenport  I kept the flow-generation logic in the benchmark to prevent dependencies on testutils, since future changes there could impact benchmark behavior and results. Let me know if you think it's overhead.

caseydavenport: I think it should be safe enough to use - especially since it's not called within the benchmark function itself, and any changes should show up in the tests!

skoryk-oleksandr: @caseydavenport Fixed :+1:

PR #11615, opened 2026-01-05
Add allowedsourceprefixes ebpf

### LINKED ISSUE #11591: allowedSourcePrefixes option is not available for eBPF mode
On certain cases, the traffic sent out from a pod will have a source IP different than the pod IP. One such case is where the pod acts as a proxy with IP_TRANSPARENT option enabled, using the client IP as the source IP of the packets. Currently, calico interfaces are configured to drop any traffic coming out of the pods that does not have the pod's IP as the source IP.  This issue seems to be fixed with the allowedSourcePrefixes annotation for iptables based Calico, but the same feature is not supported for the eBPF dataplane mode.

## Expected Behavior

Calico should have the option not to drop packets flowing out of the pod with a different source IP.

## Current Behavior

Calico drops packets going out of the pods with a different source IP. Setting `bpfLogLevel: Debug` displays the following in BPF tracelog

```
python3-2237505 [050] ..s2. 778538.987120: bpf_trace_printk: cali39c2a41992a-E: Workload RPF check src=<src-IP> skb iface=66.
python3-2237505 [050] ..s2. 778538.987120: bpf_trace_printk: cali39c2a41992a-E: Workload RPF fail: missing route.
python3-2237505 [050] ..s2. 778538.987122: bpf_trace_printk: cali39c2a41992a-E: Final result=DENY (0). Program execution time: 17473ns
```

## Possible Solution

allowedSourcePrefixes annotation should be implemented for eBPF mode as well.

1. Create a simple Python pod that acts as a proxy that preservers client IP with IP_TRANSPARENT option enabled. 
2. Set `bpfLogLevel: Debug` in felixconfiguration
3. Send Python packets to the pod
4. Run tcpdump both on the host machine running the pod and the pod itself
5. Listen to bpf logs. (E.g. command `tc exec bpf debug`

## Context

This bug arose as we were trying to implement source IP preservation with pods that act as a proxy. We encountered the issue both with an Nginx pod, as well as a custom Python script

## Your Environment

* Calico version - v3.29.2
* Calico dataplane - eBPF
* Orchestrator version - v1.31
* Operating System and version - Ubuntu 22.04

turkmenkaan:
## Description

This PR adds support for allowedSourcePrefixes annotation to Calico with eBPF mode. This annotation allows the pods to sent out traffic with a source IP other than the pod IP and the annotation is already implemented for non-eBPF Calico modes.

Calico node is the only component that is affected directly by the PR. However, there are changes at the eBPF layer that can affect traffic in the cluster.

The testing consisted of a TCP packet generating script and a UDP proxy with IP_TRANSPARENT option enabled. The testing setup will be detailed below, but the results are common for both manual tests.

**TCP Packet Generation**
A pod controlled by a DaemonSet running on one of the nodes generates empty TCP packets using Python & Scapy. The source IP is set to 192.192.192.192. The generated packets are sent to an external machine every second. A tcpdump instance is running on the target machine. 

**UDP Proxy**
A simple Python script that acts as a UDP proxy is deployed to the cluster with the updated calico-node image as a DaemonSet. Python script accepts incoming packets and uses IP_TRANSPARENT option to send them to the external machine while preserving the source IP. Once again, tcpdump is running on the target machine to observe incoming packets

**Results**
In both cases, before adding any annotations to the test pod, the egress pods are getting dropped and the following logs are observed in `bpftool prog tracelog`

```
python3-814830  [003] ..s2. 9696281.750213: bpf_trace_printk: calidce118d2ec5-E: Workload RPF check src=192.192.192.192 skb iface=82.
python3-814830  [003] ..s2. 9696281.750215: bpf_trace_printk: Allow source lookup src=192.192.192.192 found=0
python3-814830  [003] ..s2. 9696281.750215: bpf_trace_printk: calidce118d2ec5-E: Workload RPF fail: missing route.
```
*Second line was added temporarily for debugging purposes*

Once the following annotation `cni.projectcalico.org/allowedSourcePrefixes: '["192.192.192.192/32"]'` is added to the test pod, packets start reaching target machine. In addition, the following logs are observed.

```
python3-1203326 [001] ..s2. 9696468.029264: bpf_trace_printk: calia1d5ef599c0-E: Workload RPF check src=192.192.192.192 skb iface=83.
python3-1203326 [001] ..s2. 9696468.029267: bpf_trace_printk: Allow source lookup src=192.192.192.192 found=1
python3-1203326 [001] ..s2. 9696468.029268: bpf_trace_printk: calia1d5ef599c0-E: Workload RPF bypass: allowing spoofed source
```

Once the annotation is removed, the initial logs are observed again.

## Related issues/PRs

Fixes https://github.com/projectcalico/calico/issues/11591 

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
eBPF: implement allowedSourcePrefixes functionality 
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/rpf.h
```diff
@@ -18,10 +19,13 @@ static CALI_BPF_INLINE int wep_rpf_check(struct cali_tc_ctx *ctx, struct cali_rt
 {
         CALI_DEBUG("Workload RPF check src=" IP_FMT " skb iface=%d.",
                         debug_ip(ctx->state->ip_src), ctx->skb->ifindex);
-        if (!r) {
+        if (!r && !cali_allowsource_lookup(&ctx->state->ip_src, ctx->skb->ifindex)) {
```
tomastigera: This would only work for IPs that are not assigned to any other pod. If the IP is assigned to anything else that exists in the cluster (host or pod), you would have a route and your secondary check would not execute. But the RPF check would fail below because it would not be a local pod with the same interface index as your current endpoint.

tomastigera: If you only want to make it for IPs that are not part of the cluster, wouldn't it be easier to add a route to the bpf routes with the ifindex of this pod an then you would have a match, thus you would have a route and it all would work without any changes to the bpf code?

You would need to add logic to the bpf_routes_mgr to add the route, preferable to changing the bpf code.

If you need to allow IPs that may exist in the cluster, I think that would be also achievable via those routes, but that woudl require changes in many other places.

felix/bpf/allowsources/map.go
```diff
@@ -0,0 +1,99 @@
+package allowsources
```
tomastigera, line 1: allowsource is a very specific name. Could you think of a broader usecase?

felix/bpf-gpl/rpf.h
turkmenkaan: @tomastigera do we have the same constraint in iptables based implementation of allowed source prefixes? Does that only work for cluster-external IPs? 

I've also been in touch with @sridhartigera regarding this, tagging him here for visibility.

sridhartigera: > @tomastigera do we have the same constraint in iptables based implementation of allowed source prefixes? Does that only work for cluster-external IPs?
> 
> I've also been in touch with @sridhartigera regarding this, tagging him here for visibility.

@tomastigera As far as using route map is concerned, I am not sure if its a good fit as we can have multiple pods annotated with same spoof addresses.

felix/bpf/allowsources/map.go
turkmenkaan: Other options I can come up with are pretty specific since the purpose of this map is fairly specific too. I considered "rpfexceptions" etc.. Do you have any suggestions?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4606,6 +4615,122 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+        logrus.WithField("wep", wlID).WithField("cidr", prefix).WithError(mapDelError).Warn("Failed to remove allowed source CIDR")
+        return
+    }
+
+    m.allowedSourcesPerWorkload[wlID].Discard(prefix)
+
+    logrus.WithField("wep", wlID).WithField("cidr", prefix).Debug("Successfully removed allowed source CIDR")
+}
+
+func (m *bpfEndpointManager) adjustMapWithPrefix(prefix string, ifindex int, op string) error {
```
tomastigera:
```suggestion
func (m *bpfEndpointManager) changeAllowedSource(prefix string, ifindex int, op string) error {
```

something more concrete

felix/bpf/allowsources/map.go
tomastigera: perhaps something like 'sourcerules' :man_shrugging:

felix/bpf-gpl/rpf.h
tomastigera: I don't know what are the constraints, I am just flagging how the code will work :) Sounds like routes are not the right map to use.

felix/bpf-gpl/allowsources.h
```diff
@@ -0,0 +1,44 @@
...
+{
+    struct allow_sources_key k;
+#ifdef IPVER6
+    k.prefixlen = IPV6_ADDR_BITS + IFINDEX_BITS;
+#else
+    k.prefixlen = IPV4_ADDR_BITS + IFINDEX_BITS;
+#endif
+    k.addr = *addr;
+    k.ifindex = ifindex;
+    __u32 *val = cali_sprefix_lookup_elem(&k);
```
sridhartigera: you can simply return cali_sprefix_lookup_elem(&k)

felix/bpf/ut/spoof_test.go
```diff
@@ -28,6 +29,7 @@ var (
 	rtValGood       = routes.NewValueWithIfIndex(routes.FlagsLocalWorkload, 1).AsBytes()
 	rtValWrongIface = routes.NewValueWithIfIndex(routes.FlagsLocalWorkload, 2).AsBytes()
 	rtValWrongType  = routes.NewValueWithNextHop(routes.FlagsRemoteWorkload, ip.V4Addr{1, 0, 0, 0}).AsBytes()
+    allowSourcesKey = allowsources.NewKey(srcV4CIDR, 1).AsBytes()
```
sridhartigera: `rtKeySrc` and `allowSourcesKey` both use the same `srcV4CIDR`. Can you please specify a different src IP.?

turkmenkaan: I think we need to use the same srcV4CIDR here. AllowSourcesKey is used to update the new ebpf map with the source IP that packets are sent with. In this test case, packets are generated by `testPacketUDPDefault` within `runSpoofTest` and srcIP used there equals to srcV4CIDR. Using any other IP here would require changes in the underlying test helper functions without any real gain.

sridhartigera: Lets leave it as it is.

felix/bpf/allowsources/map.go
turkmenkaan: I like source rules. Lets us add any other functionality related to the source IP to the package. I'll change the package name to sourcerules, but keep the map name same as the map itself has a more specific purpose.

turkmenkaan: Now that I think about it, we can only have one map per package due to the current convention anyways. I think a name like sourcerules is too vague to describe the purpose of the map itself. What do you think?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4606,6 +4615,122 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+        managerDataplane = m.v4
+    } else if cidr.Version() == 6 {
+        entry = allowsources.NewKeyV6(cidr, ifindex)
+        managerDataplane = m.v6
+    }
+
+    if strings.ToLower(op) == "update" {
+        err = managerDataplane.AllowSourcesMap.Update(entry.AsBytes(), allowsources.DummyValue)
+    } else if strings.ToLower(op) == "delete" {
+        err = managerDataplane.AllowSourcesMap.Delete(entry.AsBytes())
```
sridhartigera: If `ipv6Enabled` is `false`, this will crash if the user adds a v6 cidr to the annotation. I think it will be better if `allowedSourcesPerWorkload` is added to `bpfEndpointManagerDataplane` struct. All updates should be placed under `if m.v4 != nil` and `if m.v6 != nil` as we have designed it for single stack ipv4, ipv6 or dual stack.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4606,6 +4615,122 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+
+func (m *bpfEndpointManager) removeAllowSourceSets(wlID types.WorkloadEndpointID, wl *proto.WorkloadEndpoint, prefix string) {
+    bpfInterface, exists := m.nameToIface[wl.Name]
+    if !exists {
+        logrus.WithField("wep", wlID).WithField("iface", wl.Name).Warn("Interface for workload endpoint not found, cannot remove allowed source CIDRs")
+        return
+    }
+
+    ifindex := bpfInterface.info.ifIndex
+    if ifindex <= 0 {
```
sridhartigera: I don't think this check is required.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4606,6 +4615,122 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+
+    if cidr.Version() == 4 {
+        entry = allowsources.NewKey(cidr, ifindex)
+        managerDataplane = m.v4
+    } else if cidr.Version() == 6 {
+        entry = allowsources.NewKeyV6(cidr, ifindex)
+        managerDataplane = m.v6
+    }
+
+    if strings.ToLower(op) == "update" {
```
sridhartigera: Can we just do it with a `bool`? something like `isAdd` instead of `update` and `delete`?

sridhartigera (changes requested):

felix/dataplane/linux/bpf_ep_mgr.go
turkmenkaan: I noticed calico-node receiving updates with ifIndex = 0. This is to avoid invalid values to the map. Not sure why we received extra WorkloadEndpointUpdate without valid indexes though. It might have been the IfaceUpdates

felix/dataplane/linux/bpf_ep_mgr.go
turkmenkaan: done

felix/dataplane/linux/bpf_ep_mgr.go
turkmenkaan: Not sure if you meant to add this comment to that code block but I gathered that you were talking about all allowedSourcesPerWorkload operations. Moved that map under the dataplane struct and added helper functions to control the operations that add/remove/update those maps. Also added dual-stack test cases to bpf_ep_mgr_test.go.

kaan-turkmen:
PR updated with test cases and conntrack entry creation surpression @tomastigera @sridhartigera

sfakam:
is there an ETA on this PR ?  once this will unblock the use case for us

felix/bpf/ut/spoof_test.go
```diff
@@ -72,6 +72,17 @@ func TestWorkloadSpoof(t *testing.T) {
 	Expect(err).NotTo(HaveOccurred())
     runSpoofTest(t, resTC_ACT_UNSPEC)
     cleanUpMaps()
+
```
sridhartigera, line 101: Think it will be better to check if there is no ct entry.

turkmenkaan: This test specifically will create CT entry as route exists. I added a check to confirm no CT entry is being created to the test above where we are missing a return route.

kaan-turkmen:
@sridhartigera formatting errors due to the master merge are fixed now, can we run the pipeline again please?

felix/bpf-gpl/rpf.h
```diff
@@ -16,12 +18,20 @@
...
 static CALI_BPF_INLINE int wep_rpf_check(struct cali_tc_ctx *ctx, struct cali_rt *r)
 {
+		bool allowsource_lookup = false;
+		if (WORKLOAD_SRC_SPOOFING_CONFIGURED) {
+			allowsource_lookup = cali_allowsource_lookup(&ctx->state->ip_src, ctx->skb->ifindex);
+		}
         CALI_DEBUG("Workload RPF check src=" IP_FMT " skb iface=%d.",
                         debug_ip(ctx->state->ip_src), ctx->skb->ifindex);
-        if (!r) {
+        if (!r && !allowsource_lookup) {
```
sridhartigera:
```suggestion
        if (!r) {
            if (WORKLOAD_SRC_SPOOFING_CONFIGURED && !cali_allowsource_lookup(&ctx->state->ip_src, ctx->skb->ifindex)) {
                CALI_INFO("Workload RPF fail: missing route.");
                return RPF_RES_FAIL;
            }
            CALI_INFO("Workload RPF bypass: allowing spoofed source IP");
            ctx->state->flags |= CALI_ST_SUPPRESS_CT_STATE;
            return RPF_RES_STRICT;
         }
            
```

sridhartigera: you can just remove 
```bool allowsource_lookup = false;
		if (WORKLOAD_SRC_SPOOFING_CONFIGURED) {
			allowsource_lookup = cali_allowsource_lookup(&ctx->state->ip_src, ctx->skb->ifindex);
		}
```

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4602,6 +4614,201 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+		}
+	}
+
+	for existingPrefix := range oldPrefixes.All() {
+		if !slices.Contains(wl.AllowSpoofedSourcePrefixes, existingPrefix) {
+			// remove stale annotations from bookkeeping map AND eBPF map
+			m.removeAllowSourceSets(wlID, wl, existingPrefix)
+		}
+	}
+
```
sridhartigera: Please remove new line

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4602,6 +4614,201 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+		return // if not an actual interface, don't add it to the map
+	}
+
+	mapAddError := m.changeAllowedSource(prefix, ifindex, true)
+	if mapAddError != nil {
+		logrus.WithField("wep", wlID).WithField("cidr", prefix).WithError(mapAddError).Warn("Failed to add allowed source CIDR")
+		return
+	}
+	m.setAllowedSourcesPerWorkload(wlID, prefix)
+
```
sridhartigera: Remove new line

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4602,6 +4614,201 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	logrus.WithField("wep", wlID).WithField("cidr", prefix).Debug("Successfully added allowed source CIDR")
+}
+
+func (m *bpfEndpointManager) cleanAllowSourceSetPerWorkload(wlID types.WorkloadEndpointID, wl *proto.WorkloadEndpoint) {
+	wlPrefixes, exists := m.getAllowedSourcesPerWorkload(wlID)
+	if !exists {
+		return
+	}
+
```
sridhartigera: Remove new line

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4602,6 +4614,201 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	if ifindex <= 0 {
+		return // if not an actual interface, don't add it to the map
+	}
+
+	mapDelError := m.changeAllowedSource(prefix, ifindex, false)
+	if mapDelError != nil {
+		logrus.WithField("wep", wlID).WithField("cidr", prefix).WithError(mapDelError).Warn("Failed to remove allowed source CIDR")
+		return
+	}
+
```
sridhartigera: Remove new line

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4602,6 +4614,201 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	m.discardFromAllowedSourcesPerWorkload(wlID, prefix)
+
```
sridhartigera: Remove new line

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4602,6 +4614,201 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+		entry = allowsources.NewKeyV6(cidr, ifindex)
+		managerDataplane = m.v6
+	}
+
+	if isAdd {
+		err = managerDataplane.AllowSourcesMap.Update(entry.AsBytes(), allowsources.DummyValue)
+	} else {
+		err = managerDataplane.AllowSourcesMap.Delete(entry.AsBytes())
+	}
+
```
sridhartigera: Remove new line

felix/dataplane/linux/bpf_ep_mgr_test.go
```diff
@@ -2864,6 +2888,214 @@ var _ = Describe("BPF Endpoint Manager", func() {
...
+		var (
+			ifaceName  string
+			cidr       ip.CIDR
+			cidrString string
+		)
+
+		JustBeforeEach(func() {
+			ifaceName = "cali12345"
+			cidrString = "2001:db8::1/128"
+
```
sridhartigera: Remove new line

felix/dataplane/linux/bpf_ep_mgr_test.go
```diff
@@ -2864,6 +2888,214 @@ var _ = Describe("BPF Endpoint Manager", func() {
...
+			genWLUpdateWithSpoofedSource("cali12345", []string{cidrString})()
+
+			var err error
+			cidr, err = ip.CIDRFromString(cidrString)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("should contain allowed source prefixes from annotation", func() {
+			// initial genWLUpdate should add the prefix to the map
+
```
sridhartigera, line 2980: Remove new line

felix/dataplane/linux/bpf_ep_mgr_test.go
```diff
@@ -2864,6 +2888,214 @@ var _ = Describe("BPF Endpoint Manager", func() {
...
+			bookkeepingSet, exists = bpfEpMgr.v6.allowedSourcesPerWorkload[workloadEndpointID]
+			Expect(exists).To(BeTrue())
+			Expect(bookkeepingSet.Len()).To(Equal(1))
+			Expect(bookkeepingSet.Contains("::1/128")).To(BeTrue())
+		})
+
+		It("should remove prefix from the map upon removal", func() {
+			// the following prefix should be removed from the map once WLUpdateEpRemove runs
+			genWLUpdateEpRemove("cali12345")()
+
```
sridhartigera: Remove new line

felix/dataplane/linux/bpf_ep_mgr_test.go
```diff
@@ -2864,6 +2888,214 @@ var _ = Describe("BPF Endpoint Manager", func() {
...
+			ifindex := bpfEpMgr.nameToIface[ifaceName].info.ifIndex
+
```
sridhartigera: Same here

sridhartigera (commented):
I noticed a whole bunch of new lines. Please fix them. Only suggestion in wep_rpf_check.

felix/bpf-gpl/rpf.h
turkmenkaan: Suggested change above would cause RPF filter to be bypassed when route is not found but WORKLOAD_SRC_SPOOFING_CONFIGURED set to false.

turkmenkaan:
> I noticed a whole bunch of new lines. Please fix them. Only suggestion in wep_rpf_check.

New lines are now fixed.

felix/bpf-gpl/rpf.h
turkmenkaan: I believe what you meant was the following. I'll run some more tests and update with this.

```
        CALI_DEBUG("Workload RPF check src=" IP_FMT " skb iface=%d.",
                        debug_ip(ctx->state->ip_src), ctx->skb->ifindex);
        if (!r) {
				if (WORKLOAD_SRC_SPOOFING_CONFIGURED && cali_allowsource_lookup(&ctx->state->ip_src, ctx->skb->ifindex)) {
					CALI_INFO("Workload RPF bypass: allowing spoofed source IP");
					ctx->state->flags |= CALI_ST_SUPPRESS_CT_STATE;
					return RPF_RES_STRICT;
				}
                CALI_INFO("Workload RPF fail: missing route.");
                return RPF_RES_FAIL;
		}
```

turkmenkaan: Updated

turkmenkaan:
merge conflicts fixed

sridhartigera:
Please go through the comments from copilot. Guess some of them are valid.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4611,6 +4623,195 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+		entry = allowsources.NewKeyV6(cidr, ifindex)
+		managerDataplane = m.v6
+	}
+
+	if isAdd {
+		err = managerDataplane.AllowSourcesMap.Update(entry.AsBytes(), allowsources.DummyValue)
+	} else {
+		err = managerDataplane.AllowSourcesMap.Delete(entry.AsBytes())
+	}
+	return err
```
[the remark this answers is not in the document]
turkmenkaan: As per https://pkg.go.dev/github.com/projectcalico/calico/felix/ip#V4CIDR, cidr.Version() can only be 4 or 6. We  can add an else clause just for logging purposes, but it should never hit that else clause.

Dataplane being nil seems like a legit issue (if an IPv4 is added to IPv6 dataplane or vice versa). This currently results in a segfault, and puts the pod in not ready state. Instead, I'll ignore the annotations with the incorrect source.

felix/bpf-gpl/globals.h
```diff
@@ -62,6 +62,7 @@ enum cali_globals_flags {
 	CALI_GLOBALS_SKIP_EGRESS_REDIRECT                 = 0x00002000,
 	CALI_GLOBALS_INGRESS_PACKET_RATE_CONFIGURED       = 0x00004000,
 	CALI_GLOBALS_EGRESS_PACKET_RATE_CONFIGURED        = 0x00008000,
+	CALI_GLOBALS_WORKLOAD_SRC_SPOOFING_CONFIGURED     = 0x00010000,
 	CALI_GLOBALS_UDP_GSO_LINEARIZE                    = 0x00010000,
```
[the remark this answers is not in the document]
turkmenkaan: Cause by the latest merge conflict, will be fixed

felix/bpf/allowsources/map.go
```diff
@@ -0,0 +1,105 @@
...
+type AllowSourcesEntry [KeySize]byte
+
+var MapParameters = maps.MapParameters{
+	Type:       "lpm_trie",
+	KeySize:    KeySize,
+	ValueSize:  4,
+	MaxEntries: 256 * 1024,
+	Name:       "cali_v4_sprefix",
+	Flags:      unix.BPF_F_NO_PREALLOC,
+}
```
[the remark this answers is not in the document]
turkmenkaan: Doesn't prevent program from loading as suggested by copilot, but will ensure they match regardless

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4611,6 +4623,195 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	if !exists {
+		logrus.WithField("iface", wl.Name).Warn("Interface for workload endpoint not found, cannot add allowed source CIDRs")
+		return
+	}
+
+	ifindex := bpfInterface.info.ifIndex
+	if ifindex <= 0 {
+		return // if not an actual interface, don't add it to the map
+	}
+
```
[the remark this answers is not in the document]
turkmenkaan: Not applicable

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4611,6 +4623,195 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	mapDelError := m.changeAllowedSource(prefix, ifindex, false)
+	if mapDelError != nil {
+		logrus.WithField("wep", wlID).WithField("cidr", prefix).WithError(mapDelError).Warn("Failed to remove allowed source CIDR")
+		return
+	}
+	m.discardFromAllowedSourcesPerWorkload(wlID, prefix)
+	logrus.WithField("wep", wlID).WithField("cidr", prefix).Debug("Successfully removed allowed source CIDR")
+}
```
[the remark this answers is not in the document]
turkmenkaan: Not applicable

sridhartigera (approved):
LGTM

PR #11619, opened 2026-01-05
Record first fragment if bypass mark is set.

sridhartigera:
## Description

This is a cherry-pick of https://github.com/tigera/calico-private/pull/10400.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -1372,12 +1372,6 @@ int calico_tc_skb_accepted_entrypoint(struct __sk_buff *skb)
 		skb_log(ctx, true);
 	}
 
-#ifndef IPVER6
-	if ((CALI_F_FROM_HOST || CALI_F_FROM_WEP) && ip_is_first_frag(ip_hdr(ctx))) {
-		frags4_record_ct(ctx);
-	}
-#endif
-
 	if ((CALI_F_FROM_WEP || CALI_F_TO_HEP) && qos_dscp_needs_update(ctx) && !qos_dscp_set(ctx)) {
 		goto deny;
 	}
@@ -1779,11 +1773,6 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct cali_tc_ctx *ctx
 	case CALI_CT_ESTABLISHED_BYPASS:
 		if (!ct_result_is_syn(state->ct_result.rc)) {
 			seen_mark = CALI_SKB_MARK_BYPASS;
-#ifndef IPVER6
-			if (ip_is_first_frag(ip_hdr(ctx))) {
-				frags4_record_ct_mark(ctx, seen_mark);
-			}
-#endif
 			CALI_DEBUG("marking CALI_SKB_MARK_BYPASS");
 		}
 		// fall through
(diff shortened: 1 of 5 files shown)


felix/bpf-gpl/bpf.h
```diff
@@ -199,7 +199,9 @@ enum calico_skb_mark {
 	CALI_SKB_MARK_BYPASS_FWD             = CALI_SKB_MARK_BYPASS  | 0x00300000,
 	/* Accepted by XDP untracked policy. */
 	CALI_SKB_MARK_BYPASS_XDP             = CALI_SKB_MARK_BYPASS  | 0x00500000,
-	CALI_SKB_MARK_BYPASS_MASK            = CALI_SKB_MARK_SEEN_MASK | 0x02700000,
+	/* Mark used for fragmented packets. */
+	CALI_SKB_MARK_BYPASS_FRAG            = CALI_SKB_MARK_BYPASS  | 0x00400000,
```
tomastigera: do we really have enough values? seems like we have (to my big surprise :sweat_smile: )

felix/bpf-gpl/tc.c
```diff
@@ -77,12 +77,8 @@ int calico_tc_main(struct __sk_buff *skb)
...
 	/* Optimisation: if another BPF program has already pre-approved the packet,
 	 * skip all processing. */
-	if (CALI_F_FROM_HOST && skb_mark_equals(skb, CALI_SKB_MARK_BYPASS, CALI_SKB_MARK_BYPASS) &&
-			/* If we are on tunnel and we do not have the key set, we cannot short-cirquit */
-			!(CALI_F_TUNNEL &&
-			 !skb_mark_equals(skb, CALI_SKB_MARK_TUNNEL_KEY_SET, CALI_SKB_MARK_TUNNEL_KEY_SET))) {
-		if (CALI_LOG_LEVEL >= CALI_LOG_LEVEL_DEBUG) {
-			/* This generates a bit more richer output for logging */
+	if (CALI_F_FROM_HOST && skb_mark_equals(skb, CALI_SKB_MARK_BYPASS, CALI_SKB_MARK_BYPASS)) {
+		if (skb_mark_equals(skb, CALI_SKB_MARK_BYPASS_FRAG, CALI_SKB_MARK_BYPASS_FRAG)) {
```
tomastigera: I think this would be also true for CALI_SKB_MARK_MASQ I think you need to consider more bits at the same time.

felix/bpf-gpl/tc.c
```diff
@@ -1775,7 +1801,11 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct cali_tc_ctx *ctx
 
 	case CALI_CT_ESTABLISHED_BYPASS:
 		if (!ct_result_is_syn(state->ct_result.rc)) {
-			seen_mark = CALI_SKB_MARK_BYPASS;
+			if (ip_is_frag(ip_hdr(ctx))) {
```
tomastigera: wouldn't it be better to test `if (ip_is_first_frag(ip_hdr(ctx))) {` also right here? Then you would not need to do it before you do the parsing etc. And then perhaps not to set the bypass instead of setting a new flag?

tomastigera (approved):

PR #11622, opened 2026-01-06
Prepare types for CRD generation

radixo:
## Description

It will be built a new calico/go-build image with a controller-gen without patches for numorstring types, those changes prepares calico repository to be built with this changes.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Prepare types for CRD generation
```

## Reminder for the reviewer

Other optional labels:


hjiawei:
Related toolchain changes in https://github.com/projectcalico/toolchain/pull/754.

api/pkg/lib/numorstring/port.go
```diff
@@ -28,6 +28,10 @@ import (
 //   - For a port range, set MinPort and MaxPort to the (inclusive) port numbers.  Set
 //     PortName to "".
 //   - For a single port, set MinPort = MaxPort and PortName = "".
+//
+// +kubebuilder:validation:Type=integer
```
hjiawei, line 32: Does this change force the type to be an integer only?

radixo: No, just set type to Integer, and the XIntOrString relax it to accept string as well. I'm using the same for DSCP

libcalico-go/config/crd/crd.projectcalico.org_felixconfigurations.yaml
```diff
@@ -429,6 +426,7 @@ spec:
                     a problem if this range overlaps with the operating systems. Both ends of the range are
                     inclusive. [Default: 20000:29999]
                   pattern: ^.*
+                  type: integer
```
hjiawei, line 429: Are these breaking changes @radixo ? Same question for other places.

felix/docs/config-params.json
```diff
@@ -1826,7 +1826,7 @@
           "ParsedDefault": "0",
           "ParsedDefaultJSON": "0",
           "ParsedType": "numorstring.Port",
-          "YAMLType": "integer or string",
+          "YAMLType": "integer",
```
hjiawei, line 1881: The felix doc for `YAMLType` is affected by the `+kubebuilder:validation:Type=integer` marker.

radixo:
Doing an alternative solution.

PR #11629, opened 2026-01-07
Auto-detect nft from kube-proxy

caseydavenport:
## Description

This PR makes it easier to switch to nftables mode on a live system, by having each node capable of auto-detecting whether to use iptables or nftables based on the local kube-proxy.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Support auto-detection of nftables or iptables based on kube-proxy configuration. 
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/config_update_test.go
+++ b/felix/fv/config_update_test.go
@@ -223,6 +223,10 @@ var _ = infrastructure.DatastoreDescribe("Config update tests, after starting fe
 		}
 
 		BeforeEach(func() {
+			if NFTMode() {
+				Skip("Skipping auto-detection tests when felix is already explicitly in nftables mode")
+			}
+
 			cfgChangeTime = time.Now()
 
 			// Create nftables rules in the "kube-proxy" table, which should trigger felix restart.


felix/config/config_params.go
```diff
@@ -182,7 +182,7 @@ type Config struct {
 	WireguardThreadingEnabled      bool          `config:"bool;false"`
 
 	// nftables configuration.
-	NFTablesMode string `config:"oneof(Enabled,Disabled);Disabled"`
+	NFTablesMode string `config:"oneof(Enabled,Disabled,Auto);Disabled"`
```
coutinhop: shouldn't it be `config:"oneof(Enabled,Disabled,Auto);Auto"` here? or am I off?

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1744,6 +1787,90 @@ func (d *InternalDataplane) checkIPVSConfigOnStateUpdate(state ifacemonitor.Stat
...
 
+// checkKubeProxyNftablesMode checks if kube-proxy's nftables mode has changed since the last check, and
+// triggers a Felix restart if it has and nftables mode is set to Auto.
+func (d *InternalDataplane) checkKubeProxyNftablesMode() {
+	if d.config.RulesConfig.NFTablesMode != "Auto" {
+		// We can skip this check if nftables is not configured to Auto.
+		return
+	}
+	if d.getKubeProxyNftablesEnabled == nil {
+		log.Fatal("BUG: kube-proxy nftables mode check function is nil")
```
fasaxc:
```suggestion
		log.Panic("BUG: kube-proxy nftables mode check function is nil")
```
Best to use panic, not fatal. If you call fatal in a UT without hooking it then it exits the process, whereas panic is caught and logged I think.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1744,6 +1787,90 @@ func (d *InternalDataplane) checkIPVSConfigOnStateUpdate(state ifacemonitor.Stat
...
+
+// kubeProxyNftablesFn returns a function that can be used to check if kube-proxy is running
+// in nftables mode. It does this by checking for the presence of the nftables chains that
+// kube-proxy creates when running in nftables mode.
+func kubeProxyNftablesFn(config Config) func() (bool, error) {
+	// Create v4 and v6 nftables interfaces. If either succeeds, we can use it to check for
+	// the presence of the kube-proxy nftables table.
+	nft, err := config.NewNftablesDataplane(knftables.IPv4Family, "kube-proxy")
+	if err != nil {
+		log.WithError(err).Warn("Failed to create nftables interface to check kube-proxy mode.")
```
fasaxc: Deliberately not returning an error?

fasaxc: Oh I see, nft will be nil so the returned func will spot it.  Probably worth a `// Handled below` comment

felix/dataplane/linux/int_dataplane.go
```diff
@@ -2491,6 +2618,10 @@ func (d *InternalDataplane) apply() {
 	// Creating a rule that references an unknown IP set fails, as does deleting an IP set that
 	// is in use.
 
+	// First, check if the dataplane mode has changed since the last apply, and restart if so.
+	// This catches when kube-proxy has changed between iptables/nftables mode on a live system.
+	d.checkKubeProxyNftablesMode()
```
fasaxc: This feels like the wrong cadence to me; apply() calls can be 10/s or, on a quiet system, once every 5 minutes.

Perhaps start a background thread that checks on a timer and sends a channel notification if it changes?

caseydavenport: Yeah, I went back and forth on this one. Happy to switch it over, though. I think you're right this could be a problem for high scale clusters.

PR #11645, opened 2026-01-12
Copy e2es from master into v3.31

lwr20:
## Description
This PR copies the contents of the .semaphore/end-to-end directory from master, plus a small change to change the default RELEASE_STREAM to v3.31.

This will allow us to run e2es against this branch, from pipelines in the branch.

This sort of PR should be unnecessary in future, since new releases are branched off master (which already has the e2es)

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/.semaphore/end-to-end/pipelines/patch-verification.yml
+++ b/.semaphore/end-to-end/pipelines/patch-verification.yml
@@ -126,7 +126,7 @@ blocks:
               value: ""
             - name: INSTALLER
               value: "helmerator"
-            - name: ENABLE_AUTO_HEP # autohep
+            - name: ENABLE_AUTOHEP # autohep
               value: "true"
 
   - name: "Manifests"
@@ -333,8 +333,6 @@ blocks:
             #   value: "private"    # I couldn't get this to provision, so I've skipped this for now
             - name: PROVISIONER
               value: "azr-aks"
-            - name: K8S_VERSION
-              value: "stable-1"
             - name: AZ_CREATE_MODE
               value: AKS
             - name: K8S_VERSION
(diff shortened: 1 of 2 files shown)


.semaphore/end-to-end/pipelines/benchmarking.yml
```diff
@@ -0,0 +1,103 @@
...
+          value: "CalicoIptables"
+        - name: PROVISIONER
+          value: "aws-kubeadm"
+        - name: INSTALLER
+          value: "operator"
+        - name: TEST_TYPE
+          value: "benchmark"
+        - name: BENCHMARKS
+          value: "benchmarker"
+        - name: BENCHMARKER_TEST_CONFIGS
```
sujeet-kr, line 89: Is this here because different clusters are getting created for iptables and ebpf?

lwr20: I'm not sure I understand the question.  

BENCHMARKER_TEST_CONFIGS contains json which defines a list of tests to run, which includes defining the dataplane.  The benchmark tool takes care of reconfiguring the cluster to run each test.

sujeet-kr: Should this be included in a separate config file in that case, instead of calling it out in the CI pipeline file.

lwr20: Possibly, but that would require changes in banzai-core.  At present this is how it expects to get the config.

PR #11660, opened 2026-01-14
[BPF]  Maglev Prometheus Metrics:  Connection counts

aaaaaaaalex:
## Description

Count how many packets get load balanced to backends both remote and local.
Adds a goroutine to bpf_ep_mgr which queries the counters map and exports the new Maglev counters as Prom metrics

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
[BPF] Adds new Prometheus metric: `felix_bpf_conntrack_maglev_entries_total` for maglev connection count.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/conntrack/scanner.go
+++ b/felix/bpf/conntrack/scanner.go
@@ -125,6 +125,9 @@ type Scanner struct {
 	revNATKeyToFwdNATInfo        map[KeyInterface]cleanupv1.ValueInterface
 	ipFamily                     int
 
+	conntrackGaugeMaglevToLocalBackend  prometheus.Gauge
+	conntrackGaugeMaglevToRemoteBackend prometheus.Gauge
+
 	wg       sync.WaitGroup
 	stopCh   chan struct{}
 	stopOnce sync.Once
@@ -171,6 +174,19 @@ func NewScanner(ctMap maps.Map, kfb func([]byte) KeyInterface, vfb func([]byte) 
 
 		}
 	}
+
+	var err error
+
+	s.conntrackGaugeMaglevToLocalBackend, err = conntrackGaugeMaglevTotal.GetMetricWithLabelValues("local", strconv.Itoa(s.ipFamily))
+	if err != nil {
+		log.WithError(err).Panic("Couldn't init (local) Maglev conntrack metric gauge")
+	}
+
+	s.conntrackGaugeMaglevToRemoteBackend, err = conntrackGaugeMaglevTotal.GetMetricWithLabelValues("remote", strconv.Itoa(s.ipFamily))
+	if err != nil {
+		log.WithError(err).Panic("Couldn't get (remote) Maglev conntrack metric gauge")
+	}
+
 	return s
 }
 
@@ -340,21 +356,10 @@ func (s *Scanner) Scan() {
 	// Run the bpf cleaner to process the remaining entries in the cleanup map.
 	cleaned += s.runBPFCleaner()
 
-	maglevConntracksToLocalBackend, err := conntrackGaugeMaglevTotal.GetMetricWithLabelValues("local", strconv.Itoa(s.ipFamily))
-	if err != nil {
-		log.WithError(err).Warn("Couldn't get (local) Maglev conntracks metric, will not update it on this iteration")
-	} else {
-		log.WithField("value", maglevEntriesToLocal).Debug("Setting local maglev conntrack entries gauge")
-		maglevConntracksToLocalBackend.Set(float64(maglevEntriesToLocal))
-	}
-
-	maglevConntracksToRemoteBackend, err := conntrackGaugeMaglevTotal.GetMetricWithLabelValues("remote", strconv.Itoa(s.ipFamily))
-	if err != nil {
-		log.WithError(err).Warn("Couldn't get (remote) Maglev conntracks metric, will not update it on this iteration")
-	} else {
-		log.WithField("value", maglevEntriesToRemote).Debug("Setting remote maglev conntrack entries gauge")
-		maglevConntracksToRemoteBackend.Set(float64(maglevEntriesToRemote))
-	}
+	log.WithField("value", maglevEntriesToLocal).Debug("Setting local maglev conntrack entries gauge")
+	s.conntrackGaugeMaglevToLocalBackend.Set(float64(maglevEntriesToLocal))
+	log.WithField("value", maglevEntriesToRemote).Debug("Setting remote maglev conntrack entries gauge")
+	s.conntrackGaugeMaglevToRemoteBackend.Set(float64(maglevEntriesToRemote))
 
 	conntrackCounterSweeps.Inc()
 	conntrackGaugeUsed.Set(float64(used))


felix/bpf-gpl/counters.h
fasaxc, line 24: Rev version since you changed the value size?

felix/bpf-gpl/fib_co_re.h
```diff
@@ -572,6 +572,16 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
 		}
 	}
 
+	if (CALI_F_TO_HOST && rc != TC_ACT_SHOT
+			&& ctx->state->ct_result.flags & CALI_CT_FLAG_MAGLEV) {
+		struct cali_rt *rt = cali_rt_lookup(&ctx->state->post_nat_ip_dst);
```
fasaxc: An extra route lookup for every packet might slow things down.  Can you not add this at the point that we would put it in the tunnel (or not); seems like there must already be a point int he code where we know if we're going to encap it and forward vs sending to local pod.

aaaaaaaalex: Was thinking the same. I've forgotten now what stopped me placing it near a pre-existing rt-lookup. I'll give it another try.

felix/bpf-gpl/fib_co_re.h
```diff
@@ -572,6 +572,23 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
...
+
+		int ret_from_tun = CALI_F_FROM_HEP &&
+			!ip_void(ctx->state->tun_ip) &&
+			ct_result_rc(ctx->state->ct_result.rc) == CALI_CT_ESTABLISHED_DNAT &&
+			ctx->state->ct_result.flags & CALI_CT_FLAG_NP_FWD;
+
+		if (ctx->state->ct_result.flags & CALI_CT_FLAG_EXT_LOCAL)  {
+			CALI_INFO("Counting Maglev packet forwarded to local workload");
+			counter_inc(ctx, CALI_REASON_MAGLEV_FORWARDED_TO_LOCAL);
+		} else if (ctx->state->ct_result.flags & CALI_CT_FLAG_NP_FWD && !ret_from_tun) {
```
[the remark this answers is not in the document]
tomastigera: i'd agree here, put brackets around, who knows what the operator prio is

felix/bpf-gpl/fib_co_re.h
```diff
@@ -588,6 +572,23 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
 		}
 	}
 
+	if (ctx->state->ct_result.flags & CALI_CT_FLAG_MAGLEV &&
+		CALI_F_TO_HOST && CALI_F_FROM_HEP && rc != TC_ACT_SHOT ) {
```
tomastigera: from HEP means to HOST

```suggestion
		CALI_F_FROM_HEP && rc != TC_ACT_SHOT ) {
```

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4606,6 +4613,57 @@ func (m *bpfEndpointManager) getIfaceTypeFromLink(link netlink.Link) IfaceType {
...
+	}
+	remote, err := bpfMaglevPacketsTotal.GetMetricWithLabelValues("to_backend", "remote")
+	if err != nil {
+		logrus.WithError(err).Warn("Error getting Maglev remote metric")
+	}
+
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+
+	logrus.Debug("Starting maglev metrics loop...")
```
tomastigera: why don't we also export other counters like this? Sounds like it would be handy, especially exporting errors.

felix/bpf/conntrack/scanner.go
```diff
@@ -231,10 +241,19 @@ func (s *Scanner) Scan() {
 	err := s.ctMap.Iter(func(k, v []byte) maps.IteratorAction {
 		ctKey := s.keyFromBytes(k)
 		ctVal := s.valueFromBytes(v)
+		ctFlags := ctVal.Flags()
 
 		used++
 		conntrackCounterCleaned.Inc()
 
+		if ctFlags&v4.FlagMaglev != 0 {
```
tomastigera, line 272: I think this should move to `case ScanVerdictOK` as you probably only care about active connections. This would also count connections that are expired and are going to be removed or do not have backed anymore etc.

Maybe you want to do it in the cleaner since you may not want to count connections in some FIN-wait state, where they may be for quite a while, but for the purpose of "can I remove this node" they do not matter anymore.

When @sridhartigera makes the change to send RSTs when the backend is no more, this would perhaps need to move again 🤔 

Obviously, the stats will be fixed up follow up iterations when connection is deleted.

felix/fv/bpf_test.go
```diff
@@ -2087,9 +2106,23 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 					})
 
 					It("should have connectivity from external client to maglev backend via cluster IP and external IP", func() {
+						Eventually(func() []int {
```
tomastigera: I think testing this in the scanner/cleaner UT would be enough and more practical.

felix/bpf/conntrack/scanner.go
```diff
@@ -54,6 +56,10 @@ var (
 		Name: "felix_bpf_conntrack_sweep_duration",
 		Help: "Conntrack sweep execution time (ns)",
 	})
+	conntrackGaugeMaglevTotal = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "felix_bpf_conntrack_maglev_entries_total",
+		Help: "Total number of Maglev entries in conntrack table.",
```
fasaxc:
```suggestion
		Help: "Total number of Maglev entries in conntrack table broken down by IP version, and, whether destination backend is remote (we're acting as a frontend) or local (we're the backend node).",
```

felix/fv/bpf_test.go
```diff
@@ -2087,9 +2106,23 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+							return probeMaglevConntrackMetric(tc.Felixes, fmt.Sprintf("felix_bpf_conntrack_maglev_entries_total{destination=\"local\",ip_family=\"%d\"}", familyInt))
+						}, "10s", "1s").Should(BeEquivalentTo([]int{0, 0, 0}), "Expected maglev local-conntrack metric to start at 0")
```
fasaxc:
```suggestion
						}, "10s", "1s").Should(Equal([]int{0, 0, 0}), "Expected maglev local-conntrack metric to start at 0")
```
I don't think you need the extra power of BeEquivalentTo here, so best to use the less surprising matcher.

felix/fv/bpf_test.go
```diff
@@ -2087,9 +2106,23 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
 						cc.ExpectSome(externalClient, TargetIP(externalIP), port)
 						cc.CheckConnectivity()
+
+						Eventually(func() []int {
+							return probeMaglevConntrackMetric(tc.Felixes, fmt.Sprintf("felix_bpf_conntrack_maglev_entries_total{destination=\"local\",ip_family=\"%d\"}", familyInt))
+						}, "10s").ShouldNot(BeEquivalentTo([]int{0, 0, 0}), "Expected maglev local-conntrack metric to have increased")
+						Eventually(func() []int {
+							return probeMaglevConntrackMetric(tc.Felixes, fmt.Sprintf("felix_bpf_conntrack_maglev_entries_total{destination=\"remote\",ip_family=\"%d\"}", familyInt))
+						}, "10s", "1s").ShouldNot(BeEquivalentTo([]int{0, 0, 0}), "Expected maglev remote-conntrack metric to have increased")
 					})
```
fasaxc: Here, it'd be good to test:
* That the counters increase instead of going negative or something
* That the ingress node showed a positive remote value
* That the backend node showed a positive local value.
* Rest show zero.

felix/bpf/conntrack/scanner.go
```diff
@@ -319,6 +340,22 @@ func (s *Scanner) Scan() {
 	// Run the bpf cleaner to process the remaining entries in the cleanup map.
 	cleaned += s.runBPFCleaner()
 
+	maglevConntracksToLocalBackend, err := conntrackGaugeMaglevTotal.GetMetricWithLabelValues("local", strconv.Itoa(s.ipFamily))
+	if err != nil {
+		log.WithError(err).Warn("Couldn't get (local) Maglev conntracks metric, will not update it on this iteration")
```
fasaxc: This isn't really something we should handle; it generally indicates a coding bug (e.g. passing two labels when only one is defined).  It's a minor thing, but I'd do the `GetMetricWithLabelValues` at init time and store off the sub-gauge.  Then panic if it fails.  That'll fail fast in UT/FV if we break the label set-up.

felix/fv/bpf_test.go
```diff
@@ -2080,16 +2104,39 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						Eventually(probeMaglevLocalConntrackMetricFunc(tc.Felixes...), "10s", "1s").Should(Equal([]int{0, 0, 0}), "Expected maglev local-conntrack metric to start at 0 for all Felixes")
+						Eventually(probeMaglevRemoteConntrackMetricFunc(tc.Felixes...), "10s", "1s").Should(Equal([]int{0, 0, 0}), "Expected maglev remote-conntrack metric to start at 0 for all Felixes")
+
 						cc.ExpectSome(externalClient, TargetIP(clusterIP), port)
 						cc.ExpectSome(externalClient, TargetIP(externalIP), port)
 						cc.CheckConnectivity()
+
+						// There is a 10-second interval between iterations of Felix's conntrack scanner (where we export the maglev conntrack metrics).
+						// This means we must be very pessimistic about timeouts when searching for the prom values we're after.
+						Eventually(probeMaglevRemoteConntrackMetricFunc(tc.Felixes[initialIngressFelix]), "12s", "1s").Should(Equal([]int{2}), "Expected maglev-ingress felix to increment the remote-conntracks metric")
```
fasaxc, line 2134: Is 2 guaranteed to be correct?  I guess, even if there was a retry, maglev would choose same backend and get same conntrack?

aaaaaaaalex: I believe two is guaranteed, yep. One for cluster IP, one for external IP.

felix/bpf/conntrack/scanner.go
aaaaaaaalex: Sadly I don't think it's viable to initialise the gauges in the `init()` method along with the others, because we need to discern between IP versions, which I'm getting from the Scanner state.

I can init them when the Scanner is initialised, but will also need to add a mutex to sync access, since many places call `Scan()`, where the gauges are set.

aaaaaaaalex: Alternatively, this `Scan` method gets called very early on in Felix's lifecycle, so I could just change these logs to a Panic?

fasaxc: Could you init them when a particular version of scanner is created ans store in fields?  That's a common pattern

aaaaaaaalex: Yep - but will need locks I think? Because a few different goroutines call Scan

PR #11681, opened 2026-01-20
[CORE-12132] Move windows e2es to azr provisioners

coutinhop:
## Description

Move the windows e2e tests to Azure provisioners (azr-aso with Calico CNI and azr-aks with Azure CNI) and re-enable them:
- add an option to build and run the tests from the local e2e binary to body_standard.sh
- add a dummy `needrestart` script for VMs where it isn't present (so no other changes are needed to the scripts)
- use f1 semaphore machines for the windows e2es
- increase timeouts for windows e2es (especially policy tests)

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/e2e/pkg/tests/policy/policy.go
+++ b/e2e/pkg/tests/policy/policy.go
@@ -41,6 +41,7 @@ var _ = describe.CalicoDescribe(
 	describe.WithTeam(describe.Core),
 	describe.WithFeature("NetworkPolicy"),
 	describe.WithCategory(describe.Policy),
+	describe.WithWindows(),
 	"Calico NetworkPolicy",
 	func() {
 		// Define variables common across all tests.
@@ -93,11 +94,11 @@ var _ = describe.CalicoDescribe(
 		// - Creating a default-deny policy applied to both namespaces using a GlobalNetworkPolicy.
 		// - Isolating both namespaces with ingress and egress policies
 		// - Asserting only same namespace connectivity exists.
-		framework.ConformanceIt("should correctly isolate namespaces [RunsOnWindows]", func() {
+		framework.ConformanceIt("should correctly isolate namespaces", func() {
 			nsA := f.Namespace
 
 			By("Creating a second namespace B")
-			ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
 			defer cancel()
 			nsB, err := f.CreateNamespace(ctx, f.BaseName+"-b", nil)
 			Expect(err).NotTo(HaveOccurred())
@@ -221,7 +222,7 @@ var _ = describe.CalicoDescribe(
 
 		framework.ConformanceIt("should support service account selectors", func() {
 			ns := f.Namespace
-			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
 			defer cancel()
 
 			By(fmt.Sprintf("Applying a default-deny policy to namespace %s", ns.Name))
@@ -286,7 +287,8 @@ var _ = describe.CalicoDescribe(
 
 		framework.ConformanceIt("should support namespace selectors", func() {
 			ns := f.Namespace
-			ctx := context.Background()
+			ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+			defer cancel()
 
 			By(fmt.Sprintf("Applying a default-deny policy to namespace %s", ns.Name))
 			defaultDeny := newDefaultDenyIngressPolicy(ns.Name)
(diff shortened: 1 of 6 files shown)


e2e/pkg/tests/policy/policy.go
```diff
@@ -35,12 +35,25 @@ import (
 	"github.com/projectcalico/calico/e2e/pkg/utils"
 	"github.com/projectcalico/calico/e2e/pkg/utils/client"
 	"github.com/projectcalico/calico/e2e/pkg/utils/conncheck"
+	"github.com/projectcalico/calico/e2e/pkg/utils/windows"
 )
 
+// Increase QPS, Burst and Timeout in framework config
+// in order to prevent context deadline errors.
+func init() {
```
caseydavenport: Feels like this shouldn't be done within a singular test file - can we do this elsewhere?

I'm also rather surprised this is needed - what were the symptoms you were seeing?

e2e/pkg/tests/policy/policy.go
```diff
@@ -221,7 +234,12 @@ var _ = describe.CalicoDescribe(
 
 		framework.ConformanceIt("should support service account selectors", func() {
 			ns := f.Namespace
-			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+
+			timeout := 30 * time.Second
```
caseydavenport: Do you know what operation was taking a long time?

e2e/pkg/utils/conncheck/conncheck.go
```diff
@@ -141,10 +141,14 @@ func (c *connectionTester) deploy() error {
 
 	// Wait for all pods to be running.
 	By("Waiting for all pods in the connection checker to be running")
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	timeout := 1 * time.Minute
+	if windows.ClusterIsWindows() {
+		timeout = 15 * time.Minute
```
caseydavenport, line 144: This is sort-of a crazy long timeout - is this really "expected" behavior on Windows?

e2e/pkg/utils/conncheck/k8s.go
```diff
@@ -50,7 +50,16 @@ func CreateServerPodAndServiceX(f *framework.Framework, namespace *v1.Namespace,
...
 	for _, port := range ports {
 		args := []string{}
-		if !windows.ClusterIsWindows() {
+		env := []v1.EnvVar{}
+
+		if windows.ClusterIsWindows() {
+			env = []v1.EnvVar{
+				{
+					Name:  fmt.Sprintf("SERVE_PORT_%d", port),
+					Value: "foo",
```
caseydavenport: The value "foo" here is a bit odd - what does this do? I think we should add a comment explaining why this is needed for Windows. 

if the value is unused, maybe we can just do `Value: port` or `Value: "value-not-used"` or some or something like that?

e2e/pkg/tests/policy/policy.go
coutinhop: I did see some apiserver throttling, but I admit it may have been from running the test manually several times, I'll remove the QPS and Burst configs and see if they're fine on Semaphore. I'm fairly sure I've seen timeouts though, not sure exactly why, but with Windows nodes it seems like things take longer...

e2e/pkg/tests/policy/policy.go
coutinhop: if I remember correctly, I was having trouble with cleaning up the namespaces and then I preemptively increased (for windows) all the ones in the file

e2e/pkg/utils/conncheck/conncheck.go
coutinhop: yes, there's windows images with a few GBs in size, so it may take long, I tried to make per @lwr20's guide lines of (paraphrasing) 2x as long as the longest observed case so as not to introduce flakiness

e2e/pkg/utils/conncheck/k8s.go
coutinhop: good point! yes, the way porter is coded it looks at the port from the env var name, not a value, but I'm trying my best not to touch it... will update here!

e2e/pkg/tests/policy/policy.go
coutinhop: removing QPS and Burst worked fine, will try removing everything then

e2e/pkg/utils/conncheck/conncheck.go
```diff
@@ -359,7 +363,11 @@ func (c *connectionTester) Execute() {
 	// Context to control overall timeout for all connections. After it times out, we'll forcefully
 	// terminate any remaining connections. This avoids deadlocking the test waiting for results if
 	// something goes wrong.
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	timeout := 10 * time.Second
```
caseydavenport: We can just bump this across the board IMO - 30s seems reasonable.

PR #11682, opened 2026-01-20
[BPF] don't defrag if first frag is seen first

tomastigera:
## Description

    [BPF] don't defrag if first frag is seen first
    
    When receiving IP fragments from the network, they may arrive out of
    order. Therefore, we need to reassemble the packet to process it through
    conntrack. If a random fragment is received before the first fragment,
    we would not have the L4 protocol properties (e.g., ports) available.
    
    However, it is quite common to receive the first fragment first. Even if
    there is some reordering in the subsequent fragments, the L4 information
    is already available. We can record this information, similar to how we
    handle "from the host" fragments (where we assume no reordering), and
    then process the remaining fragments using the tracked information.
    
    Since we cannot assume that the last fragment will always be received
    last, we do not remove the tracking entry upon receiving the last
    fragment. Instead, we rely on the LRU mechanism of the tracking table to
    reclaim the space when necessary.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf: If first fragment on HEP from outside arrives first (in-order), do not reassemble the fragments just track the fragments.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/ip_v4_fragment.h
+++ b/felix/bpf-gpl/ip_v4_fragment.h
@@ -201,7 +201,6 @@ out:
 	return false;
 }
 
-#ifdef BPF_CORE_SUPPORTED
 static int frags4_remove_ct_cb(void *map, struct frags4_fwd_key *key, __unused struct frags4_fwd_value *value)
 {
 	cali_v4_frgfwd_delete_elem(key);
@@ -211,7 +210,6 @@ static int frags4_remove_ct_cb(void *map, struct frags4_fwd_key *key, __unused s
 	}
 	return 0;
 }
-#endif
 
 #define FRAGS4_HANDLE_UNSUPPORTED	0
 #define FRAGS4_HANDLE_FIRST_IN_ORDER 	1
@@ -220,9 +218,6 @@ static int frags4_remove_ct_cb(void *map, struct frags4_fwd_key *key, __unused s
 
 static CALI_BPF_INLINE int frags4_handle(struct cali_tc_ctx *ctx)
 {
-#ifndef BPF_CORE_SUPPORTED
-	return false;
-#else
 	struct frags4_fwd_value *frag_ct_val = frags4_lookup_ct(ctx);
 
 	if (!frag_ct_val) {
@@ -323,7 +318,6 @@ static CALI_BPF_INLINE int frags4_handle(struct cali_tc_ctx *ctx)
 
 out:
 	return FRAGS4_HANDLE_STORE_ONLY;
-#endif /* BPF_CORE_SUPPORTED */
 }
 
 static CALI_BPF_INLINE void frags4_record_ct_flags(struct cali_tc_ctx *ctx, __u32 flags)
@@ -372,7 +366,6 @@ static CALI_BPF_INLINE void frags4_record_ct_flags(struct cali_tc_ctx *ctx, __u3
 
 	bpf_timer_set_callback(&val->timer, frags4_remove_ct_cb);
 	bpf_timer_start(&val->timer, IPFRAG_TIMEOUT * 1000000000ULL, 0);
-#endif
 }
 
 static CALI_BPF_INLINE void frags4_record_ct(struct cali_tc_ctx *ctx)


felix/bpf-gpl/ip_v4_fragment.h
```diff
@@ -49,12 +49,19 @@ struct frags4_fwd_key {
 	__u16 __pad;
 };
 
+#define FRAGS4_FWD_FLAG_FIRST_OOR  (1<<0) /* Signals that the first fragment was out of order */
+
 struct frags4_fwd_value {
 	__u16 sport;
```
fasaxc, line 61: If you're using LRU, you should also put a timestamp on the entries so that, when you read one, you can spot if it's stale and ignore it.  I think fragment IDs can be re-used quite quickly and we don't want to re-assemble with an hour-old fragment.

tomastigera: I was toying with using the timers, I think I might go with that since this is the same issue for traffic originating within a node/pod

felix/bpf-gpl/tc.c
```diff
@@ -62,8 +61,50 @@
...
+ * ports for non-first fragments unless we have a stored entry for the
+ * fragment stream.
+ */
+static CALI_BPF_INLINE int state_fill_from_l4(struct cali_tc_ctx *ctx, bool decap)
+{
+#ifndef IPVER6
+	if ((CALI_F_TO_HOST || CALI_F_FROM_HEP) && ip_is_frag(ip_hdr(ctx))) {
+		/* For fragments, we need to skip the FIB lookup as we may not be able to
+		 * forward because of smaller MTU on the next hop. We need to ensure that
+		 * all or none of the gragments fo through Linux stack for correct
```
fasaxc: Typos

felix/bpf-gpl/ip_v4_fragment.h
fasaxc: Any concern with either:
* Having millions of timers in flight (30s is a long time in number of packets?)
* Racing between timer clean up and overwriting an entry? Timer isn't checking that the value it's deleting is actually the value that it was scheduled to clean up.
* Timer duplicating LRU function: if we're writing entries at a rate that's triggering LRU cleanup before the timer pops, the timer is wasting resources.

felix/bpf-gpl/ip_v4_fragment.h
```diff
@@ -188,18 +201,63 @@ static CALI_BPF_INLINE bool frags4_try_assemble(struct cali_tc_ctx *ctx)
...
+			CALI_DEBUG("IP FRAG: first fragment in order will created CT entry");
+			return FRAGS4_HANDLE_FIRST_IN_ORDER;
+		} else {
+			/* Out of order fragment, store the fragment and create a CT entry
+			 * to remember that we had an out-of-order fragment.
+			 */
+			frags4_record_ct_flags(ctx, FRAGS4_FWD_FLAG_FIRST_OOR);
+		}
+	} else {
+		if (!frag_ct_val->flags /* faster test !FRAGS4_FWD_FLAG_FIRST_OOR */) {
```
fasaxc, line 238: Uneasy about the abbreviated condition, will we remember to fix this up if we add a new flag?

felix/bpf-gpl/tc.c
```diff
@@ -62,8 +61,50 @@
...
 #define HAS_HOST_CONFLICT_PROG CALI_F_TO_HEP
 
+/* Parse L4 header and fill in state ports. Handles fragments, skips filling
+ * ports for non-first fragments unless we have a stored entry for the
+ * fragment stream.
+ */
+static CALI_BPF_INLINE int state_fill_from_l4(struct cali_tc_ctx *ctx, bool decap)
+{
+#ifndef IPVER6
+	if ((CALI_F_TO_HOST || CALI_F_FROM_HEP) && ip_is_frag(ip_hdr(ctx))) {
```
fasaxc: Doesn't `CALI_F_TO_HOST` cover `CALI_F_FROM_HEP`.  I thought it was set for all programs operating in the to-host-ns direction?  Perhaps we should rename to `CALI_F_TOWARD_HOST_NS`

felix/bpf-gpl/tc.c
```diff
@@ -2153,16 +2189,23 @@ int calico_tc_skb_ipv4_frag(struct __sk_buff *skb)
...
+		deny_reason(ctx, CALI_REASON_FRAG_WAIT);
 		goto deny;
+	case FRAGS4_HANDLE_REASSEMBLED:
+		/* force it through stack to trigger any further necessary fragmentation */
+		ctx->state->flags |= CALI_ST_SKIP_REDIR_ONCE;
+		break;
+	case FRAGS4_HANDLE_FIRST_IN_ORDER:
+		/* Carry on as normal packet, do not redirect, it may get
+		 * fragmented again. */
+		// fwd_fib_set(&ctx->fwd, false);
```
fasaxc: Commented out?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -432,6 +433,42 @@ type ManagerWithHEPUpdate interface {
...
+		}
+
+		timeoutStr := strings.TrimSpace(string(data))
+		timeout, err := strconv.Atoi(timeoutStr)
+		if err != nil {
+			logrus.WithError(err).Warn("Failed to parse net.ipv4.ipfrag_time, using default of 30 seconds")
+			return 30
+		}
+
+		// Validate range - must be positive and fit in uint32
```
fasaxc: Range validation should be applied to the explicit config value too.  right now f someone specifies a too-large/small value in felix config, I think that'll pass through and just get wrapped into uint32.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -432,6 +433,42 @@ type ManagerWithHEPUpdate interface {
...
+		if timeout < 0 || timeout > int(^uint32(0)) {
+			logrus.WithField("timeout", timeout).Warn("Invalid net.ipv4.ipfrag_time value (out of range), using default of 30 seconds")
+			return 30
```
fasaxc: I might clamp into range instead of returning default.  If you ask for 8B seconds, clamping to 4B is close enough in my book.

felix/fv/bpf_test.go
```diff
@@ -1827,6 +1831,78 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 						Eventually(func() int { return tcpdump1.MatchCount("udp-frags") }).Should(Equal(24))
 					})
 
+					_ = testOpts.tunnel == "none" && It("should handle fragmented UDP from external client", func() {
```
fasaxc: Above, you removed this pattern in favour of an `if`

felix/fv/bpf_test.go
```diff
@@ -1827,6 +1831,78 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						tcpdump0.SetLogEnabled(true)
+						tcpdump0.AddMatcher("udp-pod-frags", regexp.MustCompile(
+							fmt.Sprintf("%s.* > %s.*", externalClient.IP, w[0][0].IP)))
+						tcpdump0.Start(infra, "-vvv", "src", "host", externalClient.IP, "and", "dst", "host", w[0][0].IP)
+						defer tcpdump1.Stop()
+
+						externalClient.Exec("ip", "route", "add", w[0][0].IP, "via", felixIP(0))
+
+						// Give tcpdump some time to start up!
+						time.Sleep(time.Second)
```
fasaxc: This shouldn't be needed (or if it is, it should be inside the tcpdump library).  I thought the tcpdump library waited for the "I'm listening" log to appear before returning

felix/bpf-gpl/fib_co_re.h
```diff
@@ -87,7 +91,7 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
 	}
 
 #ifndef IPVER6
-        if ((CALI_F_FROM_HOST || CALI_F_FROM_WEP) && (ctx->state->flags & CALI_ST_FIRST_FRAG)) {
+        if (ctx->state->ip_proto != IPPROTO_ICMP_46 && ctx->state->flags & CALI_ST_FIRST_FRAG) {
```
fasaxc, line 94: Why this change?

fasaxc (changes requested):

felix/bpf-gpl/fib_co_re.h
tomastigera: iirc it is because icmp never has fragments and the direction now does not matter

stevegaossou:
@tomastigera just FYI, I'm re-assigning this PR to OSS `v3.33.0` milestone. Nothing new (I think we already agreed to that). 

But just want to capture that here.

PR #11712, opened 2026-01-23
Fix race rendering confd templates

song-jiang:
## Description

Fixes a race condition in `bgp_processor.go` that causes confd to crash with `fatal error: concurrent map read and map write` when multiple BIRD configuration templates are rendered simultaneously.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/confd/pkg/backends/calico/bgp_processor_test.go
+++ b/confd/pkg/backends/calico/bgp_processor_test.go
@@ -2265,7 +2265,7 @@ func TestConfigCache_ConcurrentReadWrite(t *testing.T) {
 	c := newTestClient(cache, nil)
 
 	const numGoroutines = 10
-	const testDuration = 30 * time.Second
+	const testDuration = 10 * time.Second
 
 	// Create a context with 30-second timeout
 	ctx, cancel := context.WithTimeout(context.Background(), testDuration)


confd/pkg/backends/calico/bgp_processor_test.go
```diff
@@ -2214,3 +2217,109 @@ func TestTruncateBGPFilterName(t *testing.T) {
...
+//
+// This test simulates the real-world scenario where bird.cfg.template (IPv4)
+// and bird6.cfg.template (IPv6) are rendered simultaneously by different
+// goroutines in processor.go's monitorPrefix function. Each template calls
+// GetBirdBGPConfig with its respective IP version.
+//
+// This test will FAIL when run with the -race flag if the configCache map
+// is accessed without proper synchronization:
+//
+//	go test -race -run TestConfigCache_ConcurrentReadWrite ./confd/pkg/backends/calico/
```
nelljerram: This comment makes it sound like we don't normally run with `-race`.  But we do now do that.  Update to something like:

> We now run confd UT with `-race`, so this test will fail if the configCache map is accessed without proper synchronization.

confd/pkg/backends/calico/bgp_processor_test.go
```diff
@@ -2214,3 +2217,109 @@ func TestTruncateBGPFilterName(t *testing.T) {
...
+		"/calico/bgp/v1/host/test-node/ip_addr_v6": "fd00::1",
+		"/calico/bgp/v1/global/as_num":             "64512",
+		"/calico/bgp/v1/global/loglevel":           "info",
+	}
+
+	// Enable mesh for more realistic scenario
+	meshConfig := map[string]interface{}{
+		"enabled": true,
+	}
+	meshConfigJSON, _ := json.Marshal(meshConfig)
```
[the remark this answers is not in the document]
nelljerram: Agree with Copilot here.

confd/pkg/backends/calico/bgp_processor_test.go
```diff
@@ -2214,3 +2217,109 @@ func TestTruncateBGPFilterName(t *testing.T) {
...
+	cache["/calico/bgp/v1/global/node_mesh"] = string(meshConfigJSON)
+
+	// Add some peer nodes
+	cache["/calico/bgp/v1/host/peer-node-1/ip_addr_v4"] = "10.0.0.2"
+	cache["/calico/bgp/v1/host/peer-node-2/ip_addr_v6"] = "fd00::2"
```
nelljerram: Can we prepare `meshConfig` first and then include the last 3 lines in the `cache := ...` above.  That would be neater.

confd/pkg/backends/calico/bgp_processor_test.go
```diff
@@ -2214,3 +2217,109 @@ func TestTruncateBGPFilterName(t *testing.T) {
...
+
+			iterationCount := 0
+			for {
+				select {
+				case <-ctx.Done():
+					// Context timeout reached, exit goroutine
+					t.Logf("Goroutine %d (IPv%d) completed %d iterations", id, ipVersion, iterationCount)
+					return
+				default:
+					// This is what the template actually calls via getBGPConfig template function
```
nelljerram, line 2298: Doesn't make sense: "This is what the ..."  Should it be "This is where the ..." ?

confd/pkg/backends/calico/bgp_processor_test.go
```diff
@@ -2214,3 +2217,109 @@ func TestTruncateBGPFilterName(t *testing.T) {
...
+					config, err := c.GetBirdBGPConfig(ipVersion)
+					if err != nil {
+						// Some errors are expected if cache is cleared
+						continue
+					}
+					// Use the config to prevent optimization
```
nelljerram: Don't understand - what do you mean here?

confd/pkg/backends/calico/bgp_processor_test.go
```diff
@@ -2214,3 +2217,109 @@ func TestTruncateBGPFilterName(t *testing.T) {
...
+	cache["/calico/bgp/v1/global/node_mesh"] = string(meshConfigJSON)
+
+	// Add some peer nodes
+	cache["/calico/bgp/v1/host/peer-node-1/ip_addr_v4"] = "10.0.0.2"
+	cache["/calico/bgp/v1/host/peer-node-2/ip_addr_v6"] = "fd00::2"
+
+	c := newTestClient(cache, nil)
+
+	const numGoroutines = 10
+	const testDuration = 30 * time.Second
```
nelljerram: If you temporarily remove the fix, does the new test reliably fail with a duration of less than 30s?  If so it would be nice to reduce to the smallest duration that is needed to reproduce the failure.

confd/pkg/backends/calico/bgp_processor_test.go
song-jiang: The compiler might notice that config is never used after being assigned so it may optimise out calling 'GetBirdBGPConfig`, assign it to a blank var prevent the optimisation. 

Actually the code checks err as return value anyway so we don't need this line at all.

confd/pkg/backends/calico/bgp_processor_test.go
song-jiang: I think the original comment is correct. Templates use `{{getBGPConfig 4 .}}` or `{{getBGPConfig 6 .}}`
and that template function translates to calling `GetBirdBGPConfig(ipVersion)`, so this line is what (the actual thing that) gets called.

confd/pkg/backends/calico/bgp_processor_test.go
song-jiang: Yes. Without the lock, test failed reliably. 
I'm not sure how to get the smallest duration as it may differ on different test environment (CPU etc.) I think 30s is a good number and should not cause much waste in general.

nelljerram: If it fails reliably on semaphore, I think it will also fail reliably on dev laptops, because dev laptops will be more powerful than Semaphore.  Can we try for 10s?

confd/pkg/backends/calico/bgp_processor_test.go
nelljerram: Ah, I understand now.  OK.

confd/pkg/backends/calico/bgp_processor_test.go
nelljerram: Yes.  In general the compiler cannot optimize out function calls, because the function may have side effects.

confd/pkg/backends/calico/bgp_processor_test.go
song-jiang: Yeah, it fails within 10s. I've updated the code.

Issue #11711, opened 2026-01-23
Malformed URL for apiserver


aojea:
The code seems to add square brackets without taking into consideration if there is an IPv4 or an IPv6 address inside. 

https://github.com/projectcalico/calico/blob/c93f299b69c13f625baf8c004f60aa183eaede00/cni-plugin/pkg/install/install.go#L466-L475

As a consequence, if is an ipv4 address, it creates a malformed url (IPv4 address anonymized on the logs)

> 2026-01-20 13:15:41.454 [ERROR][4563] cni-plugin/plugin.go 162: Final result of CNI ADD was an error. error=error creating calico client: host must be a URL or a host:port pair: "https://[X.X.X.X]:443"
2026-01-20 13:15:41.789 [ERROR][4592] cni-plugin/plugin.go 593: Final result of CNI DEL was an error. error=host must be a URL or a host:port pair: "https://[X.X.X.X]:443"
2

... based on some excellent investigation from @jingyuanliang, this seems to only be a problem when golang fixed n issue with the parsing of malformed url https://github.com/golang/go/issues/75678, so new builds with latest golang patch releases for 1.24 and 1.25  (and 1.26) will start to fail

PR #11721, opened 2026-01-26
Use net.JoinHostPort() where possible.

fasaxc:
## Description

There are several places in the codebase where we need to form "host:port" addresses, where the host part may be an IPv6 address. #11711 highlighted that we were doing this incorrectly in some cases and an [upstream Go change](https://github.com/golang/go/issues/75678) made URL parsing require that IPv4 addresses are unadorned.

This PR replaces the various ad-hoc address calculations with the stdlib utility function `net.JoinHostPort()`, which handles IPv6/IPv4 addresses correctly.

## Related issues/PRs

#11711
CORE-12259
https://github.com/golang/go/issues/75678

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix that come components would add square brackets to IPv4s when forming host:port addresses (treating them like IPv6 addresses).
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/bpf_dual_stack_test.go
+++ b/felix/fv/bpf_dual_stack_test.go
@@ -398,7 +398,7 @@ func describeBPFDualStackTests(ctlbEnabled, ipv6Dataplane bool) bool {
 
 				for _, f := range tc.Felixes {
 					felixReady := func() int {
-						return healthStatus("["+f.IPv6+"]", "9099", "readiness")
+						return healthStatus(f.IPv6, "9099", "readiness")
 					}
 					Eventually(felixReady, "10s", "330ms").Should(BeGood())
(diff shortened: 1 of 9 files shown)


felix/fv/bpf_dual_stack_test.go
```diff
@@ -535,7 +535,7 @@ func describeBPFDualStackProxyHealthTests() bool {
 			}, "10s", "330ms").Should(Equal(200))
 
 			Eventually(func() int {
-				return healthStatus("["+tc.Felixes[0].IPv6+"]", strconv.Itoa(int(healthCheckNodePort)), "")
```
fasaxc, line 538: Removed brackets here, added `net.JoinHostPort` to `healthStatus()`

Issue #8159, opened 2023-10-25
ArgoCD installation failing due to `invalid: metadata.annotations`


headyj:
ArgoCD installation using custom plugins with `helm template` command is failing due to invalid `metadata.annotations`:

```
"installations.operator.tigera.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```
This issue is related to [this one](https://github.com/projectcalico/calico/issues/7826), which was closed without being solved for any reason

The proposed solution will not work for ArgoCD, as well as setting syncOption `- ServerSideApply=true` on the ArgoCD Application level, which doesn't seems to work for custom plugins using `helm template` as proposed [here](https://www.arthurkoziel.com/fixing-argocd-crd-too-long-error/)

## Expected Behavior
installation using ArgoCD with custom plugins using `helm templates` should work correctly

## Possible Solution
A quick solution would be to have the ability to add custom annotations on the [CRD](https://github.com/projectcalico/calico/blob/master/charts/tigera-operator/crds/operator.tigera.io_installations_crd.yaml#L5), so we could add the required annotation:

```
metadata:
  annotations:
    argocd.argoproj.io/sync-options: ServerSideApply=true
```
I did some tests on my side by duplicating the CRD in my templates directory and it seems to be working. The only problem I have is obviously ArgoCD complaining with duplicated CRD resource

## Steps to Reproduce (for bugs)
Just try to install tigera-operator following the guide using ArgoCD with a plugin using `helm template

* Calico version: Helm chart v3.26.3
* Orchestrator version: Kubernetes v1.26 (AWS)

headyj:
I think that even just having the possibility to add our own annotations on the CRD would already be a big plus. It's not a big deal and there would be no risk of regression.

caseydavenport:
Truthfully, the CRDs shouldn't be in the helm chart at all and they should be managed outside directly via users. Helm is pretty bad at managing CRDs (e.g., it won't upgrade them for you) so I'd rather we just cut ties between helm and the CRDs altogether, which would resolve this issue I think.

nd-at-globetel:
Same issue, any updates?

caseydavenport:
No updates I'm afraid. I think we have a few paths to a solution though:

- Short term viable: simply add the `    argocd.argoproj.io/sync-options: ServerSideApply=true` to our CRDs by default.
- Correct long-term: Move the CRDs out of our helm chart, manage them separately.

caseydavenport:
[...]

PR #11727, opened 2026-01-26
Remove CRDs from tigera-operator chart

caseydavenport:
## Description

This is to provide better flexibility w.r.t CRD management.

Helm charts cannot update / delete CRDs, and so helm best-practices include using a separate chart for CRD management: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

This PR:

- Removes the CRDs from the tigera-operator Helm chart.
- Introduces a new crd.projectcalico.org.v1 chart that includes the CRDs. 

## Related issues/PRs

fixes https://github.com/projectcalico/calico/issues/8159

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
BREAKING: The `tigera-operator` Helm chart no longer includes custom resource definitions for Calico. A new `crd.projectcalico.org.v1` companion Helm chart has been introduced which includes the CRDs instead. 
```

## Reminder for the reviewer

Other optional labels:


caseydavenport:
We would be trading one pain point for another if we do that (i.e., more pain for non-argo user upgrades) - which may still be the right call, but IMO if our release notes and upgrade documents are correct then this shouldn't be a problem. 

Folks need to read our upgrade instructions - that's why we have them.

caseydavenport:
@ozdanborne Pushed some commits that rename the tigera-operator chart to avoid the upgrade concerns raised above.

EDIT: Reverted! Turns out Argo doesn't prune objects from the CRDs directory in charts (which is totally good and reasonable behavior)

charts/crd.projectcalico.org.v1/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v2
+name: crd.projectcalico.org.v1
```
ozdanborne, line 2: Interesting chart name. I would have chosen `calico-crds` or `tigera-operator-crds` myself. Any reason you went with this instead?

charts/crd.projectcalico.org.v1/templates/calico
ozdanborne, line 1: I think you want to put these in the `templates/` directory of this new chart, not the `crds/` directory

charts/crd.projectcalico.org.v1/Chart.yaml
caseydavenport: I opted for this because I know I have a second chart in the works that installs the v3 API as CRDs, which I am planning to add as `projectcalico.org.v3` - so users will need to decide which version of the CRDs they need for their cluster based on whether they are using the API server or the CRDs directly.

PR #11736, opened 2026-01-28
Fix rendering of NatPortRange in nftables mode

### LINKED ISSUE #11738: Providing a natPortRange when using Nftables leads to a parsing issue when applied
When using setting `defaultFelixConfiguration.natPortRange: fromInt:toInt` for the `FelixConfiguration`, it will fail to apply / converge if one uses the `Nftables` dataplane.

## Expected Behavior
Apply the snat port range using the syntax `from:to`

## Current Behavior
Failing to converge the settings that are applied to the FelixConfiguration but fail to reconcile.

`Failed to run nft transaction error=/dev/stdin:226:147-151: Error: syntax error, unexpected number, expecting colon`

## Possible Solution
@nelljerram already found the issue in code when applying the masq, the `:` is not stripped, but kept, which leads to an nftables syntax issue

1. install calico 3.31.
2. set `defaultFelixConfiguration.natPortRange: 1024:64000` (or any other from:to range)
3. set `installation.calicoNetwork.linuxDataplane: Nftables`
4. Apply
5. Check `calico-node` pod logs

You should no see the exception `Failed to run nft transaction error=/dev/stdin:226:147-151: Error: syntax error, unexpected number, expecting colon` while also the entire cluster fails to network communicate entirely within the cluster.

## Your Environment

* Calico version: 3.31.3
* Calico dataplane: nftables
* Orchestrator version: k8s
* Operating System and version: ubuntu 24.04
* Link to your project (optional): https://goedit.io

nelljerram:
Fixes #11738 

## Release Note

```release-note
Bugfix: fix rendering of NatPortRange option when using nftables.
```

### DIFF
--- a/felix/fv/nat_outgoing_test.go
+++ b/felix/fv/nat_outgoing_test.go
@@ -76,3 +76,47 @@ var _ = infrastructure.DatastoreDescribe("NATOutgoing rule rendering test", []ap
 		}
 	})
 })
+
+var _ = infrastructure.DatastoreDescribe("NATPortRange rendering test", []apiconfig.DatastoreType{apiconfig.EtcdV3, apiconfig.Kubernetes}, func(getInfra infrastructure.InfraFactory) {
+	var (
+		infra  infrastructure.DatastoreInfra
+		tc     infrastructure.TopologyContainers
+		client client.Interface
+	)
+
+	BeforeEach(func() {
+		var err error
+		infra = getInfra()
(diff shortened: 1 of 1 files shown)


fasaxc (approved):
Looks like we missed an FV test for this, otherwise it would have been failing but I suppose we've got a UT now

marvin-tigera:

PR #11742, opened 2026-01-29
Fix CI against OpenStack Yoga, by removing it

nelljerram:
Yoga has been "unmaintained" - which is OpenStack terminology for a state similar to EoL - since October 2024, and is no longer of interest to our OpenStack customers.  The CI against Yoga recently broke when we updated our Semaphore platform from Ubuntu 20.04 to 22.04.  This was briefly addressed by https://github.com/projectcalico/calico/commit/29d69fa9324e85f0270e0dc02358e9d451785ec4, but since then there has been further breakage, which does not look easy to fix - fundamentally because upstream Yoga-level code was never developed and tested against Ubuntu 22.04.

## Release Note

```release-note
We continue testing Calico against OpenStack Caracal, but have stopped testing Calico against OpenStack Yoga.  This is because Yoga has now been "unmaintained" for over a year, and is not compatible with current CI platforms based on Ubuntu 22.04.
```


nelljerram:
No doc change needed, because our master docs already say only Caracal.

PR #11747, opened 2026-01-30
Cherry pick #11743 for v3.31 branch

nelljerram:
On the v3.31 branch, "UT/FV tests on new kernel" CI is failing and the artifacts show the same problem as in #11743, i.e.:
```
docker buildx build --load --platform=linux/amd64 --pull --build-arg UBI_IMAGE=registry.access.redhat.com/ubi9/ubi-minimal:latest --build-arg GIT_VERSION=v3.31.3-44-gd4da4610c064-dirty --build-arg CALICO_BASE=calico/base:ubi9-1764972441 --build-arg BPFTOOL_IMAGE=calico/bpftool:v7.5.0 -t typha:latest-amd64 -f docker-image/Dockerfile docker-image
ERROR: failed to build: Error response from daemon: client version 1.52 is too new. Maximum supported API version is 1.41: driver not connecting
make[2]: *** [Makefile:118: .calico_typha.created-amd64] Error 1
make[2]: Leaving directory '/home/ubuntu/calico/typha'
make[1]: *** [Makefile:302: build-typha] Error 2
make[1]: Leaving directory '/home/ubuntu/calico/felix'
make: Leaving directory '/home/ubuntu/calico/felix'
make: *** [Makefile:384: fv-bpf] Error 2
```

It's not identical to the master branch because v3.31 uses a separate copy of VM coding under `felix/.semaphore`.

### DIFF
--- a/.semaphore/semaphore.yml.d/blocks/20-felix.yml
+++ b/.semaphore/semaphore.yml.d/blocks/20-felix.yml
@@ -210,7 +210,7 @@
         - export IMAGE=ubuntu-2404-noble-amd64-v20250502a
         - export UBUNTU_VERSION=noble
         - export DOCKER_VERSION=5:27.5.1-1~ubuntu.24.04~noble
-        - export BUILDX_VERSION=" "
+        - export BUILDX_VERSION=0.31.1-1~ubuntu.24.04~noble
         - mkdir artifacts
         - ./.semaphore/create-test-vms ${VM_PREFIX}
     jobs:
--- a/.semaphore/semaphore-scheduled-builds.yml
+++ b/.semaphore/semaphore-scheduled-builds.yml
@@ -629,7 +629,7 @@ blocks:
           - export IMAGE=ubuntu-2404-noble-amd64-v20250502a
           - export UBUNTU_VERSION=noble
           - export DOCKER_VERSION=5:27.5.1-1~ubuntu.24.04~noble
-          - export BUILDX_VERSION=" "
+          - export BUILDX_VERSION=0.31.1-1~ubuntu.24.04~noble
           - mkdir artifacts
           - ./.semaphore/create-test-vms ${VM_PREFIX}
       jobs:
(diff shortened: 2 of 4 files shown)


.semaphore/semaphore.yml.d/blocks/20-felix.yml
```diff
@@ -210,6 +210,7 @@
         - export IMAGE=ubuntu-2404-noble-amd64-v20250502a
         - export UBUNTU_VERSION=noble
         - export DOCKER_VERSION=5:27.5.1-1~ubuntu.24.04~noble
+        - export BUILDX_VERSION=" "
```
fasaxc: Seems odd to have a space in the quotes (rather than empty string)

fasaxc: We're not letting docker float here (see line above), should we pin buildx (but to something more recent)

fasaxc (approved):
One query but looks like it should do the trick.

.semaphore/semaphore.yml.d/blocks/20-felix.yml
nelljerram: I used a space because I misread/misunderstood
```
: ${BUILDX_VERSION="=0.29.0-0~ubuntu.22.04~jammy"}
```
in `create-test-vm` as
```
: ${BUILDX_VERSION:="=0.29.0-0~ubuntu.22.04~jammy"}
```
In fact, `:=` defaults if the var is unset or empty, and `=` defaults only if the var is unset.  So it would work to set BUILDX_VERSION to an empty string here.  But IMO that's relying on some very fiddly behaviour.

Agree that we could also pin BUILDX_VERSION for 24.04.  The version today is `0.31.1-1~ubuntu.24.04~noble`, so I guess I'll put that?  Then that also makes the previous point moot.

nelljerram:
Pretty sure the one FV failure is a flake.

PR #11749, opened 2026-01-30
Update BIRD to not program cluster routes when felix is responsible for it

mazdakn:
## Description

Define a new value, named `felix` for `CALICO_NETWORKING_BACKEND` env variable. This env var is set in Calico node daemon set by Operator and also via manifests.

Operator:
https://github.com/tigera/operator/blob/468491588fc7e98f383de1c1d608112ebad517a9/pkg/render/node.go#L1573-L1601
Manifests: https://github.com/projectcalico/calico/blob/d290f2d6c781a356b322c5e0e01f8fa3cdfdaaaa/charts/calico/templates/calico-node.yaml#L80

Currently, the following values are accepted:
- **none**: Calico networking is not used. For example this is used with flannel, and only Calico policy is used.
- **bird**: BIRD service is enabled, and BIRD takes care of all sort of routing, both internal and external. IPIP encapsulation and no-encap routing are supported by BIRD. (vxlan is not supported in this mode)
- **vxlan**: BIRD service is disabled, and Felix programs vxlan encapsulation routes.

The proposal here is add the following value:
- **felix**: Extends `vxlan` value, where Felix programs all (ipip, vxlan, and no-encap) cluster routes. BIRD service is disabled by default, but can be enabled by users.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:


metadata.mk
```diff
@@ -67,7 +67,7 @@ LIBBPF_VERSION=v1.6.2
 BPFTOOL_IMAGE=calico/bpftool:v7.5.0
 
 # The operator branch corresponding to this branch.
-OPERATOR_BRANCH ?= master
+OPERATOR_BRANCH ?= felix-routing
```
nelljerram: I presume this is temporary?

mazdakn: Correct.

mazdakn:
@nelljerram regarding number 2, I've started working on a moving ip pool filter to go functions. We can address your suggestion there. These BIRD templates are getting out of control.

nelljerram:
> @nelljerram regarding number 2, I've started working on a moving ip pool filter to go functions. We can address your suggestion there. These BIRD templates are getting out of control.

Yes, that sounds consistent with the approach that @song-jiang started.  As you say, the templates have become pretty horrible to work with now, so we're trying to move more logic to Go.

PR #11759, opened 2026-02-03
Replace ippool filters in BIRD template with golang funcs

mazdakn:
## Description

Replace ippool related part of BIRD templates with a Go function to improve BIRD templates unnecessary complexity.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/confd/etc/calico/confd/templates/bird6_ipam.cfg.template
+++ b/confd/etc/calico/confd/templates/bird6_ipam.cfg.template
@@ -77,7 +77,7 @@ filter calico_kernel_programming {
   {{- end}}
 
 {{- end}}
-{{- range $line := ippoolsFilterBIRDFunc (gets "/v1/ipam/v6/pool/*") "" true "" 6 }}
+{{- range $line := ippoolsFilterBIRDFunc (gets "/v1/ipam/v6/pool/*") "reject" true "" 6 }}
 {{ $line }}
 {{- end}}
   accept;                                  {{- /* Destination is not in any ipPool, accept  */}}
--- a/confd/pkg/resource/template/template_funcs.go
+++ b/confd/pkg/resource/template/template_funcs.go
@@ -409,7 +409,7 @@ func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
 // This function generates BIRD statements for IPPool resources to be used as BIRD filters based on the following input:
 //   - pairs: IPPool resources packaged into KVPairs.
 //   - filterAction: specified action to filter generated statements. For exporting pools to BGP peers, we need to
-//     first reject disabled ippool, and then accept the rest at the end after all other filters. Allowed values are
+//     first reject disabled ippools, and then accept the rest at the end after all other filters. Allowed values are
 //     "accept", "reject", and "" (to not filter).
 //   - forProgrammingKernel: Whether the generated statements are intended for programming routes to kernel or exporting to
 //     other BGP Peers. As an example, we need to set "krt_tunnel" for programming IPIP and no-encap IPv4 routes.
--- a/confd/tests/compiled_templates/explicit_peering/global-ipv6/bird6_ipam.cfg
+++ b/confd/tests/compiled_templates/explicit_peering/global-ipv6/bird6_ipam.cfg
@@ -37,6 +37,5 @@ function calico_export_to_bgp_peers(bool internal_peer) {
 }
 
 filter calico_kernel_programming {
-  if (net ~ 2002::/64) then { accept; }
   accept;
 }
--- a/confd/tests/compiled_templates/explicit_peering/local-as-global-ipv6/bird6_ipam.cfg
+++ b/confd/tests/compiled_templates/explicit_peering/local-as-global-ipv6/bird6_ipam.cfg
@@ -37,6 +37,5 @@ function calico_export_to_bgp_peers(bool internal_peer) {
 }
 
 filter calico_kernel_programming {
-  if (net ~ 2002::/64) then { accept; }
   accept;
 }
--- a/confd/tests/compiled_templates/explicit_peering/local-as-ipv6/bird6_ipam.cfg
+++ b/confd/tests/compiled_templates/explicit_peering/local-as-ipv6/bird6_ipam.cfg
@@ -37,6 +37,5 @@ function calico_export_to_bgp_peers(bool internal_peer) {
 }
 
 filter calico_kernel_programming {
-  if (net ~ 2002::/64) then { accept; }
   accept;
 }
--- a/confd/tests/compiled_templates/mesh/bgp-export/bird6_ipam.cfg
+++ b/confd/tests/compiled_templates/mesh/bgp-export/bird6_ipam.cfg
@@ -39,8 +39,5 @@ function calico_export_to_bgp_peers(bool internal_peer) {
 }
 
 filter calico_kernel_programming {
-  if (net ~ 2002:101::/64) then { accept; }
-  if (net ~ 2002:102::/64) then { accept; }
-  if (net ~ 2002:103::/64) then { accept; }
   accept;
 }
--- a/confd/tests/compiled_templates/mesh/ipip-off/bird6_ipam.cfg
+++ b/confd/tests/compiled_templates/mesh/ipip-off/bird6_ipam.cfg
@@ -37,6 +37,5 @@ function calico_export_to_bgp_peers(bool internal_peer) {
 }
 
 filter calico_kernel_programming {
-  if (net ~ 2002::/64) then { accept; }
   accept;
 }
--- a/confd/tests/compiled_templates/mesh/password/step1/bird6_ipam.cfg
+++ b/confd/tests/compiled_templates/mesh/password/step1/bird6_ipam.cfg
@@ -37,6 +37,5 @@ function calico_export_to_bgp_peers(bool internal_peer) {
 }
 
 filter calico_kernel_programming {
-  if (net ~ 2002::/64) then { accept; }
   accept;
 }
--- a/confd/tests/compiled_templates/mesh/password/step2/bird6_ipam.cfg
+++ b/confd/tests/compiled_templates/mesh/password/step2/bird6_ipam.cfg
@@ -37,6 +37,5 @@ function calico_export_to_bgp_peers(bool internal_peer) {
 }
 
 filter calico_kernel_programming {
-  if (net ~ 2002::/64) then { accept; }
   accept;
 }
--- a/confd/tests/compiled_templates/mesh/password/step3/bird6_ipam.cfg
+++ b/confd/tests/compiled_templates/mesh/password/step3/bird6_ipam.cfg
@@ -37,6 +37,5 @@ function calico_export_to_bgp_peers(bool internal_peer) {
 }
 
 filter calico_kernel_programming {
-  if (net ~ 2002::/64) then { accept; }
   accept;
 }
--- a/confd/tests/compiled_templates/mesh/restart-time/bird6_ipam.cfg
+++ b/confd/tests/compiled_templates/mesh/restart-time/bird6_ipam.cfg
@@ -37,6 +37,5 @@ function calico_export_to_bgp_peers(bool internal_peer) {
 }
 
 filter calico_kernel_programming {
-  if (net ~ 2002::/64) then { accept; }
   accept;
 }


confd/etc/calico/confd/templates/bird_ipam.cfg.template
```diff
@@ -62,51 +71,18 @@ function calico_export_to_bgp_peers(bool internal_peer) {
...
   # Don't program static routes into kernel.
   {{- range ls $reject_key}}
     {{- $parts := split . "-"}}
     {{- $cidr := join $parts "/"}}
   if ( net ~ {{$cidr}} ) then { reject; }
   {{- end}}
-
 {{- end}}
-{{- if exists $network_key}}{{$network := getv $network_key}}
-{{range ls "/v1/ipam/v4/pool"}}{{$data := json (getv (printf "/v1/ipam/v4/pool/%s" .))}}
```
mazdakn: I'm not sure if this check is needed. We don't have it for IPv6 pool.

confd/etc/calico/confd/templates/bird6_ipam.cfg.template
```diff
@@ -1,10 +1,8 @@
 # Generated by confd
 function reject_disabled_pools ()
 {
-{{range ls "/v1/ipam/v6/pool"}}{{$data := json (getv (printf "/v1/ipam/v6/pool/%s" .))}}
-{{- if $data.disableBGPExport}}
-  if ( net ~ {{$data.cidr}} ) then { reject; }
-{{- end}}
+{{- range $line := ippoolsFilterBIRDFunc (gets "/v1/ipam/v6/pool/*") "reject" false "" 6 }}
+{{ $line }}
 {{- end}}
```
nelljerram, line 6: nit: I would now consider inlining the above two lines into `calico_export_to_bgp_peers`.  It feels like we no longed need `reject_disabled_pools` to be a separate function.

confd/etc/calico/confd/templates/bird6_ipam.cfg.template
```diff
@@ -1,10 +1,8 @@
 # Generated by confd
 function reject_disabled_pools ()
 {
-{{range ls "/v1/ipam/v6/pool"}}{{$data := json (getv (printf "/v1/ipam/v6/pool/%s" .))}}
-{{- if $data.disableBGPExport}}
-  if ( net ~ {{$data.cidr}} ) then { reject; }
-{{- end}}
+{{- range $line := ippoolsFilterBIRDFunc (gets "/v1/ipam/v6/pool/*") "reject" false "" 6 }}
```
nelljerram, line 4: nit: I would consider replacing the `"reject", false, ""` args (which are a little opaque) with a single enum that describes the context of this `ippoolsFilterBIRDFunc` call.  Think that would be more readable.

nelljerram: (And obviously also in the other cases, but with a different enum value.)

confd/pkg/resource/template/template_funcs.go
```diff
@@ -403,6 +406,128 @@ func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
 	return lines, nil
 }
 
+// This function generates BIRD statements for IPPool resources to be used as BIRD filters based on the following input:
+//   - pairs: IPPool resources packaged into KVPairs.
+//   - filterAction: specified action to filter generated statements. For exporting pools to BGP peers, we need to
+//     first reject disabled ippool, and then accept the rest at the end after all other filters. Allowed values are
```
nelljerram: nit: ippool -> ippools

confd/pkg/resource/template/template_funcs.go
```diff
@@ -403,6 +406,128 @@ func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
...
+		return []string{}, fmt.Errorf("version must be either 4 or 6")
+	}
+
+	lines := []string{}
+	for _, kvp := range pairs {
+		var ippool model.IPPool
+		err := json.Unmarshal([]byte(kvp.Value), &ippool)
+		if err != nil {
+			return []string{}, fmt.Errorf("error unmarshalling JSON: %s", err)
+		}
```
nelljerram, line 456: nit: longer term we should eliminate the cache from the loop here, and instead store IPPools natively instead of going via JSON.

confd/pkg/resource/template/template_funcs.go
```diff
@@ -403,6 +406,128 @@ func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
...
+			action = "reject"
+			comment = "BGP export is disabled."
+		case ippool.VXLANMode == encap.Always || ippool.VXLANMode == encap.CrossSubnet:
+			// VXLAN encapsulation is always handled by Felix.
+			if forProgrammingKernel {
+				// Felix always handles programming VXLAN IPPools.
+				action = "reject"
+				comment = "VXLAN routes are handled by Felix."
+			} else {
+				action = "accept"
```
nelljerram, line 472: Checking my understanding: I think this means that we export VXLAN pool routes, and that it makes sense when exporting to non-cluster peers.  Am I right that it wouldn't make sense when exporting to cluster peers, because then we already know that those peers will not program the route?

Might be nice to add some commenting about this, if we're sure of the precise logic.

nelljerram: (But it is already nicer to discuss this in code here than in the template!)

confd/pkg/resource/template/template_funcs.go
```diff
@@ -403,6 +406,128 @@ func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
...
+			}
+		case ippool.IPIPMode == encap.Always || ippool.IPIPMode == encap.CrossSubnet, // IPIP Encapsulation.
+			ippool.IPIPMode == encap.Undefined || ippool.VXLANMode == encap.Undefined: // No-encapsulation.
```
nelljerram, line 475: Is v1 `Undefined` the same as v3 `Never`?  I'd really prefer the code to read as "Never" here, because "Undefined" sounds really worrying!  Can we add
```
const Never Mode = ""
```
to libcalico-go/lib/backend/encap/ipip.go, and then use that?

confd/pkg/resource/template/template_funcs.go
```diff
@@ -403,6 +406,128 @@ func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
...
+			// IPIP encapsulation or No-Encap.
+			if forProgrammingKernel && version == 4 && len(localSubnet) != 0 {
+				// For IPv4 IPIP and no-encap routes, we need to set `krt_tunnel` variable which is needed by
```
nelljerram, line 478: This seems wrong.  I'm pretty sure we only use `krt_tunnel` in the IPIP case.

confd/pkg/resource/template/template_funcs.go
```diff
@@ -403,6 +406,128 @@ func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
...
+			action = "accept"
+		default:
+			return nil, fmt.Errorf("invalid %s ippool", kvp.Key)
+		}
+
+		// Filter statements based on provided filterAction.
+		if len(filterAction) != 0 && filterAction != action {
+			continue
+		}
+		lines = append(lines, emitFilterStatementForIPPools(cidr, extraStatement, action, comment))
```
nelljerram, line 491: nit: IMO it's not helpful to split pieces of the logic into subfunctions `extraStatementForKernelProgrammingIPIPNoEncap` and `emitFilterStatementForIPPools`

confd/pkg/resource/template/template_funcs.go
nelljerram: OK, I'm wrong.  We also set it to "" for the no-encap case.  Would love to rediscover why I did that!

confd/tests/compiled_templates/explicit_peering/global-ipv6/bird6_ipam.cfg
```diff
@@ -34,13 +33,10 @@ function calico_export_to_bgp_peers(bool internal_peer) {
...
-
-  if ( net ~ 2002::/64 ) then {
-    accept;
-  }
+  if (net ~ 2002::/64) then { accept; }
 }
 
 filter calico_kernel_programming {
-
+  if (net ~ 2002::/64) then { accept; }
```
nelljerram: This is a change - expected?

confd/tests/compiled_templates/explicit_peering/local-as-global-ipv6/bird6_ipam.cfg
```diff
@@ -34,13 +33,10 @@ function calico_export_to_bgp_peers(bool internal_peer) {
...
-
-  if ( net ~ 2002::/64 ) then {
-    accept;
-  }
+  if (net ~ 2002::/64) then { accept; }
 }
 
 filter calico_kernel_programming {
-
+  if (net ~ 2002::/64) then { accept; }
```
nelljerram: Ditto

confd/tests/compiled_templates/mesh/ipip-cross-subnet/bird_ipam.cfg
```diff
@@ -34,22 +33,11 @@ function calico_export_to_bgp_peers(bool internal_peer) {
...
-
-  if ( net ~ 192.168.0.0/16 ) then {
-    if defined(bgp_next_hop) && ( bgp_next_hop ~ 10.192.0.0/16 ) then
-      krt_tunnel = "";
-    else
-      krt_tunnel = "tunl0";
-    accept;
-  }
-
+  if (net ~ 192.168.0.0/16) then { if (defined(bgp_next_hop)&&(bgp_next_hop ~ 10.192.0.0/16)) then krt_tunnel=""; else krt_tunnel="tunl0"; accept; }
```
nelljerram, line 41: nit: it would be marginally nicer for this to be as readable in the BIRD config as it used to be.  I would consider emitting it on multiple lines and with similar indentation as it used to have.  In particular it's less clear now that the final `accept;` is outside of the preceding `if ... else`.

nelljerram (commented):
Nice change!  I have one concern, about code appearing in `calico_kernel_programming` for IPv6 when it wasn't previously there.  Everything else is nits, and you're welcome to take some of those if you like, or else they can be left to future work.

nelljerram:
> Nice change! I have one concern, about code appearing in `calico_kernel_programming` for IPv6 when it wasn't previously there. Everything else is nits, and you're welcome to take some of those if you like, or else they can be left to future work.

In case it's not already obvious, please don't rebase in this PR from now on, so that I can identify further changes.

confd/etc/calico/confd/templates/bird6_ipam.cfg.template
mazdakn: Right, I think we can simplify these templates further, but I think it might sense to leave it to another PR and not complicate this PR more.

confd/pkg/resource/template/template_funcs.go
mazdakn: Fixed.

confd/pkg/resource/template/template_funcs.go
mazdakn: make sense. Our current approach is not efficient. I'll leave this out for another PR.

confd/pkg/resource/template/template_funcs.go
mazdakn: Your point is correct, we do want to:
- export these routes to non-cluster peers
- NOT export to cluster peers since felix programs those routes.

However, we have another filter named `reject_tunnel_routes` that reject all `vxlan` and `wireguard` routes. This filter is only called for cluster peers. https://github.com/projectcalico/calico/blob/1b3afe69f0b62eeab99e63d16126d4c56ba9937e/confd/etc/calico/confd/templates/bird_ipam.cfg.template#L11

In this PR, I've only been trying to replicate the ippools logic in the templates and left other filters out for sake of complexity.

confd/pkg/resource/template/template_funcs.go
mazdakn: yes, based on the code it seems that v1 `Undefined` servers the same purpose as v3 `Never`.
I'll add your suggestion, there is no harm in doing that, and it's more precise.

confd/tests/compiled_templates/explicit_peering/global-ipv6/bird6_ipam.cfg
mazdakn: These changes are safe. We have a accept all in the end, so having these accept statements are not wrong, only unnecessary. Anyway, I made a change to not include these.

confd/tests/compiled_templates/explicit_peering/local-as-global-ipv6/bird6_ipam.cfg
mazdakn: Fixed.

confd/pkg/resource/template/template_funcs.go
mazdakn: I think this change is needed in another place. I'm working on it separately.

confd/pkg/resource/template/template_funcs.go
nelljerram: Thanks for explaining that.

confd/pkg/resource/template/template_funcs.go
nelljerram: Thanks.

nelljerram: (Doesn't need to block this PR, if that would delay things.)

PR #11762, opened 2026-02-04
[ebpf] - Send tcp rst when a backend is deleted

sridhartigera:
## Description

This PR has the following changes to reset a tcp connection when backend dies.
1. Separate BPF program to send a rst.
2. Felix changes to load the new program.
3. A new workloadRemoveScanner which gets the deleted workload IPs from bpf_ep_mgr and marks the tcp ct entries with a flag.
4. StaleNATScanner disabled for TCP - StaleNATScanner acts on ct entries when the service IP is deleted or backend is deleted. For TCP, we can let the connection flow through to the backend even if service is deleted and when backend is deleted, its better not to delete the NATForward and NATReverse CT entries, so that the connection can be reset.
5. UTs and FVs

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
ebpf - Reset the tcp connection when the backend is deleted.
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/conntrack_types.h
```diff
@@ -41,6 +41,7 @@ enum cali_ct_type {
 #define CALI_CT_FLAG_SKIP_REDIR_PEER	0x4000 /* marks connections from a client which is excluded from redir */
 #define CALI_CT_FLAG_SET_DSCP	0x8000 /* marks connections that needs to set DSCP */
 #define CALI_CT_FLAG_MAGLEV	0X10000 /* marks Maglev connections. Allows packets of an existing to arrive via a different tunnel after failover. */
+#define CALI_CT_FLAG_SEND_RESET	0x20000 /* marks connections where we have sent a TCP RST on behalf of the workload */
```
tomastigera: isn't it rather an intention?

```suggestion
#define CALI_CT_FLAG_SEND_RESET	0x20000 /* marks connections where we should sent a TCP RST on behalf of the workload */
```

felix/bpf-gpl/tc.c
```diff
@@ -1963,6 +1973,43 @@ int calico_tc_skb_icmp_inner_nat(struct __sk_buff *skb)
...
+                fwd_fib_set(&ctx->fwd, true);
+        }
+
+        if (skb_refresh_validate_ptrs(ctx, TCP_SIZE)) {
+                deny_reason(ctx, CALI_REASON_SHORT);
+                CALI_DEBUG("Too short");
+                return TC_ACT_SHOT;
+        }
+
+        tc_state_fill_from_iphdr(ctx);
```
tomastigera: do we need to fill it here. It is past CT lookup, we should have this info already, don't we?

felix/bpf-gpl/tcp4.h
```diff
@@ -0,0 +1,84 @@
...
+	th->doff = sizeof(struct tcphdr) / 4;
+	th->seq = 0;
+
+	if (th_orig.ack) {
+		th->seq = th_orig.ack_seq;
+	} else {
+		th->ack_seq = bpf_htonl(bpf_ntohl(th_orig.seq) + th_orig.syn + th_orig.fin + 
+				original_len - (th_orig.doff << 2));
+		th->ack = 1;
+	}
```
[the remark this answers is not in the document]
tomastigera: yes, it is likely that you are reusing the original TCP header and it may have some flags set. If you point somewhere else (there were ip options), it might contain complete garbage. Field that you do not change here may also contain garbage (idk if you set all)

felix/bpf-gpl/tcp4.h
```diff
@@ -0,0 +1,84 @@
...
+	th->check = 0;
+
+	__wsum ip_csum = bpf_csum_diff(0, 0, ctx->ip_header, sizeof(struct iphdr), 0);
+	__wsum tcp_csum = bpf_csum_diff(0, 0, (__u32 *)th, len - sizeof(struct iphdr) - skb_iphdr_offset(ctx), 0);
+	if (bpf_l3_csum_replace(ctx->skb,
+			skb_iphdr_offset(ctx) + offsetof(struct iphdr, check), 0, ip_csum, 0)) {
+		CALI_DEBUG("TCP reset v4 reply: set ip csum failed");
+		return -1;
+	}
+
```
tomastigera, line 74:
```suggestion
```

felix/bpf/conntrack/cleanup.go
```diff
@@ -54,6 +56,71 @@ func init() {
 	prometheus.MustRegister(conntrackCounterStaleNAT)
 }
 
+type WorkloadRemoveScanner struct {
```
tomastigera: denote somewhere that this is for TCP only?

```suggestion
type WorkloadRemoveScannerTCP struct {
```

felix/bpf/conntrack/conntrack_test.go
```diff
@@ -98,6 +100,62 @@ var _ = Describe("BPF Conntrack LivenessCalculator", func() {
...
+		for i := 0; i < 6; i++ {
+			octetA := byte(1 + (i * 2))
+			octetB := byte(14 + (i * 2))
+			ipA := net.IPv4(10, 0, 0, octetA)
+			ipB := net.IPv4(10, 0, 0, octetB)
+			ipCh <- ipA.String()
+			ipCh <- ipB.String()
+		}
+		fmt.Println("Scanning after removing 12 IPs")
+		time.Sleep(2 * time.Second) // Wait for the channel to be processed.
```
tomastigera: you could just check that there are 12 entries in the set, this is a bit random.

felix/bpf/conntrack/scanner.go
```diff
@@ -288,6 +300,26 @@ func (s *Scanner) Scan() {
...
+					continue
+				}
+				updatedVal := s.versionHelper.setRSTFlagInValue(ctVal)
+				batchK = append(batchK, ctKey.AsBytes())
+				batchV = append(batchV, updatedVal.AsBytes())
+				if len(batchK) >= rstBatchSize {
+					if debug {
+						log.Debugf("Updating RST flag on %d conntrack entries in batch.", len(batchK))
+					}
+					_, err := s.ctMap.BatchUpdate(batchK, batchV, 0)
```
tomastigera: I think it may not update all from the batch. It may update only first N.

felix/bpf/conntrack/scanner.go
```diff
@@ -356,6 +388,16 @@ func (s *Scanner) Scan() {
 	// Run the bpf cleaner to process the remaining entries in the cleanup map.
 	cleaned += s.runBPFCleaner()
 
+	if len(batchK) > 0 {
```
tomastigera: so you may need to flush it here in a loop and not discard all after the batch update returned above

felix/bpf/conntrack/scanner.go
```diff
@@ -439,6 +481,32 @@ func (s *Scanner) get(k KeyInterface) (ValueInterface, error) {
...
+	} else if v.Type() == TypeNATReverse {
+		newVal = v4.NewValueNATReverse(time.Duration(v.LastSeen()), flags,
+			v.Data().A2B, v.Data().B2A, v.Data().TunIP, v.OrigIP(), v.OrigPort())
+	} else {
+		newVal = v4.NewValueNormal(time.Duration(v.LastSeen()), flags, v.Data().A2B, v.Data().B2A)
+	}
+	return newVal
+}
+
+func createNewV6Value(v ValueInterface, flags uint32) ValueInterface {
```
tomastigera: since you are not using these functions outside of the version helpers, perhaps you should just put them there. They seems to be very specific to your usecase.

felix/bpf/ut/attach_test.go
```diff
@@ -619,10 +621,10 @@ func runAttachTest(t *testing.T, ipv6Enabled bool) {
 		err = oldProgs.Open()
 		Expect(err).NotTo(HaveOccurred())
 		pm := jumpMapDump(oldProgs)
-		programsCount := 15
+		programsCount := 17
 		oldPoliciesCount := 2
 		if ipv6Enabled {
-			programsCount = 28
+			programsCount = 32
```
tomastigera, line 627: damn, I should have tight this to the number of programs in the hooks, sorry for that 😿

felix/bpf/ut/tcp_rst_test.go
```diff
@@ -0,0 +1,122 @@
...
+		Expect(ok).To(BeTrue())
+		Expect(ipv6R.NextHeader).To(Equal(layers.IPProtocolTCP))
+		Expect(ipv6R.SrcIP).To(Equal(ipv6.DstIP))
+		Expect(ipv6R.DstIP).To(Equal(ipv6.SrcIP))
+
+		tcpL := pktR.Layer(layers.LayerTypeTCP)
+		Expect(tcpL).NotTo(BeNil())
+		tcpR, ok := tcpL.(*layers.TCP)
+		Expect(ok).To(BeTrue())
+		Expect(tcpR.RST).To(BeTrue())
```
tomastigera, line 105: any other flags to test?

felix/bpf/ut/tcp_rst_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	tcpSyn := &layers.TCP{
+		SrcPort:    54321,
+		DstPort:    7890,
+		SYN:        false,
+		Seq:        1000,
+		DataOffset: 5,
+	}
+
+	_, ipv6, l4, _, pktBytes, err := testPacketV6(nil, nil, tcpSyn, nil)
```
tomastigera: worth providing a options in one of the tests?

felix/bpf-gpl/tc.c
```diff
@@ -1963,6 +1973,43 @@ int calico_tc_skb_icmp_inner_nat(struct __sk_buff *skb)
...
+                        .res = TC_ACT_UNSPEC,
+                        .reason = CALI_REASON_UNKNOWN,
+                },
+        );
+        struct cali_tc_ctx *ctx = &_ctx;
+	int ret = 0;
+#ifndef IPVER6
+	ret = tcp_v4_rst(ctx);
+#else
+	ret = tcp_v6_rst(ctx);
```
tomastigera, line 1963: indent

felix/bpf-gpl/tc.c
```diff
@@ -450,6 +453,13 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 			ctx->state->flags |= CALI_ST_SKIP_FIB;
 		}
 		CALI_DEBUG("CT Hit");
+		/* Check for TCP RST injection */
+		if (CALI_F_TO_HOST && ctx->state->ct_result.flags & CALI_CT_FLAG_SEND_RESET) {
```
tomastigera, line 453: should we remove this flag and set conntrack to rstseen? If for any reason it would be detected like a spurious RST, and there would be more traffic it would get reset. Idk if I am not overthinking. This would also send the the RST for any follow up packet, guess we may want it in case of RST packet drop.

tomastigera: Also I guess we should only send the packet to the client. Wht if we see a packet from the server - the now removed endpoint? I guess sending RST does not hurt.

Iirc we discussed that the same connection cannot be reused fast enough to cause a reasonable race, right?

felix/bpf/ut/attach_test.go
sridhartigera: I have a task to sort this out. This is a pain.

felix/bpf-gpl/tc.c
sridhartigera: I have a new ctx for this bpf program. Hence we need to fill it here.

felix/bpf/ut/tcp_rst_test.go
sridhartigera: I think testing reset should be enough.

felix/bpf-gpl/tc.c
sridhartigera: Anyways this is to a workload which doesn't exist, so even if there is a spurious RST, its good to reset the connection. I think there cannot be a RST from the server, unless the same connection is reused. Am I missing something?

felix/bpf-gpl/tc.c
tomastigera: It fills the state, which survives. But I guess you have a new ip/tcp headers now.

felix/bpf/conntrack/scanner.go
```diff
@@ -528,6 +584,20 @@ func (h ipv4Helper) dummyKey() KeyInterface {
 	return dummyKey
 }
 
+func (h ipv4Helper) setRSTFlagInValue(v ValueInterface) ValueInterface {
```
tomastigera: If all you need is to set the flag, you may just add SetFlags() to the ValueInterface and you may get away without the helper entirely, right?

felix/bpf/ut/tcp_rst_test.go
tomastigera: I guess other flags should equal zero

felix/fv/bpf_test.go
```diff
@@ -3483,15 +3481,54 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+							tcpd.SetLogEnabled(true)
+
+							ipRegex := "IP"
+							if testOpts.ipv6 {
+								ipRegex = "IP6"
+							}
+							tcpd.AddMatcher("tcp-rst",
+								regexp.MustCompile(fmt.Sprintf(`%s %s\.\d+ > %s\.\d+: Flags \[[^\]]*R[^\]]*\]`, ipRegex, srcIP, w[1][1].IP)))
+							tcpd.Start(infra)
+							defer tcpd.Stop()
```
tomastigera: I think this is not longer necessary and it is handled by infra now.

felix/fv/bpf_test.go
```diff
@@ -3483,15 +3481,53 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
 						w[0][0].Stop()
 						By("removing the old workload from infra")
 						w[0][0].RemoveFromInfra(infra)
 
 						By("Testing connectivity continues")
-						prevCount = pc.PongCount()
-						Eventually(pc.PongCount, "15s").Should(BeNumerically(">", prevCount),
-							"Expected to see pong responses on the connection but didn't receive any")
+						if testOpts.protocol == "tcp" {
+							Eventually(func() int { return tcpd.MatchCount("tcp-rst") }, "25s").ShouldNot(BeZero(),
```
tomastigera, line 3640: is it also possible to test that the connections breaks? 🤔 So that it tests that that the RST is actually accepted?

tomastigera (approved):

PR #11775, opened 2026-02-05
Don't allocate IPs from IP pools with Disabled status

caseydavenport:
## Description

This PR introduces the concept of an IP pool status condition which can
be set by controllers to indicate to the IPAM code that the pool is not
valid for IP allocations.

This provides an asynchronous way to do cross-object consistency checks
without an apiserver in the loop.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
IPPool status now includes a Conditions array to indicate relevant status information. 
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/ippool.go
```diff
@@ -46,6 +47,19 @@ type IPPool struct {
...
 	Spec IPPoolSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
+
+	// +optional
+	Status *IPPoolStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+const (
+	// IPPoolDisabled is a condition type that indicates the IP pool has been operationally disabled
+	// by Calico due to an issue with the pool.
+	IPPoolConditionDisabled = "Disabled"
```
fasaxc: If we do this, let's be a bit more verbose.  We already have the "disabled" flag and `!all()` selectors, so it'd be great if the conditions were really clear.

* IgnoredForAllocations          <- Set on pools that should be ignored
* OverlapsWithAnotherPool    <- Set on the big and small pool so user can see there's a problem.
* ContainsBlocksWithIncorrectSize

Is there a convention for whether statuses can be positive or negative?  "Disabled: false" vs "AllowedForAllocations: true".  Positive is always easier to grok IME.

fasaxc: Could also reflect the "spec.disabled" and deletion state in the conditions?

kube-controllers/pkg/controllers/ippool/pool_controller.go
```diff
@@ -152,17 +155,171 @@ func (c *IPPoolController) Reconcile(p *v3.IPPool) error {
...
+			"index":        i,
+		}).Debug("Checking pool for overlaps")
+
+		_, net, err := cnet.ParseCIDR(pool.Spec.CIDR)
+		if err != nil {
+			logrus.WithError(err).WithField("cidr", pool.Spec.CIDR).Error("Failed to parse CIDR from IPPool")
+			continue
+		}
+
+		for _, activePool := range active {
```
fasaxc: Got a O(n^2) here; you could make this efficient for large numbers of pools by using a trie.  I think we have two or three trie impls in use in the codebase, I'm sure one of them can do the intersection check that you need.

api/pkg/apis/projectcalico/v3/ippool.go
caseydavenport: > Could also reflect the "spec.disabled" and deletion state in the conditions?

Yeah, I did consider this as well - perhaps that would make sense to do. 

I didn't go as verbose for the "Type" since I think typically that is represented via the "Reason" and then again in more detail in the "Message" section of conditions, so I think "Type" should still be "Disabled" or something like that, and then we can have different "Reasons" like the ones you have called out.

nelljerram:
@caseydavenport I thought we'd had this concept of Disabled IP pools for years already.  Did that get lost due to a regression, and that's why we now need to reinstate it?

nelljerram:
E.g. disabled IP pools are used here: https://docs.tigera.io/calico-enterprise/latest/networking/configuring/dual-tor
image

kube-controllers/pkg/controllers/ippool/pool_controller.go
caseydavenport: Yep, I did consider this and it's probably the right thing to do. I figured the number of pools would be small enough that this shouldn't matter, but probably still better to optimize.

caseydavenport:
@nelljerram we have a concept of administrative disablement - i.e., in the spec.

The disablement in this PR is a bit different - it uses a controller to validate the IP pool and report back a status of whether or not this pool should be used in IP allocations. 

We don't want to update the spec because that is user intent - the user may intend for the IP pool to be active, but we mark it as disabled due to a validation failure (it overlaps with other IP pools)

nelljerram:
@caseydavenport Ah great, that makes sense.  So we still have existing admin `disabled` as well.  But I think this will need a carefully worded release note so as not to confuse people the same way it confused me. :smile:

caseydavenport:
@nelljerram @fasaxc Maybe "Disabled" isn't the right description for this condition type given the spec field with the same name... We could call it something else to avoid confusion?

Some other potential options:

- type:Available, status:false
- type:Degraded status:true
- type:Invalid status:true
- type:NoAllocate status:true

nelljerram:
> @nelljerram @fasaxc Maybe "Disabled" isn't the right description for this condition type given the spec field with the same name... We could call it something else to avoid confusion?
> 
> Some other potential options:
> 
> * type:Available, status:false
> * type:Degraded status:true
> * type:Invalid status:true
> * type:NoAllocate status:true

I think all of those options look good, and all better than "Disabled", which sounds very admin-y to my ears.  Don't have a strong preference between them.

fasaxc:
FWIW there's some guidance here, but no naming recommendation for our case (although "Ready" and "Available" are mentioned as good top-level names): https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties.  

It does say that it can be good to define multiple conditions where one is the overall summary of the others.  In our case, if we were build from scratch, I think we have a few aspects to an IP pool that are calculated:

* Is it allowed for new allocations?
* Is it enabled for routing purposes (route advertisements/VXLAN programming etc).  I think the disbled flag impacts this too?
* Is it full?

Perhaps we should make this condition the first of those, say `AvailableForAllocations`, `Allocatable`?  Leaving a bit of wiggle room to add `Routable`/`BGPExportEnabled`/`Full` etc later?

kube-controllers/pkg/controllers/ippool/pool_controller.go
```diff
@@ -152,28 +155,187 @@ func (c *IPPoolController) Reconcile(p *v3.IPPool) error {
...
+	for _, p := range pools {
+		pool := p.(*v3.IPPool)
+
+		// If the pool is administratively disabled, reflect that in its conditions and skip it for the purposes of
+		// determining overlaps, since an administratively disabled pool should not block other pools from being active.
+		if pool.Spec.Disabled {
+			cond := metav1.Condition{
+				Type:    v3.IPPoolConditionDisabled,
+				Status:  metav1.ConditionTrue,
+				Reason:  "AdminDisabled",
```
fasaxc: Let's have enum constants in the API package for our reasons

kube-controllers/pkg/controllers/ippool/pool_controller.go
```diff
@@ -152,28 +155,187 @@ func (c *IPPoolController) Reconcile(p *v3.IPPool) error {
...
+			logrus.WithField("otherPool", pool.Name).Info("Disabling IPPool due to overlap")
+			if _, err := c.cli.ProjectcalicoV3().IPPools().UpdateStatus(ctx, pool, metav1.UpdateOptions{}); err != nil {
+				logrus.WithError(err).WithField("otherPool", pool.Name).Error("Failed to update status of IPPool")
+			}
+		}
+	}
+
+	// Make sure non-overlapping pools are enabled by removing the disabled condition if it exists.
+	for _, pool := range active {
+		if removeCondition(pool, v3.IPPoolConditionDisabled) {
```
fasaxc: After reading the k8s API conventions, sounds like it's better to keep the condition and toggle its value instead.  That signals "controller is up an paying attention and this state is still valid" (vs not being able to tell if the condition is false or controller is down).

kube-controllers/pkg/controllers/ippool/pool_controller.go
```diff
@@ -152,28 +155,187 @@ func (c *IPPoolController) Reconcile(p *v3.IPPool) error {
...
+				if _, err := c.cli.ProjectcalicoV3().IPPools().UpdateStatus(ctx, pool, metav1.UpdateOptions{}); err != nil {
+					logrus.WithError(err).WithField("pool", pool.Name).Error("Failed to update status of IPPool")
+				}
+			}
+			continue
+		}
+
+		c, err := ip.CIDRFromString(pool.Spec.CIDR)
+		if err != nil {
+			logrus.WithError(err).WithField("cidr", pool.Spec.CIDR).Error("Failed to parse CIDR from IPPool")
```
fasaxc: I guess this is almost impossible once we have good CRD validation(?) but you could set to disabled here too.

fasaxc (changes requested):

kube-controllers/pkg/controllers/ippool/pool_controller.go
caseydavenport: Yeah, the CRD validation should already be taking care of this

fasaxc (approved):
LGTM.

PR #11784, opened 2026-02-06
[BPF] Fix propafation of ctx->fwd

tomastigera:
Since ctx->fwd is not in state, it is only valid within a single program. We do set it to true early on, but that does not survive and we thus may not redirect when we could redirect.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf-gpl/fib_co_re.h
+++ b/felix/bpf-gpl/fib_co_re.h
@@ -216,7 +216,14 @@ skip_redir_ifindex:
 #endif
 
 			rc = bpf_fib_lookup(ctx->skb, fib_params(ctx), sizeof(struct bpf_fib_lookup), fib_flags);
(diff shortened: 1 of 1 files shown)


felix/bpf-gpl/tc.c
```diff
@@ -1303,23 +1299,14 @@ static CALI_BPF_INLINE struct fwd post_nat(struct cali_tc_ctx *ctx,
...
 
 deny:
-	{
-		struct fwd fwd = {
-			.res = TC_ACT_SHOT,
-			.reason = ctx->fwd.reason,
-		};
-		return fwd;
-	}
+	ctx->state->fwd.res = TC_ACT_SHOT;
```
sridhartigera: should the reason be set here?

PR #11812, opened 2026-02-10
Setting TLS 1.3 only ciphers causes API server to fail on startup #11706

### LINKED ISSUE #11706: Setting TLS 1.3 only ciphers causes API server to fail on startup
On Calico v3.31.3, when exclusively using TLS 1.3 ciphers via the `TLSCipherSuites` field on the `Installation` resource:

```yaml
apiVersion: operator.tigera.io/v1
kind: Installation
metadata:
  name: default
spec:
  tlsCipherSuites:
    - name: TLS_AES_256_GCM_SHA384
    - name: TLS_CHACHA20_POLY1305_SHA256
    - name: TLS_AES_128_GCM_SHA256
```

It was found that the API server would fail to start up properly, yielding the following error (formatting adjusted):

```
time=2026-01-21T10:31:31Z
level=error
msg=Error running API server: 
  error configuring http2: 
    http2: TLSConfig.CipherSuites is missing an HTTP/2-required AES_128_GCM_SHA256 cipher 
    (need at least one of TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 or TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256)
func=github.com/projectcalico/calico/apiserver/cmd/apiserver/server.RunServer.func3 
file=/go/src/github.com/projectcalico/calico/apiserver/cmd/apiserver/server/run_server.go:108
```

I believe the issue lies here https://github.com/projectcalico/calico/blob/master/apiserver/cmd/apiserver/server/options.go#L132. It seems the the minimum version for TLS is hardcoded at 1.2. This causes the server to perform some validation checks for ciphers needed for HTTP/2 regardless of what ciphers were set.

Ideally, there's a field or environment variable available that lets us specify what TLS version we want to use. If it's possible to provide that option, that would be great. Otherwise, an explanation on why TLS 1.2 has to be set would be good as well.

Thank you!

KameHameHa21110:
## Description

This PR fixes a bug where the Calico API server fails to start when only TLS 1.3 cipher suites are configured via the Installation resource.

**Type of fix:** Bug fix

**Why this should be merged:**
- Resolves HTTP/2 cipher validation error that prevents API server startup with TLS 1.3-only configurations
- The API server hardcoded MinTLSVersion to TLS 1.2, triggering Go's HTTP/2 validation that requires TLS 1.2-specific ciphers
- Modern security policies may require TLS 1.3-only configurations, which were previously blocked

**Components affected:**
- `crypto/pkg/tls`: Added `ParseTLSVersion()` function and updated `NewTLSConfig()`
- `apiserver/cmd/apiserver/server`: Updated to use `TLS_MIN_VERSION` environment variable

**Testing performed:**
- Added unit tests for `ParseTLSVersion()` covering valid versions (1.2, 1.3), invalid versions, and default behavior
- Added API server integration tests verifying environment variable handling
- All existing tests pass (`crypto/pkg/tls` and `apiserver/cmd/apiserver/server`)
- Verified go vet passes on changed packages

**How it works:**
The API server now reads the `TLS_MIN_VERSION` environment variable (values: "1.2" or "1.3"). When set to "1.3" with TLS 1.3-only ciphers, Go's HTTP/2 stack skips TLS 1.2 cipher validation, allowing the server to start successfully. Defaults to TLS 1.2 when not set (backward compatible).

**Documentation:**
Added comprehensive documentation at `apiserver/TLS_CONFIGURATION.md` with diagrams, tables, and usage scenarios.

## Related issues/PRs

fixes #11706

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Fix API server startup failure when configuring TLS 1.3-only cipher suites. The API server now supports the TLS_MIN_VERSION environment variable (values: "1.2" or "1.3") to control the minimum TLS version. Set TLS_MIN_VERSION=1.3 when using TLS 1.3-only cipher suites to avoid HTTP/2 cipher validation errors.
```

### DIFF
--- a/crypto/pkg/tls/tls_test.go
+++ b/crypto/pkg/tls/tls_test.go
@@ -48,10 +48,10 @@ func TestTLSVersionParsing(t *testing.T) {
 	RegisterTestingT(t)
 
 	tests := []struct {
-		name          string
-		version       string
-		expected      uint16
-		expectError   bool
+		name        string
+		version     string
+		expected    uint16
+		expectError bool
 	}{
 		{
 			name:        "empty string defaults to TLS 1.2",


apiserver/TLS_CONFIGURATION.md
```diff
@@ -0,0 +1,159 @@
+# TLS Configuration Guide
```
caseydavenport: I don't think we need this document - let's remove it.

crypto/pkg/tls/tls.go
```diff
@@ -106,7 +119,14 @@ func NewTLSConfig() (*tls.Config, error) {
 		return nil, fmt.Errorf("failed to create TLS Config: %w", err)
 	}
 	return &tls.Config{
-		MinVersion:   tls.VersionTLS12,
+		MinVersion: func() uint16 {
```
caseydavenport: I think it would be cleaner to determine the TLS version above like we do with the cipher suites and assign it to a local variable instead of in-lining a function here.

apiserver/cmd/apiserver/server/server_test.go
```diff
@@ -56,3 +58,54 @@ func TestCATypeFlagParsing(t *testing.T) {
 		}
 	}
 }
+
+func TestTLSVersionEnvironmentVariable(t *testing.T) {
```
caseydavenport: I don't think this test needs to be in the apiserver package? It's not testing any apiserver code, just `calicoctl` code. 

it actually looks like a duplicate of the test code that's in `tls_test.go` further down.

apiserver/TLS_CONFIGURATION.md
KameHameHa21110: Done, removed it.

crypto/pkg/tls/tls.go
KameHameHa21110: I've refactored it to use a local variable like we do with the cipher suites. 
Looks much cleaner now . Thanks for the suggestion!

apiserver/cmd/apiserver/server/server_test.go
KameHameHa21110: You're right, that was redundant....I've removed the test from the apiserver package 
since it's already covered in the crypto package tests.

apiserver/TLS_CONFIGURATION.md
KameHameHa21110: Removed in 6999d9a

crypto/pkg/tls/tls.go
KameHameHa21110: Fixed in 6999d9a

apiserver/cmd/apiserver/server/server_test.go
KameHameHa21110: Removed in 6999d9a

KameHameHa21110:
Fixed struct field alignment issue that was causing the CI check to fail. 
No functional changes, just formatting to pass make check-dirty.

KameHameHa21110:
The CI failures appear to be infrastructure issues (missing Docker image 'calico/go-build:1.25.7-llvm18.1.8-k8s1.34.3') unrelated to the TLS configuration changes. Can these tests be restarted or would you like me to rebase?
Screenshot 2026-02-11 070232


Screenshot 2026-02-11 070216

caseydavenport:
Sorry for the delay - agree these look unrelated and the most relevant tests have passed, so I think we can just merge this.

PR #11817, opened 2026-02-10
Send HostMetadata to BPF KubeProxy

aaaaaaaalex:
## Description

Exposes host metadata (e.g. labels) to BPF KP by registering KP as a manager in the dataplane.
Individual HostMetadataV4V6Updates are pooled into an aggregated update, and sent over a channel to the KP main loop, before finally configuring `kp.proxy` with the info.

## Related issues/PRs

https://github.com/projectcalico/calico/pull/11202 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/bpf/proxy/kube-proxy.go
+++ b/felix/bpf/proxy/kube-proxy.go
@@ -41,8 +41,8 @@ type KubeProxy struct {
 	pendingHostMetadataUpdates map[string]any
 	// hostMetadataUpdates is a size-1 channel - allows for one non-blocking write,
 	// and repeated updates get merged into older unconsumed ones.
-	hostMetadataUpdates chan map[string]any
-	inSyncWithFelix     bool
+	hostMetadataUpdates    chan map[string]any
+	inSyncWithIntDataplane bool
 
 	ipFamily      int
 	hostIPUpdates chan []net.IP
@@ -289,7 +289,7 @@ func (kp *KubeProxy) OnUpdate(msg any) {
 func (kp *KubeProxy) CompleteDeferredWork() error {
 	// If not in-sync with felix, we allow sending an empty update
 	// to signal to the KP loop that it can start looping.
-	if len(kp.pendingHostMetadataUpdates) == 0 && kp.inSyncWithFelix {
+	if len(kp.pendingHostMetadataUpdates) == 0 && kp.inSyncWithIntDataplane {
 		log.Debug("No pending host metadata updates to process")
 		return nil
 	}
@@ -313,7 +313,7 @@ func (kp *KubeProxy) CompleteDeferredWork() error {
 	log.Debug("Queueing new hostmetadata for main loop")
 	kp.hostMetadataUpdates <- updates
 	log.Debug("Successfully queued new hostmetadata")
-	kp.inSyncWithFelix = true
+	kp.inSyncWithIntDataplane = true
 	return nil
 }
 


felix/bpf/proxy/proxy.go
```diff
@@ -391,6 +398,17 @@ func (p *proxy) SetHostIPs(hostIPs []net.IP) {
...
 }
 
+func (p *proxy) SetHostMetadata(updates map[string]*proto.HostMetadataV4V6Update, requestResync bool) {
+	p.runnerLck.Lock()
+	defer p.runnerLck.Unlock()
+
+	p.hostMetadataByHostname = updates
+	if requestResync {
+		// Invoke a sync via the runner, so that we can release any locks in this goroutine.
+		p.syncDP()
```
tomastigera, line 416: this seems reasonable 👍

felix/dataplane/linux/hostmetadata_cache.go
```diff
@@ -0,0 +1,137 @@
...
+	"maps"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/calico/felix/proto"
+)
+
+type HostMetadataCache struct {
```
tomastigera: I think the cache in this setup is quite a bit one-off tool. I think this can be a generic manager-style plugin that would feed evens of interest to KP. Perhaps preprocessed and sanitized. Perhaps batch them until the CompleteDeferredWork and send them as a batch. That itself gives some throttling.

Since the KP can decide on its own when to kick the DP resync, I think it would be a more appropriate place to do any throttling if need be.

felix/bpf/proxy/kube-proxy.go
```diff
@@ -185,16 +202,24 @@ func (kp *KubeProxy) start() error {
 	go func() {
 		defer kp.wg.Done()
 		for {
+			var ok bool
 			select {
-			case hostIPs, ok := <-kp.hostIPUpdates:
+			case hostIPs, ok = <-kp.hostIPUpdates:
```
tomastigera, line 218: Is this now redundant because it is also covered by the hostMetadata updates (idk 100%)

felix/dataplane/linux/hostmetadata_cache.go
aaaaaaaalex: Right now it's batching them until CompleteDeferredWork - but every update _after_ complete deferred work is "immediate". Well, "immediately" requests a throttled flush, at-least.

I found batching with CompleteDeferredWork was not enough in the FVs to kill chattiness of HostMetadata events.

We can throttle anywhere theoretically but Im pretty hesitant to try doing the throttling in KP. I think it'd make that file much more spaghetti - and I don't see a clean way to throttle the KP channel that sends the update struct - since that channel is also being used as a coalescing-buffer.

My expectation is that to maintain the current pattern in KP where the loop pulls in coalesced updates in parallel, we'd need some fancy tool on either the sending or receiving side of the channel _anyway_, so having a throttled manager that does the throttling cleanly, and also does the batching seemed like the cleanest way

felix/bpf/proxy/kube-proxy.go
aaaaaaaalex: I'll need to check on that - these host IP updates were ultimately coming from interface updates IIRC, and I'm not sure if the same word is being used differently between that and HostMetadata updates 🤔

felix/dataplane/linux/hostmetadata_cache.go
tomastigera: > Right now it's batching them until CompleteDeferredWork - but every update after complete deferred work is "immediate". Well, "immediately" requests a throttled flush, at-least.

I don't follow. DeferredWork is called "periodically", specifically not to act on every update update immediately

tomastigera: > We can throttle anywhere theoretically but Im pretty hesitant to try doing the throttling in KP. I think it'd make that file much more spaghetti - and I don't see a clean way to throttle the KP channel that sends the update struct - since that channel is also being used as a coalescing-buffer.

Spaghetti? Whatever you make it ;-) I think the logic should be in KP. It may throttle some updades, not throttle others, imo it should make decisions on its own. May kick off the update together with service/endpoints updates etc.

aaaaaaaalex: > DeferredWork is called "periodically", specifically not to act on every update update immediately

It's flushing from both CompleteDeferredWork, _and_ OnUpdate at the moment - So that once the first CompleteDeferredWork is called, every subsequent update will get flushed via OnUpdate.. Having many update-triggers called right now doesn't matter because they are discarded by the throttle when they are redundant, and doing it this way was convenient for the way it batched initial updates.

It occurs to me now though, that the only reason I was throttling up here at all was to lower the number of restarts done to the syncer. Since we're no longer restarting the syncer for these updates, batching and throttling become irrelevant - the syncer-runner will take care of throttling updates down to the dataplane. So, I think I can just let the data fly down all the way unhindered. I will batch unconsumed updates via the channel in the KP's update handler func, and this PR will have less files changed :rocket:

felix/bpf/proxy/kube-proxy.go
```diff
@@ -248,24 +255,41 @@ func (kp *KubeProxy) OnHostIPsUpdate(IPs []net.IP) {
...
+			log.Debugf("kube-proxy OnUpdate: coalesce host metadata deletion for queued update (deleted update) %q: %+v", update.Hostname, u)
+		} else {
+			log.WithField("remove", update).Debug("No existing update in queue, sending remove event")
+			updates[update.Hostname] = msg
+		}
 	}
+
+	// Send the now-merged updates back down the channel.
+	log.Debug("Queueing new hostmetadata for main loop")
+	kp.hostMetadataUpdates <- updates
```
tomastigera: This is a weird pattern. I'd expect OnUpdate to keep accumulating updates and when  CompleteDeferredWork gets called, it would send it into the channel. I guess, if it cannot send, it can get the current unconsumed update and merge and send again (that must succeed now!

With your current pattern it feels like you may be sending maps of a single element if the consumer is fast. If it is not, you call receive from channel (not cheap, locks, cond vars) for every update

felix/bpf/proxy/kube-proxy.go
```diff
@@ -190,8 +193,11 @@ func (kp *KubeProxy) start() error {
 
 	// Wait for the initial update.
 	hostIPs := <-kp.hostIPUpdates
+
 	// Could be nil, initially.
-	hostMetadata := kp.checkHostMetadataV4V6Updates()
+	hostMetadataUpdates := kp.recvHostMetadataV4V6Updates()
+	hostMetadata := make(map[string]*proto.HostMetadataV4V6Update)
+	mergeHostMetadataV4V6Updates(hostMetadata, hostMetadataUpdates)
```
tomastigera: is there any need to wait for this update on startup? Even a single node update would satisfy it.

tomastigera (commented):
.

felix/bpf/proxy/kube-proxy.go
aaaaaaaalex: `recvHostMetadataV4V6Updates` is non-blocking

felix/bpf/proxy/kube-proxy.go
```diff
@@ -173,26 +195,41 @@ func (kp *KubeProxy) start() error {
...
 
-	// wait for the initial update
+	// Wait for the initial update.
 	hostIPs := <-kp.hostIPUpdates
 
-	err = kp.run(hostIPs)
+	// Could be nil, initially. Doesn't block.
+	hostMetadataUpdates := kp.checkHostMetadataV4V6Updates()
+	hostMetadata := make(map[string]*proto.HostMetadataV4V6Update)
+	mergeHostMetadataV4V6Updates(hostMetadata, hostMetadataUpdates)
```
tomastigera, line 207: feels like a void operation. Merging with empty is whatever you got, can't you just assign it?

felix/bpf/proxy/kube-proxy.go
```diff
@@ -221,6 +258,89 @@ func (kp *KubeProxy) OnHostIPsUpdate(IPs []net.IP) {
...
+		return nil
+	}
+}
+
+// mergeHostMetadataV4V6Updates merges the existing host metadata updates with the latest updates:
+// - A 'remove' in latest deletes the corresponding key in 'existing'.
+// - An 'update' in latest overwrites the corresponding key in 'existing'.
+// - If 'latest' or 'existing' is nil, does nothing.
+func mergeHostMetadataV4V6Updates(existing map[string]*proto.HostMetadataV4V6Update, latest map[string]any) {
+	if latest == nil || existing == nil {
```
fasaxc: `existing == nil` looks like it'd be a bug? I'd remove that check so code panics in that case (better to bail out than limp along doing something incorrect).

felix/bpf/proxy/kube-proxy.go
```diff
@@ -221,6 +258,89 @@ func (kp *KubeProxy) OnHostIPsUpdate(IPs []net.IP) {
...
+	// Send the merged updates back down the channel.
+	log.Debug("Queueing new hostmetadata for main loop")
+	kp.hostMetadataUpdates <- updates
+	log.Debug("Successfully queued new hostmetadata")
+	return nil
+}
+
+// checkHostMetadataV4V6Updates tries to read a pending host metadata update on the update channel.
+// Returns nil immediately, if nothing can be received from the updates channel.
+func (kp *KubeProxy) checkHostMetadataV4V6Updates() map[string]any {
```
fasaxc:
```suggestion
func (kp *KubeProxy) pollHostMetadataV4V6UpdatesNonBlock() map[string]any {
```

felix/bpf/proxy/kube-proxy.go
```diff
@@ -34,6 +35,14 @@ type KubeProxy struct {
 	proxy  ProxyFrontend
 	syncer DPSyncer
 
+	// Non-thread-safe map - only written/read by the Felix dataplane thread.
+	// Keyed by hostname, value is either a HostMetadataV4V6Update or HostMetadataV4V6Remove.
```
fasaxc:
```suggestion
	// pendingHostMetadataUpdates contains HostMetadataV4V6Update and HostMetadataV4V6Removes that we're batching up to send.  Only accessed from the in dataplane goroutine.
```
Go doc comments should start with the field name

felix/bpf/proxy/kube-proxy.go
```diff
@@ -173,26 +195,41 @@ func (kp *KubeProxy) start() error {
 	kp.syncer = syncer
 	kp.lock.Unlock()
 
-	// wait for the initial update
+	// Wait for the initial update.
 	hostIPs := <-kp.hostIPUpdates
 
-	err = kp.run(hostIPs)
+	// Could be nil, initially. Doesn't block.
+	hostMetadataUpdates := kp.checkHostMetadataV4V6Updates()
```
fasaxc: I think you _should_ wait for the first update to avoid flapping the dataplane on felix restart.

felix/bpf/proxy/kube-proxy.go
```diff
@@ -35,10 +35,10 @@ type KubeProxy struct {
 	proxy  ProxyFrontend
 	syncer DPSyncer
 
-	// Non-thread-safe map - only written/read by the Felix dataplane thread.
-	// Keyed by hostname, value is either a HostMetadataV4V6Update or HostMetadataV4V6Remove.
+	// pendingHostMetadataUpdates contains HostMetadataV4V6Update and HostMetadataV4V6Removes
+	// that we're batching up to send. Only accessed from the int-dataplane goroutine.
 	pendingHostMetadataUpdates map[string]any
-	// Map key is a hostname, value a host metadata update/remove.
+	// hostMetadataUpdates is keyed by hostname, value a host metadata update/remove.
```
fasaxc: Nit: it's a channel, it's not keyed on anything.

felix/bpf/proxy/kube-proxy.go
```diff
@@ -221,6 +260,89 @@ func (kp *KubeProxy) OnHostIPsUpdate(IPs []net.IP) {
...
+		return
+	}
+
+	kp.pendingHostMetadataUpdates[hostname] = msg
+}
+
+// CompleteDeferredWork implements the manager interface.
+// Avoids blocking the thread by draining & merging older updates on the channel before sending.
+func (kp *KubeProxy) CompleteDeferredWork() error {
+	if len(kp.pendingHostMetadataUpdates) == 0 {
```
fasaxc: Can't skip the send on the first CompleteDeferredWork since you're using it as the in-sync signal.  Need to send even if your map is empty.

felix/bpf/proxy/kube-proxy.go
aaaaaaaalex: Oops, mixed up between which field I was talking about. Fixed

felix/bpf/proxy/kube-proxy.go
aaaaaaaalex: Fixed - now allows sending one empty msg

felix/bpf/proxy/kube-proxy.go
```diff
@@ -221,6 +261,92 @@ func (kp *KubeProxy) OnHostIPsUpdate(IPs []net.IP) {
...
+	}
+
+	kp.pendingHostMetadataUpdates[hostname] = msg
+}
+
+// CompleteDeferredWork implements the manager interface.
+// Avoids blocking the thread by draining & merging older updates on the channel before sending.
+func (kp *KubeProxy) CompleteDeferredWork() error {
+	// If not in-sync with felix, we allow sending an empty update
+	// to signal to the KP loop that it can start looping.
```
fasaxc:
```suggestion
	// Skip if we have nothing to send unless this is the first time
	// we're in sync (need to send something in that case to signal to
	// the background goroutine).
```

felix/bpf/proxy/kube-proxy.go
```diff
@@ -34,6 +35,15 @@ type KubeProxy struct {
...
 	syncer DPSyncer
 
+	// pendingHostMetadataUpdates contains HostMetadataV4V6Update and HostMetadataV4V6Removes
+	// that we're batching up to send. Only accessed from the int-dataplane goroutine.
+	// Keyed by hostname (node name).
+	pendingHostMetadataUpdates map[string]any
+	// hostMetadataUpdates is a size-1 channel - allows for one non-blocking write,
+	// and repeated updates get merged into older unconsumed ones.
+	hostMetadataUpdates chan map[string]any
+	inSyncWithFelix     bool
```
fasaxc:
```suggestion
	inSyncWithIntDataplane     bool
```

felix/bpf/proxy/kube-proxy.go
fasaxc: Let's have a test for the empty first sync case.  I'm sure Claude can magic one up.

felix/bpf/proxy/kube-proxy.go
```diff
@@ -221,6 +261,92 @@ func (kp *KubeProxy) OnHostIPsUpdate(IPs []net.IP) {
...
+	for k, v := range kp.pendingHostMetadataUpdates {
+		updates[k] = v
+		log.WithField("nodeName", k).Debug("Queueing new host metadata update")
+		// ... And clear the pending updates after processing.
+		delete(kp.pendingHostMetadataUpdates, k)
+	}
+
+	// Send the merged updates back down the channel.
+	log.Debug("Queueing new hostmetadata for main loop")
+	kp.hostMetadataUpdates <- updates
```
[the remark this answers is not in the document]
aaaaaaaalex: This issue is also relevant for pre-existing host IP updates.
Worth fixing both in a followup PR .

felix/bpf/proxy/kube-proxy.go
aaaaaaaalex: One of those structures can contain removes aswell as updates, while the other only contains updates

fasaxc (approved):

PR #11824, opened 2026-02-11
 Fix CNI delete timer to start after acquiring IPAM lock

sudheernv:
## Description

  **Type:** Bug fix

  **Why this should be merged:**

  This PR fixes a critical bug in the CNI plugin's DELETE operation where the 90-second context timeout was incorrectly started *before* acquiring the IPAM lock, rather than after. This inconsistency
  with ADD operations (AssignIP and AutoAssign) caused DELETE operations to timeout when waiting for the lock during high pod churn.

  **Problem:**
  - During concurrent pod deletions, CNI processes queue for the host-wide IPAM lock
  - The DELETE operation started its 90-second timer before acquiring the lock
  - Lock wait time counted against the timeout budget
  - By the time a DELETE acquired the lock, the context was often already expired or near expiration
  - This resulted in "client rate limiter Wait returned an error: context deadline exceeded" errors

  **Solution:**
  In `cni-plugin/pkg/ipamplugin/ipam_plugin.go`. 
  - Lock is now acquired first, then the 90-second timer starts
  - This matches the pattern already used in ADD operations 
  - DELETE operations now get the full 90 seconds for API calls, regardless of lock wait time

  **Components affected:**
  - CNI plugin IPAM delete operations (`cmdDel` function)
  - Specifically the `ReleaseByHandle` flow

  **Impact:**
  - Low risk: Only reorders existing code, no logic changes
  - High benefit: Eliminates cascading DELETE failures during high pod churn
  - No API changes, backward compatible

  ## Related issues/PRs

https://github.com/projectcalico/calico/commit/8822f57fad8f9eebef57c0ad3cbdcac9eb0bf1ef

  ## Todos

  - [ ] Tests - Existing tests should pass; new tests for lock/timer ordering would be beneficial
  - [ ] Documentation - Code comment added explaining the timer ordering rationale
  - [ ] Release note - Included below

  ## Release Note

  ```release-note
 Fix CNI delete timeout to start after IPAM lock acquisition, preventing "context deadline exceeded" failures during high pod churn
```

### DIFF
--- a/cni-plugin/pkg/ipamplugin/ipam_plugin.go
+++ b/cni-plugin/pkg/ipamplugin/ipam_plugin.go
@@ -495,9 +495,6 @@ func cmdDel(args *skel.CmdArgs) error {
 	})
 
 	logger.Info("Releasing address using handleID")
-	ctx := context.Background()
-	ctx, cancel := context.WithTimeout(ctx, 90*time.Second)
-	defer cancel()
 
 	// Acquire a best-effort host-wide lock to prevent multiple copies of the CNI plugin trying to assign/delete
 	// concurrently. ReleaseXXX is concurrency safe already but serialising the CNI plugins means that
@@ -505,6 +502,10 @@ func cmdDel(args *skel.CmdArgs) error {
 	// number of concurrent requests with essentially no downside.
 	unlock := acquireIPAMLockBestEffort(conf.IPAMLockFile)
 	defer unlock()
+
+	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, 90*time.Second)
+	defer cancel()
 
 	if err := calicoClient.IPAM().ReleaseByHandle(ctx, handleID); err != nil {
 		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {


sudheernv:
@fasaxc Since this PR is on top of your below commit, could you please review it once? Thanks

https://github.com/projectcalico/calico/commit/8822f57fad8f9eebef57c0ad3cbdcac9eb0bf1ef

fasaxc:
Seems like a reasonable change, but CNI DELs do get retried so there's not as bigger issue on delete.

Please can you discard the AI PR description and fill in our PR template properly.  The release node should be a one-liner not a multi-paragraph description.

sudheernv:
> Seems like a reasonable change, but CNI DELs do get retried so there's not as bigger issue on delete.

@fasaxc  Thank you for the quick response! I've updated the PR description and release notes to align with community guidelines.

While CNI DELs are retried, retries don't help when the timer-before-lock bug causes cascading failures in high-churn environments. In our production cluster with 60-80 pods created/deleted per second per node, this bug eventually causes kubelet to mark nodes as "Not Ready."

Here's the production impact we're experiencing:

  Shared Lock Contention:
  - Both CREATE and DELETE operations compete for the same IPAM lock (/var/run/calico/ipam.lock)
  - CREATE: Timer starts AFTER lock ✅ → works fine
  - DELETE: Timer starts BEFORE lock ❌ → fails systematically

  Lock Queue Behavior (Mixed Workload):
  - 60 to 80 concurrent CREATEs + DELETEs serializing on same lock
  - CREATEs hold lock for 5s to 10s each
  - DELETEs waiting in queue: timer counting down during wait
  - By the time a DELETE acquires lock after waiting 90+ seconds → timer already expired → INSTANT FAIL

  Why Retries Don't Help:
  - Retries enter the same congested queue behind both CREATEs and DELETEs
  - Same timer-before-lock bug hits them again
  - Failure rate stays consistently high (60-70% in our production)

 In our production environment, failed DEL operations prevent IPs from being released and pods from being marked for garbage collection, leading to approximately ~16,000 dead pods accumulating on a single node over time. This massive pod list causes PLEG messages to containerd to exceed gRPC limits, ultimately forcing kubelet into "Not Ready" state and rendering the node unusable.

fasaxc:
>60-80 pods created/deleted per second per node

That's a sustained rate that's more than Calico IPAM can handle right now.  Our IPAM logic can handle 2-3 allocations/s on each node.  If you have sustained higher pod churn, it will get backed up and fail somehow.  If it's a burst of that many pods, it'll get backed up but it'll clear eventually.

I have a work in progress to move IPAM allocations to a daemon so that we can cache the state (should give a 100x improvement), but it's a back-burner task so can't say when it'll land.

Merging the PR for consistency with the create path (at least now the two paths will have a balanced probability of getting work done), but you may find that your workload still experiences issues like if you have sustained high churn.

fasaxc (approved):

fasaxc:
@sudheernv what's your use case for such high pod churn?

sudheernv:
> @sudheernv what's your use case for such high pod churn?

@fasaxc Thanks for merging the changes. We are running a large Kubernetes (K8s) cluster with over 700 physical nodes(128 cores/1 TB Memory per node). End users submit Airflow and Spark batch jobs, which create high number of “ephemeral" pods on each node.

sudheernv:
@fasaxc Do you have the dates for when this fix will be available in the v3.31 or v3.30 release?

fasaxc:
@sudheernv I think v3.30 is due later this month; v3.31 in mid April.  I think you may be a customer of Tigera(?) if so, please go through customer support for date commitments.

PR #11832, opened 2026-02-12
Monitor for "PASSED" in VM test output

nelljerram:
Since #11767, successful tests have "PASSED" instead of "SUCCESS".


nelljerram:
The "Felix: BPF tests on Ubuntu 25.10 with jitharden=2 (nftables)" CI job timed out.  Pretty sure that is unrelated to this PR, hence force-merged.

PR #11834, opened 2026-02-12
Improve performance of CIDR IP autodetection when there are many addrs

imbstack:
## Description

We recently started updating many of our nodes to use `IP_AUTODETECTION_METHOD=cidr`. We found a slight increase in CPU usage when doing so which is not an incredible change but is enough to cause us problems on some of our more tightly packed nodes. We looked into it and found a similar issue to one we reported in aws/amazon-ssm-agent#645 and have cooked up a patch to resolve this.

As I understand it the issue is that Go's `net.Interface.Addrs()` does a full `RTM_GETADDR` netlink dump of ALL addresses on the system for each call, then filters to one interface. `GetInterfaces`  calls `convertInterface` → `i.Addrs()` for every interface that passes the name filter. On nodes with many addresses (e.g., thousands of IPVS service VIPs on kube-ipvs0), this is `O(N_interfaces * N_total_addresses)` per invocation. The monitor-addresses process repeats this every 60 seconds.

AWS' ssm-agent [fixed this](https://github.com/aws/amazon-ssm-agent/commit/4e026e0ad923c45a10d276badc6e3abf722fa752#diff-1331bd687c96036031bec147d1791cecef85560d6320bbc38f83d7e4fa0e5956R29-R36) by doing one single load of all addresses and then using it inside the loop. I believe that approach will work here as well.

I've added some unit tests and run this in a testing environment for a while (it appears to work correctly afaict). In terms of addressing the resource consumption, I've attached a chart of CPU usage for 3 calico-node pods. The first two are base v3.30.4 calico-nodes. The first one is with `IP_AUTODETECTION_METHOD=""`, the second and third ones have it set to `cidr`. The third one is a version of v3.30.4 with this patch backported to it. As you can see the total change is _maybe_ 1/4 a CPU but we find that this can be rather important on some of our more highly-loaded nodes and this behavior seems to get worse when calico and kube-proxy are both trying to talk to netlink a lot.

image

This does create a new/different race condition where if interfaces change addresses between the initial load and us acting on it, we may have the incorrect link index but I think this race already existed in the old way as well, just in a different form? Happy to be told this is actually a problem though, we will go back to the drawing board and come up with some other fix hopefully.

## Related issues/PRs

None that I know of. Happy to go make one if we'd like to discuss solutions instead of just accepting this.

## Todos

- [x] Tests
- [x] Documentation (unnecessary, internal only)
- [x] Release note

## Release Note

```release-note
Improve resource consumption when IP_AUTODETECTION_METHOD is set to cidr on Linux instances with many addresses.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/node/pkg/lifecycle/startup/autodetection/interfaces.go
+++ b/node/pkg/lifecycle/startup/autodetection/interfaces.go
@@ -39,6 +39,13 @@ func GetInterfaces(getSystemInterfaces func() ([]net.Interface, error), includeR
 		return nil, err
 	}
 
+	// Single bulk address retrieval for all interfaces (Linux optimization)
+	addrsByIndex, err := getAllInterfaceAddrs()
+	if err != nil {
+		log.WithError(err).Debug("Failed to bulk-fetch addresses, falling back to per-interface queries")
+		addrsByIndex = nil // Fall back to i.Addrs() in convertInterface
+	}
+
 	var filteredIfaces []Interface
 	var includeRegexp *regexp.Regexp
 	var excludeRegexp *regexp.Regexp
@@ -65,7 +72,14 @@ func GetInterfaces(getSystemInterfaces func() ([]net.Interface, error), includeR
 		include := (includeRegexp == nil) || includeRegexp.MatchString(iface.Name)
 		exclude := (excludeRegexp != nil) && excludeRegexp.MatchString(iface.Name)
 		if include && !exclude {
-			if i, err := convertInterface(&iface, versions); err == nil {
+			// Look up pre-fetched addresses (or nil for fallback)
+			var addrs []net.Addr
+			if addrsByIndex != nil {
+				addrs = addrsByIndex[iface.Index]
+			}
+
+			// Convert with pre-fetched or fallback addresses
+			if i, err := convertInterface(&iface, addrs, versions); err == nil {
 				filteredIfaces = append(filteredIfaces, *i)
 			}
 		}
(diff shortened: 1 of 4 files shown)


caseydavenport:
Sorry for the delay! This looks promising - taking a look now.

node/pkg/lifecycle/startup/autodetection/interfaces.go
```diff
@@ -75,12 +89,19 @@ func GetInterfaces(getSystemInterfaces func() ([]net.Interface, error), includeR
...
+func convertInterface(i *net.Interface, addrs []net.Addr, versions []int) (*Interface, error) {
 	log.WithField("Interface", i.Name).Debug("Querying interface addresses")
-	addrs, err := i.Addrs()
-	if err != nil {
-		log.Warnf("Cannot get interface address(es): %v", err)
-		return nil, err
+
+	var err error
+	if addrs == nil {
+		// Fallback to per-interface query
```
caseydavenport, line 99: Do you foresee any scenarios where this fallback is going to be necessary? 

It shouldn't be a problem, just wondering if there was something you had in mind or if this is more defensive. I suppose there could be race conditions between querying the addresses and the interfaces.

caseydavenport (approved):
A minor question but this LGTM otherwise, thank you for the nice write up and fix!

node/pkg/lifecycle/startup/autodetection/interfaces.go
imbstack: Good question, I don't know of any way that this would fail other than maybe that race you mention but unless there are interfaces being added/removed a lot I don't expect to hit that? I imagine people use Calico in all sorts of funny ways though so it felt good to be careful here.

But yeah in general it was just choice to be extra careful here.

PR #11844, opened 2026-02-13
Cache most recently created uniquelabels.Map and re-use

fasaxc:
## Description

In the calculation graph profiles (aka "parent labels") reach several label indexes and each one calls `uniquelabels.Make` independently.  Since the calculation graph runs on a single thread, when processing a profile we call `uniquelabels.Make` 3 times in a row in this case and it returns a new but identical `uniquelabels.Map` each time.  Although the keys and values get deduped, the map itself has quite high overhead; it'd be better to only have one copy of the map...

Since profile is a v3 object, we can't just change the type of the field to a `uniquelabels.Map` (so that the map one be calculated once when parsing the JSON) and it feels a bit awkward to convert the v3 resource into a felix-specific resource at the top of the calc graph in order to make that work.

Instead, this PR adds a simple one-element cache in a wrapper around `uniquelabels.Make` when used in these indexes; this should give a 100% hit rate for this use case.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Felix now avoids storing namespace labels multiple times in different caches, reduces memory usage.
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/calc/active_bpgpeer_calculator.go
+++ b/felix/calc/active_bpgpeer_calculator.go
@@ -71,7 +71,7 @@ func NewActiveBGPPeerCalculator(hostname string) *ActiveBGPPeerCalculator {
 		allBGPPeersByName: map[string]*v3.BGPPeer{},
 		peersByWorkloadID: map[model.WorkloadEndpointKey][]string{},
 	}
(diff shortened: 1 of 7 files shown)


fasaxc:
Closing in favour of #11854

PR #11845, opened 2026-02-13
Introduce programClusterRoutes to BGPConfig to stop BIRD from programming cluster routes

mazdakn:
## Description

Introduce `programClusterRoutes` to `BGPConfiguration` to allow users configure BIRD to NOT program cluster routes. Supported values are, `Enabled` (default), and `Disabled`.

This option is needed to hand over responsibility of programming cluster route to Felix. With `programClusterRoutes` in `BGPConfiguration` set to `Disabled`, and `programClusterRoutes` in `FelixConfig` to `Enabled`, Felix will program all cluster routes instead of BIRD.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Introduce ProgramClusterRoutes option in BGPConfiguration to control how a cluster node gets a route to a
workload on another node, when that workload's IP comes from an IP Pool with vxlanMode: Never.
When ProgramClusterRoutes is Enabled, confd and BIRD program that route. When ProgramClusterRoutes is Disabled, it is expected that Felix will program that route.
Felix always programs such routes for IP Pools with vxlanMode: Always or vxlanMode: CrossSubnet. [Default: Enabled]
```

## Reminder for the reviewer

Other optional labels:


api/pkg/apis/projectcalico/v3/bgpconfig.go
```diff
@@ -65,6 +65,17 @@ const (
 	ServiceLoadBalancerAggregationDisabled ServiceLoadBalancerAggregation = "Disabled"
 )
 
+// BGPWithinCluster defines whether BGP within the cluster is enabled or disabled.
+// +kubebuilder:validation:Enum=Enabled;Disabled
+type BGPWithinCluster string
+
+const (
+	// BGPWithinClusterEnabled means BGP within the cluster is enabled (default behavior)
```
caseydavenport: I think we can give a bit more context on these comments - they are ultimately user facing. e.g., "When Enabled, Calico uses BGP to distribute route information between nodes. When disabled, Calico learns the necessary routing information from its IPAM database and running workloads on the cluster" 

Or something like that.

mazdakn: @caseydavenport what do you think about the field itself?

caseydavenport: Hm, yeah I do wonder if it's a little bit unintuitive. For example, we have `nodeToNodeMeshEnabled` already as an option, and this seems like it might overlap (conceptually) with that field a bit.

mazdakn: Exactly. In theory, felix can program cluster routes (based on ipam information), and mesh enabled to distribute external routes. What do you think about naming the field as `bgpForClusterRouting`?

caseydavenport: I suppose this config option really is less about whether BGP is enabled and more about whether routes learned from other cluster nodes are programmed into the routing table?

mazdakn: correct.

caseydavenport: Maybe something like one of these?

- clusterRouteSource: BGP / CalicoIPAM
- internalRoutingMode: BGP / Felix (CalicoIPAM?)
- clusterRoutingMode

?

frozenprocess: I like the `clusterRoutingMode` or something like `backendForClusterRouting`.

As a User I want it to be pretty clear what I'm fiddling with.

mazdakn: I think I personally like `ClusterRoutingMode`. 
What would be the supported values? `Felix` / `CalicoIPAM` and `BGP`?

api/pkg/apis/projectcalico/v3/bgpconfig.go
```diff
@@ -146,6 +146,13 @@ type BGPConfigurationSpec struct {
 	// It is recommended to use a link-local address.
 	// +optional
 	LocalWorkloadPeeringIPV6 string `json:"localWorkloadPeeringIPV6,omitempty" validate:"omitempty,ipv6"`
+
+	// ProgramClusterRoutes specifies whether BIRD or Felix should program cluster routes. When Enabled, Calico uses BGP
+	// to distribute route information between nodes. When disabled, Calico learns the necessary routing information
```
nelljerram: Does this single BGPConfig setting really do everything on its own?  I understand that it has an effect on confd - i.e. that it modifies the BIRD config so as not to propagate remote workload routes.  But does it also cause Felix to do the remote workload route programming?  I thought that the Felix part would require a parallel FelixConfig setting.

api/pkg/apis/projectcalico/v3/bgpconfig.go
```diff
@@ -146,6 +146,13 @@ type BGPConfigurationSpec struct {
...
+	// from its IPAM database and running workloads on the cluster. Felix always programs VXLAN routes. [Default: Enabled]
```
nelljerram: The "VXLAN routes" on its own is a touch confusing.  Can we be more precise here?  I think what we need is something like:

> ProgramClusterRoutes controls how a cluster node gets a route to a workload on another node, when that workload's IP comes from an IP Pool with `vxlanMode: Never`.  When ProgramClusterRoutes is Enabled, confd and BIRD program that route.  When ProgramClusterRoutes is Disabled, it is expected that Felix will program that route.  Felix always programs such routes for IP Pools with `vxlanMode: Always` or `vxlanMode: CrossSubnet`.

(Am I right about `vxlanMode: CrossSubnet`, or does ProgramClusterRoutes also control what happens for the non-cross-subnet routes in that case?)

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -823,36 +832,45 @@ func (c *client) processIPPools(config *types.BirdBGPConfig, ipVersion int) erro
...
+				// our fork of BIRD.
+				extraStatement = extraStatementForKernelProgrammingIPIPNoEncap(ippool.IPIPMode, localSubnet)
+			}
+			action = "accept"
+		} else {
+			felixHandlesRoutes(forProgrammingKernel, "Cluster")
 		}
-		action = "accept"
 	default:
 		log.WithFields(log.Fields{
```
nelljerram: For me this is a non-intuitive and tricky to understand way of coding the change when programClusterRoutes is false.  Can I suggest instead adding this after the `switch` block:
```
    if forProgrammingKernel && (action != "reject") && !(programClusterRoutes && (ippool.VXLANMode == encap.Never)) {
        action = "reject"
        comment = fmt.Sprintf("%s routes are handled by Felix.", routeType)
    }
```

Then I would also suggest:
- Default `action := "accept"` before the switch block.
- VXLAN case in the switch block is no longer needed at all.
- The case with `extraStatement` can stay as it was, but no longer needs to set `action`.
- `default` and `DisableBGPExport` cases stay as they were.

WDYT?

nelljerram: (I think that gives the same effects, but is easier to understand, at least for me.)

api/pkg/apis/projectcalico/v3/bgpconfig.go
mazdakn: We already have that option in FelixConfig: 
https://github.com/projectcalico/calico/blob/b9ddc3f6c2260547cbb01cf4ff090a98233c63de/api/pkg/apis/projectcalico/v3/felixconfig.go#L574

mazdakn: When enabled, it makes felix start dataplane managers to program cluster routes. This was added in v3.31.

api/pkg/apis/projectcalico/v3/bgpconfig.go
mazdakn: Thanks for through suggestion. I'll use it.

And also your statement about `vxlanMode: CrossSubnet` is correct. Felix is always responsible for it.

confd/pkg/backends/calico/bgp_processor.go
mazdakn: I think you are right that this function is hard to read now. I think it needs a better approach. I made some changes. WDYT about it?

api/pkg/apis/projectcalico/v3/bgpconfig.go
nelljerram: Great, that's what I thought.  So my concern is just with the wording above being a little misleading.  I would suggest changing

> ProgramClusterRoutes specifies whether BIRD or Felix should program cluster routes.

to

> ProgramClusterRoutes specifies whether Calico uses BGP to program cluster routes.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -823,49 +832,53 @@ func (c *client) processIPPools(config *types.BirdBGPConfig, ipVersion int) erro
...
+		}
+		return emitFilterStatementForIPPools(cidr, "", "accept", filterAction, "")
 	}
-	return emitFilterStatementForIPPools(cidr, extraStatement, action, comment)
+
+	log.WithFields(log.Fields{
+		"ippool":    ippool.CIDR,
+		"ipVersion": ipVersion,
+	}).Error("Invalid ippool")
+	return ""
```
nelljerram, line 881: Thanks, that refactoring also looks nice.

api/pkg/apis/projectcalico/v3/bgpconfig.go
mazdakn: I'm going to add your suggest comment above to the same option in FelixConfig since it's more explanatory.

PR #11846, opened 2026-02-13
Don't store label restrictions for every selector.

fasaxc:
Instead, cache only the most recently used label restrictions in a package-level variable.  The label restrictions are used several times when adding a selector to the index (which happens on a single thread) so caching one instance avoids almost all recalculation.

Storing a map per selector really adds up if there are tens of thousands of selectors active. In addition, some uses of selectors don't need the restrictions, so we save calculation there.

Before: 
```
goos: linux
goarch: amd64
BenchmarkIsolatedCustomers1k-12      	      25	  45382677 ns/op	        10.13 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	      1403 Msgs	       100.0 Policies	         0.04368 s	16544401 B/op	  191707 allocs/op
BenchmarkIsolatedCustomers10k-12     	       3	 388005473 ns/op	        47.56 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	     10403 Msgs	       100.0 Policies	         0.3810 s	127345704 B/op	 1471678 allocs/op
BenchmarkIsolatedCustomers100k-12    	       1	6898444469 ns/op	       401.7 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	    100403 Msgs	       100.0 Policies	         6.898 s	1213249056 B/op	14929873 allocs/op
PASS
ok  	github.com/projectcalico/calico/felix/calc	19.689s
```

After:
```
goos: linux
goarch: amd64
BenchmarkIsolatedCustomers1k-12      	      25	  42360355 ns/op	         9.677 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	      1403 Msgs	       100.0 Policies	         0.04082 s	16041110 B/op	  188152 allocs/op
BenchmarkIsolatedCustomers10k-12     	       3	 364593380 ns/op	        43.47 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	     10403 Msgs	       100.0 Policies	         0.3619 s	123024050 B/op	 1441087 allocs/op
BenchmarkIsolatedCustomers100k-12    	       1	6048047051 ns/op	       361.2 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	    100403 Msgs	       100.0 Policies	         6.048 s	1170766264 B/op	14629471 allocs/op
PASS
ok  	github.com/projectcalico/calico/felix/calc	18.434s
```

**Release note**
```release-note
Felix now defers calculation of selector label restrictions until they are needed, saving RAM when many selectors are active.
```

### DIFF
--- a/libcalico-go/lib/selector/parser/label_restriction_test.go
+++ b/libcalico-go/lib/selector/parser/label_restriction_test.go
@@ -150,6 +150,42 @@ func TestLabelRestrictions(t *testing.T) {
 	}
 }
 
+func TestLabelRestrictionsCache(t *testing.T) {
+	RegisterTestingT(t)
+
+	selA, err := Parse("has(a)")
+	Expect(err).NotTo(HaveOccurred())
+	selB, err := Parse("has(b)")
+	Expect(err).NotTo(HaveOccurred())
+
+	// Clear cache.
+	lastRestrictionSelector = nil
+	lastLabelRestrictions = nil
+
+	// First call should populate the cache.
+	lrsA1 := selA.LabelRestrictions()
+	Expect(lastRestrictionSelector).To(BeIdenticalTo(selA))
+
+	// Second call to the same selector should be a cache hit;
+	// the cached selector pointer should not change.
+	lrsA2 := selA.LabelRestrictions()
+	Expect(lastRestrictionSelector).To(BeIdenticalTo(selA),
+		"cache hit should not change the cached selector")
+	Expect(lrsA2).To(Equal(lrsA1))
+
+	// Calling on a different selector should update the cache.
+	lrsB := selB.LabelRestrictions()
+	Expect(lastRestrictionSelector).To(BeIdenticalTo(selB),
+		"cache miss should update the cached selector")
+	Expect(lrsB).NotTo(Equal(lrsA1))
+
+	// Going back to selA should update the cache again.
+	lrsA3 := selA.LabelRestrictions()
+	Expect(lastRestrictionSelector).To(BeIdenticalTo(selA),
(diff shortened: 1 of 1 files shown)


libcalico-go/lib/selector/parser/ast.go
```diff
@@ -145,8 +145,24 @@ func (sel *Selector) UniqueID() string {
 	return sel.hash
 }
 
+var (
+	lastRestrictionMutex    sync.Mutex
+	lastRestrictionSelector *Selector
+	lastLabelRestrictions   LabelRestrictions
+)
```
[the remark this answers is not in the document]
fasaxc: Not worth it; we're only leaking one selector.  The cleanup logic/tracking would likely cost more than the leak.

felix/labelindex/labelrestrictionindex/label_restriction_index.go
```diff
@@ -92,7 +92,7 @@ func (s *LabelRestrictionIndex[SelID]) AddSelector(id SelID, selector *selector.
 	optimized := false
 	debug := logrus.IsLevelEnabled(logrus.DebugLevel)
 	if found {
-		res := lrs[labelName]
+		res, _ := lrs.Get(labelName)
```
nelljerram, line 95: Presumably the zero value works fine here, same as it did before?

fasaxc: Yes, it'll return a zero struct, which has the same behaviour as it always did.  I suppose I could ditch the second parameter, but it seemed natural to mimic the typical Map interface

PR #11847, opened 2026-02-13
Add Zstd compression algorithm support for Typha syncproto

hjiawei:
## Description

New feature (performance): adds zstd as a second compression algorithm for the Typha sync protocol, alongside snappy.

**Why:** zstd compresses Typha's sync data to roughly half the size that snappy manages, even at its fastest setting. On a 10k-pod snapshot the ratio is 9.69:1 vs snappy's 4.65:1 (1.17 MB → 561 KB). That cuts Typha's egress bandwidth and speeds up client catch-up during rolling restarts, when Typha streams snapshots to many Felix instances at once. Decompression runs at ~374 MB/s, so the extra client cost is negligible. Full analysis and benchmark results: `typha/docs/compression-analysis.md`.

**What changed:**

- The client advertises its supported algorithms (`snappy`, `zstd`) in its hello message. The server picks the first algorithm from its own preference order that the client supports; the default order is `zstd,snappy`. So zstd wins when both sides support it, and older clients or servers fall back to snappy automatically. There is deliberately no user-facing config knob — negotiation handles version skew.
- Pre-computed binary snapshots are cached per (syncer type × compression algorithm); caches are only created for algorithms in the server's configured order. The snapshot metrics (`typha_snapshots_generated`, `typha_snapshots_reused`, `typha_snapshot_raw_bytes`, `typha_snapshot_compressed_bytes`) gain a `compression` label, as do `typha_connections_streaming` and `typha_client_snapshot_send_secs` (uncompressed connections report `compression="none"`), so the negotiated algorithm is observable per connection during upgrades.
- Per-connection zstd writers are bounded to one encoder goroutine and a 1 MiB window (the library defaults are GOMAXPROCS goroutines and an 8 MiB window per writer, which adds up with many connections). The snapshot-cache writer keeps the defaults, where compression ratio and speed matter most.
- On the client, one zstd stream decoder is reused across decoder restarts; it decodes the cached snapshot frames and the per-connection delta frames as one continuous stream. The `MsgDecoderRestart`/ACK protocol guarantees the server sends no bytes in a new encoding until the client has switched decoders.
- `typha/DESIGN.md` documents the negotiation, the ACK-before-new-encoding invariant, and the snapshot cache.
- `github.com/klauspost/compress` is now a direct dependency, bumped to v1.19.0.

**Testing:**

- In-process FV tests (`typha/fv-tests`) assert the negotiated algorithm for the version-skew matrix: defaults on both sides → zstd; a snappy-only (old) client → snappy; a snappy-only (old) server → snappy; no overlap or no decoder-restart support → uncompressed. Data-flow tests (snapshots, deltas, deletes, multiple clients) run against both algorithms. All suites pass with the race detector.
- Felix FV `"with Typha and Felix-Typha TLS"` (real Felix + real Typha + TLS) run for three real-binary combinations: branch↔branch (negotiated zstd), branch Felix ↔ released Typha v3.31.0 image (fell back to snappy), and pre-zstd master Felix ↔ branch Typha (fell back to snappy). All passed.
- Compression benchmarks and round-trip correctness tests in `typha/pkg/syncserver/compression_bench_test.go`.

**Components affected:** typha (server and client library); Felix and other Typha clients pick up the client-side change through the shared `syncclient` package.

## Todos

- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Typha now supports zstd compression for the sync protocol and prefers it when both client and server support it, roughly halving sync data transfer compared to snappy. Older clients and servers automatically fall back to snappy.
```

## Reminder for the reviewer

Other optional labels:


typha/pkg/syncserver/snap_precalc.go
```diff
@@ -23,48 +23,58 @@ import (
...
 	counterVecSnapshotsGenerated = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "typha_snapshots_generated",
 		Help: "Number of binary snapshots generated.",
-	}, []string{"syncer"})
+	}, snapshotMetricLabels)
 	counterVecSnapshotsReused = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "typha_snapshots_reused",
 		Help: "Number of binary snapshots that were cached and reused.",
-	}, []string{"syncer"})
+	}, snapshotMetricLabels)
```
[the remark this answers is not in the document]
hjiawei: The in-repo consumers are unaffected: the Grafana dashboard in `manifests/grafana-dashboards.yaml` aggregates these metrics (`max(typha_snapshot_raw_bytes)`, `max(typha_snapshot_compressed_bytes)`), and the FV tests only assert on `typha_connections_grace_used`, which has no `compression` label (the snapshot metrics are only logged). For external queries the label is additive: `sum without (compression)` recovers the old view. Per-compression visibility is the point of the label — it is how we can watch a zstd rollout. Keeping parallel metric names would double the metric surface for little gain, and all label combinations are pre-created at startup so scrapes stay stable.

typha/pkg/syncclient/sync_client.go
```diff
@@ -167,11 +175,25 @@ type SyncerClient struct {
...
 }
 
+// NegotiatedCompressionAlgorithm returns the compression algorithm the server
+// selected for the current connection, or "" if the stream is uncompressed
+// (or compression has not been negotiated yet).  Safe to call from any
+// goroutine; intended for tests and diagnostics.
+func (s *SyncerClient) NegotiatedCompressionAlgorithm() syncproto.CompressionAlgorithm {
+	alg, _ := s.negotiatedCompression.Load().(syncproto.CompressionAlgorithm)
+	return alg
+}
```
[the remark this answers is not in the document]
hjiawei: `atomic.Value.Load` does not panic before the first `Store` — it returns nil (it is `Store` that panics, on nil or on inconsistent types). The comma-ok type assertion turns that nil into "", which is what the doc comment promises for the not-yet-negotiated case.

hjiawei:
All green run at https://tigera.semaphoreci.com/workflows/d925463e-3a27-4882-a0a3-63d8ad31d6fc?pipeline_id=fd7ef8ab-b7b9-4b6f-a792-1da946bfd501.

typha/pkg/syncclient/sync_client.go
```diff
@@ -167,11 +175,25 @@ type SyncerClient struct {
 	connR      io.Reader
 	encoder    *gob.Encoder
 	decoder    *gob.Decoder
+	zstdReader *zstd.Decoder
```
fasaxc: Feel like this should be an interface field, the client shouldn't need to know what compression it's using once the reader is set up.

typha/pkg/syncclient/sync_client.go
```diff
@@ -553,12 +585,35 @@ func (s *SyncerClient) restartDecoder(cxt context.Context, logCxt *log.Entry, ms
...
+		logCxt.Info("Server selected zstd compression.")
+		if s.zstdReader == nil {
+			r, err := zstd.NewReader(s.connR, zstd.WithDecoderConcurrency(1))
+			if err != nil {
+				logCxt.WithError(err).Error("Failed to create zstd reader")
+				return err
+			}
+			s.zstdReader = r
+		}
+		// Reuse the existing zstd reader across decoder restarts.  It may
```
fasaxc: This doesn't sound right, the decoder restart should be synchronous even if the decoder impl isn't.  We should drain the previous decoder fully; stop it and start a fresh one.

typha/pkg/syncserver/sync_server.go
```diff
@@ -1063,16 +1162,55 @@ func (h *connection) waitForAckAndRestartEncoder() error {
...
+func (h *connection) waitForAckAndRestartEncoder() error {
+	if err := h.waitForAck(); err != nil {
+		return err
+	}
+	return h.restartEncoder()
+}
+
+// restartEncoder creates a new encoder on the connection, optionally wrapping
+// it with a compression writer based on h.chosenCompression.
+func (h *connection) restartEncoder() error {
```
fasaxc, line 1215: _Claude:_ This violates the intended encoder-restart sequence, though the bug is latent today. The intent was: the `MsgDecoderRestart` is the last message into the *old* encoder, then the old compression writer is flushed **and closed** so its stream terminator reaches the wire, and only then does the new encoder start.

As written, `sendDecoderRestartAndWaitForAck` sends the restart with only `Flush()` (zstd frame left open), and this `Close()` runs after the ACK, writing the frame epilogue into the old, already-abandoned `bufio.Writer` — it never reaches the wire.

This can't corrupt anything today because the encoder restarts exactly once per connection, always from the uncompressed state (a fall-behind client is disconnected, never re-snapshotted in place), so `compressCloser` is always nil at the sole `restartEncoder` call. But the close-previous logic here falsely advertises that repeated mid-stream restarts work. The first change that adds one (e.g. re-snapshotting a lagging client) corrupts zstd streams — the client's reused decoder hits the next frame's magic mid-frame — while snappy keeps working, which makes it a nasty zstd-only trap.

With the compressor-interface approach from the review summary, this becomes: send restart → `compressor.Close()` (flushes through to the wire, terminates the stream) → wait for ACK → build new compressor. Failing that, please replace this dead block with an assertion that `compressCloser == nil` plus a comment stating that encoder restarts are once-per-connection.

typha/pkg/syncclient/sync_client.go
```diff
@@ -553,12 +585,35 @@ func (s *SyncerClient) restartDecoder(cxt context.Context, logCxt *log.Entry, ms
...
+		// hold input that it read ahead from the connection, and it decodes
+		// consecutive zstd frames as one continuous stream, so no bytes are
+		// lost at the frame boundary.  We only need a new gob decoder since
+		// the server restarted its gob encoder.
+		s.decoder = gob.NewDecoder(s.zstdReader)
```
fasaxc: _Claude:_ Adding detail to the comment above: the claim "no bytes are lost at the frame boundary" is true today, but only because of an unstated server-side invariant — every zstd segment that precedes a `MsgDecoderRestart` must be a *terminated* frame whose final decoded payload ends exactly at the restart message. Right now the only such segment is the cached snapshot, which `snap_precalc.go` writes with `w.Close()`. Nothing enforces this on the server, and `DESIGN.md` documents only the ACK invariant.

It also leaves two contradictory models side by side: snappy makes a fresh reader on every restart, zstd reuses one and treats the connection as a continuous stream. A future maintainer has no way to tell which model is load-bearing.

A decompressor interface that guarantees no read-ahead past what it has returned (for zstd: synchronous mode + terminated frames) would let the client drop and recreate the decoder on every restart, uniformly, matching the snappy model and the original design intent.

typha/pkg/syncserver/sync_server.go
```diff
@@ -787,7 +862,18 @@ func (h *connection) handle(finishedWG *sync.WaitGroup) (err error) {
...
+				// Binary snapshot will follow: send MsgDecoderRestart and wait
+				// for ACK, but don't create a compressed encoder yet -- the
+				// snapshot is sent as raw bytes directly to the connection,
+				// bypassing the encoder.  The real encoder is created after
+				// the snapshot.
+				err = h.sendDecoderRestartAndWaitForAck(strings.Join(reasonsToRestart, ";"))
+			} else {
+				// No binary snapshot: create the compressed encoder now so
+				// the streamed snapshot goes through it.
+				err = h.restartEncodingIfSupported(strings.Join(reasonsToRestart, ";"))
```
fasaxc: _Claude:_ This `else` branch is unreachable: `reasonsToRestart` is non-empty only when `chosenCompression != ""`, and any chosen algorithm has a snapshot cache (`New()` creates one per configured algorithm × syncer type, and `doHandshake` rejects unknown syncer types), so `binSnapCache` can't be nil here. That makes `restartEncodingIfSupported` dead code, and its comment describes a "compressed streamed snapshot" flow that can't occur — anyone debugging the restart protocol has to reason about a third path that doesn't exist.

Suggest calling `sendDecoderRestartAndWaitForAck` directly with a comment (or an explicit invariant check) instead of keeping the branch.

fasaxc (commented):
_Claude:_ (summary written by Claude on Shaun's behalf; the design intent described here is Shaun's)

Thanks for this — the negotiation back-compat, the ACK gating, the snapshot cache plumbing, and the tests all look solid. The concerns below are about how the PR handles the encoder/decoder restart boundary for zstd, which diverges from the design intent of the restart protocol.

**Design intent for the restart protocol.** `MsgDecoderRestart` was meant to be a hard boundary on both sides:

- **Server/encoder side:** the restart message is the *last* message into the old encoder. The old encoder is then flushed and closed properly — so every byte, including any stream terminator, reaches the wire — before the new encoder starts.
- **Client/decoder side:** after the restart message, the old decoder is stopped and fully drained; the new decoder starts fresh from the connection with no state carried over.

The PR instead caches the zstd writer/reader across the boundary. That happens to be safe today, but only because of invariants that aren't written down or enforced (details in the inline comments): the encoder restarts exactly once per connection from the uncompressed state, and the only zstd stream that precedes a restart is the cached snapshot, which is a properly `Close()`d frame.

**Suggested direction.** The root problem is that snappy's reader/writer are synchronous while zstd's are (by default) asynchronous, and the two sides of this PR handle that difference in two different, undocumented ways. I think the clean fix is to introduce a small compressor/decompressor interface pair that *enforces* the synchronous contract the restart protocol needs, even if an implementation uses internal goroutines:

- a **compressor** interface whose `Flush()`/`Close()` guarantee that all accepted input has been compressed and written through to the underlying writer (terminating the stream/frame on `Close()`);
- a **decompressor** interface that guarantees it reads from the underlying connection only the bytes it needs (no speculative read-ahead past a frame boundary), so it can be discarded at a restart without losing connection bytes.

The server and client should then use compression only through these interfaces. That documents the contract in one place, makes the restart sequence identical for every algorithm (restart is the last message in → flush → close → new compressor; decoder discarded and recreated on restart), and stops correctness from hinging on per-library details like `zstd.WithDecoderConcurrency(1)`.

typha/pkg/syncclient/sync_client.go
hjiawei: Done in 8374de2a62: the client now holds a `syncproto.Decompressor` interface field; the concrete snappy/zstd readers live behind `syncproto.NewDecompressor()`.

typha/pkg/syncclient/sync_client.go
hjiawei: Reworked in 8374de2a62: the client now discards its decompressor and creates a fresh one on every restart, uniformly for all algorithms. The drain is guaranteed on the sender side: the server ends the old stream with the restart message as its final data (`syncproto.CloseWithFinalMessage`: flush, write the message, close), so once the client decodes the restart message the old stream is exactly consumed — klauspost's synchronous mode consumes a frame's trailing checksum in the same Read that returns the frame's final bytes.

Making this uniform surfaced a real edge case, now fixed: zstd emits an *empty* final block when Close finds nothing pending (stream data ending exactly on the 128KiB block boundary), and the synchronous decoder never consumes an empty final block — those bytes would be stranded at the boundary and corrupt the next stream. The flush-then-final-message-then-close sequence exists to prevent that; it's enforced in one helper and pinned by `TestCompressionBlockAlignedBoundary`.

typha/pkg/syncserver/sync_server.go
hjiawei: Fixed in 8374de2a62: `sendDecoderRestartAndWaitForAck` now ends the old stream via `syncproto.CloseWithFinalMessage` — the restart message is the last message into the old encoder, and the writer is closed (terminator on the wire) before we wait for the ACK — so repeated mid-stream restarts are safe for every algorithm. The dead close-previous block in `restartEncoder` is gone; each encoding is a self-contained stream built by `syncproto.NewStreamCompressor`.

typha/pkg/syncclient/sync_client.go
hjiawei: Done in 8374de2a62: added `syncproto.Compressor`/`Decompressor` with the synchronous, no-read-ahead contract documented in one place; the server and client access compression only through them, and snappy/zstd now follow the same fresh-reader-per-restart model. `TestCompressionRestartBoundaries` exercises the full uncompressed → snapshot → delta boundary chain for every algorithm, and `TestCompressionBlockAlignedBoundary` pins the block-aligned edge cases (including the empty-final-block case where zstd would otherwise strand bytes at the boundary). DESIGN.md's switching-invariant section now documents the hard boundary on both sides.

typha/pkg/syncserver/sync_server.go
hjiawei: Removed in 8374de2a62: `restartEncodingIfSupported` is deleted and the handshake path calls `sendDecoderRestartAndWaitForAck` directly. The impossible compressed-but-no-cache state now has an explicit check (`ErrNoSnapshotCache`), and `New()` panics at startup if an algorithm in the preference order has no snapshot cache implementation, so drift is caught at boot rather than as a per-client connect loop.

hjiawei:
Adopted the suggested direction in 8374de2a62: a `Compressor`/`Decompressor` interface pair in `pkg/syncproto` now carries the synchronous contract the restart protocol needs, and both server and client use compression only through it. The restart sequence is the same for every algorithm: the restart message is the last data into the old compressor, which is flushed and closed (terminator on the wire) before the ACK; the client discards its decompressor at each restart and starts fresh.

Two things came out of making the model uniform:

- A real edge case in the fresh-reader model: zstd emits an *empty* final block when `Close` finds nothing pending (stream data ending exactly on the 128KiB block boundary), and the synchronous decoder never consumes an empty final block, so those bytes would corrupt the next stream. For the cached snapshot that was a ~1-in-131072-sizes trigger. Streams now end via `CloseWithFinalMessage` (flush, write the final message, close), which keeps the final message inside a non-empty final block; `TestCompressionBlockAlignedBoundary` pins the aligned sizes and fails if the discipline is dropped.
- `TestCompressionRestartBoundaries` models the whole handshake → cached snapshot → delta-stream lifecycle over a pipe (with the ACK gating) for none/snappy/zstd, so repeated mid-stream restarts are covered, not just the paths the protocol takes today.

Also per the inline comments: the unreachable `restartEncodingIfSupported` path is deleted, and a missing snapshot-cache implementation now fails at startup rather than per connection. Typha unit tests, the new contract tests, the full typha fv-tests suite under `-race`, and felix unit tests all pass.

hjiawei:
Appended 038bdf28de (no rebase): `typha_connections_streaming` and `typha_client_snapshot_send_secs` now carry the `compression` label alongside `syncer`, matching the snapshot metrics; uncompressed connections report `compression="none"`.

Motivation: while validating this branch on a real cluster (rolling typha between snappy and zstd builds), the negotiated algorithm per connection was unobservable from metrics — watching the fleet drain from snappy to zstd during an upgrade required grepping client logs. The two metrics are now bound per connection after the handshake; all syncer × compression series are still pre-created at startup. Covered by the negotiation FV tests (zstd, snappy, and none label values) plus a unit test for the label mapping.

typha/pkg/syncclient/sync_client.go
```diff
@@ -167,11 +173,34 @@ type SyncerClient struct {
...
+	// decompressor is the (possibly pass-through) reader that the decoder
+	// reads from; it wraps connR.  Replaced together with the decoder at
+	// each decoder restart.
+	decompressor syncproto.Decompressor
+
+	// advertisedCompressionAlgs is the set of compression algorithms we
+	// offered in our hello message.  A MsgDecoderRestart naming any other
+	// algorithm is a protocol violation and drops the connection.  Only
+	// touched by the loop goroutine.
+	advertisedCompressionAlgs map[syncproto.CompressionAlgorithm]bool
```
fasaxc, line 185: Typha generally uses `set.Set[...]` instead of `map[...]bool`

typha/pkg/syncclient/sync_client.go
```diff
@@ -548,16 +593,27 @@ func (s *SyncerClient) loop(cxt context.Context, cancelFn context.CancelFunc, co
...
-		logCxt.Info("Server selected no compression.")
-		s.decoder = gob.NewDecoder(s.connR)
+	logCxt.WithFields(log.Fields{
+		"msg":         msg,
+		"compression": msg.CompressionAlgorithm,
+	}).Info("Server asked us to restart our decoder")
+	// The server must pick from the algorithms we advertised (or none).
+	// Anything else is a protocol violation; don't decode with an algorithm
+	// we never offered.
+	if msg.CompressionAlgorithm != "" && !s.advertisedCompressionAlgs[msg.CompressionAlgorithm] {
```
fasaxc, line 603: I might have added "" to the set so that the "no-op compressor" is a first-class peer of the others.

typha/pkg/syncproto/compression.go
```diff
@@ -0,0 +1,249 @@
...
+//     Close with no flush in between.  Close pushes every remaining byte --
+//     including any stream terminator -- through to the connection, and the
+//     sender sends nothing more until the client ACKs.  The next bytes on
+//     the wire begin a brand-new stream from a brand-new Compressor.  The
+//     flush-before-the-final-message step pins the message (much smaller
+//     than a compression block) inside the stream's final block; without
+//     it, a stream whose data ends exactly on a block boundary would end in
+//     an *empty* final block, which zstd's synchronous decoder never
+//     consumes -- those bytes would be stranded on the wire and corrupt the
+//     next stream.
```
fasaxc, line 41: comment could go on the Compressor object?  Or move this to a new package and put it in the package doc

This comment mixes implementation detail into the "spec".  I think it's be better to state the strict requirements: After you call CloseWithFinalMessage:
 
- the client received the final message as normal
- the decompressor is guaranteed to also read any trailing bytes from the stream that are produced by the compressor
- compressor is closed for new input and ready to throw away
- decompressor is ready to be closed
- stream is free for further data from some other source.

That said, it does seem like a bug that the zstd implementation produces an empty block when you `Close()` if you already `Flush()`ed?  Could we fix it or filter that out client-side so we don't need to remember to use a special operation?

typha/pkg/syncproto/compression.go
```diff
@@ -0,0 +1,249 @@
...
+	// through to the underlying writer.  It does not end the compressed
+	// stream.
+	Flush() error
+
+	// Close ends the compressed stream and writes it, including any stream
+	// terminator, through to the underlying writer.  The Compressor must not
+	// touch the underlying writer after Close returns.  To end a stream that
+	// a Decompressor must consume exactly, close it via
+	// CloseWithFinalMessage rather than calling Close directly.
+	Close() error
```
fasaxc, line 85: This is a foot gun, wonder if we should make this CloseUnclean or something so that it's clear that it's for the error path only

typha/pkg/syncproto/compression.go
```diff
@@ -0,0 +1,249 @@
...
+	// the underlying reader.
+	Close()
+}
+
+// CloseWithFinalMessage ends a compressed stream with a final message: it
+// flushes the Compressor, calls encodeMsg to write the message, then closes
+// the Compressor.  This is the only safe way to end a stream that a
+// Decompressor must consume exactly: the flush pins the message inside the
+// stream's final block (see the package comment above).
+func CloseWithFinalMessage(c Compressor, encodeMsg func() error) error {
```
fasaxc, line 107: This needs to be part of the interface so that the snappy and zstd impls can do it differently.  Snappy doesn't need the special "flush, send, close" treatment, it can just do "send and close".

typha/pkg/syncproto/compression.go
```diff
@@ -0,0 +1,249 @@
...
+	case "":
+		return nopCloserDecompressor{r}, nil
+	default:
+		return nil, fmt.Errorf("unknown compression algorithm: %q", algorithm)
+	}
+}
+
+// streamCompressor pairs a compression writer with the buffered writer
+// beneath it so that Flush and Close write through to the destination.
+type streamCompressor struct {
```
fasaxc, line 200: I think it's worth having SnappyCompressor and ZstdCompressor so there's a natural place for any workaround to live

fasaxc (changes requested):
A few thoughts.  I think it's getting close but I'd like to see if there's a way to solve the final empty block problem without a special API.  My gut says that the zstd compressor and zstd decompressor should just handle that somehow and make send, flush, close "just work".  Right now the workaround has leaked into the API, when it should really stay in the particular compressor/decompressor.

I think what we really want is for `s.decompressor.Close()` in `swapDecompressor` to consume the trailing bytes or for the sender not to send them; after the sender `Flush()`es the final message, we know it'll arrive, so we could just stop writing on Close()?

Asked Claude to do some diggging:

---

Why there's no option to omit the last block. The zstd frame format requires a block with Last_Block=1. Flush() only ever emits non-last blocks; Close() is the only path that sets the flag, and if Flush() already emptied the buffer it emits an empty raw last block (3 bytes) plus the 4-byte CRC. WithEncoderCRC(false) drops the CRC but not the block. WithZeroFrames only covers the "nothing was ever written" case. There is no "flush as final block" API.

Why the empty block is the actual hazard. In sync mode the decoder's nextBlockSync loops for len(current.b) == 0, so an empty last block is never surfaced to the caller: it consumes the block and CRC, then immediately parses the next frame header and its first block. That's how the old decoder eats the new decoder's bytes. A last block that carries data is fine, since the CRC is consumed in the same call and Read returns. That's the property the PR pins with the final-message trick.

Cleaner option: never terminate the frame. Make the zstd Compressor.Close() do Flush() then drop the encoder (Reset(nil), or just let it go; concurrency 1 spawns no goroutines). The wire then reads: non-last block ending in the restart message, ACK gate, new frame magic. This is safe because:

- In sync mode nextBlock(blocking=false) returns immediately, and Read returns as soon as it has any bytes, so the old decoder never reads past the block containing the restart message.
- The ACK gate means nothing follows on the wire until the client has already discarded the old decoder anyway.
- Snappy's Close() is already a wire no-op after Flush(), so both algorithms get identical semantics: Flush() is the boundary, Close() releases resources and writes nothing. No special "last message" call in the interface.

Costs: the abandoned frame has no terminator or CRC, so it's "truncated" by strict zstd standards. Only our own decoder ever sees it, and TCP/TLS already give integrity, but it's a protocol-level choice that DESIGN.md should state explicitly. The one thing the interface must then enforce is sync mode (WithDecoderConcurrency(1) / WithEncoderConcurrency(1)), since the async decoder does read ahead.

The alternative of keeping terminated frames and having the client consume the 3-or-7-byte trailer directly from the connection is possible but couples the client to zstd's block-header layout. I'd not recommend it.

result: klauspost zstd has no option to suppress the empty final block (format requires Last_Block=1); recommended alternative is to never terminate the frame (Flush as the boundary, Close writes nothing), which is safe with sync-mode decoding plus the ACK gate.

typha/pkg/syncproto/compression_test.go
```diff
@@ -0,0 +1,420 @@
...
+					_ = pw.CloseWithError(err)
+				}
+				writerErr <- err
+			}()
+
+			// Stream 0: handshake phase, always uncompressed.
+			hostnames, err := readTestStream(CompressionNone, pr, true)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(hostnames).To(Equal(expectedHostnames(0)))
+			acks <- struct{}{}
```
fasaxc, line 71: Watch out for deadlocks if the test or code under test has bugs; should we put a timeout on these sends/receives?

typha/pkg/syncclient/sync_client.go
```diff
@@ -566,6 +626,22 @@ func (s *SyncerClient) restartDecoder(cxt context.Context, logCxt *log.Entry, ms
 	return err
 }
 
+// swapDecompressor closes the current decompressor (if any) and installs a
+// fresh one reading from the connection, along with a fresh gob decoder.
+func (s *SyncerClient) swapDecompressor(algorithm syncproto.CompressionAlgorithm) error {
+	d, err := syncproto.NewDecompressor(algorithm, s.connR)
+	if err != nil {
+		return err
+	}
```
fasaxc, line 635: Think I'd close the old one first.  Don't think there's a bug right now, but it seems less error prone to have exactly one active compressor with a pointer to the reader.

typha/pkg/syncserver/sync_server.go
```diff
@@ -1063,29 +1195,45 @@ func (h *connection) waitForAckAndRestartEncoder() error {
...
+// sendMsg sends a message to the client and flushes it through to the
+// connection.  It may be called from multiple goroutines.
 func (h *connection) sendMsg(msg any) error {
+	return h.sendMsgMaybeFlush(msg, true)
+}
+
+// sendMsgWithoutFlush sends a message to the client, leaving it pending in
+// the writer.  Used only for the final message of a stream, which the caller
+// pushes out by closing the writer (see sendDecoderRestartAndWaitForAck).
+func (h *connection) sendMsgWithoutFlush(msg any) error {
```
fasaxc, line 1232: _Claude:_ Now that `Close()` writes no terminator, deferring this flush is an optimisation, not a protocol requirement: zstd may emit a block mid-message anyway when its 128KiB buffer fills, and that is exactly what `Flush()` would emit, so the stream ends on the last data-carrying block either way and the client stops on that byte. The comment ("pushes out by closing the writer") reads as if the deferred flush matters.

Suggest dropping `sendMsgWithoutFlush`/`sendMsgMaybeFlush` and having `sendDecoderRestartAndWaitForAck` call `sendMsg` then `h.writer.Close()`: one fewer special-case API, and no way to mistake the unflushed send for part of the contract.

fasaxc (commented):
_Claude:_ (on Shaun's behalf) Re-checked head 5058565e against the 7 Sept comments: all six are addressed, and the boundary design (Flush is the boundary, Close writes no terminator) checks out against the klauspost v1.20.0 encoder/decoder internals. Resolving those threads.

One CI blocker: `Check Go files` fails `make check-dirty` because `.semaphore/semaphore.yml` is stale. Typha now imports `libcalico-go/lib/set` and `typha/deps.txt` was regenerated, but the semaphore template output wasn't; it needs `'/libcalico-go/lib/set/*.go'` in typha's `change_in` list. `make gen-semaphore-yaml` and commit the result.

typha/pkg/syncserver/sync_server.go
fasaxc, line 1232: Don't think this is needed now?  I'm inclined to leave it out and make doubly sure that Flush followed by Close is a no-op on the wire.

fasaxc (changes requested):
Just some really small ones

PR #11854, opened 2026-02-16
Add direct-mapped cache to uniquelabels.Make for repeated inputs

fasaxc:
It is common for Make to be called in quick succession with the same input map from different call sites. This adds a small (128-entry) direct-mapped cache that avoids redundant handleMap allocations when the input matches a recently-computed Map.

Shows a decent win on the `IsolatedCustomers` benchmark:

Before:
```
goos: linux
goarch: amd64
BenchmarkIsolatedCustomers1k-12      	      25	  45382677 ns/op	        10.13 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	      1403 Msgs	       100.0 Policies	         0.04368 s	16544401 B/op	  191707 allocs/op
BenchmarkIsolatedCustomers10k-12     	       3	 388005473 ns/op	        47.56 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	     10403 Msgs	       100.0 Policies	         0.3810 s	127345704 B/op	 1471678 allocs/op
BenchmarkIsolatedCustomers100k-12    	       1	6898444469 ns/op	       401.7 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	    100403 Msgs	       100.0 Policies	         6.898 s	1213249056 B/op	14929873 allocs/op
PASS
ok  	github.com/projectcalico/calico/felix/calc	19.689s
```

After:
```
goos: linux
goarch: amd64
BenchmarkIsolatedCustomers1k-12      	      24	  42850227 ns/op	         9.702 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	      1403 Msgs	       100.0 Policies	         0.04367 s	16167045 B/op	  187753 allocs/op
BenchmarkIsolatedCustomers10k-12     	       3	 374682753 ns/op	        43.13 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	     10403 Msgs	       100.0 Policies	         0.3669 s	123510848 B/op	 1431711 allocs/op
BenchmarkIsolatedCustomers100k-12    	       1	6759899958 ns/op	       357.4 HeapAllocMB	       100.0 IPSets	       100.0 LocalEps	    100403 Msgs	       100.0 Policies	         6.760 s	1174845072 B/op	14529950 allocs/op
PASS
ok  	github.com/projectcalico/calico/felix/calc	19.250s
```

**Release note**:
```release-note
Felix now keeps a cache of recently computed interned label maps.  This reduces memory usage, particularly with large deployments or large numbers of namespaces.
```

### DIFF
--- a/lib/std/uniquelabels/cache.go
+++ b/lib/std/uniquelabels/cache.go
@@ -23,19 +23,19 @@ import (
 // It is common for Make to be called in quick succession with the same input
 // from different call sites. Since the keys and values are interned, the
 // handleMap is the main allocation that can be avoided by caching.
-var recentCache = makeCache{
+var recentCache = recentMapCache{
 	seed: maphash.MakeSeed(),
 }
 
-const makeCacheSize = 128 // Must be a power of two.
+const recentMapCacheSize = 128 // Must be a power of two.
 
-type makeCache struct {
+type recentMapCache struct {
 	mu      sync.Mutex
 	seed    maphash.Seed
-	entries [makeCacheSize]makeCacheEntry
+	entries [recentMapCacheSize]recentMapCacheEntry
 }
 
-type makeCacheEntry struct {
+type recentMapCacheEntry struct {
 	hash uint64
 	m    Map
 }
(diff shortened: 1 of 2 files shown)


fasaxc:
You can think of it as picking positions in the cache essentially be random (but with the guarantee that the exact same label map will always go in the same random slot).  So, the upshot of that is:

* If you have a handful of entries, they're unlikely to collide.
* As you get past `sqrt(cache size)` the birthday paradox will kick in and you're likely to start getting conflicts.
* Since this is a best-effort cache, I'm resolving conflicts by "newest wins, older gets thrown away".

You may get lucky or not, but, given the birthday paradox rule of thumb, you're likely to get lucky if the labels you're interning is one of the last ~10 that you've seen and guaranteed to get lucky if it's the most recently seen one.   That guarantee solves the problem that the user was seeing (where the same labels get interned three times in the same calc graph iteration); I'm hoping that the "probably get a hit for the most recent 10" helps with deployments, where we often see 10 pods created back-to-back with the same labels (but possibly mixed in with other pods being started).

PR #11863, opened 2026-02-18
Additional validation for the v3 API

caseydavenport:
## Summary

Adds comprehensive validation to the Calico v3 API using kubebuilder CEL XValidation rules, defaults, and constraint markers.

**Key changes:**
- **28 CEL XValidation rules** across 10+ CRD types enforcing cross-field validation (e.g., mutual exclusion of CIDR/selector fields, conditional field requirements, ICMP type/code constraints)
- **Default values** for `TierSpec.DefaultAction` (Deny) and staged policy `StagedAction` fields (Set)
- **MaxLength/MaxItems constraints** on fields referenced in CEL rules to stay within the K8s API server cost budget
- **Boundary corrections**: `EndpointPort.Port` minimum changed from 0→1, `ICMPFields.Type` maximum from 255→254
- **PrintColumn cleanup**: removed redundant columns across multiple CRDs
- **Short name normalization**: moved short names from resource types to list types
- **BGPFilterMatchOperator** constants renamed with `MatchOperator` prefix
- **New envtest-based validation test suite** (`api/test/validation/`) with 11 test files covering all new rules

**Minimum Kubernetes version:** 1.29 (CEL XValidation GA)

### DIFF
--- a/api/config/crd/projectcalico.org_bgpfilters.yaml
+++ b/api/config/crd/projectcalico.org_bgpfilters.yaml
@@ -36,6 +36,7 @@ spec:
                         type: string
                       cidr:
                         format: cidr
+                        maxLength: 18
                         type: string
                       interface:
                         type: string
@@ -90,6 +91,7 @@ spec:
                         type: string
                       cidr:
                         format: cidr
+                        maxLength: 43
                         type: string
                       interface:
                         type: string
@@ -144,6 +146,7 @@ spec:
                         type: string
                       cidr:
                         format: cidr
+                        maxLength: 18
                         type: string
                       interface:
                         type: string
@@ -198,6 +201,7 @@ spec:
                         type: string
                       cidr:
                         format: cidr
+                        maxLength: 43
                         type: string
                       interface:
                         type: string
--- a/api/config/crd/projectcalico.org_bgppeers.yaml
+++ b/api/config/crd/projectcalico.org_bgppeers.yaml
@@ -72,6 +72,7 @@ spec:
                   format: int32
                   type: integer
                 localWorkloadSelector:
+                  maxLength: 4096
                   type: string
                 maxRestartTime:
                   type: string
@@ -82,8 +83,10 @@ spec:
                     - Keep
                   type: string
                 node:
+                  maxLength: 253
                   type: string
                 nodeSelector:
+                  maxLength: 4096
                   type: string
                 numAllowedLocalASNumbers:
                   format: int32
@@ -105,8 +108,10 @@ spec:
                       x-kubernetes-map-type: atomic
                   type: object
                 peerIP:
+                  maxLength: 64
                   type: string
                 peerSelector:
+                  maxLength: 4096
                   type: string
                 reachableBy:
                   type: string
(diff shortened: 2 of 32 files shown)


api/config/crd/projectcalico.org_ippools.yaml
```diff
@@ -98,6 +111,27 @@ spec:
               required:
                 - cidr
               type: object
+              x-kubernetes-validations:
```
coutinhop, line 115: I guess we can't have the equivalent of https://github.com/projectcalico/calico/blob/8970f92f74c975e6290bac8e2c646e1f0c408623/libcalico-go/lib/validator/v3/validator.go#L1154-L1158 in CEL?

caseydavenport: We might! Let me play with it a bit.

caseydavenport: Yep, we can! Pushed a commit that adds this. Uses `!self.cidr.contains(':')` to detect IPv6 (since IPv6 CIDRs always have colons and IPv4 never does). Also added `MaxLength=48` on the CIDR field to keep the CEL cost budget happy.

caseydavenport:
### Code review

Found 3 issues:

1. CEL rule `self.protocol == 'TCP'` does not handle numeric protocol values. The `protocol` field is `int-or-string`, so `protocol: 6` (TCP's IANA number) is valid but would fail this string comparison, rejecting policies with HTTP match and numeric protocol.

https://github.com/projectcalico/calico/blob/a3a4997c7254f3b0bfdebba255c4469ea7e9b404/api/pkg/apis/projectcalico/v3/policy_common.go#L36-L38

2. Several string fields used in CEL `size()` expressions are missing `+kubebuilder:validation:MaxLength`. Without it, the K8s CEL cost estimator assumes unbounded strings and may exceed the cost budget. Affected fields: `Node`, `NodeSelector`, `PeerIP`, `PeerSelector`, `LocalWorkloadSelector` on `BGPPeerSpec`; `Node`, `InterfaceName` on `HostEndpointSpec`; `CIDR` on `BGPFilterRuleV4`/`V6`. ([CLAUDE.md](https://github.com/projectcalico/calico/blob/a3a4997c7254f3b0bfdebba255c4469ea7e9b404/.claude/CLAUDE.md) says "String operations cost proportional to maxLength -- always set +kubebuilder:validation:MaxLength")

https://github.com/projectcalico/calico/blob/a3a4997c7254f3b0bfdebba255c4469ea7e9b404/api/pkg/apis/projectcalico/v3/bgppeer.go#L56-L68

3. `self == oldSelf` on IPPool CIDR enforces exact string equality on updates, but `libcalico-go/lib/clientv3/ippool.go`'s `cidrChangeOK` allows canonical form normalization (e.g., `10.0.0.1/24` to `10.0.0.0/24`). If an existing pool has a non-canonical CIDR, updates to it through the CRD path will be blocked.

https://github.com/projectcalico/calico/blob/a3a4997c7254f3b0bfdebba255c4469ea7e9b404/api/pkg/apis/projectcalico/v3/ippool.go#L90-L92

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #11868, opened 2026-02-18
Define LiveMigration resource

nelljerram:
The LiveMigration resource is an ingredient in enabling seamless live migration for KubeVirt and OpenStack, by associating the source and target pods or VMs.  In KubeVirt it's backed by the KubeVirt VirtualMachineInstanceMigration resource.  In OpenStack our Neutron driver creates it and writes it to the etcd datastore.

Key components:
- LiveMigration type definition (internalapi) with spec fields for source and destination workload endpoints.
- K8s backend client using `kubevirt/client-go` to read KubeVirt VirtualMachineInstanceMigration resources (read-only: List, Get, Watch).  Populate LiveMigrationSpec from VMIM fields (vmiName, uid, sourcePod), filter VMIMs by phase (TargetReady/Running/Failed) in Get/List/Watch.
- clientv3 LiveMigration client with standard List/Get operations in case we later want to add calicoctl support for LiveMigration.  Note that CRUD and Watch are not supported here and will return errors.
- Model, namespace, and scheme registrations.
- Felix syncer for LiveMigration.

The K8s backend is deliberately read-only since LiveMigration state is owned by KubeVirt. The etcdv3 backend stores LiveMigration directly as a custom resource (unchanged default behavior).

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

**Release Note:**
```release-note
Add RBAC for Calico and the Tigera operator to watch KubeVirt migration resources, in preparation for upcoming full support for KubeVirt live migration.
```

### DIFF
--- a/libcalico-go/lib/clientv3/livemigration_e2e_test.go
+++ b/libcalico-go/lib/clientv3/livemigration_e2e_test.go
@@ -28,7 +28,6 @@ import (
 	"github.com/projectcalico/calico/libcalico-go/lib/clientv3"
 	"github.com/projectcalico/calico/libcalico-go/lib/options"
 	"github.com/projectcalico/calico/libcalico-go/lib/testutils"
-	"github.com/projectcalico/calico/libcalico-go/lib/watch"
 )
 
 var _ = testutils.E2eDatastoreDescribe("LiveMigration tests", testutils.DatastoreEtcdV3, func(config apiconfig.CalicoAPIConfig) {
@@ -222,7 +221,7 @@ var _ = testutils.E2eDatastoreDescribe("LiveMigration tests", testutils.Datastor
 	)
 
 	Describe("LiveMigration watch functionality", func() {
-		It("should handle watch events for different resource versions and event types", func() {
+		It("should return an error for an attempted Watch", func() {
 			c, err := clientv3.New(config)
 			Expect(err).NotTo(HaveOccurred())
 
@@ -234,9 +233,8 @@ var _ = testutils.E2eDatastoreDescribe("LiveMigration tests", testutils.Datastor
 			outList, outError := c.LiveMigrations().List(ctx, options.ListOptions{})
 			Expect(outError).NotTo(HaveOccurred())
 			Expect(outList.Items).To(HaveLen(0))
-			rev0 := outList.ResourceVersion
 
-			By("Configuring a LiveMigration namespace1/name1/spec1 and storing the response")
+			By("Configuring a LiveMigration namespace1/name1/spec1")
 			outRes1, err := c.LiveMigrations().Create(
 				ctx,
 				&internalapi.LiveMigration{
@@ -248,8 +246,8 @@ var _ = testutils.E2eDatastoreDescribe("LiveMigration tests", testutils.Datastor
 			rev1 := outRes1.ResourceVersion
 			Expect(err).NotTo(HaveOccurred())
 
-			By("Configuring a LiveMigration namespace2/name2/spec2 and storing the response")
-			outRes2, err := c.LiveMigrations().Create(
+			By("Configuring a LiveMigration namespace2/name2/spec2")
+			_, err = c.LiveMigrations().Create(
 				ctx,
 				&internalapi.LiveMigration{
 					ObjectMeta: metav1.ObjectMeta{Namespace: namespace2, Name: name2},
@@ -259,112 +257,9 @@ var _ = testutils.E2eDatastoreDescribe("LiveMigration tests", testutils.Datastor
 			)
 			Expect(err).NotTo(HaveOccurred())
 
-			By("Starting a watcher from revision rev1 - this should skip the first creation")
-			w, err := c.LiveMigrations().Watch(ctx, options.ListOptions{ResourceVersion: rev1})
-			Expect(err).NotTo(HaveOccurred())
-			testWatcher1 := testutils.NewTestResourceWatch(config.Spec.DatastoreType, w)
-			defer testWatcher1.Stop()
-
-			By("Deleting res1")
-			_, err = c.LiveMigrations().Delete(ctx, namespace1, name1, options.DeleteOptions{})
-			Expect(err).NotTo(HaveOccurred())
-
-			By("Checking for two events, create res2 and delete res1")
-			testWatcher1.ExpectEvents(internalapi.KindLiveMigration, []watch.Event{
-				{
-					Type:   watch.Added,
-					Object: outRes2,
-				},
-				{
-					Type:     watch.Deleted,
-					Previous: outRes1,
-				},
-			})
-			testWatcher1.Stop()
-
-			By("Starting a watcher from rev0 - this should get all events")
-			w, err = c.LiveMigrations().Watch(ctx, options.ListOptions{ResourceVersion: rev0})
-			Expect(err).NotTo(HaveOccurred())
-			testWatcher2 := testutils.NewTestResourceWatch(config.Spec.DatastoreType, w)
-			defer testWatcher2.Stop()
-
-			By("Modifying res2")
-			outRes3, err := c.LiveMigrations().Update(
-				ctx,
-				&internalapi.LiveMigration{
-					ObjectMeta: outRes2.ObjectMeta,
-					Spec:       spec1,
-				},
-				options.SetOptions{},
-			)
-			Expect(err).NotTo(HaveOccurred())
-			testWatcher2.ExpectEvents(internalapi.KindLiveMigration, []watch.Event{
-				{
-					Type:   watch.Added,
-					Object: outRes1,
-				},
-				{
-					Type:   watch.Added,
-					Object: outRes2,
-				},
-				{
-					Type:     watch.Deleted,
-					Previous: outRes1,
-				},
-				{
-					Type:     watch.Modified,
-					Previous: outRes2,
-					Object:   outRes3,
-				},
-			})
-			testWatcher2.Stop()
-
-			By("Starting a watcher from rev0 watching name1 - this should get all events for name1")
-			w, err = c.LiveMigrations().Watch(ctx, options.ListOptions{Namespace: namespace1, Name: name1, ResourceVersion: rev0})
-			Expect(err).NotTo(HaveOccurred())
-			testWatcher2_1 := testutils.NewTestResourceWatch(config.Spec.DatastoreType, w)
-			defer testWatcher2_1.Stop()
-			testWatcher2_1.ExpectEvents(internalapi.KindLiveMigration, []watch.Event{
-				{
-					Type:   watch.Added,
-					Object: outRes1,
-				},
-				{
-					Type:     watch.Deleted,
-					Previous: outRes1,
-				},
-			})
-			testWatcher2_1.Stop()
-
-			By("Starting a watcher not specifying a rev - expect the current snapshot")
-			w, err = c.LiveMigrations().Watch(ctx, options.ListOptions{})
-			Expect(err).NotTo(HaveOccurred())
-			testWatcher3 := testutils.NewTestResourceWatch(config.Spec.DatastoreType, w)
-			defer testWatcher3.Stop()
-			testWatcher3.ExpectEvents(internalapi.KindLiveMigration, []watch.Event{
-				{
-					Type:   watch.Added,
-					Object: outRes3,
-				},
-			})
-			testWatcher3.Stop()
-
-			By("Starting a watcher at rev0 in namespace1 - expect the events for LiveMigration in namespace1")
-			w, err = c.LiveMigrations().Watch(ctx, options.ListOptions{Namespace: namespace1, ResourceVersion: rev0})
-			Expect(err).NotTo(HaveOccurred())
-			testWatcher4 := testutils.NewTestResourceWatch(config.Spec.DatastoreType, w)
-			defer testWatcher4.Stop()
-			testWatcher4.ExpectEvents(internalapi.KindLiveMigration, []watch.Event{
-				{
-					Type:   watch.Added,
-					Object: outRes1,
-				},
-				{
-					Type:     watch.Deleted,
-					Previous: outRes1,
-				},
-			})
-			testWatcher4.Stop()
+			By("Starting a watcher from revision rev1 - should return an error")
+			_, err = c.LiveMigrations().Watch(ctx, options.ListOptions{ResourceVersion: rev1})
+			Expect(err).To(HaveOccurred())
 		})
 	})
 })


libcalico-go/lib/apis/internalapi/livemigration.go
```diff
@@ -0,0 +1,86 @@
...
+	KindLiveMigrationList = "LiveMigrationList"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// LiveMigration describes a live-migration operation in progress. It is
+// orchestrator-independent and holds the fields that Calico needs for
+// optimal live migration processing.
+type LiveMigration struct {
```
caseydavenport, line 34: We should be putting all of the v3 API objects into the `api/` package instead of here. If you take a look at the other `lib/apis/v3` objects, you'll see most of them simply import their spec from there.

libcalico-go/lib/apis/v3/livemigration.go
```diff
@@ -0,0 +1,86 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// LiveMigration describes a live-migration operation in progress. It is
+// orchestrator-independent and holds the fields that Calico needs for
+// optimal live migration processing.
+type LiveMigration struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	metav1.ObjectMeta `json:"metadata"`
+	// Specification of the LiveMigration.
```
caseydavenport:
```suggestion
```

libcalico-go/lib/apis/v3/livemigration.go
```diff
@@ -0,0 +1,86 @@
...
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// LiveMigration describes a live-migration operation in progress. It is
+// orchestrator-independent and holds the fields that Calico needs for
+// optimal live migration processing.
+type LiveMigration struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
```
caseydavenport:
```suggestion
```

libcalico-go/lib/apis/internalapi/livemigration.go
```diff
@@ -0,0 +1,86 @@
...
+
+// LiveMigrationSpec contains the specification for a LiveMigration resource.
+type LiveMigrationSpec struct {
+	// Name and namespace of the WorkloadEndpoint that this live
+	// migration operation is moving to.
+	DestinationWorkloadEndpoint types.NamespacedName `json:"destinationWorkloadEndpoint"`
+
+	// Selector for the WorkloadEndpoint that this live
+	// migration operation is moving to.
+	DestinationWorkloadEndpointSelector string `json:"destinationWorkloadEndpointSelector"`
```
caseydavenport: I do sort of wonder if we should be moving towards using native k8s selectors moving forward. They are easier to validate using CEL rules, and are more familiar to users.

They are however, less expressive than what our selector language can do. So depends on needs.

libcalico-go/lib/apis/internalapi/livemigration.go
```diff
@@ -0,0 +1,86 @@
...
+type LiveMigration struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	metav1.ObjectMeta `json:"metadata"`
+	// Specification of the LiveMigration.
+	Spec LiveMigrationSpec `json:"spec"`
+}
+
+// LiveMigrationSpec contains the specification for a LiveMigration resource.
+type LiveMigrationSpec struct {
```
caseydavenport, line 41: Are these all required fields? Any optional ones should be marked as such and be pointers.

caseydavenport (commented):
VERY possible I am jumping the gun and missing important context here, but thought I'd do a fast drive-by with some minor thoughts

libcalico-go/lib/apis/v3/livemigration.go
```diff
@@ -0,0 +1,86 @@
...
+	DestinationWorkloadEndpoint types.NamespacedName `json:"destinationWorkloadEndpoint"`
+
+	// Selector for the WorkloadEndpoint that this live
+	// migration operation is moving to.
+	DestinationWorkloadEndpointSelector string `json:"destinationWorkloadEndpointSelector"`
+
+	// Name and namespace of the WorkloadEndpoint that this live
+	// migration operation is moving from.
+	SourceWorkloadEndpoint types.NamespacedName `json:"sourceWorkloadEndpoint"`
+}
```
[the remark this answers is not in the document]
caseydavenport: This is wrong - the omitempty shouldn't be used on non-point struct fields as they are are ineffective. We should mark any optional fields as pointers with `// +optional`

libcalico-go/lib/apis/internalapi/livemigration.go
nelljerram: Does that apply even when this is intentionally an internal-only API?  In the senses that:
1. For KubeVirt it's backed by a KubeVirt resource (VirtualMachineInstanceMigration), analogously to how WorkloadEndpoint is backed by Pod.
2. For OpenStack our Calico driver will write it into the etcd datastore, based on Neutron DB state.
3. In both cases it is absolutely not intended to be a direct user API.

libcalico-go/lib/apis/v3/livemigration.go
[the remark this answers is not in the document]
nelljerram: Indeed, it was gopls that told me to remove the `omitempty` in these cases.

libcalico-go/lib/apis/internalapi/livemigration.go
caseydavenport: Yeah, that's an interesting one. I am trying to remove some of the mess that is our various API packages and representations as part of my v3 CRDs work. We have a few things:

- `api/pkg/projectcalico.org/v3` - the proper v3 API, intended for end-users. Used in our k8s API, as well as the `libcalico-go/clientv3` package.
- `libcalico-go/lib/api/crd.projectcalico.org/v1` - the structs used to generate the CRDs of the same name, internal use only. 
- `libcalico-go/lib/apis/v1` - I think this is actually largely unused at this point and we can probably remove most of it? It's legacy v1 API and is only used from other legacy v1 code IIUC.
- `libcalico-go/lib/apis/v3` - internal APIs used only by our code, often backed by k8s objects.

So, I think you are correct that this is in the right place. I am just getting turned around a little bit by naming!

I may submit a cleanup PR in this area, actually.

caseydavenport: > For KubeVirt it's backed by a KubeVirt resource (VirtualMachineInstanceMigration), analogously to how WorkloadEndpoint is backed by Pod.

I do wonder if for k8s clusters it just makes sense to read the VMI and not deal with all of this translation that we do? Just a passing thought - we could use the native type and avoid some munging ickyness (we'd still need our own type for OpenStack though)

libcalico-go/lib/apis/internalapi/livemigration.go
caseydavenport: If this isn't used for CRDs then we don't need the same level of API validation / optionality / etc., so can ignore these comments.

libcalico-go/lib/clientv3/client.go
```diff
@@ -189,6 +189,11 @@ func (c client) BlockAffinities() BlockAffinityInterface {
 	return blockAffinities{client: c}
 }
 
+// LiveMigrations returns an interface for managing LiveMigration resources.
+func (c client) LiveMigrations() LiveMigrationInterface {
+	return liveMigrations{client: c}
```
song-jiang, line 194: As discussed during standup, we should use upstream kubevirt client package to create a client. This is the kubevirt [library](https://github.com/song-jiang/calico/blob/ccd028114330da116c65d0faf919c51ca9187990/libcalico-go/lib/kubevirt/client.go) in my IPAM PR branch for your reference.

libcalico-go/lib/apis/internalapi/livemigration.go
nelljerram: > just makes sense to read the VMI ...

Assuming you meant VMIM here?  KubeVirt API has all of the possibilities :-)
- VM = VirtualMachine
- VMI = VirtualMachineInstance
- VMIM = VirtualMachineInstanceMigration

> we'd still need our own type for OpenStack though)

Yes, this is the key point for me.  I wanted everything in Felix to be as common as possible, so settled on LiveMigration as the common point for that.  The starting points were:

For OpenStack, we could have managed with adding fields to WorkloadEndpoint - since our driver is in complete control of those - and not having any separate resource like LiveMigration or VMIM.

For KubeVirt, KubeVirt is in control, and it uses VMIM, plus fields on VMI, to represent live migration; nothing special on pods, so nothing that we could flow through to WorkloadEndpoint.

Hence LiveMigration as the compromise between those.

(And the design has been [on display for a while at your local planning department in Alpha Centauri](https://www.goodreads.com/quotes/379100-there-s-no-point-in-acting-surprised-about-it-all-the) :grin:  OK, actually in Slack.)

WDYT?

libcalico-go/lib/apis/internalapi/livemigration.go
nelljerram: `SourceWorkloadEndpoint` is required.  The `Destination...` fields are alternatives, i.e. one of them must be non-zero.  OpenStack will use `DestinationWorkloadEndpoint`, KubeVirt will use `DestinationWorkloadEndpointSelector`.

Given that, can you advise how best to represent that, with pointers and/or optional markings?

libcalico-go/lib/clientv3/client.go
nelljerram: Thanks @song-jiang .  I've now pushed two more commits to make this change.  But I don't yet feel confident about the go.mod and go.sum changes that were needed - so I'll compare those against what you have in your branch.

nelljerram: You have very similar go.mod and go.sum changes in your PR - so that's reassuring to know.

nelljerram: By the way, I see that your PR is huge already - 103 files.  I wonder if there's any way of splitting that up into smaller stages, so that that review of each stage will be a bit more digestible?

libcalico-go/lib/apis/internalapi/livemigration.go
caseydavenport: Yes, sorry for rehashing design - let's stick with the agreed plan, it just got me thinking!

libcalico-go/lib/apis/internalapi/livemigration.go
caseydavenport: Yeah, I would use something like this in kubernetes APIs:

```go
// +optional
DestinationWorkloadEndpointSelector *string
```

Strictly speaking, I'd probably do this:

```go
Source *types.NamespacedName

Destination *WorkloadEndpointSelector
}


// +kubebuilder:validation:ExactlyOneOf
type WorkloadEndpointSelector {
  // +optional
  WorkloadEndpointReference *types.NamespacedName

  // +optional
  WorkloadEndpointSelector *string
}
```

That said, this is just an internal type - not sure if it matters as much here, but might be worth considering in case this ever does become something exposed in our proper API?

nelljerram: That looks nice, I'll apply that...

nelljerram: Hmm, I pushed a version of this in which `Source` is also `WorkloadEndpointIdentifier` - but now wonder if that's inviting trouble.
- Pro: it looks nicer for the API to be symmetric for Source and Destination.
- Neutral factor: it should be trivial to code the Felix calc graph to cope with either, for both Source and Destination.
- Con: but in fact we don't expect ever to need the Source.Selector case, and so it would be artificial to add testing to cover that, or else to leave a theoretical case uncovered.

WDYT?

caseydavenport: Yeah, definitely a risk of over generalizing. We could either leave the source as just a single field, or introduce a separate type for it if we want it to look symmetrical (i.e., that only has the reference field and not the selector)

nelljerram: I think the latter would still not be symmetric, so I should go back to what you proposed above.  (Except I think for "Identifier" instead of "Selector".)  Sorry for not already doing that.

nelljerram: I've done that now.  Would appreciate an opinion, if you have time today, on whether this is now close to approvable - in case I can get this done on Monday.

libcalico-go/lib/apis/internalapi/livemigration.go
nelljerram: Yes, definitely not CRDs.

libcalico-go/lib/clientv3/interface.go
```diff
@@ -41,6 +41,7 @@ type Interface interface {
 	CalicoNodeStatusClient
 	IPAMConfigurationClient
 	BlockAffinitiesClient
+	LiveMigrationsClient
```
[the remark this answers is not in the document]
nelljerram: True, but we regularly handle this kind of change in the Calico codebase.

charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
```diff
@@ -616,3 +616,13 @@ rules:
       - update
     resourceNames:
       - tigera-gateway-api-gateway-helm-certgen
+  # For monitoring KubeVirt live migration.  The operator does not do this itself, but needs to
+  # be able to assign this RBAC to Typha and calico-node.
+  - apiGroups:
```
fasaxc, line 621: New RBAC permissions should be release noted, I think.

apiserver/deps.txt
```diff
@@ -99,17 +104,21 @@ k8s.io/api v0.34.3
 k8s.io/apiextensions-apiserver v0.34.3
 k8s.io/apimachinery v0.34.3
 k8s.io/apiserver v0.34.3
-k8s.io/client-go v0.34.3
+k8s.io/client-go v12.0.0+incompatible
```
fasaxc: This looks wrong; bug in the deps calculation tool?  I'd expect this file to ppick up the version from the replace directive in go.mod, which is a v0.34.x version

libcalico-go/lib/backend/k8s/resources/livemigration.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func (c *LiveMigrationClient) Watch(ctx context.Context, list model.ListInterface, options api.WatchOptions) (api.WatchInterface, error) {
+	rlo := list.(model.ResourceListOptions)
+	k8sOpts := watchOptionsToK8sListOptions(options)
+	k8sWatch, err := c.vmimClient(rlo.Namespace).Watch(ctx, k8sOpts)
+	if err != nil {
+		return nil, K8sErrorToCalico(err, list)
+	}
+	adapted := newVMIMWatchAdapter(k8sWatch)
```
fasaxc: IIRC, Watch() gets called whenever we restart a watch (which happens fairly often).  I think that means that the `active` map that you're using to track state will get thrown away "mid flow", resulting in incorrect (lack of) events:
* Watch 1: Get an active object X, add to the cache.
* Watch terminates for some reason
* Watch restarted: cache reset
* X transitions to inactive, but we suppress the event because we don't think it's active.

Suggestion: just translate all events into Calico events withe the correct key, but present a `nil` value for the inactive VMIMs.  I believe that'll have the correct meaning downstream (and Typha will suppress a stream of `nil`s for the name key).

The only slight annoyance is the updatetype field, but that field was originally intended to represent the real state of the resource, even if we converted it to `nil` due to (say) a validation error.  I suspect there may be places where folks have got the wrong interpretation of that field so worth an audit.

manifests/operator-crds.yaml
```diff
@@ -1528,6 +1528,1199 @@ spec:
                           type: object
                       type: object
                   type: object
+                calicoWebhooksDeployment:
```
fasaxc, line 1531: Seems out of place in this PR?  Should someone else be merging this?

nelljerram: Yes, I think this is a case of operator code changing under our feet.  I'll raise a separate PR to bump this.

charts/tigera-operator/templates/tigera-operator/02-role-tigera-operator.yaml
nelljerram: That makes sense; I'll add a release note to this PR to say that.

apiserver/deps.txt
nelljerram: OK, I can take a look at that - OK to include in this PR?

(AFAIU, this weirdness is to do with the kubevirt golang client referencing a version of k8s.io/client-go that doesn't exist, and the only solution is to have this "incompatible" version in `require` and then `replace` with the real version.  I'm afraid I still don't understand go mod as well as would be ideal.  But anyway that is independent of the apparent issue in our deps calculation.)

libcalico-go/lib/backend/k8s/resources/livemigration.go
```diff
@@ -0,0 +1,286 @@
...
+		return false
+	}
+	return true
+}
+
+// convertVMIMToLiveMigrationSpec extracts the LiveMigrationSpec from a VMIM.
+// The caller must ensure the VMIM has the required fields (see vmimShouldEmitLiveMigration).
+func convertVMIMToLiveMigrationSpec(vmim *kubevirtv1.VirtualMachineInstanceMigration) internalapi.LiveMigrationSpec {
+	selector := fmt.Sprintf(
+		"kubevirt.io/vmi-name == '%s' && kubevirt.io/migrationJobUID == '%s'",
```
song-jiang: We should use a defined constant for those labels, e.g. https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/types.go#L1158

libcalico-go/lib/backend/k8s/resources/livemigration.go
```diff
@@ -0,0 +1,286 @@
...
+	return newK8sWatcherConverter(ctx, "VirtualMachineInstanceMigration", convertVMIMResourceToLiveMigration, adapted), nil
+}
+
+func (c *LiveMigrationClient) EnsureInitialized() error {
+	return nil
+}
+
+// vmimShouldEmitLiveMigration returns true if the VMIM is in a phase that
+// warrants emitting a LiveMigration resource and has the required fields set.
+func vmimShouldEmitLiveMigration(vmim *kubevirtv1.VirtualMachineInstanceMigration) bool {
```
song-jiang, line 145: Worth adding a comment 
```
Only proceed if the migration is actively preparing, running, or failing, and we have the VM Name, Source Pod, and Object UID firmly established.
```

libcalico-go/lib/backend/k8s/resources/livemigration.go
nelljerram: Thanks, I'll review this.

nelljerram: I'm aware that the VMIM resource gets lots of Status updates, without any net effect on the fields that we care about and convert to LiveMigration.  So my concern was that we will see a sequence of apparent KV updates for a LiveMigration resource - and all the consequent downstream CPU cost - but without anything actually changing.

libcalico-go/lib/backend/k8s/resources/livemigration_test.go
```diff
@@ -0,0 +1,409 @@
...
+
+			vmim := newVMIM("test-ns", "vmim-watch-1", "200", kubevirtv1.MigrationRunning, "vmi-w", "src-pod-w", "uid-w")
+			_, err = kvFake.KubevirtV1().VirtualMachineInstanceMigrations("test-ns").Create(ctx, vmim, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			wg.Wait()
+			w.Stop()
+		})
+
+		It("emits Deleted event when VMIM transitions to non-matching phase", func() {
```
song-jiang, line 262: I suggest adding a test case that simulates a VMIM progressing through its real-world phases (e.g., Pending -> Scheduling -> ... -> Succeeded) and verifies the corresponding watch events at each step.

manifests/operator-crds.yaml
nelljerram: Here is the separate PR for bumping operator CRDs:https://github.com/projectcalico/calico/pull/11903

apiserver/deps.txt
nelljerram: Addressing the deps generation in a separate PR #11909.  It also appears that the go.mod change to `v12.0.0+incompatible` was unnecessary, and I've reverted that [here](https://github.com/projectcalico/calico/pull/11868/commits/ffb0e1ade8125b585f5ccdeb5b4a7dbadeb9c7fe).

libcalico-go/lib/backend/k8s/resources/livemigration.go
nelljerram: Nice, thanks.

libcalico-go/lib/backend/k8s/resources/livemigration_test.go
nelljerram: Nice idea.

libcalico-go/lib/backend/k8s/resources/livemigration.go
nelljerram: Addressed now in https://github.com/projectcalico/calico/pull/11868/commits/06c169d4eba4acc13686a2ae5c07b8f5104b74cc.  Thanks for catching this.  (Claude did a less good job that I originally thought - but still my fault for letting that through.)

libcalico-go/lib/backend/k8s/resources/livemigration_test.go
nelljerram: Tests added in https://github.com/projectcalico/calico/pull/11868/commits/06c169d4eba4acc13686a2ae5c07b8f5104b74cc

libcalico-go/lib/clientv3/livemigration.go
```diff
@@ -0,0 +1,101 @@
...
+	res := &internalapi.LiveMigrationList{}
+	if err := r.client.resources.List(ctx, opts, internalapi.KindLiveMigration, internalapi.KindLiveMigrationList, res); err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+// Watch returns a watch.Interface that watches the LiveMigrations that match the
+// supplied options.
+func (r liveMigrations) Watch(ctx context.Context, opts options.ListOptions) (watch.Interface, error) {
```
fasaxc, line 100: As discussed, I suspect that Watch will panic on the nilled out backend events.  convertEvent calls kvPairToResource on the New field, which relies on there being a valid value there.

This is an awkward aspect of the k8s API's "internal metadata" approach, vs our "external metadata" approach.  We put the key in the KVPair wrapper, they put the key in the metadata inside the resource so nil is invalid.

nelljerram:
> LGTM. PR description seems out of date.

Thanks, I've updated the PR description now.  Will also squash when this eventually merges.

libcalico-go/lib/clientv3/livemigration.go
nelljerram: Added a firewall for this [here](https://github.com/projectcalico/calico/pull/11868/commits/2983c35a9f49fa83bebfe7b74fff44e17a82d8c8).  OK to approve now?

libcalico-go/lib/clientv3/livemigration.go
```diff
@@ -0,0 +1,106 @@
...
+	return nil, cerrors.ErrorOperationNotSupported{
+		Operation:  "Watch",
+		Identifier: internalapi.KindLiveMigration,
+		Reason:     "Watch is not supported for LiveMigration resources",
```
fasaxc, line 117: Worth a long-form comment explaining the incompatibility?  It's pretty subtle so the main value I'm thinking of here is documenting the problem that'd trip up v3 watches.

nelljerram: Sorry, of course.  Added [here](https://github.com/projectcalico/calico/pull/11868/commits/dbf04cc8d6fba7745b608a7a1018b0a2030f30c5).

PR #11870, opened 2026-02-18
Rename lib/v3 -> lib/internalapi

caseydavenport:
## Description

This is just to reduce some confusion with the public API, which is named v3.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/apis/internalapi/README.md
+++ b/libcalico-go/lib/apis/internalapi/README.md
@@ -2,13 +2,9 @@
 
 Internal Calico resource types that are not part of the public API (`api/pkg/apis/projectcalico/v3`).
 
-Contains types that need Kubernetes CRD storage but are either internal-only or carry additional
-fields beyond what the public API exposes:
+Contains types that are either stored in Kubernetes CRDs but not exposed to end users, and/or types that are used internally
+by Calico but are not backed by CRDs (i.e., WorkloadEndpoint which is stored in etcd mode, but backed by a Kubernetes Pod when using
+the Kubernetes API for storage).
 
-- **Node** / **WorkloadEndpoint** -- internal representations with fields like VXLAN tunnel
-  addresses, wireguard keys, QoS controls, and pod CIDRs that are not user-facing.
-- **IPAMBlock**, **IPAMHandle**, **IPAMConfig**, **BlockAffinity** -- IPAM implementation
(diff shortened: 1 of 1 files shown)


hack/Makefile
```diff
@@ -13,9 +13,6 @@ include ../lib.Makefile
 
 BINDIR?=bin
 
-# Create a list of files upon which the generated file depends, skip the generated file itself
-APIS_SRCS := $(filter-out ./lib/apis/v3/zz_generated.deepcopy.go, $(wildcard ./lib/apis/v3/*.go))
```
nelljerram, line 17: Is it intended to completely remove this?  What effect will that have on the targets below that depend on `$(APIS_SRCS)` (which will now be empty, I presume)?

libcalico-go/lib/apis/internalapi/README.md
```diff
@@ -0,0 +1,14 @@
+# libcalico-go/lib/apis/internalapi
+
+Internal Calico resource types that are not part of the public API (`api/pkg/apis/projectcalico/v3`).
+
+Contains types that need Kubernetes CRD storage but are either internal-only or carry additional
```
nelljerram: What do you mean by "that need Kubernetes CRD storage"?  In the WorkloadEndpoint case, my understanding is that that's only an in-memory concept, and that the only related storage is the related Pod.

nelljerram (commented):
In the etcd datastore, will the key for internalapi resources still be the same as it is now?  (Which I think is e.g. /calico/resources/v3/projectcalico.org/workloadendpoints/...)

Also two other queries below.

hack/Makefile
caseydavenport: this was actually unused, just a copy/paste error from the libcalico-go Makefile which I think was used as the original starting point for this file.

libcalico-go/lib/apis/internalapi/README.md
caseydavenport: Yes, this isn't quite correct - I will fix.

hack/Makefile
nelljerram: I'm sorry, I misread my own grep of the codebase.

PR #11875, opened 2026-02-18
Include ippools filters in BirdBGPConfig

mazdakn:
## Description

Move generating ippool filters from `IPPoolsFilterBIRDFunc` function to a client's method, named `processIPPools`, to include all those filters in the `BirdBGPConfig` struct. This improve generating BIRD configurations. Additionally, the next PR will introduce a new `BGPConfiguration` option to ignore all ippools, which is needed when Felix is responsible for all in-cluster routing.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/confd/pkg/backends/calico/bgp_processor.go
+++ b/confd/pkg/backends/calico/bgp_processor.go
@@ -750,7 +750,10 @@ func (c *client) processIPPools(config *types.BirdBGPConfig, ipVersion int) erro
 		filterActionForKernel = "reject"
 	}
 
-	localSubnet := c.localSubnet(ipVersion)
+	localSubnet, localSubnetErr := c.localSubnet(ipVersion)
+	if err != nil {
+		logCtx.WithError(err).Debug("Failed to get local host subnet")
+	}
 
 	for key, value := range kvPairs {
 		var ippool model.IPPool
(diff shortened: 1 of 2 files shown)


mazdakn:
@copilot add unit test for client's `processIPPools` method

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -716,3 +725,121 @@ func (c *client) processCommunityRules(config *types.BirdBGPConfig, ipVersion in
...
+	}
+	if len(comment) != 0 {
+		statement = fmt.Sprintf("%s %s", statement, formatComment(comment))
+	}
+	return
+}
+
+func formatComment(comment string) string {
+	return fmt.Sprintf("# %s", comment)
+}
```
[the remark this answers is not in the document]
mazdakn: @copilot add unit test for client's processIPPools method

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -716,3 +730,187 @@ func (c *client) processCommunityRules(config *types.BirdBGPConfig, ipVersion in
...
+	}
+
+	// Filter statements based on provided filterAction.
+	if len(filterAction) != 0 && filterAction != action {
+		return ""
+	}
+	return emitFilterStatementForIPPools(cidr, extraStatement, action, comment)
+}
+
+func (c *client) localSubnet(ipVersion int) string {
```
nelljerram: I'd rather return an error here in the error case, than an empty string, and then handle that in `processIPPools` before the pool loop, so that the `processIPPool` doesn't have to consider the case of `localSubnet` being empty.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -716,3 +730,187 @@ func (c *client) processCommunityRules(config *types.BirdBGPConfig, ipVersion in
...
+		return nil
+	}
+
+	// In IPv6, we only need to include statements with "reject" action, since we include a default "accept" at the end.
+	var filterActionForKernel string
+	if ipVersion == 6 {
+		filterActionForKernel = "reject"
+	}
+
+	localSubnet := c.localSubnet(ipVersion)
```
nelljerram: I think should return error here if `c.localSubnet` indicates an error.  Then the following loop doesn't need to consider `localSubnet == ""`.

confd/pkg/backends/calico/bgp_processor.go
mazdakn: I agree that returning error is a better approach. Fix it.

confd/pkg/backends/calico/bgp_processor.go
mazdakn: The issue is that when `localSubnet == "" `:
-  no filter for kernel should be generated.
- but we still need filters for bgp exports.

So anyway we need to somewhere check the value of `localSubnet`.

PR #11880, opened 2026-02-19
Bring Istio to Calico OSS

radixo:
## Description

This PR prepares Felix for supporting Istio on Calico OSS, aligning its behavior with the existing implementation in Calico Enterprise. It enables support for Istio ambient dataplane mode while preserving the original destination ports used to reach services, instead of redirecting traffic to port 15008 (the default for Istio ambient mode). This ensures that microsegmentation enforced by Network Policies is preserved.

It also introduces the EndpointComputedData implementation in the calculation graph layer. This enables efficient, non-persistent endpoint data to be computed and transmitted directly to the dataplane, reducing overhead while maintaining feature parity.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Introduces istioAmbientMode and istioDSCPMark to felix for integration with Calico Istio compilation.
```

## Reminder for the reviewer

Other optional labels:


felix/bpf-gpl/tc.c
```diff
@@ -1364,9 +1364,31 @@ int calico_tc_skb_accepted_entrypoint(struct __sk_buff *skb)
...
+		sip.addr = src_ip;
+
+		// Is the src part of the mesh?
+		if (cali_ip_sets_lookup_elem(&sip)) {
+			if (!qos_dscp_set(ctx, ISTIO_DSCP)) {
+				goto deny;
+			}
+			CALI_DEBUG("IP DSCP: set for Mesh");
+		}
+	}
```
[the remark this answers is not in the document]
radixo: This happens because the ISTIO_DSCP is set to -1 on eBPF programs that are attached to pods that are not part of the mesh.

api/pkg/apis/projectcalico/v3/constants.go
```diff
@@ -63,4 +63,9 @@ const (
 
 	// Host name used for Service LoadBalancer
 	VirtualLoadBalancer = "load-balancer"
+
+	// Tigera Istio Labels
```
caseydavenport:
```suggestion
	// Istio Labels
```

felix/calc/calc_graph.go
```diff
@@ -378,6 +378,10 @@ func NewCalculationGraph(
 	polResolver.RegisterCallback(callbacks)
 	cg.policyResolver = polResolver
 
+	if conf.IsIstioAmbientModeEnabled() {
+		_ = NewIstioCalculator(activeRulesCalc, ruleScanner, callbacks, ipsetMemberIndex, polResolver.OnEndpointComputedDataUpdate)
```
caseydavenport, line 382: Should we be doing something with this return value? Is it an error?

radixo: No, it's just the instance of the IstioCalculator, but everything is sorted on the constructor.

felix/bpf/ipsets/ipsets.go
```diff
@@ -35,6 +35,14 @@ var (
 	})
 )
 
+const (
+	// Base of range for reserved IP set IDs for Calico Enterprise.
+	ReservedIPSetBase uint64 = 1 << 32
+
+	// IP set ID for All Istio WEPs.
+	AllIstioWEPsID = ReservedIPSetBase + 1
+)
```
caseydavenport, line 44: Heads up — in calico-private, `+1` and `+2` are already taken by `TrustedDNSServersID` and `EgressGWHealthPortsID` (enterprise-only), so `AllIstioWEPsID` is `ReservedIPSetBase + 3` there. This is correct for OSS, just flagging that the cherry-pick to private will need to keep the existing `+3` offset.

felix/bpf-gpl/policy.h
```diff
@@ -5,6 +5,12 @@
 #ifndef __CALI_POLICY_H__
 #define __CALI_POLICY_H__
 
+// Base of range for reserved IP set IDs for Calico Enterprise.
+#define RESERVED_IP_SET_BASE (((__u64)1)<<32)
+
+// IP set ID for All Istio WEPs.  Must match the ID defined in ipsets.go.
+#define ALL_ISTIO_WEPS_ID (RESERVED_IP_SET_BASE + 1)
```
caseydavenport: Same note as `ipsets.go` — in private this is `RESERVED_IP_SET_BASE + 3` since `+1` and `+2` are enterprise-only IDs.

PR #11894, opened 2026-02-20
Refactor IPAM AllocationAttribute for VM-based handle support

song-jiang:
## Description

Rename AllocationAttribute fields, update GetAssignmentAttributes return type, and consolidate felix/hashutils into lib/hash. 

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/libcalico-go/lib/ipam/ipam_attributes.go
+++ b/libcalico-go/lib/ipam/ipam_attributes.go
@@ -45,24 +45,5 @@ func (c ipamClient) GetAssignmentAttributes(ctx context.Context, addr cnet.IP) (
 		return nil, err
 	}
 	block := allocationBlock{obj.Value.(*model.AllocationBlock)}
-
-	activeAttrs, err := block.attributesForIP(addr, OwnerAttributeTypeActive)
-	if err != nil {
-		return nil, err
-	}
-	alternateAttrs, err := block.attributesForIP(addr, OwnerAttributeTypeAlternate)
-	if err != nil {
-		return nil, err
-	}
-
-	handle, err := block.handleForIP(addr)
-	if err != nil {
-		return nil, err
-	}
-
-	return &model.AllocationAttribute{
-		HandleID:            handle,
-		ActiveOwnerAttrs:    activeAttrs,
-		AlternateOwnerAttrs: alternateAttrs,
-	}, nil
+	return block.allocationAttributesForIP(addr)
 }
--- a/libcalico-go/lib/ipam/ipam_block.go
+++ b/libcalico-go/lib/ipam/ipam_block.go
@@ -499,7 +499,7 @@ func (b allocationBlock) ipsByHandle(handleID string) []cnet.IP {
 	return ips
 }
 
-func (b allocationBlock) attributesForIP(ip cnet.IP, attrType OwnerAttributeType) (map[string]string, error) {
+func (b allocationBlock) attributesForIP(ip cnet.IP) (map[string]string, error) {
 	// Convert to an ordinal.
 	ordinal, err := b.IPToOrdinal(ip)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


libcalico-go/lib/hash/unique_id.go
```diff
@@ -34,3 +39,37 @@ func MakeUniqueID(prefix, content string) string {
...
 	return prefix + ":" + base64.RawURLEncoding.EncodeToString(hashBytes)
 }
+
+// GetLengthLimitedID returns an ID that consists of the given prefix and, either the given suffix,
+// or, if that would exceed the length limit, a cryptographic hash of the suffix, truncated to the
+// required length.
+// If the combined prefix+suffix fits within maxLength, it returns the original string unchanged.
+// If it's too long, it hashes the suffix using SHA256, base64-encodes it, and truncates to fit.
+// Uses shortenedPrefix ("_") to indicate when a hash was used, to avoid collisions.
+func GetLengthLimitedID(fixedPrefix, suffix string, maxLength int) string {
```
fasaxc, line 49: Note that we won't be able to change this algorithm after you use it for keys.  We already had to be careful because a change would result in felix IP set name churn

libcalico-go/lib/ipam/ipam_attributes.go
```diff
@@ -0,0 +1,68 @@
...
+		log.Errorf("Error reading pool for %s", addr.String())
+		return nil, cerrors.ErrorResourceDoesNotExist{Identifier: addr.String(), Err: errors.New("No valid IPPool")}
+	}
+	blockCIDR := getBlockCIDRForAddress(addr, pool)
+	obj, err := c.blockReaderWriter.queryBlock(ctx, blockCIDR, "")
+	if err != nil {
+		log.Errorf("Error reading block %s: %v", blockCIDR, err)
+		return nil, err
+	}
+	block := allocationBlock{obj.Value.(*model.AllocationBlock)}
```
fasaxc, line 47: Can we share block lookup logic with Release?  Surely that needs to do the same?

libcalico-go/lib/ipam/ipam_block.go
```diff
@@ -499,7 +499,7 @@ func (b allocationBlock) ipsByHandle(handleID string) []cnet.IP {
 	return ips
 }
 
-func (b allocationBlock) attributesForIP(ip cnet.IP) (map[string]string, error) {
+func (b allocationBlock) attributesForIP(ip cnet.IP, attrType OwnerAttributeType) (map[string]string, error) {
```
fasaxc: Think I might have introduced a second function that returns both and then keep this function returning the primary attribute.  Then your new logic can retrieve both in one call, no enum switch required

libcalico-go/lib/hash/unique_id.go
song-jiang: That's a good point. The core algorithm is identical, the only differences are:

- Empty suffix guard (new): if len(suffix) == 0 { suffix = shortenedPrefix } , this prevent panic if caller passes an empty suffix, which the original code never hit. 

- charsLeftForHash panic guard (new): if charsLeftForHash <= 0 { log.Panicf(...) } — this only fires for invalid inputs that would have panicked anyway (slice out of bounds).

I think the above improvements shouldn't affect existing use case.

libcalico-go/lib/ipam/ipam_attributes.go
song-jiang: `ReleaseIPs` has different steps in block lookup. 
- It handles the case where an IP isn't in any pool by doing a full block scan (getBlockForIP).
- It groups IPs by block to minimize datastore writes.

`GetAssignmentAttributes` has much simpler logic. I don't see there's much benefit of sharing the code.

PR #11899, opened 2026-02-20
[CORE-12378] fix(QoS): Use QdiscReplace() instead of QdiscAdd()

coutinhop:
Use QdiscReplace() instead of QdiscAdd() so that adding the TBF qdiscs needed for QoS controls with tc does not error out when there is an existing non-default (handle != 0) qdisc on the interface for any reason.

Add a test case to the felix FVs to cover this.

Also, enable felix debug logging on the QoS felix FVs, and remove overzealous Skip() that was resulting in no test cases running on iptables/nftables modes.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Fix failure to enable ingress bandwidth QoS controls when a non-default qdisc previously existed on the workload interface (handle != 0).
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/dataplane/linux/qos/qos.go
+++ b/felix/dataplane/linux/qos/qos.go
@@ -242,20 +242,12 @@ func TeardownIfb(deviceName string) error {
 	return nil
 }
 
-func CreateIngressQdisc(tbs *TokenBucketState, workloadDeviceName string) error {
+func CreateOrUpdateIngressQdisc(tbs *TokenBucketState, workloadDeviceName string) error {
 	workloadDevice, err := netlink.LinkByName(workloadDeviceName)
 	if err != nil {
 		return fmt.Errorf("get host device %s: %w", workloadDeviceName, err)
 	}
-	return createTBF(tbs, workloadDevice)
-}
-
-func UpdateIngressQdisc(tbs *TokenBucketState, workloadDeviceName string) error {
-	workloadDevice, err := netlink.LinkByName(workloadDeviceName)
-	if err != nil {
-		return fmt.Errorf("get host device %s: %w", workloadDeviceName, err)
-	}
-	return updateTBF(tbs, workloadDevice)
+	return createOrUpdateTBF(tbs, workloadDevice)
 }
 
 func CreateEgressQdisc(tbs *TokenBucketState, workloadDeviceName string, ifbDeviceName string) error {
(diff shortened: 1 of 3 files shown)


felix/fv/qos_controls_test.go
```diff
@@ -230,6 +226,7 @@ var _ = infrastructure.DatastoreDescribe(
 
 					topt.DelayFelixStart = true
 					topt.TriggerDelayedFelixStart = true
+					topt.FelixLogSeverity = "Debug"
```
nelljerram, line 229: Do you mean to check this change in?

felix/fv/qos_controls_test.go
```diff
@@ -320,7 +317,15 @@ var _ = infrastructure.DatastoreDescribe(
 						if BPFMode() && BPFAttachType() == "tc" {
 							Skip("Skipping QoS control bandwidth tests on BPF TC attach mode.")
 						}
+
+						By("Removing all limits from workloads")
+						for i := range len(w) {
+							w[i].WorkloadEndpoint.Spec.QoSControls = nil
+							w[i].UpdateInInfra(infra)
+							Eventually(tc.Felixes[i].ExecOutputFn("ip", "r", "get", fmt.Sprintf("10.65.%d.2", i)), "10s").Should(ContainSubstring(w[i].InterfaceName))
```
nelljerram, line 325: Is this check effective?  I presume the route was already there even with the QoS in place?

felix/dataplane/linux/qos/qos.go
nelljerram: createTBF and updateTBF are now identical.  Can we deduplicate them?

coutinhop: thanks for catching it! done

felix/fv/qos_controls_test.go
coutinhop: I believe I arrived at this because the route churns when the workload endpoint is updated (if my memory doesn't fail me), so I added a wait until it shows up again...

felix/fv/qos_controls_test.go
coutinhop: yes, absolutely! the issue we're fixing only shows up in the debug logs, so I think it makes sense to have them in this test

nelljerram: I guess it's OK to leave this in.  But you only mean for debugging the problem, right?  I.e. it is not the case that there was only an issue when log level was debug.

felix/fv/qos_controls_test.go
nelljerram: Hmm, that sounds surprising.  I would expect TC updates to be independent of route programming, and that a TC update would not require a route change.  Might be a real problem hiding here if you really saw route churn.

coutinhop: I don't think it's related to tc, but the FV test infra, or maybe I was too zealous when writing those into every workload update, I'll merge this PR and investigate separately

PR #11903, opened 2026-02-23
Operator CRD update (make generate)

nelljerram:


nelljerram:
Corresponding PR in calico-private master is https://github.com/tigera/calico-private/pull/10873

PR #11904, opened 2026-02-23
Enhance Felix route table for elevated priority programming

nelljerram:
JIRA ticket: CORE-12272

On its own, this PR does not yet change how Felix programs local routes.  `SetRoutes` callers do not currently set the new `Priority` field, so programmed `Priority` will be 0 for IPv4 routes and 1024 for IPv6 routes, as was the case before this PR.

Upcoming, but separate, live migration work will:
- add Felix configuration fields for "normal" and "elevated" priority values
- change route programming outside of live migration to use the "normal" priority value
- program routes with the "elevated" priority value during a live migration, as part of ensuring the best possible handover

### DIFF
--- a/felix/routetable/route_table.go
+++ b/felix/routetable/route_table.go
@@ -347,60 +347,39 @@ func New(
 		handlemgr.WithSocketTimeout(netlinkTimeout),
 	)
 
-	// Interaction between conntrack cleanup and multiple route priorities
-	// ================================================================
+	// The conntrack tracker (ConntrackCleanupManager) cleans up conntrack state for a workload
+	// IP - i.e. for flows going to or from that IP, on this node - when Felix stops programming
+	// a local route for that workload.  The latter happens during live migration when the VM
+	// was originally on this node: a remote BIRD-programmed route appears for the IP, then
+	// Felix removes the old local route for the IP, which means that conntrack state for that
+	// IP is cleaned up; yet we want existing connections to the migrating IP to be impacted as
+	// little as possible.
 	//
-	// The conntrack tracker (ConntrackCleanupManager) tracks ownership of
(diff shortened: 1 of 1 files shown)


felix/routetable/route_table.go
```diff
@@ -459,15 +513,16 @@ func (r *RouteTable) SetRoutes(routeClass RouteClass, ifaceName string, targets
...
 	}
 
 	// Figure out what has changed.
 	oldTargetsToCleanUp := r.ifaceToRoutes[routeClass][ifaceName]
-	newTargets := map[ip.CIDR]Target{}
+	newTargets := map[RouteKey]Target{}
 	for _, t := range targets {
-		delete(oldTargetsToCleanUp, t.CIDR)
-		newTargets[t.CIDR] = t
+		routeKey := r.normalizeRouteKey(t.RouteKey)
```
fasaxc, line 502: Just pondering if there's a way to _force_ normalization to avoid forgetting to do it on some new code path in future.  It was forced before because we only stored CIDR and then the conversion to kernel route did the trick.

We _could_ index the internal fields with a private `normalizedRouteKey` type, which is the same as `RouteKey` but the constructor normalizes.  Probably overkill!

felix/routetable/route_table.go
```diff
@@ -347,6 +347,60 @@ func New(
 		handlemgr.WithSocketTimeout(netlinkTimeout),
 	)
 
+	// Interaction between conntrack cleanup and multiple route priorities
```
fasaxc: Comment is welcome, but maybe a bit big?

nelljerram: Updated now, PTAL when you have time.

nelljerram: I'm not sure I made it a lot shorter; but definitely more relevant I hope.

PR #11905, opened 2026-02-23
Trim PR template and simplify cherry-pick headings

fasaxc:
Simplify the PR template to just a description comment and the release-note
block. The verbose sections (Description, Related issues/PRs, Todos, Reminder
for the reviewer) were rarely filled in properly and added clutter.

Update cherry-pick-pull to use bold text instead of markdown headings for
cherry-pick history and multi-PR separators. Improve re-pick handling to
absorb old bullet points (both old `##` and new `**` formats) into the header
so picks stack correctly with newest at the top.

**Release note:**
```release-note
TBD
```

### DIFF
--- a/hack/cherry-pick-pull
+++ b/hack/cherry-pick-pull
@@ -178,8 +178,9 @@ trap return_to_kansas EXIT
 #   rel               - release branch basename (e.g. "release-v3.14")
 #   PULL_TITLES[]     - array of PR titles
 #   PULL_BODIES[]     - array of PR bodies
-#   PULL_LABELS[]     - array of PR labels (newline-separated per element)
-#   PULLLINK[]        - array of markdown links (e.g. "org/repo#123")
+#   PULL_LABELS[]     - array of PR labels (one element per PR, labels
+#                       within each element separated by newlines)
+#   PULLLINK[]        - array of PR links (e.g. "org/repo#123")
 #   SRC_MAIN_REPO_ORG, SRC_MAIN_REPO_NAME
 #   DST_MAIN_REPO_ORG, DST_MAIN_REPO_NAME
 #   EXTRA_LABELS      - (optional) extra comma-separated labels to append


hack/cherry-pick-pull
```diff
@@ -170,31 +170,38 @@ function return_to_kansas {
...
-  # Build up the PR text.
+# Pure text-processing function: builds the PR title, body, and labels from
+# pre-fetched PR metadata.  Makes NO gh/git calls so it can be unit-tested
+# with canned input.
+#
+# Inputs (must be set before calling):
+#   rel               - release branch basename (e.g. "release-v3.14")
+#   PULL_TITLES[]     - array of PR titles
+#   PULL_BODIES[]     - array of PR bodies
+#   PULL_LABELS[]     - array of PR labels (newline-separated per element)
```
nelljerram: I'm not sure what "newline-separated per element" means.  Can you add an example here?

hack/cherry-pick-pull
```diff
@@ -170,31 +170,38 @@ function return_to_kansas {
...
+#   PULLLINK[]        - array of markdown links (e.g. "org/repo#123")
```
nelljerram: Perhaps I'm asking too many questions, but what are these doing?  I'd expect links to be inline in `PULL_BODIES`, like `[text](link)`.

nelljerram: Oh no worries, I see, these are links to the PRs themselves, right?  In that case "PR links" would be much clearer than "markdown links".

hack/cherry-pick-pull
nelljerram: Or maybe just point to [here](https://github.com/projectcalico/calico/pull/11905/changes#diff-4e0a8211d0b38f4360cb2f9b2053c7bc5de865bdb0b88fca638a73873669b3a3R60)

PR #11906, opened 2026-02-23
Fix IPAM GC hanging on rapid node cleanup

caseydavenport:
## Description

Bug fix for the IPAM GC controller getting stuck during rapid node scale-down.

When ReleaseHostAffinities fails (e.g., because a block still has tunnel IP allocations that haven't been GC'd yet), syncIPAM() would return an error, which prevented syncComplete() from ever running. Without syncComplete(), dirty nodes accumulate, each subsequent sync processes more and more nodes, contention increases, and the controller effectively freezes.

The fix:
- Change releaseNodes() to return failed nodes instead of an error, so that node cleanup failures are non-fatal
- Always call syncComplete(), ensuring incremental progress on every sync pass
- Re-dirty failed nodes after syncComplete() so they get retried on the next pass

This is a simpler alternative to the approach in #10333, which improved throughput but didn't address the core ordering issue between tunnel IP GC and node affinity release.

Related issues/PRs

Fixes https://github.com/projectcalico/calico/issues/8643

Related: https://github.com/projectcalico/calico/pull/10333

```release-note
Fix calico-kube-controllers IPAM GC controller getting stuck when cleaning up nodes during rapid scale-down.
```

### DIFF
--- a/kube-controllers/pkg/controllers/node/ipam_allocation.go
+++ b/kube-controllers/pkg/controllers/node/ipam_allocation.go
@@ -297,10 +297,7 @@ func (t *allocationState) markDirty(node string, reason string) {
 }
 
 // markClean removes a single node from the dirty set after it has been
-// successfully processed. This replaces the old bulk syncComplete() approach,
-// ensuring that only nodes we actually handled get cleared. Nodes that were
-// never processed (e.g., due to an early error) or that failed cleanup remain
-// dirty and will be retried on the next sync.
+// successfully processed.
 func (t *allocationState) markClean(node string, reason string) {
 	if _, ok := t.dirtyNodes[node]; ok {
 		log.WithFields(log.Fields{"node": node, "reason": reason}).Debug("Node is no longer dirty")


kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1087,18 +1087,25 @@ func (c *IPAMController) syncIPAM() error {
...
-		kick(c.syncChan)
+	// Re-dirty any nodes whose cleanup failed so they are retried on the next sync pass.
+	// This must happen AFTER syncComplete() so the dirty flags aren't immediately cleared.
+	for _, node := range failedNodes {
+		c.allocationState.markDirty(node, "node cleanup failed, will retry")
+	}
+
+	// If there is remaining work, return an error so the retry controller schedules
+	// the next sync with exponential backoff, avoiding tight-loop retries.
+	if len(failedNodes) > 0 || len(c.confirmedLeaks) > 0 {
```
skoryk-oleksandr, line 1105: I might be wrong here, but I think this changes the retry behavior for remaining confirmed leaks. The original code used `kick(c.syncChan)`
  which triggered an immediate next pass, while returning an error here goes through retryController.ScheduleRetry() with exponential backoff.
  If there are e.g. **50k** leaked IPs being processed in **5** batches of **10k** each, would this add unnecessary delays between batches?

  If so, might be worth splitting the two cases:
```
  if len(failedNodes) > 0 {
      return fmt.Errorf("node cleanup failed for %d nodes", len(failedNodes))
  }
  if len(c.confirmedLeaks) > 0 {
      log.WithField("num", len(c.confirmedLeaks)).Info("Confirmed leaks still exist, scheduling another pass")
      kick(c.syncChan)
  }
  return nil
```

  But maybe the backoff is intentional here, curious what you think.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1087,18 +1087,25 @@ func (c *IPAMController) syncIPAM() error {
...
 	// nodes are no longer in the Kubernetes API, and have no valid allocations, so can be cleaned up entirely
-	// from Calico IPAM.
-	if err = c.releaseNodes(nodesToRelease); err != nil {
-		return err
-	}
+	// from Calico IPAM. Node cleanup failures are non-fatal — failed nodes are re-dirtied below
+	// so they are retried on the next sync. This ensures syncComplete() always runs, preventing
+	// the controller from getting stuck when cleanup fails (e.g., due to "block not empty" errors
+	// from concurrent IPAM operations).
+	failedNodes := c.releaseNodes(nodesToRelease)
```
skoryk-oleksandr, line 1100: Looks like `checkAllocations()`, `garbageCollectKnownLeaks()`, and `releaseUnusedBlocks()`
  above still have `if err != nil { return err }` that skip `syncComplete()` the same way. I don't think those paths hit the same ordering issue in
  practice, but would it be worth using a `defer` for `syncComplete()` to guarantee it always runs regardless of which step fails?

kube-controllers/pkg/controllers/node/ipam.go
caseydavenport: Yeah, I thought putting that through backoff would be OK and in many cases desirable anyway. A small delay should be fine, and I think it's safer to be confident we never have a tight loop here?

kube-controllers/pkg/controllers/node/ipam.go
caseydavenport: I think a defer for _just_ syncComplete() is wrong - we want to run some logic after syncComplete finishes, but perhaps I can refactor this a bit more to make it cleaner. Let me think about this.

kube-controllers/pkg/controllers/node/ipam_test.go
```diff
@@ -1851,6 +1851,171 @@ var _ = Describe("IPAM controller UTs", func() {
...
+	// The fix changes releaseNodes() to return failed nodes instead of an error,
+	// ensuring syncComplete() always runs. Failed nodes are re-dirtied after
+	// syncComplete() so they are retried on the next pass.
+	//
+	// This test creates three deleted nodes, injects cleanup errors for two of them,
+	// and verifies that:
+	//   - The healthy node is cleaned up immediately
+	//   - The failing nodes are retained and retried
+	//   - As errors are cleared, each node is eventually cleaned up
+	It("should not get stuck when node cleanup fails (issue #8643)", func() {
```
skoryk-oleksandr, line 1926: Right now 1 node always succeeds. Would be good to also test the case where nothing makes progress in a 
  pass, all nodes fail and all get re-dirtied.

kube-controllers/pkg/controllers/node/ipam_test.go
```diff
@@ -1851,6 +1851,171 @@ var _ = Describe("IPAM controller UTs", func() {
...
+	//
+	// The fix changes releaseNodes() to return failed nodes instead of an error,
+	// ensuring syncComplete() always runs. Failed nodes are re-dirtied after
+	// syncComplete() so they are retried on the next pass.
+	//
+	// This test creates three deleted nodes, injects cleanup errors for two of them,
+	// and verifies that:
+	//   - The healthy node is cleaned up immediately
+	//   - The failing nodes are retained and retried
+	//   - As errors are cleared, each node is eventually cleaned up
```
skoryk-oleksandr, line 1925: Does it make sense to also verify that **dirty nodes don't accumulate across syncs**? e.g., pause the controller after a few failed sync passes and assert that `len(c.allocationState.dirtyNodes)` stays equal to the number of failing nodes, rather than growing with each pass. That would directly validate the core fix.

caseydavenport:
@skoryk-oleksandr thanks for the review! Just pushed a commit that addresses your comments. 

I refactored the dirty / clean logic a bit to avoid the need for a defer. Instead, We mark nodes as clean as we iterate them, to avoid the need to do an explicit cleanup at the end of the function.

PR #11908, opened 2026-02-23
Fix persistent connection teardown race in BPF spoof test

tomastigera:
When Stop() is called on a PersistentConnection, it creates a loop file to signal test-connection to exit, then waits for the process. However, the tryLoopFile loop checks the loop file only in ls.Next(), which comes AFTER the Receive() call. If the TCP connection gets reset (e.g. due to Felix reprogramming BPF state after a WEP reconfiguration), Receive() returns a non-timeout error and hits log.Fatal before ever checking the loop file. This causes test-connection to exit with code 1, which makes Stop() fail the test even though all assertions passed.

Fix: in the Receive() error handler, check if the loop file exists (meaning we were asked to stop). If so, exit cleanly instead of fatally.

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Other optional labels:

### DIFF
--- a/felix/fv/test-connection/test-connection.go
+++ b/felix/fv/test-connection/test-connection.go
@@ -488,9 +488,12 @@ func (tc *testConn) tryLoopFile(loopFile string, logPongs bool, timeout, sleep t
 			continue
 		} else {
 			fmt.Printf("err = %+v\n", err)
-			// If the loop file exists, we were asked to stop. The connection
-			// error is expected during shutdown, so exit cleanly.
-			if ls.loopFile != "" {
+			// If the initial exchange has completed and the loop file exists,
+			// we were asked to stop. The connection error is expected during
+			// shutdown, so exit cleanly. We only check after sentInitial so
+			// that startup failures still fail loudly (the loop file is
+			// expected to exist before the first successful exchange).
+			if ls.sentInitial && ls.loopFile != "" {
 				if _, statErr := os.Stat(ls.loopFile); statErr == nil {
 					log.WithError(err).Info("Connection error during shutdown, exiting cleanly")
 					if rmErr := os.Remove(ls.loopFile); rmErr != nil {


felix/fv/fv_suite_test.go
```diff
@@ -131,16 +132,21 @@ var _ = BeforeSuite(func() {
...
+		return err
 	}
 	totalBatches, err := strconv.Atoi(os.Getenv("FV_NUM_BATCHES"))
-	if err != nil || totalBatches < 1 {
-		return fmt.Errorf("invalid FV_NUM_BATCHES value: %s", os.Getenv("FV_NUM_BATCHES"))
+	if err != nil {
+		return err
+	}
+
+	if totalBatches < 1 || currentBatch <= 0 {
```
sridhartigera: I think Jiawei already merged a fix for this.

tomastigera: ok, it will get resolved when merging :+1:

felix/fv/test-connection/test-connection.go
```diff
@@ -488,6 +488,15 @@ func (tc *testConn) tryLoopFile(loopFile string, logPongs bool, timeout, sleep t
 			continue
 		} else {
 			fmt.Printf("err = %+v\n", err)
+			// If the loop file exists, we were asked to stop. The connection
+			// error is expected during shutdown, so exit cleanly.
+			if ls.loopFile != "" {
+				if _, statErr := os.Stat(ls.loopFile); statErr == nil {
+					log.WithError(err).Info("Connection error during shutdown, exiting cleanly")
```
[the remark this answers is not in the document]
tomastigera: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/projectcalico/calico/pull/11908#discussion_r2843606038)

sridhartigera (approved):

PR #11911, opened 2026-02-24
Add support for SourceAddr when programming kernel routes for Pod Networks

Cellebyte:
## Observations
When running a Kubernetes node with a single interface and adding an IP to that interface which is lower than the NodeIP, the kernel takes the lowest IP for source address selection.

There is an option to define the `deviceSourceAddress` and `deviceSourceAddressIPv6` in Felix for IPv4 and IPv6.
This is already configured and adds the desired `src` on the directly attached calico interface routes.

This option only applies for local node traffic so, local Node -> local Pod traffic, but not Node -> Node -> Pod traffic.

The IPv6 addresses are shortened I use correct IPv6 addresses on the setup.

When you use e.g. kube-proxy with masquerade for the SNAT in any mode, it uses the kernel selection mechanism to determine the source address for the masqueraded packet. Due to the missing `src` it could use the wrong IP of the Interface.

```console
$ ip a
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc prio state UP group default qlen 1000
    link/ether 00:50:56:b9:38:c4 brd ff:ff:ff:ff:ff:ff
    altname enp11s0
    inet 10.0.0.4/27 brd 10.0.0.2 scope global eth0  # node-ip
       valid_lft forever preferred_lft forever
    inet 10.0.0.3/32 scope global eth0                      # other ip
       valid_lft forever preferred_lft forever
    inet6 ::2/128 scope global                                   # other-ip
       valid_lft forever preferred_lft forever
    inet6 ::3/64 scope global                                     # node-ip
       valid_lft forever preferred_lft forever

$ ip -6 r s
::/64 dev eth0 proto kernel metric 256 pref medium
fd00:172:30:31d2:32b3:6b1:5d3f:c300/122 via ::4 dev eth0 proto bird metric 1024 pref medium
fd00:172:30:452d:f80d:7071:f347:3580/122 via ::7 dev eth0 proto bird metric 1024 pref medium
fd00:172:30:4dd1:b828:2dab:2cf7:d340/122 via ::5 dev eth0 proto bird metric 1024 pref medium
fd00:172:30:8ed8:7d36:d84:30d9:2880 dev calief6f80a244b src ::3 metric 1024 pref medium
fd00:172:30:8ed8:7d36:d84:30d9:2881 dev calibb1f3303ca0 src ::3 metric 1024 pref medium
fd00:172:30:8ed8:7d36:d84:30d9:2882 dev cali6e113f50928 src ::3 metric 1024 pref medium
blackhole fd00:172:30:8ed8:7d36:d84:30d9:2880/122 dev lo proto bird metric 1024 pref medium
fd00:172:30:912c:b19b:1bf6:6310:4640/122 via ::6 dev eth0 proto bird metric 1024 pref medium
fe80::/64 dev eth0 proto kernel metric 256 pref medium
fe80::/64 dev calief6f80a244b proto kernel metric 256 pref medium
fe80::/64 dev calibb1f3303ca0 proto kernel metric 256 pref medium
fe80::/64 dev cali6e113f50928 proto kernel metric 256 pref medium
default via ::1 dev eth0 proto static metric 1024 pref medium
 ```

To also apply this onto routes learned from the pod network I would like to add the source hint also to learned routes.

Do you see any issues with this?

I don't know why it is not working :(
ref: https://marc.info/?l=bird-users&m=175036644407969&w=2

## What this PR does / why we need it

The goal of this PR — programming a source address (`krt_prefsrc`) onto kernel
routes learned from BGP peers — requires BIRD 2+. `calico/node` shipped the
long-deprecated `calico/bird` fork (BIRD 1.6.8), so this PR migrates
`calico/node`'s BGP data plane to **upstream BIRD 3.3.0** and implements the
source-address feature on top of it.

**Motivation:** on a node with several addresses on one interface, the kernel
selects the lowest address as the source for traffic to remote pod networks
(e.g. for kube-proxy SNAT), which may be the wrong address. Felix's
`deviceRouteSourceAddress`/`deviceRouteSourceAddressIPv6` already fix this for
*directly-attached* pod routes, but not for routes *learned from other nodes*.
Setting `krt_prefsrc` on BGP-learned routes makes node→remote-pod traffic use a
deterministic source address (the node's BGP source IP).

## Type of change
- Feature: source address on BGP-programmed kernel routes
- Dependency upgrade: `calico/bird` 1.6.x fork → upstream BIRD 3.3.0

## What's included
- **build: add calico/bird image from upstream BIRD 3.3.0** — new in-repo `bird/`
  producer image built from the upstream release tarball on a glibc base (no
  libssh/RPKI) so `bird`/`birdcl` are copy-compatible with node's UBI runtime;
  `node` consumes it via `FROM ${BIRD_IMAGE}`.
- **feat: migrate node and confd to upstream BIRD 3 and set kernel route source
  address** — BIRD 3 runs a single daemon for both families over one control
  socket. confd renders three consolidated config files
  (`bird.cfg`/`bird_ipam.cfg`/`bird_aggr.cfg`) using channel syntax; the
  `bird6.*` templates/tomls/service are removed. IPIP is expressed with
  recursive static tunnel routes (no `krt_tunnel`). Node health checks, status
  populators and `calicoctl node status` parse the single daemon's output per
  address family and tolerate BIRD 3's capitalised `show protocols` headers,
  `Table masterN:` route headers and millisecond timestamps. The kernel
  programming filter sets `krt_prefsrc` to the peer's source address (the node
  IP for mesh/`UseNodeIP` peers) on learned routes.
- **ci: build the bird image and validate rendered BIRD configs** — builds the
  image in CI and parses every rendered golden config with a real BIRD 3 binary
  (`bird -p -c`), catching syntax errors golden text comparison cannot.
- **refactor(felix): generate standalone BIRD peers into a shared conf.d** — the
  standalone (OpenStack/bare-metal) BIRD generators now write both IPv4 and IPv6
  peers into a single `/etc/bird/conf.d` drop-in directory with AFI-prefixed
  filenames.

**Release note:**
```release-note
Upgrade calico/node's BGP backend from the deprecated calico/bird 1.6.x fork to upstream BIRD 3.3.0 (a single daemon serving both IPv4 and IPv6).
calico/node now sets a preferred source address (krt_prefsrc) on BGP-learned kernel routes, so traffic from a node to remote pod networks uses a deterministic source address (the node's BGP source IP) instead of the kernel's default lowest-address selection.
```


Cellebyte:
Tested and it seems bird ignores the configuration or it is still the wrong section in the config.

Cellebyte:
I think this requires a bird update.

Issue #11856, opened 2026-02-16
Segfault in calico-kube-controllers (etcd mode) in loadbalancer_controller.go on upgrade (v3.29.7 -> v3.30.6)


VannTen:
## Expected Behavior

calico-kube-controllers does not crash

## Current Behavior

calico-kube-controllers crash loop when attempting (from what I can tell) to reconcile loadbalancer object (?) (NOTE: we don't have any LoadBalancer type service in the Kubernetes clusters), after upgrading from 3.29.7 to 3.30.6 Here is the log:
```
[...] # Let me know if the log before this are needed
I0216 16:10:18.692748      14 shared_informer.go:320] Caches are synced for nodes
I0216 16:10:18.692772      14 shared_informer.go:313] Waiting for caches to sync for pods
I0216 16:10:18.693279      14 shared_informer.go:320] Caches are synced for loadbalancer
2026-02-16 16:10:18.693 [INFO][14] kube-controllers/policy_controller.go 255: Successfully updated network policy key="tyforge-integ-defectdojo/knp.default.defectdojo-redis"
2026-02-16 16:10:18.693 [INFO][14] kube-controllers/policy_controller.go 226: Create/Update NetworkPolicy in Calico datastore key="tyforge-integ-zammad/knp.default.zammad-redis"
2026-02-16 16:10:18.693 [INFO][14] kube-controllers/loadbalancer_controller.go 234: Will run periodic IPAM sync every 5m0s
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1d89ab5]

goroutine 152 [running]:
github.com/projectcalico/calico/kube-controllers/pkg/controllers/loadbalancer.(*loadBalancerController).handleBlockUpdate(0xc0006be5a0, {{0x28cafa0, 0xc00053ccc0}, {0x23b0f20, 0xc0009ba8c0}, {0xc000caa330, 0x9}, 0x0, 0x0})
	/go/src/github.com/projectcalico/calico/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go:303 +0xb55
github.com/projectcalico/calico/kube-controllers/pkg/controllers/loadbalancer.(*loadBalancerController).handleUpdate(0xc0006be5a0, {0x236a340?, 0xc00099cf80?})
[...]

PR #11913, opened 2026-02-24
prevent nil pointer dereference in handleBlockUpdate

MichalFupso:
Blocks with leaked IPs can have nil affinity while still containing  allocations. The previous condition allowed these blocks to fall through into processing logic that assumes load balancer attributes (HandleID, namespace, service) are present.. Only process blocks explicitly affine to "virtual:LoadBalancer".

**Release note:**
```release-note
LoadBalancer controller prevent nil pointer dereference in handleBlockUpdate
```

Fixes: https://github.com/projectcalico/calico/issues/11856

### DIFF
--- a/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
+++ b/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
@@ -291,7 +291,12 @@ func (c *loadBalancerController) handleUpdate(update any) {
 
 func (c *loadBalancerController) handleBlockUpdate(kvp model.KVPair) {
 	block, ok := kvp.Value.(*model.AllocationBlock)
-	if !ok || block == nil {
+	if !ok {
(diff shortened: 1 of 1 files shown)


kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -290,17 +290,17 @@ func (c *loadBalancerController) handleUpdate(update any) {
 }
 
 func (c *loadBalancerController) handleBlockUpdate(kvp model.KVPair) {
-	if kvp.Value == nil {
+	block, ok := kvp.Value.(*model.AllocationBlock)
+	if !ok || block == nil {
```
caseydavenport: Do we actually expect `!ok` here ever? Sounds like a bug if we hit that branch. I'd suggest handling it explicitly with an error log.

PR #11917, opened 2026-02-24
fix: advertise /32 LB IPs assigned from IPPool via BGP

caseydavenport:
## Summary
- Fixes a bug where BIRD does not advertise /32 service LoadBalancer IP addresses via BGP
- When `serviceLoadBalancerIPs` contains a `/32` CIDR and aggregation is enabled (the default), the global path correctly filters out /32s to defer to per-service logic, but the per-service path only checked the deprecated `svc.Spec.LoadBalancerIP` field — which is empty when IPs are assigned from a Calico IPPool. The actual assigned IP in `svc.Status.LoadBalancer.Ingress` was never checked.
- Introduces `hasAnySingleLoadBalancerIP()` which checks both spec and status ingress IPs, fixing advertisement for IPPool-assigned /32 LB IPs.

```release-note
Fix advertisement of /32 LB IP addresses when not present in the Service Spec
```

### DIFF
--- a/confd/pkg/backends/calico/routes.go
+++ b/confd/pkg/backends/calico/routes.go
@@ -394,6 +394,23 @@ func (rg *routeGenerator) isAllowedLoadBalancerIP(loadBalancerIP string) bool {
 	return false
 }
 
+// hasAnySingleLoadBalancerIP checks whether the service has any LoadBalancer IP
+// that matches a single-IP (/32 or /128) entry in serviceLoadBalancerIPs.
+// It checks both svc.Spec.LoadBalancerIP (deprecated but still used) and the
+// actual assigned IPs from svc.Status.LoadBalancer.Ingress, since IPs assigned
+// from a Calico IPPool only appear in status, not spec.
+func (rg *routeGenerator) hasAnySingleLoadBalancerIP(svc *v1.Service) bool {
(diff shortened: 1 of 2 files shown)


confd/pkg/backends/calico/routes.go
```diff
@@ -394,6 +394,23 @@ func (rg *routeGenerator) isAllowedLoadBalancerIP(loadBalancerIP string) bool {
 	return false
 }
 
+// hasAnySingleLoadBalancerIP checks whether the service has any LoadBalancer IP
+// that matches a single-IP (/32 or /128) entry in serviceLoadBalancerIPs.
+// It checks both svc.Spec.LoadBalancerIP (deprecated but still used) and the
+// actual assigned IPs from svc.Status.LoadBalancer.Ingress, since IPs assigned
+// from a Calico IPPool only appear in status, not spec.
+func (rg *routeGenerator) hasAnySingleLoadBalancerIP(svc *v1.Service) bool {
```
MichalFupso, line 402: Should this also check for externalIP? They are supposed to be managed externally and routeable to the nodes, so might not have to check

caseydavenport: We have a separate code path for external IPs

PR #11919, opened 2026-02-25
BPF: remove overlay device IP requirement from BPF dataplane

tomastigera:
The BPF dataplane doesn't need a separate tunnel device IP (HOST_TUNNEL_IP) — HOST_IP serves the same role for inner packet source addressing and bpf_skb_set_tunnel_key local IP.

- Remove host_tunnel_ip from BPF globals struct and HOST_TUNNEL_IP macro
- Set local_ipv4/v6 = HOST_IP in bpf_skb_set_tunnel_key (fib_co_re.h)
- Remove now-unnecessary SNAT-to-tunnel-IP fixup from tc.c
- Remove HostTunnelIPv4/v6 from Go TC attach point and libbpf interop
- Skip IP assignment on VXLAN/IPIP overlay devices in BPF mode
- Skip IPIP/VXLAN tunnel masquerade iptables rules in BPF mode
- Update BPF unit tests and FV source IP assertions accordingly

### BPFOverlayIPOnDevice option

A new `BPFOverlayIPOnDevice` FelixConfiguration parameter (default `false`) is added to optionally restore the legacy behavior. When set to `true`:
- IPIP/VXLAN tunnel devices get an IP address assigned (as before this PR)
- BPF uses the tunnel device IP (`HOST_TUNNEL_IP`) as the encapsulation source instead of the node IP
- Masquerade rules for IPIP/VXLAN tunnels are generated in iptables/nftables
- The SNAT conflict resolution path in `tc.c` activates for IPIP/VXLAN (not just WireGuard)

This has no effect on WireGuard tunnels, which always use a tunnel device IP regardless of this setting.

The BPF C code (`fib_co_re.h`) now checks at runtime: if `HOST_TUNNEL_IP` is set, it uses that as the encap source; otherwise it falls back to `HOST_IP`. This means the same compiled BPF programs work for both modes — the Go side simply controls whether `HOST_TUNNEL_IP` is populated in the BPF globals map.

Standalone FV tests (`describeBPFOverlayIPOnDeviceTests`) verify host-networked and pod connectivity with the option enabled for both IPIP and VXLAN.

**Release note:**
```release-note
ebpf: vxlan and ipip do not have any IP assigned, therefore when a node talks to a pod via the overlay, it always uses the host's main device IP (as configured and presented in k8s as internal IP). The legacy behavior (IP on overlay device) can be restored by setting BPFOverlayIPOnDevice to true.
```

**Is upgrade to this feature safe?**

  Pod-to-pod via tunnel (the common case): Zero impact. The removed SNAT only applied to host-originated traffic (CALI_F_TO_HEP && (CALI_F_LO || CALI_F_MAIN)), not workload endpoints. Inner packets always use pod IPs. The outer source IP change doesn't
  matter — receiving side has NORMAL CT entries with no tun_ip check.

  Host-to-remote-pod via tunnel: Safe during rolling upgrade because of three layers of protection:

  1. Tunnel device keeps its IP — ensureAddressOnLink is skipped when addr="", but nothing removes the old IP. The masquerade address notifier never fires. Kernel conntrack entries survive.
  2. Kernel conntrack handles existing connections — the actual SNAT was done by iptables MASQUERADE in POSTROUTING (before BPF TC egress). Removing the masquerade rule only affects new connections. Established connections continue to be SNATed/un-SNATed
  via cached conntrack entries.
  3. BPF conntrack persists — pinned BPF maps survive Felix restart. BPF NORMAL entries for these flows remain valid.

  New connections after upgrade: Use host_ip as source (no SNAT). Return traffic comes back via direct routing instead of tunnel. Works because host IPs are always directly routable between nodes.

  If issues arise, set `BPFOverlayIPOnDevice: true` in FelixConfiguration to restore the old behavior without rolling back.


felix/dataplane/linux/route_mgr.go
```diff
@@ -582,8 +582,10 @@ func (m *routeManager) configureTunnelDevice(
 	}
 
 	if m.maintainIPOnly {
-		if err := m.ensureAddressOnLink(addr, link); err != nil {
-			return fmt.Errorf("failed to ensure address of interface: %s", err)
+		if addr != "" {
+			if err := m.ensureAddressOnLink(addr, link); err != nil {
```
fasaxc: Do we care about also removing the IP when it's not needed?  Its presence will affect source IP selection so I think it'd be good to clean it up to avoid "memory" of the past state.

felix/fv/bpf_test.go
```diff
@@ -101,6 +101,134 @@ var _ = describeBPFTests(withProto("tcp"),
...
 	withBPFLogLevel("off"))
 
+// Standalone tests for BPFOverlayIPOnDevice — verifies that overlay connectivity works when the
+// tunnel device is assigned an IP address (the legacy behaviour).  The primary scenario is
+// host-networked traffic to/from remote tunneled workloads, which is where HOST_TUNNEL_IP and
+// the SNAT conflict resolution logic come into play.
+var _ = describeBPFOverlayIPOnDeviceTests("ipip")
+var _ = describeBPFOverlayIPOnDeviceTests("vxlan")
+
+func describeBPFOverlayIPOnDeviceTests(tunnel string) bool {
```
fasaxc: New file?  LEt's not grow bpf_test.go

fasaxc (commented):
I like the idea but I'm wondering if we should either:

- Keep the IP on the interface and maintain it as now.  IP largely unused, unless client binds to it explicitly (which would presumably still work?).
- Actively clean up the IP.  Make a clean break, no left-over IP to accidentally bind to.

Otherwise, feels a little messy that we'll keep allocating these IPs but not use them at all.

felix/dataplane/linux/route_mgr.go
tomastigera: It is now being removed since the feature is no opt-in and thus the user will be aware of possible disruption. It was there to facilitate smooth upgrade when the feature was turned on by default

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -708,6 +708,14 @@ type FelixConfigurationSpec struct {
 	// BPFEnabled, if enabled Felix will use the BPF dataplane. [Default: false]
 	BPFEnabled *bool `json:"bpfEnabled,omitempty" validate:"omitempty"`
 
+	// BPFOverlayIPOnDevice, if enabled (the default), Felix assigns an IP address to overlay tunnel devices
+	// (IPIP/VXLAN) in BPF mode and uses it as the encapsulation source IP.  When disabled, BPF programs use
+	// the node IP directly for encapsulation without requiring a separate tunnel device IP.  This option has
+	// no effect on WireGuard tunnels, which always use a tunnel device IP.  Defaults to true to preserve the
+	// legacy behaviour for clusters upgraded from earlier releases; set to false to opt in to the streamlined
+	// no-tunnel-IP behaviour.  [Default: true]
+	BPFOverlayIPOnDevice *bool `json:"bpfOverlayIPOnDevice,omitempty" validate:"omitempty"`
```
fasaxc: Let's avoid bool for the usual reasons
```
bpfHostOverlayInnerSourceIP: PreserveHostAddress | SNATToTunnelAddress 
```

fasaxc (changes requested):

fasaxc (approved):

Issue #11424, opened 2025-11-21
Kernel panic in skb_segment() with UDP GSO packets over VXLAN overlay


echkenluo:
<!--- Provide a general summary of the issue in the Title above -->
<!--- If you believe the issue may have security implications, please report it as a vulnerability -->
<!--- Report a vulnerability: https://github.com/projectcalico/calico/security -->

## Expected Behavior
Calico should handle UDP GSO packets traversing VXLAN overlay network without causing kernel panics, especially when BPF DSR mode is enabled for external services.  

## Current Behavior
 Nodes experience kernel panics with null pointer dereference in skb_segment() when processing UDP GSO packets over VXLAN. The crash occurs during packet forwarding after VXLAN encapsulation, specifically in the GSO              
  segmentation path.   

 Crash backtrace:                                                                                                                                                                                                                    
```
  PID: 0      TASK: ff16ba3686914000  CPU: 7   COMMAND: "swapper/7"                                                                                                                                                                   
      [exception RIP: skb_segment+558]                                                                                                                                                                                                
      RIP: ffffffffaa037f7e  RSP: ff3ca05846b44708  RFLAGS: 00010246                                                                                                                                                                  
      RAX: 0000000000000000  RBX: 0000000000000000  RCX: 0000000000000000                                                                                                                                                             
      RDX: 0000000000000000  RSI: ff16ba3aa69472c0  RDI: 0000000000000544                                                                                                                                                             
   #8 [ff3ca05846b447e0] tcp_gso_segment at ffffffffaa1219c5                                                                                                                                                                          
   #9 [ff3ca05846b44830] inet_gso_segment at ffffffffaa136f4b                                                                                                                                                                         
  #10 [ff3ca05846b44880] skb_mac_gso_segment at ffffffffaa04cb9d                                                                                                                                                                      
  #11 [ff3ca05846b448a8] skb_udp_tunnel_segment at ffffffffaa12bbae                                                                                                                                                                   
  #12 [ff3ca05846b44908] inet_gso_segment at ffffffffaa136f4b                                                                                                                                                                         
  #13 [ff3ca05846b44958] skb_mac_gso_segment at ffffffffaa04cb9d                                                                                                                                                                      
  #14 [ff3ca05846b44980] __skb_gso_segment at ffffffffaa04ccbd                                                                                                                                                                        
  #15 [ff3ca05846b449a8] validate_xmit_skb at ffffffffaa04d15e                                                                                                                                                                        
[...]

PR #11920, opened 2026-02-25
[BPF] Fix kernel crash on UDP GSO FRAGLIST packets after bpf_skb_pull_data

tomastigera:
A partial bpf_skb_pull_data() on a UDP GSO_FRAGLIST packet (e.g. GRO-aggregated QUIC) corrupts the fraglist geometry, causing a NULL pointer dereference in skb_segment() when the kernel tries to segment the packet.

After the normal partial pull succeeds, check if the packet is UDP GSO. If so, fully linearize it by pulling skb->len bytes, which collapses all non-linear data into the linear head (data_len=0, frag_list=NULL). This is rarely hit since GRO packets normally have headers in the linear portion and the partial pull is skipped entirely.

Also move skb_is_gso() above skb_refresh_validate_ptrs() so it can be called from there.

Fixes: https://github.com/projectcalico/calico/issues/11424

**Release note:**
```release-note
ebpf: Fix kernel crash on UDP GSO FRAGLIST packets after partial bpf_skb_pull_data by fully linearizing the packet. Auto-detected via kernel version (fixed in 6.16+), overridable via FeatureDetectOverride.
```


api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -933,6 +933,14 @@ type FelixConfigurationSpec struct {
 	//+kubebuilder:validation:Enum=Enabled;Disabled
 	BPFProfiling string `json:"bpfProfiling,omitempty"`
 
+	// BPFUDPGSOLinearize controls whether Felix fully linearizes UDP GSO packets
+	// after a partial bpf_skb_pull_data() to work around a kernel bug that causes
+	// crashes with GSO_FRAGLIST packets. The kernel fix is in 6.16+.
+	// When set to "Auto", Felix detects the kernel version and only enables
+	// linearization on kernels older than 6.16. [Default: Auto]
+	// +kubebuilder:validation:Enum=Enabled;Disabled;Auto
+	BPFUDPGSOLinearize string `json:"bpfUDPGSOLinearize,omitempty"`
```
fasaxc: Will we need the config flag long term (don't we have a general method to override feature detection, so you could just use that feature flag)?

PR #11932, opened 2026-02-25
Add renovate bot for dependency management

MichalFupso:
Adds renovate bot to keep k8s, go and golang.org/x/* libraries up to date. Works with dependabot alert database to update libraries with CVEs

**Release note:**
```release-note
TBD
```

### DIFF
--- a/.github/workflows/renovate-bot.yml
+++ b/.github/workflows/renovate-bot.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run Renovate
-        uses: renovatebot/github-action@v46.1.1
+        uses: renovatebot/github-action@v46.1.4
         with:
           token: ${{ secrets.MARVIN_RENOVATE_PAT }}
           configurationFile: renovate.json


.github/workflows/renovate-bot.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  renovate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Run Renovate
+        uses: renovatebot/github-action@v46.1.1
```
Behnam-Shobiri: Can we update to latest? 
Currently, 46.1.3

.github/workflows/renovate-bot.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          token: ${{ secrets.MARVIN_RENOVATE_PAT }}
+          configurationFile: renovate.json
+          mount-docker-socket: true
+          docker-user: root
```
Behnam-Shobiri, line 21: Do we. need root and docker-socket? 
Root -> extra perm in the container 
docker socket -> may lead to privilege escalation

MichalFupso: We need it to be able to run `make gen-deps-files` in calico > 3.31. It generates deps files that we're not able to store without root, and the socket is needed to run the make target. It runs inside of our build container and the docker socket allows for "docker in docker" inside the renovate container

PR #11939, opened 2026-02-26
Fix BPF FV flake: sync NAT maps on all felixes before connecting

sridhartigera:
## Summary
- The "should have connectivity after a backend is replaced by a new one" BPF FV test was flaky because it only checked felix-1's NAT maps before starting a persistent connection
- In the shared-cgroup FV environment, felix-0's CTLB cgroup program intercepts connects from all containers and reads from felix-0's maps
- If felix-0 hadn't synced yet, the CTLB program returned NAT_NO_BACKEND → EPERM, causing test-connection to exit before deleting the loop file, resulting in a timeout
- Fix: check all felixes' NAT maps so both the CTLB and TC paths are ready

## Test plan
- [ ] Existing BPF FV tests pass (this fixes a flaky one)
- [ ] No new test needed — this is a test fix

### DIFF
--- a/felix/fv/bpf_test.go
+++ b/felix/fv/bpf_test.go
@@ -3559,19 +3559,23 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 								natFtKey = nat.NewNATKey(net.ParseIP(ip), port, numericProto)
 								family = 4
 							}
-							Eventually(func() bool {
-								m, be, _ := dumpNATMapsAny(family, tc.Felixes[1])
+							// Check all felixes. In the shared-cgroup FV environment,
+							// felix-0's CTLB program intercepts connects from all
+							// containers, so its NAT maps must be synced too.
+							for _, f := range tc.Felixes {
+								Eventually(func() bool {
+									m, be, _ := dumpNATMapsAny(family, f)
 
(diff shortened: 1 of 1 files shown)


felix/fv/bpf_test.go
```diff
@@ -3559,18 +3559,24 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 								natFtKey = nat.NewNATKey(net.ParseIP(ip), port, numericProto)
 								family = 4
 							}
+							// Check all felixes. In the shared-cgroup FV environment,
+							// felix-0's CTLB program intercepts connects from all
+							// containers, so its NAT maps must be synced too.
 							Eventually(func() bool {
-								m, be, _ := dumpNATMapsAny(family, tc.Felixes[1])
+								for _, f := range tc.Felixes {
```
tomastigera: woudn't it make a little bit more sense to have the loop outside of eventually. This way it keeps querying felix that is possibly already in sync. It should way for them one by one with higher prob that once one is in sync the other will be too.

sridhartigera: Good point! Moved the loop outside `Eventually` so each felix gets its own poll — once one is confirmed synced, we move on without re-querying it.

PR #11961, opened 2026-03-01
Support spec.loadBalancerIP in LoadBalancer controller

### LINKED ISSUE #11815: LoadBalancer IPAM: support IP assignment with `Service.spec.externalIPs/loadBalancerIP`
We use Calico for BGP peering and announcing Service IPs to the upstream fabric.

`BGPConfiguration` allows Services from pre-defined ranges to be used and announced:
```shell
› kubectl get bgpconfiguration default -oyaml | yq .spec.serviceLoadBalancerIPs
- cidr: 10.10.10.0/24
- cidr: 192.168.0.0/27
```

It would be great if Loadbalancer controller would not only tolerate `projectcalico.org/loadBalancerIPs` for [manual assignment](https://docs.tigera.io/calico/latest/networking/ipam/service-loadbalancer#manual-service-ip-address-assignment)  but also actual `Service.spec.externalIPs` and `Service.spec.loadBalancerIP`.

This would help to gradually migrate existing charts when payload is deployed to clusters where Calico LoadBalancer IP address management is used without having to switch the whole chart fleet to the Calico-specific annotation.

nebojsaj1726:
Fixes #11815

Adds support for spec.loadBalancerIP as a fallback in the LoadBalancer controller when no Calico-specific annotation is set.                              
   
 **Changes:**
 - spec.loadBalancerIP is now recognized as an opt-in signal in RequestedServicesOnly mode via  IsCalicoManagedLoadBalancer
 - spec.loadBalancerIP is used as the requested IP in IPAM when no projectcalico.org/loadBalancerIPs annotation is  present

 **Testing:**                                                                                                                                
 Added a unit test that verifies the new behavior:
 - should recognize spec.loadBalancerIP as calico-managed in RequestedServicesOnly mode

  **Release note:**

  ```release-note
 LoadBalancer controller now supports spec.loadBalancerIP as a fallback when no Calico-specific annotation is set.
```

### DIFF
--- a/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
+++ b/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
@@ -701,6 +701,15 @@ func (c *loadBalancerController) assignIP(svc *v1.Service) ([]string, error) {
 		return nil, err
 	}
 
+	// Fall back to spec.loadBalancerIP when no specific IP was requested via annotation.
+	if loadBalancerIPs == nil && svc.Spec.LoadBalancerIP != "" {
+		ip := cnet.ParseIP(svc.Spec.LoadBalancerIP)
+		if ip == nil {
+			return nil, fmt.Errorf("invalid IP in spec.loadBalancerIP: %s", svc.Spec.LoadBalancerIP)
+		}
+		loadBalancerIPs = []cnet.IP{*ip}
+	}
+
 	var assignedIPs []string
 
 	metadataAttrs := map[string]string{
@@ -912,10 +921,11 @@ func IsCalicoManagedLoadBalancer(svc *v1.Service, assignIPs api.AssignIPs) bool 
 
 		if svc.Annotations[annotationIPv4Pools] != "" ||
 			svc.Annotations[annotationIPv6Pools] != "" ||
-			svc.Annotations[annotationLoadBalancerIP] != "" {
+			svc.Annotations[annotationLoadBalancerIP] != "" ||
+			svc.Spec.LoadBalancerIP != "" {
 
 			if svc.Spec.LoadBalancerClass != nil && *svc.Spec.LoadBalancerClass != calicoLoadBalancerClass {
-				log.WithFields(log.Fields{"svc": svc.Name, "ns": svc.Namespace}).Warn("calico LoadBalancer annotation set with spec.LoadBalancerClass != calico is not supported")
+				log.WithFields(log.Fields{"svc": svc.Name, "ns": svc.Namespace}).Warn("calico LoadBalancer IP request set with spec.LoadBalancerClass != calico is not supported")
 				return false
 			}
 			return true
--- a/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller_ut_test.go
+++ b/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller_ut_test.go
@@ -540,6 +540,21 @@ var _ = Describe("LoadBalancer controller UTs", func() {
 		Expect(cli.IPAMUpgradeCallCount()).To(Equal(2))
 	})
 
+	It("should recognize spec.loadBalancerIP as calico-managed in RequestedServicesOnly mode", func() {
+		// With no annotations and no spec.loadBalancerIP, not managed in RequestedServicesOnly mode.
+		managed := IsCalicoManagedLoadBalancer(&svc, apiv3.RequestedServicesOnly)
+		Expect(managed).To(BeFalse())
+
+		// Setting spec.loadBalancerIP should make it calico-managed (user is requesting a specific IP).
+		svc.Spec.LoadBalancerIP = "10.0.0.5"
+		managed = IsCalicoManagedLoadBalancer(&svc, apiv3.RequestedServicesOnly)
+		Expect(managed).To(BeTrue())
+
+		// Should still be managed in AllServices mode.
+		managed = IsCalicoManagedLoadBalancer(&svc, apiv3.AllServices)
+		Expect(managed).To(BeTrue())
+	})
+
 	It("should handle invalid IP addresses in allocation tracker without panicking", func() {
 		svcKey, err := serviceKeyFromService(&svc)
 		Expect(err).ToNot(HaveOccurred())

### CODE REVIEW

**Comment by caseydavenport** [comment]
CC @MichalFupso

**Comment by defo89** [comment]
Kindly ping. Is it still planned to include this in `Calico v3.32.0`?

File: kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go [hunk 1]
```diff
@@ -417,6 +417,21 @@ func (c *loadBalancerController) ensureDatastoreUpgraded() error {
 // - Updates the controllers internal state tracking of which IP addresses are allocated.
 // - Updates the IP addresses in the Service Status to match the IPAM DB.
 func (c *loadBalancerController) syncService(svcKey serviceKey) {
+	// If the service has spec.externalIPs, reflect them directly in status without Calico IPAM.
+	if externalSvc, err := c.serviceLister.Services(svcKey.namespace).Get(svcKey.name); err == nil && len(externalSvc.Spec.ExternalIPs) > 0 {
```
caseydavenport: confd already advertises `spec.externalIPs` straight from the spec, gated by `serviceExternalIPs` - see [`getAllRoutesForService`](https://github.com/projectcalico/calico/blob/816b3bd1037b05fb901dd81451f142a95f3db05b/confd/pkg/backends/calico/routes.go#L272-L281). So what does writing them into LB status actually buy us?

I think our confd implementation is now going to advertise these IPs if they match against the BGPConfiguration.Spec.LoadBalancerIPs now, which is a slightly awkward behavioral change.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -417,6 +417,21 @@ func (c *loadBalancerController) ensureDatastoreUpgraded() error {
...
+		var ingress []v1.LoadBalancerIngress
+		for _, ip := range externalSvc.Spec.ExternalIPs {
+			ingress = append(ingress, v1.LoadBalancerIngress{IP: ip})
+		}
+		externalSvc.Status.LoadBalancer.Ingress = ingress
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if _, updateErr := c.clientSet.CoreV1().Services(externalSvc.Namespace).UpdateStatus(ctx, externalSvc, metav1.UpdateOptions{}); updateErr != nil {
```
caseydavenport: This does an unconditional `UpdateStatus` on every sync, and `syncIPAM` runs on a resync timer, so we'd rewrite status every resync for every externalIPs service. The IPAM path gates writes behind `needsStatusUpdate` - we should do the same here instead of writing every time.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -871,7 +895,8 @@ func IsCalicoManagedLoadBalancer(svc *v1.Service, assignIPs api.AssignIPs) bool
 
 		if svc.Annotations[annotationIPv4Pools] != "" ||
 			svc.Annotations[annotationIPv6Pools] != "" ||
-			svc.Annotations[annotationLoadBalancerIP] != "" {
+			svc.Annotations[annotationLoadBalancerIP] != "" ||
```
caseydavenport, line 924: Worth noting that this will not unconditionally copy LB IPs to the status, whereas before users had some control over whether an IP should be handled by Calico (with an annotation specific to us) or someone else.

Not sure it matters, and I think respecting it is probably the right thing regardless (even though the field is deprecated). Just though I'd mention it.

caseydavenport (changes requested):
Apologies for the long delay on this one, and understood if you're not able to pick it back up.

But I've added some thoughts / questions that I'd like your input on.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
nebojsaj1726: Hi, thanks for the review.

I wasn't aware that confd already handles BGP advertisement of spec.externalIPs via the serviceExternalIPs allow-list. So the gap I was filling isn't BGP advertisement but status.loadBalancer.ingress population, which some external tooling (external-dns, ingress controllers) relies on rather than reading the spec directly. 

On the behavioral change concern - yes, writing into status causes the BGPConfiguration.Spec.LoadBalancerIPs filter to also apply to these IPs, which is unintended. I can add a guard to avoid the status update if the service already has its externalIPs advertised via the confd path, or we could scope this to only populate status without triggering the LoadBalancerIPs advertisement path. 

My question is whether populating status for spec.externalIPs is needed here at all - I can drop that part and focus only on spec.loadBalancerIP if you think the externalIPs case is already sufficiently handled by confd.

caseydavenport: Yeah, at a minimum we should split that out into a second PR so we can discuss it more distinctly - I think it's a separate concern from the LB IPs and it will be easier to reason about them separately.

I'm not sure if there is value in copying the external IPs into the status, but we could explore use-cases to find out.

nebojsaj1726: Done - removed the externalIPs block from this PR. The description and release note have been updated to reflect the narrowed scope (spec.loadBalancerIP only). 

This PR is ready for another look when you get a chance.

kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go
```diff
@@ -912,7 +921,8 @@ func IsCalicoManagedLoadBalancer(svc *v1.Service, assignIPs api.AssignIPs) bool
...
+			svc.Spec.LoadBalancerIP != "" {
 
 			if svc.Spec.LoadBalancerClass != nil && *svc.Spec.LoadBalancerClass != calicoLoadBalancerClass {
 				log.WithFields(log.Fields{"svc": svc.Name, "ns": svc.Namespace}).Warn("calico LoadBalancer annotation set with spec.LoadBalancerClass != calico is not supported")
```
caseydavenport: a nit, but this log refers to the annotation but can now get hit by spec.LoadBalancerIP - so we should tidy that up to avoid confusion.

nebojsaj1726: Updated to "IP request set" instead of "annotation set" to cover both cases, let me know if you prefer different wording.

caseydavenport (approved):

caseydavenport:
LGTM, thanks!

defo89:
Hi @caseydavenport @nebojsaj1726 do you plan to cherry-pick this to the v3.31 and v3.32 branches?

Let me know if I can help - happy to work on cherry picks if needed.

nebojsaj1726:
Thanks @defo89. Backport decisions are not up to me, no objection from my side if you want to open the cherry-picks.

caseydavenport:
I think it would be OK to backport this to v3.32. Worth noting that the v3.31 release branch is nearing EOL, not sure if there will be another patch on that branch.

defo89:
Thank you @nebojsaj1726 @caseydavenport - I have opened a backport to v3.32 https://github.com/projectcalico/calico/pull/14001 and agree about dropping v3.31 if it's nearing EOL.

PR #11964, opened 2026-03-02
Add SEMAPHORE_AGENT_UPLOAD_JOB_LOGS variable

lwr20:
Set environment variable to save job logs as an artifact in the case that the logs are bigger than 16MB.

**Release note:**
```release-note
TBD
```


Glen-Tigera (commented):
LGTM

lwr20:
Looks like we already have this set in all jobs.

PR #11969, opened 2026-03-03
CORE-12071: Added IPAM GC for kubevirt VMs

skoryk-oleksandr:
- Add `IPAM GC` for `KubeVirt VM/VMI` allocations in `kube-controllers` — validates allocations against live `VM/VMI` resources with a 5-minute grace period for VM recreation
  events

**Release note:**
```release-note
TBD
```

### DIFF
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -925,8 +925,13 @@ func (c *IPAMController) checkAllocations() ([]string, error) {
 			} else if a.isVMAllocation() {
 				// VM/VMI allocation with no backing VM or standalone VMI.
 				// Use a dedicated grace period to tolerate transient gaps
-				// during VM restarts and migrations.
-				a.markLeak(c.vmRecreationGracePeriod)
+				// during VM restarts and migrations. If the operator configured
+				// a longer LeakGracePeriod, respect that as well.
+				gracePeriod := c.vmRecreationGracePeriod
+				if c.config.LeakGracePeriod != nil && c.config.LeakGracePeriod.Duration > gracePeriod {
+					gracePeriod = c.config.LeakGracePeriod.Duration
+				}
+				a.markLeak(gracePeriod)
 			} else if c.config.LeakGracePeriod != nil {
 				// The allocation is NOT valid, but the Kubernetes node still exists, so our confidence is lower.
 				// Mark as a candidate leak. If this state remains, it will switch


kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1074,65 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
 	return false
 }
 
+// isVMOrStandaloneVMIExists checks whether the backing VM or standalone VMI for
+// a VMI-type IP allocation still exists. Returns true on any error or missing
+// data to avoid incorrectly releasing in-use IPs.
+func (c *IPAMController) isVMOrStandaloneVMIExists(a *allocation) bool {
```
caseydavenport:
```suggestion
func (c *IPAMController) isVMAllocationValid(a *allocation) bool {
```

^ I think this is a cleaner name.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1074,65 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
...
+	ns := a.attrs[ipam.AttributeNamespace]
+	vmName := a.attrs[ipam.AttributeVMIName]
+	logc := log.WithFields(a.fields())
+
+	// If we can't reason about it, don't GC it.
+	// (We can tighten this later once attrs are guaranteed.)
+	if ns == "" || vmName == "" || c.virtClient == nil {
+		logc.Debugf("Insufficient data to validate VMI allocation, assuming valid. Namespace = %s, vmName = %s", ns, vmName)
```
caseydavenport: I think `logc`  already includes ns / vmName? We should be using WithFields anyway instead of inline.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1074,65 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
...
+	// VM not found — check if a standalone VMI exists.
+	vmi, err := kubevirt.GetVMIResourceByName(context.Background(), c.virtClient, ns, vmName)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			logc.WithError(err).Warnf("Failed to query VMI, skipping VMI existence check for %s/%s", ns, vmName)
+			return true
+		}
+	}
+	if vmi != nil {
+		return true
```
caseydavenport: Debug log in this branch would probably be good.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1074,65 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
...
+	}
+
+	// Neither VM nor standalone VMI found.
+	logc.Debug("Neither VM nor standalone VMI found")
```
caseydavenport, line 1134: Also, the comment here is redundant with the log, can remove it.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1074,65 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
...
+	return false
+}
+
+func (c *IPAMController) getVMByName(
```
caseydavenport: Don't think this signature is long enough to warrant splitting it across lines.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1074,65 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
...
+	ns, vmName string,
+	logc *log.Entry,
+) (*kubevirtv1.VirtualMachine, error) {
+	vm, err := c.virtClient.VirtualMachine(ns).
+		Get(context.Background(), vmName, metav1.GetOptions{})
```
caseydavenport: Same here - we can go up to 180 characters before we should split lines.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1074,65 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logc.WithError(err).Debugf("VM is not found, vmName = %s, namespace = %s", vmName, ns)
```
caseydavenport: logc should include these fields

kube-controllers/pkg/controllers/node/ipam.go
caseydavenport: Do we really need a helper function for this? It's just wrapping a client call. I think this makes the code a bit harder to follow.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1074,65 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
...
+		return true
+	}
+
+	vm, err := c.getVMByName(ns, vmName, logc)
+	if err != nil {
+		logc.WithError(err).Error("Failed to get VM resource. Assuming allocation is valid")
+		return true
+	}
+	if vm != nil {
+		return true
```
caseydavenport, line 1110: Debug log here?

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1074,65 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
...
+	// Neither VM nor standalone VMI found.
+	logc.Debug("Neither VM nor standalone VMI found")
+	return false
+}
+
+func (c *IPAMController) getVMByName(
+	ns, vmName string,
+	logc *log.Entry,
+) (*kubevirtv1.VirtualMachine, error) {
+	vm, err := c.virtClient.VirtualMachine(ns).
```
caseydavenport: We really shouldn't be using a direct client call here - we should have an informer set up so we can read from the cache. Adding direct client calls when there are many VMs will have a major performance impact.

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -366,17 +367,23 @@ func getClients(kubeconfig string) (*kubernetes.Clientset, client.Interface, cli
...
 	var v3c clientset.Interface
 	v3c, err = clientset.NewForConfig(k8sconfig)
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf("failed to build Calico Kubernetes v3 client: %s", err)
+		return nil, nil, nil, nil, fmt.Errorf("failed to build Calico Kubernetes v3 client: %s", err)
 	}
 
-	return k8sClientset, libcalicoClient, v3c, nil
+	// KubeVirt client is optional — log and continue without it on failure.
+	kubevirtClient, err := kubevirt.TryCreateVirtClient(k8sconfig)
```
caseydavenport: This only runs at startup, so if the kubevirt resources are added to a running cluster we won't notice and will need to reboot this pod.

This might not be too bad, but we'll need to at least comment / document it.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -44,10 +45,17 @@ import (
...
 	"github.com/projectcalico/calico/libcalico-go/lib/ipam"
+	"github.com/projectcalico/calico/libcalico-go/lib/kubevirt"
 	cnet "github.com/projectcalico/calico/libcalico-go/lib/net"
 	"github.com/projectcalico/calico/libcalico-go/lib/options"
 )
 
+const (
+	// defaultVMRecreationGracePeriod is the default time window to allow VM recreation.
+	// During this period, if a VM doesn't exist we assume the VM is being recreated (e.g., restart, etc.)
+	defaultVMRecreationGracePeriod = 15 * time.Minute
```
caseydavenport: Can put this with the other const?

caseydavenport: We allow for a configurable leak grace period for pod IPs - do we need that for this as well? 

Worth adding a comment explaining why we need two different grace periods.

caseydavenport (changes requested):
First pass. Thanks @skoryk-oleksandr !

kube-controllers/cmd/kube-controllers/main.go
skoryk-oleksandr: > This might not be too bad, but we'll need to at least comment / document it.

@caseydavenport Do you mean adding information in the docs about installing kubevirt, mentioning that existing pods should be restarted ?

kube-controllers/pkg/controllers/node/ipam.go
skoryk-oleksandr: VM recreation (restart, live migration) typically takes longer than pod rescheduling, so they need a separate grace period. Updated the comment to explain why they're separate.                                                                                                                                                                                                                                         
For now I would keep it as a hardcoded constant (5 minutes) to keep things simple. We can make it configurable via KubeControllersConfiguration if users need to tune it, but I'd rather not add the API field until there's a real need

skoryk-oleksandr:
@caseydavenport Thanks a lot for the code review :+1: . I addressed your comments. Feel free make a second pass.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -71,6 +72,14 @@ const (
 
 	// key for ratelimited sync retries.
 	retryKey = "ipamSyncRetry"
+
+	// defaultVMRecreationGracePeriod is the default time window to allow VM recreation
+	// before treating the allocation as a leak.
+	//
+	// This is separate from the configurable LeakGracePeriod (which applies to pod IP allocations)
+	// because VM recreation (e.g., restart, live migration) can take significantly longer than pod
```
caseydavenport: 🤖 The PR description says "15-minute grace period" but this is set to 5 minutes. Which one is correct?

libcalico-go/lib/kubevirt/client.go
```diff
@@ -50,3 +69,36 @@ func (v *virtClientAdapter) VirtualMachine(namespace string) VMInterface {
 func (v *virtClientAdapter) VirtualMachineInstanceMigration(namespace string) VMIMInterface {
 	return v.client.VirtualMachineInstanceMigrations(namespace)
 }
+
+// TryCreateVirtClient attempts to create a KubeVirt client.
+// Returns nil if KubeVirt is not available.
+func TryCreateVirtClient(restConfig *rest.Config) (VirtClientInterface, error) {
```
caseydavenport: 🤖 `TryCreateVirtClient` is added here but nothing in this PR calls it. The informer creation in `informers.go` creates its own `kubevirtcorev1` client directly. Is this dead code, or is it intended to be used elsewhere?

kube-controllers/pkg/controllers/node/ipam_test.go
```diff
@@ -25,10 +25,12 @@ import (
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/kubernetes/fake"
+	fake "k8s.io/client-go/kubernetes/fake"
```
caseydavenport: 🤖 Nit: the package is already named `fake`, so the explicit alias is redundant.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -904,6 +924,11 @@ func (c *IPAMController) checkAllocations() ([]string, error) {
 				// - The node the allocation belongs to no longer exists.
 				// - The pod owning this allocation no longer exists.
```
caseydavenport, line 923: I know we already spoke of this, and nothing to do in this PR, but I do think we should consider a follow-on to at least respect the config option if it is larger than the default for VM addresses. 

An operator who sets a longer grace period might expect it to cover VM allocations too.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1076,70 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
 	return false
 }
 
+// isVMAllocationValid checks whether the backing VM or standalone VMI for
+// a VMI-type IP allocation still exists. Returns true on any error or missing
+// data to avoid incorrectly releasing in-use IPs.
+func (c *IPAMController) isVMAllocationValid(a *allocation) bool {
+	ns := a.attrs[ipam.AttributeNamespace]
+	vmName := a.attrs[ipam.AttributeVMIName]
```
caseydavenport: Nit: this reads from `AttributeVMIName` but stores it in `vmName`, which is a bit misleading.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1048,6 +1076,70 @@ func (c *IPAMController) allocationIsValid(a *allocation, preferCache bool) bool
...
+	logc := log.WithFields(a.fields())
+
+	// If we can't reason about it, don't GC it.
+	// (We can tighten this later once attrs are guaranteed.)
+	if ns == "" || vmName == "" || c.vmIndexer == nil {
```
caseydavenport: This checks `c.vmIndexer == nil` but doesn't check `c.vmiIndexer`.

I know in the current code this isn't possible, but in the future if `vmIndexer` is non-nil but `vmiIndexer` is somehow nil, `c.vmiIndexer.GetByKey(key)` below will panic. 

```go
if ns == "" || vmName == "" || c.vmIndexer == nil || c.vmiIndexer == nil {
```

kube-controllers/pkg/controllers/node/ipam.go
skoryk-oleksandr: It was initially set to 15 minutes, but that was too long. I reduced it to 5 minutes and forgot to update the PR description.

libcalico-go/lib/kubevirt/client.go
skoryk-oleksandr: You are right. Dead code. Removed it.

kube-controllers/pkg/controllers/node/ipam_test.go
skoryk-oleksandr: Fixed.

kube-controllers/pkg/controllers/node/ipam.go
skoryk-oleksandr: Makes sense. I changed it to `max(defaultVMRecreationGracePeriod, config.LeakGracePeriod)`

kube-controllers/pkg/controllers/node/ipam.go
skoryk-oleksandr: Fixed

kube-controllers/pkg/controllers/node/ipam.go
skoryk-oleksandr: Fixed

PR #11971, opened 2026-03-03
CORE-12209: kubevirt changes for calicoctl

skoryk-oleksandr:
- Add `--kubevirt-ip-persistence=<Enabled|Disabled>` flag to calicoctl ipam configure for controlling KubeVirt VM IP persistence across lifecycle events
  - Make `--strictaffinity` optional so individual IPAM settings can be changed independently
  - Add `AlternateOwnerAttrs` support to IPAM migration, check, and show commands
  - Detect VMI allocation type in borrowed IP display (`calicoctl ipam show --show-borrowed`)
  - Show `HandleID`, `Active/Alternate Owner` Attributes separately in `calicoctl ipam show --ip`
  - Plumb `KubeVirtVMAddressPersistence` through IPAM config backend conversion

**Release note:**
```release-note
`calicoctl ipam configure` now accepts `--kubevirt-ip-persistence=<Enabled|Disabled>` to toggle whether KubeVirt VMs retain a persistent IP across reboots, live migrations, and pod evictions. Pairs with the IPAM-side support added in #11865.
```

### DIFF
--- a/calicoctl/calicoctl/commands/ipam/configure.go
+++ b/calicoctl/calicoctl/commands/ipam/configure.go
@@ -32,7 +32,7 @@ import (
 func updateIPAMConfig(
 	ctx context.Context,
 	ipamClient ipamlib.Interface,
-	enabled bool,
+	strictAffinity *bool,
 	maxBlocks *int,
 	persistence *ipamlib.VMAddressPersistence,
 ) error {
@@ -41,16 +41,17 @@ func updateIPAMConfig(
 		return fmt.Errorf("error: %v", err)
 	}
 
-	// If StrictAffinity == true => an address from a block can only be assigned by
-	// host with block affinity.
-	ipamConfig.StrictAffinity = enabled
+	// Update StrictAffinity if specified.
+	if strictAffinity != nil {
+		ipamConfig.StrictAffinity = *strictAffinity
+	}
 
-	// Set MaxBlocksPerHost if specified
+	// Set MaxBlocksPerHost if specified.
 	if maxBlocks != nil {
 		ipamConfig.MaxBlocksPerHost = *maxBlocks
 	}
 
-	// Update KubeVirtVMAddressPersistence if specified
+	// Update KubeVirtVMAddressPersistence if specified.
 	if persistence != nil {
 		ipamConfig.KubeVirtVMAddressPersistence = persistence
 	}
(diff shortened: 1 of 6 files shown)


calicoctl/calicoctl/commands/ipam/configure.go
```diff
@@ -26,51 +26,89 @@ import (
 	"github.com/projectcalico/calico/calicoctl/calicoctl/commands/common"
 	"github.com/projectcalico/calico/calicoctl/calicoctl/commands/constants"
 	"github.com/projectcalico/calico/calicoctl/calicoctl/util"
-	"github.com/projectcalico/calico/libcalico-go/lib/ipam"
+	ipamlib "github.com/projectcalico/calico/libcalico-go/lib/ipam"
```
caseydavenport: Is this import alias necessary?

calicoctl/calicoctl/commands/ipam/show.go
```diff
@@ -91,12 +92,18 @@ func getBorrowedIPs(ctx context.Context, ippoolClient clientv3.IPPoolInterface,
 					if blockOwner != borrowingNode {
 						bIP := borrowedIP{block: b.CIDR.IPNet.String(), blockOwner: blockOwner, borrowingNode: borrowingNode}
 						bIP.addr = b.OrdinalToIP(i).String()
-						if _, ok := attributes.ActiveOwnerAttrs[model.IPAMBlockAttributePod]; ok {
-							bIP.allocatedTo = fmt.Sprintf("%s/%s", attributes.ActiveOwnerAttrs[model.IPAMBlockAttributeNamespace],
-								attributes.ActiveOwnerAttrs[model.IPAMBlockAttributePod])
+
+						if attributes.HandleID != nil {
+							bIP.allocatedTo = *attributes.HandleID
```
caseydavenport: Is this right? This is going to produce pretty unreadable output for most things (handle IDs for pods are not user-facing) whereas the previous logic would emit the namespace and name of the pod.

caseydavenport (changes requested):
Couple minor points / questions

calicoctl/calicoctl/commands/ipam/show.go
skoryk-oleksandr: Hmm, good catch. You are right, it would print unreadable text. Fixed it.

calicoctl/calicoctl/commands/ipam/configure.go
skoryk-oleksandr: Nope, it's not needed. I removed it.

skoryk-oleksandr:
@caseydavenport Thanks for review :+1: . I addressed your comments.

caseydavenport (approved):
LGTM - looks like a lot of this code is not covered by UTs? We should probably fix that, but doesn't need to be in this PR.

PR #11973, opened 2026-03-03
Enforce CRD CEL validation rules client-side in libcalico-go

caseydavenport:
## Description

Right now, CEL validation rules (`x-kubernetes-validations`) on our CRDs are only enforced by the Kubernetes API server. In etcd mode there's no API server in the picture, so these validations are just silently skipped — which means you can create resources that violate the CEL rules without any error.

This PR adds a new package `libcalico-go/lib/validator/cel/` that compiles the CEL rules from our embedded CRD schemas into Go-executable validators at init time, using the K8s `apiextensions-apiserver` CEL libraries. The pipeline is roughly:

1. Load all embedded CRDs via `config.AllCRDs()`
2. Convert `v1.JSONSchemaProps` → internal → `structuralschema.Structural`
3. Compile CEL validators via `celvalidation.NewValidator()` (returns nil for CRDs with no CEL rules)
4. On validation, convert typed objects to unstructured and run the compiled CEL programs

This is wired into the existing `v3.Validate()` function, so both Go struct validation and CEL validation run together and errors from both are collected into a single `ErrorValidation`. Currently `Tier` and `ClusterNetworkPolicy` are the only CRDs with CEL rules, but any future `x-kubernetes-validations` will automatically be picked up.

**Note:** This is a prototype / proof-of-concept. A few things that could be improved in follow-ups:
- Update-time validation with `oldSelf` (the current `Validate()` signature doesn't take an old object)
- MutatingAdmissionPolicy enforcement (the K8s admission machinery is too tightly coupled to reuse directly, so this would likely need a different approach)

### DIFF
--- a/libcalico-go/lib/validator/v3/crd_validation_test.go
+++ b/libcalico-go/lib/validator/v3/crd_validation_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+	"github.com/projectcalico/api/pkg/lib/numorstring"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
@@ -148,6 +149,45 @@ func TestCRDValidation_BGPPeer(t *testing.T) {
 				Spec:       apiv3.BGPPeerSpec{NextHopMode: &invalidMode},
 			},
 			errSubstr: "supported values",
+		},
+	})
+}
+
+// TestCRDValidation_BGPPeer_ASNumber tests CEL rules involving asNumber, which
+// is a uint32 Go type. Verifies the JSON round-trip in toUnstructured() produces
+// the correct int type for CEL evaluation.
+func TestCRDValidation_BGPPeer_ASNumber(t *testing.T) {
+	runCRDTests(t, []crdTestCase{
+		{
+			name: "localWorkloadSelector with asNumber passes",
+			obj: &apiv3.BGPPeer{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-peer"},
+				Spec: apiv3.BGPPeerSpec{
+					LocalWorkloadSelector: "color == 'red'",
+					ASNumber:              numorstring.ASNumber(65401),
+				},
+			},
+		},
+		{
+			name: "localWorkloadSelector without asNumber fails",
+			obj: &apiv3.BGPPeer{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-peer"},
+				Spec: apiv3.BGPPeerSpec{
+					LocalWorkloadSelector: "color == 'red'",
+				},
+			},
+			errSubstr: "asNumber is required",
+		},
+		{
+			name: "peerSelector with asNumber fails",
+			obj: &apiv3.BGPPeer{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-peer"},
+				Spec: apiv3.BGPPeerSpec{
+					PeerSelector: "has(label)",
+					ASNumber:     numorstring.ASNumber(65401),
+				},
+			},
+			errSubstr: "asNumber must be empty",
 		},
 	})
 }
--- a/libcalico-go/lib/validator/v3/crd_validation.go
+++ b/libcalico-go/lib/validator/v3/crd_validation.go
@@ -16,6 +16,7 @@ package v3
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 
@@ -185,10 +186,50 @@ func toUnstructured(obj runtime.Object) (any, error) {
 		return u.Object, nil
 	}
 
-	// Convert typed object to unstructured map.
-	data, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	// Use JSON round-trip instead of runtime.DefaultUnstructuredConverter so
+	// that numeric types match what the API server sees. The converter uses
+	// reflection and preserves Go types (e.g. uint32 → uint64), but the API
+	// server receives JSON where all numbers decode as float64 or int64.
+	// Without this, CEL rules fail at runtime with type mismatches like
+	// "expected int, got uint64" for fields such as numorstring.ASNumber.
+	raw, err := json.Marshal(obj)
 	if err != nil {
 		return nil, err
 	}
+	var data map[string]any
+	if err := json.Unmarshal(raw, &data); err != nil {
+		return nil, err
+	}
+	// encoding/json decodes all numbers as float64 into interface{}, but CEL
+	// expects integer-typed fields as int64. Convert exact-integer float64
+	// values to int64, matching the k8s API server's internal representation.
+	normalizeNumbers(data)
 	return data, nil
+}
+
+// normalizeNumbers recursively walks an unstructured map and converts float64
+// values that represent exact integers to int64.
+func normalizeNumbers(v any) {
+	switch val := v.(type) {
+	case map[string]any:
+		for k, elem := range val {
+			if f, ok := elem.(float64); ok {
+				if f == float64(int64(f)) {
+					val[k] = int64(f)
+				}
+			} else {
+				normalizeNumbers(elem)
+			}
+		}
+	case []any:
+		for i, elem := range val {
+			if f, ok := elem.(float64); ok {
+				if f == float64(int64(f)) {
+					val[i] = int64(f)
+				}
+			} else {
+				normalizeNumbers(elem)
+			}
+		}
+	}
 }


libcalico-go/lib/validator/cel/cel.go
```diff
@@ -0,0 +1,168 @@
...
+		version := crd.Spec.Versions[0]
+		if version.Schema == nil || version.Schema.OpenAPIV3Schema == nil {
+			continue
+		}
+
+		// Convert v1 JSONSchemaProps to internal.
+		internalSchema := &apiextensions.JSONSchemaProps{}
+		if err := apiextensionsv1.Convert_v1_JSONSchemaProps_To_apiextensions_JSONSchemaProps(
+			version.Schema.OpenAPIV3Schema, internalSchema, nil,
+		); err != nil {
```
mazdakn: nit: since it's a multi line condition, IMO it would be more readable to check err in a separate line.

libcalico-go/lib/validator/cel/cel.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		// Convert to structural schema.
+		structural, err := structuralschema.NewStructural(internalSchema)
+		if err != nil {
+			log.WithError(err).WithField("kind", kind).Warn("Failed to create structural schema from CRD")
+			continue
+		}
+
+		// Compile CEL validator. Returns nil if no x-kubernetes-validations exist.
```
mazdakn: how about other cel validations like: 
```
// +kubebuilder:validation:MinItems=1
```
Is there anyway to validate those?

libcalico-go/lib/validator/cel/cel.go
```diff
@@ -0,0 +1,168 @@
...
+// limitations under the License.
+
+package cel
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	log "github.com/sirupsen/logrus"
```
mazdakn: nit: let's use `logrus` instead of `log`

libcalico-go/lib/validator/cel/cel.go
```diff
@@ -0,0 +1,168 @@
...
+
+// validators maps CRD Kind to its compiled CEL validator.
+// Populated once via initValidators.
+var validators map[string]*celvalidation.Validator
+
+// schemas maps CRD Kind to its structural schema (needed by Validate signature but unused internally).
+var schemas map[string]*structuralschema.Structural
+
+var initOnce sync.Once
+var initErr error
```
mazdakn: nit: cleaner to use a single var block.

libcalico-go/lib/validator/cel/cel.go
```diff
@@ -0,0 +1,168 @@
...
+
+// initValidators loads all embedded CRDs, converts their schemas, and compiles
+// CEL validators for any that contain x-kubernetes-validations rules.
+func initValidators() {
```
mazdakn: why not using an `init` function instead? any reason for going with this approach?

libcalico-go/lib/validator/v3/cel_integration_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	tests := []struct {
+		name      string
+		tier      *apiv3.Tier
+		expectErr bool
+		errSubstr string
+	}{
+		{
+			name: "default tier with Deny passes",
+			tier: func() *apiv3.Tier {
```
mazdakn: nit: it's a bit unnecessary that a `func()` call is embedded to return a tier resource.

libcalico-go/lib/validator/v3/cel_integration_test.go
```diff
@@ -0,0 +1,189 @@
...
+					t.Fatal("expected validation error, got nil")
+				}
+				verr, ok := err.(errors.ErrorValidation)
+				if !ok {
+					t.Fatalf("expected ErrorValidation, got %T: %v", err, err)
+				}
+				// Verify that CEL error details are included.
+				found := false
+				for _, f := range verr.ErroredFields {
+					if contains(f.Reason, tt.errSubstr) {
```
mazdakn: We can just use starings.Contains instead.

libcalico-go/lib/validator/v3/cel_integration_test.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func searchSubstring(s, substr string) bool {
+	for i := 0; i <= len(s)-len(substr); i++ {
+		if s[i:i+len(substr)] == substr {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
mazdakn: yeap. this is valid.

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -145,10 +147,34 @@ var (
...
 	// individual fields are valid format.
 	if err := validate.Struct(current); err != nil {
-		return convertError(err)
+		verr = convertError(err)
+	}
+
+	// Run CRD validation rules (OpenAPI schema constraints + CEL
+	// x-kubernetes-validations). In Kubernetes datastore mode, the API server
+	// enforces these. In etcd mode there is no API server, so we enforce them here.
+	if rObj, ok := current.(runtime.Object); ok {
```
caseydavenport, line 161: Wonder if there is any time that this won't be a runtime.Object? Should we just require the `current` to be a runtime.Object in the function args?

libcalico-go/lib/validator/v3/crd_validation_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+// TestCRDValidation_Tier tests CEL x-kubernetes-validations on Tier resources.
+func TestCRDValidation_Tier(t *testing.T) {
+	defaultOrder := apiv3.DefaultTierOrder
+	kubeAdminOrder := apiv3.KubeAdminTierOrder
+	customOrder := float64(100)
+	deny := apiv3.Deny
+	pass := apiv3.Pass
+
+	runCRDTests(t, []crdTestCase{
```
caseydavenport, line 76: Would prefer we split the function call from the testcase slice definition.

libcalico-go/lib/validator/v3/crd_validation_test.go
```diff
@@ -0,0 +1,231 @@
...
+				ObjectMeta: metav1.ObjectMeta{Name: "default"},
+				Spec:       apiv3.TierSpec{Order: &defaultOrder, DefaultAction: &pass},
+			},
+			errSubstr: "default",
+		},
+	})
+}
+
+// TestCRDValidation_BGPPeer tests OpenAPI schema enum constraints on BGPPeer.
+func TestCRDValidation_BGPPeer(t *testing.T) {
```
caseydavenport, line 133: We should make sure we have test cases in this style for every CRD validation we currently support.

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -145,10 +147,34 @@ var (
...
+		if crdErrs := validateCRD(context.Background(), rObj, nil); len(crdErrs) > 0 {
```
caseydavenport, line 162: This supports an old object - do we ever get on in the libcalico validator?

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -145,10 +148,34 @@ var (
...
+				if name == "" || name == "<nil>" {
+					name = rObj.GetObjectKind().GroupVersionKind().Kind
+				}
+				verr.ErroredFields = append(verr.ErroredFields, errors.ErroredField{
+					Name:   name,
+					Reason: e.Detail,
+				})
+			}
+		}
+	}
```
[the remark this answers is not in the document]
caseydavenport: Done - we now fall back to `reflect.TypeOf(obj).Elem().Name()` which gives us the Go struct name (matches CRD Kind for all Calico types). Added a test case without TypeMeta to verify.

libcalico-go/lib/validator/cel/cel.go
```diff
@@ -0,0 +1,168 @@
...
+	initOnce.Do(initValidators)
+	if initErr != nil {
+		return field.ErrorList{field.InternalError(nil, initErr)}
+	}
+
+	// Determine Kind from the object.
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	kind := gvk.Kind
+	if kind == "" {
+		return nil
```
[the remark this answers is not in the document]
caseydavenport: Fixed - same approach as above. Kind is now inferred from the Go type when TypeMeta isn't set.

libcalico-go/lib/validator/cel/cel.go
```diff
@@ -0,0 +1,168 @@
...
+
+	for _, crd := range crds {
+		kind := crd.Spec.Names.Kind
+		if len(crd.Spec.Versions) == 0 {
+			continue
+		}
+
+		// Use the first (storage) version's schema.
+		version := crd.Spec.Versions[0]
+		if version.Schema == nil || version.Schema.OpenAPIV3Schema == nil {
```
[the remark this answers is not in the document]
caseydavenport: Done - we now iterate to find the version with `Storage: true`, falling back to `Versions[0]` if none is marked.

libcalico-go/lib/validator/v3/cel_integration_test.go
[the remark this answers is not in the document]
caseydavenport: Done, switched to `strings.Contains`.

libcalico-go/lib/validator/v3/cel_integration_test.go
```diff
@@ -0,0 +1,189 @@
...
+		{
+			name: "default tier with Deny passes",
+			tier: func() *apiv3.Tier {
+				deny := apiv3.Deny
+				return &apiv3.Tier{
+					TypeMeta:   metav1.TypeMeta{Kind: "Tier", APIVersion: "projectcalico.org/v3"},
+					ObjectMeta: metav1.ObjectMeta{Name: "default"},
+					Spec:       apiv3.TierSpec{Order: &defaultOrder, DefaultAction: &deny},
+				}
+			}(),
```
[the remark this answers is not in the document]
caseydavenport: Added a test case without TypeMeta - confirms CEL enforcement still works via the reflect-based Kind inference.

libcalico-go/lib/validator/cel/cel.go
caseydavenport: Done.

libcalico-go/lib/validator/cel/cel.go
caseydavenport: Yep, good call. I've added OpenAPI schema validation alongside CEL - this covers enum, minItems, maxItems, maxLength, pattern, etc. from the CRD schemas. Same `init()` path loads both validators. I've also started removing the duplicate Go struct validators that are now covered by the CRD schema (action, ipipMode, vxlanMode, logLevel, etc.) in a follow-up branch.

libcalico-go/lib/validator/cel/cel.go
caseydavenport: Done.

libcalico-go/lib/validator/cel/cel.go
caseydavenport: Done.

libcalico-go/lib/validator/cel/cel.go
caseydavenport: No strong reason - switched to `init()`.

libcalico-go/lib/validator/v3/cel_integration_test.go
caseydavenport: Done - removed the func() wrappers, structs are built inline now.

libcalico-go/lib/validator/v3/cel_integration_test.go
caseydavenport: Done.

libcalico-go/lib/validator/v3/validator.go
caseydavenport: Tracked in CORE-12504 — will change the signature to take `runtime.Object` directly.

libcalico-go/lib/validator/v3/crd_validation_test.go
caseydavenport: Done in latest push — split the `runCRDTests` call from the test case slice.

libcalico-go/lib/validator/v3/crd_validation_test.go
caseydavenport: Added BGPPeer (enum) and IPPool (enum) test cases alongside the existing Tier (CEL) ones. More can be added as we expand coverage — tracked in CORE-12504.

libcalico-go/lib/validator/v3/validator.go
caseydavenport: No, the libcalico validator only ever validates on create — never has the old object. Removed the `oldObj` parameter from `validateCRD`, it always passes nil anyway.

marvin-tigera:

PR #11977, opened 2026-03-03
Add support for selector-scoped FelixConfiguration resources

tomastigera:
## Description

This PR adds support for selector-scoped FelixConfiguration resources, allowing Felix configuration to be applied conditionally to nodes based on label selectors. This enables more granular configuration management in heterogeneous clusters where different node types may require different Felix settings.

### Key Changes

1. **ConfigBatcher Enhancement** (`felix/calc/config_batcher.go`):
   - Added tracking of local node labels via Node resource updates
   - Added storage and evaluation of selector-scoped FelixConfiguration resources
   - Implemented selector matching logic to determine which configs apply to the local node
   - Added `ExtractConfigFromFelixSpec()` function to extract config key-value pairs from FelixConfigurationSpec using reflection, sharing the same special-case value converters as the configUpdateProcessor (ProtoPort, RouteTableRange, HealthTimeoutOverrides, BPFConntrackTimeouts)
   - Modified `OnConfigUpdate` callback signature to include a new `selectorConfig` parameter
   - Annotation-based config overrides (`config.projectcalico.org/*`) are supported on selector-scoped resources, consistent with default/per-node handling

2. **Startup Configuration Loading** (`felix/daemon/daemon.go`):
   - Added `loadSelectorScopedFelixConfig()` function to load and evaluate selector-scoped configs during startup
   - Integrated selector config loading into the initial configuration bootstrap process
   - Added `DatastorePerSelector` source to the configuration priority chain

3. **Configuration Processor Updates** (`libcalico-go/lib/backend/syncersv1/updateprocessors/configurationprocessor.go`):
   - Modified to skip fields tagged with `confignamev1:"-"` (like NodeSelector)
   - Added pass-through handling for selector-scoped FelixConfiguration resources (those not named "default" or "node.*")
   - These resources are now passed to ConfigBatcher as raw v3 resources rather than being decomposed into v1 config KVPairs

4. **API Validation** (`libcalico-go/lib/validator/v3/validator.go`):
   - Added validation to ensure NodeSelector is only set on selector-scoped FelixConfiguration resources
   - Prevents NodeSelector from being set on "default" or "node.*" named resources

5. **FelixConfigurationSpec** (`api/pkg/apis/projectcalico/v3/felixconfig.go`):
   - Added `NodeSelector` field to FelixConfigurationSpec with documentation on usage and semantics

6. **Event Sequencer** (`felix/calc/event_sequencer.go`):
   - Updated to handle the new `selectorConfig` parameter in config updates

### Configuration Priority

The configuration priority order (highest to lowest) is now:
1. Internal overrides
2. Environment variables
3. Config file
4. Datastore per-host
5. **Datastore per-selector** (new)
6. Datastore global
7. Defaults

### Overlapping Selector-Scoped Configs

At most one selector-scoped FelixConfiguration should match any given node. If multiple selector-scoped resources match the same node, this is treated as a **misconfiguration**: a warning is logged and all selector-scoped config is ignored, falling back to default + per-host config only.

This keeps the semantics simple and avoids ambiguous merge ordering. The intended use case is distinct, non-overlapping groups of nodes (e.g., control-plane vs. worker nodes), not layered config composition.

### Testing

Added comprehensive unit tests in `felix/calc/config_batcher_test.go` covering:
- Matching selector-scoped FelixConfiguration resources
- Non-matching selector-scoped resources
- Label changes causing config re-evaluation
- Deletion of selector-scoped resources
- Ignoring node updates for other hosts
- Multiple overlapping selectors treated as misconfiguration (all selector config ignored)
- Invalid selector strings (ignored)
- `ExtractConfigFromFelixSpec` with complex field types (ProtoPort, RouteTableRange, RouteTableRanges, HealthTimeoutOverrides, Duration, NodeSelector skipping)

**Release note:**
```release-note
[Tech Preview] FelixConfiguration resources now support an optional `nodeSelector` field that restricts configuration to nodes matching a label selector, enabling per-node-group Felix configuration without requiring individual per-node resources. At most one selector-scoped FelixConfiguration should match any given node; if multiple match, the oldest by creation time wins to avoid disrupting existing working configuration. Overlapping selectors are treated as a misconfiguration and this behavior may change in future releases.
```


nelljerram:
@tomastigera Just thinking of the API here; I haven't looked at the implementation yet.

First of all, great!  I've thought for a while that we could have different instances of FelixConfiguration for different kinds of node.  For example, if we already had this capability, we wouldn't need all the "Windows..." fields [here](https://github.com/tigera/calico-private/blob/master/api/pkg/apis/projectcalico/v3/felixconfig.go#L1142-L1158).  Instead we could just use the corresponding non-Windows fields in a FelixConfig with a nodeSelector identifying Windows nodes.

Secondly, have you already proposed and agreed this at design/API level, and can you link to that if so?  I have two particular queries:

1. What happens when there are multiple FelixConfigurations applying to the same node?  This will need to be well-defined, and documented.

2. Technically it would be nice to evolve our existing conventions to use the new nodeSelector API.  In other words:

   - Currently FelixConfig with `name: default` applies to all nodes.  Could evolve that to `nodeSelector: "all()"`.
   - Currently FelixConfig with `name: node.<nodename>` applies to just `<nodename>`.  Could evolve that to a nodeSelector that selects just that node.

   But I suppose we have to keep the existing conventions for back-compatibility.  So perhaps we just grandfather in the new implementation, by converting those naming conventions to selectors.  What are your thoughts on this?

tomastigera:
> @tomastigera Just thinking of the API here; I haven't looked at the implementation yet.
> 
> First of all, great! I've thought for a while that we could have different instances of FelixConfiguration for different kinds of node. For example, if we already had this capability, we wouldn't need all the "Windows..." fields [here](https://github.com/tigera/calico-private/blob/master/api/pkg/apis/projectcalico/v3/felixconfig.go#L1142-L1158). Instead we could just use the corresponding non-Windows fields in a FelixConfig with a nodeSelector identifying Windows nodes.
> 
> Secondly, have you already proposed and agreed this at design/API level, and can you link to that if so? I have two particular queries:

This was the original plan reviewed by @fasaxc in this branch, got removed, so I resurrected is and placed it here https://github.com/tigera/designs/pull/43

> 1. What happens when there are multiple FelixConfigurations applying to the same node?  This will need to be well-defined, and documented.

Yeah, that is the right question, I updated the description, definitely needs doc update.

Overlapping Selector-Scoped Configs
When multiple selector-scoped FelixConfiguration resources match the same node, their fields are merged in alphabetical order by resource name. If two matching resources set the same field, the value from the last resource in alphabetical order wins.

> 2. Technically it would be nice to evolve our existing conventions to use the new nodeSelector API.  In other words:
>    
>    * Currently FelixConfig with `name: default` applies to all nodes.  Could evolve that to `nodeSelector: "all()"`.
>    * Currently FelixConfig with `name: node.<nodename>` applies to just `<nodename>`.  Could evolve that to a nodeSelector that selects just that node.

I kept full backwards compatibility. I like that the default just applies to all. Technically we can eventually deprecate the "per-node" variant since that is covered by the node selector.

The motivation for this was to say configure control nodes differently than worker nodes - real world usecase.

caseydavenport:
> When multiple selector-scoped FelixConfiguration resources match the same node, their fields are merged in alphabetical order by resource name. If two matching resources set the same field, the value from the last resource in alphabetical order wins.

Something we have done recently for IP pools is to use an external controller to spot these conflicts and mark one as "blessed" via status - then Felix would just ignore ones that don't have the status.

Is there a strong use-case for the merging behavior? It sounds complicated / confusing / easy to get wrong, so enforcing "at most one felix config per node" seems like a good win?

Keeping "default can be overridden by a singular per-node / selector" seems reasonable and intuitive, though.

nelljerram:
> > When multiple selector-scoped FelixConfiguration resources match the same node, their fields are merged in alphabetical order by resource name. If two matching resources set the same field, the value from the last resource in alphabetical order wins.
> 
> Something we have done recently for IP pools is to use an external controller to spot these conflicts and mark one as "blessed" via status - then Felix would just ignore ones that don't have the status.
> 
> Is there a strong use-case for the merging behavior? It sounds complicated / confusing / easy to get wrong, so enforcing "at most one felix config per node" seems like a good win?
> 
> Keeping "default can be overridden by a singular per-node / selector" seems reasonable and intuitive, though.

FWIW that was my thinking too.  It feels like it should suffice for most use cases to have default + one more specific FelixConfig, applying to a single given node.  And if the selectors result in more FelixConfigs applying than that, we can treat that as an error.

tomastigera:
> Is there a strong use-case for the merging behavior? It sounds complicated / confusing / easy to get wrong, so enforcing "at most one felix config per node" seems like a good win?

tbh I would really treat it as a misconfiguration, reject node-selector configs and go with default if that is acceptable. I do not see a usecase for overlapping nodeselectors unless we can properly order the hierarchy, but that is outside of the scope of this work. The usecase is that you have distinct non-overlapping groups of nodes. The only overlap is with default and per-node as they are defined atm and there is strict decision on who wins.

So if you are fine with that, I can just reject overlapping node-selector based configs. Or any other suggestions?

nelljerram:
> > Is there a strong use-case for the merging behavior? It sounds complicated / confusing / easy to get wrong, so enforcing "at most one felix config per node" seems like a good win?
> 
> tbh I would really treat it as a misconfiguration, reject node-selector configs and go with default if that is acceptable. I do not see a usecase for overlapping nodeselectors unless we can properly order the hierarchy, but that is outside of the scope of this work. The usecase is that you have distinct non-overlapping groups of nodes. The only overlap is with default and per-node as they are defined atm and there is strict decision on who wins.
> 
> So if you are fine with that, I can just reject overlapping node-selector based configs. Or any other suggestions?

Yes, it sounds like we're all thinking the same thing here - @caseydavenport can you confirm?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -168,6 +168,20 @@ const (
 // +kubebuilder:validation:XValidation:rule="!has(self.deviceRouteSourceAddress) || size(self.deviceRouteSourceAddress) == 0 || (isIP(self.deviceRouteSourceAddress) && ip(self.deviceRouteSourceAddress).family() == 4)",message="deviceRouteSourceAddress must be a valid IPv4 address",reason=FieldValueInvalid
 // +kubebuilder:validation:XValidation:rule="!has(self.deviceRouteSourceAddressIPv6) || size(self.deviceRouteSourceAddressIPv6) == 0 || (isIP(self.deviceRouteSourceAddressIPv6) && ip(self.deviceRouteSourceAddressIPv6).family() == 6)",message="deviceRouteSourceAddressIPv6 must be a valid IPv6 address",reason=FieldValueInvalid
 type FelixConfigurationSpec struct {
+	// NodeSelector is an optional label selector that restricts this FelixConfiguration
+	// to only apply to nodes that match the given selector. This field is only valid
+	// on FelixConfiguration resources whose name is not "default" and does not start
+	// with "node.". For resources named "default", the configuration applies globally
+	// to all nodes. For resources named "node.<nodename>", the configuration applies to
+	// the named node only.
```
fasaxc:
```
This field is only valid
	// on FelixConfiguration resources whose name is not "default" and does not start
	// with "node."
```
Enforce with CEL?

felix/calc/config_batcher.go
```diff
@@ -117,9 +255,123 @@ func (cb *ConfigBatcher) maybeSendCachedConfig() {
...
+	}
+}
+
+// AnnotationConfigPrefix is the prefix for config override annotations on
+// FelixConfiguration resources. Must match the prefix in configurationprocessor.go.
+const AnnotationConfigPrefix = "config.projectcalico.org/"
+
+// ExtractConfigFromFelixSpec extracts the configuration key-value pairs from a
+// FelixConfigurationSpec using reflection. This mirrors the logic in the
+// configUpdateProcessor but produces a simple map rather than v1-model KVPairs.
```
fasaxc: Lets move this to library function and share it with the configUpdateProcessor.

felix/calc/config_batcher.go
```diff
@@ -117,9 +255,123 @@ func (cb *ConfigBatcher) maybeSendCachedConfig() {
...
-	cb.callbacks.OnConfigUpdate(globalConfigCopy, hostConfigCopy)
+	cb.callbacks.OnConfigUpdate(globalConfigCopy, selectorConfigMerged, hostConfigCopy)
 	cb.configDirty = false
 }
+
+// mergeMatchingSelectorConfigs evaluates all selector-scoped FelixConfiguration
+// resources against the local node's labels. If exactly one resource matches,
+// its config is returned. If multiple resources match, this is treated as a
+// misconfiguration: a warning is logged and an empty map is returned (falling
+// back to default + per-host config only).
```
fasaxc: Not sure about the tie-breaker being "ignore both".  I might have either:
* Sorted on name and taken first match.
* Sorted on name and applied the union of all matches by applying them all to a map in sequence so later ones overwrite earlier.

felix/daemon/daemon.go
```diff
@@ -977,6 +996,115 @@ func loadConfigFromDatastore(
...
+	felixConfigs, err := client.List(ctx, model.ResourceListOptions{
+		Kind: apiv3.KindFelixConfiguration,
+	}, "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to list FelixConfiguration resources: %w", err)
+	}
+
+	// Collect matching selector-scoped configs. If more than one matches,
+	// treat it as a misconfiguration and return empty (fall back to
+	// default + per-host config only).
```
fasaxc: Looks like this logic could be shared with the calc graph piece, it's complex enough to want to avoid duplication.

libcalico-go/lib/validator/v3/validator.go
```diff
@@ -889,6 +890,26 @@ func validateFelixConfigSpec(structLevel validator.StructLevel) {
 	}
 }
 
+func validateFelixConfiguration(structLevel validator.StructLevel) {
```
caseydavenport: We shouldn't be adding new validation to the validator package at this point - we should add CEL XValidation rules to the underlying CRD.

The validation code in Golang imports those CEL rules and will execute them in the etcd / apiserver code path, but without them we don't get any validation in the v3 CRD mode.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -168,6 +168,20 @@ const (
 // +kubebuilder:validation:XValidation:rule="!has(self.deviceRouteSourceAddress) || size(self.deviceRouteSourceAddress) == 0 || (isIP(self.deviceRouteSourceAddress) && ip(self.deviceRouteSourceAddress).family() == 4)",message="deviceRouteSourceAddress must be a valid IPv4 address",reason=FieldValueInvalid
 // +kubebuilder:validation:XValidation:rule="!has(self.deviceRouteSourceAddressIPv6) || size(self.deviceRouteSourceAddressIPv6) == 0 || (isIP(self.deviceRouteSourceAddressIPv6) && ip(self.deviceRouteSourceAddressIPv6).family() == 6)",message="deviceRouteSourceAddressIPv6 must be a valid IPv6 address",reason=FieldValueInvalid
 type FelixConfigurationSpec struct {
+	// NodeSelector is an optional label selector that restricts this FelixConfiguration
+	// to only apply to nodes that match the given selector. This field is only valid
```
caseydavenport:
```suggestion
	// to apply only to nodes that match the given selector. This field is only valid
```

nit, but I think this is the meaning you intend :)

fasaxc (changes requested):
Gave it a quick skim.

Would be nice to handle parsing of all three groups with common code, perhaps we pass all felix configs to the calc graph and generate the KVs there?

I don't see it emitting the KV pair version? That might affect VPP or Dikastes, which may still use that.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -168,6 +168,20 @@ const (
...
+	// with "node.". For resources named "default", the configuration applies globally
+	// to all nodes. For resources named "node.<nodename>", the configuration applies to
+	// the named node only.
+	//
+	// At most one selector-scoped FelixConfiguration should match any given node.
+	// If multiple selector-scoped resources match, this is treated as a
+	// misconfiguration: all selector-scoped config is ignored and the node
+	// falls back to the default and per-host configuration only.
+	// +optional
+	NodeSelector string `json:"nodeSelector,omitempty" validate:"omitempty,selector" confignamev1:"-"`
```
caseydavenport:
```suggestion
	NodeSelector *string `json:"nodeSelector,omitempty"`
```

Optional fields should be pointers. 
We can get rid of the "validate" tag once we're using CEL instead of Go code validation.
I think confignamev1 is unused at this point?

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -168,6 +168,20 @@ const (
...
+	// on FelixConfiguration resources whose name is not "default" and does not start
+	// with "node.". For resources named "default", the configuration applies globally
+	// to all nodes. For resources named "node.<nodename>", the configuration applies to
+	// the named node only.
+	//
+	// At most one selector-scoped FelixConfiguration should match any given node.
+	// If multiple selector-scoped resources match, this is treated as a
+	// misconfiguration: all selector-scoped config is ignored and the node
+	// falls back to the default and per-host configuration only.
+	// +optional
```
caseydavenport, line 184: We should add a MaxLengh kubebuilder tag here for a reasonable size to avoid exploding CEL complexity calculations (think we do this for other selectors like network policy if you want to look for an example of this).

caseydavenport (commented):
First pass, but still looking.

caseydavenport:
I do wonder if the merging logic really belongs in Felix though - I was imagining something more along the lines of a controller that can:

- See all FelixConfigurations
- Resolve conflicts
- Updates a status field with the resolution (e.g., "This is invalid due to overlap" or "This is active on node(s) X" 

Then the felix code is simply "Does this FelixConfiguration say it is valid, and does it match me" 

It is a larger architectural change though, and could be done in a separate pass so not too concerned about doing it here. But I think having the reporting back via `kubectl get` is actually important for this feature. 

You can see how we do it for v3.IPPool, where we also use an additional report column in the CRD to explicitly include detailed conflict information in the output from `kubectl`.

felix/calc/config_batcher.go
caseydavenport: We solve this in the IP pool controller by sorting on a few different fields, CreationTimestamp being an important one - it means if you accidentally create a conflicting object, you don't invalidate the existing state that may have been working for you!

tomastigera: We would still have the default config. If anything does not work, it should be clear from logs, right? It is essentially misconfiguration and how would we tell the user about it :thinking:

tomastigera:
## Manual Test Results — Selector-Scoped FelixConfiguration

Tested on a 4-node GCP cluster (1 control-plane + 3 workers, K8s v1.34.6, operator-managed) by building and deploying custom `calico/node`, `calico/typha`, and `calico/apiserver` images from this branch.

### Setup
- Built images from branch `claude/felix-config-node-labels-03U9P` (commit `f9d794f826`)
- Applied updated `crd.projectcalico.org` CRD with `nodeSelector` field
- Deployed to cluster via daemonset/deployment image patches (operator reconciliation paused via `unsupported.operator.tigera.io/ignore` annotation)

### Test 1: CRD Field Present
- Verified `nodeSelector` field exists in the `felixconfigurations.crd.projectcalico.org` CRD schema (type: string, maxLength: 1024)
- **PASS**

### Test 2: Basic Selector Match
- Labeled `node-0` and `node-1` with `test-role=gpu`
- Created `gpu-config` FelixConfiguration with `nodeSelector: "test-role == 'gpu'"` and `logSeverityScreen: Debug`
- Felix on `node-0` and `node-1` switched to DEBUG log level
- Felix on `node-2` and control-plane remained at INFO
- **PASS**

### Test 3: Conflict Resolution (Oldest Wins)
- Created `gpu-config-2` with same selector but `logSeverityScreen: Warning`
- Felix logged: `Multiple selector-scoped FelixConfigurations match this node; using the oldest by creation time. This is likely a misconfiguration. matching=["gpu-config-2", "gpu-config"] winner="gpu-config"`
- Felix remained at DEBUG (from `gpu-config`, the older resource)
- **PASS**

### Test 4: Dynamic Label Change
- Removed `test-role=gpu` from `node-1`
- Felix on `node-1` detected label change via `config_batcher.go:232: Local node labels changed`, re-evaluated selectors, and reverted to INFO (`selector=map[string]string{}`)
- **PASS**

### Test 5: Config Deletion
- Deleted both `gpu-config` and `gpu-config-2`
- Felix on all nodes reverted to global config only (`LogSeverityScreen: Info`)
- `config_batcher.go:160: Selector-scoped FelixConfiguration deleted` logged for each
- **PASS**

### Test 6: CEL Validation
- Attempted to set `nodeSelector` on the `default` FelixConfiguration via both v3 and v1 APIs
- **NOT ENFORCED** — the request succeeded on both paths. The Go struct validator was removed in favor of CEL, but the CEL rule (`self.metadata.name == 'default' || self.metadata.name.startsWith('node.') ? !has(self.spec.nodeSelector) : true`) did not land in the `crd.projectcalico.org` CRD, and the calico-apiserver (AA layer) doesn't enforce CRD-level CEL rules.
- **NEEDS FOLLOW-UP** — either restore a Go-based check in the apiserver validation path, or ensure the CEL rule is generated into the v1 CRD schema.

### Summary
The core feature works correctly: selector matching, config precedence (global < selector < per-node), conflict resolution via CreationTimestamp, dynamic label re-evaluation, and config deletion all behave as expected. The only gap is CEL validation enforcement for preventing `nodeSelector` on `default`/`node.*` resources.

tomastigera:
## Update: CEL Validation Now Enforced

The previous test run found that nodeSelector validation on `default`/`node.*` resources was not enforced on either API path. This is now fixed in commits `8bcb5e1b0c` and `d3ad34099e`.

### What was added
- **v3 API** (calico-apiserver): Go validation in the strategy's `Validate()`/`ValidateUpdate()` methods
- **v1 CRD** (`crd.projectcalico.org`): CEL XValidation rule on the v1 FelixConfiguration type (previously only on v3)

### Re-test results
```
Test 2 - nodeSelector on "default" via v3 API:
  The FelixConfiguration "default" is invalid: spec.nodeSelector: Forbidden:
  nodeSelector must not be set on the 'default' or per-node ('node.*') FelixConfiguration
  PASS (rejected)

Test 3 - nodeSelector on "node.test" via v3 API:
  The FelixConfiguration "node.test" is invalid: spec.nodeSelector: Forbidden:
  nodeSelector must not be set on the 'default' or per-node ('node.*') FelixConfiguration
  PASS (rejected)

Selector-scoped resource "gpu-config" with nodeSelector: PASS (accepted)
```

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -168,6 +169,22 @@ const (
...
 type FelixConfigurationSpec struct {
+	// NodeSelector is an optional label selector that restricts this FelixConfiguration
+	// to apply only to nodes that match the given selector. This field is only valid
+	// on FelixConfiguration resources whose name is not "default" and does not start
+	// with "node.". For resources named "default", the configuration applies globally
+	// to all nodes. For resources named "node.<nodename>", the configuration applies to
+	// the named node only.
+	//
+	// At most one selector-scoped FelixConfiguration should match any given node.
+	// If multiple selector-scoped resources match, the oldest (by creation
```
fasaxc, line 180: Not keen on this tie-breaker.  Consider git-ops and disaster recovery; nothing guarantees that resources get re-applied in the same order.  Also, why forbid something that is useful?  It'd useful to be able to say "apply config X to nodes in region Y" and "override with config Z for nodes in rack P"

felix/daemon/daemon.go
```diff
@@ -977,6 +997,128 @@ func loadConfigFromDatastore(
...
+			for i, m := range matches {
+				names[i] = m.name
+			}
+			log.WithFields(log.Fields{
+				"matching": names,
+				"winner":   winner.name,
+			}).Warn("Multiple selector-scoped FelixConfigurations match this node at startup; " +
+				"using the oldest by creation time. This is likely a misconfiguration.")
+		}
+		maps.Copy(result, winner.config)
```
fasaxc: Still feels like too much duplication between here and calc graph.  One tweak to the merging algorithm and it could cause a felix crash loop.

fasaxc (dismissed):

api/pkg/apis/projectcalico/v3/felixconfig.go
caseydavenport: I think it is worth keeping the overlap / nesting restricted in the first pass at least. No reason we can't expand it, but once it is out there we can't take it back. My gut tells me it is more of a footgun than a helper - you can get the same behavior with selector adjustments (although not quite as easily). 

The timestamp + gitops interaction is a good call out for fresh starts, but I think still useful in the more applicable "error case" of accidentally checking in something new after the cluster has been running for a while.

felix/daemon/daemon.go
tomastigera: Agreed, I'll deduplicate this. Plan:

1. Export `SelectorConfigEntry` and extract a shared `MergeSelectorConfigs(configs []*SelectorConfigEntry, nodeLabels map[string]string) map[string]string` function in `felix/calc/`.
2. Both `config_batcher.go:mergeMatchingSelectorConfigs()` and `daemon.go:loadSelectorScopedFelixConfig()` call through to that shared function.
3. This eliminates the duplicate selector evaluation, merge/conflict resolution, and warning logging.

caseydavenport (approved):

PR #11979, opened 2026-03-03
Migrate remaining QoS e2e tests from k8s-e2e

caseydavenport:
Migrates the two remaining QoS test cases from k8s-e2e that weren't covered by the initial bandwidth test migration:

- **Combined ingress+egress bandwidth test** -- deploys a pod with both `ingressBandwidth` and `egressBandwidth` annotations and verifies both directions are limited independently.
- **Packet rate limit test** -- tests `ingressPacketRate` and `egressPacketRate` annotations using UDP traffic with known packet sizes, then verifies the throughput is capped to the expected rate.

Also adds some shared helpers that these tests need:

- `iperfcheck.WithUDP()`, `WithPacketLength()`, `WithTargetBandwidth()` -- UDP mode support for the existing iperf3 tester, plus a fallback in the JSON parser for UDP's `end.sum` field (iperf3 reports UDP results differently than TCP).
- `utils.GetCalicoNodePodOnNode()`, `GetPodInterfaceName()`, `GetPodInterfaceIndex()` -- calico-node pod lookup and interface discovery helpers.
- `utils.ExecInCalicoNode()` -- kubectl exec wrapper that targets the `calico-node` container explicitly (the pod has multiple containers).

### Differences from the k8s-e2e version

The k8s-e2e packet rate test used **iperf2** for UDP testing. We use **iperf3** here with `-u -l -b` flags, which has the same capability but avoids pulling in a new container image (the existing iperf3 image already handles both TCP and UDP).

Interface lookup is now **prefix-agnostic** -- uses `ip route get <podIP>` to find the host-side veth name rather than grepping by a platform-specific prefix (`cali`, `eni`, `azv`, `gke`). Works on any platform without configuration. Interface index lookup uses `/sys/class/net/<ifname>/ifindex` rather than parsing `ip addr` output.

`GetPodInterfaceName` accepts a workload `Pod` directly and auto-detects the calico-node pod on the same node internally, so callers don't need to wire up both pieces themselves.

Dataplane detection and **low-level rule verification** (inspecting iptables chains, nftables rules, BPF map entries) were removed -- the throughput assertions alone prove the feature works e2e, and coupling the test to dataplane implementation details adds brittleness without meaningful additional signal.

No version checks -- the monorepo tests are co-versioned with the product, so we don't need the `CompareSemVer` / `ReleaseStream` gating that k8s-e2e required.


e2e/pkg/tests/networking/qos.go
```diff
@@ -136,5 +142,316 @@ var _ = describe.CalicoDescribe(
...
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to measure egress packet rate limited throughput")
+			logrus.Infof("Egress packet-rate-limited throughput (bps): %.0f", egressResult.AverageRate)
+			Expect(egressResult.AverageRate).To(BeNumerically("<=", 1000*8*100*1.2), "egress packet rate limit not effective")
 		})
 	})
+
+// verifyPacketRateRules checks that the packet rate dataplane rules are programmed
+// for the given direction (ingress or egress) on the given interface.
+func verifyPacketRateRules(
```
caseydavenport: Is this too low level for an e2e test? Do we really need this?

e2e/pkg/utils/caliconode.go
```diff
@@ -0,0 +1,95 @@
...
+			return &podList.Items[i]
+		}
+	}
+	return nil
+}
+
+// GetPodInterfaceName returns the host-side veth interface name for a workload pod
+// by running `ip route get <podIP>` in the calico-node container. This is
+// prefix-agnostic and works with cali, eni, azv, gke, or any other prefix.
+func GetPodInterfaceName(calicoNodePod *corev1.Pod, workloadPodIP string) string {
```
caseydavenport: Might be nicer if we could simply pass a Pod object (i.e., auto detect the calico/node from the pod's Node and avoid the need to pass both)

Issue #11331, opened 2025-11-06
Taint kubernetes when NetworkUnavailable=true


r2ronoha:
Calico now [waits until networking is fully established before setting the `NetworkUnavailable=false` condition.](https://github.com/projectcalico/calico/pull/10866)

Nonetheless, that is not something that the kubelet uses as condition for a node readiness.

We have the calico-node daemonset as part of the resources that need to be Ready for the node to transition to Ready itself. The problem is that, if calico-node is in a `"calico/node is not ready: BIRD is not ready: BGP not established` with loop, at this point the pod is already transitioned to Healthy, and hence the Kubelet considers it and the node as Ready. And during that time, if pods are scheduled on the node, they may fail to resolve services.

It would be great if calico could taint the nodes while `NetworkUnavailable=false`, so that workloads can be ensured to not be scheduled until networking is fully set up.

Azahorscak:
There are similar patterns in use by the[ aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/install.md) and [cilium](https://docs.cilium.io/en/latest/installation/taints/) that calico could consider adopting. 
This would only require calico to remove a taint upon healthy startup, not necessarily apply one.

caseydavenport:
Yeah, I think putting this logic in our operator would make sense - we'd need that as a third-party to manage the taint addition.

caseydavenport:
I did start this prototype a little while back that I need to come back to when I have some time: https://github.com/projectcalico/calico/pull/11985

awalker125:
If we look hard we often see sandbox errors on pod startup on new nodes. I was hoping #10866 would help with that problem. Unfortunately NetworkUnavailable=True condition only ever seems to get set if the ccm returns true in its implementation of https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L64. This is normally only the case if the underlying cloud/infra has some kind of VPC that needs programmed with the routes for new nodes before connectivity is available.

If there are no routes needed at the infra level and so the ccm returns false then workload pods will start to schedule as soon as calico writes the cni config to disk.

I've considered various hacks from setting NetworkUnavailable=True manually or applying the taint node.kubernetes.io/network-unavailable but both dont feel the right approach.

The cilium pattern of allowing a taint to be applied which is later removed would probably work well in cases with or without the route controller.

PR #11985, opened 2026-03-03
Monitor NetworkUnavailable at runtime and taint nodes until networking is ready

### LINKED ISSUE #5233: Update Kubernetes Node condition `NetworkUnavailable` in case of problems
Hi, 
we are sometimes encountering problems with `calico-node` on Kubernetes. 

**Issue Description**
The condition type `NetworkUnavailable` on a Kubernetes node seems to be only [initialised to false on startup](https://github.com/projectcalico/node/pull/89/files).
However, sometimes we are experiencing issues such as (from calico-node logs):

```
bird: Mesh_10_250_10_60: Socket error: bind: Address in use
bird: Mesh_10_250_19_13: Socket error: bind: Address in use
bird: Mesh_10_250_21_169: Socket error: bind: Address in use
```

The aforementioned node condition  is  still set to `False` with reason `CalicoIsUp` indicating that everything is working fine.

## Expected Behavior

I would expect that calico's state is reflected in the Kubernetes node condition `NetworkUnavailable` (also in case of abnormalities during runtime).

## Current Behavior

The condition type `NetworkUnavailable` on a Kubernetes node is only [initialised to false on startup](https://github.com/projectcalico/node/pull/89/files) so problem during runtime are not reflected in the node condition.

Another possibility is, that calico sets the `NetworkUnavailable` too early during startup, with the actual error happening after the initialisation.

## Context

We are creating Kubernetes node via the [Machine Controller Manager](https://github.com/gardener/machine-controller-manager) with [calcio-node as deamon set](https://github.com/gardener/gardener-extension-networking-calico/blob/master/charts/internal/calico/templates/node/daemonset-calico-node.yaml).
We are monitoring the Node conditions including the `NetworkUnavailable` to be able to replace a node in case calico has problems.

## Your Environment

* calico/node version: <TODO>
* Operating System and version: Ubuntu 18.04.4 LTS (though observed independently)

caseydavenport:
Calico sets the NetworkUnavailable node condition once at startup and once on shutdown. If calico-node crashes, gets OOMKilled, or Felix/BIRD fail at runtime, the condition goes stale and the node looks healthy when it isn't. And on clusters with no cloud route controller, the condition doesn't gate scheduling at all, so pods can land on a node before BGP is established and fail to reach services.

### Keep NetworkUnavailable accurate at runtime

kube-controllers watches calico-node pod readiness and keeps the condition in step with it, in both directions. The kubelet readiness probe already runs `calico node health --felix-ready --bird-ready`, so pod readiness is the same signal calico-node would poll for itself, and it additionally catches the crashes and OOMKills an in-pod loop can't observe.

One writer owns each transition. calico-node keeps startup and shutdown; kube-controllers owns everything at runtime. Startup has to stay on the node: new nodes are tainted at admission time and kube-controllers needs pod networking, so it can't schedule until calico-node has cleared the first node.

The controller reconciles against the node's current condition and taints rather than tracking what it wrote, so restarting it neither re-marks nor strands a node. It leaves a NetworkUnavailable condition it didn't set alone, since a cloud route controller writes the same condition and clearing that one would claim a node is routable while its routes are still missing.

Nodes with no calico-node pod are skipped rather than marked. The DaemonSet passes over Windows, virtual-kubelet and Fargate nodes on purpose, and Calico has nothing to say about their networking. calico-node pods are identified by their container name, which is the same across Calico and Canal, rather than by the `k8s-app` label, which isn't.

Two cases where Calico stops speaking for a node it previously spoke for:

- The pod disappears entirely, from an uninstall or a drained node. The DaemonSet replaces a deleted pod within seconds, so after five minutes with no pod Calico gives up the condition and the taint, rather than leaving a node it had marked permanently unschedulable. New nodes are given the same five minutes, so the admission-time taint survives normal bringup.
- Another CNI provides pod networking, read from `CALICO_NETWORKING_BACKEND` on calico-node's own container so it works in a manifest install as well as under the operator. The condition describes that CNI rather than Calico, so it is left alone and only the taint is cleared.

### Taint nodes until networking is ready (opt-in)

When enabled by the operator, Calico taints nodes with `node.projectcalico.org/network-not-ready:NoSchedule` until networking is ready. New nodes are tainted at creation via a MutatingAdmissionPolicy; calico-node removes the taint once Felix and BIRD are ready, and kube-controllers re-adds it if they fail later.

Adding the taint is gated on an environment variable, removing it is not. That matters twice:

- Disabling the feature drains the taint from nodes that already carry it, rather than leaving them permanently unschedulable with nothing left to clear it.
- The admission policy is safe in a plain manifest install, where nothing sets the variable. Nodes are tainted at registration and calico-node clears the taint once it's up; both host-networked components already tolerate every NoSchedule taint, so nothing is blocked from starting.

### RBAC

Neither service account could perform the node writes this needs. calico-kube-controllers gains `nodes/status` patch and `nodes` update; calico-node gains `nodes` update, which it had only in KDD mode.

Fixes #5233
Related: #11331

Companion operator PR: https://github.com/tigera/operator/pull/5072

Related: [CORE-13302](https://tigera.atlassian.net/browse/CORE-13302)

## AI assistance

Written with Claude Code.

```release-note
Calico now keeps the NetworkUnavailable node condition accurate at runtime, updating it when calico-node fails after startup rather than only at start and shutdown.
```

```release-note
Adds an opt-in option to taint nodes until Calico networking is ready, so that workloads are not scheduled onto a node before Calico can service them.
```

[CORE-13302]: https://tigera.atlassian.net/browse/CORE-13302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/admission/nodetaint.mutatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,33 @@
...
+        apiVersions: ["v1"]
+        operations: ["CREATE"]
+        resources: ["nodes"]
+  matchConditions:
+    # Don't add the taint if it's already there (e.g. the node was registered with it).
+    - name: taint-absent
+      expression: "!has(object.spec.taints) || !object.spec.taints.exists(t, t.key == 'node.projectcalico.org/network-not-ready')"
+  # Never block node registration if the policy fails to evaluate - a node that can't register is
+  # worse than one that schedules a little early.
+  failurePolicy: Ignore
```
fasaxc, line 25: Given that the taint is opt-in, I'd be inclined to fail closed instead of using Ignore (or to make the failurePolicy configurable).

api/admission/nodetaint.mutatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,33 @@
...
+  reinvocationPolicy: Never
+  mutations:
+    - patchType: "JSONPatch"
+      jsonPatch:
+        expression: |
+          has(object.spec.taints) ?
```
fasaxc, line 31: Ugly, but I guess it's needed?

kube-controllers/pkg/controllers/node/node_condition.go
```diff
@@ -0,0 +1,407 @@
...
+	if age := c.nowFn().Sub(node.CreationTimestamp.Time); age < c.unmanagedGracePeriod {
+		// New node. Its pod is on the way, and the admission-time taint has to survive until
+		// calico-node arrives to clear it.
+		c.workqueue.AddAfter(node.Name, c.unmanagedGracePeriod-age)
+		return nil
+	}
+
+	// The DaemonSet controller replaces a deleted pod within seconds, so a sustained absence
+	// means Calico is gone from this node - uninstalled, or the node is on its way out.
+	return c.releaseNode(ctx, node)
```
fasaxc, line 262: Surprised to see this, I thought one of the cases we wanted to cover was a controller manager being down or slow to schedule.  We can't really deal with not being installed (kube-controllers would be killed almost immediately anyway)

kube-controllers/pkg/controllers/node/node_condition.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+// releaseNode gives up whatever Calico wrote on a node it can no longer speak for. Without it an
+// uninstall would leave every node it had marked permanently unschedulable.
+func (c *nodeConditionController) releaseNode(ctx context.Context, node *v1.Node) error {
+	// Same writes as the healthy path: clear our condition, clear our taint, touch nothing else.
+	return c.markAvailable(ctx, node)
+}
```
fasaxc, line 270: Redundant func; it's name isn't obvious "release" as in untaint?

kube-controllers/pkg/controllers/node/node_condition.go
```diff
@@ -0,0 +1,407 @@
...
+			return nil, fmt.Errorf("expected *v1.Pod in the pod index, got %T", obj)
+		}
+		pods = append(pods, pod)
+	}
+	return pods, nil
+}
+
+// calicoNodePodsByNode indexes calico-node pods by the node they run on. Pods are matched on the
+// container name rather than a label, because Canal labels the same DaemonSet "canal".
+func calicoNodePodsByNode(obj any) ([]string, error) {
```
fasaxc, line 358: Weird signature?  I'd expect this method to look up the pod, but it's passed in and then converted to a slice

kube-controllers/pkg/controllers/node/node_condition.go
```diff
@@ -0,0 +1,407 @@
...
+		return c.releaseNode(ctx, node)
+	}
+
+	if anyPodReady(pods) {
+		return c.markAvailable(ctx, node)
+	}
+
+	if remaining := c.gracePeriod - c.nowFn().Sub(notReadySince(pods)); remaining > 0 {
+		log.WithField("node", nodeName).WithField("remaining", remaining).Info("No Ready calico-node pod")
+		c.workqueue.AddAfter(nodeName, remaining)
```
fasaxc, line 242: Claude spotted a race here:
* Queue the re-sched
* calico-node comes up, clears the taint itself
* kubelet hasn't probed calico-node readiness yet
* re-sched happens, kube-contrellers sees calico-node down, re-adds the taint

Likely fix: check the last transition time to see if calico-node might have just set it.

kube-controllers/pkg/controllers/node/node_condition.go
```diff
@@ -0,0 +1,407 @@
...
+	c.handleErr(c.syncNode(context.Background(), nodeName), nodeName)
+	return true
+}
+
+func (c *nodeConditionController) handleErr(err error, nodeName string) {
+	if err == nil {
+		c.workqueue.Forget(nodeName)
+		return
+	}
+	if c.workqueue.NumRequeues(nodeName) < nodeConditionMaxRetries {
```
fasaxc, line 200: Claude says we don't have any periodic resync so if we hit this condition that's it, we'll never fix the Node?

kube-controllers/pkg/converter/transformer.go
fasaxc, line 104: Claude spotted that notReadySince checks the CreationTimestamp, but you're not copying it into the transformed Pod.

kube-controllers/pkg/converter/transformer.go
```diff
@@ -23,6 +24,68 @@ import (
...
+
+func keptCalicoNodeEnv(pod *v1.Pod) []v1.EnvVar {
+	var kept []v1.EnvVar
+	for _, container := range pod.Spec.Containers {
+		if container.Name != calicoNodeContainerName {
+			continue
+		}
+		for _, env := range container.Env {
+			if slices.Contains(calicoNodeCachedEnv, env.Name) {
+				kept = append(kept, v1.EnvVar{Name: env.Name, Value: env.Value})
```
fasaxc, line 57: Note there's also ValueFrom in the struct, and we use it for CALICO_NETWORKING_BACKEND so this is lossy.

node/pkg/node/command.go
```diff
@@ -149,7 +149,8 @@ func newStartupCommand() *cobra.Command {
 			logrus.SetFormatter(&logrusr.Formatter{Component: "startup"})
 			startup.Run()
 			if completeStartup {
-				if err := startup.MarkNetworkAvailable(); err != nil {
+				if err := startup.MarkNetworkAvailable(context.Background(), 5*time.Minute); err != nil {
```
fasaxc, line 152: Claude spotted that this now waits for 5 minutes for Felix/BIRD on Windows where it previously patched the condition and returned without health checks.

I didn't look at the detail, but sounds like something we should investigate!

fasaxc (changes requested):
I'm not sure about the design of auto-releasing nodes after a grace period; I think we should fail closed unless told otherwise (maybe by looking at the calico-node daemonset selector or something?)

Claude /code-review spotted several issues too.

PR #9804, opened 2025-02-06
Use app policy for staged policies

mazdakn:
## Related issues/PRs
Based on https://github.com/projectcalico/calico/pull/9746

Changes are mainly picks of the following PRs: 
https://github.com/tigera/calico-private/pull/8162
https://github.com/tigera/calico-private/pull/8163
https://github.com/tigera/calico-private/pull/8195
https://github.com/tigera/calico-private/pull/8217
https://github.com/tigera/calico-private/pull/8273
https://github.com/tigera/calico-private/pull/8283
https://github.com/tigera/calico-private/pull/8445
https://github.com/tigera/calico-private/pull/8573
https://github.com/tigera/calico-private/pull/8582

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
TBD
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #11986, opened 2026-03-04
fix(dikastes): register ALPCheckProvider to restore L7 policy enforcement

### LINKED ISSUE #11857: L7 Policy: 403 Forbidden even with valid NetworkPolicy
## Expected Behavior

I am unable to get NetworkPolicy to permit traffic. Despite having a NetworkPolicy that should allow GET and POST requests on prefix /, the Istio-proxy and dikastes consistently returns a 403 Forbidden. This has been tested on both Minikube (Docker driver) and a bare-metal kubeadm cluster. I have also tried to create a allow-all rule but without success.

I followed the guide https://docs.tigera.io/calico/latest/network-policy/istio/app-layer-policy

## Current Behavior

Curl output when trying to POST to a service running with sidecar (kubeadm and nodeport):
```
curl -v -X POST http://10.10.0.111:30001/
*   Trying 10.10.0.111:30001...
* Connected to 10.10.0.111 (10.10.0.111) port 30001
> POST / HTTP/1.1
> Host: 10.10.0.111:30001
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 403 Forbidden
< date: Mon, 16 Feb 2026 19:00:40 GMT
< server: istio-envoy
< x-envoy-decorator-operation: grecho.default.svc.cluster.local:8001/*
< content-length: 0
<
* Connection #0 to host 10.10.0.111 left intact
```

1. Deploy a cluster using kubeadm or minikube (docker driver)
2. Setup Calico CNI using the guides in the docs (iptables)
3. Install Istio 1.29.0 using the Istio docs and operator manifest in the Calico docs (istio-operator-dikastes.yaml)
4. Annotate pods to inject Dikastes
5. Expose the service's http port (tested NodePort and metallb LoadBalancer)
6. Create a NetworkPolicy that allows TCP and HTTP (multiple combinations tried)
7. Try to access the exposed service and get a 403 Unauthorized

## Context

I don't see any errors or anything suspicious in any logs.
The goal of this is to include performance testing of this for my thesis work, so any help is appreciated.

## Your Environment

* Calico version: 3.31.3
* Orchestrator version (e.g. kubernetes, openshift, etc.): Kubernetes 1.35.1
* Operating System and version: Ubuntu 24.04

electricjesus:
L7 application layer policy (ALP) via dikastes has been completely broken in OSS since v3.30.0. The staged policies refactoring in 931244d78c (#9804) introduced a `CheckProvider` plugin interface in the checker package, but never registered a provider in OSS. When `authServer.Check()` runs, it iterates over an empty `checkProviders` slice — no policy evaluation ever happens, and every ext_authz request returns `UNKNOWN`, which Envoy treats as a denial (403).

This PR:
- Adds an `ALPCheckProvider` that wraps the existing `checkStore` policy evaluation logic
- Registers it in dikastes via a shared `newCheckServer()` function (single source of truth for provider wiring)
- Uses the existing `RegisterGRPCServices()` method, cleaning up redundant v2/v2alpha imports
- Adds FV tests that exercise the full ext_authz gRPC flow (sync server → store → Check RPC → response) using the same `newCheckServer()` as production

Fixes #11857

**Release note:**
```release-note
Fix dikastes L7 application layer policy enforcement being broken since v3.30.0 due to missing ALPCheckProvider registration.
```

### DIFF
--- a/app-policy/cmd/dikastes/dikastes.go
+++ b/app-policy/cmd/dikastes/dikastes.go
@@ -28,8 +28,6 @@ import (
 	"time"
 
 	"github.com/docopt/docopt-go"
-	authz_v2 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2"
-	authz_v2alpha "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2alpha"
 	authz "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
@@ -106,13 +104,7 @@ func runServer(arguments map[string]any) {
 	// Check server
 	gs := grpc.NewServer()
 	storeManager := policystore.NewPolicyStoreManager()
-	checkServer := checker.NewServer(ctx, storeManager,
-		checker.WithRegisteredCheckProvider(checker.NewALPCheckProvider()),
-	)
-	authz.RegisterAuthorizationServer(gs, checkServer)
-	checkServerV2 := checkServer.V2Compat()
-	authz_v2alpha.RegisterAuthorizationServer(gs, checkServerV2)
-	authz_v2.RegisterAuthorizationServer(gs, checkServerV2)
+	newCheckServer(ctx, gs, storeManager)
 
 	// Synchronize the policy store
 	opts := uds.GetDialOptions()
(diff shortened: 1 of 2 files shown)


app-policy/checker/alp_check_provider.go
```diff
@@ -0,0 +1,50 @@
...
+// endpoint configured, which is the case in per-pod (sidecar) mode.
+func (a *ALPCheckProvider) EnabledForRequest(ps *policystore.PolicyStore, _ *authz.CheckRequest) bool {
+	return ps.Endpoint != nil
+}
+
+// Check evaluates the request against the configured policies and profiles
+// for the local workload endpoint and returns the authorization decision.
+func (a *ALPCheckProvider) Check(ps *policystore.PolicyStore, req *authz.CheckRequest) (*authz.CheckResponse, error) {
+	flow := NewCheckRequestToFlowAdapter(req)
+	st := checkStore(ps, ps.Endpoint, rules.RuleDirIngress, flow)
```
caseydavenport, line 48: Is this hard coded rule dir correct?

electricjesus: Yes that's correct unfortunately ALP is ingress-only. In the Istio doc about ALP [1] we setup Dikastes as an ext_authz filter so we will only really have request (inbound) data and we'll not have any data to use for egressing policies (we won't have HTTP response headers, body, etc to evaluate egress ALP policy against)

we _could_ have an ext_proc filter as an enhancement later on...


1: https://docs.tigera.io/calico/latest/network-policy/istio/enforce-policy-istio

PR #11989, opened 2026-03-04
Sync comment for ProgramClusterRoutes in FelixConfig and BGPConfig

mazdakn:
Use the same doc comments used for `ProgramClusterRoutes` in BGPConfig that was added in the below PR for the same config option in FelixConfig 
https://github.com/projectcalico/calico/pull/11845

**Release note:**
```release-note
TBD
```


mazdakn:
@frozenprocess feel free to suggest a better one.

PR #11990, opened 2026-03-04
fix(windows): use 'crane ls' to find latest kube-proxy-windows patch version

coutinhop:
Use 'crane ls' to find out the latest kube-proxy-windows patch version and use that. For a few days, k8s v1.34.5 was available but the latest kube-proxy-windows v1.34 version was v1.34.4, which worked fine in clusters. This shouldn't happen longer than a few days, but with these changes it will no longer cause issues with spinning up ASO clusters.

**Release note:**
```release-note
TBD
```

### DIFF
--- a/process/testing/aso/install-calico.sh
+++ b/process/testing/aso/install-calico.sh
@@ -193,10 +193,15 @@ timeout --foreground 600 bash -c "while ! ${KUBECTL} wait pod -l k8s-app=calico-
 echo "Calico is ready on Windows nodes"
 
 # Create the kube-proxy-windows daemonset. Use 'crane ls' to use the latest patch release on the same minor version as KUBE_VERSION, in
-# case the kube-proxy-windows for the exact patch version hasn't been publishded yet
-KUBE_PROXY_WIN_TAG=$(${CRANE} ls sigwindowstools/kube-proxy | grep "^${KUBE_VERSION%.*}.*-calico.*" | sort -Vr | head -1)
-KUBE_PROXY_WIN_VERSION="${KUBE_PROXY_WIN_TAG%%-calico*}"
-echo "Install kube-proxy-windows ${KUBE_VERSION} from sig-windows-tools"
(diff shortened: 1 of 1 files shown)


process/testing/aso/install-calico.sh
```diff
@@ -191,10 +192,13 @@ echo "Wait for Calico to be ready on Windows nodes..."
 timeout --foreground 600 bash -c "while ! ${KUBECTL} wait pod -l k8s-app=calico-node-windows --for=condition=Ready -n calico-system --timeout=30s; do sleep 5; done"
 echo "Calico is ready on Windows nodes"
 
-# Create the kube-proxy-windows daemonset
+# Create the kube-proxy-windows daemonset. Use 'crane ls' to use the latest patch release on the same minor version as KUBE_VERSION, in
+# case the kube-proxy-windows for the exact patch version hasn't been publishded yet
+KUBE_PROXY_WIN_TAG=$(${CRANE} ls sigwindowstools/kube-proxy | grep "^${KUBE_VERSION%.*}.*-calico.*" | sort -Vr | head -1)
```
skoryk-oleksandr: What if `crane ls` fails or the `grep` matches nothing? Probably both `KUBE_PROXY_WIN_TAG` and `KUBE_PROXY_WIN_VERSION` will be empty. The **sed** command will then inject an empty string into the
  `kube-proxy` manifest?

process/testing/aso/install-calico.sh
```diff
@@ -191,10 +192,13 @@ echo "Wait for Calico to be ready on Windows nodes..."
...
+KUBE_PROXY_WIN_VERSION="${KUBE_PROXY_WIN_TAG%%-calico*}"
 echo "Install kube-proxy-windows ${KUBE_VERSION} from sig-windows-tools"
```
skoryk-oleksandr: should we echo `${KUBE_PROXY_WIN_VERSION}` instead of `${KUBE_VERSION}` here ?

PR #11992, opened 2026-03-04
Make BIRD to not advertise cluster routes to iBGP peers when felix programs cluster routes

mazdakn:
Generate BIRD filters to not advertise cluster routes (only IPIP and no-encap ippools) programmed by Felix to other iBGP peers, since those are also (most likely) calico nodes and already have the intended routes.

In case of vxlan and wireguard, we use a filter based on interface to do the same thing:
https://github.com/projectcalico/calico/blob/691ebda3a308855688701e73f4e83075621f82c7/confd/etc/calico/confd/templates/bird_ipam.cfg.template#L4-L12

**Release note:**
```release-note
TBD
```

### DIFF
--- a/confd/pkg/backends/calico/bgp_processor.go
+++ b/confd/pkg/backends/calico/bgp_processor.go
@@ -877,8 +877,8 @@ func processIPPool(
 			// Exporting routes to iBGP peers.
 			if forInternalPeers {
 				// Do not generate any statement, since "reject_tunnel_routes" filter rejects all routes related to
-				// *.calico and *.cali interfaces which rejects all vxlan, and wireguard routes.
-				// Also, BIRD does not re-advertise ipip and no-encap routes, as those are programmed by BIRD itself.
+				// *.calico and *.cali interfaces, which rejects all VXLAN, and WireGuard routes.
+				// Also, BIRD does not re-advertise IPIP and no-encap routes, as those are programmed by BIRD itself.
 				return ""
 			}
 			// Exporting routes to all peers.


confd/etc/calico/confd/templates/bird6_ipam.cfg.template
```diff
@@ -9,6 +9,9 @@ function reject_tunnel_routes () {
         reject;
      }
   }
+{{- range $line := $config.InternalBGPExportFilter }}
+{{ $line }}
+{{- end}}
```
mazdakn, line 14: In the next PR, I'm going to inline all the filters in this function. Not included here to minimize the number of changes, i.e. generated test files.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -772,20 +772,26 @@ func (c *client) processIPPools(config *types.BirdBGPConfig, ipVersion int) erro
 		}
 
 		// Generate statements for rejecting disabled ippools in the filter for exporting routes to other peers.
-		statement := c.processIPPool(&ippool, programClusterRoutes, false, "reject", "", ipVersion)
+		statement := processIPPool(&ippool, programClusterRoutes, false, false, "reject", "", ipVersion)
```
mazdakn, line 775: I'll clean up all the values passed to `processIPPool` function in a separate PR. Keep it out to simplify this PR.

mazdakn:
@nelljerram this PR is basically to sync the behavior with vxlan and wireguard, as those two have filters to rejects exporting routes to other nodes.
We introduced those filters since users found issues with extra and unnecessary routes in BIRD state. IIRC, several routes were exported to other nodes for the same subnet. 

>we shouldn't restrict export to other nodes, because we don't know for certain how those other nodes are configured

The other nodes anyway know about the subnet since felix would program it to the kernel, and bird reads it from kernel.

confd/etc/calico/confd/templates/bird_ipam.cfg.template
```diff
@@ -9,6 +9,9 @@ function reject_tunnel_routes () {
         reject;
      }
   }
+{{- range $line := $config.BGPExportFilterForInternalPeers }}
+{{ $line }}
+{{- end}}
```
nelljerram, line 14: It looks odd to place this inside `reject_tunnel_routes` given that I think it also applies to no-encap routes.  Am I misunderstanding?

confd/tests/compiled_templates/felix_cluster_routing/bird6_ipam.cfg
nelljerram, line 4: So am I right that this actually refers to VXLAN and Wireguard routes?  I understand why you're not changing the comment in this PR - so just checking my understanding.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -864,13 +874,27 @@ func (c *client) processIPPool(
 				// our fork of BIRD.
 				extraStatement = extraStatementForKernelProgrammingIPIPNoEncap(ippool.IPIPMode, localSubnet)
 			}
+			// Exporting routes to iBGP peers.
+			if forInternalPeers {
+				// Do not generate any statement, since "reject_tunnel_routes" filter rejects all routes related to
+				// *.calico and *.cali interfaces, which rejects all VXLAN, and WireGuard routes.
+				// Also, BIRD does not re-advertise IPIP and no-encap routes, as those are programmed by BIRD itself.
```
nelljerram, line 881: I don't understand this comment.  Instead I would expect it to say "Don't reject export of IP-IP or no-encap routes because we *want* BIRD to program these."

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -864,13 +874,27 @@ func (c *client) processIPPool(
...
 
 		// Felix is responsible for programming cluster routes, not BIRD.
 		if forProgrammingKernel {
 			return emitFilterStatementForIPPools(cidr, "", "reject", filterAction, "Cluster routes are handled by Felix.")
 		}
+		// Reject exporting routes to iBGP peers when Felix is responsible for programming cluster routes.
+		// Disabled ippools are handled above.
+		if forInternalPeers && !ippool.DisableBGPExport {
+			return emitFilterStatementForIPPools(cidr, "", "reject", filterAction, "")
+		}
```
nelljerram, line 896: I think the behaviour here has changed for `ippool.DisableBGPExport && !forProgrammingKernel && forInternalPeers`.  Previously we had
```
return emitFilterStatementForIPPools(cidr, "", "reject", filterAction, "BGP export is disabled.")
```
Now I think we have
```
return emitFilterStatementForIPPools(cidr, "", "accept", filterAction, "")
```

confd/etc/calico/confd/templates/bird_ipam.cfg.template
mazdakn: I was thinking to eventually inline this filter function. But then inlining these lines is more aligned with the approach. Fixed now.

confd/tests/compiled_templates/felix_cluster_routing/bird6_ipam.cfg
mazdakn: correct. I'll update these comments later. Touching these makes all compiled files to change, and I wanted to keep this PR as small as possible.

mazdakn:
@nelljerram is your first point a valid/supported scenario? cause we already filter vxlan and wireguard routes for iBGP peers.

mazdakn:
IMO, it does make sense to not hard code filters, and instead give users a tool to fix their issues. For example, in this case, maybe the better approach is to NOT complicate our static logic, and rely on user defined filters, i.e. BGPFilter resource. WDYT?

PR #11993, opened 2026-03-04
Fix Felix FV infrastructure flakiness

caseydavenport:
Three recurring infra issues have been causing FV test flakes, identified via fv-tests-guru analysis of recent failures:

**Exponential backoff for infra retry loops** — All the retry loops in `infra_k8s.go` poll at a fixed 100ms interval, which hammers the API server when it's already overloaded (especially when it starts returning 429s). Switched to exponential backoff that doubles up to a 5s cap. The CRD availability loop also bumps the delay to at least 1s on 429 so we stop making things worse.

**BPF ready timeout 30s → 90s** — When the API server is recovering from overload, BPF program loading can take way longer than 30s since dataplane programming gets delayed. 90s gives enough headroom for apiserver recovery + BPF loading when starting several felix nodes in parallel.

**Drop `--asm` from `bpfDumpPolicy()`** — The `--asm` flag makes `calico-bpf` disassemble every BPF program, which is expensive and starves Felix of CPU when called repeatedly in Eventually/Consistently loops. None of the 35+ callers need the disassembly — they only check structural markers (policy names, tier boundaries) that show up without `--asm`. Added a separate `bpfDumpPolicyAsm()` for anything that needs it in the future.

### DIFF
--- a/felix/fv/bpf_test.go
+++ b/felix/fv/bpf_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021-2025 Tigera, Inc. All rights reserved.
+// Copyright (c) 2021-2026 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


felix/fv/infrastructure/felix.go
```diff
@@ -454,7 +454,7 @@ func (f *Felix) WaitForReady() {
 	if BPFMode() {
 		// BPF mode has to load BPF programs at startup, this can take a while
 		// when starting several felix nodes in parallel.
-		timeout = "30s"
+		timeout = "90s"
 	}
 	EventuallyWithOffset(1, f.Ready, timeout, "100ms").Should(BeTrue(),
```
tomastigera, line 459: If we expect 90s should we test every 100ms?

felix/fv/infrastructure/infra_k8s.go
```diff
@@ -53,6 +53,17 @@ import (
 
 const DefaultBPFLogByteLimit = 64 * 1024 * 1024
 
+// retryBackoff sleeps for the given delay and returns the next delay,
+// doubling each time up to maxDelay.
+func retryBackoff(current, maxDelay time.Duration) time.Duration {
+	time.Sleep(current)
+	next := current * 2
```
tomastigera, line 60: isn't *2 too fast? 100ms gets into seconds pretty quickly.

felix/fv/infrastructure/felix.go
caseydavenport: Yeah, we could up this, but I think in the mainline case we still want to keep things snappy?

felix/fv/infrastructure/infra_k8s.go
caseydavenport: But we cap out at 5s, so I think it's OK - the point here is to back off pretty quickly so we stop hammering the API server.

PR #11994, opened 2026-03-05
Enable parallel Ginkgo execution for e2e tests

caseydavenport:
Switch e2e test execution from direct binary invocation to the `ginkgo` CLI with `-procs=N` (default 4), so tests run in parallel across multiple processes. This cuts conformance test wall time roughly in half.

The main change is wiring `ginkgo -procs=$(E2E_PROCS)` into the Makefile's `e2e-run-test` target and the CI script in `body_standard.sh`. Override with `E2E_PROCS=1` to go back to sequential if needed.

Supporting changes to make existing tests parallel-safe:

- **`e2e_test.go`**: Add `testing.Init()` before `flag.Parse()` so the ginkgo CLI can pass `-test.timeout`, `-test.count`, etc. to child processes.
- **`kubectl.go`**: `PortForward` now allocates a random local port instead of reusing the remote port, avoiding bind conflicts when multiple tests port-forward simultaneously.
- **`policy.go` / `service_policy.go`**: Remove `CleanDatastore` calls — this deletes *all* GNPs and NetworkPolicies cluster-wide, which obviously races with other tests running in parallel. Each test already cleans up its own resources via `defer`.
- **`policy.go`**: Randomize GNP names with `GenerateRandomName` to avoid `AlreadyExists` collisions.
- **`service_policy.go`**: Scope the `allow-kube-dns` GNP to the test namespace via `NamespaceSelector` instead of `Selector: "all()"` — the latter matches every pod in the cluster, which breaks other tests' connectivity checks.
- **`staged_policy.go`**: Updated to use the port returned by `PortForward`.


e2e/pkg/tests/apis/owner_references.go
```diff
@@ -44,14 +44,10 @@ var _ = describe.CalicoDescribe(
 		)
 
 		ginkgo.BeforeEach(func() {
-			// Ensure a clean starting environment before each test.
 			var err error
 			cli, err = client.New(f.ClientConfig())
 			Expect(err).NotTo(HaveOccurred())
 
-			// Ensure a clean starting environment before each test.
```
caseydavenport, line 52: CleanDatastore is a major problem for running in parallel - it just blats away the config for any in-progress test.

caseydavenport: Tests just need to track and clean up their own resources, which is pretty easy with `DeferCleanup` calls. We might even be able to hook this into the Client natively to force the defer cleanup so it is never missed?

caseydavenport:
> I'm surprised that PROCS=4 only gives a 2x speedup. Any idea why that might be?

@lwr20 I think it's because the SERIAL tests still need to run after the parallel tests finish, so not all tests are parallelized.

PR #12002, opened 2026-03-05
Enhance BGPFilter API with new match criteria and route operations

song-jiang:
## Summary

Extends the BGPFilter API with new match criteria and route modification operations to support advanced BGP use cases such as KubeVirt live migration route priority propagation.

### New match criteria on BGPFilterRuleV4/V6

- **PeerType** (`eBGP`/`iBGP`): Filter rules can target specific peer types. When used, the generated BIRD function takes a `bool is_internal` parameter and wraps rules in `if (is_internal)` / `if (!is_internal)` guards.
- **Communities**: Match routes carrying specific standard (`aa:nn`) or large (`aa:nn:mm`) BGP communities. Supported on both import and export.
- **ASPathPrefix**: Match routes whose AS path begins with a specified sequence of AS numbers.
- **Priority**: Match routes with a specific priority (krt_metric) value.

### New operations (only valid with `action: Accept`)

- **AddCommunity**: Attach a BGP community to matching routes.
- **PrependASPath**: Prepend AS numbers to the route's AS path.
- **SetPriority**: Set the route's priority (krt_metric).

### Validation

- CEL validation (`x-kubernetes-validations`) enforces:
  - Operations are only allowed with `action: Accept`
  - Exactly one operation field per `BGPFilterOperation` entry
  - Community values conform to standard (16-bit) or large (32-bit) format
- Doc comments added to all pre-existing and new fields on BGPFilterRuleV4/V6

### Implementation

- **API types** (`api/pkg/apis/projectcalico/v3/bgpfilter.go`): New types and fields with kubebuilder markers
- **confd template functions** (`confd/pkg/resource/template/template_funcs.go`): BIRD filter generation for all new match criteria and operations
- **confd BGP processor** (`confd/pkg/backends/calico/bgp_processor.go`): PeerType-aware function call generation (passes `is_internal` bool when needed)
- **Tests**: Unit tests for all new template functions and BGP processor changes

**Release note:**
```release-note
BGPFilter: Add new match criteria (PeerType, Communities, ASPathPrefix, Priority) and route operations (AddCommunity, PrependASPath, SetPriority) to BGPFilter rules, enabling advanced BGP policy use cases such as route priority propagation for KubeVirt live migration.
```

### DIFF
--- a/api/pkg/apis/projectcalico/v3/bgpfilter_test.go
+++ b/api/pkg/apis/projectcalico/v3/bgpfilter_test.go
@@ -60,7 +60,7 @@ func TestBGPFilterValidation(t *testing.T) {
 			name: "basic valid BGPFilter",
 			obj: &v3.BGPFilter{
 				ObjectMeta: metav1.ObjectMeta{Name: "valid-bgpfilter"},
-				Spec:       v3.BGPFilterSpec{ExportV4: []v3.BGPFilterRuleV4{{CIDR: "10.0.0.0/24", Action: v3.Accept}}},
+				Spec:       v3.BGPFilterSpec{ExportV4: []v3.BGPFilterRuleV4{{CIDR: "10.0.0.0/24", MatchOperator: v3.MatchOperatorIn, Action: v3.Accept}}},
 			},
 			valid: true,
 		},


api/pkg/apis/projectcalico/v3/bgpfilter.go
```diff
@@ -60,38 +61,150 @@ type BGPFilterSpec struct {
...
+	// "matches" the given CIDR - where the definition of "matches" is according to
+	// MatchOperator and PrefixLength.  CIDR should be in conventional CIDR notation,
+	// <prefix>/<length>.
 	// +kubebuilder:validation:Format=cidr
 	CIDR string `json:"cidr,omitempty" validate:"omitempty,netv4"`
 
+	// PrefixLength further constrains the CIDR match by restricting the range of allowed
+	// prefix lengths.  For example, CIDR "10.0.0.0/8" with MatchOperator "In" and
+	// PrefixLength {min: 16, max: 24} matches any route within 10.0.0.0/8 whose prefix
+	// length is between /16 and /24.  Only meaningful when CIDR is also specified; if
```
fasaxc: Why is it only meaningful if CIDR is also specified; I'd expect the default for CIDR to be 0.0.0.0/0, so the default prefix length range is 0-32 and could be refined from there.

api/pkg/apis/projectcalico/v3/bgpfilter.go
```diff
@@ -139,6 +252,80 @@ const (
...
+	// Values is a list of BGP community values to match against.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=1
+	Values []BGPCommunityValue `json:"values" validate:"required"`
+}
+
+// BGPFilterOperation is a discriminated union representing a single route modification.
+// Exactly one field must be set.
+// +mapType=atomic
+// +kubebuilder:validation:XValidation:rule="(has(self.addCommunity) ? 1 : 0) + (has(self.prependASPath) ? 1 : 0) + (has(self.setPriority) ? 1 : 0) == 1",message="exactly one operation must be set"
```
fasaxc: I think there's a shortcut for this that doesn't need CEL.  We should only use CEL if there's no way to do it with normal validation.

charts/crd.projectcalico.org.v1/templates/operator.tigera.io_goldmanes.yaml
fasaxc, line 1448: Seems out of place in this PR?  Missing CRD update on master?

song-jiang:
I have prepared a follow-up [PR](https://github.com/song-jiang/calico/pull/9) with additional BGPFilter test cases once this PR get merged. All new test cases passed locally.

api/pkg/apis/projectcalico/v3/bgpfilter.go
```diff
@@ -139,6 +252,80 @@ const (
...
+	// ...RoutePriority fields in FelixConfiguration.
+	// +optional
+	SetPriority *BGPFilterSetPriority `json:"setPriority,omitempty" validate:"omitempty"`
+}
+
+// BGPFilterAddCommunity specifies a BGP community to add to a route.
+// +mapType=atomic
+type BGPFilterAddCommunity struct {
+	// Value is the BGP community to add.
+	Value BGPCommunityValue `json:"value" validate:"required"`
```
caseydavenport: Think these should be pointers if we ever intend to support additional sub-fields in this struct and make this optional. If we can't foresee this being optional then it can stay as-is, though. Same for the SetPriority type.

song-jiang: Good point. We should use a pointer so that we can expand the struct later, e.g. adding a named Community or Priority.

api/pkg/apis/projectcalico/v3/bgpfilter.go
song-jiang: Good point. We should use the following annotation
```
// +kubebuilder:validation:MinProperties=1
// +kubebuilder:validation:MaxProperties=1
```

api/pkg/apis/projectcalico/v3/bgpfilter.go
song-jiang: This is current implementation, PrefixLength without CIDR is silently ignored (no CIDR condition generated, rule matches all routes). We can remove this limitation by a follow-up PR if we need.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -606,7 +644,11 @@ func (c *client) buildImportFilter(filters []string, ipVersion int) string {
 				// Check if import rules exist based on IP version
 				if (ipVersion == 4 && len(filter.Spec.ImportV4) > 0) || (ipVersion == 6 && len(filter.Spec.ImportV6) > 0) {
 					truncatedName := truncateBGPFilterName(filterName)
-					filterLines = append(filterLines, fmt.Sprintf("'bgp_%s_importFilter%s'();", truncatedName, filterSuffix))
+					if filterHasPeerType(&filter, "import", ipVersion) {
+						filterLines = append(filterLines, fmt.Sprintf("'bgp_%s_importFilter%s'(%v);", truncatedName, filterSuffix, sameAS))
+					} else {
+						filterLines = append(filterLines, fmt.Sprintf("'bgp_%s_importFilter%s'();", truncatedName, filterSuffix))
```
nelljerram, line 650: I realize they already existed before this PR, but why are single quotes used here?

confd/pkg/resource/template/template_funcs.go
```diff
@@ -184,6 +225,88 @@ func filterMatchInterface(iface string) (string, error) {
...
+	}
+	if len(asPathPrefix) == 1 {
+		return fmt.Sprintf("(bgp_path.first = %s)", asPathPrefix[0].String()), nil
+	}
+	// Multiple ASNs: bgp_path ~ [= ASN1 ASN2 ... * =]
+	var asns []string
+	for _, asn := range asPathPrefix {
+		asns = append(asns, asn.String())
+	}
+	return fmt.Sprintf("(bgp_path ~ [= %s * =])", strings.Join(asns, " ")), nil
```
nelljerram, line 259: Can we simplify a little by using the general form for the `len(asPathPrefix) == 1 ` case as well?

confd/pkg/resource/template/template_funcs.go
```diff
@@ -210,51 +333,90 @@ type filterArgs struct {
...
+
+// hasPeerTypeRules returns true if any of the given filterArgs have a PeerType set.
+func hasPeerTypeRules(rules []filterArgs) bool {
+	for _, r := range rules {
+		if r.peerType != "" {
+			return true
+		}
+	}
+	return false
+}
```
nelljerram, line 382: There's a very similar function in bgp_processor.go.  Do we need both of them?

confd/pkg/resource/template/template_funcs.go
```diff
@@ -210,51 +333,90 @@ type filterArgs struct {
...
-//	  "  if ((net ~ 44.0.0.0/16)) then { accept; }",
-//	  "  if ((net ~ 44.1.0.0/16)) then { reject; }",
-//	  "}",
-//	  "function 'bgp_test-bgpfilter_exportFilterV4'() {",
-//	  "  if ((net ~ 77.0.0.0/16)) then { accept; }",
-//	  "  if ((net ~ 77.1.0.0/16)) then { reject; }",
-//	  "}",
-//	 }
+// Rules with PeerType are then wrapped in if (is_same_as) / if (!is_same_as) guards.
 func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
```
nelljerram, line 416: What is the input here (i.e. what is `pairs`) ?

What is the output?

nelljerram: It looks like it's a set of v3.BGPFilter.  But what set?  Those needed for a particular BGPPeer?  Or the union of those needed for all BGPPeers?  And why the wrapping as KV pairs?

confd/pkg/resource/template/template_funcs.go
```diff
@@ -210,51 +333,90 @@ type filterArgs struct {
...
+		cidr:           rule.CIDR,
+		prefixLengthV4: rule.PrefixLength,
+		source:         rule.Source,
+		iface:          rule.Interface,
+		peerType:       rule.PeerType,
+		communities:    rule.Communities,
+		asPathPrefix:   rule.ASPathPrefix,
+		priority:       rule.Priority,
+		action:         rule.Action,
+		operations:     rule.Operations,
```
nelljerram, line 353: I think the code is more difficult to understand overall because of moving data around to internal structs with different names.  Do we really need to do that?  Could the apparent problem by addressed with templating instead?

confd/pkg/resource/template/template_funcs.go
```diff
@@ -210,51 +333,90 @@ type filterArgs struct {
...
+		case v3.BGPFilterPeerTypeEBGP:
+			filterRule = fmt.Sprintf("if (!is_same_as) then { %s }", filterRule)
+		}
+
+		lines = append(lines, fmt.Sprintf("  %s", filterRule))
+	}
+	return lines, nil
 }
 
 // BGPFilterBIRDFuncs generates a set of BIRD functions for BGPFilter resources that have been packaged into KVPairs.
```
nelljerram, line 407: Suggest:
```
// BGPFilterBIRDFuncs generates the definitions of a set of BIRD functions for all configured BGPFilter resources.
```

confd/pkg/resource/template/template_funcs.go
```diff
@@ -210,51 +333,90 @@ type filterArgs struct {
...
-// template code into our BIRD config templates that is both difficult to read and prone to errors
-//
-// e.g. for a BGPFilter resource specified as follows:
-//
-// kind: BGPFilter
-// apiVersion: projectcalico.org/v3
-// metadata:
-//
-//	name: test-bgpfilter
+// template code into our BIRD config templates that is both difficult to read and prone to errors.
```
nelljerram, line 409: I don't think this para tells me anything, can we delete?

confd/pkg/resource/template/template_funcs.go
```diff
@@ -298,99 +455,70 @@ func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
...
+			if hasPeerTypeRules(ruleFields) {
+				line = fmt.Sprintf("function %s(bool is_same_as) {", filterFuncName)
+			} else {
+				line = fmt.Sprintf("function %s() {", filterFuncName)
 			}
-
-			line = "}"
 			lines = append(lines, line)
+
+			ruleLines, err := emitFilterRules(ruleFields)
```
nelljerram, line 478: Please inline the logic of emitFilterRules here.

confd/pkg/resource/template/template_funcs.go
```diff
@@ -298,99 +455,70 @@ func BGPFilterBIRDFuncs(pairs memkv.KVPairs, version int) ([]string, error) {
...
+			if err != nil {
+				return []string{}, err
+			}
+			lines = append(lines, ruleLines...)
+
+			lines = append(lines, "}")
```
nelljerram, line 517: Can this be reworked to avoid duplicating the whole code twice?  (It feels to me that should be pretty easy.)

confd/pkg/resource/template/template_funcs.go
song-jiang: The difference is that one uses the v3 struct and the other uses the internal ones. We definitely should consolidate them, but let's do it in a separate follow-up PR.

confd/pkg/backends/calico/bgp_processor.go
song-jiang: The single quotes are required by BIRD syntax, they allow special characters (e.g. my-bgp-filter) in identifiers.

confd/pkg/resource/template/template_funcs.go
song-jiang: Good point.

confd/pkg/resource/template/template_funcs.go
song-jiang: `pairs` is all configured BGPFilter resources, passed as KV pairs because that's the standard interface confd's template engine uses to pass datastore resources to template functions. In this case, the template function is bgpFilterBIRDFuncs, see bird.cfg.template below.
```
{{- range $line := bgpFilterBIRDFuncs (gets "/resources/v3/projectcalico.org/bgpfilters/*") 4 }} 
``` 

Each pair has the datastore path as key (e.g. /calico/resources/v3/projectcalico.org/bgpfilters/my-filter) and the JSON-serialized v3.BGPFilter as value. The output is a list of BIRD config lines defining the filter functions.

confd/pkg/resource/template/template_funcs.go
song-jiang: The reason filterArgs exists is that `BGPFilterRuleV4` and `BGPFilterRuleV6` are separate types (V4 has `*BGPFilterPrefixLengthV4`, V6 has `*BGPFilterPrefixLengthV6`). Without
  filterArgs, every function like filterStatement, filterMatchCIDR, etc. would need to be duplicated, one for V4 and one for V6.

PR #12005, opened 2026-03-05
Handle pod-not-found gracefully in CNI DEL for virt-launcher pods

song-jiang:
During KubeVirt live migration, multiple CNI DEL calls can arrive in
parallel for the same pod. If the pod is already deleted by the time
a subsequent DEL calls `getVMIInfoForPod`, the Pods().Get() returns a
404 error, causing the entire CNI DEL to fail. 

This PR handles the not-found case by returning (nil, nil) so that
cmdDel falls through to the standard release path, which already
handles missing resources gracefully.

Also update the KubeVirtVMAddressPersistence API field documentation
to clarify that creating a live migration target pod is not supported
when IP persistence is disabled.

**Release note:**
```release-note
TBD
```

### DIFF
--- a/cni-plugin/pkg/ipamplugin/ipam_plugin.go
+++ b/cni-plugin/pkg/ipamplugin/ipam_plugin.go
@@ -17,6 +17,7 @@ package ipamplugin
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"net"
@@ -44,7 +45,7 @@ import (
 	"github.com/projectcalico/calico/cni-plugin/pkg/upgrade"
 	"github.com/projectcalico/calico/libcalico-go/lib/apiconfig"
 	client "github.com/projectcalico/calico/libcalico-go/lib/clientv3"
-	"github.com/projectcalico/calico/libcalico-go/lib/errors"
+	cerrors "github.com/projectcalico/calico/libcalico-go/lib/errors"
 	"github.com/projectcalico/calico/libcalico-go/lib/ipam"
 	"github.com/projectcalico/calico/libcalico-go/lib/ipam/vmipam"
 	"github.com/projectcalico/calico/libcalico-go/lib/kubevirt"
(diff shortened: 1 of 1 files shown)


cni-plugin/pkg/ipamplugin/ipam_plugin.go
```diff
@@ -549,9 +550,10 @@ func acquireIPAMLockBestEffort(path string) unlockFn {
 
 // getVMIInfoForPod retrieves KubeVirt VirtualMachineInstance (VMI) information for a given pod.
 // Returns (vmiInfo, nil) if the pod is a valid virt-launcher pod.
-// Returns (nil, nil) if the pod is not a virt-launcher pod.
+// Returns (nil, nil) if the pod is not a virt-launcher pod, or if isDel is true and the pod
+// has already been deleted (e.g., due to parallel CNI DEL calls during live migration).
 // Returns (nil, error) if there was an error retrieving or validating VMI information.
-func getVMIInfoForPod(conf types.NetConf, epIDs *utils.WEPIdentifiers, logger *logrus.Entry) (*kubevirt.PodVMIInfo, error) {
+func getVMIInfoForPod(conf types.NetConf, epIDs *utils.WEPIdentifiers, logger *logrus.Entry, isDel bool) (*kubevirt.PodVMIInfo, error) {
```
fasaxc: Bool behaviour toggle arguments are a code smell; better to return a specific error if pod is not found and then the caller can decide what to do with that information.

cni-plugin/pkg/ipamplugin/ipam_plugin.go
```diff
@@ -654,7 +663,14 @@ func cmdDel(args *skel.CmdArgs) error {
 	// Always detect VMI info for logging, but only use VM-based handle when persistence is enabled.
 	vmiInfo, err := getVMIInfoForPod(conf, epIDs, logger)
 	if err != nil {
-		return fmt.Errorf("failed to get VMI info: %w", err)
+		if err == errPodNotFound {
```
fasaxc: You should probably use `errors.Is()` here in case the error gets wrapped later.

PR #12006, opened 2026-03-05
Fix tier RBAC authorization for policy tier changes

caseydavenport:
Now that policies can change tiers without changing their name, UPDATE
operations need to authorize against both the old and new tier. Previously
the apiserver REST storage only checked the old tier, and the webhook only
checked the new tier — either path allowed a user to move a policy into or
out of a tier they don't have permission for.

The apiserver `Update()` methods now resolve the new object via
`objInfo.UpdatedObject()` and check the new tier if it differs from the old
one. The webhook `authorize()` function now also parses `OldObject` on UPDATE
operations and checks the old tier if it differs.

Also adds test coverage for the tier-migration path in Felix's calc graph:
- Policy sorter unit tests for `UpdatePolicy` and `OnUpdate` with tier changes
- Policy lookups cache test verifying tier ref counting across tier changes
- Calc graph FV test exercising a policy moving between two non-default tiers
  (40 expanded test variations covering different flush strategies and orderings)

### DIFF
--- a/webhooks/pkg/rbac/rbac.go
+++ b/webhooks/pkg/rbac/rbac.go
@@ -172,7 +172,7 @@ func (h *tieredRBACHook) authorize(ar v1.AdmissionReview) *v1.AdmissionResponse 
 	if tier == "" {
 		tier = oldTier
 	}
-	logCtx = logCtx.WithField("tier", tier)
+	logCtx = logCtx.WithFields(logrus.Fields{"newTier": newTier, "oldTier": oldTier})
 	if err = h.authz.AuthorizeTierOperation(ctx, obj.GetName(), tier); err != nil {
 		logCtx.WithError(err).Warn("User is not authorized")
 		return &v1.AdmissionResponse{


webhooks/pkg/rbac/rbac.go
matthewdupre: Would it be worth changing this error to "Failed to parse policy metadata or tier" since the tier isn't strictly metadata?  Not sure how possible it is for us to fail to parse a tier though, in which case it may be moot.

webhooks/pkg/rbac/rbac.go
```diff
@@ -161,6 +161,37 @@ func (h *tieredRBACHook) authorize(ar v1.AdmissionReview) *v1.AdmissionResponse
...
+					Status:  metav1.StatusFailure,
+					Message: fmt.Sprintf("Failed to parse old policy metadata: %v", err),
+					Reason:  metav1.StatusReasonInvalid,
+				},
+			}
+		}
+		if oldTier != tier {
+			logCtx.WithField("oldTier", oldTier).Debug("Tier changed, authorizing old tier")
+			if err = h.authz.AuthorizeTierOperation(ctx, obj.GetName(), oldTier); err != nil {
+				logCtx.WithError(err).Warn("User is not authorized for old tier")
```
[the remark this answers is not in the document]
matthewdupre: No Copilot - parsePolicy() calls through to getTier() which is modified below to call TierOrDefault().

webhooks/pkg/rbac/rbac.go
```diff
@@ -161,6 +162,37 @@ func (h *tieredRBACHook) authorize(ar v1.AdmissionReview) *v1.AdmissionResponse
 		}
 	}
 
+	// For UPDATE operations, also authorize the old tier. A tier change requires permission on both
+	// the old tier (to remove the policy from it) and the new tier (already checked above).
+	if ar.Request.Operation == v1.Update && len(ar.Request.OldObject.Raw) > 0 {
+		_, oldTier, err := h.parsePolicy(ar.Request.Kind.Kind, ar.Request.OldObject.Raw)
```
matthewdupre: I think it's correct in all cases, but I don't really like the way the tier and the oldTier get used here from a naming sense.

This is what I think is really happening:
- When deleting, tier is the old object
- Otherwise it's the new one.
- When updating, the extra variable oldTier is added which is the old one.

Please either comment the variables to describe exactly what they are and why the code is correct, or if you/Claude can be bothered I think a reflowing would be possible where you use oldTier/newTier and have them really mean that.

webhooks/pkg/rbac/rbac.go
caseydavenport: Yeah, good call — I simplified it further to just "Failed to parse policy" since `parsePolicy` handles both the metadata and the tier extraction. Felt like calling out "metadata or tier" was a bit awkward when it's really one function doing both.

webhooks/pkg/rbac/rbac.go
caseydavenport: Yep, agreed the naming was confusing. Restructured the function to parse `Object.Raw` and `OldObject.Raw` separately so `newTier` always comes from the new object and `oldTier` always comes from the old one. For DELETE (no new object), it falls back to `oldTier` for the authorization check. Should be a lot clearer now.

webhooks/pkg/rbac/rbac.go
```diff
@@ -148,7 +167,12 @@ func (h *tieredRBACHook) authorize(ar v1.AdmissionReview) *v1.AdmissionResponse
 		}
 	}
 
-	// Run the RBAC authorizer to check if the user is authorized to perform the operation.
+	// Authorize the new tier for CREATE/UPDATE, or the old tier for DELETE.
+	tier := newTier
```
caseydavenport, line 171: Let's just include both values in the context logger.

PR #12010, opened 2026-03-05
Add ClusterInformation write-protection webhook

caseydavenport:
When running in aggregated API server mode, the Calico API server explicitly blocks Create/Update/Delete on ClusterInformation resources (returns `MethodNotSupported`), making the resource effectively read-only through the API. In v3 CRD mode, there's no equivalent protection — users can modify ClusterInformation directly since it's just a regular CRD.

This adds a validating admission webhook handler that blocks writes to ClusterInformation unless the request comes from a known Calico system service account (`calico-node` or `calico-kube-controllers`). The handler is registered at `/cluster-info` on the existing webhook server.

The operator will need a corresponding change to include `clusterinformations` in the `ValidatingWebhookConfiguration` rules so the K8s API server actually routes admission requests to this handler.

Ref: CORE-12369

### DIFF
--- a/webhooks/deps.txt
+++ b/webhooks/deps.txt
@@ -108,5 +108,6 @@ local:libcalico-go/lib/namespace
 local:libcalico-go/lib/net
 local:pkg/buildinfo
 local:webhooks/cmd
+local:webhooks/pkg/clusterinfo
 local:webhooks/pkg/rbac
 local:webhooks/pkg/utils


webhooks/pkg/clusterinfo/clusterinfo.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/projectcalico/calico/webhooks/pkg/utils"
+)
+
+// allowedServiceAccounts is the set of service account names (not fully-qualified) that are
+// permitted to write ClusterInformation. These are the Calico system components that need
+// to create and update ClusterInformation during normal operation.
+var allowedServiceAccounts = map[string]struct{}{
+	"calico-node":             {},
+	"calico-kube-controllers": {},
```
caseydavenport, line 37: Do we need Typha here? Should do a review of who writes to this.

webhooks/pkg/clusterinfo/clusterinfo.go
```diff
@@ -0,0 +1,87 @@
...
+// "system:serviceaccount:<namespace>:<name>".
+func isAllowedUser(username string) bool {
+	parts := strings.Split(username, ":")
+	if len(parts) != 4 || parts[0] != "system" || parts[1] != "serviceaccount" {
+		return false
+	}
+	saName := parts[3]
+	_, ok := allowedServiceAccounts[saName]
+	return ok
+}
```
coutinhop: is there any benefit to checking the namespace (`parts[2]`) here as well?

PR #12013, opened 2026-03-06
kube-controllers: eliminate internal config types, use v3 API directly

caseydavenport:
## Description

The kube-controllers config system has been maintaining two parallel type hierarchies — the v3 API types (`v3.NodeControllerConfig`, `v3.PolicyControllerConfig`, etc.) and an internal set of `RunConfig` types (`config.NodeControllerConfig`, `config.GenericControllerConfig`, `config.AutoHostEndpointConfig`, etc.) — with ~400 lines of merge code mapping between them. This made the config really hard to read and modify, and meant every new config field needed to be plumbed through both representations.

This PR eliminates the internal types entirely and drives all controller configuration directly off `v3.KubeControllersConfigurationSpec`. The key principle is that all user-facing configuration should be expressible in the v3 API structure, and there's no reason to maintain a second representation of the same data.

### What changed

- **`runconfig.go`**: Deleted all internal config types (`RunConfig`, `ControllersConfig`, `GenericControllerConfig`, `NodeControllerConfig`, `AutoHostEndpointConfig`, `LoadBalancerControllerConfig`, `MigrationControllerConfig`). Replaced `mergeConfig()` and its 8 helper functions with a single `resolveSpec()` that takes the API spec, applies env var overrides, fills defaults, and returns a fully-resolved `v3.KubeControllersConfigurationSpec`. Status reporting becomes trivial — `status.RunningConfig = &resolved`.
- **`main.go`**: `controllerControl.restart` channel now carries `v3.KubeControllersConfigurationSpec` directly. Controllers are instantiated with v3 types.
- **Controllers**: Controllers that took `config.GenericControllerConfig` now just take `time.Duration` (the reconciler period). The node controller takes `v3.NodeControllerConfig` + a `deleteNodes bool` (derived from datastore type at the call site, since it's a runtime decision not user config). The loadbalancer controller takes `v3.LoadBalancerControllerConfig`.
- **`config.go`**: Removed `NumberOfWorkers` fields and env var parsing — these were hardcoded to 1 everywhere and mostly dead code.

Related design doc: https://github.com/tigera/designs/pull/4


kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -185,58 +186,58 @@ func main() {
...
-		runCfg.LogLevelScreen = logLevel
+		// Set some global defaults for flannelmigration.
+		// Note that we now ignore the HEALTH_ENABLED environment variable in the case of flannel migration.
+		runCfg.HealthChecks = v3.Enabled
+		runCfg.LogSeverityScreen = logLevel.String()
 
-		// this channel will never receive, and thus flannelmigration will never
+		// This channel will never receive, and thus flannelmigration will never
 		// restart due to a config change.
-		controllerCtrl.restart = make(chan config.RunConfig)
```
caseydavenport: "controllerCtrl" is a mouthful. We can probably rename this to "manager" or "framework" or something more indicative of what this is.

kube-controllers/pkg/config/runconfig.go
```diff
@@ -39,78 +39,19 @@ var title = cases.Title(language.English)
...
-type LoadBalancerControllerConfig struct {
-	// AssignIPs indicates if LoadBalancer controller will auto-assign all ip addresses or only if asked to do so via annotation
-	AssignIPs v3.AssignIPs
-}
+	defaultReconcilerPeriod = 5 * time.Minute
+	defaultCompactionPeriod = 10 * time.Minute
+	defaultLeakGracePeriod  = 15 * time.Minute
+)
 
 type RunConfigController struct {
```
caseydavenport:
```suggestion
type ConfigurationMonitor struct {
```

PR #12016, opened 2026-03-06
Refactor kind cluster setup out of node/ into shared infra

caseydavenport:
Moves the kind cluster setup infrastructure out of `node/tests/k8st/` into shared locations (`hack/test/kind/` and `lib.Makefile`) so it can be reused by other components' test suites (e.g., `e2e/`). Also adds incremental image loading to avoid re-transferring unchanged images onto the cluster.

Note: the dependency tracking improvements (deps.txt, IMAGE_DEPS, SRC_FILES, register fixes) that were originally part of this PR have already been merged separately via #12037. This PR now only contains the kind cluster refactoring.

### Shared infrastructure

The kind cluster creation scripts, helm values, and supporting manifests were previously buried in `node/tests/k8st/`. This PR moves them to `hack/test/kind/` and `hack/test/kind/infra/`, and moves the kind-related Make targets into `lib.Makefile` so any component can use them.

The old `node/tests/k8st/` scripts and `infra/` directory are deleted — `node/Makefile` now delegates to the shared targets.

### Kind target naming

All kind-related targets now live under a clear `kind-` namespace:

| Target | Description |
|--------|-------------|
| `kind-build-images` | Build all container images needed for the kind cluster |
| `kind-cluster-create` | Create the kind cluster (no images, no Calico) |
| `kind-cluster-destroy` | Tear down the kind cluster |
| `kind-deploy` | Load images + install Calico + wait for readiness |
| `kind-up` | Build images + deploy (full cluster bringup) |
| `kind-down` | Alias for `kind-cluster-destroy` |
| `kind-reload` | Reload images onto an existing cluster (incremental) |

### Incremental image loading

The image loading script (`hack/test/kind/load_images.sh`) compares local Docker image IDs against what's already on the cluster (via `crictl images` on the kind node) and only loads images that have actually changed. This makes the `kind-reload` / `kind-deploy` cycle much faster during iterative development.

**Benchmarks** (14 images, ~1.4GB combined):

| Scenario | Time |
|----------|------|
| All images changed (cold load) | ~54s |
| No images changed | ~1s |
| Single image changed (e.g., rebuilt felix) | ~10-15s |

### Other cleanup

- Removed unused `test-webserver.yaml` deployment from kind cluster setup (wasn't referenced by any test code)
- Fixed kube-controllers wrapper binary not rebuilding on source changes
- `KIND_NAME` is now consistently passed through to all scripts that interact with the cluster

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,7 +189,9 @@ CLUSTER_ROUTING ?= BIRD
 
 ## Build all test images, create a kind cluster, and deploy Calico on it.
 .PHONY: kind-up
-kind-up: kind-build-images kind-deploy
+kind-up: kind-build-images
+	$(MAKE) kind-cluster-create CALICO_API_GROUP=$(KIND_CALICO_API_GROUP)
+	$(MAKE) kind-deploy
 
 ## Create a kind cluster and run all e2e tests.
 e2e-test:
--- a/lib.Makefile
+++ b/lib.Makefile
@@ -1612,11 +1612,13 @@ kind-build-images: $(KIND_IMAGE_MARKERS) $(REPO_ROOT)/.stamp.operator
 # Create a kind cluster and deploy Calico on it via Helm. Assumes images are
 # already built and tagged as test-build in the local Docker daemon. If a
 # cluster already exists (stamp file present), the creation step is skipped.
-# Default to v3 CRDs for kind clusters. Override with CALICO_API_GROUP=crd.projectcalico.org/v1 if needed.
+# Default to v3 CRDs for kind clusters. Override with KIND_CALICO_API_GROUP=crd.projectcalico.org/v1 if needed.
 KIND_CALICO_API_GROUP ?= projectcalico.org/v3
 
+# Load images, install Calico via Helm, and wait for readiness on an existing
+# kind cluster. Use kind-up for end-to-end bringup (images + cluster + deploy).
 .PHONY: kind-deploy
-kind-deploy: kind-cluster-create
+kind-deploy:
 	$(MAKE) -C $(REPO_ROOT) chart CALICO_API_GROUP=$(KIND_CALICO_API_GROUP)
 	REPO_ROOT=$(REPO_ROOT) \
 	KUBECONFIG=$(KIND_KUBECONFIG) \


lib.Makefile
```diff
@@ -1488,6 +1488,36 @@ helm: bin/helm
...
+
+# Helm chart tarballs needed for kind cluster setup.
+KIND_SETUP_CHARTS=$(REPO_ROOT)/bin/tigera-operator-$(GIT_VERSION).tgz \
+	$(REPO_ROOT)/bin/crd.projectcalico.org.v1-$(GIT_VERSION).tgz \
+	$(REPO_ROOT)/bin/projectcalico.org.v3-$(GIT_VERSION).tgz
+
+# Deploy Calico on an existing kind cluster. Assumes images are already built
+# and tagged as test-build in the local Docker daemon.
+.PHONY: kind-k8st-setup
+kind-k8st-setup: $(KIND_SETUP_CHARTS) kind-cluster-create
```
caseydavenport:
```suggestion
kind-setup: $(KIND_SETUP_CHARTS) kind-cluster-create
```

^ Not specific to k8st

Makefile
```diff
@@ -188,17 +188,24 @@ K8S_NETPOL_SUPPORTED_FEATURES ?= "ClusterNetworkPolicy"
 K8S_NETPOL_UNSUPPORTED_FEATURES ?= ""
 CLUSTER_ROUTING ?= BIRD
 
+## Build all test images, tag them as test-build, and set up a kind cluster with Calico.
+kind-test-cluster: kind-test-images kind-k8st-setup
+
+## Build and tag all component images needed for kind cluster testing.
+kind-test-images:
+	$(MAKE) -C node kind-test-images
```
caseydavenport: Would like to move the `kind-test-images` build logic out of node entirely. Probably better in lib.Makefile or a new Makefile in the `hack/test/kind` directory.

hack/test/kind/deploy_resources_on_kind_cluster.sh
```diff
@@ -48,7 +88,6 @@ function collect_diags() {
     fi
   done
 
-  # Also check for Running pods that aren't fully ready (e.g., 0/1, 1/2).
```
caseydavenport, line 51: These comments are useful - we should keep them.

hack/test/kind/deploy_resources_on_kind_cluster.sh
```diff
@@ -145,7 +169,6 @@ if ! ( ${kubectl} wait --for=create --timeout=60s tigerastatus/calico &&
 fi
 
 # Wait for the Calico API server to be available, if not using the projectcalico.org/v3 CRDs.
-# If using the projectcalico.org/v3 CRDs, there is no Calico API server to wait for.
```
caseydavenport, line 148: This one too

hack/test/kind/deploy_resources.sh
```diff
@@ -167,29 +190,26 @@ if [[ "$CLUSTER_ROUTING" == "FELIX" ]]; then
...
-${kubectl} apply -f tests/k8st/infra/test-webserver.yaml
+${kubectl} apply -f ${INFRA_DIR}/test-webserver.yaml
 
 echo "Wait for client and webserver pods to be ready..."
 wait_pod_ready -l pod-name=client
 wait_pod_ready -l app=webserver
 echo "client and webserver pods are running."
 echo
 
-# Show all the pods running for diags purposes.
```
caseydavenport, line 188: And this

lib.Makefile
```diff
@@ -1488,6 +1488,129 @@ helm: bin/helm
...
+
+# Helm chart tarballs needed for kind cluster setup.
+KIND_SETUP_CHARTS=$(REPO_ROOT)/bin/tigera-operator-$(GIT_VERSION).tgz \
+	$(REPO_ROOT)/bin/crd.projectcalico.org.v1-$(GIT_VERSION).tgz \
+	$(REPO_ROOT)/bin/projectcalico.org.v3-$(GIT_VERSION).tgz
+
+# Deploy Calico on an existing kind cluster. Assumes images are already built
+# and tagged as test-build in the local Docker daemon.
+.PHONY: kind-setup
+kind-setup: $(KIND_SETUP_CHARTS) kind-cluster-create
```
caseydavenport: This claims to deploy on an existing cluster, but it also depends on `kind-cluster-create` so will deploy one if there isn't one running. The comment is wrong, but perhaps we should have two targets?

I think it's getting a bit confusing, with all of these doing something slightly different, but with very similar names. 

- kind-setup
- kind-cluster-create
- kind-test-cluster

Similarly, `load-container-images` and `deploy_resources_on_kind_cluster.sh` are similarly named but do different / overlapping things. We should clean up the terminology to be really crisp.

PR #12019, opened 2026-03-06
Fix VXLAN device FlowBased mode detection in vxlanLinksIncompat

tomastigera:
## Summary
- Add explicit `FlowBased` flag check as the first comparison in `vxlanLinksIncompat`, since it is the most fundamental mode difference between eBPF and iptables/nftables dataplanes.
- Without this, switching dataplanes produced a misleading "vni: 0 vs 4096" message instead of a clear "flow-based mode" incompatibility.
- Add `DescribeTable` test covering identical devices, flow-based vs legacy mismatches, and other field mismatches (VNI, port, GBP, L2miss).

## Test plan
- [x] New `vxlanLinksIncompat` table-driven unit tests pass (8/8)
- [ ] Existing VXLAN manager tests still pass
- [ ] Manual verification: switch Felix from iptables to BPF dataplane and confirm VXLAN device is properly recreated

### DIFF
--- a/felix/dataplane/linux/vxlan_mgr.go
+++ b/felix/dataplane/linux/vxlan_mgr.go
@@ -352,6 +352,14 @@ func vxlanLinksIncompat(l1, l2 netlink.Link) string {
 	v1 := l1.(*netlink.Vxlan)
 	v2 := l2.(*netlink.Vxlan)
 
+	// FlowBased is the most fundamental mode difference — a flow-based device
+	// (used by the BPF dataplane with BTF) has no VNI, VtepDevIndex, or SrcAddr
+	// set on the device; the BPF program sets the tunnel key per-packet.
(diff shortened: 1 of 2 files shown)


felix/dataplane/linux/vxlan_mgr.go
coutinhop, line 347: not exactly related to these changes, but how about either
- printing both v1 and v2 for debugging so that when multiple values are different; or
- accumulating all differences and returning all of them?

coutinhop (approved):
lgtm, one suggestion/random thought (feel free to ignore), we should add a release note to the PR

felix/dataplane/linux/vxlan_mgr.go
tomastigera: tried it after your commend, I think failing on one thing and reporting one thing at a time is more pointed

PR #12022, opened 2026-03-06
Reduce Felix FV test runtime by removing unnecessary waits

caseydavenport:
Felix FVs have been taking longer and occasionally timing out on Semaphore. Tracked down several sources of unnecessary waiting and fixed them.

**Combine sequential Consistently blocks in staged flow log tests.** There were 8 occurrences where 2-4 sequential `Consistently(..., "5s")` blocks ran back-to-back, each sleeping for the full 5s. In BPF mode, 4 sequential checks meant 20s of waiting per occurrence. Combined each group into a single `Consistently` with a closure that checks all conditions in one poll iteration, bringing each down to 5s. This also fixes a correctness bug — `bpfDumpPolicy()` returns a `string`, not a `func() string`, so Gomega was wrapping a static value and re-checking it without ever re-polling. The new closures actually re-execute the BPF dump on each iteration.

**Remove unnecessary `time.Sleep` calls in staged flow log helpers.** `configureStagedAllow` had a 3s sleep and `configureStagedAllow`/`configureStagedPass` in the pending-policies suite had 5s sleeps, all placed after `Consistently` blocks that already waited 5s. No documented purpose — just belt-and-suspenders delays.

**Reduce QoS `Consistently` timeouts from 10s to 5s.** All 19 `Consistently` blocks in `qos_controls_test.go` verified absence of config on the "other" direction (e.g., no egress rules after setting ingress). Since these follow an `Eventually` that already confirmed the related config was applied, Felix has completed its programming pass and 5s is plenty.

**Use exponential backoff for node creation retries.** The node creation loop in `infra_k8s.go` used a fixed 3s sleep between attempts. Switched to the existing `retryBackoff` helper (100ms → 5s cap), consistent with the other retry loops in the file.

**Remove redundant 5s sleep from wireguard egress policy test.** There was a `time.Sleep(5s)` immediately before an `Eventually` that polls for the same condition (policy applied in datastore). The sleep is pure waste since the `Eventually` already handles the wait. Bumped the `Eventually` timeout from 5s to 10s to compensate.

**Poll BPF NAT map instead of sleeping in external IP overlap test.** The "should handle temporary overlap of external IPs" test had a fixed `time.Sleep(20s)` after deleting a service, waiting for Felix to process the deletion. Replaced with an `Eventually` that polls the BPF NAT frontend map for the deleted service's cluster IP key to be removed. This runs across 12 test parameter combinations, so the savings add up (~200s total).

**Increase BPF device MTU check timeout in MTU FV test.** The `expectMTU` helper checked `bpfin.cali`/`bpfout.cali` MTU with a 5s timeout, but the MTU file is written early in `NewInternalDataplane` while the BPF devices aren't created until `NewBPFEndpointManager` runs later. Under CI load (especially with IPv6 and its larger set of BPF maps), the init gap can exceed 5s. Bumped to 30s.

Rough estimate of total savings: ~7 minutes per FV run in BPF mode, ~2 minutes in iptables mode.

### DIFF
--- a/guardian/pkg/tunnel/dial_test.go
+++ b/guardian/pkg/tunnel/dial_test.go
@@ -31,13 +31,11 @@ import (
 	"github.com/projectcalico/calico/guardian/test/utils"
 )
 
-// handleConnection accepts a connection from the listener, sets up a yamux session,
-// reads a message from the client, and writes a response. Errors are reported via
-// errCh rather than calling Expect() (which would panic from a non-test goroutine).
-// doneCh is closed when the function completes all its work.
-func handleConnection(t *testing.T, listener net.Listener, errCh chan<- error, doneCh chan<- struct{}) {
-	defer close(doneCh)
-
+// handleConnection accepts a connection, sets up a yamux session, reads a
+// message from the client, writes a response, then waits for clientDoneCh
+// before returning. This ensures the server doesn't tear down its session
+// while the client is still reading.
+func handleConnection(t *testing.T, listener net.Listener, errCh chan<- error, clientDoneCh <-chan struct{}) {
 	conn, err := listener.Accept()
 	if err != nil {
 		errCh <- fmt.Errorf("Accept: %v", err)
@@ -78,6 +76,10 @@ func handleConnection(t *testing.T, listener net.Listener, errCh chan<- error, d
 		errCh <- fmt.Errorf("stream.Write: %v", err)
 		return
 	}
+
+	// Wait for the client to finish before returning, so our deferred
+	// session/stream/conn close doesn't race with the client's read.
+	<-clientDoneCh
 }
 
 func TestDial(t *testing.T) {
@@ -89,13 +91,13 @@ func TestDial(t *testing.T) {
 		defer func() { _ = listener.Close() }()
 
 		errCh := make(chan error, 1)
-		doneCh := make(chan struct{})
-		go handleConnection(t, listener, errCh, doneCh)
+		clientDoneCh := make(chan struct{})
+		go handleConnection(t, listener, errCh, clientDoneCh)
 
-		dialer, err := tunnel.NewTLSSessionDialer(listener.Addr().String(), nil)
+		dialer, err := tunnel.NewTLSSessionDialer(listener.Addr().String(), nil, tunnel.WithDialerKeepAliveSettings(false, time.Second))
 		Expect(err).NotTo(HaveOccurred())
 
-		assertExpectations(t, dialer, doneCh)
+		assertExpectations(t, dialer, clientDoneCh)
 
 		select {
 		case err := <-errCh:
@@ -124,8 +126,8 @@ func TestDial(t *testing.T) {
 		defer func() { _ = listener.Close() }()
 
 		errCh := make(chan error, 1)
-		doneCh := make(chan struct{})
-		go handleConnection(t, listener, errCh, doneCh)
+		clientDoneCh := make(chan struct{})
+		go handleConnection(t, listener, errCh, clientDoneCh)
 
 		certPool := x509.NewCertPool()
 		caCert, err := os.ReadFile(serverCrt.Name())
@@ -134,9 +136,9 @@ func TestDial(t *testing.T) {
 
 		dialer, err := tunnel.NewTLSSessionDialer(listener.Addr().String(), &tls.Config{
 			RootCAs: certPool,
-		})
+		}, tunnel.WithDialerKeepAliveSettings(false, time.Second))
 		Expect(err).NotTo(HaveOccurred())
-		assertExpectations(t, dialer, doneCh)
+		assertExpectations(t, dialer, clientDoneCh)
 
 		select {
 		case err := <-errCh:
@@ -146,7 +148,7 @@ func TestDial(t *testing.T) {
 	})
 }
 
-func assertExpectations(t *testing.T, dialer tunnel.SessionDialer, serverDone <-chan struct{}) {
+func assertExpectations(t *testing.T, dialer tunnel.SessionDialer, clientDoneCh chan<- struct{}) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


felix/fv/flow_logs_goldmane_staged_test.go
```diff
@@ -821,18 +856,36 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ aggregation of flow log wit
...
 			bpfWaitForNetworkPolicy(tc.Felixes[0], ep1_1.InterfaceName, "ingress", "default", "tier1.np1-1")
 			bpfWaitForNetworkPolicy(tc.Felixes[1], ep2_1.InterfaceName, "egress", "default", "tier1.np1-1")
 			bpfWaitForNetworkPolicy(tc.Felixes[1], ep2_1.InterfaceName, "ingress", "default", "tier1.np1-1")
 			// When policies are programmed, make sure no staged policy is programmed. Staged policies must be skipped.
-			Consistently(bpfDumpPolicy(tc.Felixes[0], ep1_1.InterfaceName, "ingress"), "5s", "1s").ShouldNot(ContainSubstring("staged"))
-			Consistently(bpfDumpPolicy(tc.Felixes[0], ep1_1.InterfaceName, "egress"), "5s", "1s").ShouldNot(ContainSubstring("staged"))
-			Consistently(bpfDumpPolicy(tc.Felixes[1], ep2_1.InterfaceName, "ingress"), "5s", "1s").ShouldNot(ContainSubstring("staged"))
-			Consistently(bpfDumpPolicy(tc.Felixes[1], ep2_1.InterfaceName, "egress"), "5s", "1s").ShouldNot(ContainSubstring("staged"))
+			Consistently(func() error {
+				for _, c := range []struct {
```
caseydavenport: This is redefined a lot - might be worth have an actual local struct type to avoid this ugliness.

felix/fv/qos_controls_test.go
```diff
@@ -359,7 +359,7 @@ var _ = infrastructure.DatastoreDescribe(
 						// ingress config should be present
 						Eventually(getQdisc, "10s", "1s").Should(MatchRegexp(`qdisc tbf \d+: dev ` + regexp.QuoteMeta(w[1].InterfaceName) + ` root refcnt \d+ rate ` + regexp.QuoteMeta("10Mbit") + `.* peakrate ` + regexp.QuoteMeta("100Mbit")))
 						// egress config should not be present
-						Consistently(getQdisc, "10s", "1s").ShouldNot(MatchRegexp(`qdisc tbf \d+: dev bwcali.* root refcnt \d+ rate ` + regexp.QuoteMeta("10Mbit")))
```
caseydavenport, line 362: Is 10s meaningful in some way? i.e., does this need to be this large due to some retry timer or something? Or is it arbitrary? @coutinhop

felix/fv/infrastructure/infra_k8s.go
```diff
@@ -832,11 +832,12 @@ func (kds *K8sDatastoreInfra) AddNode(felix *Felix, v4CIDR *net.IPNet, v6CIDR *n
 	log.WithField("nodeIn", nodeIn).Debug("Node defined")
 	var nodeOut *v1.Node
 	var err error
+	delay := 100 * time.Millisecond
 	for i := range 5 {
 		nodeOut, err = kds.K8sClient.CoreV1().Nodes().Create(context.Background(), nodeIn, metav1.CreateOptions{})
 		if err != nil {
 			log.WithError(err).WithField("try number", i).Debug("Error creating node")
-			time.Sleep(3 * time.Second)
+			delay = retryBackoff(delay, 5*time.Second)
```
nelljerram, line 840: IIUC we now only allow 3.2s overall, compared to 15s previously.  Is 3.2s always enough?

felix/fv/flow_logs_goldmane_staged_test.go
```diff
@@ -346,10 +351,22 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ goldmane flow log with stag
...
+					f     *infrastructure.Felix
+					iface string
+				}{
+					{tc.Felixes[0], ep1_1.InterfaceName},
+					{tc.Felixes[1], ep2_1.InterfaceName},
+				} {
+					for _, hook := range []string{"ingress", "egress"} {
+						if strings.Contains(bpfDumpPolicy(c.f, c.iface, hook), "staged") {
+							return fmt.Errorf("found staged policy in BPF dump for %s %s", c.iface, hook)
```
nelljerram: It's likely we'd want to know which Felix here.  So change the `for` to `for i, c := ...` and log `i` here.

nelljerram: (or use f.hostname as below)

nelljerram: + same for all similar occurrences below

felix/fv/mtu_test.go
```diff
@@ -48,11 +48,11 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ VXLAN topology before addin
 			EventuallyWithOffset(1, func() string {
 				out, _ := felix.ExecOutput("ip", "link", "show", "dev", "bpfin.cali")
 				return out
-			}, "5s", "500ms").Should(ContainSubstring(fmt.Sprintf("mtu %d", mtu)))
+			}, "30s", "500ms").Should(ContainSubstring(fmt.Sprintf("mtu %d", mtu)))
```
nelljerram: This is a huge increase; could it be smaller?  Is this a test that we've seen flaking?  Let's at least record more context for this change.

caseydavenport: Yes, sorry - this got added to this PR inadvertently. It was a flake fix for a test I have seen flaking recently, but it was a bit optimistic. I'll remove it from here for now.

felix/fv/qos_controls_test.go
coutinhop: it is arbitrary, I picked it as a reasonable interval for the config to be applied, but my gut feeling is it's much shorter than that (order of 1s or less?). I'd say it's probably fine if you want to reduce these to 5s or even 3s (if it causes increased flakiness we can always make them larger again)

PR #12023, opened 2026-03-07
ci: add watchdog to kill hung VM test batches before job timeout

tomastigera:
## Summary

Two fixes for CI test batch reliability:

- **Watchdog timer for hung batches**: `run-tests-on-vms` now starts a watchdog that fires 5 minutes before the Semaphore job timeout (`JOB_TIMEOUT_MINUTES`). It kills any still-running batch subshells, generates JUnit XML failure reports for the timed-out batches, and lets the job exit cleanly as "failed" instead of being hard-killed by Semaphore. This ensures the epilogue runs, artifacts are uploaded, and test results are visible.

- **Fix BPF UT pipeline hang**: `startBPFLogging()` was starting `bpftool prog tracelog` with inherited stdout/stderr. When the test binary runs in a pipeline (`bpf_ut.test |& gotestsum`), bpftool inherits the pipe file descriptors. If the test binary exits without killing bpftool (panic, crash, signal), bpftool keeps the pipe open and gotestsum blocks forever. Fixed by redirecting bpftool output to a file and adding a SIGKILL fallback with timeout in `stopBPFLogging()`.

## Test plan

- [x] Watchdog tested on a CI run where the UT batch hung — job exited as failed with JUnit report instead of timing out silently
- [x] BPF UT hang root-caused and fix verified: bpftool no longer inherits the pipeline pipe

### DIFF
--- a/.semaphore/vms/run-tests-on-vms
+++ b/.semaphore/vms/run-tests-on-vms
@@ -121,8 +121,20 @@ done
 monitor_pattern="${monitor_pattern:1}" # Strip leading '|'
 
 test_pid=()
+watchdog_pids=()
 monitor_pids=()
 log_files=()
+
+# Calculate watchdog timeout before starting batches.
+# JOB_TIMEOUT_MINUTES can be set to match the Semaphore execution_time_limit;
+# the watchdog fires 5 minutes before that to leave time for cleanup.
+: "${JOB_TIMEOUT_MINUTES:=60}"
+if [ "$JOB_TIMEOUT_MINUTES" -le 5 ] 2>/dev/null; then
+  echo "WARNING: JOB_TIMEOUT_MINUTES ($JOB_TIMEOUT_MINUTES) is too low for watchdog, disabling."
+  watchdog_seconds=0
+else
+  watchdog_seconds=$(( (JOB_TIMEOUT_MINUTES - 5) * 60 ))
+fi
 
 # Format batch name for log files: zero-pad numeric values to 3 digits
 format_batch_for_log() {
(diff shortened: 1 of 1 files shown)


felix/bpf/ut/bpf_prog_test.go
```diff
@@ -298,14 +299,26 @@ type testLogger interface {
...
+	f, err := os.Create("/tmp/bpf-trace.log")
+	if err != nil {
+		log.WithError(err).Warn("Failed to create BPF trace log file")
+		return nil
+	}
 	cmd := exec.Command("/usr/bin/bpftool", "prog", "tracelog")
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-	err := cmd.Start()
+	cmd.Stdout = f
```
fasaxc: In the FVs, we use a pipe and capture the logs into the log stream of the test; are you happy with sending to a temp file instead?

In any case, you should make sure this goes into the job artifacts so we have it in CI.

.semaphore/semaphore.yml.d/blocks/20-felix.yml
```diff
@@ -273,6 +273,8 @@
             value: "bpf-25.10-nft-no-fv-with-ut"
           - name: FELIX_FV_BPFATTACHTYPE
             value: "tc"
+          - name: JOB_TIMEOUT_MINUTES
```
fasaxc, line 284: Think a few more jobs need this

felix/bpf/ut/bpf_prog_test.go
tomastigera: Reverted the file-redirect approach. Instead, bpftool now keeps writing to stdout/stderr (so trace logs stay inline in the test output) and we set `Pdeathsig: syscall.SIGKILL` on the child process. This tells the kernel to automatically SIGKILL bpftool when the parent thread exits — so if the test binary panics or crashes without running `stopBPFLogging`, bpftool is killed immediately and releases the pipe FDs, letting gotestsum get EOF instead of blocking forever.

The `stopBPFLogging` SIGKILL-with-timeout fallback is kept for clean shutdown.

.semaphore/vms/run-tests-on-vms
```diff
@@ -215,6 +215,60 @@ done
...
+fi
+if [ "$watchdog_seconds" -gt 0 ]; then
+(
+  sleep "$watchdog_seconds"
+  echo
+  echo "===== WATCHDOG: approaching job timeout (${JOB_TIMEOUT_MINUTES}m), killing hung batches ====="
+  for i in "${!batches[@]}"; do
+    batch=${batches[$i]}
+    pid=${test_pid[$i]}
+    if kill -0 "$pid" 2>/dev/null; then
```
fasaxc: I think we wait on this pid elsewhere so, if job was successful and then pid was reused, you might wait on the wrong pid here.  Could we set a flag somewhere (if all else fails, a touchfile) that says "job N exited cleanly" and skip clean jobs explicitly?

fasaxc: Or, start one watchdog process per job and have the per-job logic kill the job-specific watchdog after the job finishes.

PR #12024, opened 2026-03-07
Reduce node k8st test runtime by fixing unnecessary waiting

caseydavenport:
The k8st tests have accumulated a bunch of places where they wait way longer than necessary, do redundant work on every test method, or just have overly generous timeouts for a kind cluster. The cumulative effect is pretty rough — the full suite was taking ~28 minutes when it really doesn't need to. This gets it down to ~11 minutes.

## Description

Changes roughly grouped by category:

**Move expensive setup to `setUpClass`**

Several test classes were creating and destroying Docker containers with external BIRD instances on every single test method. Moving that to `setUpClass`/`tearDownClass` does it once per class instead:

- `test_bgp_filter.py`: 9 tests were each creating two external Docker containers (v4 + v6) with `docker run` + `apk install` + BIRD config — 18 container lifecycles total, now 2.
- `test_bgp_advert.py` / `test_bgp_advert_v6.py`: Same pattern — external node + BGPPeer + Secret creation moved to class-level setup. Also removed a redundant `wait_for_deployment` call in `test_node_exclusion` that was waiting on a deployment that hadn't been scaled.
- `test_ipip_spoofing.py`: Namespace + pod creation shared across both spoof tests, eliminating one redundant namespace lifecycle and calico-node restart cycle.

**Fix `DiagsCollector` to only collect on failure**

`DiagsCollector` was unconditionally collecting full cluster diagnostics after every test — `kubectl get` across all namespaces, `calicoctl get` for 5+ resource types, `docker exec` on all nodes, `kubectl logs` for calico-node, and 6 confd template files per node. Now it only fires on test failure.

**Remove redundant retry loops**

- BGP advert tests had `for i in range(10)` loops wrapping `retry_until_success(curl, ...)` — but `retry_until_success` already retries up to 90 times. One call is enough.
- The v6 curl path was using `retries=200, wait_time=5` which could block for 16+ minutes on a single attempt.

**Reduce excessive timeouts and polling intervals**

- `wait_ready`/`wait_not_ready`: 300s → 60s (pods on kind don't need 5 minutes)
- `delete_and_confirm`: 10s polling → 2s polling
- `update_ds_env`: poll every 3s with 120s timeout instead of 10s polling with no timeout
- External node container startup: 20s sleep → 2s sleep with a 30s retry cap
- BGP filter route checks: 270s max → 30s max

**Reduce test scope where overkill**

- `test_graceful_restart`: 8 iterations of delete-and-wait-for-calico-node → 3. Each iteration takes ~36s (calico-node startup time), and 3 is plenty to verify no route churn during graceful restart.
- `test_many_services`: 300 services → 50. Still exercises the bulk-create-and-verify path without spending a minute creating services.
- `test_methodology`: removed an unconditional `time.sleep(5)` after a `retry_until_success` that already confirmed BIRD was running.

**Remove unused namespace from BGP filter tests**

The 9 `test_bgp_filter` tests were each creating and deleting a namespace that nothing was ever deployed into. Pure overhead — ~4s per test for no reason.

**Fix pre-existing `TestLocalBGPPeerRR` regex bug**

The regex in `assertRegex` was hardcoding `Node_172_18_0_` but kind clusters can use different subnets (e.g., 172.19.x.x). Changed to match `Node_` followed by anything.

### DIFF
--- a/node/tests/k8st/test_base.py
+++ b/node/tests/k8st/test_base.py
@@ -230,37 +230,6 @@ class TestBase(TestCase):
     def get_routes(self):
         return run("docker exec kube-node-extra ip r")
 
-    def update_ds_env(self, ds, ns, env_vars):
-        config.load_kube_config(os.environ.get('KUBECONFIG'))
-        api = client.AppsV1Api(client.ApiClient())
-        node_ds = api.read_namespaced_daemon_set(ds, ns, exact=True, export=False)
-        for container in node_ds.spec.template.spec.containers:
-            if container.name == ds:
-                for k, v in env_vars.items():
-                    logger.info("Set %s=%s", k, v)
-                    env_present = False
-                    for env in container.env:
-                        if env.name == k:
-                            env_present = True
-                    if not env_present:
-                        v1_ev = client.V1EnvVar(name=k, value=v, value_from=None)
-                        container.env.append(v1_ev)
-        api.replace_namespaced_daemon_set(ds, ns, node_ds)
-
-        # Wait until the DaemonSet reports that all nodes have been updated.
-        start = time.time()
-        while time.time() - start < 120:
-            node_ds = api.read_namespaced_daemon_set_status(ds, ns)
-            logger.info("%d/%d nodes updated",
-                      node_ds.status.updated_number_scheduled,
-                      node_ds.status.desired_number_scheduled)
-            if node_ds.status.updated_number_scheduled == node_ds.status.desired_number_scheduled:
-                break
-            time.sleep(3)
-        else:
-            raise Exception("DaemonSet %s/%s failed to roll out within 120s" % (ns, ds))
-
-
     def scale_deployment(self, deployment, ns, replicas):
         return kubectl("scale deployment %s -n %s --replicas %s" %
                        (deployment, ns, replicas)).strip()
--- a/node/tests/k8st/utils/utils.py
+++ b/node/tests/k8st/utils/utils.py
@@ -95,15 +95,7 @@ def start_external_node_with_bgp(name, bird_peer_config=None, bird6_peer_config=
     # Check how much space there is inside the container.  We may need
     # to retry this, as it may take a while for the image to download
     # and the container to start running.
-    for attempt in range(30):
-        try:
-            run("docker exec %s df -h" % name)
-            break
-        except subprocess.CalledProcessError:
-            _log.info("Container not ready yet (attempt %d)", attempt)
-            time.sleep(2)
-    else:
-        raise Exception("Container %s failed to start after 30 attempts" % name)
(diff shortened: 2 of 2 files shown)


node/tests/k8st/tests/test_bgp_filter.py
```diff
@@ -82,45 +78,26 @@ def setUp(self):
...
-            raise teardown_exception
-
-    def _get_bird_conf(self, ipv6=False):
-        if ipv6:
-            return bird_conf % (self.egress_node_ip6)
-
-        return bird_conf % (self.egress_node_ip)
+    def setUp(self):
+        super(TestBGPFilter, self).setUp()
+        self.egress_calico_pod = self.get_calico_node_pod(self.egress_node)
```
matthewdupre, line 100: This previously created a namespace "bgpfilter-test" (and tore it down in tearDown()).  Is there a reason to remove that?

node/tests/k8st/utils/utils.py
```diff
@@ -91,13 +95,15 @@ def start_external_node_with_bgp(name, bird_peer_config=None, bird6_peer_config=
     # Check how much space there is inside the container.  We may need
     # to retry this, as it may take a while for the image to download
     # and the container to start running.
-    while True:
+    for attempt in range(30):
```
matthewdupre: Should use retry_until_success

node/tests/k8st/test_base.py
```diff
@@ -222,14 +222,45 @@ def is_it_gone_yet(res_name, res_type):
...
-        retry_until_success(is_it_gone_yet, retries=10, wait_time=10, function_args=[name, resource_type])
+        retry_until_success(is_it_gone_yet, retries=30, wait_time=2, function_args=[name, resource_type])
 
     class StillThere(Exception):
         pass
 
     def get_routes(self):
         return run("docker exec kube-node-extra ip r")
 
+    def update_ds_env(self, ds, ns, env_vars):
```
matthewdupre: You removed this unused method in this PR yesterday: https://github.com/projectcalico/calico/pull/12203.  It's still unused as far as I can see.

caseydavenport: whoopsie

node/tests/k8st/tests/test_bgp_filter.py
caseydavenport: The namespace was created and deleted each test but nothing was ever deployed into it - no pods, services, or network policies reference it. So it was pure overhead.

node/tests/k8st/utils/utils.py
caseydavenport: Done, switched to retry_until_success.

node/tests/k8st/test_base.py
caseydavenport: Removed, thanks for catching the reintroduction.

PR #12025, opened 2026-03-07
Lazily detect KubeVirt in kube-controllers

skoryk-oleksandr:
## Lazily detect KubeVirt in kube-controllers                                                                                                                                 
                                                                                                                                                                             
Previously, KubeVirt detection happened once at startup. If KubeVirt was installed after calico-kube-controllers was already running, KubeVirt-aware IPAM garbage collection never activated. VM allocations would be validated using the regular pod-based path, which doesn't account for VM lifecycle events like live migration or VM restart.                                                                                                                                                                 
                                                                                                                                                                             
Now kube-controllers polls for KubeVirt in a background goroutine. Once detected, it creates and syncs VM/VMI informers and atomically publishes their indexers to the IPAM controller. No restart required.
                                                                                                                                                                             
## Changes                                                                                                                                                                  

  - Add `KubeVirtState` with `atomic.Pointer` for **thread-safe** lazy access to VM/VMI indexers                                                                                     
  - Add InitInformers function that tries immediately, then polls every **30s** until KubeVirt appears
  - Replace direct `vmIndexer/vmiIndexer` fields in IPAMController with `*KubeVirtState`                                                                                         
  - Remove `VM/VMI` informer sync blocking from NodeController.Run()                                                                                                           
  - Move `KubeVirt` informer lifecycle out of controllerControl into the InitInformers goroutine                                                                               
  - Accept InformerFactory function for testability                                                                                                                          
                                                         

<!-- For any user-visible change, replace TBD with a one-line description. -->
**Release note:**
```release-note
TBD
```

### DIFF
--- a/libcalico-go/lib/kubevirt/state.go
+++ b/libcalico-go/lib/kubevirt/state.go
@@ -27,9 +27,9 @@ type kvIndexers struct {
 	vmi cache.Indexer
 }
 
-// GetIndexerFunc creates KubeVirt VM and VMI informers.
+// GetIndexInformerFunc creates KubeVirt VM and VMI informers.
 // Returns (nil, nil, nil) when KubeVirt is not installed.
-type GetIndexerFunc func() (cache.SharedIndexInformer, cache.SharedIndexInformer, error)
+type GetIndexInformerFunc func() (cache.SharedIndexInformer, cache.SharedIndexInformer, error)
 
 // DeferredInformers provides thread-safe access to KubeVirt cache indexers
 // that may be populated lazily if KubeVirt is installed after startup.
@@ -42,7 +42,7 @@ type DeferredInformers struct {
 // goroutine that polls for KubeVirt, creates informers, and publishes their
 // indexers once synced. If KubeVirt is already installed, the informers are
 // created immediately.
-func NewDeferredInformers(createInformers GetIndexerFunc, pollInterval time.Duration, stop <-chan struct{}) *DeferredInformers {
+func NewDeferredInformers(createInformers GetIndexInformerFunc, pollInterval time.Duration, stop <-chan struct{}) *DeferredInformers {
 	d := &DeferredInformers{}
 	go d.initInformers(createInformers, pollInterval, stop)
 	return d
@@ -84,7 +84,7 @@ func (d *DeferredInformers) setIndexers(vm, vmi cache.Indexer) {
 // initInformers detects KubeVirt, creates VM/VMI informers, and publishes
 // their indexers. If KubeVirt is not installed at call time it polls every
 // pollInterval until it appears or stop is closed.
-func (d *DeferredInformers) initInformers(createInformers GetIndexerFunc, pollInterval time.Duration, stop <-chan struct{}) {
+func (d *DeferredInformers) initInformers(createInformers GetIndexInformerFunc, pollInterval time.Duration, stop <-chan struct{}) {
 	// Try immediately first.
 	vmInf, vmiInf, err := createInformers()
 	if err != nil {
--- a/libcalico-go/lib/kubevirt/state_test.go
+++ b/libcalico-go/lib/kubevirt/state_test.go
@@ -38,9 +38,9 @@ func newFakeInformers(stop <-chan struct{}) (cache.SharedIndexInformer, cache.Sh
 	return vmInf, vmiInf
 }
 
-// newFakeInformerFactory returns a GetIndexerFunc that returns nil for the
+// newFakeInformerFactory returns a GetIndexInformerFunc that returns nil for the
 // first nilCount calls, then returns real (fake) informers.
-func newFakeInformerFactory(nilCount int, stop <-chan struct{}) kubevirt.GetIndexerFunc {
+func newFakeInformerFactory(nilCount int, stop <-chan struct{}) kubevirt.GetIndexInformerFunc {
 	var calls atomic.Int32
 	return func() (cache.SharedIndexInformer, cache.SharedIndexInformer, error) {
 		n := int(calls.Add(1))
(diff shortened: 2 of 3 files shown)


libcalico-go/lib/kubevirt/state.go
```diff
@@ -0,0 +1,92 @@
...
+	"sync/atomic"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/tools/cache"
+)
+
+// kubeVirtIndexers holds the VM and VMI cache indexers.
+type kubeVirtIndexers struct {
+	VM  cache.Indexer
```
caseydavenport: Think these can be private fields?

libcalico-go/lib/kubevirt/state.go
```diff
@@ -0,0 +1,92 @@
...
+	vmInf, vmiInf, err := createInformers()
+	if err != nil {
+		log.WithError(err).Warn("Failed to create KubeVirt informers, will keep polling")
+	}
+
+	if vmInf == nil {
+		// KubeVirt not found — poll until it appears.
+		ticker := time.NewTicker(pollInterval)
+		defer ticker.Stop()
+		for vmInf == nil {
```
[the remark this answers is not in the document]
skoryk-oleksandr: fixed

libcalico-go/lib/kubevirt/state.go
```diff
@@ -0,0 +1,92 @@
...
+
+// kubeVirtIndexers holds the VM and VMI cache indexers.
+type kubeVirtIndexers struct {
+	VM  cache.Indexer
+	VMI cache.Indexer
+}
+
+// KubeVirtState provides thread-safe access to KubeVirt cache indexers.
+// It is safe for concurrent reads and a single writer (the lazy init goroutine).
+type KubeVirtState struct {
```
caseydavenport: A bit generic of a name, what about something like `DeferredInformer`? 

I'd probably also tidy this up with a constructor instead of asking callers to instantiate directly inline, in case we want to add new fields eventually. 

```go
func NewDeferredInformer(factory InformerFactory) *DeferredInformer
```

Then I might even move `InitInformers` to be private (`initInformers`) and start it in the constructor even, but not 100% sure about that.

libcalico-go/lib/kubevirt/state.go
```diff
@@ -0,0 +1,92 @@
...
+				}
+			}
+		}
+	}
+
+	// Start the informers and wait for initial sync.
+	go vmInf.Run(stop)
+	go vmiInf.Run(stop)
+	if !cache.WaitForCacheSync(stop, vmInf.HasSynced, vmiInf.HasSynced) {
+		log.Warn("Failed to sync KubeVirt informers")
```
caseydavenport, line 115: Hm, if we fail to sync, should we still return the indexers from Get()?

libcalico-go/lib/kubevirt/state.go
```diff
@@ -0,0 +1,92 @@
...
+
+// KubeVirtState provides thread-safe access to KubeVirt cache indexers.
+// It is safe for concurrent reads and a single writer (the lazy init goroutine).
+type KubeVirtState struct {
+	indexers atomic.Pointer[kubeVirtIndexers]
+}
+
+// Get returns the VM and VMI indexers, or (nil, nil) if KubeVirt has not been
+// detected yet.
+func (s *KubeVirtState) Get() (vm cache.Indexer, vmi cache.Indexer) {
```
caseydavenport:
```suggestion
func (s *KubeVirtState) GetIndexers() (vm cache.Indexer, vmi cache.Indexer) {
```

Or maybe even:

```go
func (s *KubeVirtState) VMIndexer() (cache.Indexer) 
func (s *KubeVirtState) VMInstanceIndexer() (cache.Indexer) 
```

^ This makes the calling code feel more self-documenting.

skoryk-oleksandr:
@caseydavenport Thanks a lot for review :+1: . I addressed your comments, see [this commit](https://github.com/projectcalico/calico/pull/12025/changes/e10099f8224da715033eefb877b9cb4cd5d356c8) . Feel free to take a look again.

libcalico-go/lib/kubevirt/state.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+type kvIndexers struct {
+	vm  cache.Indexer
+	vmi cache.Indexer
+}
+
+// InformerFactory creates KubeVirt VM and VMI informers.
+// Returns (nil, nil, nil) when KubeVirt is not installed.
+type InformerFactory func() (cache.SharedIndexInformer, cache.SharedIndexInformer, error)
```
caseydavenport: A bit of a nit, but I think we should rename this - there's already an InformerFactory type in the upstream k8s libraries, so could be confusing. 

Maybe `GetIndexersFn`?

caseydavenport: A bit of a nit, but I think we should rename this - there's already an InformerFactory type in the upstream k8s libraries, so could be confusing. 

Maybe `GetIndexersFn`?

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -517,12 +509,12 @@ func (cc *controllerControl) InitControllers(
 		cc.controllers["NetworkPolicy"] = policyController
 	}
 	if cfg.Controllers.Node != nil {
-		nodeController := node.NewNodeController(ctx, k8sClientset, calicoClient, *cfg.Controllers.Node, nodeInformer, podInformer, dataFeed, vmInformer, vmiInformer)
+		deferredInformers := kubevirt.NewDeferredInformers(func() (cache.SharedIndexInformer, cache.SharedIndexInformer, error) {
+			return kubevirt.TryCreateInformers(k8sconfig, 5*time.Minute)
```
caseydavenport: I'd probably tidy this up with a new helper in the kubevirt package:

```go
func NewIndexerFunc(rest.Config, time.Duration) GetIndexerFunc
```

Then this becomes:

```go
kubevirt.NewDeferredInformers(kubevirt.NewIndexerFunc(k8sconfig, 5*time.Minute))
```

libcalico-go/lib/kubevirt/state.go
```diff
@@ -0,0 +1,113 @@
...
+	idx := d.indexers.Load()
+	if idx == nil {
+		return nil
+	}
+	return idx.vmi
+}
+
+// SetIndexers atomically publishes the VM and VMI indexers. Exported for use
+// in tests.
+func (d *DeferredInformers) SetIndexers(vm, vmi cache.Indexer) {
```
caseydavenport: I think the only reason this is public is so it can be called from tests? If we can, would be nice to keep this package scoped - production code really shouldn't be calling it AFAIK.

skoryk-oleksandr:
@caseydavenport 
Hey Casey, thanks for the review! Addressed all three:                                                                                                                                                                               
  - Renamed InformerFactory -> GetIndexerFunc                            
  - Added NewIndexerFunc helper                                                                                                                                                                              
  - Made SetIndexers package-scoped, added a NewDeferredInformersWithIndexers constructor for tests instead

libcalico-go/lib/kubevirt/state.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+type kvIndexers struct {
+	vm  cache.Indexer
+	vmi cache.Indexer
+}
+
+// GetIndexerFunc creates KubeVirt VM and VMI informers.
+// Returns (nil, nil, nil) when KubeVirt is not installed.
+type GetIndexerFunc func() (cache.SharedIndexInformer, cache.SharedIndexInformer, error)
```
caseydavenport: I know I suggested this name, but I think it should be `GetIndexInformerFunc` since it returns IndexInformers (not sure if I just misread before or if the return type changed!)

libcalico-go/lib/kubevirt/state_test.go
```diff
@@ -0,0 +1,137 @@
...
+	It("should not publish indexers when stop is closed before KubeVirt appears", func() {
+		// Factory always returns nil — KubeVirt never appears.
+		factory := func() (cache.SharedIndexInformer, cache.SharedIndexInformer, error) {
+			return nil, nil, nil
+		}
+
+		di := kubevirt.NewDeferredInformers(factory, 10*time.Millisecond, stop)
+
+		// Let it poll a few times, then stop.
+		time.Sleep(50 * time.Millisecond)
```
caseydavenport: This test relies on `time.Sleep(50ms)` + `time.Sleep(20ms)` for synchronization, which is fragile under CI load. Consider using `Consistently` to assert the indexers remain nil over a short window, rather than sleeping for fixed durations.

Also, the `stop` reassignment on line 121 to avoid double-close in `AfterEach` is a bit subtle. A cleaner pattern might be to use a dedicated local channel for this test case and skip the `AfterEach` close entirely.

PR #12026, opened 2026-03-07
Implement tier deletion lifecycle with finalizers and webhook protection

caseydavenport:
## Description

Adds the infrastructure for graceful tier deletion, modeled after the namespace and IPPool deletion patterns. The idea is that when a user deletes a tier, we don't just yank it out from under any policies that reference it — instead, we block deletion until those policies are cleaned up.

### Webhook: block policy creation in terminating tiers

The RBAC admission webhook now checks that the target tier actually exists and isn't terminating before allowing policy CREATE/UPDATE. Previously, a user could create a policy referencing a non-existent tier and the webhook would happily allow it through.

- `checkTierExists` returns a deny response if the tier is not found or has a `DeletionTimestamp`
- DELETE operations skip this check (removing a policy from a tier is always fine)

### Tier controller with finalizers

New controller in kube-controllers that manages tier lifecycle via finalizers (same pattern as the IPPool controller):

- Adds a `projectcalico.org/tier-controller` finalizer to tiers on creation
- When a tier is deleted, uses field selectors on `spec.tier` to count remaining policies across all four policy types (GlobalNetworkPolicy, NetworkPolicy, StagedGlobalNetworkPolicy, StagedNetworkPolicy)
- Updates the tier's status with a `Ready=False/Terminating` condition showing which kinds of policies still exist, e.g. "Waiting for policies to be removed: 3 GlobalNetworkPolicies, 1 NetworkPolicy"
- Removes the finalizer once all policies are gone, allowing the tier to be fully deleted

### API changes

- Added `+kubebuilder:subresource:status` to `Tier`, with `TierStatus` containing `Conditions`
- Added a `Status` printer column showing `Ready` condition reason
- Fixed `+kubebuilder:selectablefield` marker on `GlobalNetworkPolicy` — it was on the List type instead of the resource type

### Built-in tier protection

ValidatingAdmissionPolicy + binding in `api/config/admission/` that blocks deletion of the `default`, `kube-admin`, and `kube-baseline` tiers.

Related: [CORE-12443](https://tigera.atlassian.net/browse/CORE-12443)

[CORE-12443]: https://tigera.atlassian.net/browse/CORE-12443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/kube-controllers/run-uts
+++ b/kube-controllers/run-uts
@@ -27,9 +27,15 @@ for PKG in "$@"; do
 	HAS_TESTS=$(find ${PKG} -name "ut.test")
 	if [ ! -z "${HAS_TESTS}" ]; then
 		echo "Running tests for package: ${PKG}";
-		# Generate a unique JUnit report name from the package path.
 		REPORT_NAME=$(echo "${PKG}" | tr '/./' '_' | sed 's/^_//')
-		${PKG}/ut.test --ginkgo.junit-report=report/${REPORT_NAME}.xml ${GINKGO_ARGS} || ((RC++))
+
+		# Only pass Ginkgo flags to Ginkgo-based test binaries. Vanilla go test
+		# binaries don't recognize --ginkgo.* flags and will fail immediately.
+		if ${PKG}/ut.test --help 2>&1 | grep -q ginkgo; then
+			${PKG}/ut.test --ginkgo.junit-report=report/${REPORT_NAME}.xml ${GINKGO_ARGS} || ((RC++))
+		else
+			${PKG}/ut.test -test.v || ((RC++))
+		fi
 	else
 		echo "WARNING: No tests to run in ${PKG}, skipping"
 	fi


kube-controllers/pkg/controllers/tier/tier_controller.go
```diff
@@ -0,0 +1,312 @@
...
+
+func pluralY(n int) string {
+	if n == 1 {
+		return "y"
+	}
+	return "ies"
+}
+
+// countPoliciesInTier uses field selectors to count all policies that reference the given tier.
+func (c *TierController) countPoliciesInTier(tierName string) (policyCounts, error) {
```
MichalFupso, line 308: What do you think about using the informer cache to get this info instead of calling the api? Maybe add a custom index function to the indexer to be able to filter by tier name?

kube-controllers/pkg/controllers/tier/tier_controller.go
```diff
@@ -0,0 +1,312 @@
...
+		logrus.WithField("type", fmt.Sprintf("%T", obj)).Warn("Unknown policy type in tier controller")
+		return ""
+	}
+}
+
+// reconcileTierByName fetches the tier from the API and reconciles it.
+func (c *TierController) reconcileTierByName(name string) {
+	logCtx := logrus.WithField("name", name)
+	tier, err := c.cli.ProjectcalicoV3().Tiers().Get(c.ctx, name, metav1.GetOptions{})
+	if err != nil {
```
MichalFupso: Nit: I think we should distinguish between notFound Error and general error. On general error I think we could change the log level to error

kube-controllers/pkg/controllers/tier/tier_controller.go
```diff
@@ -0,0 +1,312 @@
...
+	if _, err := tierInformer.AddEventHandler(tierHandlers); err != nil {
+		logrus.WithError(err).Fatal("Failed to register event handler for Tier")
+	}
+
+	// Watch policy resources so that when a policy is deleted, we re-reconcile the
+	// owning tier (which may now be ready to have its finalizer removed).
+	policyHandlers := cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj any) {
+			tierName := tierNameFromPolicy(obj)
+			if tierName == "" {
```
MichalFupso, line 125: I think this is worth logging an error for, policy should always belong to a tier

kube-controllers/pkg/controllers/tier/tier_controller.go
caseydavenport: Yeah, good call. Switched to using a `"byTier"` cache index on each policy informer — `countPoliciesInTier` now does `ByIndex` lookups and type-switches the cached objects to count per-kind. No more API List calls.

kube-controllers/pkg/controllers/tier/tier_controller.go
caseydavenport: Yep, done — not-found now logs at Debug (expected when tier is already gone), everything else at Error.

libcalico-go/lib/names/policy.go
```diff
@@ -32,6 +34,28 @@ func TierOrDefault(tier string) string {
 	}
 }
 
+// TierFromPolicy extracts the tier name from a policy object using reflection.
+// It looks for a Spec.Tier string field, which all Calico policy types have
+// (GlobalNetworkPolicy, NetworkPolicy, StagedGlobalNetworkPolicy,
+// StagedNetworkPolicy, StagedKubernetesNetworkPolicy). Returns the tier name
```
caseydavenport: StagedKubernetesNetworkPolicy doesn't have a tier field!

PR #12030, opened 2026-03-07
Reduce libcalico-go test suite runtime

caseydavenport:
Reduces the libcalico-go FV and UT test suite runtime through several targeted fixes.

### FV: Replace `time.Sleep` with `Eventually` polling (clientv3 suite, ~52s saved)

The old pattern for testing etcd TTL expiry slept fixed durations:

```go
time.Sleep(1 * time.Second)  // wait for key to still exist
_, err = c.Resource().Get(ctx, name, options.GetOptions{})
time.Sleep(2 * time.Second)  // wait for TTL expiry
_, err = c.Resource().Get(ctx, name, options.GetOptions{})
Expect(err).To(HaveOccurred())
```

This wastes ~1s per site because etcd's coarse TTL resolution means keys expire at `TTL + tick` regardless. The new pattern polls every 200ms and catches expiry immediately:

```go
Eventually(func() error {
    _, err := c.Resource().Get(ctx, name, options.GetOptions{})
    return err
}, 5*time.Second, 200*time.Millisecond).Should(HaveOccurred())
```

Also removes 7 unnecessary `time.Sleep(1s)` calls after delete operations where the subsequent check already has its own polling or the delete is synchronous. Applied across 18 clientv3 e2e test files.

### FV: Add context timeout to invalid-revision List calls (k8s backend, ~76s saved)

Two tests call `c.List()` with a fabricated too-high resource version to test error handling. With `ResourceVersionMatch=NotOlderThan`, the K8s API server blocks waiting for that version to appear, timing out after ~43 seconds each. Adding a 5-second context timeout makes the tests fail fast while still validating the error behavior.

### FV: Reduce overkill test parameters (~37s saved)

- "Many policies" pagination tests: 1000→550 resources (still triggers pagination at page size 500, but creates 45% fewer resources via the K8s API)
- Watch prefix deletion test: 5000→500 BGPPeers (tests the same bulk watch event delivery, 10x faster)

### UT improvements

- `metrics_server_test.go`: RSA key size 4096→2048, reduced startup wait and retry delays (**15s → 1.7s**)
- `health_test.go`: Reduced reporter timeout and sleep duration (**2.7s → 0.85s**)
- `watchersyncer_test.go`: Tightened watch intervals and poll timeouts, fixed a `Duration*Duration` bug (**16s → 8s**)

### Makefile

Adds `$(GINKGO_ARGS)` passthrough to the `fv-fast` target for consistency with `ut`.


libcalico-go/lib/backend/k8s/client_test.go
```diff
@@ -1112,7 +1112,11 @@ var _ = testutils.E2eDatastoreDescribe("Test Syncer API for Kubernetes backend",
 		})
 
 		By("Listing all Tiers, using an invalid revision", func() {
-			_, err := c.List(ctx, model.ResourceListOptions{Kind: apiv3.KindTier}, fmt.Sprintf("1%s", kvp2b.Revision))
+			// Use a short timeout context - the API server will wait for the invalid (too-high)
+			// resource version to appear, which would take ~43s with the default timeout.
+			shortCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+			defer cancel()
+			_, err := c.List(shortCtx, model.ResourceListOptions{Kind: apiv3.KindTier}, fmt.Sprintf("1%s", kvp2b.Revision))
```
nelljerram: I don't follow why `kvp2b.Revision` is considered to be an invalid version here.  IIUC it comes from `kvpRes.Revision`, which was the version for a Tier creation, and so valid in the datastore overall.

nelljerram: (And if it is actually valid, I think that questions the logic of this change?)

libcalico-go/lib/backend/watchersyncer/watchersyncer_test.go
```diff
@@ -261,7 +261,7 @@ var _ = Describe("Test the backend datastore multi-watch syncer", func() {
 		// - list succeeds
 		// - watch succeeds ... total 6s
 		By("Driving a bunch of List complete, Watch fail events for the 3rd resource type")
-		expectedDuration = watchersyncer.WatchPollInterval * watchersyncer.ListRetryInterval
+		expectedDuration = watchersyncer.WatchPollInterval + watchersyncer.ListRetryInterval
```
nelljerram, line 264: I don't understand how this test could have worked before, so I worry that I'm somehow misunderstanding, and the fix might not be right.  It seems that:
- `expectedDuration` and `minDuration` would have been huge before
- either the test would have taken hours to run, or the first `rs.allEventsHandled()` check would be true and then the `Expect(duration).To(BeNumerically(">", minDuration))` check would fail.

What am I missing?

libcalico-go/lib/clientv3/globalnetworkpolicy_e2e_test.go
```diff
@@ -282,47 +282,40 @@ var _ = testutils.E2eDatastoreDescribe("GlobalNetworkPolicy tests", testutils.Da
...
 			}
 
 			By("Attempting to delete GlobalNetworkPolicy (name2) again")
-			_, outError = c.GlobalNetworkPolicies().Delete(ctx, name2, options.DeleteOptions{})
-			Expect(outError).To(HaveOccurred())
-			Expect(outError.Error()).To(ContainSubstring("resource does not exist: GlobalNetworkPolicy(" + name2 + ") with error:"))
+			Eventually(func() error {
+				_, err := c.GlobalNetworkPolicies().Delete(ctx, name2, options.DeleteOptions{})
+				return err
+			}, 5*time.Second, 200*time.Millisecond).Should(HaveOccurred())
```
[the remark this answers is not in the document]
nelljerram: I think this test code was better as it was.  I believe the Delete here should reliably fail the first time, and should need any retrying.

libcalico-go/lib/clientv3/stagedglobalnetworkpolicy_e2e_test.go
```diff
@@ -276,47 +276,40 @@ var _ = testutils.E2eDatastoreDescribe("StagedGlobalNetworkPolicy tests", testut
...
 			}
 
 			By("Attempting to delete StagedGlobalNetworkPolicy (name2) again")
-			_, outError = c.StagedGlobalNetworkPolicies().Delete(ctx, name2, options.DeleteOptions{})
-			Expect(outError).To(HaveOccurred())
-			Expect(outError.Error()).To(ContainSubstring("resource does not exist: StagedGlobalNetworkPolicy(" + name2 + ") with error:"))
+			Eventually(func() error {
+				_, err := c.StagedGlobalNetworkPolicies().Delete(ctx, name2, options.DeleteOptions{})
+				return err
+			}, 5*time.Second, 200*time.Millisecond).Should(HaveOccurred())
```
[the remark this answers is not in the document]
nelljerram: Code should be as it was; retry should not be needed.

libcalico-go/lib/metricsserver/metrics_server_test.go
```diff
@@ -147,7 +147,7 @@ func TestServePrometheusMetricsHTTPS(t *testing.T) {
 			select {
 			case err := <-done:
 				recordConnectionError(err)
-			case <-time.After(1 * time.Second):
+			case <-time.After(200 * time.Millisecond):
```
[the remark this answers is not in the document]
nelljerram: I think the question is: is 200ms (or alternatively 1s) long enough for the server startup to generate an error, if there is a reason why it _should_ generate an error?  It does feel safer to stay at 1s in this case.

libcalico-go/lib/clientv3/stagedglobalnetworkpolicy_e2e_test.go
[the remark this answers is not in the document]
caseydavenport: Yeah, agreed

libcalico-go/lib/backend/k8s/client_test.go
caseydavenport: The revision being used here is actually `fmt.Sprintf("1%s", kvp2b.Revision)` which is invalid because it will be too high of a value compared to the latest revision in the apiserver.

But I am actually not sure about the reasoning here regardless - I will double check this one.

libcalico-go/lib/backend/watchersyncer/watchersyncer_test.go
caseydavenport: yeah, this one is baffling me a little bit too.... doing some digging.

caseydavenport: @nelljerram OK this is what Claude was able to tell me:

> The code was indeed buggy (Duration * Duration produces nonsensical nanosecond-squared values), but it "worked" because integer overflow made minDuration wrap to a negative number, so the > minDuration check was trivially true. And maxDuration overflowed to a still-huge positive number (~2424 hours), so the < maxDuration check also trivially passed. The timing assertions were effectively dead code — they weren't actually validating anything. The fix to use +  
  instead of * is correct.

libcalico-go/lib/clientv3/bgpfilter_e2e_test.go
```diff
@@ -264,27 +264,21 @@ var _ = testutils.E2eDatastoreDescribe("BGPFilter tests", testutils.DatastoreAll
...
-						_, outError = c.BGPFilter().Get(ctx, name2, options.GetOptions{})
-						Expect(outError).NotTo(HaveOccurred())
-						time.Sleep(2 * time.Second)
-						_, outError = c.BGPFilter().Get(ctx, name2, options.GetOptions{})
-						Expect(outError).To(HaveOccurred())
-						Expect(outError.Error()).To(ContainSubstring("resource does not exist: BGPFilter(" + name2 + ") with error:"))
+						Eventually(func() error {
+							_, err := c.BGPFilter().Get(ctx, name2, options.GetOptions{})
+							return err
+						}, 5*time.Second, 200*time.Millisecond).Should(HaveOccurred())
```
caseydavenport: We should continue to assert on the error string output here and elsewhere in this PR.

libcalico-go/lib/backend/k8s/client_test.go
nelljerram: Ah thanks, that answers my concern, I missed that `1`.  Happy to leave further review of the reasoning to you.

libcalico-go/lib/backend/watchersyncer/watchersyncer_test.go
nelljerram: Thanks.  Good to have an understanding of this.

PR #12035, opened 2026-03-08
Speed up kube-controllers unit tests by ~20%

caseydavenport:
Three categories of improvements to kube-controllers test runtime, measured on the same machine running the full suite via `make -C kube-controllers ut`:

**Convert FV tests to Ordered containers** — Most FV test suites were using `BeforeEach`/`AfterEach` to spin up and tear down etcd, apiserver, and kube-controller-manager Docker containers for every single spec. This is the single largest source of overhead in the suite. Converting to Ginkgo's `Ordered` container pattern with `BeforeAll`/`AfterAll` starts the infrastructure once per suite and reuses it across specs. Per-spec `AfterEach` blocks handle lightweight resource cleanup (deleting CRDs, pods, nodes) so each spec still starts from a clean state.

**Fix kdd_ipam_gc AfterEach pod/node deletion order** — The AfterEach in the KDD IPAM GC tests was deleting the node before the pods. This orphaned the pods, and the kube-controller-manager's pod GC only runs every ~20s, so cleanup took ~50s per spec. Reordering to delete pods first (with `GracePeriodSeconds=0`) then the node avoids the orphan GC entirely. This saves 30-50s on each of the 4 affected specs.

**Use direct indexer population for scale test** — The "should detect a leaked IP reasonably quickly" spec creates 1000 nodes and 5000 pods through the fake K8s client, waiting for each object to propagate through the informer's create→watch→inform pipeline. This took ~60s of pure setup. Populating the informer caches directly via `cache.Indexer.Add()` achieves the same result in under 1s.

**Convert etcd-mode FV tests to in-process controllers** — The serviceaccount, networkpolicy, and etcd IPAM GC FV tests previously ran the kube-controllers binary in a Docker container alongside a real K8s apiserver and controller-manager. These tests now instantiate the controller directly in-process using `fake.NewSimpleClientset()` for the K8s client, eliminating 3 Docker containers per test suite. This required changing the controller constructors to accept `kubernetes.Interface` instead of `*kubernetes.Clientset`, which also makes these controllers more testable going forward.

### Results (master vs branch, same machine)

| Package | Master | Branch | Delta |
|---------|--------|--------|-------|
| node | 602.6s | 423.4s | -179.2s (-30%) |
| config | 50.5s | 26.4s | -24.1s (-48%) |
| networkpolicy | 122.8s | 110.8s | -12.0s (-10%) |
| pod | 29.8s | 18.3s | -11.5s (-39%) |
| loadbalancer | 84.0s | 78.2s | -5.8s (-7%) |
| serviceaccount | 29.3s | 25.1s | -4.2s (-14%) |
| namespace | 28.4s | 29.7s | +1.3s (noise) |
| flannelmigration | 88.0s | 116.0s | +28.0s (noise) |
| **Total** | **1035.4s** | **827.9s** | **-207.5s (-20%)** |

The flannelmigration variance is from running both suites concurrently on the same machine — it has long container startups sensitive to system load. All 146 specs pass on the branch.

### DIFF
--- a/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller_fv_test.go
+++ b/kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller_fv_test.go
@@ -454,16 +454,19 @@ var _ = Describe("Calico loadbalancer controller FV tests (etcd mode)", Ordered,
 				return service.Status.LoadBalancer.Ingress
 			}, time.Second*15, 500*time.Millisecond).Should(Not(BeEmpty()))
 
-			service, err = k8sClient.CoreV1().Services(testNamespace).Get(context.Background(), serviceIpv4PoolSpecified.Name, metav1.GetOptions{})
-			Expect(err).NotTo(HaveOccurred())
-
-			// Update the service to have the same IP as the service we have created above
-			service.Annotations = map[string]string{
-				"projectcalico.org/loadBalancerIPs": fmt.Sprintf("[\"%s\"]", specificIpFromAutomaticPool),
-			}
-
-			_, err = k8sClient.CoreV1().Services(testNamespace).Update(context.Background(), service, metav1.UpdateOptions{})
-			Expect(err).NotTo(HaveOccurred())
+			// Update the service to have the same IP as the service we have created above.
+			// Retry on conflict since the loadbalancer controller may concurrently update the service.
+			Eventually(func() error {
+				svc, err := k8sClient.CoreV1().Services(testNamespace).Get(context.Background(), serviceIpv4PoolSpecified.Name, metav1.GetOptions{})
+				if err != nil {
+					return err
+				}
+				svc.Annotations = map[string]string{
+					"projectcalico.org/loadBalancerIPs": fmt.Sprintf("[\"%s\"]", specificIpFromAutomaticPool),
+				}
+				_, err = k8sClient.CoreV1().Services(testNamespace).Update(context.Background(), svc, metav1.UpdateOptions{})
+				return err
+			}, time.Second*5, 500*time.Millisecond).Should(Succeed())
 
 			// The service ingress should be empty
 			Eventually(func() []v1.LoadBalancerIngress {
--- a/kube-controllers/tests/testutils/flannel_migration_utils.go
+++ b/kube-controllers/tests/testutils/flannel_migration_utils.go
@@ -145,15 +145,17 @@ func (f *FlannelCluster) AddFlannelNode(nodeName, podCidr, backend, mac, ip stri
 }
 
 func (f *FlannelCluster) AddDefaultCalicoConfigMap() {
-	_, err := f.k8sClient.CoreV1().ConfigMaps(metav1.NamespaceSystem).Create(context.Background(),
-		&v1.ConfigMap{
-			TypeMeta: metav1.TypeMeta{Kind: "ConfigMap", APIVersion: "v1"},
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "calico-config",
-			},
-			Data: map[string]string{"veth_mtu": "1450"},
+	cm := &v1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{Kind: "ConfigMap", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "calico-config",
 		},
-		metav1.CreateOptions{})
+		Data: map[string]string{"veth_mtu": "1450"},
+	}
+	_, err := f.k8sClient.CoreV1().ConfigMaps(metav1.NamespaceSystem).Create(context.Background(), cm, metav1.CreateOptions{})
+	if apierrs.IsAlreadyExists(err) {
+		_, err = f.k8sClient.CoreV1().ConfigMaps(metav1.NamespaceSystem).Update(context.Background(), cm, metav1.UpdateOptions{})
+	}
 	Expect(err).NotTo(HaveOccurred())
 }
 


kube-controllers/pkg/controllers/networkpolicy/policy_controller_fv_test.go
```diff
@@ -21,95 +21,110 @@ import (
...
+	"github.com/projectcalico/calico/kube-controllers/pkg/config"
+	"github.com/projectcalico/calico/kube-controllers/pkg/controllers/networkpolicy"
 	"github.com/projectcalico/calico/kube-controllers/tests/testutils"
 	"github.com/projectcalico/calico/libcalico-go/lib/apiconfig"
 	client "github.com/projectcalico/calico/libcalico-go/lib/clientv3"
 	"github.com/projectcalico/calico/libcalico-go/lib/options"
 )
 
-var _ = Describe("Calico networkpolicy controller FV tests (etcd mode)", func() {
+var _ = Describe("Calico networkpolicy controller FV tests (etcd mode)", Ordered, func() {
```
coutinhop: when using `Ordered`, a failure will stop the tests then and there (https://onsi.github.io/ginkgo/#failure-handling-in-ordered-containers), is this something we'd want? Otherwise we can add `ContinueOnFailure` so that all other tests do run even when one fails...

caseydavenport: Mmm, good catch. We probably want ContinueOnFailure here. Let me look.

kube-controllers/tests/testutils/cleanup.go
```diff
@@ -0,0 +1,379 @@
...
+// is running in the test.
+//
+// FelixConfigurations and BGPConfigurations named "default" are preserved when
+// keepDefaultConfigs is true.
+func CleanupAllResources(
+	ctx context.Context,
+	k8sClient kubernetes.Interface,
+	calicoClient client.Interface,
+	bc api.Client,
+	deletePodsBeforeNodes bool,
```
MichalFupso: Might be worth it to create a cleanupOptions 

```
type cleanupOptions struct {
	deletePodsBeforeNodes bool
	keepDefaultConfigs    bool
}
```

kube-controllers/pkg/controllers/namespace/namespace_controller_fv_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017,2021 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017,2021,2026 Tigera, Inc. All rights reserved.
```
MichalFupso:
```suggestion
// Copyright (c) 2017-2026 Tigera, Inc. All rights reserved.
```

kube-controllers/pkg/controllers/networkpolicy/policy_controller_fv_test.go
```diff
@@ -21,95 +21,110 @@ import (
...
-		controllerManager.Stop()
-		kubeControllers.Stop()
-		apiserver.Stop()
 		etcd.Stop()
 	})
 
+	AfterEach(func() {
+		close(stopCh)
+
+		// Clean up Calico network policies.
```
MichalFupso: This functionality is in the new cleanup.go, does it make sense to make it public and call that function instead

kube-controllers/pkg/controllers/namespace/namespace_controller_fv_test.go
```diff
@@ -107,6 +109,15 @@ var _ = Describe("Calico namespace controller FV tests (etcd mode)", func() {
 			}, time.Second*15, 500*time.Millisecond).ShouldNot(BeNil())
 		})
 
+		AfterEach(func() {
+			ctx := context.Background()
```
MichalFupso: I think similar functionality is in cleanup.go, cleanupK8sNamespaces could take in a list of namespaces to not delete and it could be used here

kube-controllers/pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -50,11 +51,18 @@ type policyController struct {
...
 
 // NewPolicyController returns a controller which manages NetworkPolicy objects.
-func NewPolicyController(ctx context.Context, clientset *kubernetes.Clientset, c client.Interface, cfg config.GenericControllerConfig) controller.Controller {
+func NewPolicyController(ctx context.Context, clientset kubernetes.Interface, c client.Interface, cfg config.GenericControllerConfig) controller.Controller {
 	policyConverter := converter.NewPolicyConverter()
 
 	// Create a NetworkPolicy watcher.
-	listWatcher := cache.NewListWatchFromClient(clientset.NetworkingV1().RESTClient(), "networkpolicies", "", fields.Everything())
+	listWatcher := &cache.ListWatch{
+		ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
```
MichalFupso: Deprecated
```suggestion
		ListWithContextFunc: func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
```

kube-controllers/pkg/controllers/networkpolicy/policy_controller.go
```diff
@@ -50,11 +51,18 @@ type policyController struct {
...
+			return clientset.NetworkingV1().NetworkPolicies("").List(ctx, opts)
+		},
+		WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
```
MichalFupso: Deprecated
```suggestion
		WatchFuncWithContext: func(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error) {
```

PR #12038, opened 2026-03-09
FelixConfiguration and Felix changes for live migration

nelljerram:
## Summary

- Add live migration support to Felix, enabling seamless VM migration between hosts with minimal
network disruption
- Introduce a LiveMigration API resource and calc graph calculator that tracks migration
source/target roles for workload endpoints
- Implement a per-workload FSM (Base → Target → Live → TimeWait → Base) in the dataplane that
adjusts route priorities during migration and detects migration completion via GARP/RARP
- Suppress proxy ARP for workload's own IPs using a new nftables ARP family table, preventing the
host from answering a VM's ARP probes for its own address

### Calc graph: LiveMigrationCalculator
- New `LiveMigrationCalculator` processes LiveMigration resources and annotates WorkloadEndpoint
proto messages with `live_migration_role` (source/target)
- Supports both direct name references and label-selector-based target matching
- Buffers early selector matches that arrive before WEP tracking (dispatcher ordering fix)
- Calc graph FV tests covering source, target-by-name, and target-by-selector scenarios

### Dataplane: live migration FSM and route priority
- `liveMigrationMonitor` manages per-workload FSMs that track migration state
- **Target** state: routes suppressed (workload hasn't migrated yet)
- **Live** state: elevated route priority after GARP/RARP detection confirms VM is active
- **TimeWait** state: elevated priority maintained for configurable convergence period
(`LiveMigrationRouteConvergenceTime`, default 30s) before reverting to normal
- GARP/RARP detection via AF_PACKET (pcapgo) on workload interfaces

### Dataplane: ARP suppression
- New nftables table in the `arp` family (`calico-arp`) with per-workload chains
- Drops ARP replies from the host containing the workload's own IP, preventing proxy ARP confusion
- Always active regardless of dataplane mode (iptables/nftables/eBPF)
- Dispatch via verdict map matching on output interface name

### Configuration
- `*RoutePriority` fields (int) — priority metric values for normal and elevated priority routing
- `LiveMigrationRouteConvergenceTime` (duration) — time to hold elevated route priority after
migration completes

## Test plan

- [x] Unit tests for `LiveMigrationCalculator` (bug fixes, selector matching, lifecycle)
- [x] Calc graph FV tests for LiveMigration resource processing
- [x] Unit tests for `liveMigrationMonitor` FSM (24-case transition table, lifecycle, multi-step
scenarios)
- [x] Unit tests for endpoint manager route priority adjustment during migration
- [x] Unit tests for nftables ARP match builder (`ARPOperation`, `ARPSrcIP`)
- [x] Unit tests for `NewARPTable` (base chains, family, rule insertion)
- [x] Unit tests for endpoint manager ARP chain programming (add/remove/multi-IP/nil-table)
- [ ] CI green on all Felix UT/FV suites
- [ ] Manual verification of ARP suppression with live migration workflow

**Release note:**
```release-note
Support live migration for KubeVirt VMs.
```

### DIFF
--- a/felix/calc/calc_graph.go
+++ b/felix/calc/calc_graph.go
@@ -229,10 +229,7 @@ func NewCalculationGraph(
 	// Create and register the live migration calculator on localEndpointDispatcher BEFORE
 	// the active rules calculator.  This ensures the LMC sees WEP updates first, so its
 	// wepData is already populated when the ARC fires computed selector match callbacks.
-	cg.liveMigrationCalculator = NewLiveMigrationCalculator(
-		cg.activeRulesCalculator,
-		nil, // onEndpointComputedData wired up later, after polResolver is created
-	)
+	cg.liveMigrationCalculator = NewLiveMigrationCalculator(cg.activeRulesCalculator)
 	localEndpointDispatcher.Register(
 		model.WorkloadEndpointKey{},
 		cg.liveMigrationCalculator.OnUpdate,
@@ -395,7 +392,7 @@ func NewCalculationGraph(
 	// Wire up the live migration calculator's output callback and remaining dispatcher
 	// registration, now that polResolver exists.  (The LMC was created and registered on
 	// localEndpointDispatcher earlier, before the ARC, to ensure correct dispatch ordering.)
-	cg.liveMigrationCalculator.onEndpointComputedData = polResolver.OnEndpointComputedDataUpdate
+	cg.liveMigrationCalculator.OnEndpointComputedData = polResolver.OnEndpointComputedDataUpdate
 	allUpdDispatcher.Register(
 		model.ResourceKey{},
 		cg.liveMigrationCalculator.OnUpdate,
--- a/felix/calc/live_migration_calculator.go
+++ b/felix/calc/live_migration_calculator.go
@@ -31,7 +31,7 @@ import (
 // proto.WorkloadEndpoints that are sent to the dataplane.
 type LiveMigrationCalculator struct {
 	activeRulesCalc        *ActiveRulesCalculator
-	onEndpointComputedData EndpointComputedDataUpdater
+	OnEndpointComputedData EndpointComputedDataUpdater
 	weps                   map[wepOwnerID]*wepData
 	liveMigrations         map[model.ResourceKey]internalapi.LiveMigration
 	// endpointKeys tracks LiveMigration resources that reference a specific endpoint
@@ -74,15 +74,11 @@ type wepData struct {
 	targetSelectorMatched string
 }
 
-func NewLiveMigrationCalculator(
-	activeRulesCalc *ActiveRulesCalculator,
-	onEndpointComputedDataUpdater EndpointComputedDataUpdater,
-) *LiveMigrationCalculator {
+func NewLiveMigrationCalculator(activeRulesCalc *ActiveRulesCalculator) *LiveMigrationCalculator {
 	lmc := &LiveMigrationCalculator{
-		activeRulesCalc:        activeRulesCalc,
-		onEndpointComputedData: onEndpointComputedDataUpdater,
-		weps:                   map[wepOwnerID]*wepData{},
-		liveMigrations:         map[model.ResourceKey]internalapi.LiveMigration{},
+		activeRulesCalc: activeRulesCalc,
+		weps:            map[wepOwnerID]*wepData{},
+		liveMigrations:  map[model.ResourceKey]internalapi.LiveMigration{},
 		endpointKeys: [numSourceOrTarget]map[wepOwnerID]set.Set[model.ResourceKey]{
 			map[wepOwnerID]set.Set[model.ResourceKey]{},
 			map[wepOwnerID]set.Set[model.ResourceKey]{},
@@ -238,10 +234,11 @@ func (lmc *LiveMigrationCalculator) OnComputedSelectorMatch(cs string, epKey mod
 				wd.targetSelectorMatched = cs
 			})
 		}
-		// If the WEP isn't tracked yet, we can ignore this callback.  The LMC is
-		// registered on localEndpointDispatcher before the ARC, so it always sees WEP
-		// updates first.  A selector callback for an unknown WEP means the WEP was
-		// filtered out (e.g. non-local) and we don't need to track it.
+		// If the WEP isn't tracked yet, we can ignore this callback.  The LMC is registered
+		// on localEndpointDispatcher before the ARC, so it always sees WEP updates first.
+		// The unknown WEP case here (wd == nil) shouldn't occur because
+		// LiveMigrationCalculator sees the same set of local WEPs as ActiveRulesCalculator,
+		// but it's harmless to be defensive here.
 	}
 }
 
@@ -501,7 +498,7 @@ func (lmc *LiveMigrationCalculator) indicateRole(key model.WorkloadEndpointKey, 
 		"role": role,
 		"uid":  uid,
 	}).Debug("LiveMigrationCalculator: emitting role for WEP")
-	lmc.onEndpointComputedData(key, EPCompDataKindLiveMigration, &liveMigrationRole{role: role, uid: uid})
+	lmc.OnEndpointComputedData(key, EPCompDataKindLiveMigration, &liveMigrationRole{role: role, uid: uid})
 }
 
 type liveMigrationRole struct {
--- a/felix/calc/live_migration_calculator_test.go
+++ b/felix/calc/live_migration_calculator_test.go
@@ -47,13 +47,11 @@ func newTestLMCEnv() *testLMCEnv {
 	arc.RuleScanner = &noopRuleScanner{}
 	env.arc = arc
 
-	env.lmc = NewLiveMigrationCalculator(
-		arc,
-		func(key model.WorkloadEndpointKey, kind EndpointComputedDataKind, data EndpointComputedData) {
-			lmr := data.(*liveMigrationRole)
-			env.roleEvents = append(env.roleEvents, roleEvent{key: key, role: lmr.role, uid: lmr.uid})
-		},
-	)
+	env.lmc = NewLiveMigrationCalculator(arc)
+	env.lmc.OnEndpointComputedData = func(key model.WorkloadEndpointKey, kind EndpointComputedDataKind, data EndpointComputedData) {
+		lmr := data.(*liveMigrationRole)
+		env.roleEvents = append(env.roleEvents, roleEvent{key: key, role: lmr.role, uid: lmr.uid})
+	}
 	return env
 }
 


felix/nftables/table.go
```diff
@@ -406,6 +418,127 @@ func NewTable(
...
+// from workloads for their own IPs. The ARP family table has a single
+// filter-OUTPUT base chain.
+func NewARPTable(
+	name string,
+	hashPrefix string,
+	featureDetector environment.FeatureDetectorIface,
+	options TableOptions,
+	required bool,
+) *NftablesTable {
+	ourChainsRegexp := regexp.MustCompile("^(cali|filter)-.*")
```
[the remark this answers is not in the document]
fasaxc: I think this comes from porting over from iptables; the unprefixed chains are the top-level hook chains.

felix/nftables/table.go
```diff
@@ -406,6 +418,127 @@ func NewTable(
 	return table
 }
 
+// NewARPTable creates a new NftablesTable in the ARP family. This is used for
+// proxy ARP suppression, preventing the host from responding to ARP requests
+// from workloads for their own IPs. The ARP family table has a single
+// filter-OUTPUT base chain.
+func NewARPTable(
```
fasaxc, line 328: Think the meat of this constructor should be combined with the normal one. Looks like everything can be shared apart from needing to select the right base chains?

felix/dataplane/linux/live_migration.go
```diff
@@ -0,0 +1,349 @@
...
+func (m *liveMigrationMonitor) OnTimerPop(id types.WorkloadEndpointID) {
+	m.executeFSM(id, liveMigrationInputTimerPop)
+}
+
+// detectGARP reads packets from the given handle and sends the workload ID
+// to garpC when a GARP or RARP packet is detected.  It is a one-shot
+// goroutine: it returns after the first detection or when the handle is closed.
+func detectGARP(logCtx *logrus.Entry, id types.WorkloadEndpointID,
+	handle garpHandle, garpC chan<- types.WorkloadEndpointID) {
+	defer handle.Close()
```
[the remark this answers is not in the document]
nelljerram: You're right, it's not safe; I thought I had looked at this before and concluded it was OK, but it's clearer anyway only to close the handle in one place in stopGARPDetection().  Will fix this.

felix/dataplane/linux/live_migration.go
```diff
@@ -0,0 +1,349 @@
...
+	})
+}
+
+func (f *liveMigrationFSM) stopElevatedRoutingTimer() {
+	if f.timer != nil {
+		f.logCtx.Debug("Stopping elevated routing timer")
+		f.timer.Stop()
+		f.timer = nil
+	}
+}
```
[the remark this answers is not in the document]
nelljerram: Agreed.  Will fix in OnTimerPop by taking care not to recreate the FSM.

libcalico-go/lib/backend/model/workloadendpoint.go
```diff
@@ -95,15 +95,20 @@ func (key WorkloadEndpointKey) String() string {
...
-func (key WorkloadEndpointKey) GetNamespace() string {
+// WorkloadID is expected to be in the format "name" or "namespace/name".
+func (key WorkloadEndpointKey) GetNameAndNamespace() (string, string) {
 	parts := strings.SplitN(key.WorkloadID, "/", 2)
 	if len(parts) == 2 {
-		return parts[0]
+		return parts[1], parts[0]
 	}
-	return ""
+	return parts[0], ""
```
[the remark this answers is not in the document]
nelljerram: OK, will beef up.

felix/dataplane/linux/live_migration_test.go
```diff
@@ -0,0 +1,692 @@
...
+
+			g.Expect(fsm.currentState).To(Equal(tt.wantState), "FSM state")
+			if tt.wantState != tt.from {
+				g.Expect(m.pendingUpdates).To(HaveLen(1), "should emit one update")
+				g.Expect(m.pendingUpdates[0].State).To(Equal(tt.wantState), "emitted state")
+			} else {
+				g.Expect(m.pendingUpdates).To(BeEmpty(), "should not emit any updates")
+			}
+		})
+	}
```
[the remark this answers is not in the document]
nelljerram: Agreed, will fix.

felix/calc/states_for_test.go
```diff
@@ -3272,6 +3273,81 @@ var wireguardV4V6 = empty.withKVUpdates(
...
+
+	// A remote WEP name that doesn't match any local WEP.
+	remoteWEPName = k8stypes.NamespacedName{Name: "remote-wep"}
+
+	// NamespacedNames matching localWlEpKey1 (WorkloadID "wl1" → name="wl1", namespace="").
+	localWlEp1NN = k8stypes.NamespacedName{Name: "wl1"}
+
+	// A selector that matches localWlEp1's labels (has label "a").
+	liveMigrationTargetSelector = "has(a)"
+)
```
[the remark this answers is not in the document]
nelljerram: Don't see any actual problem here, or anything that needs clarification.

felix/dataplane/linux/endpoint_mgr.go
```diff
@@ -881,7 +918,38 @@ func (m *endpointManager) resolveWorkloadEndpoints() {
...
+
+		// Build the ARP dispatch chain with a vmap lookup.
+		arpDispatchChain := &generictables.Chain{
+			Name: rules.ChainARPDispatch,
+			Rules: []generictables.Rule{
+				{
+					Match: nftables.Match().OutInterfaceVMAP(rules.NftablesARPDispatchMap),
+				},
+			},
+		}
```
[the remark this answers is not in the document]
nelljerram: Believe this one is OK as is.  Compare the existing VMAP uses in dispatch.go:617 and dispatch.go:621 uses the same pattern: Match only, no Action.

felix/dataplane/linux/live_migration.go
```diff
@@ -0,0 +1,349 @@
...
+				m.executeFSM(id, liveMigrationInputTarget)
+			}
+		}
+	case *proto.WorkloadEndpointRemove:
+		id := types.ProtoToWorkloadEndpointID(msg.GetId())
+		delete(m.roles, id)
+		delete(m.ifaceNames, id)
+		m.executeFSM(id, liveMigrationInputDeleted)
+	}
+}
```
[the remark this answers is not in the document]
nelljerram: I think Copilot here is confusing the typical pattern in the dataplane (which is for all managers to see everything) with the typical pattern in the calc graph (which is for calculators to register for the types they want to process).

felix/calc/live_migration_calculator.go
```diff
@@ -0,0 +1,351 @@
...
+		}
+	}
+}
+
+type LiveMigrationRole struct {
+	role proto.LiveMigrationRole
+}
+
+func (lmr *LiveMigrationRole) ApplyTo(wep *proto.WorkloadEndpoint) {
+	wep.LiveMigrationRole = lmr.role
```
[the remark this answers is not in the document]
nelljerram: Agreed, will fix.

felix/nftables/table.go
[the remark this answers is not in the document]
nelljerram: Will add comment.  Also spotted a nearby typo.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -2422,6 +2482,10 @@ func (d *InternalDataplane) processMsgFromCalcGraph(msg any) {
 	for _, mgr := range d.allManagers {
 		mgr.OnUpdate(msg)
 	}
+	// The live migration monitor may have accumulated FSM state changes from the
+	// message we just fanned out.  Drain them and fan out to all managers as
+	// pseudo-proto messages.
+	d.dispatchLiveMigrationUpdates()
```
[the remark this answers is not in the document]
nelljerram: Hm, on balance, Claude and I think it's OK for LiveMigrationMonitor to remain as a manager.  Claude's reasoning, which I agree with:
> I agree it's fine. The alternative — making it not a manager and special-casing its WEP message delivery — would actually be more confusing and break the pattern that all WEP consumers are managers. Silently ignoring unhandled message types is exactly what every other manager does. The fact that its outputs are fanned out separately is a pragmatic choice: the monitor needs to process WEP updates first, then its derived state updates need to reach the endpoint manager in the same apply cycle. That ordering constraint is the real reason for the two-phase dispatch, not a design inconsistency.

felix/rules/rule_defs.go
```diff
@@ -91,6 +91,10 @@ const (
 	NftablesToWorkloadDispatchMap   = ChainNamePrefix + "to-wl-dispatch"
 	NftablesFromWorkloadDispatchMap = ChainNamePrefix + "from-wl-dispatch"
 
+	WorkloadARPPfx         = ChainNamePrefix + "arp-"
+	NftablesARPDispatchMap = ChainNamePrefix + "arp-dispatch"
+	ChainARPDispatch       = ChainNamePrefix + "arp-dispatch"
```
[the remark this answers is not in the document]
nelljerram: Arguable, but we have exactly the same in existing code here:
```
	ChainToWorkloadDispatch   = ChainNamePrefix + "to-wl-dispatch"

	NftablesToWorkloadDispatchMap   = ChainNamePrefix + "to-wl-dispatch"
```
I'd rather follow that existing convention.

libcalico-go/lib/backend/model/workloadendpoint.go
```diff
@@ -95,15 +95,23 @@ func (key WorkloadEndpointKey) String() string {
...
 }
 
-// GetNamespace extracts and returns the namespace from the WorkloadID.
-// WorkloadID is expected to be in the format "namespace/name".
-// Returns an empty string if the WorkloadID doesn't contain a namespace.
-func (key WorkloadEndpointKey) GetNamespace() string {
+// WorkloadID is expected to be in the format "name" or "namespace/name".
+// Returns name first, then namespace, with namespace being empty if WorkloadID
+// doesn't have any namespace.
+func (key WorkloadEndpointKey) GetNameAndNamespace() (string, string) {
```
fasaxc: Think I'd return the ns first in line with the input

felix/calc/live_migration_calculator.go
```diff
@@ -0,0 +1,351 @@
...
+	numSourceOrTarget
+)
+
+// LiveMigrationCalculator tracks local workload endpoints and LiveMigration resources and
+// correlates between them, in order to correctly set the live_migration_role field in
+// proto.WorkloadEndpoints that are sent to the dataplane.
+type LiveMigrationCalculator struct {
+	activeRulesCalc        *ActiveRulesCalculator
+	onEndpointComputedData EndpointComputedDataUpdater
+	weps                   map[types.NamespacedName]*wepData
```
fasaxc: As discussed, we have a mismatch between WEP and Pod.  One pod may have multiple weps in a multi-NIC scenario.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1097,29 +1110,53 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		filterMaps = filterTableV4.(nftables.MapsDataplane)
 	}
 
+	// Create nftables ARP table for proxy ARP suppression (always enabled, regardless of dataplane mode).
```
fasaxc, line 1113: Might be good to have an enabled-by-default feature flag to disable this in case someone is running on an old/exotic distro and suddenly can't upgrade calico due to nftables being disabled.

I'm imagining a feature flag "NftablesSupported" that could be auto-detected, but actually we just hard-code its default to enabled so that we have a workaround if its ever needed.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1307,25 +1344,31 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
 			"",
 			filterMapsV6,
-			config.BPFEnabled,
-			config.BPFAttachType,
 			nil,
 			callbacks,
-			config.FloatingIPsEnabled,
-			nftablesEnabled,
 			linkAddrsManagerV6,
+			nil, // arpTable - ARP is IPv4 only
```
fasaxc, line 1371: And we don't need an NDP equivalent?  (I can't remember what we did for NDP in IPv6, did we abandon proxying?)

felix/dataplane/linux/int_dataplane.go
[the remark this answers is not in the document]
fasaxc: Any reason that LiveMigrationController can't use the existing UpdateBatchResolver interface to fan out its updates?  I think we added it for exactly this case.
```
type UpdateBatchResolver interface {
	// Opportunity for a manager component to resolve state that depends jointly on the updates
	// that it has seen since the preceding CompleteDeferredWork call.  Processing here can
	// include passing resolved state to other managers.  It should not include any actual
	// dataplane updates yet.  (Those should be actioned in CompleteDeferredWork.)
	ResolveUpdateBatch() error
}
```
When we've needed cross-manager communication in the past we have not tried to force the updates through `OnUpdates()`, we just define interfaces between the managers.  c.f. `hepListener` in `endpoint_mgr.go`.

networking-calico/networking_calico/datamodel_v3.py
```diff
@@ -116,7 +118,10 @@ def put(
         value["metadata"]["labels"] = labels
     # Set the new spec (overriding whatever may already be there).
     value["spec"] = spec
-    return etcdv3.put(key, json.dumps(value), mod_revision=mod_revision)
+    uid = value["metadata"]["uid"]
```
fasaxc: Is this change meant to be in this PR?  I don't see anywhere that's using the UID in anger.

felix/dataplane/linux/endpoint_mgr.go
```diff
@@ -447,6 +456,20 @@ func (m *endpointManager) OnUpdate(protoBufMsg any) {
...
+			"id":    msg.ID,
+			"state": msg.State,
+		}).Debug("Live migration state update")
+		if msg.State == liveMigrationStateBase {
+			delete(m.pendingLiveMigrationStates, msg.ID)
+		} else {
+			m.pendingLiveMigrationStates[msg.ID] = msg.State
+		}
+		// Mark the endpoint dirty so CompleteDeferredWork re-evaluates its routes.
+		if ep := m.activeWlEndpoints[msg.ID]; ep != nil {
```
fasaxc: Do you need to check that there wasn't already a pending one (caused by a real WEP update)?

felix/dataplane/linux/endpoint_mgr.go
fasaxc: Nice time to introduce a utility method?
```
routeTargets := m.calculateRoutes(id)
```

felix/calc/live_migration_calculator.go
```diff
@@ -0,0 +1,522 @@
...
+			}
+
+			// Check for direct-name matches: both exact endpoint-level and workload-level.
+			workloadID := exactID.workloadLevel()
+			for _, sot := range []sourceOrTarget{source, target} {
+				if lmc.directNameKeys[sot][exactID] != nil {
+					wd.directNameKeys[sot].AddSet(lmc.directNameKeys[sot][exactID])
+				}
+				if workloadID != exactID {
+					if lmc.directNameKeys[sot][workloadID] != nil {
```
fasaxc: Seems a bit convoluted to store both types of match criteria in one map and then to have to have to deal with this aliasing case; could we just use two maps?

felix/calc/live_migration_calculator.go
```diff
@@ -0,0 +1,522 @@
...
+			logrus.WithFields(logrus.Fields{
+				"wep":      exactID,
+				"selector": cs,
+			}).Debug("LiveMigrationCalculator: selector matched WEP")
+			lmc.withRoleUpdateIfNeeded(wd, func() {
+				wd.targetSelectorMatched = cs
+			})
+		} else {
+			// WEP not yet tracked; the ARC handler runs before ours in the
+			// dispatcher chain, so we may receive selector match callbacks
```
fasaxc: Could we just move it earlier in the chain to avoid?  

Maybe the extra complexity keeps us decoupled so it's not clear cut (but usually we've handled this kind of thing by choosing an order that works).

felix/calc/live_migration_calculator.go
```diff
@@ -0,0 +1,522 @@
...
+// referenced by the given set of keys, or "" if none has a UID.
+func (lmc *LiveMigrationCalculator) uidFromLMKeys(keys set.Set[model.ResourceKey]) string {
+	if keys == nil {
+		return ""
+	}
+	for lmKey := range keys.All() {
+		if lm, ok := lmc.liveMigrations[lmKey]; ok && string(lm.UID) != "" {
+			return string(lm.UID)
+		}
+	}
```
fasaxc, line 342: Let's make the return value deterministic.  I'd iterate the whole set and take the "greatest" UID

felix/dataplane/linux/live_migration.go
```diff
@@ -0,0 +1,408 @@
...
+		switch input {
+		case liveMigrationInputTarget:
+			next = liveMigrationStateTarget
+			f.startGARPDetection()
+		}
+	case liveMigrationStateTarget:
+		switch input {
+		case liveMigrationInputGARPDetected:
+			next = liveMigrationStateLive
+			f.stopGARPDetection()
```
fasaxc: Is there a way to represent the FSM that we can avoid every transition needing to be paired with the action? For example
```
if current == liveMigrationStateTarget && next != liveMigrationStateTarget {
  stopGARPDetection()
}
```
or
```
if entering(liveMigrationStateTarget) {
  startGARPDetection()
} else if leaving(liveMigrationStateTarget) {
  stopGARPDetection()
}
```
Just conscious of the ease of forgetting one of those stop() calls.

felix/dataplane/linux/live_migration.go
```diff
@@ -0,0 +1,408 @@
...
+}
+
+func (f *liveMigrationFSM) stopGARPDetection() {
+	if f.pcapHandle != nil {
+		if err := f.pcapHandle.Close(); err != nil {
+			f.logCtx.WithError(err).Debug("Error closing GARP detection handle")
+		}
+		f.pcapHandle = nil
+	}
+}
```
fasaxc, line 337: Should we have a waitgroup to wait for bg goroutine to exit (assuming it should stop almost immediately)?  It's good to avoid getting a stray kick from an old goroutine.

felix/dataplane/linux/live_migration.go
```diff
@@ -0,0 +1,408 @@
...
+
+func (f *liveMigrationFSM) startElevatedRoutingTimer() {
+	f.logCtx.WithField("duration", f.monitor.convergenceTime).Debug("Starting elevated routing timer")
+	id := f.id
+	f.timer = time.AfterFunc(f.monitor.convergenceTime, func() {
+		select {
+		case f.monitor.timerC <- id:
+		default:
+			logrus.WithField("id", id).Warn("Live migration timer channel full, dropping timer pop")
```
fasaxc: Handling these on the receive side should be pretty trivial; think I'd have just done a blocking send and made sure that I drain the channel (with `drainChan()`) on the receive side to deal with all in progress sends.

libcalico-go/lib/backend/model/workloadendpoint.go
nelljerram: We've ended up not needing the reason why I changed this area, so I'll revert here and felix/collector/types/endpoint/endpoint.go

felix/calc/live_migration_calculator.go
nelljerram: Addressed now.

felix/dataplane/linux/int_dataplane.go
nelljerram: It does already degrade gracefully in this scenario.  Tested at https://github.com/projectcalico/calico/pull/12038/changes#diff-04d98980b3f0580fc047f33d14cad5095b6b77a6d3485eb06330d9102ef0551aR107.  `NewARPTable` returns nil if nft is not available, and int_dataplane.go allows for that.

felix/dataplane/linux/int_dataplane.go
nelljerram: We probably do.  endpoint_mgr.go has:
```
		// Enable proxy NDP, similarly to proxy ARP, described above.
		err := writeProcSys(fmt.Sprintf("/proc/sys/net/ipv6/conf/%s/proxy_ndp", name), "1")
```
But I think it makes sense to split this to a separate ticket: https://tigera.atlassian.net/browse/CORE-12498

networking-calico/networking_calico/datamodel_v3.py
nelljerram: No sorry, I'll move this one to the OpenStack-specific PR.

felix/dataplane/linux/live_migration.go
nelljerram: Have done this, and also removed buffer capacity from the channels, as there is no need for it.

nelljerram: Thought about it, but unfortunately there is no other natural timer for when the send side could give up on trying to send.

felix/calc/live_migration_calculator.go
nelljerram: Thanks, nice cleanup here.

felix/dataplane/linux/int_dataplane.go
fasaxc: What if the binary is available but it errors due to missing kernel function?

felix/calc/calc_graph.go
```diff
@@ -224,9 +224,22 @@ func NewCalculationGraph(
...
 	activeRulesCalc := NewActiveRulesCalculator()
-	activeRulesCalc.RegisterWith(localEndpointDispatcher, allUpdDispatcher)
 	cg.activeRulesCalculator = activeRulesCalc
 
+	// Create and register the live migration calculator on localEndpointDispatcher BEFORE
+	// the active rules calculator.  This ensures the LMC sees WEP updates first, so its
+	// wepData is already populated when the ARC fires computed selector match callbacks.
+	cg.liveMigrationCalculator = NewLiveMigrationCalculator(
+		cg.activeRulesCalculator,
+		nil, // onEndpointComputedData wired up later, after polResolver is created
```
fasaxc: I'd remove from the constructor if the mainline is to wire it up later.

felix/calc/live_migration_calculator.go
```diff
@@ -26,7 +26,22 @@ import (
 	"github.com/projectcalico/calico/libcalico-go/lib/set"
 )
 
-const EPCompDataKindLiveMigration = EndpointComputedDataKind("LiveMigration")
+// LiveMigrationCalculator tracks local workload endpoints and LiveMigration resources and
+// correlates between them, in order to correctly set the live_migration_role field in
+// proto.WorkloadEndpoints that are sent to the dataplane.
+type LiveMigrationCalculator struct {
+	activeRulesCalc        *ActiveRulesCalculator
+	onEndpointComputedData EndpointComputedDataUpdater
```
fasaxc: I'd make it a public field since it's set externally.

felix/calc/live_migration_calculator.go
```diff
@@ -45,6 +60,209 @@ type wepOwnerID struct {
...
+				"wep":      exactID,
+				"selector": cs,
+			}).Debug("LiveMigrationCalculator: selector matched WEP")
+			lmc.withRoleUpdateIfNeeded(wd, func() {
+				wd.targetSelectorMatched = cs
+			})
+		}
+		// If the WEP isn't tracked yet, we can ignore this callback.  The LMC is
+		// registered on localEndpointDispatcher before the ARC, so it always sees WEP
+		// updates first.  A selector callback for an unknown WEP means the WEP was
```
fasaxc: I don't think the comment about non-local is correct.  ARC only sees local WEPs.

felix/calc/calc_graph.go
nelljerram: I wondered about that; will do.

PR #12051, opened 2026-03-10
Add host-identifying prefix to BPF log output in FV tests

sridhartigera:
## Summary
- Adds a short prefix (f0-, f1-, f2-) to the BPF `iface_name` log field for non-workload interfaces when running inside FV test containers
- Makes it easy to identify which Felix instance a BPF log line belongs to when debugging multi-node FV tests
- Zero production impact — prefix is only active when hostname matches the FV pattern (`felix-<id>-...-felixfv`)
- Workload interfaces (cali* veths) are excluded since they already have unique names

## Test plan
- [ ] Run BPF FV test: `make -C felix fv-bpf GINKGO_FOCUS="BPF dual stack basic in-cluster connectivity" GINKGO_ARGS="-ginkgo.v"`
- [ ] Verify BPF logs show `f0-eth0-I:` / `f1-eth0-E:` instead of `eth0-I:` / `eth0-E:`
- [ ] Verify workload interface logs remain unchanged (e.g. `cali1234abcde`)

### DIFF
--- a/felix/bpf/tc/attach.go
+++ b/felix/bpf/tc/attach.go
@@ -34,6 +34,7 @@ import (
 	"github.com/projectcalico/calico/felix/bpf/libbpf"
 	"github.com/projectcalico/calico/felix/bpf/maps"
 	tcdefs "github.com/projectcalico/calico/felix/bpf/tc/defs"
+	bpfutils "github.com/projectcalico/calico/felix/bpf/utils"
 	"github.com/projectcalico/calico/felix/dataplane/linux/qos"
 )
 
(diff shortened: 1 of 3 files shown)


felix/bpf/tc/attach.go
```diff
@@ -520,8 +521,14 @@ func (ap *AttachPoint) Configure() *libbpf.TcGlobalData {
 		globalData.JumpsV6[tcdefs.ProgIndexPolicy] = uint32(ap.PolicyIdxV6)
 	}
 
+	logIface := ap.Iface
```
tomastigera, line 524: I love the idea. But can't we rather rename the interfaces when we create the containers in fv infra than polluting prod code?

Issue #11910, opened 2026-02-24
calico update 3.30 -> 3.31 breaks hostnetwork vxlan on unupgraded nodes


juliantaylor:
While upgrading from calico 3.30.4 to 3.31.4 connection from hostnetwork pods via vxlan to kubernetes services break until all calico-node pods are fully updated.

The nodes running 3.31 respond seemingly correctly but the v3.30 drop the packets after vxlan decapsulations

tcpdump hostnetwork pod with calico 3.30.4 -> kubernetes service -> pod 100.65.107.133 on node with calico 3.31.4

```
# request to kubernetes service from hostnetwork pod
12:24:33.244816 eno1  Out IP node1.47053 > node2.4789: VXLAN, flags [I] (0x08), vni 4096
IP node1.43700 > 100.65.107.133.15010: Flags [S], seq 1137234856, win 64860, options [mss 1410,sackOK,TS val 189658141 ecr 0,nop,wscale 7], length 0
# response Synack, not delivered to pod
12:24:33.245120 eno1  In  IP node2.48238 > node1.4789: VXLAN, flags [I] (0x08), vni 4096
IP 100.65.107.133.15010 > node1.43700: Flags [S.], seq 558739255, ack 1137234857, win 64308, options [mss 1410,sackOK,TS val 2058776255 ecr 189658141,nop,wscale 7], length 0
# retransmit
12:24:34.276628 eno1  In  IP node2.40538 > node1.4789: VXLAN, flags [I] (0x08), vni 4096
IP 100.65.107.133.15010 > node1.43700: Flags [S.], seq 558739255, ack 1137234857, win 64308, options [mss 1410,sackOK,TS val 2058777287 ecr 189658141,nop,wscale 7], length 0
...

```

bpf trace logs look fine, same as on the working 3.31->3.31 hostnetwork<->service pods (including the fib failure)

```
          <idle>-0       [003] ..s2. 306139.765821: bpf_trace_printk: eno1------------I: New packet at ifindex=4; mark=0
          <idle>-0       [003] ..s2. 306139.765823: bpf_trace_printk: eno1------------I: IP id=48833 len=110
          <idle>-0       [003] ..s2. 306139.765825: bpf_trace_printk: eno1------------I: IP s=10.238.194.8 d=10.238.196.2
          <idle>-0       [003] ..s2. 306139.765826: bpf_trace_printk: eno1------------I: IP ihl=20 bytes
[...]

PR #12054, opened 2026-03-10
Fix BPF VXLAN packet drop during rolling upgrade to 3.31

tomastigera:
## Summary

- Fix VXLAN packets being dropped on 3.30 nodes during a rolling upgrade to 3.31. In 3.31 the VXLAN device uses flow-based mode, which changes the inner destination MAC in encapsulated packets. When a 3.30 node receives these packets, the kernel sets `pkt_type` to `PACKET_OTHERHOST` (because the inner MAC doesn't match `vxlan.calico`), causing them to be silently dropped after decapsulation.
- The fix calls `bpf_skb_change_type(PACKET_HOST)` for allowed packets arriving on the VXLAN interface, ensuring the kernel delivers them correctly regardless of the inner MAC address.

Fixes https://github.com/projectcalico/calico/issues/11910

## Release note
```release-note
ebpf: Fixed issue where VXLAN traffic was dropped on nodes still running Calico 3.30 during a rolling upgrade to 3.31. This caused connectivity failures for host-networked pods communicating via Kubernetes services across nodes at different versions. Upgrading to this 3.30 patch release before upgrading to 3.31 resolves the issue.
```

## Test plan

- [ ] Deploy a mixed 3.30/3.31 cluster with BPF dataplane and VXLAN encapsulation
- [ ] Verify host-networked pod → Kubernetes service → pod connectivity works across 3.31→3.30 node boundaries
- [ ] Verify 3.30→3.30 traffic is unaffected (no-op path)
- [ ] Run BPF FV tests with VXLAN enabled

### DIFF
--- a/felix/bpf-gpl/fib_co_re.h
+++ b/felix/bpf-gpl/fib_co_re.h
@@ -6,6 +6,7 @@
 #define __CALI_FIB_CO_RE_H__
 
 #include "profiling.h"
+#include <linux/if_packet.h>
 
 static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
 {
@@ -358,6 +359,9 @@ allow:
 			CALI_INFO("Final result=DENY (%d). Program execution time: %lluns",
 					reason, prog_end_time-state->prog_start_time);
 		} else {
+			if (CALI_F_VXLAN && CALI_F_TO_HOST) {
+				bpf_skb_change_type(ctx->skb, PACKET_HOST);
+			}
 			CALI_INFO("Final result=ALLOW rc %d. Program execution time: %lluns",
 					rc, prog_end_time-state->prog_start_time);
 		}


felix/bpf-gpl/fib_co_re.h
```diff
@@ -358,6 +359,9 @@ static CALI_BPF_INLINE int forward_or_drop(struct cali_tc_ctx *ctx)
 			CALI_INFO("Final result=DENY (%d). Program execution time: %lluns",
 					reason, prog_end_time-state->prog_start_time);
 		} else {
+			if (CALI_F_VXLAN && CALI_F_TO_HOST) {
+				bpf_skb_change_type(ctx->skb, PACKET_HOST);
```
aaaaaaaalex, line 363: Any issue with the fact that we are now _always_ changing the type? Does it remove ang benefits to having the other type?

tomastigera: The unconditional CALI_F_VXLAN && CALI_F_TO_HOST check is the right approach — no need to match a specific MAC. It's a no-op for 3.30→3.30 traffic (where pkt_type is already PACKET_HOST), and fixes 3.31→3.30 traffic regardless of what the inner MAC happens to be.

PR #12064, opened 2026-03-11
[CORE-12269] Seamless live migration for OpenStack

nelljerram:
This PR updates Calico's OpenStack driver code to use the the new Calico-internal Live Migration API, so as to gain more seamless live migration support.  The key changes here are:
1. Using the LiveMigration API to describe a live migration operation from the "pre-live-migration" point.
2. Creating the WorkloadEndpoint object on the target node at the "pre-live-migration" point, instead of waiting until OpenStack says that live migration has completed.
3. Enabling the use of `live_migration_wait_for_vif_plug`, by signaling into Neutron when initial networking and security programming is in place for the migrating VM on the target node.

**Release note:**
```release-note
Make live migration more seamless in Calico for OpenStack; specifically in that there should be much less latency after the VM becomes active on the target node, before network traffic can be successfully sent to and from that VM.
```

### DIFF
--- a/networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
+++ b/networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
@@ -129,13 +129,9 @@ class WorkloadEndpointSyncer(ResourceSyncer):
                 )
 
         # Delete orphaned LiveMigration resources.
-        for name, _, mod_revision in datamodel_v3.get_all(
-            "LiveMigration", namespace
-        ):
+        for name, _, mod_revision in datamodel_v3.get_all("LiveMigration", namespace):
             if name not in expected_lm_names:
-                LOG.warning(
-                    "LiveMigration resync: deleting orphaned LM %s", name
-                )
+                LOG.warning("LiveMigration resync: deleting orphaned LM %s", name)
                 self.delete_live_migration(name, mod_revision=mod_revision)
 
         LOG.info("LiveMigration resync done")
--- a/networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_plugin_etcd.py
+++ b/networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_plugin_etcd.py
@@ -2080,7 +2080,6 @@ class TestLiveMigration(TestPluginEtcdBase):
         # Should NOT have called notify_port_active_direct.
         self.db.nova_notifier.notify_port_active_direct.assert_not_called()
 
-
     def _trigger_resync(self):
         """Trigger a periodic resync by advancing simulated time.
 
@@ -2107,12 +2106,8 @@ class TestLiveMigration(TestPluginEtcdBase):
         self._trigger_resync()
 
         # The resync should have created the LiveMigration and dest WEP.
-        self.assertIn(
-            self._lm_key(self.DEST_HOST), self.recent_writes
-        )
-        self.assertIn(
-            self._ep_key(self.DEST_HOST), self.recent_writes
-        )
+        self.assertIn(self._lm_key(self.DEST_HOST), self.recent_writes)
+        self.assertIn(self._ep_key(self.DEST_HOST), self.recent_writes)
 
     def test_resync_deletes_stale_live_migration(self):
         """Resync deletes orphaned LiveMigration with no migrating port."""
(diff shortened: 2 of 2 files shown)


networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -978,16 +1053,58 @@ def update_port_postcommit(self, context):
...
+
+                # Create destination WEP.  Skip DB re-read because this
+                # is a synthetic port dict with the destination host.
+                dest_port = port.copy()
+                dest_port["binding:host_id"] = dest_host
+                self.endpoint_syncer.write_endpoint(
+                    dest_port, plugin_context, reread=False
+                )
+
+                # Create LiveMigration resource.
```
fasaxc: Should you make the migration resource first?   If felix sees WEP before migration then it'll treat it as non-migration WEP and do route programming.  In etcd land, write ordering is generally preserved so it seems sensible to write in the correct order.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -691,6 +694,36 @@ def _try_to_update_port_status(self, admin_context, port_status_key):
...
+                            port_id,
+                            hostname,
+                        )
+                        # notify_port_active_direct expects a db
+                        # model (not a dict), matching the pattern
+                        # used by OVN and ML2 RPC callers.
+                        db_port = ml2_db.get_port(admin_context, port_id)
+                        if db_port:
+                            self.db.nova_notifier.notify_port_active_direct(db_port)
+                except Exception:
```
fasaxc: Seems like a very broad except given that you don't re-raise; do we really want to squash all exceptions here?

networking-calico/networking_calico/datamodel_v3.py
```diff
@@ -225,6 +230,15 @@ def delete(resource_kind, namespace, name, mod_revision=None):
     return etcdv3.delete(key, mod_revision=mod_revision)
 
 
+def get_uid(resource_kind, namespace, name):
+    """Read the metadata UID of a Calico v3 resource."""
+    try:
+        value, _ = _get_with_metadata(resource_kind, namespace, name)
+        return value.get("metadata", {}).get("uid", "unknown")
+    except Exception:
```
fasaxc: Another broad except; do we really want to squash this?

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -959,6 +992,48 @@ def update_port_postcommit(self, context):
...
+                        migration_uid,
+                        port["id"],
+                        original["binding:host_id"],
+                        port["binding:host_id"],
+                    )
+                    self.endpoint_syncer.delete_endpoint(original)
+
+                # Skip all remaining processing — migration cleanup is
+                # complete.
+                return
```
fasaxc: Any possibility that other changes are batched into one call and this early return will ignore them?

felix/dataplane/linux/live_migration.go
```diff
@@ -0,0 +1,408 @@
...
+	return updates
+}
+
+func (m *liveMigrationMonitor) OnUpdate(protoBufMsg any) {
+	switch msg := protoBufMsg.(type) {
+	case *proto.WorkloadEndpointUpdate:
+		id := types.ProtoToWorkloadEndpointID(msg.GetId())
+		m.ifaceNames[id] = msg.Endpoint.Name
+		if uid := msg.Endpoint.LiveMigrationUid; uid != "" {
+			m.migrationUIDs[id] = uid
```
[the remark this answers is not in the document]
nelljerram: Good spot, but in this case I think it's more helpful to continue logging the old UID in case there are any further FSM transitions (for the same FSM instance) after the LiveMigration resource has been deleted.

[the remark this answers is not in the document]
nelljerram: (I've added a comment to say that in the code.)

felix/calc/live_migration_calculator.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	return
+}
+
+func (lmc *LiveMigrationCalculator) withRoleUpdateIfNeeded(wepData *wepData, updateFunc func()) {
+	oldRole, _ := lmc.liveMigrationRoleAndUID(wepData)
+	updateFunc()
+	newRole, newUID := lmc.liveMigrationRoleAndUID(wepData)
+	if newRole != oldRole {
```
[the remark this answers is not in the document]
nelljerram: This is a trickier but similar case.  The main transient overlap case will be when role changes for the same WEP from TARGET to SOURCE.  Even though our code allows for it, I'm not sure if there is ever going to be a real overlap case with the role staying the same.  Until we see a real case like that, on balance I think it's better for logs, for a given FSM instance, to continue showing the same UID.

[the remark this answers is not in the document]
nelljerram: Again, I've added a comment in the code.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -691,6 +692,31 @@ def _try_to_update_port_status(self, admin_context, port_status_key):
...
+                        )
+                        LOG.info(
+                            "Live migration %s: destination port %s "
+                            "active on %s, notifying Nova",
+                            migration_uid,
+                            port_id,
+                            hostname,
+                        )
+                        self.db.nova_notifier.notify_port_active_direct(port)
+                except Exception:
```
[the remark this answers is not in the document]
nelljerram: The code here has since moved on, and this comment is outdated.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: Agreed, changed to write LiveMigration first, then WEP.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: Agreed; the preceding code in this function didn't catch any exceptions, except for "defensive" ones which I take to mean "not actually expected in modern OpenStack".  And I'm not aware of any exceptions that are expected in this new code.

We do want to take care to avoid terminating the loop in `_loop_writing_port_statuses` though.  Therefore I've removed the except here but added a broad except in `_loop_writing_port_statuses`.

networking-calico/networking_calico/datamodel_v3.py
nelljerram: Changed this one to just `etcdv3.KeyNotFound`, consistent with elsewhere in this file.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: Agreed, have reworked a little so as to cover other port changes, and to be generally a bit tidier.

nelljerram:
> Some comments inline; also looks like resync processing is missing so we won't clean up orphaned live migration objects or recreate them if they're missing. When server is restarted, port updates may have been lost leading to incorrect live migration state

Thanks, good catch; I've added coding and tests for this now.

PR #12067, opened 2026-03-11
[CORE-12378] Enable (optional) publishing for openstack packages on PRs

coutinhop:
**Release note:**
```release-note
TBD
```

### DIFF
--- a/.semaphore/push-images/packaging.yaml
+++ b/.semaphore/push-images/packaging.yaml
@@ -35,20 +35,22 @@ global_job_config:
       - install-package --no-install-recommends devscripts moreutils patchelf jq
 
 blocks:
+  # This promotion is only _automatic_ for the master branch.  For other
+  # branches, and PRs, it is available but not automatic.  This means a
+  # developer can trigger the building and publishing of release branch and
+  # pre-merge packages by clicking the "Publish openstack packages" button in
+  # the Semaphore UI.  When building from the master branch, packages are
+  # published to the "master" PPA.  When building from any other branch,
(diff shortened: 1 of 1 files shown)


.semaphore/push-images/packaging.yaml
```diff
@@ -35,20 +35,22 @@ global_job_config:
...
-      # packages" button in the Semaphore UI.  When building from the master
-      # branch, packages are published to the "master" PPA.  When building from
-      # any other branch, packages are published to the "testing" PPA.
-      when: "branch !~ '.+'"
     task:
       jobs:
         - name: "packages"
           commands:
             - if [ -z "${SEMAPHORE_GIT_PR_NUMBER}" ]; then make -C release/packaging release-publish VERSION=$SEMAPHORE_GIT_BRANCH; fi
+            - if [ -n "${SEMAPHORE_GIT_PR_NUMBER}" ]; then make -C release/packaging release-publish VERSION=pr-$SEMAPHORE_GIT_PR_NUMBER; fi
```
danudey, line 56: Is this going to create an appropriate (valid) version number in the Debian packages? What is the version number going to look like as a result?

Also, if we're building these packages on every PR build it might make sense to add this build to the 'Push calico-node images' promotion, since it will avoid re-building libbpf/felix a second time. Alternately, if we could cache those build results...

danudey: Edit: we're removing image push promotions, so scratch that part.

danudey (approved):

PR #12071, opened 2026-03-11
Fix three Felix FV test flakes

caseydavenport:
## Summary

Three independent flake fixes for Felix FV tests, one commit each.

### 1. PersistentConnection.IsConnectionReset race (batch 3)

The "should have connectivity after a backend is replaced by a new one" test
waits for a TCP RST in tcpdump, then immediately checks `pc.IsConnectionReset()`.
The RST has arrived at the network level but the test-connection process hasn't
exited yet, so `connectionReset` is still false.

Fix: make `connectionReset` access thread-safe and use `Eventually` for the check.

```
[FAIL] BPF tests (ipv4 tcp, ct=true, log=debug, tunnel=wireguard, dsr=false)
  ... should have connectivity after a backend is replaced by a new one
  felix/fv/bpf_test.go:3676
```

### 2. BPF affinity map assertions inside Eventually (batch 13)

The affinity map test used hard `Expect` assertions inside an `Eventually`
callback. When the BPF datapath hasn't written the affinity entry yet, the
`Expect` panics. Ginkgo catches it and retries, but this interacts poorly with
the expensive `cc.CheckConnectivity()` call that also runs on each retry.

Also removed dead code -- the first `if testOpts.ipv6` block returned early,
making the duplicate IPv6 handling below it unreachable.

Fix: replace `Expect` with soft conditional checks that return nil to trigger retry.

```
[FAIL] BPF tests (ipv6 tcp, ct=true, log=debug, tunnel=none, dsr=false)
  ... with affinity should have connectivity from a workload to a
  service with multiple backends
  felix/fv/bpf_test.go:3414
```

### 3. Container startup retry on transient docker failures (batch 16)

The apiserver container got a 504 Gateway Timeout during image pull, causing
`docker run` to exit immediately. `WaitUntilRunning` detected this and failed
the test with no retry.

Fix: retry `docker run` up to 3 times with backoff on transient startup failures.

```
[FAIL] cluster routing using Felix programming routes ... [BeforeEach]
  should have correct connectivity
  felix/fv/containers/containers.go:638
  Container apiserver-...-felixfv failed before being listed in 'docker ps'
  stderr: docker: received unexpected HTTP status: 504 Gateway Time-out.
```

### DIFF
--- a/felix/fv/bpf_test.go
+++ b/felix/fv/bpf_test.go
@@ -3441,7 +3441,10 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 									return nil
 								}
 								return aff[mkey.(nat.AffinityKey)].Backend()
-							}, 60*time.Second, time.Second).ShouldNot(Equal(mVal.Backend()))
+							}, 60*time.Second, time.Second).Should(SatisfyAll(
+								Not(BeNil()),
+								Not(Equal(mVal.Backend())),
+							))
 						})
 					}
 
@@ -3677,7 +3680,7 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 							Eventually(func() int { return tcpd.MatchCount("tcp-rst") }, "25s").ShouldNot(BeZero(),
 								"Expected to see TCP RSTs on the connection after backend change")
 							Eventually(pc.IsConnectionReset, "5s").Should(BeTrue(),
-							"Expected the persistent connection to detect the reset after RST seen in tcpdump")
+								"Expected the persistent connection to detect the reset after RST seen in tcpdump")
 						} else {
 							prevCount = pc.PongCount()
 							Eventually(pc.PongCount, "15s").Should(BeNumerically(">", prevCount),
--- a/felix/fv/containers/containers.go
+++ b/felix/fv/containers/containers.go
@@ -654,13 +654,18 @@ func (c *Container) waitUntilRunning() error {
 
 	// Set up so we detect if container startup fails.
 	stoppedChan := make(chan struct{})
+	// Capture runCmd locally so the goroutine doesn't race with a retry
+	// that reassigns c.runCmd.
+	runCmd := c.runCmd
 	go func() {
 		defer close(stoppedChan)
-		err := c.runCmd.Wait()
+		err := runCmd.Wait()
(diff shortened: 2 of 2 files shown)


felix/fv/containers/containers.go
```diff
@@ -242,33 +244,56 @@ func RunWithFixedName(name string, opts RunOpts, args ...string) (c *Container)
...
+				"container": c.Name,
+				"attempt":   attempt,
+				"lastErr":   lastErr,
+			}).Warn("Retrying container startup after transient failure")
+			// Wait for log goroutines from the previous attempt to finish.
+			c.logFinished.Wait()
+			// Clean up any remnants of the failed container.
+			cleanCmd := utils.Command("docker", "rm", "-f", c.Name)
+			_ = cleanCmd.Run()
+			time.Sleep(time.Duration(attempt-1) * time.Second)
```
caseydavenport: This feels excessive - can just sleep 1?

felix/fv/containers/containers.go
```diff
@@ -242,33 +244,56 @@ func RunWithFixedName(name string, opts RunOpts, args ...string) (c *Container)
...
+
+		// Note: it might take a long time for the container to start running,
+		// e.g. if the image needs to be downloaded.
+		lastErr = c.waitUntilRunning()
+		if lastErr == nil {
+			break
+		}
+		log.WithError(lastErr).WithField("container", c.Name).Warn("Container failed to start")
+	}
+	Expect(lastErr).NotTo(HaveOccurred(),
```
caseydavenport: Can just be a single line.

felix/fv/bpf_test.go
```diff
@@ -3406,37 +3406,36 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
-									Expect(aff).To(HaveKey(mkey.(nat.AffinityKeyV6)))
-
-									return aff[mkey.(nat.AffinityKeyV6)].Backend()
-								}
-
-								if testOpts.ipv6 {
-									aff := dumpAffMapV6(tc.Felixes[0])
-									Expect(aff).To(HaveLen(0))
+									if len(aff) != 0 {
+										return nil // Map not empty yet after delete, retry.
```
caseydavenport: No inline comments please.

PR #12087, opened 2026-03-12
Add e2e tests on GCP kubeadm promotion for local binary testing

lwr20:
Add a new Semaphore pipeline that runs e2e tests against a gcp-kubeadm cluster using the locally-built e2e binary (via RUN_LOCAL_TESTS). This triggers automatically when files in the e2e/ directory change.

The pipeline provisions a gcp-kubeadm cluster via banzai, builds the e2e binary from the monorepo checkout, and runs sig-calico Conformance tests with the CalicoBPF dataplane using parallel ginkgo execution.

**Release note:**
```release-note
TBD
```

### DIFF
--- a/.semaphore/semaphore-scheduled-builds.yml
+++ b/.semaphore/semaphore-scheduled-builds.yml
@@ -581,7 +581,9 @@ blocks:
       secrets:
         - name: banzai-secrets
       prologue:
-        commands_file: .semaphore/end-to-end/scripts/global_prologue.sh
+        commands:
+          - checkout
+          - source .semaphore/end-to-end/scripts/global_prologue.sh
       epilogue:
         always:
           commands:
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -581,7 +581,9 @@ blocks:
       secrets:
         - name: banzai-secrets
       prologue:
-        commands_file: .semaphore/end-to-end/scripts/global_prologue.sh
+        commands:
+          - checkout
+          - source .semaphore/end-to-end/scripts/global_prologue.sh
       epilogue:
         always:
           commands:
--- a/.semaphore/semaphore.yml.d/blocks/20-e2e-gcp.yml
+++ b/.semaphore/semaphore.yml.d/blocks/20-e2e-gcp.yml
@@ -7,7 +7,9 @@
     secrets:
       - name: banzai-secrets
     prologue:
-      commands_file: .semaphore/end-to-end/scripts/global_prologue.sh
+      commands:
+        - checkout
+        - source .semaphore/end-to-end/scripts/global_prologue.sh
     epilogue:
       always:
         commands:


.semaphore/end-to-end/pipelines/e2e-gcp.yml
```diff
@@ -0,0 +1,74 @@
...
+          commands:
+            - ~/calico/.semaphore/end-to-end/scripts/body_standard.sh
+          matrix:
+            - env_var: PROVISIONER
+              values: ["gcp-kubeadm"]
+            - env_var: K8S_VERSION
+              values: ["stable"]
+            - env_var: INSTALLER
+              values: ["operator"]
+            - env_var: DATAPLANE
```
caseydavenport: I think we should have USE_API_SERVER=false for these (selfishly)

.semaphore/end-to-end/pipelines/e2e-gcp.yml
```diff
@@ -0,0 +1,74 @@
...
+          execution_time_limit:
+            hours: 3
+          commands:
+            - ~/calico/.semaphore/end-to-end/scripts/body_standard.sh
+          matrix:
+            - env_var: PROVISIONER
+              values: ["gcp-kubeadm"]
+            - env_var: K8S_VERSION
+              values: ["stable"]
+            - env_var: INSTALLER
```
caseydavenport: We should figure out how we can install the local code from the PR / commit on this test env (as a follow up).

.semaphore/end-to-end/pipelines/e2e-gcp.yml
```diff
@@ -0,0 +1,74 @@
...
+    - name: TEST_TYPE
+      value: k8s-e2e
+    - name: PRODUCT
+      value: calico
+    - name: FUNCTIONAL_AREA
+      value: "e2e-gcp.yml"
+    - name: RUN_LOCAL_TESTS
+      value: "true"
+    - name: K8S_E2E_FLAGS
+      value: --ginkgo.focus=(\[sig-calico\].*\[Conformance\]) --ginkgo.skip=(\[Slow\]|\[Disruptive\]|\[DataPath\])
```
caseydavenport: For this "real cluster" run, I think we want to run more than just the conformance tests - we should run all Calico tests that we think will pass on this env.

.semaphore/semaphore.yml.d/03-promotions.yml
```diff
@@ -103,6 +103,10 @@ promotions:
     pipeline_file: push-images/packaging.yaml
     auto_promote:
       when: "branch =~ 'master'"
+  - name: E2E tests on GCP kubeadm
```
caseydavenport: Since this is a promotion, does that mean it will only run _after_ all other jobs finish? 

I think we want this to run in parallel with other jobs to minimize wall clock time?

PR #12089, opened 2026-03-12
BIRD config: propagate route priority via BGP LOCAL_PREF for live migration

nelljerram:
Support live migration by propagating elevated route priority through BGP between BIRD instances on different nodes.

API: Add IPv4NormalRoutePriority and IPv6NormalRoutePriority fields to BGPConfigurationSpec, so BIRD knows the boundary between normal and elevated-priority routes (default 1024, matching Felix).

Import filter (buildImportFilter): Default krt_metric to NormalRoutePriority on all imports. For iBGP peers, convert BGP LOCAL_PREF back to krt_metric (inverting: krt_metric = INT_MAX - bgp_local_pref). If the resulting krt_metric is below NormalRoutePriority, set BIRD preference=200 so the BGP-learned route overrides the local workload route in the kernel.

Export filter (buildExportFilter): For iBGP peers, convert krt_metric to BGP LOCAL_PREF (inverting: bgp_local_pref = INT_MAX - krt_metric) before calling calico_export_to_bgp_peers.

Aggregation (calico_aggr templates): Accept elevated-priority routes (those with krt_metric < NormalRoutePriority) within a confirmed block, between the block-level exact-match accept and the block-level subnet reject. This lets /32 live migration routes pass through aggregation instead of being dropped.

Mesh peers now use buildImportFilter/buildExportFilter (same as explicit peers) instead of hardcoded filter strings, unifying filter generation.

**Release note:**
```release-note
Calico's BIRD configuration (generated by the confd subcomponent of calico-node) has been enhanced to propagate route priorities within a contiguous AS (autonomous system).  This is one of a set of changes needed for seamless live migration of VMs.
```


confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -587,10 +620,30 @@ func truncateBGPFilterName(name string) string {
...
+		fmt.Sprintf("krt_metric = %d;", normalRoutePriority),
+	)
+
+	// For iBGP peers, convert from LOCAL_PREF to krt_metric.  Higher LOCAL_PREF = higher
+	// priority, but lower krt_metric = higher priority, so we invert: krt_metric = INT_MAX -
+	// bgp_local_pref.
+	if peerAS == nodeAS {
+		filterLines = append(filterLines,
+			"if (defined(bgp_local_pref)) then {",
+			"  krt_metric = 2147483647 - bgp_local_pref;",
```
song-jiang: Suggest to define a constant `birdIntMaxValue` and update this code with `  fmt.Sprintf("bgp_local_pref = %d - krt_metric;", birdIntMaxValue)`

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -612,14 +665,37 @@ func (c *client) buildImportFilter(filters []string, ipVersion int) string {
 		}
 	}
 
-	filterLines = append(filterLines, "accept; # Prior to introduction of BGP Filters we used \"import all\" so use default accept behaviour on import")
+	// If we now have a higher than normal priority route, set 'preference' attribute to allow
+	// it to override an existing local workload route in the kernel.
+	filterLines = append(filterLines,
+		fmt.Sprintf("if (krt_metric < %d) then", normalRoutePriority),
+		"  preference = 200;",
```
song-jiang: Better to define a constant for preference as well.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -612,14 +665,37 @@ func (c *client) buildImportFilter(filters []string, ipVersion int) string {
...
+	ipVersion, normalRoutePriority int,
+) string {
 	var filterLines []string
 
+	// For iBGP peers, convert from krt_metric to LOCAL_PREF.  Higher LOCAL_PREF = higher
+	// priority, but lower krt_metric = higher priority, so we invert: bgp_local_pref = INT_MAX
+	// - krt_metric.
+	if peerAS == nodeAS {
+		filterLines = append(filterLines,
+			"bgp_local_pref = 2147483647 - krt_metric;",
```
song-jiang: Suggest to use `birdIntMaxValue` constant.

I have a concern that this apply to all exported routes include the aggregated route for IPAM blocks, service IPs for service advertisement etc. Since those routes are from `static` protocol and has no krt_metric value, the template always generates max value on `bgp_local_pref `and be treated as high priority routes on iBGP peers. We should have something like `if (defined(krt_metric)), bgp_local_pref = 2147483647 - krt_metric;`

song-jiang (changes requested):
Looking good with a few nits.

confd/pkg/backends/calico/bgp_processor.go
nelljerram: Good catch, thanks.  But in such cases I think we should default krt_metric to normalRoutePriority.  I'll do that.

PR #12090, opened 2026-03-12
Add make get-operator-crds to ci-preflight-checks

MichalFupso:
Add a get-operator-crds to make ci-preflight-checks. This was causing issues with hashreleases where during PR we did not check for the generated files, but that target was run during hashrelease job. This would fail hashrelease as the files were marked as dirty

**Release note:**
```release-note
TBD
```


caseydavenport:
@MichalFupso we used to have this, but removed it based on a discussion with @fasaxc - so you guys should chat :p

PR #12101, opened 2026-03-12
Run get-operator-crds before dirty check

alexh-tigera:
The hashrelease stage runs a check to make sure the operator CRDs are up to date. We should run this at the start of the build to fail fast.

**Release note:**
```release-note
TBD
```


alexh-tigera:
> We used to have this and removed it because it was too annoying. We shouldn't add it back.

Ok, makes sense!

PR #12102, opened 2026-03-12
Remove unused docker-credential-gcr dependency from Windows release targets

radixo:
### Summary
- Remove unused `docker-credential-gcr` download and Make targets from `lib.Makefile`
- This dependency was left behind after the OIDC auth migration (commit 08ead585d4) which switched crane to run directly on the host instead of inside a Docker container
- The binary was being downloaded but never executed, and its download failure (GitHub 502) was causing hashrelease builds to fail

### Problem
The `release-windows-with-tag` target had `docker-credential-gcr-binary` as a prerequisite, which downloads a binary from GitHub releases. When GitHub returned transient 502 errors, the entire hashrelease pipeline failed — even though the binary is no longer used.

Example failure: https://tigera.semaphoreci.com/jobs/0dfbda26-784c-4756-8cd4-bba5ebe6d06a

**Release note:**
```release-note-not-required
TBD
```


lib.Makefile
coutinhop: `CRANE_BINDMOUNT` and `CRANE_BINDMOUNT_CMD` seem to also no longer be needed

```suggestion

```

coutinhop: `CRANE_BINDMOUNT` and `CRANE_BINDMOUNT_CMD` seem to also no longer be needed

```suggestion

```

PR #12103, opened 2026-03-12
Make publishing images more resilient to transient external registry issues 

stevegaossou:
Add a retry loop around the docker manifest create command in lib.Makefile to handle transient registry errors.

Changes in this PR: 
  - MANIFEST_RETRIES (default 5) — number of attempts for each docker manifest command
  - MANIFEST_RETRY_DELAY (default 5 seconds) — wait between retries
  - Both manifest create and manifest push are wrapped in retry loops that:
    - break on success
    - Log a warning with the attempt number on failure
    - exit 1 after exhausting all retries (preserving the original failure behavior)
  - Both variables use ?= so they can be overridden from the command line or environment if needed
  
### Background

Here's some context for what this PR is trying to guard against. 

We sometimes run into issues when attempting to publish images to Quay (e.g. our hashrelease images). 

You'll see in our builds a transient quay.io registry error like this: 
```
  docker manifest create quay.io/calico/envoy-proxy:master --amend quay.io/calico/envoy-proxy:master-amd64 --amend
  quay.io/calico/envoy-proxy:master-arm64
  unsupported manifest media type and no default available: application/grpc
```

The assumption is: 
- This is an internal quay.io infrastructure issue (gRPC is used internally between quay components, and it leaked through to the client)
- It is transient 

We often kick CI builds after running into Quay issues and the problem vanishes on follow-up builds. 

**Release note:**
```release-note
TBD
```

### DIFF
--- a/lib.Makefile
+++ b/lib.Makefile
@@ -1031,41 +1031,42 @@ ifeq ($(findstring quay.io,$(REGISTRY)),quay.io)
 	)
 endif
 
+# retry_docker_cmd retries a docker command up to a specified number of times.
+# Usage: $(call retry_docker_cmd,<description>,<docker command>,<max retries>,<retry delay>)
+define retry_docker_cmd
+	i=1; \
+	while [ $$i -le $(3) ]; do \
+		$(2) && break; \
+		echo "WARNING: $(1) failed (attempt $$i/$(3)), retrying in $(4)s..."; \
+		if [ $$i -eq $(3) ]; then exit 1; fi; \
+		sleep $(4); \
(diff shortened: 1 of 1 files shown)


stevegaossou:
Going to merge this. The CI build failed on a single eBPF test flake: 

```
FV: Felix 24.04 BPF (iptables)
- [It] _BPF-SAFE_ endpoint-to-host-action tests (kubernetes backend) endpoint to host tests INPUT=DROP, ep-to-host=Accept, HEP=*, HEP-pol=Allow, expected connectivity=true

```

PR #12111, opened 2026-03-13
[BPF] Migrate BPF events from perf event array to ring buffer

lucastigera:
## Context
Felix's BPF dataplane currently delivers all kernel events through a single shared `BPF_MAP_TYPE_PERF_EVENT_ARRAY` (`cali_perf_evnt`), multiplexing them through this one channel and demuxing in Go by a type header field. Since this per-CPU map is shared across all event types, high event volume from any single source creates pressure on the entire channel, increasing the risk of map overflow and dropped events.

This PR replaces the shared perf array with a single shared `BPF_MAP_TYPE_RINGBUF`, which is more memory efficient and paves the way for events that might require separated ring buffers.

### Problems with the perf event array
1. **Per-CPU memory waste.** `BPF_MAP_TYPE_PERF_EVENT_ARRAY` allocates a separate mmap'd ring per CPU. On a 64-core node that's 64 independent rings. Each must be sized for peak load, and unused capacity on one CPU cannot be used by another.

2. **Context pointer requirement.** `bpf_perf_event_output()` requires a program context pointer (`struct __sk_buff *` for TC, `struct pt_regs *` for kprobes). This restricts which BPF program types can emit events. Syscall tracepoints, for example, have a raw `void *ctx` that doesn't satisfy TC context requirements.

3. **Single consumer bottleneck.** The Go-side `bpfEventPoller` reads all events sequentially in one goroutine, parsing headers to dispatch by type. A slow handler for one event type delays consumption of all others.

### Benefits of a ring buffer

1. **Memory efficiency.** `BPF_MAP_TYPE_RINGBUF` is a single shared ring (not per-CPU). Total memory usage is lower, and all CPUs contribute to and consume from the same ring.

2. **Reserve-then-write semantics.** `bpf_ringbuf_reserve()` returns a pointer directly into ring memory. The BPF program writes fields in-place — no stack bounce needed. This is critical for large events (e.g., 572-byte TLS events exceed the 512-byte BPF stack limit for `bpf_perf_event_output`).

3. **No context pointer required.** `bpf_ringbuf_reserve()`/`bpf_ringbuf_submit()` do not require a program context pointer, enabling any BPF program type (including syscall tracepoints) to emit events.

## Changes summary
- Replace `BPF_MAP_TYPE_PERF_EVENT_ARRAY` (`cali_perf_evnt`) with `BPF_MAP_TYPE_RINGBUF` (`cali_rb_evnt`) for all BPF event delivery
- Add `cali_rb_drops` per-CPU counter map to track events dropped when the ring buffer is full, preserving `ErrLostEvents` metrics and rate-limited warning logs
- Scale ring buffer size by `NumCPU` to match the same total buffer capacity as the old per-CPU perf array
- Pure-Go mmap ring buffer reader (no CGo on the hot path), consistent with the existing perf reader pattern

**Release note:**
```release-note-not-required
[eBPF] Replace per-CPU perf event array with a shared ring buffer (BPF_MAP_TYPE_RINGBUF) for event delivery.
```

### DIFF
--- a/felix/bpf-gpl/ringbuf.h
+++ b/felix/bpf-gpl/ringbuf.h
@@ -24,13 +24,17 @@ struct {					\
 // max_entries IS the buffer size in bytes.
 CALI_RINGBUF(cali_rb_evnt, 1024 * 1024)
 
-// Shared drop counter and flush timestamp, protected by a spinlock.
+// Shared drop counter and flush timestamp.
 // Drops are emitted as EVENT_LOST_EVENTS through the ring buffer itself;
 // Go never reads this map directly.
+//
+// No bpf_spin_lock: kprobe/tracing programs cannot reference maps whose
+// value type contains a spin lock. Instead, count uses atomic XCHG
+// (__sync_lock_test_and_set) for the read-and-zero in flush, and
+// __sync_fetch_and_add for increments.
 #define CALI_RB_FLUSH_INTERVAL_NS (5ULL * 1000000000ULL) /* 5 seconds */
 
 struct rb_drops_val {
-	struct bpf_spin_lock lock;
 	__u64 count;
 	__u64 last_flush_ts;
 };
@@ -48,20 +52,16 @@ static CALI_BPF_INLINE void ringbuf_flush_drops(void)
 	}
 
 	__u64 now = bpf_ktime_get_ns();
-	
-	if (val->count == 0 || now - val->last_flush_ts < CALI_RB_FLUSH_INTERVAL_NS) {
+	if (now - val->last_flush_ts < CALI_RB_FLUSH_INTERVAL_NS) {
 		return;
 	}
 
-	bpf_spin_lock(&val->lock);
 	__u64 dropped = __sync_lock_test_and_set(&val->count, 0);
 	if (dropped == 0) {
-		bpf_spin_unlock(&val->lock);
 		return;
 	}
-	bpf_spin_unlock(&val->lock);
 
-	/* Cannot call bpf_ringbuf_output while holding the lock. */
+	/* Emit the accumulated drop count as a lost-events notification. */
 	struct {
 		struct event_header hdr;
 		__u64 count;
--- a/felix/bpf/ringbuf/ringbuf.go
+++ b/felix/bpf/ringbuf/ringbuf.go
@@ -104,14 +104,14 @@ func SetMapSize(size int) {
 }
 
 // DropsMap returns the shared array map used by the BPF side to track dropped
-// events. The value is struct rb_drops_val { bpf_spin_lock(4) + pad(4) +
-// count(8) + last_flush_ts(8) } = 24 bytes. Go never reads this map; it
-// exists only for EnsureExists() to create/pin the map.
+// events. The value is struct rb_drops_val { count(8) + last_flush_ts(8) } =
+// 16 bytes. Go never reads this map; it exists only for EnsureExists() to
+// create/pin the map.
 func DropsMap() maps.Map {
 	return maps.NewPinnedMap(maps.MapParameters{
 		Type:       "array",
 		KeySize:    4,
-		ValueSize:  24,
+		ValueSize:  16,
 		MaxEntries: 1,
 		Name:       DropsMapName,
 		Version:    1,
--- a/felix/bpf/ut/ringbuf_events_test.go
+++ b/felix/bpf/ut/ringbuf_events_test.go
@@ -152,9 +152,9 @@ func TestRingBufFillup(t *testing.T) {
 	// Drain any leftover events from previous tests and reset the drops map
 	// so we start with a completely clean state.
 	rb.Drain()
-	// Reset the single-entry drops map (struct rb_drops_val = 24 bytes).
+	// Reset the single-entry drops map (struct rb_drops_val = 16 bytes).
 	k := make([]byte, 4) // key = 0
-	zeroVal := make([]byte, 24)
+	zeroVal := make([]byte, 16)
 	err = ringBufDropsMap.Update(k, zeroVal)
 	Expect(err).NotTo(HaveOccurred())
 


felix/bpf-gpl/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+// Shared drop counter and flush timestamp.
+//   key 0: accumulated drop count
+//   key 1: last flush timestamp (nanoseconds, from bpf_ktime_get_ns)
+// Drops are emitted as EVENT_LOST_EVENTS through the ring buffer itself;
+// Go never reads this map directly.
+#define CALI_RB_DROPS_KEY_COUNT 0
+#define CALI_RB_DROPS_KEY_TS    1
+#define CALI_RB_FLUSH_INTERVAL_NS (5ULL * 1000000000ULL) /* 5 seconds */
+
+struct {
```
tomastigera: there is a macro for versioned maps which should be used

felix/bpf-gpl/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+	if (!last_ts) {
+		return;
+	}
+
+	__u64 now = bpf_ktime_get_ns();
+	if (now - *last_ts < CALI_RB_FLUSH_INTERVAL_NS) {
+		return;
+	}
+
+	__u64 dropped = __sync_lock_test_and_set(cnt, 0);
```
tomastigera: 4. Race window in drop counter flush (Low)

  ringbuf.h:72-80: Between __sync_lock_test_and_set(cnt, 0) and the bpf_ringbuf_output call, if the output fails and we __sync_fetch_and_add(cnt, dropped), concurrent CPUs may have already
  incremented cnt — so the restore adds back dropped on top of new increments. This is technically correct (no drops are lost), but the count in the next flush event could overcount. This is
  acceptable for metrics but worth a comment.

tomastigera: we could use spin lock

felix/bpf-gpl/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+static CALI_BPF_INLINE void ringbuf_flush_drops(void)
+{
+	__u32 cnt_key = CALI_RB_DROPS_KEY_COUNT;
+	__u64 *cnt = bpf_map_lookup_elem(&cali_rb_drops, &cnt_key);
+	if (!cnt || *cnt == 0) {
+		return;
+	}
+
+	__u32 ts_key = CALI_RB_DROPS_KEY_TS;
+	__u64 *last_ts = bpf_map_lookup_elem(&cali_rb_drops, &ts_key);
```
tomastigera: we do not use these directly so we do not need to fix it if we need to upgrade the map's version, see the macros for maps

felix/bpf-gpl/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+// event through the ring buffer, at most once every CALI_RB_FLUSH_INTERVAL_NS.
+static CALI_BPF_INLINE void ringbuf_flush_drops(void)
+{
+	__u32 cnt_key = CALI_RB_DROPS_KEY_COUNT;
+	__u64 *cnt = bpf_map_lookup_elem(&cali_rb_drops, &cnt_key);
+	if (!cnt || *cnt == 0) {
+		return;
+	}
+
+	__u32 ts_key = CALI_RB_DROPS_KEY_TS;
```
tomastigera: why not to have a struct in the map? A single lookup would provide both

felix/bpf-gpl/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	__u64 now = bpf_ktime_get_ns();
+	if (now - *last_ts < CALI_RB_FLUSH_INTERVAL_NS) {
+		return;
+	}
+
+	__u64 dropped = __sync_lock_test_and_set(cnt, 0);
+	if (dropped > 0) {
+		struct {
```
tomastigera: now would be the time to grab the lock

felix/bpf-gpl/ringbuf.h
```diff
@@ -0,0 +1,104 @@
...
+			},
+			.count = dropped,
+		};
+		if (bpf_ringbuf_output(&cali_rb_evnt, &evt, sizeof(evt), 0) != 0) {
+			/* Ring still full — restore counter. */
+			__sync_fetch_and_add(cnt, dropped);
+			return;
+		}
+		*last_ts = now;
+	}
```
tomastigera, line 82: now would be the time to release the lock

tomastigera (commented):
6. Missing unit tests for the Go ring buffer reader (Medium)

  The old perf_test.go had pure-Go unit tests for the ring parsing logic (wrapping, lost events, etc.) without needing real BPF programs. The new ringbuf.go has no equivalent — only BPF-level
   integration tests. Consider adding Go-level tests for readOne / roundupLen with synthetic mmap'd memory, especially for edge cases like wraparound reads and busy/discard bit handling.

  7. TypeLostEvents value change from 0 to 8 (Info)

  The old TypeLostEvents was 0 (a synthetic value from the perf reader, not a BPF event type). Now it's 8 (a real BPF event type EVENT_LOST_EVENTS). This is correct but is a semantic change —
   any downstream consumer checking for event type 0 would silently stop working. The PR correctly updates the Go constant. Just flagging for awareness during review.

felix/bpf-gpl/ringbuf.h
```diff
@@ -0,0 +1,106 @@
...
+
+	__u64 now = bpf_ktime_get_ns();
+
+	if (val->count == 0 || now - val->last_flush_ts < CALI_RB_FLUSH_INTERVAL_NS) {
+		return;
+	}
+
+	bpf_spin_lock(&val->lock);
+	__u64 dropped = val->count;
```
tomastigera: it should recheck the count holding the lock. The check o line 52 is a fast path, but some other CPU could have seen the same and got the lock before us.

The read should use test and set something like this:

```
  bpf_spin_lock(&val->lock);
  __u64 dropped = __sync_lock_test_and_set(&val->count, 0);
  if (dropped == 0) {
      bpf_spin_unlock(&val->lock);
      return;
  }
  bpf_spin_unlock(&val->lock);
```

felix/bpf/ringbuf/ringbuf_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+package ringbuf
+
+import (
+	"encoding/binary"
+	"testing"
+	"unsafe"
+
+	. "github.com/onsi/gomega"
+)
```
tomastigera, line 23: nit: real systems will have positions >> ring size; testing with e.g. startPos = 1<<40 would verify the mask arithmetic at scale

felix/bpf-gpl/ringbuf.h
lucastigera: nice catch!

What if we get the lock before the check?

```
	__u64 now = bpf_ktime_get_ns();
	
	bpf_spin_lock(&val->lock);
	if (val->count == 0 || now - val->last_flush_ts < CALI_RB_FLUSH_INTERVAL_NS) {
		bpf_spin_unlock(&val->lock);
		return;
	}

	__u64 dropped = val->count;
	val->count = 0;
	bpf_spin_unlock(&val->lock);
```

Could it slow down the fast path?

sridhartigera: I think Tomas' suggestion is correct. Lets say CPU0 flushes the counter. and CPU1 finds a drop and increments the counter.

CPU0 - Takes the lock, reads the counter. Lets say it reads a value 3.
CPU1 - without lock, its going to read and increment the value to 4, in one atomic operation using __sync_fetch_and_add.
CPU0 - Now resets to 0.

We will report the lost events as 3 whereas actually there are 4. 

When you do `__sync_lock_test_and_set` with lock,
CPU0 - Holds the lock, reads the value 3 and immediately sets it to 0 atomically.
CPU1 - Now increments the value to 1.

Eventually this totals to 4. 3 events reported lost in this window and 1 in the next window. The idea is if you read and increment atomically, you should read and set it to 0 atomically.

lucastigera: That's correct. I've updated the code to cover all val->count updates within a spin_lock block.

nelljerram:
@lucastigera Please can you add to the PR description to explain **why** you are making this change.

felix/bpf-gpl/ringbuf.h
lucastigera: For the record, after some back and forth, we decided not to use bpf_spin_lock for two main reasons:

1. Atomic operations don’t require lock protection. The increment path uses `__sync_fetch_and_add`, and the read-and-zero path uses `__sync_lock_test_and_set` - both are single atomic CPU instructions that are inherently safe under concurrency. Wrapping them in a spinlock would add overhead without any real benefit.
2. BPF spinlocks are not available in all program types. For example, kprobe programs cannot use bpf_spin_lock - the verifier rejects it - which would limit its applicability.

nelljerram:
Thanks so much @lucastigera for the description updates - very clear now.

PR #12122, opened 2026-03-13
IPIP: no need to program specific routes for remote tunnel endpoints

mazdakn:
For VXLAN encap, Felix programs routes for remote VTEPs like:
```
10.65.0.0/26 via 10.65.0.0 dev vxlan.calico onlink mtu 1450
10.65.0.1 dev vxlan.calico scope link mtu 1450
```

This is valid since VXLAN tunnels are L2 networks, and remote VTEPs can be reached directly.
For IPIP tunnels this is not the case as IPIP tunnels are L3, and we can't program such routes. However, so far Felix has been programming an exact route per remote tunnel endpoint like:

```
192.168.110.128 via 172.18.0.2 dev tunl0 proto 80 onlink mtu 1480
192.168.110.128/26 via 172.18.0.2 dev tunl0 proto 80 onlink mtu 1480
```

these exact routes are redundant and do nothing specific. This PR ensure Felix won't program such routes, and just keep the singular route for a allocated block.

The behavior for borrowed IPs is not changed.

UPDATE: The bug intended to fixed in this PR is actually addressed and fixed in this one: https://github.com/projectcalico/calico/pull/12180
This PR is now only include a few improvements, and more test coverage for the issue.

**Release note:**
```release-note
TBD
```

### DIFF
--- a/felix/dataplane/linux/ipip_mgr_test.go
+++ b/felix/dataplane/linux/ipip_mgr_test.go
@@ -390,7 +390,7 @@ var _ = Describe("IPIPManager", func() {
 		err := ipipMgr.CompleteDeferredWork()
 		Expect(err).NotTo(HaveOccurred())
 
-		// Expect no route.
+		// Expect one exact route.
 		Expect(rt.currentRoutes[dataplanedefs.IPIPIfaceName]).To(HaveLen(1))
 		Expect(rt.currentRoutes[dataplanedefs.IPIPIfaceName][0]).To(Equal(
 			routetable.Target{


felix/dataplane/linux/ipip_mgr.go
```diff
@@ -158,6 +158,10 @@ func (m *ipipManager) CompleteDeferredWork() error {
 }
 
 func (m *ipipManager) tunnelRoute(cidr ip.CIDR, r *proto.RouteUpdate) *routetable.Target {
+	if isRemoteVTEPRoute(r, proto.IPPoolType_IPIP) {
```
caseydavenport: A bit confusing to call a function with VTEP in the name on IP in IP routes :)

felix/dataplane/linux/route_mgr.go
```diff
@@ -181,7 +172,7 @@ func (m *routeManager) OnUpdate(protoBufMsg any) {
 			m.routesDirty = true
 		}
 
-		if isRemoteTunnelRoute(msg, m.ippoolType) {
+		if isRemoteVTEPRoute(msg, m.ippoolType) || isBorrowedRoute(msg, m.ippoolType) {
 			m.logCtx.WithField("msg", msg).Debug("Route manager received route update for remote tunnel endpoint")
```
caseydavenport, line 174: This log doesn't seem right any more, could be borrowed. Best to log precisely the type.

felix/dataplane/linux/route_mgr.go
```diff
@@ -129,31 +129,22 @@ func calculateRouteProtocol(dpConfig Config) netlink.RouteProtocol {
...
-	}
-
-	var isRemoteTunnel bool
-	var isBlock bool
-	isRemoteTunnel = isType(msg, proto.RouteType_REMOTE_TUNNEL)
-	isBlock = isType(msg, proto.RouteType_REMOTE_WORKLOAD)
+// This happens when tunnel addresses are selected from an IP pool with blocks of a single address.
+// These also need routes of the form "<IP> dev vxlan.calico" rather than "<block> via <TunnelEndpoint>".
+// This is only applicable to VXLAN encapsulation.
+func isRemoteVTEPRoute(msg *proto.RouteUpdate, ippoolType proto.IPPoolType) bool {
```
caseydavenport: Feels like the previous name made more sense? We still call this on both types of encap, we just handle those routes differently?

felix/dataplane/linux/ipip_mgr.go
mazdakn: Changed it back to `isRemoteTunnelRoute`

felix/dataplane/linux/route_mgr.go
mazdakn: changed it back.

felix/dataplane/linux/route_mgr.go
mazdakn: Fixed

mazdakn: Added a separate if clause for the borrowed case.

PR #12133, opened 2026-03-16
Fix OpenShift agent override breaking pipeline compilation

lwr20:
## Summary

- Commit a419047 ("Use f1-standard-2 for OpenShift jobs") placed `agent` overrides at the **job level** in `upgrade.yml` and `patch-verification.yml`, which Semaphore's schema does not allow — only **task-level** `agent` is valid
- Both pipelines fail immediately at compilation with `"Schema does not allow additional properties"` at `#/blocks/N/task/jobs/M/agent`, meaning **zero jobs run**
- Fix: extract each OpenShift job into its own block so `agent` sits at the task level, matching the pattern already used correctly in `iptables.yml`

## Test plan

- [ ] Re-run the upgrade scheduler task and confirm the pipeline compiles and jobs start
- [ ] Re-run the patch-verification scheduler task and confirm the pipeline compiles and jobs start
- [ ] Verify the OpenShift jobs use f1-standard-2 machines (visible in Semaphore job details)

### DIFF
--- a/.semaphore/end-to-end/pipelines/upgrade.yml
+++ b/.semaphore/end-to-end/pipelines/upgrade.yml
@@ -158,7 +158,7 @@ blocks:
           type: f1-standard-2
           os_image: ubuntu2204
       jobs:
-        - name: Openshift - 4.16
+        - name: Openshift
           execution_time_limit:
             hours: 4
           commands:


lwr20:
Addressed Copilot feedback — renamed "Openshift - 4.16" to just "Openshift" in upgrade.yml since the version comes from the matrix (`OPENSHIFT_VERSION=4.19.20`).

PR #12144, opened 2026-03-16
Fix flannel/canal migration script path in pipeline configs

lwr20:
## Summary
- The `body_flannel-migration.sh` script was added to `.semaphore/end-to-end/scripts/` in #12112, but the pipeline YAML files still reference the old banzai-calico path (`./scripts/body_flannel-migration.sh`)
- This causes both the Flannel migration and Canal migration jobs to fail immediately with "No such file or directory"
- Update the path to match `body_standard.sh`: `~/calico/.semaphore/end-to-end/scripts/body_flannel-migration.sh`
- Affects `patch-verification.yml` (2 jobs) and `iptables.yml` (1 job)

## Example failing jobs
- [Flannel migration (patch-verification)](https://tigera-delivery.semaphoreci.com/jobs/16e2b820-4d2a-43c9-a046-018999f55c07)
- [Canal migration (patch-verification)](https://tigera-delivery.semaphoreci.com/jobs/42f171fa-f0e0-4e14-a292-f56345a0dda7)

## Test plan
- [ ] Run patch-verification pipeline — Flannel migration and Canal migration jobs should now find the script and proceed to provision

### DIFF
--- a/.semaphore/end-to-end/pipelines/iptables.yml
+++ b/.semaphore/end-to-end/pipelines/iptables.yml
@@ -748,7 +748,7 @@ blocks:
           execution_time_limit:
             hours: 7
           commands:
-            - ./scripts/body_flannel-migration.sh
+            - ~/calico/.semaphore/end-to-end/scripts/body_flannel-migration.sh
           matrix:
             - env_var: DOWNLEVEL_MANIFEST
               values:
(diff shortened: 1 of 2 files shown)


lwr20:
Responding to copilot: 
The fix is consistent with the existing pattern. Changing to ${HOME}/${SEMAPHORE_GIT_DIR} for just the flannel script while everything else uses ~/calico would be inconsistent — and making that change across all scripts would be a separate cleanup.

PR #12146, opened 2026-03-16
[CORE-12502] Add NFTablesSupported feature flag (default Enabled)

nelljerram:
So that we have an escape hatch if nftables is not available at all on someone's distribution, or if their kernel errors when nftables is used for proxy ARP filtering.

Note that this flag is different from nftables _mode_.  The latter configures if we use nftables for policy programming; the former configures if we can use nftables for other purposes.

**Release note:**
```release-note
Add NFTablesSupported feature flag, defaulting to enabled.  This controls Calico's use of nftables for purposes other than policy programming, such as proxy ARP filtering.  In case your distribution or kernel does not support nftables, you can disable this by setting `NFTablesSupported=false` in the `FeatureDetectOverride` field in `FelixConfiguration`.
```


caseydavenport:
@nelljerram perhaps you could expand on why we would want nftables for one purpose but not the other? I'm a bit confused by this one (and expect the similarly named config options are going to be really confusing for users - we already get a lot of head scratching with IptablesBackend vs NftablesMode!)

nelljerram:
> @nelljerram perhaps you could expand on why we would want nftables for one purpose but not the other? I'm a bit confused by this one (and expect the similarly named config options are going to be really confusing for users - we already get a lot of head scratching with IptablesBackend vs NftablesMode!)

In #12038 I have used nftables with the "arp" family to suppress Calico sending a proxy ARP response to a migrating VM for its own IP.  This use of nftables is independent of whether we use nftables/iptables/eBPF for policy programming.  My understanding is that there are still scaling concerns with nftables for policy in some setups, so we didn't want to say "if you want to do that, you must also use nftables for policy"; hence the independence point.  But then @fasaxc said: what if someone is using a distro or kernel where nftables use isn't available or causes a problem - we should have a way to disable the ARP suppression use if necessary.  And that more or less works for the function concerned here - live migration proxy ARP suppression - because it's only needed in quite specific setups, and then we're no worse off than we were before all the live migration work.  Hence I've added this new setting to allow us to do that.

Hopefully that explains the situation and the motivation.  I'm certainly happy to change the naming and commenting to make everything clearer.

caseydavenport:
@nelljerram thanks - that makes more sense to me now!

I might still just have an `ARPSuppression: Enalbed | Disabled` toggle then? Or even better maybe just plug this into the existing feature-detection / overrides capabilities Felix already has? 

IIUC, this sounds like a pretty niche case - we expect this to work regardless of dataplane in most cases.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1115,15 +1115,21 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
-		LookPathOverride: config.LookPathOverride,
-		OnStillAlive:     dp.reportHealth,
-		OpRecorder:       dp.loopSummarizer,
-		NewDataplane:     config.NewNftablesDataplane,
+	// If the NFTablesProxyARPFilter feature is enabled, create nftables ARP table for proxy ARP
+	// suppression.  Note that that feature is different from nftables _mode_.  The latter
+	// configures if we use nftables for policy programming; the former configures if we can use
+	// nftables for proxy ARP filtering.
+	var arpRootTable *nftables.NftablesTable
+	if featureDetector.GetFeatures().NFTablesProxyARPFilter {
```
fasaxc: This doesn't read right to me.   featureDetector.Features is meant to tell you about features of the environment, not features of Calico.  So it should be something like `featureDetector.GetFeatures().NFTablesSupported`.

PR #12147, opened 2026-03-16
Bump k8s to 1.35.2

MichalFupso:
Bump k8s to 1.35.2

**Release note:**
```release-note
TBD
```


apiserver/pkg/storage/calico/resource.go
```diff
@@ -99,6 +99,11 @@ func (rs *resourceStore) SetKeysFunc(_ storage.KeysFunc) {
 	logrus.Error("STUB: SetKeysFunc() not supported by Calico client.")
 }
 
+func (rs *resourceStore) EnableResourceSizeEstimation(_ storage.KeysFunc) error {
+	logrus.Error("STUB: EnableResourceSizeEstimation() not supported by Calico client.")
```
stevegaossou, line 103: Question: 
- When would this not be true (i.e. when would we need to have a working implementation of this function)? Might be worth calling out that we need to revisit this in the future.

stevegaossou, line 103: I think Error log level is a little too harsh, no? 

What is the likelihood, that this gets called repeatedly for some scenario, and that leads to spam Error level logs? 

I'm thinking about any users that have monitoring on error logs in their clusters. We want to avoid producing unwanted errors for them. 

How about Warn or Info level for this? Is there an existing pattern of us doing Error level for these kinds of stub (not implemented) functions?

MichalFupso: We have an existing patter that logs with Error if the function is not defined.

MichalFupso: I don't think we will need to implement, we don't have a functionality for autoscaling and resource scheduling

PR #12148, opened 2026-03-16
refactor(release): remove operator-init image from release tooling

radTuti:
For hashreleases, we were previously building an `operator-init` image. Since delegating to using the tooling in operator (), this is no longer being done as operator tooling does not build the `operator-init` image. Additionally, the `operator-init` image is not used anywhere and is actually just `operator` retagged as `operator-init`

**Release note:**
```release-note
TBD
```


marvin-tigera:

PR #12155, opened 2026-03-17
Swap IPAM owner attributes on live migration completion

skoryk-oleksandr:
This PR ensures correct IPAM ownership after migration completes.  IPAM owner attributes are supposed to be updated, to point to the VM on the target node, once the VM becomes active on that node; but this was accidentally forgotten in #12038 .


felix/dataplane/linux/int_dataplane.go
```diff
@@ -286,6 +287,11 @@ type Config struct {
 
 	LiveMigrationRouteConvergenceTime time.Duration
 
+	// IPAMClient is the Calico IPAM client used to swap owner attributes
+	// when a KubeVirt live migration completes.  May be nil if live migration
```
nelljerram: I don't see anything in the code that corresponds to this "May be nil" comment.  Can you explain further what is meant here?

felix/daemon/daemon.go
```diff
@@ -455,6 +456,12 @@ configRetry:
 		log.Panic("Graceful shutdown took too long")
 	}
 
+	// Get the IPAM client for live migration owner attribute swaps.
+	var ipamClient ipam.Interface
+	if v3Client != nil {
```
nelljerram: v3Client *must* exist here, so can remove this condition.

felix/dataplane/linux/live_migration.go
```diff
@@ -61,9 +67,14 @@ type liveMigrationMonitor struct {
 	migrationUIDs   map[types.WorkloadEndpointID]string
 	newGARPHandle   func(string) (garpHandle, error)
 	convergenceTime time.Duration
+
+	// ipamClient is the Calico IPAM client used to swap owner attributes
+	// when live migration completes.  May be nil if not available.
```
nelljerram: I don't think the "May be nil" case is possible.

felix/dataplane/linux/live_migration.go
```diff
@@ -415,6 +430,61 @@ func isGARPOrRARP(packet gopacket.Packet) bool {
...
 }
 
+// startIPAMOwnerSwap begins an asynchronous IPAM owner attribute swap when a
+// live migration completes (Target→Live transition).  It looks up the VMI name
+// from the target pod's labels via the Kubernetes API, then calls
+// vmipam.EnsureActiveVMOwnerAttrs to atomically promote the alternate (target)
+// owner attributes to active.  This runs in a goroutine to avoid blocking the
+// dataplane main loop.
+func (f *liveMigrationFSM) startIPAMOwnerSwap() {
+	if f.monitor.ipamClient == nil || f.monitor.kubeClientSet == nil {
```
nelljerram: Only need to check for kubeClientSet being nil.

felix/dataplane/linux/live_migration.go
```diff
@@ -415,6 +430,61 @@ func isGARPOrRARP(packet gopacket.Packet) bool {
...
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+
+		// Look up the VMI name from the target pod's labels.
+		pod, err := f.monitor.kubeClientSet.CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
+		if err != nil {
+			logCtx.WithError(err).Warn("Failed to look up pod for IPAM owner swap")
+			return
+		}
```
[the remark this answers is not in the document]
nelljerram: Agreed.  The Labels are available on model.WorkloadEndpoint and I think we should create a live_migration_vmi_name field in proto.WorkloadEndpoint and populate that in ModelWorkloadEndpointToProto.

felix/dataplane/linux/live_migration.go
```diff
@@ -415,6 +430,61 @@ func isGARPOrRARP(packet gopacket.Packet) bool {
...
+		vmiName := pod.Labels["vmi.kubevirt.io/id"]
+		if vmiName == "" {
+			logCtx.Debug("Pod has no vmi.kubevirt.io/id label, skipping IPAM owner swap")
+			return
+		}
+
+		logCtx = logCtx.WithField("vmiName", vmiName)
+		logCtx.Info("Starting IPAM owner attribute swap for live migration")
+
+		err = vmipam.EnsureActiveVMOwnerAttrs(ctx, f.monitor.ipamClient, "", namespace, vmiName, podName)
```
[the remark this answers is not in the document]
nelljerram: Unfortunately I don't see any evidence for the "cni.projectcalico.org/network" annotation being populated, and in any case we'd prefer not to need a Pod read here.  I think empty string is fine for now.  We can improve this later if needed by plumbing the correct network name through Felix.

felix/dataplane/linux/live_migration.go
```diff
@@ -415,6 +430,61 @@ func isGARPOrRARP(packet gopacket.Packet) bool {
...
+		if err != nil {
+			logCtx.WithError(err).Warn("Failed to swap IPAM owner attributes")
+			return
+		}
+		logCtx.Info("Successfully swapped IPAM owner attributes")
```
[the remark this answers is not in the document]
nelljerram: Agreed.  The commenting says that we should retry if the error is not one of (nil, ErrAlternateOwnerEmpty, ErrAlternateOwnerMismatch).

felix/dataplane/linux/live_migration.go
```diff
@@ -415,6 +430,61 @@ func isGARPOrRARP(packet gopacket.Packet) bool {
...
+	}()
```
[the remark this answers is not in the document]
nelljerram: Yes please.  (Claude should be able to do this for you, if you like that approach.)  Also means that the test code does not need to pass nil for these client objects, which will be more realistic.

felix/dataplane/linux/live_migration_test.go
```diff
@@ -50,7 +50,7 @@ func (f *fakeLiveMigrationListener) drain() []liveMigrationStateUpdate {
 // newTestMonitor creates a liveMigrationMonitor with a no-op GARP handle factory
 // and a fake listener, preventing tests from opening real AF_PACKET sockets.
 func newTestMonitor(convergenceTime time.Duration) *liveMigrationMonitor {
-	m := newLiveMigrationMonitor(convergenceTime)
+	m := newLiveMigrationMonitor(convergenceTime, nil, nil)
```
nelljerram: Should prepare fake clients and pass those here, instead of nil, and then add a test that uses them.

nelljerram (approved):

PR #12164, opened 2026-03-17
Defer BPF program attachment until after IP sets are updated

fasaxc:
## Description

This change separates BPF program attachment from the `CompleteDeferredWork()` method by introducing a new `ApplyBPFPrograms()` method that is called explicitly after IP sets have been written to the dataplane.

### Why This Change

When BPF programs are attached to interfaces before IP sets are updated, policy rules that reference IP sets can produce spurious denies for newly-added workloads whose IPs are freshly added to IP sets. By deferring BPF program attachment until after IP set updates complete, we ensure that:

1. IP sets are fully populated before BPF programs start enforcing policies
2. Newly-added workloads don't experience transient policy denies due to missing IP set entries
3. The ordering of dataplane updates is more predictable and correct

### Changes Made

**bpf_ep_mgr.go:**
- Moved BPF program attachment logic (`applyProgramsToDirtyDataInterfaces()`, `updateWEPsInDataplane()`, `removeDirtyPolicies()`) from `CompleteDeferredWork()` to a new `ApplyBPFPrograms()` method
- Moved conntrack map delta copy logic to the end of `CompleteDeferredWork()` to ensure it happens before program attachment
- `CompleteDeferredWork()` now focuses on preparing state without attaching programs

**int_dataplane.go:**
- Added `bpfEndpointManager` field to `InternalDataplane` to enable explicit control over when BPF programs are applied
- Updated `apply()` method to call `ApplyBPFPrograms()` after IP sets are updated but before iptables rules are applied
- Added appropriate error handling and health reporting for the new call

**bpf_ep_mgr_test.go:**
- Updated all test cases to call `ApplyBPFPrograms()` after `CompleteDeferredWork()` to maintain test correctness
- Added 60+ test invocations to ensure the new method is properly exercised

### Testing

All existing unit tests have been updated to call the new `ApplyBPFPrograms()` method in the appropriate sequence. The test changes ensure that the separation of concerns is properly validated across all BPF endpoint manager test scenarios.

**Release note:**
```release-note
Fixed a race condition where BPF programs could enforce policies before IP sets were fully updated, causing spurious policy denies for newly-added workloads.
```

https://claude.ai/code/session_014RZxxt6Tf6YQY7K2T6CbjM

### DIFF
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -1735,11 +1735,11 @@ func (m *bpfEndpointManager) CompleteDeferredWork() error {
 	return nil
 }
 
-// Apply attaches BPF programs to interfaces and updates related dataplane state.
-// It is called from the main apply loop after IP sets have been written to the
-// dataplane, ensuring that policy rules referencing IP sets don't produce spurious
+// ApplyBPFPrograms attaches BPF programs to interfaces and updates related dataplane
+// state.  It is called from the main apply loop after IP sets have been written to
+// the dataplane, ensuring that policy rules referencing IP sets don't produce spurious
 // denies for newly-added workloads whose IPs are freshly added to IP sets.
-func (m *bpfEndpointManager) Apply() error {
+func (m *bpfEndpointManager) ApplyBPFPrograms() error {
 	m.applyProgramsToDirtyDataInterfaces()
 	m.updateWEPsInDataplane()
 	if m.bpfPolicyDebugEnabled {
--- a/felix/dataplane/linux/bpf_ep_mgr.go
+++ b/felix/dataplane/linux/bpf_ep_mgr.go
@@ -1717,21 +1717,6 @@ func (m *bpfEndpointManager) CompleteDeferredWork() error {
 		}
 	}
 
-	// Copy data from old map to the new map
-	m.copyDeltaOnce.Do(func() {
-		logrus.Info("Copy delta entries from old map to the new map")
-		var err error
-		if m.v6 != nil {
-			err = m.v6.CtMap.CopyDeltaFromOldMap()
-		}
-		if m.v4 != nil {
-			err = m.v4.CtMap.CopyDeltaFromOldMap()
-		}
-		if err != nil {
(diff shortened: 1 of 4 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1742,6 +1717,37 @@ func (m *bpfEndpointManager) CompleteDeferredWork() error {
 		}
 	}
 
+	// Copy data from old map to the new map
+	m.copyDeltaOnce.Do(func() {
```
fasaxc: I don't think this should be re-ordered; I think it needs to happen after the programs are attached to prevent races.  Move it to the apply function to maintain the ordering

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1723,14 +1705,7 @@ func (m *bpfEndpointManager) CompleteDeferredWork() error {
 		logrus.Info("BPF counters synced.")
 	})
 
-	m.applyProgramsToDirtyDataInterfaces()
-	m.updateWEPsInDataplane()
-	if m.bpfPolicyDebugEnabled {
-		m.removeDirtyPolicies()
-	}
-
 	bpfEndpointsGauge.Set(float64(len(m.nameToIface)))
```
fasaxc: @claude this gauge should be updated after applying programs too; it's meant to mean "count of programs that we successfully applied"

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1742,6 +1715,41 @@ func (m *bpfEndpointManager) CompleteDeferredWork() error {
...
+	if m.bpfPolicyDebugEnabled {
+		m.removeDirtyPolicies()
+	}
+
+	bpfEndpointsGauge.Set(float64(len(m.nameToIface)))
+
+	// Copy data from old conntrack map to the new map.  Must happen after
+	// programs are applied above so that old programs are no longer writing
+	// to the old map.
+	m.copyDeltaOnce.Do(func() {
```
fasaxc, line 1737: @claude let's not re-order this relative to the code below unless there's a reason to do so.

PR #12167, opened 2026-03-17
[BPF] Add native netkit BPF attachment support to Felix

tomastigera:
## Summary

Native netkit BPF attachment support for Felix. When Felix detects a workload endpoint using a netkit device (instead of veth), it attaches BPF programs via the netkit attach API rather than TC/TCX. Full BPF FV suite passes with netkit on Ubuntu 25.10.

### Core netkit support

- **Netkit detection and attachment**: Felix auto-detects netkit workload interfaces and uses `BPF_NETKIT_PRIMARY`/`BPF_NETKIT_PEER` attachment instead of TC/TCX
- **Separate prog_array maps**: Netkit programs have a different `expected_attach_type` and cannot share `prog_array` maps with TC/TCX programs. New maps with pin path overrides
- **host_ifindex in globals**: Netkit peer programs see the peer ifindex in `skb->ifindex`; added `host_ifindex` to globals for correct FIB, QoS, counter, and RPF map lookups
- **No bpf_redirect_peer for netkit**: Netkit programs run in xmit context where `bpf_redirect_peer` silently drops; uses plain `bpf_redirect` via FIB instead
- **Non-cgo stub**: `AttachNetkit` stub in `libbpf_stub.go` for architectures without BPF support

### BPF dataplane fixes for netkit

- **Workload RPF check**: Use `host_ifindex` from globals so the route ifindex comparison works correctly
- **QoS packet rate**: Use `host_ifindex` for QoS map key lookup
- **Netkit peer MTU**: Set MTU on the peer (container-side) LinkAttrs so it matches the host side

### Felix jump map cleanup for netkit

- **jumpMapDelete / reclaimPolicyIdx / reclaimFilterIdx / removePolicyProgram / ensureNoProgram**: Select correct jump map (netkit vs TC) based on interface type
- **Preserve ifaceType on interface down**: When `getIfaceLink` fails during down events, preserve the previously-known type for correct cleanup
- **syncIfStateMap**: Clean both TC and netkit jump maps for gone interfaces

### FV test fixes

- **Backend replacement test**: Use workload-side tcpdump (inside pod namespace) instead of host-side
- **Program recovery test**: Use `bpfProgPinDir()` helper to select `NetkitPinDir` vs `TcxPinDir`
- **Policy scale test**: Query correct netkit jump maps in `BPFNumPolProgramsByEntryPoint`
- **UT fix**: Initialize `NetkitJumpMaps` in `bpf_ep_mgr_test.go` common maps setup

### CI

- Full BPF FV run with netkit on Ubuntu 25.10 (replaces UT-only 25.10 check)
- FV test infrastructure: `--netkit` flag for test-workload, `FELIX_FV_NETKIT=Enabled` env var

```release-note
ebpf: Add native netkit BPF attachment support. Felix auto-detects netkit workload interfaces and attaches BPF programs via the netkit API instead of TC/TCX, with correct handling of ifindex differences, separate prog_array maps, and jump map cleanup.
```

## Test plan

- [x] Full BPF FV suite with netkit on Ubuntu 25.10 CI runner -- all green
- [x] BPF FV suite with TCX (default) -- no regressions
- [x] bpf_ep_mgr unit tests pass

### DIFF
--- a/felix/bpf/hook/map.go
+++ b/felix/bpf/hook/map.go
@@ -179,11 +179,19 @@ func MergeLayouts(layouts ...Layout) Layout {
 	return ret
 }
 
+// programCacheKey combines AttachType with the BPF program attach type string
+// ("TC", "TCX", or "Netkit") to differentiate cache entries for programs that
+// share the same object file but require different expected_attach_type values.
+type programCacheKey struct {
+	AttachType
+	ProgAttachType string
+}
+
 type ProgramsMap struct {
 	*bpfmaps.PinnedMap
 
 	programsLock sync.Mutex
-	programs     map[AttachType]*program
+	programs     map[programCacheKey]*program
 
 	expectedAttachType string
 	nextIdx            atomic.Int64
@@ -325,7 +333,7 @@ func NewEgressProgramsMap() bpfmaps.Map {
 func newProgramsMap(ProgramsMapParameters bpfmaps.MapParameters, expectedAttachType string) bpfmaps.Map {
 	return &ProgramsMap{
 		PinnedMap:          bpfmaps.NewPinnedMap(ProgramsMapParameters),
-		programs:           make(map[AttachType]*program),
+		programs:           make(map[programCacheKey]*program),
 		expectedAttachType: expectedAttachType,
 	}
 }
@@ -340,7 +348,7 @@ func NewXDPProgramsMap() bpfmaps.Map {
 			Name:       "xdp_cali_progs",
 			Version:    3,
 		}),
-		programs: make(map[AttachType]*program),
+		programs: make(map[programCacheKey]*program),
 	}
 }
 
@@ -351,7 +359,8 @@ func (pm *ProgramsMap) LoadObj(at AttachType, progType string, disabledOptional 
 	}
 	log.WithField("AttachType", at).Debugf("Looked up file for attach type: %s", file)
 
-	pi := pm.getOrCreateProgramInfo(at)
+	cacheKey := programCacheKey{AttachType: at, ProgAttachType: progType}
+	pi := pm.getOrCreateProgramInfo(cacheKey)
 
 	// Loading is protected by the program lock to ensure that we do not
 	// load the same object multiple times in parallel.  Two goroutines may
@@ -388,13 +397,13 @@ func (pm *ProgramsMap) LoadObj(at AttachType, progType string, disabledOptional 
 	}, nil
 }
 
-func (pm *ProgramsMap) getOrCreateProgramInfo(at AttachType) *program {
+func (pm *ProgramsMap) getOrCreateProgramInfo(key programCacheKey) *program {
 	pm.programsLock.Lock()
 	defer pm.programsLock.Unlock()
-	pi, ok := pm.programs[at]
+	pi, ok := pm.programs[key]
 	if !ok {
 		pi = &program{}
-		pm.programs[at] = pi
+		pm.programs[key] = pi
 	}
 	return pi
 }
@@ -618,7 +627,7 @@ func (pm *ProgramsMap) ResetForTesting() {
 	// We keep the same pinned map but reset the accounting as the map is
 	// replaced by repinning by the user.
 	pm.nextIdx.Store(0)
-	pm.programs = make(map[AttachType]*program)
+	pm.programs = make(map[programCacheKey]*program)
 }
 
 func (pm *ProgramsMap) Programs() map[AttachType]Layout {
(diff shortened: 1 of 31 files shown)


felix/dataplane/linux/bpf_ep_mgr_test.go
```diff
@@ -460,6 +460,8 @@ var _ = Describe("BPF Endpoint Manager", func() {
 		jumpMapEgr = mock.NewMockMap(progsParamsEg)
 		commonMaps.JumpMaps = append(commonMaps.JumpMaps, jumpMapIng)
 		commonMaps.JumpMaps = append(commonMaps.JumpMaps, jumpMapEgr)
+		commonMaps.NetkitJumpMaps = append(commonMaps.NetkitJumpMaps, mock.NewMockMap(progsParamsIng))
+		commonMaps.NetkitJumpMaps = append(commonMaps.NetkitJumpMaps, mock.NewMockMap(progsParamsEg))
```
tomastigera, line 471: do we actually run this test with netkit devices?

felix/bpf/hook/map.go
```diff
@@ -437,6 +545,19 @@ func (pm *ProgramsMap) configureMapsAndPrograms(obj *libbpf.Obj, file, progAttac
 				return fmt.Errorf("error setting attach type for program %s: %w", prog.Name(), err)
 			}
 		}
+	case "Netkit":
+		for prog, err := obj.FirstProgram(); prog != nil && err == nil; prog, err = prog.NextProgram() {
```
sridhartigera: The loop is repetitive with TCX case.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1676,59 +1729,118 @@ func (m *bpfEndpointManager) syncIfaceProperties() error {
...
 			if hk == hook.Ingress {
 				attachType = libbpf.AttachTypeTcxIngress
 			}
 			if err := obj.SetAttachType(p.Name(), attachType); err != nil {
-				return fmt.Errorf("error setting attach type for program %s: %w", p.Name(), err)
+				return 0, 0, fmt.Errorf("error setting attach type for program %s: %w", p.Name(), err)
+			}
+		}
+	case tc.AttachOptionNetkit:
+		for p, err := obj.FirstProgram(); p != nil && err == nil; p, err = p.NextProgram() {
```
sridhartigera, line 1805: This loop is repetitive with TCX

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -3115,7 +3260,7 @@ func (m *bpfEndpointManager) getEndpointType(ifaceName string) tcdefs.EndpointTy
 	ifaceType := m.nameToIface[ifaceName].info.ifaceType
 	m.ifacesLock.Unlock()
 	switch ifaceType {
-	case IfaceTypeData, IfaceTypeVXLAN, IfaceTypeBond, IfaceTypeBondSlave:
+	case IfaceTypeData, IfaceTypeVXLAN, IfaceTypeBond, IfaceTypeBondSlave, IfaceTypeNetkit:
```
sridhartigera: I think `IfaceTypeNetkit` is not needed here as this is only for host ifaces.

felix/bpf/tc/attach.go
```diff
@@ -206,6 +251,62 @@ func (ap *AttachPoint) ProgPinPath() string {
 	return path.Join(bpfdefs.TcxPinDir, fmt.Sprintf("%s_%s", strings.ReplaceAll(ap.Iface, ".", ""), ap.Hook))
 }
 
+func (ap *AttachPoint) NetkitProgPinPath() string {
+	return path.Join(bpfdefs.NetkitPinDir, fmt.Sprintf("%s_%s", strings.ReplaceAll(ap.Iface, ".", ""), ap.Hook))
+}
+
+func (ap *AttachPoint) attachNetkitProgram(binaryToLoad string) error {
```
sridhartigera, line 279: `attachTCXProgram` and `attachNetkitProgram` are identical except for the attach type. Can we make it a single function?

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: Tried this and the BPF FVs panicked with `Unsupported ifaceName eth20` — in FV netkit mode the test harness creates host-side data interfaces (e.g. eth20 in multi-NIC RPF / external-client topologies) as netkit devices, so they reach this switch with `ifaceType == IfaceTypeNetkit` but `isWorkloadIface == false`. So `IfaceTypeNetkit` does need to stay here, just for that FV-only path. Reverted.

sridhartigera (approved):
LGTM. Thanks

PR #12172, opened 2026-03-18
Investigate why hashrelease build is dirty

nelljerram:
**Release note:**
```release-note
TBD
```


nelljerram:
No longer needed, and probably outdated now.

PR #12176, opened 2026-03-18
Delete accidentally committed hack/test/kind/infra/operator directory

nelljerram:
This directory is checked in but was being deleted by the clean target in node/Makefile:

    rm -rf $(REPO_ROOT)/hack/test/kind/infra/operator

This might be the cause of the hashrelease build currently failing with

    FATAL[0067]main.go:85 main.main() Error running app                             error="there are uncommitted changes in the repository, please commit or stash them before publishing the release"

I've reproduced a failing build with "sem debug" and then "git status" shows:

    semaphore@semaphore-vm:~/calico/release$ git status
    On branch master
    Your branch is up to date with 'origin/master'.

    Changes not staged for commit:
      (use "git add/rm <file>..." to update what will be committed)
      (use "git restore <file>..." to discard changes in working directory)
    	deleted:    ../hack/test/kind/infra/operator

    no changes added to commit (use "git add" and/or "git commit -a")

So it looks like the immediate problem in the hashrelease build is hack/test/kind/infra/operator having been deleted.  I don't know for sure if that is because of a make clean in node, but it seems like a good guess.

It looks like the hack/test/kind/infra/operator directory was an accidental inclusion in https://github.com/projectcalico/calico/pull/12126, so let's just delete it again.


nelljerram:
Added `skip...` label because a similar accidental commit did not happen in calico-private, either on master or on release-calient-v3.23.

PR #12177, opened 2026-03-18
Exclude istio/ and third_party/ from gen-deps-files

radixo:
These directories do not contain project source code — they download, patch, and build vendored code at specific versions. Generating deps.txt for them is incorrect since their Go files are transient build artifacts, not tracked dependencies.

If some run `make gen-deps-files` after a `make -C istio build` it will result creating an `istio/deps.txt` file.

Remove the erroneously generated istio/deps.txt and add both directories to GO_DIRS_EXCLUDE. Also refactor the exclude list to use Make's addprefix for cleaner, more maintainable grep flags.

**Release note:**
```release-note-not-required
TBD
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,8 @@ gen-semaphore-yaml:
 	                          RELEASE_BRANCH_PREFIX=$(RELEASE_BRANCH_PREFIX) \
 	                          go run ./hack/cmd/deps $(DEPS_ARGS) generate-semaphore-yamls"
 
-GO_DIRS=$(shell find -name '*.go' | grep -v -e './lib/' -e './pkg/' | grep -o --perl '^./\K[^/]+' | sort -u)
+GO_DIRS_EXCLUDE=./lib/ ./pkg/ ./istio/ ./third_party/
+GO_DIRS=$(shell find -name '*.go' | grep -v $(addprefix -e ,$(GO_DIRS_EXCLUDE)) | grep -o --perl '^./\K[^/]+' | sort -u)
 DEP_FILES=$(patsubst %, %/deps.txt, $(GO_DIRS))
 
 gen-deps-files:
--- a/istio/deps.txt
+++ b/istio/deps.txt
@@ -1,5 +1,0 @@
-!!! GENERATED FILE, DO NOT EDIT !!!
-This file contains the list of modules that this package depends on
-in order to trigger CI on changes
-
-go 1.25.7


nelljerram:
@radixo The PR description sounds sensible, but can you also explain why we need to do this?  I.e. is something currently broken in practice, and this PR will fix that?

Makefile
```diff
@@ -121,7 +121,8 @@ gen-semaphore-yaml:
 	                          RELEASE_BRANCH_PREFIX=$(RELEASE_BRANCH_PREFIX) \
 	                          go run ./hack/cmd/deps $(DEPS_ARGS) generate-semaphore-yamls"
 
-GO_DIRS=$(shell find -name '*.go' | grep -v -e './lib/' -e './pkg/' | grep -o --perl '^./\K[^/]+' | sort -u)
+GO_DIRS_EXCLUDE=./lib/ ./pkg/ ./istio/ ./third_party/
```
fasaxc, line 124: Why the lib exclude?  That directory contains libraries that are part of calico and should have dep tracking.  ./pkg too, isn't that packaging scripts; which might/might not use go.

Shouldn't istio be inside third_party?  That's what that directory is for: thrird-party builds that have their own rules.

radixo: The `./lib/` and `./pkg/` were already excluded before, I just appended `./istio` and `./third_party`.

I'm not so sure the reason istio was created on root, and only third_party/istio-ztunnel for ztunnel component, I believe @electricjesus and @hjiawei can explain us why.

radixo:
> @radixo The PR description sounds sensible, but can you also explain why we need to do this? I.e. is something currently broken in practice, and this PR will fix that?

I believe the second paragraph I just added to the PR description brings more light to the problem this PR sorts.

Makefile
radixo: Hey Shaun, the `./istio` is on the root because it is patched with our customizations. As other projects like `fluent-bit`.
But what makes me think why `istio-ztunnel` is on`third-party` dir, once it has customizations as well.
I don't know about you guys, but looking at those changes using patches, I feel like if all of them were inside third_party would be more well addressed.

Maybe a task to sort this?

WDYT @hjiawei , @fasaxc ?

fasaxc (approved):
Sorry for the noise, step int eh right direction, but yes, I think istio should go in third_party

PR #12180, opened 2026-03-18
Fix tunnel IPs getting REMOTE_WORKLOAD from containing IPAM block

caseydavenport:
When a tunnel IP (VXLAN/IPIP) falls within an IPAM block -- which is the normal case, since tunnel IPs are allocated from the same pool as workloads -- the `l3_route_resolver` was unconditionally OR'ing `REMOTE_WORKLOAD` into the route type from the parent block entry. Combined with `REMOTE_TUNNEL` from the tunnel ref, this produced route type 17 (`REMOTE_TUNNEL | REMOTE_WORKLOAD`), which caused `isRemoteTunnelRoute()` in `route_mgr.go` to return true. The route manager then programmed a spurious /32 directly-connected route on the tunnel device:

```
192.168.110.128/26 via 172.18.0.2 dev vxlan.calico onlink mtu 1450   # block route (correct)
192.168.110.128 dev vxlan.calico scope link mtu 1450                  # /32 route (spurious)
```

The trie walk in `flush()` iterates from root to leaf, accumulating type flags from pools, blocks, and refs. The fix defers the block-derived workload type flags (`LOCAL_WORKLOAD`/`REMOTE_WORKLOAD`) and only applies them after the full walk if the route is not a tunnel ref. Workload refs (including borrowed IPs) still get the block types as before -- this is important because borrowed workloads legitimately need both `REMOTE_WORKLOAD` (from the block's per-allocation entry on a different node) and `LOCAL_WORKLOAD` (from the local WEP ref).

The bug was order-dependent in the trie: it only manifested when the IPAM block existed before the tunnel IP ref was added, because `UpdateBlockRoute` doesn't call `markChildrenDirty`. The existing calc graph FV tests happened to process KVs in an order that avoided it. A new unit test (included) reproduces the issue by setting up the block first, then adding the node with the tunnel IP.

Builds on the failing test from #12179. Related to @mazdakn's work in #12122.

### DIFF
--- a/felix/calc/l3_route_resolver.go
+++ b/felix/calc/l3_route_resolver.go
@@ -747,6 +747,8 @@ func (c *L3RouteResolver) flush() {
 		poolAllowsCrossSubnet := false
 		var blockSeen bool
 		var blockNodeName string
+		var blockTypes proto.RouteType
+		var hasTunnelRef bool
 		for _, entry := range buf {
 			ri := entry.Data.(RouteInfo)
 			if len(ri.Pools) > 0 {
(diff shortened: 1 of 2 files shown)


mazdakn (approved):
LGTM

PR #12186, opened 2026-03-18
Add postrelease test for operator images list [master]

danudey:
This adds a test to the postrelease test suite which does the following:

1. Pulls the Operator image for this release and runs it with `--print-images=list`
2. Finds any `quay.io/calico` images
3. Verifies that those images are being checked for in our postrelease tests

For example, if we add a new image, `enfrobulate`, to Operator, then it will be listed as `quay.io/calico/enfrobulate:v3.32.3`. When the release tool triggers these tests it passes a list of images to test for; if `enfrobulate` is not one of those images, this test will fail.

This ensures that when new images are added we don't risk missing them in our postrelease tests, and potentially in other places where release tool may need to know about or interact with images.

### DIFF
--- a/release/pkg/postrelease/operator_images_test.go
+++ b/release/pkg/postrelease/operator_images_test.go
@@ -29,7 +29,7 @@ func TestOperatorPrintedImagesInExpectedList(t *testing.T) {
 
 	// Run the operator image with --print-images=list to get the list of images it uses.
 	t.Logf("Running operator image %s with --print-images=list", fqOperatorImage)
-	runCmd := exec.Command("docker", "run", "--rm", fqOperatorImage, "--print-images=list")
+	runCmd := exec.Command("docker", "run", "--rm", fqOperatorImage, "--print-images=listcalico")
 	out, err := runCmd.CombinedOutput()
 	if err != nil {
 		t.Fatalf("failed to run operator image %s with --print-images=list: %v\n%s", fqOperatorImage, err, string(out))
@@ -51,7 +51,7 @@ func TestOperatorPrintedImagesInExpectedList(t *testing.T) {
 		if line == "" {
 			continue
 		}
-		if !strings.HasPrefix(line, tigeraPrefix) {
+		if strings.Contains(line, operator.DefaultImage) {
 			continue
 		}
 		// Extract the image name from the fully qualified image reference.
--- a/release/pkg/postrelease/operator_images_test.go
+++ b/release/pkg/postrelease/operator_images_test.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/projectcalico/calico/release/internal/command"
 	"github.com/projectcalico/calico/release/internal/registry"
 	"github.com/projectcalico/calico/release/pkg/manager/operator"
 )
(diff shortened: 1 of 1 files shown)


release/pkg/postrelease/operator_images_test.go
```diff
@@ -0,0 +1,81 @@
...
+	// Pull the operator image.
+	t.Logf("Pulling operator image %s", fqOperatorImage)
+	pullCmd := exec.Command("docker", "pull", fqOperatorImage)
+	if out, err := pullCmd.CombinedOutput(); err != nil {
+		t.Fatalf("failed to pull operator image %s: %v\n%s", fqOperatorImage, err, string(out))
+	}
+
+	// Run the operator image with --print-images=list to get the list of images it uses.
+	t.Logf("Running operator image %s with --print-images=list", fqOperatorImage)
+	runCmd := exec.Command("docker", "run", "--rm", fqOperatorImage, "--print-images=list")
```
radTuti:
```suggestion
	runCmd := exec.Command("docker", "run", "--rm", fqOperatorImage, "--print-images=listcalico")
```
this gets just the calico images

release/pkg/postrelease/operator_images_test.go
```diff
@@ -0,0 +1,81 @@
...
+	tigeraPrefix := registry.DefaultCalicoRegistry + "/"
+	var missing []string
+	for line := range strings.SplitSeq(strings.TrimSpace(string(out)), "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
+		if !strings.HasPrefix(line, tigeraPrefix) {
+			continue
+		}
```
radTuti: probably have this first to not process operator image any further.
```suggestion
		if strings.Contains(line, operator.DefaultImage) {
			continue
		}
```
also won't need to eliminate tigera images

release/pkg/postrelease/operator_images_test.go
```diff
@@ -0,0 +1,81 @@
...
+		// Extract the image name from the fully qualified image reference.
+		// e.g. "quay.io/calico/node:v3.31.1" -> "node"
+		withoutRegistry := strings.TrimPrefix(line, tigeraPrefix)
+		imageName, _, _ := strings.Cut(withoutRegistry, ":")
+		if imageName == "" {
+			continue
+		}
+
+		// The operator image itself is not expected in the images list.
+		if strings.Contains(operator.DefaultImage, imageName) {
```
radTuti: this is probably too late in the process as the imageName will be "operator" at this point 
(see suggestion above for placement).

release/pkg/postrelease/operator_images_test.go
```diff
@@ -0,0 +1,81 @@
...
+			missing = append(missing, fmt.Sprintf("%s (from %s)", imageName, line))
+		}
+	}
+
+	if len(missing) > 0 {
+		t.Fatalf("The following quay.io/calico images printed by the operator are not in the expected images list:\n  %s\nExpected images: %s",
+			strings.Join(missing, "\n  "),
+			images)
+	}
+	t.Logf("All quay.io/calico images from operator --print-images=list are in the expected images list")
```
radTuti: nit, do not specify the registry

release/pkg/postrelease/operator_images_test.go
```diff
@@ -0,0 +1,76 @@
...
+	t.Logf("Pulling operator image %s", fqOperatorImage)
+	pullCmd := exec.Command("docker", "pull", fqOperatorImage)
+	if out, err := pullCmd.CombinedOutput(); err != nil {
+		t.Fatalf("failed to pull operator image %s: %v\n%s", fqOperatorImage, err, string(out))
+	}
+
+	// Run the operator image with --print-images=list to get the list of images it uses.
+	t.Logf("Running operator image %s with --print-images=listcalico", fqOperatorImage)
+	runCmd := exec.Command("docker", "run", "--rm", fqOperatorImage, "--print-images=listcalico")
+	out, err := runCmd.CombinedOutput()
```
radTuti: use `command.Run(...)`

release/pkg/postrelease/operator_images_test.go
```diff
@@ -0,0 +1,76 @@
...
+	fqOperatorImage := fmt.Sprintf("%s/%s:%s", operator.DefaultRegistry, operator.DefaultImage, operatorVersion)
+
+	// Pull the operator image.
+	t.Logf("Pulling operator image %s", fqOperatorImage)
+	pullCmd := exec.Command("docker", "pull", fqOperatorImage)
+	if out, err := pullCmd.CombinedOutput(); err != nil {
+		t.Fatalf("failed to pull operator image %s: %v\n%s", fqOperatorImage, err, string(out))
+	}
+
+	// Run the operator image with --print-images=list to get the list of images it uses.
```
radTuti: nit, update comment.
also can it be either have the comment or the log? they are the same thing.

radTuti (approved):

PR #12197, opened 2026-03-19
Replace tabs in ASCII art in the felix Makefile

twz123:
That way, the right formatting is more portable and not depending on a tab width of eight.

**Release note:**
```release-note
TBD
```


twz123:
👋

nelljerram (approved):

PR #12202, opened 2026-03-20
[BPF] Fix TC program detachment from veth host interfaces

tomastigera:
## Summary

- On some kernels (notably GCP), veth `ParentIndex` references the peer interface. This caused Felix's interface tree to incorrectly classify host veths (like eth0) as children of other veths (like eth20), triggering "XDP only" mode which detached TC programs.
- Without TC programs on eth0, VXLAN return traffic was not decapsulated, breaking NodePort forwarding through secondary NICs in FV tests.
- Treat veth and netkit devices as standalone in the interface tree since their `ParentIndex` is a peer reference, not a real hierarchy parent (bond/bridge).

```release-note
Fix that BPF programs could be incorrectly removed from workload interfaces on recent kernels due to change in kernel use of IFLA_LINK netlink message.
```

## Test plan

- [x] Verified the failing BPF FV test passes on GCP VM
- [ ] Verify bond/VLAN interface tree logic still works (existing unit tests)
- [ ] Run full BPF FV suite to check for regressions

### DIFF
--- a/felix/bpf-gpl/Makefile
+++ b/felix/bpf-gpl/Makefile
@@ -188,9 +188,23 @@ test_xdp%.ll: xdp.c xdp.d calculate-flags
 test_xdp%_v6.ll: xdp.c xdp_v6.d calculate-flags
 	$(COMPILE)
 
+tc.d: tc.c
+	$(COMPILE_DEPS)
 tc_v6.d: tc.c
 	$(COMPILE_DEPS)
+xdp.d: xdp.c
+	$(COMPILE_DEPS)
 xdp_v6.d: xdp.c
+	$(COMPILE_DEPS)
+connect_balancer.d: connect_balancer.c
+	$(COMPILE_DEPS)
+connect_balancer_v6.d: connect_balancer_v6.c
+	$(COMPILE_DEPS)
+connect_balancer_v46.d: connect_balancer_v46.c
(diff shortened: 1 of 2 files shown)


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -4927,12 +4927,22 @@ func (trees bpfIfaceTrees) addIface(link netlink.Link) {
 		children:    make(map[int]*bpfIfaceNode),
 	}
 
-	if attrs.MasterIndex == 0 && attrs.ParentIndex == 0 {
-		trees.addIfaceStandAlone(intf)
-	} else if attrs.MasterIndex != 0 {
+	// Veth and netkit devices use ParentIndex to reference their peer, not a
```
fasaxc, line 4930: `ParentIndex` is a misnomer in the netlink Go API.  Sounds like the underlying `IFLA_LINK` field has a history of being used for different things including the peer veth.

fasaxc:
Please make sure the release note says which kernel we found the problem on

tomastigera:
> Please make sure the release note says which kernel we found the problem on

This is purely fv tests issue and the kernels are asorted. The issue is with 6.14 in gcp but not with 6.14 generic, 6.17 gcp but idk which else.

PR #11948, opened 2026-02-27
Build Calico with RHEL 9 and Go 1.26

hjiawei:
This changeset upgrades Calico toolchain base to RHEL 9, Golang to v1.26, and K8s to v1.35 release train.

**Release note:**
```release-note
Calico is now built using the RHEL 9 toolchain with Go 1.26, and is updated to use the Kubernetes 1.35 release series.
```

PR #12212, opened 2026-03-20
Fix minor issues from PR #11948 review comments

hjiawei:
Address review feedback from the RHEL 9 toolchain PR (#11948):

- Fix `/n` typo to `\n` in calicoctl node run error formatting, which caused output lines to not split correctly and print literal `/n`.
- Use `bpftool` (PATH lookup) instead of hardcoded `/usr/sbin/bpftool` in BPF UT, consistent with all other bpftool invocations in the repo.
- Use `grep -E` instead of `grep -P` in mockery config check script for better portability (PCRE not always available).
- Bump CRANE_VERSION to v0.21.3.

**Release note:**
```release-note
TBD
```


calicoctl/calicoctl/commands/node/run.go
```diff
@@ -373,8 +373,8 @@ Description:
 	if output, err := runCmd.CombinedOutput(); err != nil {
 		var errStr strings.Builder
 		fmt.Fprintf(&errStr, "Error executing command: %v\n", err)
-		for line := range strings.SplitSeq(string(output), "/n") {
-			fmt.Fprintf(&errStr, " | %s/n", line)
+		for line := range strings.SplitSeq(string(output), "\n") {
```
hjiawei, line 376: `/n` was originally added in https://github.com/projectcalico/calicoctl/pull/1649. I believe it was a typo from the original author. Docker's `CombinedOutput()` returns normal `\n`-delimited bytes.

PR #12216, opened 2026-03-21
goldmane: fix bugs found in code review

caseydavenport:
Fix four bugs found during a goldmane code review:

- **Nil stream panic on backfill**: when max streams is reached, `register()` returns nil. That nil was sent straight to the backfill channel, which panics when `backfill()` calls methods on it. Added a nil guard before sending.

- **numClients metric not registered**: the `goldmane_collector_num_clients` gauge was declared and Inc/Dec'd on connect/disconnect, but never registered with Prometheus. The metric silently didn't appear in output.

- **ExpiringFlowCache.Run() goroutine leak**: `Run()` was an infinite loop with no shutdown path — no context, no done channel. Added a context parameter and threaded it through from the daemon and flow client.

- **saveState() Create/Update ignore timeout**: the `Get` call in `saveState()` correctly used a 15s timeout context, but `Create` and `Update` used `context.Background()`. If the API server is unreachable, those calls block forever. Now all three use the same timeout context.


goldmane/pkg/stream/stream_manager.go
```diff
@@ -153,7 +153,9 @@ func (m *streamManager) Run(ctx context.Context) {
 		case req := <-m.streamRequests:
 			stream := m.register(req)
 			req.respCh <- stream
-			m.backfillRequests <- stream
+			if stream != nil {
```
skoryk-oleksandr, line 156: nit: since we're fixing up the nil-stream path, the nil that goes through `req.respCh` ends up in `goldmane.go:Stream()` which returns a plain fmt.Errorf("failed to establish new stream"). gRPC wraps that as `codes.Unknown`, which doesn't tell the client anything useful. Might be worth using status.Error(codes.ResourceExhausted, "max concurrent streams reached") there so clients can distinguish capacity limits from actual errors. Not blocking, just a natural follow-up since you're already here.

PR #12218, opened 2026-03-21
confd: add native Go template rendering tests

caseydavenport:
Replaces a large chunk of the bash-based confd test suite with native Go tests.

- Leave test inputs (YAML fixtures) and expected outputs (compiled_templates/) as they are
- Get rid of bash runners — no more Docker, no binary builds
- Use `go test` with standard Go/Gomega assertions
- Run confd in-process via `RunWithContext()` against envtest (KDD) or standalone etcd
- Dual-backend: each test runs against both KDD and etcd automatically
- Daemon-mode support for testing dynamic behaviors (e.g., filter deletion)

Covers: mesh, explicit peering, BGP filters, TTL security, next hop, reverse peering, felix cluster routing, ignored interfaces. Not yet migrated: communities, static routes, route reflectors, password lifecycle.

Run with `make -C confd fv`.

```release-note
NONE
```

### DIFF
--- a/confd/tests/helpers_test.go
+++ b/confd/tests/helpers_test.go
@@ -548,6 +548,49 @@ func applyCalicoNode(t *testing.T, be *datastoreBackend, yamlBytes []byte) func(
 	}
 }
 
+// waitForResources polls the Calico client until all CRD resources from the
+// input YAML are visible via List. Under CI load, the envtest API server can
+// briefly return stale list results after a Create.
+func waitForResources(t *testing.T, cc client.Interface, inputPath string) {
+	t.Helper()
+
+	data, err := os.ReadFile(inputPath)
+	require.NoError(t, err)
+
+	expected := countResourcesByKind(data)
+	ctx := context.Background()
+	listCount := map[string]func() int{
+		"BGPPeer":          func() int { l, _ := cc.BGPPeers().List(ctx, options.ListOptions{}); return len(l.Items) },
+		"BGPConfiguration": func() int { l, _ := cc.BGPConfigurations().List(ctx, options.ListOptions{}); return len(l.Items) },
+		"BGPFilter":        func() int { l, _ := cc.BGPFilter().List(ctx, options.ListOptions{}); return len(l.Items) },
+		"IPPool":           func() int { l, _ := cc.IPPools().List(ctx, options.ListOptions{}); return len(l.Items) },
+	}
+
+	require.Eventually(t, func() bool {
+		for kind, want := range expected {
+			if counter, ok := listCount[kind]; ok && counter() < want {
+				return false
+			}
+		}
+		return true
+	}, 5*time.Second, 50*time.Millisecond, "timed out waiting for resources to be visible via Calico client")
+}
+
+func countResourcesByKind(data []byte) map[string]int {
+	counts := map[string]int{}
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(data), 4096)
+	for {
+		var raw map[string]any
+		if err := decoder.Decode(&raw); err != nil {
+			break
+		}
+		if kind, _ := raw["kind"].(string); kind != "" {
+			counts[kind]++
+		}
+	}
+	return counts
+}
+
 // blockAffinityEntry defines a block affinity for test setup.
 type blockAffinityEntry struct {
 	cidr  string
@@ -691,6 +734,11 @@ func runConfdTest(t *testing.T, be *datastoreBackend, inputYAML, goldenDir strin
 	// Create a fresh Calico client for confd so its syncer gets a clean snapshot.
 	confdCalicoClient, err := client.New(apiconfig.CalicoAPIConfig{Spec: be.datastoreConfig})
 	require.NoError(t, err, "creating confd Calico client")
+
+	// Wait for all resources to be visible through the Calico client. Under
+	// CI load the envtest API server can return stale list results briefly
+	// after a Create returns, causing the syncer to miss resources.
+	waitForResources(t, confdCalicoClient, inputPath)
 
 	confdConfig := &config.Config{
 		ConfDir:  confDir,


confd/tests/helpers_test.go
```diff
@@ -0,0 +1,558 @@
...
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths:        []string{crdDir()},
+		ControlPlaneStopTimeout:  2 * time.Second,
+	}
+
+	var err error
+	restConfig, err = testEnv.Start()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to start envtest: %v\n", err)
+		os.Exit(1)
```
caseydavenport, line 154: Fail instead of exit?

confd/tests/helpers_test.go
```diff
@@ -0,0 +1,558 @@
...
+	}
+
+	os.Setenv("DATASTORE_TYPE", "kubernetes")
```
caseydavenport: Need both etcd and kdd mode

confd/tests/helpers_test.go
```diff
@@ -0,0 +1,558 @@
...
+	if err := testEnv.Stop(); err != nil {
+		fmt.Fprintf(os.Stderr, "failed to stop envtest: %v\n", err)
+	}
+	os.Exit(code)
+}
+
+// writeKubeconfig writes a kubeconfig file for the envtest API server and
+// returns the path. Needed because the Calico client reads KUBECONFIG to
+// configure its backend syncer.
+func writeKubeconfig(cfg *rest.Config) string {
```
caseydavenport, line 300: Can move non-test helpers into a utils_test.go

confd/tests/template_render_test.go
```diff
@@ -0,0 +1,558 @@
...
+		{ObjectMeta: objectMeta("kube-master-10-1-0-0-24"), Spec: apiv3.BlockAffinitySpec{Node: "kube-master", CIDR: "10.1.0.0/24"}},
+		{ObjectMeta: objectMeta("kube-master-10-2-0-1-32"), Spec: apiv3.BlockAffinitySpec{Node: "kube-master", CIDR: "10.2.0.1/32"}},
+		{ObjectMeta: objectMeta("kube-master-192-168-221-192-26"), Spec: apiv3.BlockAffinitySpec{Node: "kube-master", CIDR: "192.168.221.192/26"}},
+		{ObjectMeta: objectMeta("kube-master-192-168-221-64-26"), Spec: apiv3.BlockAffinitySpec{Node: "kube-master", CIDR: "192.168.221.64/26", State: "confirmed"}},
+		{ObjectMeta: objectMeta("kube-master-192-168-221-0-26"), Spec: apiv3.BlockAffinitySpec{Node: "kube-master", CIDR: "192.168.221.0/26", State: "pending"}},
+	}
+	for _, a := range affinities {
+		_, err := cc.BlockAffinities().Create(ctx, &a, options.SetOptions{})
+		if err != nil {
+			// May already exist if IPPool deletion didn't clean it up.
```
caseydavenport: Why would this happen? We either need this, or we don't.

confd/tests/template_render_test.go
```diff
@@ -0,0 +1,558 @@
...
+			require.NoError(t, getErr, "BlockAffinity %s creation failed and get also failed: create err: %v", a.Name, err)
+			a.ResourceVersion = existing.ResourceVersion
+			_, err = cc.BlockAffinities().Update(ctx, &a, options.SetOptions{})
+			require.NoError(t, err, "updating BlockAffinity %s", a.Name)
+		}
+	}
+}
+
+// crdDir finds the Calico CRD directory by walking up from the current directory.
+func crdDir() string {
```
caseydavenport: We should use a common helper function for this - don't the other envtest tests in api/ and kube-controllers/ already have this?

confd/tests/helpers_test.go
```diff
@@ -0,0 +1,558 @@
...
+	t.Helper()
+	data, err := os.ReadFile(path)
+	require.NoError(t, err, "reading %s", path)
+
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(data), 4096)
+	ctx := context.Background()
+
+	for {
+		var raw map[string]any
+		if err := decoder.Decode(&raw); err != nil {
```
caseydavenport, line 347: I think we can use the universal deserializer with the appropriate schema registrations to decode files directly into Go objects, and then create them with a controller-runtime client rather than have per-type handling (we may still need it for Node, but not the other types).

confd/tests/template_render_test.go
```diff
@@ -0,0 +1,558 @@
...
+		}
+		if kind != "Node" {
+			require.NoError(t, err, "creating %s from %s", kind, path)
+		}
+	}
+}
+
+// deleteTestResources removes per-test Calico resources. Uses t.Logf for errors
+// instead of require so that cleanup always runs to completion.
+func deleteTestResources(t *testing.T, cc client.Interface) {
```
caseydavenport: Each test should know the resources it created - can't we be more selective in how we do this? i.e., delete or revert each resource the test created instead of querying everything?

confd/tests/template_render_test.go
```diff
@@ -0,0 +1,558 @@
...
+
+	confdConfig := &config.Config{
+		ConfDir:  confDir,
+		Onetime:  true,
+		SyncOnly: true,
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	err = run.RunWithContext(
+		ctx, confdConfig, confdCalicoClient, sharedK8sClient,
```
caseydavenport: Each argument on its own line please.

confd/tests/template_render_test.go
```diff
@@ -0,0 +1,558 @@
...
+		apiconfig.CalicoAPIConfigSpec{DatastoreType: apiconfig.Kubernetes},
+		&syncclientutils.TyphaConfig{},
+		nil,
+	)
+	require.NoError(t, err, "confd RunWithContext")
+
+	// Compare output files against golden files.
+	goldenFiles := []string{
+		"bird.cfg", "bird6.cfg",
```
caseydavenport: Each on its own line.

confd/tests/template_render_test.go
```diff
@@ -0,0 +1,558 @@
...
+		"bird_aggr.cfg", "bird6_aggr.cfg",
+	}
+	for _, f := range goldenFiles {
+		t.Run(f, func(t *testing.T) {
+			got, err := os.ReadFile(filepath.Join(outputDir, f))
+			require.NoError(t, err, "reading output %s", f)
+
+			want := readGolden(t, goldenDir, f)
+			if normalizeBlankLines(string(got)) != normalizeBlankLines(want) {
+				t.Errorf("output mismatch for %s\n\n--- want ---\n%s\n\n--- got ---\n%s", f, want, string(got))
```
caseydavenport: Can we make a nicer error message? Perhaps formatted with detailed information and contextual diffs? Also include paths to the generated / expected files on disk.

confd/tests/helpers_test.go
```diff
@@ -0,0 +1,558 @@
...
+
+func readGolden(t *testing.T, goldenDir, filename string) string {
+	t.Helper()
+	path := filepath.Join("compiled_templates", goldenDir, filename)
+	data, err := os.ReadFile(path)
+	require.NoError(t, err, "reading golden file %s", path)
+	return string(data)
+}
+
+func normalizeBlankLines(s string) string {
```
caseydavenport, line 804: Why do we need this?

confd/tests/helpers_test.go
```diff
@@ -0,0 +1,511 @@
...
+
+			expectedPath := filepath.Join("compiled_templates", goldenDir, f)
+			want, err := os.ReadFile(expectedPath)
+			require.NoError(t, err, "reading golden file %s", expectedPath)
+
+			gotNorm := normalizeBlankLines(string(got))
+			wantNorm := normalizeBlankLines(string(want))
+			if gotNorm != wantNorm {
+				t.Errorf("output mismatch for %s\nexpected: %s\nactual:   %s\n\n--- expected ---\n%s\n\n--- actual ---\n%s",
+					f, expectedPath, actualPath, string(want), string(got))
```
caseydavenport: Can we generate a minimal diff to include in the output?

confd/Makefile
```diff
@@ -38,70 +38,24 @@ endif
...
 	$(DOCKER_RUN) $(CALICO_BUILD) sh -c 'cd /go/src/$(PACKAGE_NAME) && ginkgo -r -race --junit-report=confd_ut.xml --output-dir=report/ .'
 
+.PHONY: fv
+## Run confd functional verification tests (envtest-based, replaces bash test suite).
+fv: setup-envtest
+	$(DOCKER_RUN) \
+		-v $$(ls -d $(ENVTEST_DIR)/k8s/*-$(BUILDOS)-$(BUILDARCH) | tail -1):/envtest:ro \
+		-e KUBEBUILDER_ASSETS=/envtest \
+		$(CALICO_BUILD) \
+		go test -count=1 -v -timeout 600s ./confd/tests/...
```
nelljerram: `-count=1` is default, so can we delete this?  Otherwise I'm wondering why it's there.

caseydavenport: Yeah, `-count=1` is there intentionally to disable test caching - we want them to always run fresh instead of just feeding cached results. I added a comment in the Makefile to make that clear.

caseydavenport:
So, I think I'd prefer to keep related tests in separate files rather than cramming everything into 3 big files - it makes it easier to find what you're looking for and keeps individual files short. But I agree the duplicated loop boilerplate across all the oneshot files was hiding the common pattern.

I've added a shared `oneshotTestCase` type and `runOneshotTests` helper that all the oneshot test files now use. Every oneshot test is now just a case list + one function call:

```go
func TestMeshTemplates(t *testing.T) {
    runOneshotTests(t, []oneshotTestCase{
        {name: "bgp-export", goldenDir: "mesh/bgp-export"},
        {name: "ipip-always", goldenDir: "mesh/ipip-always"},
        {name: "hash", goldenDir: "mesh/hash", envVars: map[string]string{"CALICO_ROUTER_ID": "hash"}},
        {name: "static-routes", goldenDir: "mesh/static-routes", kddOnly: true},
    })
}
```

The helper handles the backend loop, kddOnly skipping, envVar setup, and derives `inputYAML` from `goldenDir` (since it's always `goldenDir + "/input.yaml"`). Named fields mean the common case is just `{name, goldenDir}` with no `nil, false` noise.

If you still think this would be better as fewer files though, happy to consolidate further - it'd be a pretty mechanical change at this point.

confd/tests/README.md
```diff
@@ -1,55 +1,28 @@
...
 
-Similar to above, the compiled templates were pulled from a working cluster matching
-the configuration described by the `mock_data` directory structure.  It should be noted
-that there is no difference in compiled templates between etcd and KDD.
-
-If changes are needed, and simple enough, the compiled templates can just be adjusted
-by hand.  However if there were large changes to the templates, it would be better
-to spin up a new cluster, configure it, and pull the compiled templates from there
-so it can be verified that the new templates also work in a live cluster.
+If changes are needed, and simple enough, the compiled templates can just be adjusted by hand. However if there were large changes to the templates, it would be better to spin up a new cluster, configure it, and pull the compiled templates from there so it can be verified that the new templates also work in a live cluster.
```
nelljerram: The README text here was outdated.  In fact we've the UPDATE_EXPECTED_DATA option in Makefile - is there an equivalent for that in this refactoring?  (I think yes, and that I've seen it in passing.  So probably want to mention that here.)

confd/tests/bgpfilter_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+import "testing"
+
+func TestBGPFilterTemplates(t *testing.T) {
+	runOneshotTests(t, []oneshotTestCase{
+		{name: "single_filter/global_peer", goldenDir: "bgpfilter/single_filter/global_peer"},
+		{name: "single_filter/explicit_peer", goldenDir: "bgpfilter/single_filter/explicit_peer"},
+		{name: "multi_filter/global_peer", goldenDir: "bgpfilter/multi_filter/global_peer"},
+		{name: "multi_filter/explicit_peer", goldenDir: "bgpfilter/multi_filter/explicit_peer"},
+		{name: "node_mesh", goldenDir: "bgpfilter/node_mesh"},
```
nelljerram, line 25: Are we missing these here:
```
  test_bgp_filter_deletion
  test_bgp_filter_names
```

confd/tests/daemon_test.go
```diff
@@ -0,0 +1,393 @@
...
+// when iterating through many BGPPeers with password references. The deadlock
+// required the iteration to take >100ms, so we create many nodes/peers.
+func TestBGPPasswordDeadlock(t *testing.T) {
+	for _, be := range activeBackends {
+		if be.ctrlClient == nil {
+			continue // KDD-only: needs K8s API for Secrets
+		}
+		t.Run(be.name, func(t *testing.T) {
+			ctx := context.Background()
+			scale := 30
```
nelljerram: Is this really enough to see the deadlock?  We had 99 in shell.

confd/tests/bgpfilter_test.go
caseydavenport: Both are covered - `filter_names` is in `bgpfilter_test.go` line 29 (oneshot), and `filter_deletion` is in `daemon_test.go` as `TestBGPFilterDeletion`.

confd/tests/daemon_test.go
caseydavenport: Good call, bumped it back to 99 to match the old test.

caseydavenport:
Thanks for the thorough review Nell! Pushed a commit addressing the feedback:

**Daemon mode coverage:** Added `TestDaemonModeRendering` which runs representative scenarios (mesh, explicit peering, bgpfilter) through a single long-lived confd daemon, exercising the watch/re-render loop that the old `execute_tests_daemon` covered. Removed the unused `withTypha`/`startTypha` code - Typha doesn't work correctly with envtest (ClusterInformation CRD isn't served under the `crd.projectcalico.org` API group), and the watch/re-render path is the same with or without Typha.

**Step counts:** I double-checked and they actually match the old shell tests exactly - the Claude analysis was counting setup/teardown as separate steps. Mesh password has 5 verification points in both, BGP filter deletion has 2 in both, BGP peer password has 6 in both.

**Deadlock scale:** Bumped back to 99 to match the old test.

**README:** Updated to describe how to update golden files.

Replied to the inline comments too - `filter_names` and `filter_deletion` are both present, just in different files.

PR #12219, opened 2026-03-21
goldmane: optimize ingestion pipeline hot path

caseydavenport:
Three targeted optimizations to the BucketRing ingestion hot path, reducing end-to-end AddFlow latency by ~30%.

**Eliminate redundant findBucket call** — `AddFlow` was calling both `Window()` (which calls `findBucket`) and then `findBucket` again directly. Now calls `findBucket` once and derives the time window from the returned bucket. Also removes the now-unused `Window` method.

**Make findBucket O(1)** — The ring has a fixed interval per bucket, so we can compute the target index directly from `(headStart - t) / interval` instead of scanning linearly. Keeps a linear fallback (with a warning log) as a safety net in case of clock skew or unexpected bucket state.

**Use Clear() instead of iterate-and-Discard in bucket Reset** — `set.Set` already has a `Clear()` method backed by Go's builtin `clear()`, which is O(1) for map clearing. The old code was iterating all entries to call `Discard()` one at a time.

### Benchmark results

End-to-end `BucketRing.AddFlow` with 100 unique flow keys against a 242-bucket ring (production config):

| Metric | before | after | improvement |
|--------|--------|-------|-------------|
| ns/op | 2,085 | 1,463 | **30% faster** |
| B/op | 1,624 | 1,104 | **32% less memory** |
| allocs/op | 21 | 14 | **33% fewer allocs** |

### DIFF
--- a/goldmane/pkg/storage/bucket_ring.go
+++ b/goldmane/pkg/storage/bucket_ring.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2025 Tigera, Inc. All rights reserved.
+// Copyright (c) 2025-2026 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -336,15 +336,15 @@ func (r *BucketRing) Rollover(sink Sink) int64 {
 }
 
 func (r *BucketRing) AddFlow(flow *types.Flow) {
-	// Find the window for this Flow based on the global bucket ring. We use the ring to ensure
-	// that time windows are consistent across all DiachronicFlows.
-	start, end, err := r.Window(flow)
-	if err != nil {
+	// Find the bucket for this flow's timestamp. This determines the time window for
+	// the DiachronicFlow as well, so we only need one lookup.
+	_, bucket := r.findBucket(flow.StartTime)
+	if bucket == nil {
 		logrus.WithFields(logrus.Fields{
-			"start": flow.StartTime,
-			// "now":   a.nowFunc().Unix(),
+			"start":  flow.StartTime,
+			"oldest": r.BeginningOfHistory(),
+			"newest": r.EndOfHistory(),
 		}).WithFields(flow.Key.Fields()).
-			WithError(err).
 			Warn("Unable to sort flow into a bucket")
 		return
 	}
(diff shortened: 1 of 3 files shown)


goldmane/pkg/storage/bucket_ring.go
```diff
@@ -365,26 +365,11 @@ func (r *BucketRing) AddFlow(flow *types.Flow) {
...
-		}).Warn("Failed to find bucket, unable to ingest flow")
-		return
+	if logrus.IsLevelEnabled(logrus.DebugLevel) {
+		logrus.WithFields(bucket.Fields()).WithField("flowStart", flow.StartTime).Debug("Adding flow to bucket")
 	}
-
-	// TODO: Adding the flow to the bucket can currently block if the bucket is being accessed
-	// by another thread. This should be relatively rare and short, but ideally we'd make adding a Flow
-	// to a bucket non-blocking.
-	fields := bucket.Fields()
```
Brian-McM, line 384: Just wanted to double check (since it's not quite obvious to me, likely I'm not grokking something simple), you meant to delete this? Why don't we need this anymore?

caseydavenport: The log just moved within an `IsLevelEnabled` check above, but yeah it's missing the extra fields being added which I don't think was intentional.

caseydavenport: Putting it within the `IsLevelEnabled` check prevents us doing unnecessary work building the context logger when debug isn't enabled.

goldmane/pkg/storage/bucket_ring.go
```diff
@@ -336,15 +336,15 @@ func (r *BucketRing) Rollover(sink Sink) int64 {
 }
 
 func (r *BucketRing) AddFlow(flow *types.Flow) {
-	// Find the window for this Flow based on the global bucket ring. We use the ring to ensure
-	// that time windows are consistent across all DiachronicFlows.
-	start, end, err := r.Window(flow)
-	if err != nil {
+	// Find the bucket for this flow's timestamp. This determines the time window for
+	// the DiachronicFlow as well, so we only need one lookup.
+	_, bucket := r.findBucket(flow.StartTime)
```
skoryk-oleksandr, line 341: The refactor in `AddFlow` changed the error ordering. Previously `Window()` failed before creating a `DiachronicFlow`. Now `findBucket` returning nil happens after the `DiachronicFlow` is already created and added to all indices. Looks like now, flow with an out-of-range timestamp will leak an empty `DiachronicFlow` into `r.diachronics` and every index until rollover cleans it up.

caseydavenport: The findBucket nil check is actually at the top of AddFlow already (before the DiachronicFlow creation) -- I think the diff view might be confusing here since the old code had it further down. The early return on line 342-350 happens before we touch r.diachronics.

goldmane/pkg/storage/bucket_ring.go
caseydavenport: It moved into the IsLevelEnabled guard above to avoid the bucket.Fields() allocation when debug is off. I dropped the "head" field by accident though -- just pushed a fix to add it back.

skoryk-oleksandr (approved):

PR #12225, opened 2026-03-23
Consolidate Calico components into a single calico binary and image

caseydavenport:
Consolidates most Calico components into a single `calico/calico` image built from one Go binary. The binary dispatches via `calico component <name>` (cobra subcommands), with `calicoctl` shipped as a symlink.

Image count drops from 13 to 3: `calico/calico`, `calico/node`, `calico/whisker`. Node stays separate for CGO + BPF + runit. Whisker stays separate as a TypeScript frontend.

Needs the matching operator change: https://github.com/tigera/operator/pull/4608

## Commits

Each commit stands on its own; build and tests pass at every step.

1. **Harden cmdwrapper restart loop** - adds `WrapSelf` so exit-129 components still restart when invoked as `calico component <name>`.
2. **Migrate component CLIs to cobra** - every component exports `NewCommand()`. typha and calicoctl off docopt.
3. **Add combined calico binary and image** - new `cmd/calico/` + single `docker/calico/Dockerfile`.
4. **Rebuild node image around the combined binary** - CGO build for BPF; strips per-component image targets.
5. **Switch deployments and tests to the combined image** - Helm, manifests, kind, felix FV, KC FV, release tooling.
6. **Regenerate per-component deps.txt**.
7. **Wire up Semaphore CI for the combined image** - new `20-cmd-calico` block; folds in some incidental CI cleanup.
8. **Remove retired per-component entry points and docker image builds**.

```release-note
Calico now ships a single "quay.io/calico/calico" image containing most sub-components, drastically reducing the number of images required to install Calico.
```
```release-note
A number of calico/ docker images have been removed and subsumed by the new calico/calico image: calicoct, cni, kube-controllers, typha, etc.
```
```release-note
Release artifacts have changed as a result of consolidation from many images to the new calico/calico image.
```


nelljerram:
@caseydavenport From the PR title and first para of the description, I'm not totally clear if this is binary/executable consolidation, or image consolidation, or both.  Could you clarify?  It feels like image consolidation would be an even bigger deal.  If it is that, I recommend the title says "uber image" instead of "uber binary".

caseydavenport:
@nelljerram it's both! I'll clarify the title and description.

app-policy/cmd/dikastes/dikastes.go
```diff
@@ -15,29 +15,10 @@
...
-	"strconv"
-	"sync"
-	"syscall"
-	"time"
-
 	"github.com/docopt/docopt-go"
-	authz "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
-	log "github.com/sirupsen/logrus"
-	"google.golang.org/grpc"
+	logrus "github.com/sirupsen/logrus"
```
hjiawei: nit. no need to use `logrus` alias.

app-policy/cmd/healthz/healthz.go
hjiawei, line 1: This healthz differs significantly in enterprise. It may be better to align their implementation to minimize merge conflicts.

charts/crd.projectcalico.org.v1/templates/operator.tigera.io_imagesets.yaml
```diff
@@ -20,7 +20,7 @@ spec:
             ImageSet is used to specify image digests for the images that the operator deploys.
             The name of the ImageSet is expected to be in the format `<variant>-<release>`.
             The `variant` used is `enterprise` if the InstallationSpec Variant is
-            `CalicoEnterprise` or `TigeraSecureEnterprise`, otherwise it is `calico`.
+            `TigeraSecureEnterprise` otherwise it is `calico`.
```
hjiawei: Did we deprecate `TigeraSecureEnterprise` in favor of `CalicoEnterprise`?

cmd/calico/Makefile
hjiawei, line 1: Shall we move the `calico/calico` image build steps from the top-level Makefile into this component Makefile?

docker/calico/Dockerfile
```diff
@@ -0,0 +1,49 @@
+ARG CALICO_BASE
+ARG UBI_IMAGE
+
+FROM ${UBI_IMAGE} AS builder
+
+# pprof profile dumps for kube-controllers.
+RUN mkdir /profiles
+RUN touch /profiles/mem.pprof && chown 999 /profiles/mem.pprof
+RUN touch /profiles/cpu.pprof && chown 999 /profiles/cpu.pprof
```
hjiawei: Historically, Calico OSS components used different non-root UIDs (998, 999, 1000, 1001, etc.). We’ve since standardized on 10001:10001 for most components. Should we use 10001 here as well?

docker/calico/Dockerfile
```diff
@@ -0,0 +1,49 @@
+ARG CALICO_BASE
+ARG UBI_IMAGE
+
+FROM ${UBI_IMAGE} AS builder
+
+# pprof profile dumps for kube-controllers.
+RUN mkdir /profiles
```
hjiawei: Do we still need to create an empty directory and files for kube-controller to run properly? Would it be possible to create tem under /tmp at runtime when profiling is enabled?

docker/calico/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+LABEL org.opencontainers.image.licenses="Apache-2.0"
+
+LABEL description="Calico combined binary containing all Calico components."
+LABEL maintainer="maintainers@tigera.io"
+LABEL name="Calico"
+LABEL release=1
+LABEL summary="Calico combined binary containing all Calico components."
+LABEL vendor="Project Calico"
+LABEL version="${GIT_VERSION}"
+
```
hjiawei, line 55: Should we add a `USER 10001:10001` instruction here as a fallback?

felix/bpf-gpl/Makefile
hjiawei, line 1: This change removes ppc64le build support. Shall we revert it?

felix/bpf/libbpf/libbpf.go
hjiawei, line 1: This change removes ppc64le build support. Shall we revert it?

felix/bpf-gpl/conntrack.h
hjiawei, line 1: I'm not sure if this change belongs in this PR. Shall we commit it separately?

felix/bpf/ut/pkt_byte_counts_test.go
hjiawei, line 1: I'm not sure if this change belongs in this PR. Shall we commit it separately?

felix/fv/flow_logs_goldmane_test.go
```diff
@@ -553,7 +553,6 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ goldmane nat outgoing flow
 		}, "30s", "3s").ShouldNot(HaveOccurred())
 	})
 })
-
```
hjiawei, line 556: nit. keep the new line?

felix/Makefile
hjiawei, line 1: These changes remove the ppc64le eBPF build. Is this intentional?

felix/etc/calico-diags
hjiawei, line 1: If these changes are still needed, shall we commit them in a separate PR?

hack/test/kind/infra/calico_versions.yml
```diff
@@ -50,3 +50,5 @@ components:
     version: test-build
   istio-proxyv2:
     version: test-build
+  calico:
+    version: test-build
```
hjiawei: nit. order the components in this yaml alphabetically?

docker/calico/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# (hash)releases — only the staged layer above it changes from one build to
+# the next, so the base layer is reused from the registry cache.
+FROM scratch AS source
+
+ARG TARGETARCH
+ARG BIN_DIR
+
+COPY ${BIN_DIR}/calico-${TARGETARCH} /usr/bin/calico
+# Provide calicoctl for backward compatibility with scripts that exec into the container.
+COPY ${BIN_DIR}/calico-${TARGETARCH} /usr/bin/calicoctl
```
hjiawei: Do we have a busybox-style symlink + argv[0] detection for `calicoctl` so we don't need to copy the real `calicoctl` binary to the image?

hjiawei: Two ways to avoid the duplicate copy, both with a small change in `calico`'s `main`:

**1. Symlink + argv[0] dispatch (busybox-style)**

```dockerfile
RUN ln -s calico /usr/bin/calicoctl
```

```go
if filepath.Base(os.Args[0]) == "calicoctl" {
    os.Args = append([]string{"calico", "ctl"}, os.Args[1:]...)
}
```

**2. Shebang shim (the UBI/coreutils pattern — `cat /usr/bin/cat` to see it)**

```dockerfile
RUN printf '#!/usr/bin/calico --shebang=ctl\n' > /usr/bin/calicoctl && chmod +x /usr/bin/calicoctl
```

Shebang is a kernel feature (`binfmt_script`), no shell required. The kernel rewrites `calicoctl get nodes` into `calico --shebang=ctl /usr/bin/calicoctl get nodes`, so `calico` strips both:

```go
if len(os.Args) >= 3 && strings.HasPrefix(os.Args[1], "--shebang=") {
    sub := strings.TrimPrefix(os.Args[1], "--shebang=")
    os.Args = append([]string{os.Args[0], sub}, os.Args[3:]...)
}
```

Symlink is simpler and zero-overhead; shebang is a regular file (friendlier for `rpm -V`, self-documenting). Either avoids the ~50MB duplicate.

hack/set_quay_expiry.py
```diff
@@ -422,7 +422,7 @@ def main():
 
     for image_spec in args.image_names:
         image_spec = image_spec.removeprefix("https://")
-        if not image_spec.startswith("quay.io/"):
+        if not image_spec.startswith("quay.io"):
```
hjiawei: Revert this change? It was added in https://github.com/projectcalico/calico/pull/12321 as a security fix.

node/Makefile
hjiawei, line 1: Some of the changes in this Makefile remove the ppc64le eBPF build. Is this intentional?

caseydavenport: Nope, good catch - a lot of these I think came from bad merges with master.

cmd/calico/main.go
```diff
@@ -0,0 +1,71 @@
...
+	_, filename := filepath.Split(os.Args[0])
+	switch filename {
+	case "calico-ipam", "calico-ipam.exe":
+		ipamplugin.Main(buildinfo.Version)
+		return
+	case "calicoctl", "calicoctl.exe":
+		// Dispatch to ctl subcommand when invoked as calicoctl.
+		os.Args = append([]string{"calico", "ctl"}, os.Args[1:]...)
+	}
+	if os.Getenv("CNI_COMMAND") != "" {
```
hjiawei: Minor footgun: this `CNI_COMMAND` check happens after the `calicoctl` argv rewrite above, but doesn't gate on whether the caller asked for a specific subcommand. If a process happens to have `CNI_COMMAND` set in its env (e.g. someone debugging in a CNI-install pod) and runs `calicoctl get nodes`, the binary silently dispatches to `plugin.Main` instead of calicoctl. Cheap fix: only consult `CNI_COMMAND` when invoked as plain `calico` with no subcommand args.

hjiawei (commented):
A few items that don't appear to have been raised yet — leaving inline. Most are minor; the `massageError` loss and the `calico health` TLS gap stood out.

Process items I'd suggest adding to the PR description: (1) deprecation story for the standalone images (`calico/cni`, `calico/typha`, `calico/kube-controllers`, `calico/dikastes`, `calico/key-cert-provisioner`) — kept, frozen, removed? (2) link to the matching operator PR. (3) no tests under `cmd/calico/` for the dispatch logic in `main.go` — the four branches (`calicoctl` basename, `calico-ipam` basename, `CNI_COMMAND` env, normal) are load-bearing for every container invocation and would be very cheap to cover with a table-driven test.

Also: Copilot's claim that `DatastoreMigration` was dropped is **stale** — it's re-registered at `kube-controllers/pkg/kubecontrollers/run.go:486-527`. Worth resolving that thread.

cmd/calico-windows/main.go
```diff
@@ -0,0 +1,122 @@
...
+// This binary imports only Windows-safe packages, avoiding the Linux-only
+// dependencies in the full uber binary (typha, pod2daemon, dikastes, etc.).
+//
+// When symlinked or copied as "calico-ipam.exe", dispatches to the IPAM
+// plugin. When CNI_COMMAND is set, dispatches to the CNI plugin. Otherwise,
+// runs the Cobra command tree.
+func main() {
+	_, filename := filepath.Split(os.Args[0])
+	switch filename {
+	case "calico-ipam", "calico-ipam.exe":
```
hjiawei, line 42: Asymmetry vs. `cmd/calico/main.go`: the Linux dispatcher handles `calicoctl.exe` (`cmd/calico/main.go:58` — dead code there) but this Windows dispatcher only handles `calico-ipam.exe`, not `calicoctl.exe`. If we want the calicoctl backward-compat shim to work on Windows, add `"calicoctl", "calicoctl.exe"` here too. Otherwise, drop the `.exe` cases from the Linux file.

cmd/calico/health.go
```diff
@@ -0,0 +1,66 @@
...
+HealthAggregator pattern. Sends an HTTP GET to the component's
+/readiness or /liveness endpoint and exits with code 0 on success.`,
+		Run: func(cmd *cobra.Command, args []string) {
+			if checkType != "readiness" && checkType != "liveness" {
+				fmt.Fprintf(os.Stderr, "Invalid check type %q, expected \"readiness\" or \"liveness\"\n", checkType)
+				os.Exit(1)
+			}
+
+			url := fmt.Sprintf("http://%s:%d/%s", host, port, checkType)
+			client := &http.Client{Timeout: 5 * time.Second}
```
hjiawei: Three small design gaps for a generic health probe:

1. **No TLS support** — URL is hardcoded `http://`. If any component (apiserver in particular) ever serves health over TLS, this can't reach it. Add `--scheme` + optional `--insecure-skip-verify`/`--ca-cert`.
2. **Hardcoded 5s timeout** — fine in steady state, but not configurable for slow-startup components. Add `--timeout`.
3. **`>= 200 && < 400` is unusually permissive** — HealthAggregator never returns 3xx so this is benign today, but `< 300` is the conventional check and would catch a misconfigured proxy that sneaks in a redirect.

calicoctl/calicoctl/commands/command.go
```diff
@@ -0,0 +1,146 @@
...
+// NewCommand returns a cobra command tree for calicoctl.
+func NewCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "ctl",
+		Short: "Calico CLI tool for managing Calico resources",
+		Long: `The calicoctl command line tool is used to manage Calico network and security
+policy, to view and manage endpoint configuration, and to manage a Calico
+node instance.`,
+		SilenceUsage:  true,
+		SilenceErrors: true,
```
hjiawei, line 72: With `SilenceErrors: true` + `SilenceUsage: true`, error printing falls through to `cmd/calico/main.go:67-69`, which prints raw `err.Error()`. That **drops the `massageError` post-processing** from the legacy main (`calicoctl/calicoctl/calicoctl.go:152-164`) — the bit that strips the confusing `error unmarshaling JSON: while decoding JSON: json:` prefix and rewrites `unknown field "X"` into a friendlier message. End-user error UX for malformed YAML/JSON will regress.

Either re-apply `massageError` here (in a wrapper around `RunE` returns) or expose it from the commands package and call it from the parent `Execute()` site.

cmd/deps.txt
```diff
@@ -0,0 +1,556 @@
+!!! GENERATED FILE, DO NOT EDIT !!!
```
hjiawei, line 1: 556 lines, hand-maintained — will bit-rot when packages move. Worth a follow-up to generate this from `go list -deps -f '{{ if not .Standard }}{{ .Dir }}{{ end }}' ./cmd/calico/...` (or similar) and add a `make check-calico-deps` target so CI flags staleness.

cmd/calico/main.go
```diff
@@ -0,0 +1,71 @@
...
+	// When installed as a CNI plugin on the host, the binary may be invoked
+	// directly by the container runtime. Detect this and dispatch accordingly.
+	_, filename := filepath.Split(os.Args[0])
+	switch filename {
+	case "calico-ipam", "calico-ipam.exe":
+		ipamplugin.Main(buildinfo.Version)
+		return
+	case "calicoctl", "calicoctl.exe":
+		// Dispatch to ctl subcommand when invoked as calicoctl.
+		os.Args = append([]string{"calico", "ctl"}, os.Args[1:]...)
```
hjiawei: This rewrite drops the original `os.Args[0]` (`/usr/bin/calicoctl`) — afterwards anything that reads `os.Args[0]` for log prefixes, panic traces, or help text will see `calico` instead. Probably fine in practice, but worth a quick grep through the `ctl` subtree — if any subcommand uses `os.Args[0]` to construct user-facing strings, preserve the original name (e.g. set `commands.InvokedAs = "calicoctl"` before rewriting).

docker/calico/Dockerfile
caseydavenport: Good idea — dropped the duplicate COPY and shipped calicoctl as a symlink. Saves ~210MB per image.

cmd/calico/health.go
caseydavenport: Added `--timeout`. Passing on the other two:

- TLS: HealthAggregator only does plain HTTP (`ListenAndServe` in `libcalico-go/lib/health/health.go`), and `calico health` is scoped to HealthAggregator endpoints, so scheme/CA/insecure flags don't buy us anything today.
- `< 300`: HealthAggregator explicitly documents `200 <= status < 400` as good (`libcalico-go/lib/health/health.go:344-357`), so I'd rather mirror the contract than diverge.

cmd/calico/Makefile
caseydavenport: Done — moved `build`, `build-cgo`, `image`, `image-all` into `cmd/calico/Makefile` and the marker to `cmd/calico/.image.created-$(ARCH)`, matching the typha/node/whisker pattern.

cmd/deps.txt
caseydavenport: Already auto-generated — see the `!!! GENERATED FILE !!!` header at the top and `make gen-deps-files` (wired into `make generate`) in the root Makefile, which regenerates via `hack/cmd/deps`. CI runs `gen-deps-files` + `check-dirty` in the prerequisites block (`.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml:45-55`), so staleness fails the build.

app-policy/cmd/healthz/healthz.go
caseydavenport: Yeah, the divergence predates this PR — OSS uses the custom `dikastesproto.Healthz`, enterprise uses `grpc_health_v1`. This PR just extracts the existing OSS client into a package for the uber binary.

Full protocol alignment is a bigger job (remove `proto/healthz.proto`, switch dikastes to `grpc_health_v1`, update probe configs). Filed as CORE-12663 — out of scope for this PR.

charts/crd.projectcalico.org.v1/templates/operator.tigera.io_imagesets.yaml
caseydavenport: Good catch — the matching operator branch (`casey-uber-operator-v2`) was ~60 commits behind `tigera/operator:master`, including the commit that added `CalicoEnterprise` as the preferred variant name (https://github.com/tigera/operator/pull/4607). That staleness was polluting the CRDs pulled via `make get-operator-crds` in this PR.

I've merged master into the operator branch, force-pushed, and re-ran `make generate` here. This comment is addressed, plus it cleans up ~45K lines of unrelated CRD churn that other reviewers would otherwise have hit.

felix/bpf-gpl/Makefile
caseydavenport: Reverted — bad-merge noise, master always had ppc64le support.

felix/bpf/libbpf/libbpf.go
caseydavenport: Reverted — bad-merge noise, master always had ppc64le support.

felix/Makefile
caseydavenport: Reverted — bad-merge noise, master always had ppc64le support.

felix/bpf-gpl/conntrack.h
caseydavenport: Reverted — bad-merge noise, wasnt meant to be here.

felix/bpf/ut/pkt_byte_counts_test.go
caseydavenport: Reverted — bad-merge noise, wasnt meant to be here.

felix/etc/calico-diags
caseydavenport: Reverted — bad-merge noise.

hack/set_quay_expiry.py
caseydavenport: Reverted — good catch on the security fix.

node/Makefile
caseydavenport: Cleaned up in the latest push — node/Makefile now has just the intentional uber-binary hunks.

Makefile
```diff
@@ -69,9 +69,6 @@ check-images-availability: bin/crane bin/yq
 check-language:
 	./hack/check-language.sh
 
-check-mockery-config:
-	./hack/check-mockery-config.sh
-
 check-ginkgo-v2:
 	./hack/check-ginkgo-v2.sh
```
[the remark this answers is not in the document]
caseydavenport: `check-mockery-config` is still defined at `Makefile:74-75` and CI still invokes it (`.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml:82`). Nothing was removed here — this PR does not modify that target or its caller. Dismissing.

Makefile
```diff
@@ -186,7 +183,7 @@ image:
 E2E_FOCUS ?= "sig-network.*Conformance|sig-calico.*Conformance|BGP"
 E2E_SKIP ?= ""
 E2E_PROCS ?= 4
-K8S_NETPOL_SUPPORTED_FEATURES ?= "ClusterNetworkPolicy,ClusterNetworkPolicyNamedPorts"
+K8S_NETPOL_SUPPORTED_FEATURES ?= "ClusterNetworkPolicy"
 K8S_NETPOL_UNSUPPORTED_FEATURES ?= ""
```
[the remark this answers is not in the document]
caseydavenport: Misread — `K8S_NETPOL_SUPPORTED_FEATURES` still contains `ClusterNetworkPolicy,ClusterNetworkPolicyNamedPorts` in master and in this PR (`Makefile:230`). The e2e features list is unchanged. Dismissing.

app-policy/cmd/dikastes/dikastes.go
caseydavenport: Dropped the alias.

felix/fv/flow_logs_goldmane_test.go
caseydavenport: Restored.

hack/test/kind/infra/calico_versions.yml
caseydavenport: Moved `calico` into alphabetical position (between apiserver and cni).

kube-controllers/cmd/kube-controllers/main.go
```diff
@@ -551,49 +543,6 @@ func (cc *controllerControl) InitControllers(
 		cc.controllers["NetworkPolicyMigrator"] = policyMigrator
 	}
```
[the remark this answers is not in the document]
caseydavenport: Stale — `DatastoreMigration` is re-registered in the shared run package at `kube-controllers/pkg/kubecontrollers/run.go:487-526`.

cmd/calico/main.go
caseydavenport: Fixed — CNI_COMMAND now only dispatches when invoked as plain `calico` with no subcommand args. The basename switch handles `calicoctl` and `calico-ipam` before the CNI_COMMAND check is reached.

cmd/calico-windows/main.go
caseydavenport: Dropped the `.exe` cases from the Linux dispatcher — no calicoctl.exe ships for Windows, so that branch was dead code.

calicoctl/calicoctl/commands/command.go
caseydavenport: Fixed. Exported `commands.MassageError` and added a `wrapRunEWithMassageError` walk in `NewCommand()` that wraps every RunE in the tree. The standalone calicoctl now reuses the exported function too (deleted the local copy).

cmd/calico/main.go
caseydavenport: Fixed — the rewrite now preserves `os.Args[0]` by inserting `ctl` between argv[0] and the rest, rather than replacing argv[0]. Grepped the ctl subtree; only `calicoctl/calicoctl/util/docstring.go:18` reads argv[0] (for kubectl plugin detection) and now sees the original `calicoctl` path.

pod2daemon/pkg/csi/run.go
```diff
@@ -0,0 +1,58 @@
...
+		logrus.WithError(err).Fatal("Failed to retrieve CSI driver config")
+	}
+
+	if logLevel != "" {
+		config.LogLevel = logLevel
+	}
+
+	level, err := logrus.ParseLevel(config.LogLevel)
+	if err != nil {
+		logrus.WithError(err).Fatalf("Could not parse the log level")
```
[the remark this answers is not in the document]
caseydavenport: Already fixed — see commit c519db2a06 (`WithError(err).WithField("logLevel", driverCfg.LogLevel).Fatal(...)`).

kube-controllers/pkg/kubecontrollers/run.go
```diff
@@ -0,0 +1,494 @@
...
+		}
+		cancel()
+
+		healthCtx, cancel = context.WithTimeout(ctx, timeout)
+		healthStatus := 0
+		result := k8sClientset.Discovery().RESTClient().Get().AbsPath("/healthz").Do(healthCtx).StatusCode(&healthStatus)
+		cancel()
+		if healthStatus != http.StatusOK {
+			logrus.WithError(result.Error()).WithField("status", healthStatus).Errorf("Received bad status code from apiserver")
+			s.SetReady("KubeAPIServer", false, fmt.Sprintf("Error reaching apiserver: %v with http status code: %d", err, healthStatus))
```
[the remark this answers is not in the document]
caseydavenport: Already fixed — see commit c519db2a06; the message now uses `result.Error()` from the `/healthz` request instead of the stale datastore err.

docker/calico/Dockerfile
caseydavenport: The chown lines went away with the pprof files themselves — see the reply on the `/profiles` thread. If we bring them back we can use 10001:10001.

docker/calico/Dockerfile
caseydavenport: Dropped entirely. Grepped calico + calico-private + operator — nothing reads or writes `/profiles/*.pprof`. The setup was inherited from the old kube-controllers Dockerfile copy-paste and never served a purpose, so removing the files rather than relocating them.

docker/calico/Dockerfile
caseydavenport: Added `USER 10001:10001` in the final stage as a fallback default.

caseydavenport:
@hjiawei @nelljerram — updated the PR description with the process items you flagged: clearer lead paragraph, deprecation story (clean break next release), link to the matching operator PR, and the new dispatch tests at `cmd/calico/main_test.go`.

All threads addressed. Ready for another look.

calicoctl/calicoctl/commands/ipam/split.go
```diff
@@ -204,6 +204,99 @@ Examples:
 	return nil
 }
 
+// SplitPool splits an IP pool identified by CIDR or name into the given number of smaller pools.
+func SplitPool(ctx context.Context, config, cidr, poolName string, splitNum int) error {
```
caseydavenport, line 97: Why is this diff just new code? Where was the old implementation of this?

cni-plugin/Makefile
```diff
@@ -64,33 +45,24 @@ clean-windows: clean-windows-builder
 ###############################################################################
 # Building the binary
 ###############################################################################
+# Linux binaries are only built locally for running unit tests. The production
```
caseydavenport, line 48: We should probably use the `calico` binary for tests too - then we can ditch these targets entirely?

felix/fv/utils/utils.go
```diff
@@ -48,7 +48,8 @@ type EnvConfig struct {
 	FelixImage   string `default:"calico/felix-test:latest"`
 	EtcdImage    string `default:"quay.io/coreos/etcd"`
 	K8sImage     string `default:"calico/go-build:latest"`
-	TyphaImage   string `default:"calico/typha:latest"`
+	TyphaImage   string `default:"calico/calico:latest"`
+	TyphaCmd     string `default:"component typha"`
```
caseydavenport: Why do we need a configurable TyphaCmd? I think it's always "component typha" ?

felix/k8sfv/run-test
```diff
@@ -212,13 +212,15 @@ if ${USE_TYPHA}; then
        --name ${prefix}-typha \
        -e CALICO_DATASTORE_TYPE=kubernetes \
        -e TYPHA_LOGSEVERITYSCREEN=${TYPHA_LOG_LEVEL} \
+       -e TYPHA_LOGFILEPATH=none \
```
caseydavenport: Think this isn't needed since the image includes typha.cfg now.

goldmane/cmd/goldmane/command.go
```diff
@@ -0,0 +1,32 @@
...
+package daemon
+
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+)
+
+// NewCommand returns a cobra command that runs the Goldmane flow aggregation service.
+func NewCommand() *cobra.Command {
```
caseydavenport, line 26: Probably should standardize on `calico/<component>/cmd/<component>` for the `NewCommand` entrypoints across all.

That keeps them consistently located, while still keeping the package names correct for callers to use clearly:

```go
goldmane.NewCommand()
```

hack/test/kind/infra/calico_versions.yml
```diff
@@ -22,6 +22,8 @@ components:
     version: test-build
   apiserver:
```
caseydavenport: We should remove the old images now as they are unused.

hack/test/kind/infra/values.yaml
```diff
@@ -104,6 +104,10 @@ goldmane:
     spec:
       template:
         spec:
+          # Pin to control-plane so goldmane starts as soon as calico-node
+          # is ready there, without waiting for worker nodes.
+          nodeSelector:
+            kubernetes.io/hostname: kind-control-plane
```
caseydavenport: This seems like an unrelated change. We should revert it.

hack/test/kind/deploy_resources.sh
```diff
@@ -184,9 +184,6 @@ for extra in ${EXTRA_VALUES_FILES:-}; do
 done
 echo "Helm values files: ${VALUES_FILE} ${EXTRA_VALUES_FILES:-}"
 ${HELM} install calico ${CHART} "${helm_values_args[@]}" -n tigera-operator --create-namespace
-
-echo "Install calicoctl as a pod"
-${kubectl} apply -f ${INFRA_DIR}/calicoctl.yaml
```
caseydavenport, line 189: Also unrelated?

kube-controllers/pkg/controllers/flannelmigration/flannel_migration_fv_test.go
```diff
@@ -159,7 +159,7 @@ var _ = Describe("flannel-migration-controller FV test", Ordered, ContinueOnFail
 
 			startController()
 
-			w := migrationController.WatchStderrFor(regexp.MustCompile(`.*no migration process is needed.*`))
+			w := migrationController.WatchStdoutFor(regexp.MustCompile(`.*no migration process is needed.*`))
```
caseydavenport, line 162: For reviewers, this is a consequence of the new cmdwrapper logic.

kube-controllers/pkg/controllers/flannelmigration/k8s_resources.go
```diff
@@ -492,51 +499,124 @@ func (n k8snode) execCommandInPod(k8sClientset *kubernetes.Clientset, namespace,
 		return "", fmt.Errorf("failed to execute command in pod. Pod %s is not ready", pod.Name)
 	}
 
-	cmdArgs := []string{"exec", pod.Name, fmt.Sprintf("--namespace=%s", namespace), fmt.Sprintf("-c=%s", containerName), "--"}
-	cmdArgs = append(cmdArgs, args...)
-	out, err := exec.Command("/usr/bin/kubectl", cmdArgs...).CombinedOutput()
+	req := k8sClientset.CoreV1().RESTClient().Post().
```
caseydavenport, line 502: For reviewers: we used to ship `kubectl` inside the flannelmigration image. I didn't want to add that to calico/calico, so I've re-implemented the logic that depended on it to avoid needing the kubectl binary.

pkg/cmdwrapper/cmdwrapper.go
```diff
@@ -15,95 +15,162 @@
...
-	})
 	if len(os.Args) < 2 {
 		logrus.Fatalf("Invalid invocation of command wrapper, expected: %s <wrapped command>", os.Args[0])
 	}
-	prog := os.Args[1]
-	args := os.Args[2:]
+	runLoop(os.Args[1], os.Args[2:], nil)
+}
+
+// WrapSelf provides the same restart-on-RestartReturnCode semantics as Run,
```
caseydavenport: Do we still use Run() anywhere? Should this just be a library function at this point?

pod2daemon/pkg/flexvol/flexvol.go
```diff
@@ -46,11 +46,12 @@ type Response struct {
 	VolumeName string `json:"volumename,omitempty"`
 }
 
+// Capabilities describes the driver capabilities reported to kubelet.
 type Capabilities struct {
 	Attach bool `json:"attach"`
 }
 
-// Response to the 'init' command.
+// InitResponse is the response to the 'init' command.
```
caseydavenport, line 54: I think we've moved entirely over to CSI instead of flexvol - wonder if we could remove this entirely in a follow-on.

process/testing/winfv-felix/setup-fv.sh
```diff
@@ -104,19 +104,50 @@ function get_logs(){
...
 
-  # Get logs from windows pod
-  ${KUBECTL} --kubeconfig="${KUBECONFIG}" logs -n calico-system -l k8s-app=calico-node-windows -c uninstall-calico > ./pod-logs/win-uninstall-calico.log || echo "Failed to get logs for win-uninstall-calico"
-  ${KUBECTL} --kubeconfig="${KUBECONFIG}" logs -n calico-system -l k8s-app=calico-node-windows -c install-cni > ./pod-logs/win-install-cni.log || echo "Failed to get logs for win-install-cni"
-  ${KUBECTL} --kubeconfig="${KUBECONFIG}" logs -n calico-system -l k8s-app=calico-node-windows -c node > ./pod-logs/win-node.log || echo "Failed to get logs for win-node"
-  ${KUBECTL} --kubeconfig="${KUBECONFIG}" logs -n calico-system -l k8s-app=calico-node-windows -c felix > ./pod-logs/win-felix.log || echo "Failed to get logs for win-felix"
-
-  # Get logs from linux pod
-  ${KUBECTL} --kubeconfig="${KUBECONFIG}" logs -n calico-system -l k8s-app=calico-node -c calico-node > ./pod-logs/linux-calico-node.log || echo "Failed to get logs for linux-calico-node"
+  # Dump cluster state and container logs to stdout as well as files so that
```
caseydavenport, line 107: For reviewers: The changes here were to help with diagnostics when debugging windows failures in this PR. Could revert, but probably useful to keep.

release/internal/pinnedversion/pinnedversion_test.go
```diff
@@ -55,36 +55,32 @@ func TestImageComponents(t *testing.T) {
...
 	}
+	// The combined calico image (from cmd/calico) bundles typha, apiserver,
+	// kube-controllers, dikastes, goldmane, guardian, whisker-backend,
+	// key-cert-provisioner, CSI, flexvol, webhooks, and the Linux CNI
+	// plugin. Those components no longer get independent entries.
+	// "calico" is listed in the pinned version file as a meta-component
+	// representing the product release, but it is filtered out by
+	// ImageComponents (see noImageComponents), so the expected maps below
+	// do not include it.
+	commonComponents := map[string]registry.Component{
```
caseydavenport, line 66: OK, so not quite 3 images... but istio and envoy don't count!

webhooks/pkg/webhook/command.go
```diff
@@ -14,10 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package webhook
 
 import (
-	"crypto/tls"
+	crypto_tls "crypto/tls"
```
caseydavenport:
```suggestion
	cryptotls "crypto/tls"
```

webhooks/pkg/webhook/command.go
```diff
@@ -46,6 +43,11 @@ import (
 	"github.com/projectcalico/calico/webhooks/pkg/utils"
 )
 
+// maxRequestBodyBytes is the maximum size of an admission review request body.
+// The Kubernetes API server limits API objects to 3MB, so an AdmissionReview
+// wrapping a Calico resource will not legitimately exceed this.
+const maxRequestBodyBytes = 3 << 20 // 3MB
```
caseydavenport: This was already in master I think? Diff seems out of date.

webhooks/pkg/webhook/command.go
```diff
@@ -170,16 +167,13 @@ func readyFn() func(http.ResponseWriter, *http.Request) {
...
 }
 
-// handleFn implements utils.HandleFn to allow registration of webhooks.
 func handleFn(handler utils.AdmissionReviewHandler) func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
 		handleRequest(w, r, handler)
 	}
 }
 
-// handleRequest handles an incoming HTTP request, decodes the AdmissionReview, processes it, and writes the response.
```
caseydavenport, line 180: A lot of unrelated changes here moving code / removing comments. We should minimize the number of changed lines.

caseydavenport, line 180: Let's avoid making unrelated changes - e.g., removing comments or reordering code.

This PR is big enough as it is.

metadata.mk
```diff
@@ -66,8 +66,8 @@ LIBBPF_VERSION=v1.6.2
 BPFTOOL_IMAGE=calico/bpftool:v7.5.0
 
 # The operator branch corresponding to this branch.
-OPERATOR_BRANCH ?= master
-OPERATOR_ORGANIZATION ?= tigera
+OPERATOR_BRANCH ?= casey-uber-operator-v2
```
caseydavenport, line 69: Gotta remember to revert this - but there's a bit of a chicken / egg here as the two branches depend on each other.

manifests/generate.sh
```diff
@@ -64,7 +64,7 @@ ${HELM} -n tigera-operator template \
 	--set tigeraOperator.image=$OPERATOR_IMAGE \
 	--set tigeraOperator.registry=$OPERATOR_REGISTRY \
 	--set calicoctl.tag=$CALICO_VERSION \
-	--set calicoctl.image=$REGISTRY/ctl \
+	--set calicoctl.image=$REGISTRY/calico \
```
radTuti, line 72: should it still be "calicoctl" here? it will be confusing as this no longer just the calicoctl image?

manifests/generate.sh
```diff
@@ -158,7 +153,7 @@ ${HELM} template \
 	--set tigeraOperator.image=$OPERATOR_IMAGE \
 	--set tigeraOperator.version=$OPERATOR_VERSION \
 	--set tigeraOperator.registry=$OPERATOR_REGISTRY \
-	--set calicoctl.image=$REGISTRY/ctl \
+	--set calicoctl.image=$REGISTRY/calico \
```
radTuti, line 161: same concern as above - calico/calicoctl mismatch

release/pkg/manager/calico/manager.go
```diff
@@ -1076,13 +1024,8 @@ func (r *CalicoManager) buildReleaseTar() error {
...
 		images := map[string]string{
-			fmt.Sprintf("%s/node:%s", registry, r.calicoVersion):                         filepath.Join(imgDir, "calico-node.tar"),
-			fmt.Sprintf("%s/typha:%s", registry, r.calicoVersion):                        filepath.Join(imgDir, "calico-typha.tar"),
-			fmt.Sprintf("%s/cni:%s", registry, r.calicoVersion):                          filepath.Join(imgDir, "calico-cni.tar"),
-			fmt.Sprintf("%s/kube-controllers:%s", registry, r.calicoVersion):             filepath.Join(imgDir, "calico-kube-controllers.tar"),
-			fmt.Sprintf("%s/pod2daemon-flexvol:%s", registry, r.calicoVersion):           filepath.Join(imgDir, "calico-pod2daemon.tar"),
-			fmt.Sprintf("%s/dikastes:%s", registry, r.calicoVersion):                     filepath.Join(imgDir, "calico-dikastes.tar"),
-			fmt.Sprintf("%s/flannel-migration-controller:%s", registry, r.calicoVersion): filepath.Join(imgDir, "calico-flannel-migration-controller.tar"),
+			fmt.Sprintf("%s/calico:%s", registry, r.calicoVersion): filepath.Join(imgDir, "calico.tar"),
+			fmt.Sprintf("%s/node:%s", registry, r.calicoVersion):   filepath.Join(imgDir, "calico-node.tar"),
```
radTuti: not for PR but just noting down - multi arch support

radTuti, line 1028: Can we add a line to the release-note block mentioning the tarball contents change?

radTuti (commented):
Scoping this review mainly to release tooling changes.

manifests/generate.sh
caseydavenport: No, the goal of this PR is to remove images. Everything will be run through the calico/calico image!

calicoctl/calicoctl/commands/ipam/split.go
caseydavenport: Refactored — `Split` is now a docopt shim that delegates to `SplitPool`. See f716cffe.

cni-plugin/Makefile
caseydavenport: Filed as https://tigera.atlassian.net/browse/CORE-12712 to follow on.

felix/fv/utils/utils.go
caseydavenport: Yeah agreed — hardcoded to `component typha` now. See d91210542e.

felix/k8sfv/run-test
caseydavenport: Dropped. See d91210542e.

goldmane/cmd/goldmane/command.go
caseydavenport: Done. Moved goldmane, guardian, typha, and whisker-backend to `<component>/cmd/<component>`. See cb2e5a2737.

hack/test/kind/infra/calico_versions.yml
caseydavenport: Reverted the whole file — unrelated to this PR. See e7d2926c06.

hack/test/kind/infra/values.yaml
caseydavenport: Reverted. See e7d2926c06.

hack/test/kind/deploy_resources.sh
caseydavenport: Reverted. See e7d2926c06.

pkg/cmdwrapper/cmdwrapper.go
caseydavenport: Deleted — nothing calls it anymore. See 49e0582518.

pod2daemon/pkg/flexvol/flexvol.go
caseydavenport: Filed as https://tigera.atlassian.net/browse/CORE-12713 to follow on.

webhooks/pkg/webhook/command.go
caseydavenport: Reverted — the alias isn't needed. See 3911828a4d.

webhooks/pkg/webhook/command.go
caseydavenport: Reverted the unrelated bits and rewrote to minimize the diff — file is now a clean rename of `cmd/main.go` with just the NewCommand() wrapping. See 3911828a4d.

webhooks/pkg/webhook/command.go
caseydavenport: Minimized — see 3911828a4d.

caseydavenport: Minimized — see 3911828a4d.

.semaphore/semaphore.yml.d/blocks/20-e2e.yml
```diff
@@ -55,8 +33,16 @@
             value: "crd.projectcalico.org/v1"
           - name: E2E_TEST_CONFIG
             value: "e2e/config/kind-felix-routing.yaml"
-          - name: STORE_BUILD_CACHE
-            value: "true"
+      - name: "E2E tests: ClusterNetworkPolicy (cluster routing: Felix)"
```
mazdakn, line 36: The idea here is to run all e2e tests (not just ClusterNetworkPolicy) with Felix programming cluster routes. Eventually this mode should be set as default, then we can get rid of the other job.

Makefile
```diff
@@ -248,12 +249,11 @@ kind-migration-test:
 	KIND_CALICO_API_GROUP=crd.projectcalico.org/v1 $(MAKE) kind-up
 	$(REPO_ROOT)/hack/test/kind/migration/run_test.sh
 
-## Create a kind cluster and run all e2e tests.
+## Create a kind cluster and run the conformance e2e tests.
 e2e-test:
 	$(MAKE) -C e2e build
 	CLUSTER_ROUTING=$(CLUSTER_ROUTING) $(MAKE) kind-up
 	$(MAKE) e2e-run KUBECONFIG=$(KIND_KUBECONFIG)
-	$(MAKE) e2e-run-cnp KUBECONFIG=$(KIND_KUBECONFIG)
```
mazdakn, line 256: Why removing this conformance test?

caseydavenport: I've moved it to its own block in CI so that it runs in parallel instead of sequentially @mazdakn

.semaphore/semaphore.yml.d/blocks/20-e2e.yml
caseydavenport: Yep - I have three blocks now:

- E2E with BIRD
- E2E with Felix
- CNP E2E with Felix

mazdakn: Right. I misinterpreted the change.

Makefile
mazdakn: Got it. Make sense.

nelljerram (approved):
Let's go for it!

metadata.mk
nelljerram: Another reminder to revert this.

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
```diff
@@ -126,9 +126,132 @@
           - tar --use-compress-program="zstd -3" -cf /tmp/working-copy.tar.zstd $CALICO_DIR_NAME
           - gcs_branch_cache_dir="${GCS_CACHE_DIR}/${SEMAPHORE_GIT_BRANCH}"
           - gcloud storage cp /tmp/working-copy.tar.zstd "$gcs_branch_cache_dir/working-copy.tar.zstd"
+- name: "Pull: go-build image"
+  # Populate the GCS cache of calico/go-build so that every downstream job can
+  # docker-load it from GCS (faster, and avoids Docker Hub rate limits) rather
+  # than pulling cold from Docker Hub. The cache is keyed by GO_BUILD_VER so
+  # the tarball survives across workflows — the common case is a no-op
+  # `gcloud storage ls` on a cache hit.
```
nelljerram, line 134: Is this a separate change that we might want on existing release branches?

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
```diff
@@ -126,9 +126,132 @@
...
+          # cmd/calico build-cgo` above transitively built it via the Felix
+          # CGO dependency chain; the test-or-build below keeps this a
+          # no-op on a warm tree but still produces the artifact if the
+          # dependency chain is ever severed. Consumers (Felix: Build,
+          # Felix: UT) extract this into felix/bpf-gpl/ to skip the clone,
+          # avoiding the known github.com/libbpf/libbpf HTTP 500 flake
+          # and saving the ~30s build step across all Felix jobs.
+          - test -f felix/bpf-gpl/libbpf/src/amd64/libbpf.a || make -C felix libbpf
+          - LIBBPF_VERSION=$(grep -E '^LIBBPF_VERSION[[:space:]]*=' metadata.mk | cut -d'=' -f2 | tr -d '[:space:]')
+          - tar --use-compress-program="zstd -3" -cf /tmp/libbpf-amd64.tar.zst -C felix/bpf-gpl libbpf ".libbpf-${LIBBPF_VERSION}"
```
nelljerram, line 203: Worth a check that `.libbpf-${LIBBPF_VERSION}` really exists?  if we calculate LIBBPF_VERSION wrongly I guess tar might silently ignore it, and then the only symptom will be builds taking longer because stuff missing from the cache

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
```diff
@@ -126,9 +126,132 @@
...
+          - gcloud storage cp /tmp/libbpf-amd64.tar.zst "${GCS_WORKFLOW_DIR}/libbpf-amd64.tar.zst"
+          # TEMPORARY for PR #12225: warm master's "calico" build cache from this
+          # PR so downstream PR-consumer blocks get a realistic warm-cache
+          # measurement. REVERT BEFORE MERGE — once this lands on master, the
```
nelljerram: Revert reminder - although I can't actually see how the following code references this PR

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
```diff
@@ -126,9 +126,132 @@
...
+          # normal branch-build epilogue repopulates the cache.
+          - |-
+            pushd $HOME
+            tar --use-compress-program="zstd -3" -cf /tmp/build-cache.tar.zstd go ${CALICO_DIR_NAME}/.go-pkg-cache
+            gcloud storage cp /tmp/build-cache.tar.zstd "${GCS_CACHE_DIR}/master/build-cache-calico.tar.zstd"
+            popd
+- name: "Build: node image"
+  # Builds the node image on top of the combined calico binary. The underlying
+  # Go state lives in the "calico" build cache populated by "Build: calico
+  # image", so this block does not participate in build-cache storage.
```
nelljerram, line 208: Do you mean this isn't using the cache and so is repeating lots of the build?

nelljerram: Why is this now here instead of in 20-node.yml?

nelljerram: Is it just transitional that we still build node and whisker images at all?

charts/calico/templates/calico-kube-controllers.yaml
```diff
@@ -72,17 +73,17 @@ spec:
             # Choose which controllers to run.
             - name: ENABLED_CONTROLLERS
               value: policy,namespace,serviceaccount,workloadendpoint,node,loadbalancer
-{{- if .Values.kubeControllers.env }}
-{{ toYaml .Values.kubeControllers.env | indent 12 }}
-{{- end }}
```
nelljerram, line 77: Didn't somebody need this?

charts/calico/templates/calico-node.yaml
```diff
@@ -51,9 +51,9 @@ spec:
         # It can be deleted if this is a fresh installation, or if you have already
         # upgraded to use calico-ipam.
         - name: upgrade-ipam
-          image: {{.Values.cni.registry}}/{{.Values.cni.image}}:{{ .Values.version }}
+          image: {{.Values.calico.registry}}/{{.Values.calico.image}}:{{ .Values.version }}
           imagePullPolicy: {{.Values.imagePullPolicy}}
-          command: ["/opt/cni/bin/calico-ipam", "-upgrade"]
+          command: ["/usr/bin/calico-ipam", "-upgrade"]
```
nelljerram, line 56: Could we be consistent everywhere about whether we're using args of the unified binary, or a separate executable name that does the same thing?

charts/tigera-operator/README.md
```diff
@@ -168,7 +168,7 @@ tigeraOperator:
   image: tigera/operator
   registry: quay.io
 calicoctl:
-  image: quay.io/calico/ctl
+  image: quay.io/calico/calico
 
 # Optionally configure the host and port used to access the Kubernetes API server.
 kubernetesServiceEndpoint:
```
nelljerram, line 174: I think this is no longer supported, right - need to use podAnnotations instead.

felix/.semaphore/cache-test-artifacts
```diff
@@ -22,7 +22,9 @@ set -x
 
 artifact push job bin &
 docker save -o /tmp/calico-felix-test.tar calico/felix-test:latest-amd64
-docker save -o /tmp/felixtest-typha.tar felix-test/typha:latest-amd64
+# The combined calico image provides Typha (via `component typha`) and other
+# components used by the FVs.
+docker save -o /tmp/calico-image.tar calico/calico:latest-amd64
```
nelljerram, line 27: Does this still apply?  Now that calico/calico is pre-built, I would guess not?

nelljerram: (same for calico-image.tar below)

felix/bpf-gpl/Makefile
```diff
@@ -129,15 +129,23 @@ ut-objs: $(UT_OBJS)
...
 $(LIBBPF_FILE_CREATED):
-ifeq ("$(wildcard ./$(LIBBPF_DIR))", "")
-	$(info "Directory does not exist.")
-	@echo "Directory does not exist."
-	env -u GIT_DIR -u GIT_WORK_TREE git clone --depth 1 --single-branch https://github.com/libbpf/libbpf.git
-endif
-	rm -rf .libbpf-*
-	cd $(LIBBPF_DIR) && \
-		env -u GIT_DIR -u GIT_WORK_TREE git fetch --tags && env -u GIT_DIR -u GIT_WORK_TREE git checkout $(LIBBPF_VERSION)
+	@if [ -n "$$NO_LIBBPF_CLONE" ]; then \
```
nelljerram, line 142: Can we generally not use @ in Makefiles?  I almost always find later that it would have been helpful to know if the line happened or not.

felix/docker-image/calico-felix-wrapper
```diff
@@ -45,8 +45,11 @@ while true; do
   source /extra-env.sh
 
   echo "calico-felix-wrapper: Starting calico-felix"
+  # calico-felix here is the combined calico binary, dispatched via cobra to
+  # the felix subcommand. argv[0] stays "calico-felix" so /proc comm and
+  # GetSinglePID("calico-felix") continue to work.
   # Only enable tracebacks for Felix, not other binaries.
-  GOTRACEBACK=${FELIX_GOTRACEBACK} calico-felix &
+  GOTRACEBACK=${FELIX_GOTRACEBACK} calico-felix component felix &
```
nelljerram, line 52: So "calico-felix" here is just a copy of the "calico" binary, and the binary internally doesn't have special dispatch for being called "calico-felix"?  This is a bit confusing.

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
caseydavenport: Yep, I think it could be!

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
caseydavenport: I moved it here since it's being used as a prereq now for some downstream jobs so we don't duplicate the image build for e.g., the e2e jobs. 

We could probably keep it in the old location, though.

charts/tigera-operator/README.md
caseydavenport: Is that so? I thought this was still meant to work - but also, unrelated to this PR I think?

nelljerram: I was inferring from these lines being removed from 02-tigera-operator.yaml:
```
-        {{- if .Values.kubernetesServiceEndpoint.host }}
-        checksum/endpoints-config: {{ include (print $.Template.BasePath "/tigera-operator/01-kubernetes-services-endpoint.yaml") . | sha256sum }}
-        {{- end }}
-        {{- with .Values.podAnnotations }}
         {{- toYaml . | nindent 8 }}
-        {{- end }}
```

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
nelljerram: Sorry, I think this is actually fine; at the time I had forgotten that we do still build a node image separately from calico uber.

PR #12226, opened 2026-03-23
Kube controllers leak tests

lwr20:
**Release note:**
```release-note
TBD
```

### DIFF
--- a/kube-controllers/pkg/controllers/node/ipam_leak_test.go
+++ b/kube-controllers/pkg/controllers/node/ipam_leak_test.go
@@ -1,0 +1,699 @@
+// Copyright (c) 2025 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


kube-controllers/pkg/controllers/node/ipam_leak_test.go
```diff
@@ -0,0 +1,699 @@
...
+				return err == nil && exists
+			}, 2*time.Second, 50*time.Millisecond).Should(BeTrue(),
+				"pod should appear in informer cache")
+
+			obj, exists, err := podInformer.GetIndexer().GetByKey("default/live-pod")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(exists).To(BeTrue())
+			cached := obj.(*v1.Pod)
+
+			// THE LEAK GUARD: the cached pod must be the slim version.
```
caseydavenport, line 380: I'm not sure this one is actually a leak? It's just storing stuff we don't need? Or am I misunderstanding?

PR #12283, opened 2026-03-30
Fix bgp_local_pref handling in BGP filters with SetPriority

song-jiang:
## Description

BGPFilter's SetPriority modifies `krt_metric`, but the generated BIRD config didn't always reflect this in `bgp_local_pref`. Two issues fixed:

**Import filter:** `krt_metric` set in BIRD's BGP import filter doesn't survive to the kernel export filter (BIRD drops kernel-protocol-specific attributes in the RIB). The kernel export filter re-derives `krt_metric` from `bgp_local_pref`, so after BGPFilter processing with SetPriority we now emit `bgp_local_pref = INT_MAX - krt_metric` to persist the value. This is only emitted when a BGPFilter actually has SetPriority operations.

**Export filter:** For iBGP, `bgp_local_pref` was derived from `krt_metric` *before* the BGPFilter ran, so SetPriority in export rules had no effect on the advertised LOCAL_PREF. Moved the conversion to after BGPFilter processing, and only emit it when SetPriority operations are present.

### Changes
- Added `filterHasSetPriority()` helper that inspects BGPFilter rules for SetPriority operations
- `buildImportFilter()`: conditionally emit `bgp_local_pref = INT_MAX - krt_metric` only when SetPriority is present
- `buildExportFilter()`: moved `bgp_local_pref` conversion to after filter processing, conditional on iBGP + SetPriority
- Updated compiled templates to match (only 4 scenarios with SetPriority retain the line)
- Added comprehensive tests for the new conditional behavior

### DIFF
--- a/confd/pkg/backends/calico/bgp_processor.go
+++ b/confd/pkg/backends/calico/bgp_processor.go
@@ -738,18 +738,34 @@ func (c *client) buildExportFilter(
 ) string {
 	var filterLines []string
 
-	// Default krt_metric to our normal route priority, if not already set.
+	// Ensure both krt_metric and bgp_local_pref are set for the rest of
+	// the export filter. BGPFilter Priority matching needs bgp_local_pref,
+	// and calico_aggr() needs krt_metric (it checks "krt_metric < 1024"
+	// to let elevated-priority /32s escape aggregation into /26 blocks).
+	//
+	// There are two cases depending on where the route came from:
+	//
+	// Re-exported route (learned via iBGP): this happens when a node acts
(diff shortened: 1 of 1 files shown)


nelljerram:
I think this PR has now been superseded by #12302

nelljerram:
Are you continuing work on this PR?  I thought we agreed to merge #12302 instead.

song-jiang:
> Are you continuing work on this PR? I thought we agreed to merge #12302 instead.

Sorry, my mistake. I'm working on a confd fix and reused the PR branch. I'll create a new branch instead.

PR #12298, opened 2026-03-31
e2e: fix staged policy test cleanup order

lwr20:
## Summary

The three `enforcing staged-policies` tests (`StagedKubernetesNetworkPolicy`, `StagedNetworkPolicy`, `StagedGlobalNetworkPolicy`) were failing in `[AfterEach]` with:

```
Cannot delete a non-empty tier
operation delete is not supported on e2e-staged-tier-XXXXX: Cannot delete a non-empty tier
```

**Root cause:** In Ginkgo v2, `DeferCleanup` registered inside an `It` block runs *after* all `AfterEach` nodes. The tests used `DeferCleanup` to delete the staged and enforced policies, but the outer `AfterEach` deleted the tier first — while those policies were still inside it.

**Fix:** Promote policy variables to context scope and delete them in an inner `AfterEach` per context. Ginkgo runs inner `AfterEach` nodes before outer ones, so the policies are removed from the tier before the tier itself is deleted.

The actual staged policy enforcement was working correctly throughout — the failure was purely in test cleanup.

## Test plan

- [ ] Re-run the `[sig-calico] staged network policy enforcing staged-policies` tests and confirm all three `should enforce a deny policy` specs pass cleanly with no AfterEach errors

### DIFF
--- a/e2e/pkg/tests/policy/staged_policy.go
+++ b/e2e/pkg/tests/policy/staged_policy.go
@@ -247,6 +247,22 @@ var _ = describe.CalicoDescribe(
 			})
 
 			Context("StagedKubernetesNetworkPolicy", func() {
+				var (
+					knpPolicy   *v3.StagedKubernetesNetworkPolicy
+					knpEnforced *networkingv1.NetworkPolicy
+				)
+
+				AfterEach(func() {
+					if knpEnforced != nil {
+						Expect(cli.Delete(context.TODO(), knpEnforced)).ShouldNot(HaveOccurred())
+						knpEnforced = nil
+					}
+					if knpPolicy != nil {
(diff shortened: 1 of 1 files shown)


e2e/pkg/tests/policy/staged_policy.go
```diff
@@ -254,11 +254,11 @@ var _ = describe.CalicoDescribe(
 
 				AfterEach(func() {
 					if knpEnforced != nil {
-						Expect(cli.Delete(context.TODO(), knpEnforced)).ShouldNot(HaveOccurred())
+						Expect(ctrlclient.IgnoreNotFound(cli.Delete(context.TODO(), knpEnforced))).ShouldNot(HaveOccurred())
```
caseydavenport, line 257: I think we probably want to be using DeferCleanup for these after each is created, anyway?

PR #12343, opened 2026-04-03
Kubevirt live migration e2es

skoryk-oleksandr:
## Summary

This PR adds KubeVirt live migration e2e tests covering:

- IP persistence across live migration and workload recreation
- Route programming after migration
- IPAM owner attribute swap after migration completion
- TCP connectivity continuity across migrations (iBGP and eBGP)
- Kubernetes NetworkPolicy enforcement survives migration

### IP persistence tests (`ip_persistency.go`)

| # | Test |
|---|------|
| 1 | IP preservation across live migration |
| 2 | IP preservation across pod recreation (eviction) |
| 3 | IP preservation across VMI recreation (reboot) |
| 4 | IP preservation when source pod is deleted during migration |

### Live migration tests (`live_migration.go`)

| # | Test |
|---|------|
| 1 | IPAM owner attribute swap on migration completion |
| 2 | TCP connection over iBGP survives two consecutive migrations with 0 gaps |
| 3 | TCP connection from eBGP external client across two consecutive migrations with 0 gaps |
| 4 | NetworkPolicy enforcement survives live migration |

### Prerequisites

- KubeVirt cluster with `kubeVirtVMAddressPersistence: Enabled`
- At least 3 schedulable worker nodes
- For eBGP test, an external TOR node with BIRD eBGP peering via L2TP

**Release note:**
```release-note
TBD
```


song-jiang (approved):

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+	"github.com/projectcalico/calico/e2e/pkg/utils"
+	"github.com/projectcalico/calico/e2e/pkg/utils/externalnode"
+	"github.com/projectcalico/calico/e2e/pkg/utils/images"
+	"github.com/projectcalico/calico/libcalico-go/lib/apiconfig"
+	"github.com/projectcalico/calico/libcalico-go/lib/clientv3"
+	cnet "github.com/projectcalico/calico/libcalico-go/lib/net"
+	"github.com/projectcalico/calico/libcalico-go/lib/options"
+)
+
+func init() {
```
caseydavenport: Side effects in package `init()` on every test run are surprising - this fires for any test in this package, not just the eBGP one that needs `EXT_KEY`. We should resolve `EXT_KEY` lazily inside the eBGP test instead.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+	if bzRoot != "" && os.Getenv("EXT_KEY") == "" {
+		keyPath := bzRoot + "/.local/external_key"
+		if _, err := os.Stat(keyPath); err == nil {
+			os.Setenv("EXT_KEY", keyPath)
+		}
+	}
+}
+
+const (
+	defaultVMImage = "mcas/kubevirt-ubuntu-20.04:latest"
```
caseydavenport: We should move this (and `calico/bird:v0.3.3-211-g9111ec3c` below) into `e2e/pkg/utils/images/`, and pin the VM image by digest rather than `:latest` - otherwise our test results depend on whatever happens to be in that registry tag at the time.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+
+	// testTimeout is the per-test context timeout.
+	testTimeout = 10 * time.Minute
```
caseydavenport: 10 min per test is a lot, especially with `WithSerial`. Test 3 chains multiple `Eventually` blocks with 45s/120s budgets across two migrations and can easily eat 8-10 min wall-clock on its own. We should tighten budgets where we can - see also the 3-minute ICMP ping retry in `expectPingSuccess` below, which is hard to defend.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+func (v *testVM) Delete() {
+	logrus.Infof("Cleaning up VM %s/%s", v.namespace, v.name)
+	err := v.kvClient.VirtualMachines(v.namespace).Delete(context.Background(), v.name, metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		Expect(err).NotTo(HaveOccurred())
+	}
+}
+
+// Stop sets RunStrategy to Halted, causing KubeVirt to delete the VMI and pod.
+func (v *testVM) Stop(ctx context.Context) {
```
caseydavenport: `Stop` and `Start` are byte-identical except for the `RunStrategy` constant. We should collapse to one helper that takes the strategy as an arg.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+		if vmim.Status.Phase != kubevirtv1.MigrationSucceeded {
+			return fmt.Errorf("phase is %s", vmim.Status.Phase)
+		}
+		return nil
+	}, 5*time.Minute, 1*time.Second).Should(Succeed())
+}
+
+// expectPingSuccess verifies ICMP connectivity from a pod to the target IP.
+// Retries for up to 3 minutes to allow for route convergence after migration.
+func expectPingSuccess(ns, podName, targetIP string) {
```
caseydavenport: 3 minutes for route convergence seems hard to defend? When do we expect route convergence to take that long?

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+	if attr == nil {
+		return nil, nil, fmt.Errorf("no IPAM assignment found for %s", ipStr)
+	}
+	return attr.ActiveOwnerAttrs, attr.AlternateOwnerAttrs, nil
+}
+
+// checkConnectionToTCPServer verifies that the client pod can connect to the VM's TCP
+// server on port 9999 and receive at least one "seq=" line. Retries for up to 2 minutes
+// to accommodate slow-booting VMs where the TCP server may not yet be listening.
+func checkConnectionToTCPServer(ns, podName, vmIP string) {
```
caseydavenport: Two things: (1) per our e2e helper convention, anything that wraps `Eventually(...).Should(Succeed())` should be `expect*` not `check*`, and should call `GinkgoHelper()` so failures attribute to the caller's line. Same applies to `checkTCPConnectionBlocked` below. (2) See the broader conncheck point on `expectPingSuccess` above.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+		}
+		return nil
+	}, 10*time.Second, 2*time.Second).Should(Succeed(),
+		"TCP connection to %s:9999 should be blocked by policy", vmIP)
+}
+
+// tcpStreamLineCount returns the number of lines in the given file inside the given pod.
+// Uses `wc -l < <file>` so the output is a bare integer (no filename suffix) and returns
+// a descriptive error if the kubectl exec fails or the output cannot be parsed.
+func tcpStreamLineCount(ns, podName, streamFile string) (int, error) {
```
caseydavenport: These three `*LineCount` helpers (this one, `torStreamLineCount`, and `torContainerLineCount` below) are near-identical `wc -l` wrappers + `Sscanf`. We should refactor to one `lineCount(execFn)` that takes an exec-shaped closure so the parsing isn't duplicated three times.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+		}
+	}
+	return
+}
+
+// runOnTOR executes a shell command on the external TOR node via SSH and returns stdout.
+// Logs a warning on error but does not fail the test — use this for fire-and-forget
+// commands (process spawn, cleanup). For commands whose result drives an assertion, use
+// runOnTORE so SSH failures are surfaced rather than silently producing an empty string.
+func runOnTOR(tor *externalnode.Client, cmd string) string {
```
caseydavenport: `runOnTOR` swallows SSH errors with a `logrus.Warnf` and callers downstream operate on empty strings as if the call succeeded - that's a foot-gun. We should propagate the error, or at least make callers explicitly check for empty. And `runOnTORE` then duplicates the same logic with the error returned - one function returning `(string, error)` is enough; let callers ignore if they really want.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+func startBirdOnTOR(tor *externalnode.Client, torIP string, peersConf string) {
+	By("Starting BIRD container on TOR")
+
+	// Remove any prior container.
+	runOnTOR(tor, "sudo docker rm -f tor-bird 2>/dev/null || true")
+
+	// Start the container with host networking. The calico/bird image ships with
+	// a base bird.conf that defines router id, protocol kernel, and protocol device.
+	runOnTOR(tor, "sudo docker run -d --privileged --name tor-bird --network host "+
+		"calico/bird:v0.3.3-211-g9111ec3c")
```
caseydavenport: Hardcoded BIRD image version in test code - should go in `e2e/pkg/utils/images/` so we have one place to update it.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+	// The controller-runtime client fails resource discovery when the Calico API
+	// server aggregated endpoint (projectcalico.org/v3) is registered but not running.
+	lcgc := newLibcalicoClient(f)
+	ctx := context.Background()
+
+	// Collect node BGP IPs from the projectcalico.org/IPv4Address annotation.
+	nodeList, err := f.ClientSet.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	Expect(err).NotTo(HaveOccurred(), "failed to list nodes")
+
+	bgpSubnet := &net.IPNet{
```
caseydavenport: `172.16.8.0/24` (here) and the pod CIDR `192.168.0.0/16` baked into the BIRD filter generated below are Banzai-specific. On any cluster with a different layout this will silently filter out the real pod CIDR and the test will fail in confusing ways. We should discover both from `Installation`/`IPPool`.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+
+	// Keep the BGP mesh enabled — nodes need iBGP for inter-node routing
+	// (required for KubeVirt live migration). The eBGP peer to the TOR is
+	// additive: the master advertises routes to the TOR via eBGP while all
+	// nodes continue to exchange routes with each other via the iBGP mesh.
+
+	// Debug: check master's BIRD config and routes after BGPPeer creation.
+	// This helps diagnose SSH breakage by showing what confd generated.
+	By("Debugging: checking master BIRD config after BGPPeer creation")
+	time.Sleep(3 * time.Second) // Give confd time to regenerate config
```
caseydavenport: We don't want bare `time.Sleep` in e2e. Replace with an `Eventually` that polls for the actual confd-generated state (e.g. `bird.cfg` containing the new peer block). The rest of the file does `Eventually` correctly - this one's an outlier.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+
+		// Show BIRD protocol status.
+		birdProto, _ := kubectl.NewKubectlCommand("calico-system", "exec", masterPodName, "-c", "calico-node", "--",
+			"birdcl", "show", "protocols").Exec()
+		logrus.Infof("DEBUG: Master BIRD protocols:\n%s", birdProto)
+
+		// Show BIRD routing table.
+		birdRoutes, _ := kubectl.NewKubectlCommand("calico-system", "exec", masterPodName, "-c", "calico-node", "--",
+			"birdcl", "show", "route", "export", "Node_172_16_8_5").Exec()
+		logrus.Infof("DEBUG: Master BIRD routes exported to TOR peer:\n%s", birdRoutes)
```
caseydavenport: This whole `DEBUG:` block (~lines 822-861) reads like development scaffolding rather than steady-state diagnostics - dumping `bird.cfg`, `bird_ipam.cfg`, `bird_aggr.cfg`, full BIRD routes, exported routes, and protocols on every test run is a lot of noise. We should either delete it or wrap it so it only fires on failure. Same critique applies to the per-route `logrus.Infof` calls inside `queryTORRoute` further down - that's noise on every poll.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+type torRouteSnapshot struct {
+	Host32    torPrefixState `json:"host32"`
+	Block26   torPrefixState `json:"block26"`
+	KernelVia string         `json:"kernelVia"`
+}
+
+// parseBIRDRouteOutput parses the output of "birdcl show route <prefix> all"
+// and returns the list of routes. Returns nil if the output contains
+// "Network not in table" or is empty.
+func parseBIRDRouteOutput(output string) []torBIRDRoute {
```
caseydavenport, line 805: The `fmt.Sscanf` returns are silently discarded here (and in `queryWorkerMetric` below). On parse failure we end up with zero values that then drive assertions - which is going to make a real bug look like "no `BGP.local_pref` line in output" rather than "parse failed". We should at least log the parse error.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+			LabelSelector: "k8s-app=calico-typha",
+		})
+		if err == nil && len(pods.Items) > 0 {
+			typhaPods = pods.Items
+			typhaNamespace = ns
+			break
+		}
+	}
+	if len(typhaPods) == 0 {
+		Skip("Typha not deployed; live-migration tests require Typha")
```
caseydavenport: We don't want self-skips - they hide configuration mistakes in CI. The YAML exclusion via `Feature:KubeVirt` already controls whether the test runs at all, so once selected, missing prereqs is a real failure that should `Fail` with a clear message, not silently `Skip`. Same applies to the `Skip` for missing `EXT_IP/EXT_KEY/EXT_USER` in `live_migration.go`.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+		// when the VM migrates, the elevated-priority route advertisement from the target
+		// node reaches the TOR via eBGP, causing the TOR's kernel routing table to switch
+		// to the new next-hop without dropping the TCP connection. Two consecutive migrations
+		// verify the route priority reverts correctly after the TimeWait period and can be
+		// re-elevated for the second migration.
+		// Requires EXT_IP, EXT_KEY, EXT_USER env vars pointing to the TOR/external node.
+		It("should maintain TCP connection from eBGP external client across two consecutive migrations", func() {
+			tor := externalnode.NewClient()
+			if tor == nil {
+				Skip("External node not configured (set EXT_IP, EXT_KEY, EXT_USER)")
```
caseydavenport: e2e tests should never self-skip. Should `Fail` with a clear message instead, and we update the test to include the `RequiresExternalNode` label so callers can skip.

e2e/pkg/tests/kubevirt/ip_persistency.go
```diff
@@ -0,0 +1,276 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			originalVMIUID := vmi.UID
+			logrus.Infof("Original VMI UID: %s, IP: %s", originalVMIUID, originalIP)
+
+			By("Stopping the VM")
+			vm.Stop(ctx)
+
+			By("Waiting for VMI to be deleted")
+			// Check for specific NotFound error, not any error.
+			Eventually(func() bool {
```
caseydavenport: `Eventually(func() bool {...}).Should(BeTrue())` - we should return an `error` with diagnostic context instead, so a timeout failure is informative. As written, when this times out the message is just "Expected <bool>: false to be true" and you have to go re-read the test to figure out what was being checked.

e2e/config/base.yaml
```diff
@@ -17,3 +17,6 @@ exclude:
 
     - label: ExternalNode
       reason: "requires a non-cluster node with EXT_IP/EXT_KEY/EXT_USER configured"
+
+    - label: Feature:KubeVirt
+      reason: "e2e infra integration with Banzai not ready, currently run manually"
```
caseydavenport, line 22: Longer term, do we think we're going to be able to run this as part of repo CI? What's left on the bz side?

e2e/pkg/tests/kubevirt/ip_persistence.go
```diff
@@ -0,0 +1,276 @@
+// Copyright (c) 2026 Tigera, Inc. All rights reserved.
```
caseydavenport, line 1: `ip_persistence.go`

e2e/pkg/tests/kubevirt/ip_persistency.go
```diff
@@ -0,0 +1,276 @@
...
+		f := utils.NewDefaultFramework("calico-kubevirt-ip")
+
+		var kvClient kubevirtcorev1.KubevirtV1Interface
+
+		BeforeEach(func() {
+			// Live migration needs at least 2 nodes to migrate between.
+			utils.RequireNodeCount(f, 2)
+
+			var err error
+			kvClient, err = kubevirtcorev1.NewForConfig(f.ClientConfig())
```
caseydavenport: We should register kubevirtcorev1 with the scheme used by our controller-runtime client in this directory instead of instantiating a dedicated typed client. 

See here: https://github.com/caseydavenport/calico/blob/5f1fa4124959d26bdbb1280b1b43dae53b8f57a6/e2e/pkg/utils/client/client.go#L86-L112

e2e/pkg/tests/kubevirt/ip_persistency.go
```diff
@@ -0,0 +1,276 @@
...
+			defer cancel()
+			ns := f.Namespace.Name
+			vmName := "e2e-migration-test"
+			vm := &testVM{name: vmName, namespace: ns, kvClient: kvClient}
+
+			pingPod := setupPingPod(ctx, f, ns)
+			vm.Create(ctx)
+			DeferCleanup(vm.Delete)
+
+			originalIP, sourceNode := vm.WaitForRunningWithIP(ctx)
```
caseydavenport: We have a standard of putting connection checks into the conncheck library - I think we can and should use that here as well.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			By("Creating client pod on a different node than server VM")
+			clientPod := setupAntiAffinityPod(ctx, f, ns, node1)
+			checkConnectionToTCPServer(ns, clientPod.Name, serverIP)
+
+			// Use nohup to prevent SIGHUP when kubectl exec session closes. The Exec()
+			// return is intentionally ignored: `nohup ... &` backgrounds immediately and
+			// always returns success, so its error value tells us nothing about whether
+			// nc actually connected. We verify liveness explicitly below.
+			By("Starting TCP client")
+			_, _ = kubectl.NewKubectlCommand(ns, "exec", clientPod.Name, "--",
```
caseydavenport: It would require some additions to the conncheck framework, but we should still move this there. 

We really want to avoid one-off / bespoke connection tooling in the tests.

Should be able to enhance the conncheck tooling with either a new Target type / options, or a new ExpectX method that allows for this.

caseydavenport: At a minimum we can / should use the conncheck tooling to manage the pod lifecycles - the tool gives you access to the underlying v1.Pod object so you can still interact with it.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			migration2 := &testVMIM{name: serverVMName + "-migration2", namespace: ns, vmiName: serverVMName, kvClient: kvClient}
+			migration2.Create(ctx)
+			DeferCleanup(migration2.Delete)
+			migration2.WaitForSuccess(ctx)
+			vmi, err = kvClient.VirtualMachineInstances(ns).Get(ctx, serverVMName, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			node3 := vmi.Status.NodeName
+			// With 3 worker nodes, second migration moves away from node2.
+			// It could return to node1 — that's fine, we only require it left node2.
+			Expect(node3).NotTo(Equal(node2))
```
caseydavenport:
```suggestion
			Expect(node3).NotTo(Equal(node2), "VM didn't migrate off node 2")
```

We should make sure we're giving good context on each expectation.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			// Also require the server lands on a different node than the
+			// client pod. Otherwise the post-migration "iBGP cross-node"
+			// claim degrades to a same-node loopback path, which would
+			// trivially have zero seq gaps regardless of routing behaviour.
+			Expect(node3).NotTo(Equal(clientPod.Spec.NodeName),
+				"server must end up on a different node from the client to actually exercise the iBGP path")
```
caseydavenport:
```suggestion
				"server was scheduled to the same node as the client")
```

^ Describe the check, not why we're making it.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			var midLines int
+			Eventually(func() (int, error) {
+				var err error
+				midLines, err = tcpStreamLineCount(ns, clientPod.Name, "/tmp/tcp_stream")
+				return midLines, err
+			}, 30*time.Second, 2*time.Second).Should(BeNumerically(">=", preLines+5),
+				"TCP data should have grown after first migration")
+			logrus.Infof("After first migration: %d lines", midLines)
+
+			By("Second migration")
```
caseydavenport:
```suggestion
			By("Migrating the VM a second time")
```

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			var preLines int
+			Eventually(func() (int, error) {
+				var err error
+				preLines, err = tcpStreamLineCount(ns, clientPod.Name, "/tmp/tcp_stream")
+				return preLines, err
+			}, 30*time.Second, 2*time.Second).Should(BeNumerically(">=", 5),
+				"TCP data should be flowing")
+			logrus.Infof("Pre-migration: %d lines on client pod", preLines)
+
+			By("First migration")
```
caseydavenport:
```suggestion
			By("Migrating the VM")
```

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			migration1 := &testVMIM{name: serverVMName + "-migration1", namespace: ns, vmiName: serverVMName, kvClient: kvClient}
+			migration1.Create(ctx)
```
caseydavenport: Might just be me, but I find this phrasing pretty unorthodox and hard to follow. 

Any reason not to use a more standard approach?

```go
vmim1 := migration.NewVMIMigration(...)
cli.Create(vmim1)
```

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			// leaves the cluster with natOutgoing in the wrong state.
+			//
+			// Register the re-enable DeferCleanup BEFORE the disable patch so a panic
+			// in the propagation wait below cannot leave the cluster with natOutgoing
+			// permanently disabled. DeferCleanups run in LIFO after the test body, so
+			// the re-enable still runs after every test action.
+			const vmIPPoolName = "default-ipv4-ippool"
+			DeferCleanup(func() {
+				By("Re-enabling natOutgoing via Installation")
+				patchInstallationPoolNATOutgoing(vmIPPoolName, "Enabled")
```
caseydavenport: We have a `utils.ConfigureWithCleanup` that we should use whenever we're changing global shared state that needs to be reverted at the end of the test. We do this elsewhere for BGPConfiguration and FelixConfiguration, as an example.

caseydavenport: Also, if this test relies on the operator, we should have a BeforeEach() fail-fast assertion if this isn't an operator-managed cluster.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			})
```
caseydavenport:
```suggestion
			})
			
```

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			By("Disabling natOutgoing via Installation to prevent masquerade breaking TCP after migration")
+			patchInstallationPoolNATOutgoing(vmIPPoolName, "Disabled")
```
caseydavenport:
```suggestion
			patchInstallationPoolNATOutgoing(vmIPPoolName, "Disabled")
			
```

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			// Wait for operator to reconcile the IPPool and Felix to drain the masq ipset.
+			// When natOutgoing is disabled, the field is omitted from the IPPool spec
+			// (jsonpath returns empty string rather than "false").
+			logrus.Info("Waiting for natOutgoing=Disabled to propagate...")
+			Eventually(func() string {
+				out, _ := kubectl.NewKubectlCommand("", "get", "ippool", vmIPPoolName,
```
caseydavenport: We shouldn't be using kubectl for this - we have a controller-runtime client that tests should use to interact with the k8s API.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+					return fmt.Errorf("TCP server not sending data from TOR (output=%q)", out)
+				}
+				return nil
+			}, 2*time.Minute, 5*time.Second).Should(Succeed(),
+				"TCP server not reachable from TOR")
+			logrus.Infof("TCP server on %s:9999 is reachable and sending data from TOR", vmIP)
+
+			By("Starting TCP client container on TOR connecting to VM")
+			runOnTOR(tor, fmt.Sprintf("sudo docker rm -f %s 2>/dev/null || true", ncClientContainer))
+			runOnTOR(tor, fmt.Sprintf(
```
caseydavenport: We have an external node utility package - I think it already has helpers for running things on external nodes?

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+				if !strings.Contains(out, "seq=") {
+					return fmt.Errorf("TCP server not sending data from TOR (output=%q)", out)
+				}
+				return nil
+			}, 2*time.Minute, 5*time.Second).Should(Succeed(),
+				"TCP server not reachable from TOR")
+			logrus.Infof("TCP server on %s:9999 is reachable and sending data from TOR", vmIP)
+
+			By("Starting TCP client container on TOR connecting to VM")
+			runOnTOR(tor, fmt.Sprintf("sudo docker rm -f %s 2>/dev/null || true", ncClientContainer))
```
caseydavenport: IMO instead of `|| true` this should return an err and we can explicitly check / skip checking the error.

e2e/pkg/tests/kubevirt/live_migration.go
caseydavenport: Either way, feels like this whole chunk is mostly boilerplate / setup logic that could be extracted (cleaning, starting, waiting for running) instead of embedding it in the test - makes the test much shorter and easier to read.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			timeline.record(routeTimelineEntry{
+				Phase:    "pre-migration",
+				VMNode:   node1,
+				TOR:      queryTORSnapshot(tor, vmIP),
+				TCPLines: preLines,
+			})
+
+			// ---- First migration ----
+			By("First migration")
+			logrus.Infof("TIMELINE: starting first migration")
```
caseydavenport: Do we need both the By() and the log? 

The point of `By()` statements is to show the steps the test is taking and they all show up in the log as `STEP: msg`, so I think we can drop the logrus call here.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+			preTime := time.Now()
+			timeline.record(routeTimelineEntry{
+				Phase:    "pre-migration",
+				VMNode:   node1,
+				TOR:      queryTORSnapshot(tor, vmIP),
+				TCPLines: preLines,
+			})
+
+			// ---- First migration ----
+			By("First migration")
```
caseydavenport:
```suggestion
			By("Starting first migration")
```

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+				Phase:    "convergence-done",
+				VMNode:   node2,
+				TOR:      queryTORSnapshot(tor, vmIP),
+				TCPLines: convergenceLines,
+			})
+
+			midLines, _ := torContainerLineCount(tor, ncClientContainer)
+			logrus.Infof("TIMELINE: after first migration wait lines=%d (pre=%d, delta=%d)", midLines, preLines, midLines-preLines)
+			Expect(midLines).To(BeNumerically(">", preLines), "TCP should still be flowing after first migration")
+			// Cordon node1 (the original node) so the second migration cannot
```
caseydavenport: Is there no way to specify a destination node as part of the test?

Cordoning node 1 is a pretty disruptive action we're taking on the cluster and scares me a bit - especially if the cleanup fails in some way.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,817 @@
...
+
+			logrus.Infof("TIMELINE: test complete — analyse route-monitor and seq-monitor logs above")
+		})
+
+		// Test 4: Kubernetes NetworkPolicy enforcement survives live migration.
+		// After migration, KubeVirt creates a new virt-launcher pod on the target node.
+		// Network policies that selected the old pod must also apply to the new pod.
+		// This test verifies that an allowed client can still connect and a denied
+		// client is still blocked after the VM migrates to a different node.
+		It("should enforce NetworkPolicy after live migration", func() {
```
caseydavenport, line 476: Is this test really necessary? Do we have any reason network policy won't continue to work? Is there some end-to-end level thing we're testing here? Seems like we already have migration tests, we already have NP tests, we don't necessarily need the cross product?

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+	"github.com/projectcalico/calico/e2e/pkg/utils/externalnode"
+	"github.com/projectcalico/calico/e2e/pkg/utils/images"
+	"github.com/projectcalico/calico/libcalico-go/lib/apiconfig"
+	"github.com/projectcalico/calico/libcalico-go/lib/clientv3"
+	cnet "github.com/projectcalico/calico/libcalico-go/lib/net"
+	"github.com/projectcalico/calico/libcalico-go/lib/options"
+)
+
+func init() {
+	// Auto-resolve EXT_KEY from BZ_ROOT_DIR if not explicitly set.
```
caseydavenport: This is a bit of a smell - our tests really shouldn't be assuming anything about BZ, and this should all be done via the config package default setting.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+const (
+	defaultVMImage = "mcas/kubevirt-ubuntu-20.04:latest"
+
+	// testTimeout is the per-test context timeout.
+	testTimeout = 10 * time.Minute
+)
+
+// vmImage returns the container disk image for test VMs. Configurable via the
+// KUBEVIRT_TEST_VM_IMAGE env var; defaults to defaultVMImage.
+func vmImage() string {
```
caseydavenport, line 64: Should be done via the config package as well - no secret config buried in per-test utility packages.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+	if img := os.Getenv("KUBEVIRT_TEST_VM_IMAGE"); img != "" {
+		return img
+	}
+	return defaultVMImage
+}
+
+const defaultCloudInit = "#cloud-config\npassword: testpass\nchpasswd: { expire: False }\nssh_pwauth: True\n"
+
+const tcpServerCloudInit = `#cloud-config
```
caseydavenport, line 78: A bit of documentation here would be good - what's the point of this init script?

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+password: testpass
+chpasswd: { expire: False }
+ssh_pwauth: True
+write_files:
+  - path: /usr/local/bin/tcp-server.py
```
caseydavenport, line 83: Where does this come from?

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+      s.listen(5)
+      while True:
+          conn, addr = s.accept()
+          threading.Thread(target=handle, args=(conn, addr), daemon=True).start()
+runcmd:
+  - [bash, -c, "nohup python3 /usr/local/bin/tcp-server.py &"]
+`
+
+// testVM encapsulates a KubeVirt VirtualMachine for e2e tests.
+type testVM struct {
```
caseydavenport:
```suggestion
type kubeVirtVM struct {
```

caseydavenport: ^ Self-describing name.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+						}}},
+					},
+				},
+			},
+		},
+	}
+}
+
+// Create creates the VirtualMachine in the cluster.
+func (v *testVM) Create(ctx context.Context) {
```
caseydavenport: Yeah, no need to bind these to the VM type IMO - the code is more clear when a VM doesn't reference a client that is then used to create itself - keep the concepts more distinct and clear.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+		if err != nil {
+			return fmt.Errorf("update VM: %w", err)
+		}
+		return nil
+	}, 1*time.Minute, 5*time.Second).Should(Succeed(), "failed to start VM %s", v.name)
+}
+
+// WaitForRunningWithIP polls the VMI until it reaches Running phase and has an IP
+// address assigned. Returns the IP and the node where the VMI is scheduled.
+func (v *testVM) WaitForRunningWithIP(ctx context.Context) (ip, node string) {
```
caseydavenport: I do still think we could fold this into conncheck as a new client / server type and implement the interfaces within.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+		}
+	}
+	return nil, fmt.Errorf("no running virt-launcher pod found for VM %s (total pods: %d)", v.name, len(pods.Items))
+}
+
+// setupPingPod creates a long-running Alpine pod for connectivity checks (ping, nc).
+// The pod is scheduled on any Linux node and cleaned up after the test.
+func setupPingPod(ctx context.Context, f *framework.Framework, ns string) *corev1.Pod {
+	By("Creating a ping pod for connectivity checks")
+	pod := &corev1.Pod{
```
caseydavenport: Definitely should be conncheck.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+	Expect(err).NotTo(HaveOccurred(), "pod %s not Running", created.Name)
+	return created
+}
+
+// setupAntiAffinityPod creates a long-running Alpine pod with a node anti-affinity rule
+// that prevents scheduling on avoidNode. Used for TCP tests where the client must be on
+// a different node than the server VM to exercise cross-node BGP routing.
+func setupAntiAffinityPod(ctx context.Context, f *framework.Framework, ns, avoidNode string) *corev1.Pod {
+	By(fmt.Sprintf("Creating client pod avoiding node %s", avoidNode))
+	pod := &corev1.Pod{
```
caseydavenport: Also conncheck - there's a customizer that can be used to set affinities.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+		}
+		return nil
+	}, 2*time.Minute, 5*time.Second).Should(Succeed(), "TCP server not ready on VM %s", vmIP)
+	logrus.Infof("TCP server ready on %s:9999", vmIP)
+}
+
+// checkTCPConnectionBlocked verifies that the client pod cannot connect to the VM's TCP
+// server on port 9999. Uses Consistently to confirm that nc never receives "seq=" data
+// over the check window, indicating the connection is blocked by network policy.
+func checkTCPConnectionBlocked(ns, podName, vmIP string) {
```
caseydavenport: Same check -> expect naming convention

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1387 @@
...
+  add paths on;
+  graceful restart;
+  connect delay time 2;
+  connect retry time 5;
+  error wait time 5,30;
+}
+
+`)
+	for i, nodeIP := range nodeIPs {
+		sb.WriteString(fmt.Sprintf(`protocol bgp node_%d from bgp_template {
```
caseydavenport: This is a bit hard to parse / easy to mess up in the future due to formatting large inline string segments. maybe better to define the templates as their own variable so we can reference that?

e2e/pkg/tests/kubevirt/utils.go
caseydavenport: We also have a BIRD_VERSION in metadata.mk that probably could get plumbed through here?

caseydavenport (changes requested):
OK - first pass. Lots of comments but most / many are nits. Just a few more structural requests. Let's chat about any of it if I got anything wrong!

e2e/pkg/tests/kubevirt/live_migration.go
song-jiang: This test verifies that end-to-end assumption: that label propagation from `vm.spec.template.metadata.labels` to the virt-launcher pod works correctly across a live migration, and that the NetworkPolicy remains enforced throughout. If KubeVirt somehow changes its label propagation behaviour, this test will catch the regression and give us an early signal to find a solution.

e2e/config/base.yaml
skoryk-oleksandr: Yes, it is being implemented within this ticket : https://tigera.atlassian.net/browse/CORE-12051

e2e/pkg/tests/kubevirt/utils.go
skoryk-oleksandr: it's inlined in the cloud-init write_files block right here in utils.go

e2e/pkg/tests/kubevirt/utils.go
skoryk-oleksandr: Both Sscanf calls now logrus.Warnf on parse failure

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,701 @@
...
+// The live-migration tests mutate cluster-global state — the Installation
+// resource (natOutgoing), global BGPPeer/BGPFilter resources with fixed names
+// (kubevirt-lm, tor-ebgp-peer), and BIRD configuration on the TOR. They must
+// not run in parallel with each other or with other tests that touch Calico
+// configuration.
+var _ = describe.CalicoDescribe(
+	describe.WithTeam(describe.Core),
+	describe.WithFeature("KubeVirt"),
+	describe.WithCategory(describe.Networking),
+	describe.WithSerial(),
```
caseydavenport: The `WithSerial()` justification (Installation patch, fixed-name BGPPeer/BGPFilter, TOR BIRD state) only applies to Test 3. Tests 1, 2, and 4 create per-namespace VMs and could safely run in parallel. I think we should move `WithSerial()` onto the eBGP `Context` only - CI will thank us.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -0,0 +1,701 @@
...
+		})
+
+		// Test 1: active IPAM owner promotes from source to target after migration.
+		// Asserts Active=target,Alternate=empty (Felix EnsureActiveVMOwnerAttrs) by
+		// matching the real target pod/node, not just "differs from source".
+		It("should promote target pod to active IPAM owner after migration", func() {
+			ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+			defer cancel()
+			ns := f.Namespace.Name
+			vmName := "e2e-attr-promote"
```
caseydavenport, line 87: VM names here and elsewhere (`e2e-attr-promote`, `e2e-tcp-double-srv`, `e2e-netpol-vm`, `e2e-ebgp-tcp`, etc.) are hardcoded. Namespace-scoped so it works under serial, but if we drop the blanket `WithSerial()` per the comment above we'll want `utils.GenerateRandomName`. Same applies to the cluster-scoped `kubevirt-lm` BGPFilter and `tor-ebgp-peer` BGPPeer in `setupEBGPPeering` - those are collision-prone across reruns where cleanup fails.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1204 @@
...
+	GinkgoHelper()
+	By(fmt.Sprintf("Waiting for TCP server on %s:9999", vmIP))
+	Eventually(func() error {
+		output, _ := kubectl.NewKubectlCommand(ns, "exec", podName, "--",
+			"sh", "-c", fmt.Sprintf("timeout 5 nc %s 9999", vmIP)).Exec()
+		if !strings.Contains(output, "seq=") {
+			return fmt.Errorf("TCP server not ready (no seq= in output)")
+		}
+		return nil
+	}, 2*time.Minute, 5*time.Second).Should(Succeed(), "TCP server not ready on VM %s", vmIP)
```
caseydavenport: Same family of feedback as before: we have `conncheck.StreamProbe` now, so these kubectl-exec-`nc` helpers could become a TCP-stream target/assertion inside conncheck rather than a parallel path. Not blocking, but it would close out the last bit of bespoke connection tooling in this file.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1204 @@
...
+//
+// The kubectl exec error is intentionally ignored: `timeout 5 nc` always exits
+// 143 (SIGTERM from `timeout`) once the budget elapses, even when the
+// connection succeeded and stdout contains the expected seq= lines. Checking
+// stdout is the only honest signal here.
+func expectConnectionToTCPServer(ns, podName, vmIP string) {
+	GinkgoHelper()
+	By(fmt.Sprintf("Waiting for TCP server on %s:9999", vmIP))
+	Eventually(func() error {
+		output, _ := kubectl.NewKubectlCommand(ns, "exec", podName, "--",
```
caseydavenport: The `_` discard is exactly what the e2e skill (P23) calls out. The comment explains why (`timeout 5 nc` always exits 143), but I'd rather see us check `err` against the expected exit status, or at least name it (`_, _ = ...`) with a louder inline comment so it doesn't look like an oversight.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1204 @@
...
+	"github.com/projectcalico/calico/libcalico-go/lib/apiconfig"
+	"github.com/projectcalico/calico/libcalico-go/lib/clientv3"
+	cnet "github.com/projectcalico/calico/libcalico-go/lib/net"
+	"github.com/projectcalico/calico/libcalico-go/lib/options"
+)
+
+const (
+	// testTimeout caps each spec at 6m. The longest test (eBGP double-migration)
+	// finishes in ~5m on healthy clusters.
+	testTimeout = 6 * time.Minute
```
caseydavenport: 6m is fine as a cap, but Tests 1 and 2 realistically run in ~3m. Per-test budgets surface regressions sooner - I think we should tighten each spec to what it actually needs and only keep the 6m budget on the eBGP double-migration test.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1204 @@
...
+	logrus.Infof("queryWorkerMetric(%s, %s): no metric field (route: %s)", nodeName, ip, out)
+	return -1
+}
+
+// routeTimelineEntry captures the route state at a single point in time.
+type routeTimelineEntry struct {
+	Timestamp string           `json:"ts"`
+	Phase     string           `json:"phase"`
+	VMNode    string           `json:"vmNode,omitempty"`
+	TOR       torRouteSnapshot `json:"tor"`
```
caseydavenport: `routeTimeline`, `startRouteMonitor`, the `seq-monitor` goroutine, and `writeToTOR` of the JSON timeline read like instrumentation that was needed to chase a bug, not steady-state diagnostics. Now that the test is green I think we should retire what we can and gate the rest behind `CurrentSpecReport().Failed()` like `logBIRDDiagnostics` is - otherwise this is going to rot.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1204 @@
...
+	_, err = runOnTOR(tor, cmd)
+	if err != nil {
+		logrus.WithError(err).Warn("Route timeline: failed to write JSON to TOR")
+		return
+	}
+	logrus.Infof("Route timeline: wrote %d entries (%d bytes) to TOR:/tmp/route-timeline.json",
+		len(entries), len(data))
+}
+
+func ptrInt64(v int64) *int64 { return &v }
```
caseydavenport: Use `ptr.To(int64(0))` from `k8s.io/utils/ptr` instead of rolling our own - it's the upstream idiom and we already pull it in elsewhere.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1204 @@
...
+	}
+	Fail(fmt.Sprintf("no third worker node found (node1=%s, node2=%s); need at least 3 schedulable workers for the double-migration eBGP test", node1, node2))
+	return ""
+}
+
+// requireOperatorManagedCluster fails the calling test if the cluster does not
+// have an operator-managed Installation/default. The live-migration tests patch
+// the Installation CR to disable natOutgoing; without an Installation that path
+// is invalid (manifest installs configure IPPool directly).
+func requireOperatorManagedCluster(ctx context.Context, cli ctrlclient.Client) {
```
caseydavenport: We already have `utils.GetInstallation(cli)` in `e2e/pkg/utils/installation.go` which returns nil when the Installation isn't there. Reuse it: `if utils.GetInstallation(cli) == nil { Fail(...) }`.

e2e/pkg/tests/kubevirt/utils.go
```diff
@@ -0,0 +1,1204 @@
...
+}
+
+// waitForMasterCalicoNodePod polls until the calico-node pod on the
+// control-plane node is observable. Returns the pod name. Used as a
+// precondition for any test step that needs to exec birdcl on the master.
+func waitForMasterCalicoNodePod(ctx context.Context, f *framework.Framework, masterName string) string {
+	GinkgoHelper()
+	var name string
+	Eventually(func() error {
+		pods, err := f.ClientSet.CoreV1().Pods("calico-system").List(ctx, metav1.ListOptions{
```
caseydavenport: Two existing helpers we should be using here and in similar spots below (lines 1029, 714, 782, 789): `utils.CalicoNamespace(f)` instead of hardcoding `"calico-system"` (won't survive on a `kube-system` install), `utils.GetCalicoNodePodOnNode(clientset, nodeName)` for the pod lookup, and `utils.ExecInCalicoNode(pod, cmd)` for the exec. Same applies to `queryWorkerMetric` at L1026.

caseydavenport (commented):
Some more thoughts

caseydavenport (approved):

skoryk-oleksandr:
The CI failure on the gcp-bpf pipeline is not caused by changes in this PR.

The failing Tiered-RBAC tests (in particular `default tier should not be deletable`) fail because the gcp-bpf pipeline runs with `USE_API_SERVER=false` (v3 CRD mode), and in that mode the only thing that is supposed to enforce built-in tier deletion is a ValidatingAdmissionPolicy. That VAP was sitting under `api/config/admission/` and was never bundled into any chart or operator manifest, so nothing was actually preventing deletion of the default tier. The CEL expressions also read `object.metadata.name`, which is null on DELETE requests.

Casey is fixing the real bug in #12824 (wires the VAP into the chart, fixes the CEL expressions, and adds the matching CI exclude). Once that lands, these tests should pass on this pipeline too.

PR #12353, opened 2026-04-06
fix(bpf): enforce policy on established ambient mesh connections

electricjesus:
## Summary

- Fix BPF fast-path (`CALI_CT_ESTABLISHED_BYPASS`) bypassing policy re-evaluation for Istio ambient mesh (ztunnel) connections
- Disable CTLB in Istio ambient e2e tests (incompatible with ambient mode)
- Temporary workaround for calico-system tier policies blocking cross-node ztunnel<->istiod in BPF mode without CTLB

## Problem

When Istio ambient mode is enabled on a CalicoBPF cluster, ztunnel maintains persistent HBONE connections to workload pods. These get both `a_to_b.approved` and `b_to_a.approved` set in conntrack, resulting in `CALI_CT_ESTABLISHED_BYPASS`. All subsequent packets skip policy — so applying a new deny policy has no effect on existing ztunnel connections.

## Approach

Add `CALI_CT_FLAG_AMBIENT` to mark conntrack entries for ambient mesh workloads (detected via `ISTIO_DSCP` on the interface). For connections with this flag, the conntrack lookup suppresses the bypass (returns `ESTABLISHED` instead of `BYPASS`), forcing per-packet policy re-evaluation on both ingress (TO_WEP) and egress (FROM_WEP) paths. This ensures policy changes take immediate effect on ztunnel's persistent connections.

On the TO_WEP path, the `tc.c` change handles ambient detection and routes ambient ESTABLISHED connections through `syn_force_policy` for full policy evaluation. FROM_WEP ambient connections also hit policy because the bypass is suppressed in conntrack.

The `daemon.go` workaround creates an allow-all override policy for ztunnel+istiod when BPF+ambient+CTLB-disabled is detected, working around operator-managed policies that don't resolve correctly cross-node without CTLB.

### Known limitation: egress policy on OSS ambient

Egress policies with destination selectors (e.g., `app == 'allowed-svc'`) cannot distinguish between HBONE-tunneled destinations on open-source Calico. All ambient egress goes through ztunnel on port 15008 — BPF sees the HBONE tunnel, not the original destination. This is a policy model limitation with upstream Istio ambient, not a bug in the fix.

## Test plan

- [x] BPF unit test `TestIstioAmbientPolicyEnforcementV4` — ambient deny enforced + non-ambient fast-path preserved
- [x] Existing `TestIstioDSCPv4` passes (no regression)
- [x] Manual e2e on CalicoBPF cluster: deploy Istio ambient, apply ingress deny, verify blocked, remove, verify resumes
- [x] Manual e2e: selective ingress deny (one service blocked, another unaffected)
- [ ] Istio ambient e2e tests pass

**Release note:**
```release-note
ebpf: Fix dataplane to enforce network policy changes on established Istio ambient mesh (ztunnel) connections. Previously, policy updates had no effect on existing ztunnel connections due to the conntrack fast-path bypass.
```

### DIFF
--- a/felix/bpf-gpl/types.h
+++ b/felix/bpf-gpl/types.h
@@ -171,6 +171,8 @@ enum cali_state_flags {
 	CALI_ST_SET_DSCP   = 0x2000,
 	/* CALI_ST_FIRST_FRAG is set if this packet is the first fragment of a fragmented IP packet */
 	CALI_ST_FIRST_FRAG        = 0x4000,
+	/* CALI_ST_AMBIENT is set when the packet is on an ambient mesh workload interface. */
+	CALI_ST_AMBIENT           = 0x8000,
 };
 
 struct cali_tc_ctx {
--- a/e2e/pkg/tests/istio/istio.go
+++ b/e2e/pkg/tests/istio/istio.go
@@ -71,6 +71,30 @@ var _ = describe.CalicoDescribe(
 			var err error
 			cli, err = client.New(f.ClientConfig())
 			gomega.Expect(err).NotTo(gomega.HaveOccurred(), "Failed to create controller-runtime client")
+
+			// Disable CTLB (connect-time load balancing) for Istio ambient mode tests.
+			// CTLB intercepts connections at the socket level, which can bypass ztunnel's
+			// traffic interception and break policy enforcement.
+			originalFC := v3.NewFelixConfiguration()
+			gomega.Eventually(func() error {
+				return cli.Get(context.Background(), types.NamespacedName{Name: "default"}, originalFC)
+			}, 10*time.Second, 1*time.Second).Should(gomega.Succeed(), "failed to get default FelixConfiguration")
+
+			originalCTLB := originalFC.Spec.BPFConnectTimeLoadBalancing
+			ctlbDisabled := v3.BPFConnectTimeLBDisabled
+			gomega.Eventually(func() error {
+				return utils.UpdateFelixConfig(cli, func(spec *v3.FelixConfigurationSpec) {
+					spec.BPFConnectTimeLoadBalancing = &ctlbDisabled
+				})
+			}, 10*time.Second, 1*time.Second).Should(gomega.Succeed(), "failed to disable CTLB")
+
+			ginkgo.DeferCleanup(func() {
+				gomega.Eventually(func() error {
+					return utils.UpdateFelixConfig(cli, func(spec *v3.FelixConfigurationSpec) {
+						spec.BPFConnectTimeLoadBalancing = originalCTLB
+					})
+				}, 10*time.Second, 1*time.Second).Should(gomega.Succeed(), "failed to restore CTLB")
+			})
 		})
 
 		// Test: Full Istio Ambient Mode lifecycle with traffic encryption and Calico policy enforcement.
--- a/felix/bpf-gpl/conntrack.h
+++ b/felix/bpf-gpl/conntrack.h
@@ -898,8 +898,13 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_lookup(struct cali_tc_c
 		CALI_CT_VERB("B: approved %d.", v->b_to_a.approved);
 
 		if (v->a_to_b.approved && v->b_to_a.approved) {
-			result.rc = CALI_CT_ESTABLISHED_BYPASS;
-			ct_result_set_flag(result.rc, CT_RES_CONFIRMED);
+			if (ct_value_get_flags(v) & CALI_CT_FLAG_AMBIENT) {
+				result.rc = CALI_CT_ESTABLISHED;
+				CALI_CT_DEBUG("Ambient connection: suppressing bypass");
+			} else {
+				result.rc = CALI_CT_ESTABLISHED_BYPASS;
+				ct_result_set_flag(result.rc, CT_RES_CONFIRMED);
+			}
 		} else {
 			result.rc = CALI_CT_ESTABLISHED;
 		}
@@ -911,6 +916,11 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_lookup(struct cali_tc_c
 			src_to_dst = &v->b_to_a;
 			dst_to_src = &v->a_to_b;
 		}
+
+		/* Only propagate the AMBIENT flag for NORMAL entries. Other flags
+		 * were previously not propagated and doing so could cause side
+		 * effects (e.g., SKIP_FIB being unexpectedly honored). */
+		result.flags = ct_value_get_flags(v) & CALI_CT_FLAG_AMBIENT;
 
 		break;
 	default:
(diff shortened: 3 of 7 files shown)


felix/bpf-gpl/types.h
```diff
@@ -171,6 +171,8 @@ enum cali_state_flags {
 	CALI_ST_SET_DSCP   = 0x2000,
 	/* CALI_ST_FIRST_FRAG is set if this packet is the first fragment of a fragmented IP packet */
 	CALI_ST_FIRST_FRAG        = 0x4000,
+	/* CALI_ST_AMBIENT is set when the packet is on an ambient mesh workload interface. */
+	CALI_ST_AMBIENT           = 0x8000,
```
tomastigera, line 175: I'd name it more generically as "always enforce policy on every packet"

tomastigera:
why is (non-)CTLB the problem?

tomastigera:
# Code Review

## Overview

The PR fixes a real bug: `CALI_CT_ESTABLISHED_BYPASS` lets ztunnel's persistent HBONE connections bypass policy re-evaluation, so new deny policies never take effect. The fix adds `CALI_CT_FLAG_AMBIENT` / `CALI_ST_AMBIENT`, detects ambient workloads via `ISTIO_DSCP`, and narrows the bypass so ambient flows take `CALI_CT_ESTABLISHED` instead — forcing the policy program on every packet.

## Major issues

- **PR description vs diff mismatch.** The description talks in three places about a `daemon.go` workaround that creates an allow-all override for ztunnel↔istiod under BPF+ambient+CTLB-disabled, plus a "temporary workaround for calico-system tier policies." No `daemon.go` (or any felix Go-side) change is in the 7-file diff. Either the description is stale, or the implementation is missing. If it was intentionally dropped, rewrite the description; if not, the PR is incomplete.

- **DESIGN.md not updated.** `felix/CLAUDE.md` states the update rule explicitly: a BPF dataplane PR that adds **a new CT flag, new mark bit, new map field**, or changes the packet path must update `felix/bpf/DESIGN.md` in the same PR. This PR adds `CALI_CT_FLAG_AMBIENT`, `CALI_ST_AMBIENT`, and narrows the bypass — the DSCP/Istio section (§20) and the fast-path section (§21) should be updated.

- **Fast-path discipline (§21) not addressed.** This is exactly the case §21 singles out: *"A change that suppresses or narrows an existing fast-path shortcut for a class of flows is the same kind of change as adding a per-packet map lookup."* For every packet on an ambient flow you now pay: CT lookup + full policy program re-entry. That's a per-packet cost delta for a class of flows with no discussion in the PR. §21 requires one of: a benchmark, a scoping mechanism that restores the shortcut in steady state (e.g. a gen-counter bumped on policy update), or an explicit case that the affected class is small. Please pick one and write it into the PR description / DESIGN.md.

## Correctness concerns

- **`conntrack.h:920`** — `result.flags = ct_value_get_flags(v) & CALI_CT_FLAG_AMBIENT;` unconditionally overwrites `result.flags` for the NORMAL type. The comment says "previously not propagated," but the assignment is an overwrite, not an OR — if anything adds flags later in the same block, the ordering matters. Safer as `|=`.

- **`tc.c` ambient detection placement in `calico_tc_skb_accepted_entrypoint` (~line 1386).** Gate is `ct_result_is_syn(...) || ct_result_rc == CALI_CT_NEW`. For NEW flows this is correct (flag is copied into the CT entry in `calico_tc_skb_new_flow_entrypoint`). For mid-flow SYN on an already-established entry, the flag is set on state but the CT entry isn't rewritten, so nothing changes. Worth confirming that's intentional; if so, the `ct_result_is_syn` arm is a no-op and should be dropped or commented defensively.

- **`bpf/ut/bpf_prog_test.go`** — the new `if topts.istioDSCP != 0 { ... } else { globals.IstioDSCP = -1 }` treats `0` as the sentinel for "disabled." `0` is a legal DSCP value. Existing tests that explicitly set `istioDSCP = 0` now silently flip to `-1`. Use a `*uint8`, or a separate `istioDSCPEnabled bool`, instead of overloading `0`.

## Test issues

- **`istio_test.go`: dead-import hack.** `_ = gopacket.Default` and `_ = layers.LayerTypeEthernet` with the comment "Ignore the unused import warnings" — `gopacket` and `layers` are already used by other tests in this file (`layers.TCP`, `layers.IPProtocolTCP`, …). These two lines are no-ops; remove them.

- **`TestIstioAmbientPolicyEnforcementV4` coverage gap.** No IPv6 variant, despite `TestIstioDSCPv6` establishing the pattern. The CT flag affects both paths.

- **Missing assertion on the CT flag itself.** The test observes end-to-end behavior (allow vs deny after policy change) but never reads the conntrack entry and asserts `CALI_CT_FLAG_AMBIENT` is set. Adding that would pin the contract and catch regressions where detection stops firing but the bypass suppression still matches some other path.

- **`e2e/pkg/tests/istio/istio.go`.** The CTLB disable `Eventually`s on the API write succeeding but not on the dataplane reload. BPF programs are re-attached when `BPFConnectTimeLoadBalancing` flips — the next test step can race a transitional state. Poll a dataplane signal (e.g., absence of `cali_ctlb_v4` attachment, or a felix-ready probe) before proceeding.

## Minor

- `conntrack.h:905` — `CALI_CT_DEBUG("Ambient connection: suppressing bypass")` has no flow context; the existing `CALI_CT_VERB` lines carry leg state. Either add the 5-tuple or drop to `CALI_CT_VERB`.

## Recommendation

Hold. Blockers: (1) reconcile the description with the diff (daemon.go workaround), (2) update `felix/bpf/DESIGN.md` §20/§21, (3) address the §21 fast-path question in the description with a benchmark or a scoping argument. Then fix the `istioDSCP == 0` sentinel, drop the `_ = gopacket.Default` cruft, and add a CT-flag assertion to the UT.

radixo:
I don't understand some things:
#### 1 - This summary item
> * Temporary workaround for calico-system tier policies blocking cross-node ztunnel<->istiod in BPF mode without CTLB

I don't remind of this problem. The real problem of CTLB is that it overwrite the service IP on the destination and ZTunnel can't identify when to send the packet to a configured waypoint for the service.

#### 2 - CTLB disabled on e2e tests
The other thing I don't understand is why was necessary to disable CTLB on Istio tests, are we testing anything with waypoints? Would not be better to disable the CTLB when the src IP is part of the ISTIO mesh IPSet?

## Late rule evaluation

I don't see it as an Istio problem, it will happen with all long live connections that were created before the rule is applied. So I think we could sort it for everything not only for Istio, proposal:

Felix bumps a `policy_revision` in BPF globals every time it installs a policy program. Each conntrack entry stores the revision it was last validated against. The `calico_ct_lookup` bypass branch returns `CALI_CT_ESTABLISHED_BYPASS` when the entry's revision matches the current one, and `CALI_CT_ESTABLISHED` (re-evaluate once) when it doesn't. After the policy program approves the packet, the entry's revision is re-stamped, restoring the fast path.

Net effect: zero steady-state cost (fast path resumes once policy is stable), one re-eval on the first packet after a policy change, no TCP disruption, and it generalises to all flows rather than just ambient ones.

WDYT? @electricjesus , @tomastigera 

\* I believe we also should think how to sort this to iptables/nftables dataplane :(

PR #12420, opened 2026-04-09
Centralize build and test defaults into defaults.yaml

hjiawei:
Replace metadata.mk with a centralized defaults.yaml file that serves as the single source of truth for build toolchain versions, dependency pins, architecture lists, registry configuration, and other build/test defaults that were previously scattered across dozens of Makefiles and scripts.

Key changes:

- Add defaults.yaml with structured YAML sections: toolchain, versions, base_images, architectures, registries, git, windows, fips, and ci.

- Delete metadata.mk; move its _read_default/_read_default_list helpers and frequently-used variables into lib.Makefile. Variables referenced 3+ times are kept as Make variables for caching; the rest are read inline via $(call _read_default,...) at point of use.

- Remove `include metadata.mk` from all 33 component Makefiles. They now get everything through lib.Makefile.

- Update pin-update functions (update_go_build_pin, update_calico_base_pin) to use yq for reading/writing defaults.yaml instead of grep/sed on the old Make format.

- Fix manifests/generate.sh for Helm v3.20.0 compatibility: --show-only now requires the crds/ prefix for CRD files.

- Update Semaphore CI templates and scripts to reference defaults.yaml instead of metadata.mk in change_in() paths and version reads.

- Bump stale etcd image in charts/calico configure-canal template to match the pinned version.

- Fix hardcoded BIRD image in node k8st Python test to use the ROUTER_IMAGE environment variable already passed by the Makefile.

- Update winfv-cni-plugin setup script to read etcd image from defaults.yaml via yq instead of hardcoding.

- Simplify api/Makefile by removing standalone-repo conditionals (monorepo only).

- Convert execution_time_limit from minutes to hours where >60 min in Semaphore YAML templates (node: 2h, release: 6h).

**Release note:**
```release-note
TBD
```


.semaphore/end-to-end/scripts/test_scripts/e2e-validation.sh
```diff
@@ -1,14 +1,12 @@
...
-# Install yq
-wget -q https://github.com/mikefarah/yq/releases/download/v4.11.0/yq_linux_amd64 -O yq && chmod +x yq
 echo [INFO] Checking pipeline file syntax
 FAILED="false"
 for file in .semaphore/end-to-end/pipelines/*
 do
     echo [INFO] Checking "$file"
-    if bash -c "cat $file | ./yq eval > /dev/null"; then
+    if bash -c "cat $file | yq eval > /dev/null"; then
         echo "OK"
```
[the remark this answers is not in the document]
hjiawei: These Semaphore scripts run on Sem CI agents which have mikefarah/yq pre-installed. No bootstrap needed here.

.semaphore/end-to-end/scripts/body_standard.sh
```diff
@@ -93,7 +93,7 @@ else
       echo "[INFO] starting e2e testing from local binary..."
       pushd "${HOME}/calico"
       make -C e2e build |& tee >(gzip --stdout > "${BZ_LOGS_DIR}/${TEST_TYPE}-build.log.gz")
-      GO_BUILD_VER=$(grep '^GO_BUILD_VER=' ./metadata.mk | cut -d= -f2)
+      GO_BUILD_IMAGE=$(yq '.toolchain.go_build_image + ":" + .toolchain.go_build_ver' ./defaults.yaml)
```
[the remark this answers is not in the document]
hjiawei: This script runs on Semaphore CI agents which have mikefarah/yq pre-installed. Bare `yq` on PATH is intentional here.

process/testing/winfv-cni-plugin/setup-fv.sh
```diff
@@ -42,7 +42,7 @@ function setup_etcd() {
   ${MASTER_CONNECT_COMMAND} "sudo ctr -n k8s.io container delete calico-etcd" || true
 
   # Start etcd container
-  ETCD_CONTAINER="quay.io/coreos/etcd:v3.4.6"
+  ETCD_CONTAINER="$(yq '.base_images.etcd' "${CALICO_HOME}/defaults.yaml"):$(yq '.versions.etcd' "${CALICO_HOME}/defaults.yaml")"
   ${MASTER_CONNECT_COMMAND} "sudo ctr -n k8s.io images pull ${ETCD_CONTAINER}"
```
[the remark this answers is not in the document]
hjiawei: This runs on Semaphore CI agents which have mikefarah/yq pre-installed. Bare `yq` is fine here.

lib.Makefile
```diff
@@ -1,3 +1,72 @@
...
+    || { >&2 echo "yq not found on PATH; downloading $(_YQ_VERSION) to $(_YQ_LOCAL)..." \
+         && mkdir -p $(_DEFAULTS_DIR)bin \
+         && _arch=$$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/') \
+         && _os=$$(uname -s | tr A-Z a-z) \
+         && curl -sSfL --retry 3 -o $(_YQ_LOCAL).tmp \
+              "https://github.com/mikefarah/yq/releases/download/$(_YQ_VERSION)/yq_$${_os}_$${_arch}" \
+         && chmod +x $(_YQ_LOCAL).tmp \
+         && mv $(_YQ_LOCAL).tmp $(_YQ_LOCAL) \
+         && echo $(_YQ_LOCAL); })
+
```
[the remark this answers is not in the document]
hjiawei: In practice the Python yq (kislyuk/yq) is not present on the platforms where this runs (Semaphore CI, developer Linux/macOS machines). The auto-download fallback only triggers when no system yq is found at all. Adding version validation would add complexity for a scenario we have not seen in practice.

lib.Makefile
```diff
@@ -471,40 +531,38 @@ define get_go_build_version
...
-	bash -c '\
-		if [[ "$(new_ver)" > "$(old_ver)" ]]; then \
-			sed -i "s/^GO_BUILD_VER[[:space:]]*=.*/GO_BUILD_VER=$(new_ver)/" $(1); \
-			echo "GO_BUILD_VER is updated to $(new_ver)"; \
-		else \
-			echo "no need to update GO_BUILD_VER"; \
-		fi'
+	@if [ "$(new_ver)" \> "$(old_ver)" ]; then \
+		$(_YQ) -i '.toolchain.go_build_ver = $(new_ver)' $(1); \
+		echo "go_build_ver is updated to $(new_ver)"; \
```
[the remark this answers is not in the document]
hjiawei: Good catch — fixed in 15a6ef5. Added quotes around the RHS so yq treats it as a string literal.

lib.Makefile
```diff
@@ -471,40 +531,38 @@ define get_go_build_version
...
-	bash -c '\
-		if [[ "$(new_ver)" > "$(old_ver)" ]]; then \
-			sed -i "s/^CALICO_BASE_VER[[:space:]]*=.*/CALICO_BASE_VER=$(new_ver)/" $(1); \
-			echo "CALICO_BASE_VER is updated to $(new_ver)"; \
-		else \
-			echo "no need to update CALICO_BASE_VER"; \
-		fi'
+	@if [ "$(new_ver)" \> "$(old_ver)" ]; then \
+		$(_YQ) -i '.versions.calico_base = $(new_ver)' $(1); \
+		echo "calico_base is updated to $(new_ver)"; \
```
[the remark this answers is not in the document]
hjiawei: Fixed in 15a6ef5 along with the go_build_ver quoting.

release/pkg/manager/calico/manager.go
```diff
@@ -1808,3 +1807,45 @@ func ownerFromRemoteURL(raw string) (string, error) {
...
+	}
+
+	var doc yaml.Node
+	if err := yaml.Unmarshal(data, &doc); err != nil {
+		return fmt.Errorf("parsing %s: %w", filePath, err)
+	}
+
+	// doc.Content[0] is the root mapping node.
+	node := doc.Content[0]
+	for i, key := range keyPath {
```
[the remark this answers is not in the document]
hjiawei: Good catch — added a guard in 15a6ef5 that checks `len(doc.Content) > 0` and verifies the root is a MappingNode before accessing it.

node/tests/k8st/tests/test_local_bgp_peer.py
```diff
@@ -331,12 +331,12 @@ def create_workload_pod(self, host, name, ns="default", color="red"):
...
   containers:
   - name: bird
-    image: calico/bird:v0.3.3-211-g9111ec3c
+    image: %s
     securityContext:
       privileged: true
   nodeName: %s
   terminationGracePeriodSeconds: 0
-""" % (name, ns, color, host))
+""" % (name, ns, color, os.environ["ROUTER_IMAGE"], host))
```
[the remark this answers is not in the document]
hjiawei: Good catch — fixed in 15a6ef5. Now imports and uses the `ROUTER_IMAGE` constant from `utils.py` which has the default fallback.

defaults.yaml
```diff
@@ -0,0 +1,135 @@
...
+  use_ssh: true
+
+  # Release
+  dev_tag_suffix: 0.dev
+  release_branch_prefix: release
+
+  # Operator repo
+  operator_branch: master
+  operator_organization: tigera
+  operator_repo: operator
```
radTuti, line 63: I think I will prefer something like this:
```suggestion
  operator:
    branch: master
    organization: tigera
    repo: operator
```
or is the idea to have just one level?

lib.Makefile
```diff
@@ -1,3 +1,71 @@
...
+         && mv $(_YQ_LOCAL).tmp $(_YQ_LOCAL) \
+         && echo $(_YQ_LOCAL); })
+
+ifeq ($(_YQ),)
+$(error yq not found and auto-download failed. Install yq: https://github.com/mikefarah/yq)
+endif
+
+# _read_default reads a scalar value from defaults.yaml.
+# _read_default_list reads a YAML array and returns a space-separated Make list.
+_read_default = $(shell $(_YQ) '$(1)' $(_DEFAULTS_YAML))
```
radTuti, line 39: i think this will return "null" if a key does not exist or is mistyped and needs to be guarded against that.

e.g. `$(call _read_default,.toolchian.go_build_ver)` will result in something like `calico/go-build:null`

lib.Makefile
```diff
@@ -471,40 +530,38 @@ define get_go_build_version
...
+	@echo "latest go_build_ver=$(new_ver)"
 
-	bash -c '\
-		if [[ "$(new_ver)" > "$(old_ver)" ]]; then \
-			sed -i "s/^GO_BUILD_VER[[:space:]]*=.*/GO_BUILD_VER=$(new_ver)/" $(1); \
-			echo "GO_BUILD_VER is updated to $(new_ver)"; \
-		else \
-			echo "no need to update GO_BUILD_VER"; \
-		fi'
+	@if [ "$(new_ver)" \> "$(old_ver)" ]; then \
```
radTuti, line 543: this will like evaluate `X.10.Z` as less than `X.9.Z`, can this use `sort -V` instead?

lib.Makefile
```diff
@@ -471,40 +530,38 @@ define get_go_build_version
...
+	@echo "latest calico_base=$(new_ver)"
 
-	bash -c '\
-		if [[ "$(new_ver)" > "$(old_ver)" ]]; then \
-			sed -i "s/^CALICO_BASE_VER[[:space:]]*=.*/CALICO_BASE_VER=$(new_ver)/" $(1); \
-			echo "CALICO_BASE_VER is updated to $(new_ver)"; \
-		else \
-			echo "no need to update CALICO_BASE_VER"; \
-		fi'
+	@if [ "$(new_ver)" \> "$(old_ver)" ]; then \
```
radTuti, line 559: same as above, switch to `sort -V | tail ...`

lib.Makefile
```diff
@@ -1,3 +1,71 @@
...
+# working yq binary.
+_YQ_LOCAL := $(_DEFAULTS_DIR)bin/yq
+_YQ := $(shell command -v yq 2>/dev/null \
+    || { test -x $(_YQ_LOCAL) && echo $(_YQ_LOCAL); } \
+    || { >&2 echo "yq not found on PATH; downloading latest to $(_YQ_LOCAL)..." \
+         && mkdir -p $(_DEFAULTS_DIR)bin \
+         && _arch=$$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/') \
+         && _os=$$(uname -s | tr A-Z a-z) \
+         && curl -sSfL --retry 3 -o $(_YQ_LOCAL).tmp \
+              "https://github.com/mikefarah/yq/releases/latest/download/yq_$${_os}_$${_arch}" \
```
radTuti, line 28: we should keep using a pinned version instead of "latest"

have `versions.yq` in `defaults.yaml` and read with `grep`

Makefile
```diff
@@ -283,15 +282,14 @@ release-prep: release/bin/release bin/gh
 
 # Install ghr for publishing to github.
 bin/ghr:
-	$(DOCKER_RUN) -e GOBIN=/go/src/$(PACKAGE_NAME)/bin/ $(CALICO_BUILD) go install github.com/tcnksm/ghr@$(GHR_VERSION)
+	$(DOCKER_RUN) -e GOBIN=/go/src/$(PACKAGE_NAME)/bin/ $(CALICO_BUILD) go install github.com/tcnksm/ghr@latest
```
radTuti, line 285: we should stick to pinned versions and have it in `defaults.yaml` either under `versions`

Makefile
```diff
@@ -283,15 +282,14 @@ release-prep: release/bin/release bin/gh
...
-# Install GitHub CLI
+# Install GitHub CLI (latest release). The version is needed because asset
+# filenames embed it (gh_X.Y.Z_linux_amd64.tar.gz).
 bin/gh:
 	@mkdir -p bin
-	@curl -sSL --retry 5 -o bin/gh.tgz https://github.com/cli/cli/releases/download/v$(GITHUB_CLI_VERSION)/gh_$(GITHUB_CLI_VERSION)_linux_amd64.tar.gz
-	@tar -zxvf bin/gh.tgz -C bin/ gh_$(GITHUB_CLI_VERSION)_linux_amd64/bin/gh --strip-components=2
-	@chmod +x $@
-	@rm bin/gh.tgz
+	$(eval GH_VER := $(shell curl -sSf https://api.github.com/repos/cli/cli/releases/latest | jq -r '.tag_name | ltrimstr("v")'))
```
radTuti, line 291: same as above with "latest" version

process/testing/aso/Makefile
```diff
@@ -108,23 +106,20 @@ $(BINDIR)/asoctl:
...
+	curl -sSf -L --retry 5 https://get.helm.sh/helm-$(HELM_VERSION)-linux-$(ARCH).tar.gz | tar xz -C $(@D) --strip-components=1 linux-$(ARCH)/helm
 	chmod +x $@
 	touch $@
 
 CRANE_ARCH = $(subst amd64,x86_64,$(ARCH))
 $(BINDIR)/crane:
 	mkdir -p $(@D)
-	curl -sSfL --retry 5 -o /tmp/crane.tar.gz https://github.com/google/go-containerregistry/releases/download/$(CRANE_VERSION)/go-containerregistry_Linux_$(CRANE_ARCH).tar.gz
-	tar xz -C $(@D) -f /tmp/crane.tar.gz crane && rm -f /tmp/crane.tar.gz
+	curl -sSfL --retry 5 https://github.com/google/go-containerregistry/releases/latest/download/go-containerregistry_Linux_$(CRANE_ARCH).tar.gz | tar xz -C $(@D) crane
```
radTuti, line 122: crane version should be pinned and stored in `defaults.yaml`

release/pkg/manager/calico/manager.go
```diff
@@ -1852,3 +1851,48 @@ func ownerFromRemoteURL(raw string) (string, error) {
...
+		if !found {
+			return fmt.Errorf("key %q not found in %s", strings.Join(keyPath[:i+1], "."), filePath)
+		}
+	}
+
+	out, err := yaml.Marshal(&doc)
+	if err != nil {
+		return fmt.Errorf("marshaling %s: %w", filePath, err)
+	}
+	return os.WriteFile(filePath, out, 0644)
```
radTuti, line 1897: nit, it should probably keep the mode the file is on

radTuti (changes requested):

PR #12421, opened 2026-04-09
fix(manifests): use regex for image registry matching in generate.sh

radTuti:
The static manifest image replacement in `manifests/generate.sh` used an exact match on `${defaultRegistry}/${img}` to find and replace image references. This fails when the registry in the generated manifests doesn't match `defaultRegistry` — the sed pattern silently matches nothing and leaves those images unchanged.

Replace with a regex pattern (`[a-zA-Z0-9/._-]*/${img}`) that matches any registry prefix, so images are correctly updated regardless of the source registry.

**Release note:**
```release-note
None
```


danudey (approved):
lgtm

PR #12433, opened 2026-04-10
[CORE-12479] Protect against clobbering with ExtraComputedSelector usage

nelljerram:
AddExtraComputedSelector and RemoveExtraComputedSelector now use reference counting so that multiple callers (e.g. LiveMigrationCalculator and IstioCalculator) can independently register the same selector without one caller's Remove inadvertently deactivating another's.  The selector is only removed from the label index when the last caller removes it.

The functions are now package-level generics parameterised on a comparable caller identity type, so that each caller passes itself (or another stable identifier) as the caller argument.

Added unit tests for multi-caller scenarios: both callers receiving callbacks, selector remaining active after one caller removes, full deactivation after both remove, duplicate adds from the same caller, remove without prior add, and re-add after full removal.


felix/calc/active_rules_calculator.go
```diff
@@ -291,16 +296,30 @@ func (arc *ActiveRulesCalculator) OnUpdate(update api.Update) (_ bool) {
...
 		log.WithError(err).Panicf("Failed to parse computed selector %#v", cs)
 	}
-	arc.labelIndex.UpdateSelector(computedSelector(cs), sel)
+	callers := arc.computedSelectorCallers[cs]
+	if callers == nil {
+		arc.computedSelectorCallers[cs] = set.New[any]()
+		callers = arc.computedSelectorCallers[cs]
+		arc.labelIndex.UpdateSelector(computedSelector(cs), sel)
+	}
+	callers.Add(any(caller))
```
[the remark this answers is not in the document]
nelljerram: https://github.com/projectcalico/calico/pull/12433/commits/082153c6951cca01392fc55caaba2a79a44270ba

felix/calc/active_rules_calculator.go
```diff
@@ -288,19 +293,43 @@ func (arc *ActiveRulesCalculator) OnUpdate(update api.Update) (_ bool) {
...
+//
+// The underlying selector is added to the label index when the first caller registers it, and it
+// is only removed after the last caller removes its registration.  Callbacks for matches/stops
+// matching continue to be delivered to all registered PolicyMatchListeners while the selector is
+// present.
+func AddExtraComputedSelector[T comparable](arc *ActiveRulesCalculator, cs string, caller T) {
+	callers := arc.computedSelectorCallers[cs]
+	if callers == nil {
+		arc.computedSelectorCallers[cs] = set.New[any]()
+		callers = arc.computedSelectorCallers[cs]
```
radixo: Just to avoid double map lookup:

```suggestion
		callers = set.New[any]()
		arc.computedSelectorCallers[cs] = callers
```

radixo (approved):
LGTM, just a small comment.

felix/calc/active_rules_calculator.go
```diff
@@ -288,19 +293,43 @@ func (arc *ActiveRulesCalculator) OnUpdate(update api.Update) (_ bool) {
...
+	callers := arc.computedSelectorCallers[cs]
+	if callers == nil {
+		callers = set.New[any]()
+		arc.computedSelectorCallers[cs] = callers
+		sel, err := selector.Parse(cs)
+		if err != nil {
+			log.WithError(err).Panicf("Failed to parse computed selector %#v", cs)
+		}
+		arc.labelIndex.UpdateSelector(computedSelector(cs), sel)
 	}
```
fasaxc, line 319: I think you need an `else` branch here that sends the existing matches to the new caller.

felix/calc/active_rules_calculator.go
```diff
@@ -288,19 +293,43 @@ func (arc *ActiveRulesCalculator) OnUpdate(update api.Update) (_ bool) {
...
+// Registration is tracked per caller.  The caller identity must therefore be stable, and the
+// same caller value (including the same inferred type T) must be passed to
+// RemoveExtraComputedSelector to remove that caller's registration.  Repeated adds from the same
+// caller are deduplicated.
+//
+// The underlying selector is added to the label index when the first caller registers it, and it
+// is only removed after the last caller removes its registration.  Callbacks for matches/stops
+// matching continue to be delivered to all registered PolicyMatchListeners while the selector is
+// present.
+func AddExtraComputedSelector[T comparable](arc *ActiveRulesCalculator, cs string, caller T) {
```
fasaxc, line 309: I think the generics dont' really do much here, except add a bit of complexity.  `any` is `comparable` for back-compatibility, so the generics don't even guard against passing a non-comparable.

Something like this would force you to go through a constructor to get an "owner" object (but might be overklll vs just using `string`.
```
type ComputedSelectorOwner struct {
    string
}

func (arc *ActiveRulesCalculator) AddExtraComputedSelector(owner ComputedSelectorOwner, cs string)
```

PR #12445, opened 2026-04-12
Move API code generation into a shared script

caseydavenport:
The API Go code generation (defaults, deep copies, OpenAPI, and the clientset/listers/informers) lived inline in `api/Makefile`. The standalone `projectcalico/api` and `tigera/api` repos mirror `api/build/` but keep their own `Makefile`, so a fix to the generation logic here never reaches them. That just bit us: `deepcopy-gen` dropped `--bounding-dirs` in the k8s 1.36 build image and it was removed here, but the mirrors still pass it, so their `make gen-files` breaks.

This moves all of the Go generation into `api/build/codegen.sh` (replacing `update-client-gen.sh`) so it rides the `build/` mirror and lives in one place. CRD generation stays in the Makefile since it uses the in-repo Calico-patched controller-gen, which doesn't exist in the mirror repos.

The mirror repos still need a one-time `Makefile` cutover to call `codegen.sh` (they keep `Makefile` local); after that this class of drift goes away.

```release-note
None
```


api/Makefile
```diff
@@ -72,68 +72,13 @@ gen-files .generate_files: lint-cache-dir clean-generated
...
-	# generate all pkg/client contents
-	$(DOCKER_RUN) $(CALICO_BUILD) \
-	   sh -c '$(GIT_CONFIG_SSH) $(BUILD_DIR)/update-client-gen.sh'
+	# Generate the Go artifacts (defaults, deep copies, OpenAPI definitions, and
+	# the clientset/listers/informers). The logic lives in build/codegen.sh so it
+	# can be shared with the standalone projectcalico/api and tigera/api repos,
+	# which mirror build/ but keep their own Makefile. CRD generation stays above
+	# because it uses the monorepo-only Calico-patched controller-gen.
+	$(DOCKER_RUN) -e PACKAGE_NAME=$(PACKAGE_NAME) $(CALICO_BUILD) \
+	   sh -c '$(GIT_CONFIG_SSH) $(BUILD_DIR)/codegen.sh'
```
caseydavenport:
```suggestion
	$(DOCKER_RUN) -e PACKAGE_NAME=$(PACKAGE_NAME) $(CALICO_BUILD) sh -c '$(GIT_CONFIG_SSH) $(BUILD_DIR)/codegen.sh'
```

radixo (approved):

PR #12450, opened 2026-04-13
Generate BIRD config files with changes of programClusterRoutes in BGPConfig

mazdakn:
This PR fixes a bug where BIRD configs are not updated with `programClusterRoutes` changes in `BGPConfiguration`. This result in different issues like both Felix and BIRD try to program cluster routes (as reported by https://tigera.atlassian.net/browse/CORE-12641) or neither one programs cluster routes leading to connectivity issue. 

**Release note:**
```release-note
TBD
```


caseydavenport (approved):
LGTM - but we should have a unit test - what would have caught this issue?

PR #12451, opened 2026-04-13
Per Pod/LB IP proxy ARP

MichalFupso:
Adds a functionality that allows calico-node to reply to ARP requests for local Pod or LB IPs for pods and loadbalancer IPs from a no-encap pool in node CIDRs

**Release note:**
```release-note
Felix now responds to ARP requests for pod and LoadBalancer IPs on the node's subnet, enabling L2 reachability without BGP or overlays.
```


felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,721 @@
...
+	"net/netip"
+	"regexp"
+	"strings"
+	"sync/atomic"
+	"time"
+
+	"github.com/mdlayher/arp"
+	"github.com/mdlayher/ethernet"
+	"github.com/mdlayher/ndp"
+	log "github.com/sirupsen/logrus"
```
mazdakn: nit: just use `logrus` instead of `log` alias.

api/pkg/apis/projectcalico/v3/felixconfig.go
```diff
@@ -95,6 +95,14 @@ const (
 	FloatingIPsDisabled FloatingIPType = "Disabled"
 )
 
+// +kubebuilder:validation:Enum=Disabled;PodsAndLoadBalancers
+type LocalSubnetL2ReachabilityMode string
+
+const (
+	LocalSubnetL2ReachabilityDisabled             LocalSubnetL2ReachabilityMode = "Disabled"
+	LocalSubnetL2ReachabilityPodsAndLoadBalancers LocalSubnetL2ReachabilityMode = "PodsAndLoadBalancers"
```
mazdakn, line 103: IMO, we should add a `Pods` value to exclude load balancers, since the PMREQ is mainly for egw use case.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -747,6 +750,20 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		}
 	}
 
+	// Register the proxy neighbor managers when enabled. They listen on raw sockets
+	// and respond to ARP (IPv4) / NDP (IPv6) requests for pod and LB IPs that fall
+	// within the same subnet as a host physical interface.
+	if config.LocalSubnetL2Reachability == "PodsAndLoadBalancers" {
```
mazdakn: nit: if you check `config.LocalSubnetL2Reachability != "Disabled"`, you would basically support future values.

mazdakn: Instead of string literal you should use the defined const, i.e. `LocalSubnetL2ReachabilityDisabled`

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,729 @@
...
+
+	// localWorkloadIPs maps workload endpoint ID to the pod's IP strings.
+	localWorkloadIPs map[types.WorkloadEndpointID][]string
+
+	// lbServiceIPs maps service ID to LoadBalancer ingress IP strings.
+	lbServiceIPs map[serviceID][]string
+
+	// clusterNodes is the set of hostnames known to the cluster. Used as the
+	// hash ring for LB VIP node selection.
+	clusterNodes map[string]struct{}
```
mazdakn: we should use `set.Set` for this.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,729 @@
...
+	ndpFactory ndpConnFactory
+
+	ctx    context.Context
+	cancel context.CancelFunc
+}
+
+func newProxyNeighManager(dpConfig Config, ipVersion uint8) *proxyNeighManager {
+	nl, err := netlinkshim.NewRealNetlink()
+	if err != nil {
+		log.WithError(err).Panic("Failed to create netlink handle for proxy neighbor manager")
```
mazdakn: I don't think we should panic here. In other managers, we ignore the error. The better approach could be just to log an error message, and leave it as it is.

mazdakn:
@MichalFupso to make the review easier, it would be better to split this PR into multiple small ones:
- One for the functionality only for pods (including UT and FVs)
- One for including load balancers
- One for E2E

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,729 @@
...
+		// they can't host pods/VIPs in our family, so they shouldn't be in
+		// the hash ring.
+		addr := msg.Ipv4Addr
+		if m.ipVersion == 6 {
+			addr = msg.Ipv6Addr
+		}
+		if addr == "" {
+			return
+		}
+		log.WithField("hostname", msg.Hostname).Debug("Proxy neighbor manager received HostMetadataV4V6Update")
```
mazdakn:
```suggestion
		log.WithField("hostname", msg.Hostname).Debug("Proxy neighbor manager received HostMetadataUpdate")
```

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,729 @@
...
+		m.clusterNodes[msg.Hostname] = struct{}{}
+		m.dirty = true
+
+	case *proto.HostMetadataRemove:
+		if _, ok := m.clusterNodes[msg.Hostname]; ok {
+			log.WithField("hostname", msg.Hostname).Debug("Proxy neighbor manager received HostMetadataV4V6Remove")
```
mazdakn:
```suggestion
			log.WithField("hostname", msg.Hostname).Debug("Proxy neighbor manager received HostMetadataRemove")
```

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,729 @@
...
+		wlKey := types.ProtoToWorkloadEndpointID(msg.GetId())
+		if _, ok := m.localWorkloadIPs[wlKey]; ok {
+			log.WithField("workload", wlKey).Debug("Proxy neighbor manager received WorkloadEndpointRemove")
+			delete(m.localWorkloadIPs, wlKey)
+			m.dirty = true
+		}
+
+	case *proto.ServiceUpdate:
+		svcKey := serviceID{Namespace: msg.Namespace, Name: msg.Name}
+		if msg.Type != "LoadBalancer" {
```
mazdakn: Can we use const instead instead of string literal

api/pkg/apis/projectcalico/v3/felixconfig.go
MichalFupso: I think it's okay to keep it together "PodsAndLoadBalancers", you have to create the no-encap LB pool which acts as another config gate

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,732 @@
...
+
+	"github.com/projectcalico/calico/felix/netlinkshim"
+	"github.com/projectcalico/calico/felix/proto"
+	"github.com/projectcalico/calico/felix/types"
+	"github.com/projectcalico/calico/libcalico-go/lib/set"
+)
+
+// desiredIPSet is an immutable set of IPs that a listener goroutine reads to
+// decide whether to respond to an ARP/NDP request. Swapped atomically.
+type desiredIPSet map[string]bool
```
mazdakn: We should use `set.Set` type. Any reason for not using it?

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,732 @@
...
+	logrus.WithField("iface", l.ifaceName).Info("Stopped proxy neighbor listener")
+}
+
+// runARPListener listens for ARP requests on a raw socket and replies for
+// IPs in the desired set.
+func (m *proxyNeighManager) runARPListener(ctx context.Context, l *ifaceListener) {
+	defer close(l.done)
+
+	for {
+		_ = l.arpCli.SetReadDeadline(time.Now().Add(1 * time.Second))
```
mazdakn: nit: we should define a constant for the 1 second deadline.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,732 @@
...
+	SetReadDeadline(t time.Time) error
+	Close() error
+}
+
+type arpClientFactory func(ifaceName string) (arpClient, net.HardwareAddr, error)
+type ndpConnFactory func(ifaceName string) (ndpConn, net.HardwareAddr, error)
+
+// ifaceListener manages a raw socket listener for a single host interface.
+type ifaceListener struct {
+	ifaceName string
```
mazdakn: nit: `ifaceListener` is used as a `map[iface name]ifaceListerner`, so adding a separate `ifaceName` field seems redundant.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,732 @@
...
+	arpCli    arpClient // IPv4 only
+	ndpCli    ndpConn   // IPv6 only
+	hwAddr    net.HardwareAddr
+	cancel    context.CancelFunc
+	done      chan struct{}
+}
+
+// proxyNeighManager automatically responds to ARP (IPv4) and NDP (IPv6) requests for
+// pod and LoadBalancer IPs that fall within the same L2 subnet as a host interface.
+// Instead of programming kernel proxy neighbor entries, it listens on raw sockets and
```
mazdakn:
```suggestion
// It listens on raw sockets and
```

mazdakn: `Instead of programming kernel proxy neighbor entries` looks out of context. It's based on design doc, but doesn't add any value here.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,732 @@
...
+// startListener opens a raw socket on the given interface and starts a
+// listener goroutine.
+func (m *proxyNeighManager) startListener(ifaceName string) error {
+	l := &ifaceListener{
+		ifaceName: ifaceName,
+		done:      make(chan struct{}),
+	}
+
+	ctx, cancel := context.WithCancel(m.ctx)
+	l.cancel = cancel
```
mazdakn: It doesn't seem that we do anything with the new context here. I think we should just remove these two lines. We can just use the context created initially.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,732 @@
...
+			}).Debug("NDP listener: failed to send NA")
+		}
+	}
+}
+
+// sendGARP sends a gratuitous ARP (IPv4) or unsolicited NA (IPv6) for the
+// given IP using the listener's raw socket.
+func (m *proxyNeighManager) sendGARP(l *ifaceListener, ipStr string) {
+	if m.ipVersion == 4 {
+		ip, err := netip.ParseAddr(ipStr)
```
mazdakn: This condition is common for both ip families. Just move it to the outside of ip family condition.

mazdakn: Also, it would be more readable by reducing the nesting. For example you can move the logic for each ip family to a separate function, and just call them here.:

```
if m.ipVersion == 4 {
  m.SerndGARPv4()
  return
}

m.SendGARPv6()
```

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,732 @@
...
+type ndpConnFactory func(ifaceName string) (ndpConn, net.HardwareAddr, error)
+
+// ifaceListener manages a raw socket listener for a single host interface.
+type ifaceListener struct {
+	ifaceName string
+	desired   atomic.Pointer[desiredIPSet]
+	arpCli    arpClient // IPv4 only
+	ndpCli    ndpConn   // IPv6 only
+	hwAddr    net.HardwareAddr
+	cancel    context.CancelFunc
```
mazdakn: what's the purpose of the `cancel` field? doesn't seem to be used.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,732 @@
...
+	listeners map[string]*ifaceListener
+
+	dirty    bool
+	nlHandle netlinkshim.Interface
+
+	arpFactory arpClientFactory
+	ndpFactory ndpConnFactory
+
+	ctx    context.Context
+	cancel context.CancelFunc
```
mazdakn, line 121: This field seems unnecessary. Any reason for it?

mazdakn (commented):
Left a few more comments

felix/dataplane/linux/proxy_neigh_mgr.go
MichalFupso: It's used in tests to stop the manager and the garp go routines

felix/dataplane/linux/proxy_neigh_mgr.go
MichalFupso: The new ctx is passed to the GARP go routine, when we call l.cancel() the go routine is closed

felix/dataplane/linux/proxy_neigh_mgr.go
MichalFupso: I would keep it here, we use it in logs and the consumers only receive ifaceListener. Otherwise we would have to pass the name as a separate argument

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,743 @@
...
+}
+
+func newProxyNeighManager(dpConfig Config, ipVersion uint8) *proxyNeighManager {
+	nl, err := netlinkshim.NewRealNetlink()
+	if err != nil {
+		logrus.WithError(err).Error("Failed to create netlink handle for proxy neighbor manager")
+		return nil
```
mazdakn: we should not return nil. It seems in other managers we just ignore the error and continue, which is no ideal. But we most likely don't expect to see any error at this stage. 

Here, It would be enough to log an error, and just continue like the rest of managers.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,743 @@
...
+		ctx:              ctx,
+		cancel:           cancel,
+	}
+}
+
+func (m *proxyNeighManager) OnUpdate(protoBufMsg any) {
+	switch msg := protoBufMsg.(type) {
+	case *ifaceAddrsUpdate:
+		if m.wlIfacesRegexp.MatchString(msg.Name) {
+			return
```
mazdakn, line 239: worth adding a debug log to explain why it's ignored. Something like "Workload interface, ignoring"

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,743 @@
...
+			"ips":           ips,
+			"migrationRole": ep.LiveMigrationRole,
+		}).Debug("Proxy neighbor manager received WorkloadEndpointUpdate")
+		if ep.LiveMigrationRole == proto.LiveMigrationRole_SOURCE {
+			delete(m.localWorkloadIPs, wlKey)
+		} else if len(ips) > 0 {
+			m.localWorkloadIPs[wlKey] = ips
+		} else {
+			delete(m.localWorkloadIPs, wlKey)
+		}
```
mazdakn, line 273: nit: using multiple if/else if levels reduces readability. It's better to use `switch` instead. 

but here you can summarize this into:

```
if ep.LiveMigrationRole != proto.LiveMigrationRole_SOURCE && len(ips) > 0 {
  m.localWorkloadIPs[wlKey] = ips
} else {
  delete(m.localWorkloadIPs, wlKey)
}
```

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,743 @@
...
+	}
+	if m.ipVersion == 4 {
+		return ip.To4() != nil
+	}
+	return ip.To4() == nil
+}
+
+func isNoEncapPool(pool *proto.IPAMPool) bool {
+	return (pool.IpipMode == "" || strings.EqualFold(pool.IpipMode, "Never")) &&
+		(pool.VxlanMode == "" || strings.EqualFold(pool.VxlanMode, "Never"))
```
mazdakn: We should not use string literals. we have v3 constant for all of these.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,743 @@
...
+		if cidr.Contains(ip) {
+			return true
+		}
+	}
+	return false
+}
+
+// removeHostIface drops any tracked subnet info for ifaceName. Used when the
+// interface goes away.
+func (m *proxyNeighManager) removeHostIface(ifaceName string) {
```
mazdakn: This func is used only in one place (in `OnUpdate`), and it's better to just embed it like other `delete` statements.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,743 @@
...
+			lbIP := net.ParseIP(ipStr)
+			if lbIP == nil {
+				continue
+			}
+			if !m.isInNoEncapPool(lbIP) {
+				continue
+			}
+			m.addMatchingIPs(desiredByIface, lbIP)
+		}
+	}
```
mazdakn, line 454: Worth breaking this into different functions. For example, for you can have a function for creating and compiling `desiredByIface`

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,759 @@
...
+
+	return desiredByIface
+}
+
+// reconcileListeners starts a listener for each interface that newly needs one
+// and stops listeners for interfaces that no longer appear in the desired
+// state. On a start failure it re-marks the manager dirty (so the next
+// CompleteDeferredWork retries) and returns the error.
+func (m *proxyNeighManager) reconcileListeners(desiredByIface map[string]set.Set[string]) error {
+	var startErr error
```
mazdakn: nit: it's more idiomatic to name this `err` and just use it. You won't need to init a new one in `err := m.startListener()` and setting `startErr` later on.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,759 @@
...
+// appearing on that interface.
+func (m *proxyNeighManager) publishDesiredIPs(desiredByIface map[string]set.Set[string]) {
+	for ifaceName, desired := range desiredByIface {
+		l, ok := m.listeners[ifaceName]
+		if !ok {
+			continue
+		}
+		// Atomically swap in the new desired set; the previous pointer tells
+		// us which IPs are new and need a GARP/UNA.
+		d := desired
```
mazdakn: why declaring a new var here?

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,759 @@
...
+			if old == nil || !(*old).Contains(ip) {
+				m.sendGARP(l, ip)
+			}
+		}
+	}
+}
+
+// addMatchingIPs adds the IP to the desired set for every host interface whose subnet
+// contains it.
+func (m *proxyNeighManager) addMatchingIPs(desiredByIface map[string]set.Set[string], ip net.IP) {
```
mazdakn: This function is used in `buildDesiredState`. It would be cleaner to place it right after that function. It would be like:

```
func buildDesiredState() {
...
  addMatchingIPs()
...
}

func addMatchingIPs() {
}
```

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,759 @@
...
+}
+
+// stopListener cancels and cleans up a listener.
+func (m *proxyNeighManager) stopListener(l *ifaceListener) {
+	l.cancel()
+	if l.arpCli != nil {
+		_ = l.arpCli.Close()
+	}
+	if l.ndpCli != nil {
+		_ = l.ndpCli.Close()
```
mazdakn: we should log these two error cases. If either one fails, it helps to detect it, and prevent a potential resource leakage.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,759 @@
...
+	logrus.WithField("iface", l.ifaceName).Info("Stopped proxy neighbor listener")
+}
+
+// runARPListener listens for ARP requests on a raw socket and replies for
+// IPs in the desired set.
+func (m *proxyNeighManager) runARPListener(ctx context.Context, l *ifaceListener) {
+	defer close(l.done)
+
+	for {
+		_ = l.arpCli.SetReadDeadline(time.Now().Add(readDeadlineInterval))
```
mazdakn: we should log an error here instead of ignoring the error.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,759 @@
...
+
+		pkt, _, err := l.arpCli.Read()
+		if err != nil {
+			if ctx.Err() != nil {
```
mazdakn: we should check for context situation first thing in the for loop.

mazdakn: You also need to differentiate between a timeout and a real error. You need to cast the error to net.Error and check the `Timeout` value. Something like:

```
for {
  if err := ctx.Err(); err != nil {
    return err
  }

  l.arpCli.SetReadDeadline(time.Now().Add(readDeadlineInterval))

  pkt, _, err := l.arpCli.Read()
  if err != nil {
    if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
      continue
    }
    // Persistent error, log here, and we should do something more probably.
  }
}
```

mazdakn: This is the same for `runNDPListener` function.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,759 @@
...
+
+	logrus.WithFields(logrus.Fields{
+		"numWorkloads":  len(m.localWorkloadIPs),
+		"numHostIfaces": len(m.hostIfaceToCIDRs),
+		"numLBServices": len(m.lbServiceIPs),
+		"numNodes":      m.clusterNodes.Len(),
+	}).Debug("Proxy neighbor manager CompleteDeferredWork")
+
+	desiredByIface := m.buildDesiredState()
+	startErr := m.reconcileListeners(desiredByIface)
```
mazdakn: nit: just name this `err`

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,795 @@
...
+func (m *proxyNeighManager) runARPListener(ctx context.Context, l *ifaceListener) {
+	defer close(l.done)
+
+	for {
+		if ctx.Err() != nil {
+			return
+		}
+
+		if err := l.arpCli.SetReadDeadline(time.Now().Add(readDeadlineInterval)); err != nil {
+			logrus.WithError(err).WithField("iface", l.ifaceName).Debug("Failed to set ARP read deadline")
```
mazdakn: This is a more serious error, and warrant bumping it to either warn or error.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,795 @@
...
+// appearing on that interface.
+func (m *proxyNeighManager) publishDesiredIPs(desiredByIface map[string]set.Set[string]) {
+	for ifaceName, desired := range desiredByIface {
+		l, ok := m.listeners[ifaceName]
+		if !ok {
+			continue
+		}
+		// Atomically swap in the new desired set; the previous pointer tells
+		// us which IPs are new and need a GARP/UNA.
+		old := l.desired.Swap(new(desired))
```
mazdakn: using `new(desired)` is similar to the previous code. We can just use `Swap(&desired)`.

e2e/pkg/describe/describe.go
```diff
@@ -152,6 +153,15 @@ func RequiresExternalNode() any {
 	return framework.WithLabel("RequiresExternalNode")
 }
 
+// RequiresKindNetwork marks tests that need to attach a docker container to the
+// host's "kind" network as an L2-adjacent peer of the cluster nodes. These
+// tests can only run where the test runner has direct access to the local
+// docker daemon (i.e., kind clusters), and must be skipped elsewhere via
+// --ginkgo.skip=KindNetwork.
+func RequiresKindNetwork() any {
```
caseydavenport: This is a red flag - we shouldn't be writing tests that are so tightly tied to a specific cluster installer.

If this requires kind, it sounds like it really belongs more as part of our kind k8st ST suite and not part of e2es.

e2es are intended to verify cross-platform feature support. I think we need to either:

- Split this such that the test code is orthogonal to the infrastructure layer
- Move this test to the k8s ST suite (which is Python, but we can still run Go tests there).

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,795 @@
...
+		}
+
+		if err := l.arpCli.SetReadDeadline(time.Now().Add(readDeadlineInterval)); err != nil {
+			logrus.WithError(err).WithField("iface", l.ifaceName).Debug("Failed to set ARP read deadline")
+		}
+
+		pkt, _, err := l.arpCli.Read()
+		if err != nil {
+			if err, ok := errors.AsType[net.Error](err); ok && err.Timeout() {
+				logrus.WithField("iface", l.ifaceName).Debug("ARP read deadline expired with no packet")
```
mazdakn: I think this will lead to log spam. Imagine, 3 ideal goroutines, spitting out a debug message every second. Probably better to remove it.

e2e/pkg/describe/describe.go
caseydavenport: On this note - I just gave the same feedback to @song-jiang here - https://github.com/projectcalico/calico/pull/12892#discussion_r3345727819

I think the lack of a Golang suite that runs on kind that isn't our e2es is pushing us to write tests in e2es that sholudn't actually be in the e2es.

caseydavenport: But if we can factor this as "The tests just need an external node that is L2 reachable to the cluster" and "Kind is one way of providing that" then I think it changes the picture - it's just that the test shouldn't be so tightly coupled to kind - it should just declare it's abstract environmental requirements.

e2e/pkg/tests/networking/proxy_neigh.go
```diff
@@ -0,0 +1,659 @@
...
+	for i := range pBytes {
+		pBytes[i] |= sBytes[i]
+	}
+	return fmt.Sprintf("%s/%d", netip.AddrFrom16(pBytes), suffix.Bits()), nil
+}
+
+// createIPPool creates the given pool and registers a DeferCleanup to delete
+// it. We don't return the cleanup func because every pool we create in this
+// test should be torn down at the end of the spec.
+func createIPPool(ctx context.Context, cli ctrlclient.Client, pool *v3.IPPool) {
```
caseydavenport: I don't think we need a helper for this? but if we do, we should call it `mustCreate` to indicate it fails, and it should use `ExpectWithOffset`

caseydavenport: Alternatively, have this function return a cleanupFn that can be passed directly to DeferCleanup and it earns its value back sort of.

e2e/pkg/tests/networking/proxy_neigh.go
```diff
@@ -0,0 +1,659 @@
...
+}
+
+// setLocalSubnetL2Reachability mutates the default FelixConfiguration to put
+// LocalSubnetL2Reachability into the requested mode and returns a function
+// that restores the original value. The restore function is idempotent.
+func setLocalSubnetL2Reachability(cli ctrlclient.Client, mode v3.LocalSubnetL2ReachabilityMode) func() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	cfg := &v3.FelixConfiguration{}
```
caseydavenport: We have a helper - ConfigureWithCleanup - that should be used for operations like this.

e2e/pkg/tests/networking/proxy_neigh.go
```diff
@@ -0,0 +1,659 @@
...
+		}, "20s", "2s").Should(Succeed(), "restoring original LocalSubnetL2Reachability")
+	}
+}
+
+// deployServer brings up a conncheck Server backed by a pod whose IP is
+// allocated from the given IP pool. It returns the conncheck Server (so the
+// caller can register cleanup with ct.Stop) and the pod's IP of the requested
+// family. The conncheck framework handles pod creation, readiness, and
+// re-fetching the pod after it's running so Status.PodIPs is populated.
+func deployServer(
```
caseydavenport: Why write a wrapper for this? 

I think it just makes it harder to add multiple pods to the connchecker - the idea of conncheck is that it can do things in parallel, so:

```
Add()
Add()
Add()
Deploy()
```

e2e/pkg/tests/networking/proxy_neigh.go
```diff
@@ -0,0 +1,659 @@
...
+// network (the kind network by default). Tests use it to send ARP/NDP probes
+// from a location that is layer-2 adjacent to the cluster nodes but is not
+// itself a cluster node.
+//
+// The container is given NET_RAW (for raw ARP/NDP sockets) and NET_ADMIN (for
+// netshoot tools that tweak interfaces/routes) and runs `sleep infinity` so we
+// can docker exec into it. Callers MUST defer Close() to remove the container.
+
+const (
+	kindNetwork = "kind"
```
caseydavenport: The kind network may not always be named this.

e2e/pkg/tests/networking/proxy_neigh.go
```diff
@@ -0,0 +1,659 @@
...
+	extL2Iface  = "eth0"
+)
+
+// extL2Peer wraps a single docker container attached to a bridge network.
+type extL2Peer struct {
+	name string
+}
+
+func newExtL2Peer(name, network string) (*extL2Peer, error) {
```
caseydavenport: Think this should be set up outside of the tests (per my other comments)

e2e/pkg/tests/networking/proxy_neigh.go
```diff
@@ -0,0 +1,659 @@
...
+// netshoot tools that tweak interfaces/routes) and runs `sleep infinity` so we
+// can docker exec into it. Callers MUST defer Close() to remove the container.
+
+const (
+	kindNetwork = "kind"
+	extL2Iface  = "eth0"
+)
+
+// extL2Peer wraps a single docker container attached to a bridge network.
+type extL2Peer struct {
```
caseydavenport: Ideally this would implement the conncheck.Client interface so that it can be used in that framework - that's what we moved to for e.g., external nodes

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,795 @@
...
+		desired := l.desired.Load()
+		if desired == nil || !(*desired).Contains(targetIP) {
+			continue
+		}
+
+		if err := l.arpCli.Reply(pkt, l.hwAddr, pkt.TargetIP); err != nil {
+			logrus.WithError(err).WithFields(logrus.Fields{
+				"iface": l.ifaceName,
+				"ip":    targetIP,
+			}).Debug("ARP listener: failed to send reply")
```
mazdakn: This should be at least a warning log.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,795 @@
...
+		}
+	}
+}
+
+// sendGARP parses ipStr and dispatches to the IPv4 (gratuitous ARP) or IPv6
+// (unsolicited NA) sender for the manager's address family.
+func (m *proxyNeighManager) sendGARP(l *ifaceListener, ipStr string) {
+	addr, err := netip.ParseAddr(ipStr)
+	if err != nil {
+		return
```
mazdakn: need to log an error here.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,795 @@
...
+		}
+
+		// If the NS source is the unspecified address (::), this is a
+		// Duplicate Address Detection probe (RFC 4862). We can't unicast
+		// back to ::, so reply to the all-nodes multicast address and
+		// clear the Solicited flag (RFC 4861 §4.4).
+		dst := srcAddr
+		solicited := true
+		if !srcAddr.IsValid() || srcAddr.IsUnspecified() {
+			dst = netip.MustParseAddr("ff02::1")
```
mazdakn: `ff02::1` is used in a couple of place. Worth declaring a constant for it.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,795 @@
...
+	}
+	if m.ipVersion == 4 {
+		return ip.To4() != nil
+	}
+	return ip.To4() == nil
+}
+
+func isNoEncapPool(pool *proto.IPAMPool) bool {
+	return (pool.IpipMode == "" || v3.IPIPMode(pool.IpipMode) == v3.IPIPModeNever) &&
+		(pool.VxlanMode == "" || v3.VXLANMode(pool.VxlanMode) == v3.VXLANModeNever)
```
mazdakn: You can also use v3.Never for empty string: `v3.EncapMode(pool.Ipipode) == v3.Never`.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -381,6 +383,7 @@ type InternalDataplane struct {
 	allManagers             []Manager
 	managersWithRouteTables []ManagerWithRouteTables
 	managersWithRouteRules  []ManagerWithRouteRules
+	proxyNeighManagers      []*proxyNeighManager
```
mazdakn: I don't think we need this.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,806 @@
...
+		// the hash ring.
+		addr := msg.Ipv4Addr
+		if m.ipVersion == 6 {
+			addr = msg.Ipv6Addr
+		}
+		if addr == "" {
+			return
+		}
+		logrus.WithField("hostname", msg.Hostname).Debug("Proxy neighbor manager received HostMetadataUpdate")
+		m.clusterNodes.Add(msg.Hostname)
```
mazdakn: before adding the host name, and setting dirty flag, we should check if the hostname already exists in the set (using Contain). If so, then we should just ignore the update.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,806 @@
...
+
+func (m *proxyNeighManager) OnUpdate(protoBufMsg any) {
+	switch msg := protoBufMsg.(type) {
+	case *ifaceAddrsUpdate:
+		if m.wlIfacesRegexp.MatchString(msg.Name) {
+			logrus.WithFields(logrus.Fields{
+				"ifaceName": msg.Name,
+			}).Debug("Ignoring ifaceAddrsUpdate for workload interface")
+			return
+		}
```
mazdakn, line 240: Do we also need to consider `InterfaceExclude` config here?

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,806 @@
...
+	Close() error
+}
+
+// ndpConn abstracts the mdlayher/ndp.Conn for testing.
+type ndpConn interface {
+	ReadFrom() (ndp.Message, *ipv6.ControlMessage, netip.Addr, error)
+	WriteTo(m ndp.Message, cm *ipv6.ControlMessage, dst netip.Addr) error
+	SetReadDeadline(t time.Time) error
+	Close() error
+}
```
mazdakn, line 78: nit: You don't need to name all parameters. It only worth naming when it add some context.
For example, in `SetReadDeadline(t time.Time)` using `t` doesn't add anything. It's cleaner to just say `SetReadDeadline(time.Time)` instead.

For `ReadFrom` and `WriteTo` is a bit fuzzy. IMO, it's also OK to remove the parameters name, but at least naming the last one adds some context. In that case you can do something like:
```
WriteTo(_  ndp.Message, _ ipv6.ControlMessage, dst netip.Addr) error
```

mazdakn (changes requested):

felix/dataplane/linux/proxy_neigh_mgr.go
MichalFupso: I think that is handled upstream in iface_monitor

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,802 @@
...
+	logrus.WithFields(logrus.Fields{
+		"numWorkloads":  len(m.localWorkloadIPs),
+		"numHostIfaces": len(m.hostIfaceToCIDRs),
+		"numLBServices": len(m.lbServiceIPs),
+		"numNodes":      m.nodeRing.Len(),
+	}).Debug("Proxy neighbor manager CompleteDeferredWork")
+
+	desiredByIface := m.buildDesiredState()
+	err := m.reconcileListeners(desiredByIface)
+	m.publishDesiredIPs(desiredByIface)
```
mazdakn, line 396: nit: It seems we want to call `publishDesiredIPs` independent of return value of `reconcileListeners`. In that case, I think it's better to change the order:

```
m.publishDesiredIPs(desiredByIface)
return m.reconcileListeners(desiredByIface)
```

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,802 @@
...
+}
+
+// runARPListener listens for ARP requests on a raw socket and replies for
+// IPs in the desired set.
+func (m *proxyNeighManager) runARPListener(ctx context.Context, l *ifaceListener) {
+	defer close(l.done)
+
+	for {
+		if ctx.Err() != nil {
+			return
```
mazdakn: worth adding a debug log here that the context is cancelled.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,802 @@
...
+		go m.runNDPListener(ctx, l)
+	}
+
+	m.listeners[ifaceName] = l
+	logrus.WithField("iface", ifaceName).Info("Started proxy neighbor listener")
+	return nil
+}
+
+// stopListener cancels and cleans up a listener.
+func (m *proxyNeighManager) stopListener(l *ifaceListener) {
```
mazdakn: `stopListerner` should be a method of `ifaceListerner` not `proxyNeighManager`. It does not use anything from the latter. 

You can make it a method of `ifaceListener` and rename it to `stop`.

mazdakn: This looks the same for `runARPListener`, `runNDPListener`, `sendGARPV4` and `sendGARPv6`. This way all ARP/NDP related code are encapsulated nicely in the `ifaceListener` struct.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,802 @@
...
+				m.dirty = true
+			}
+		}
+	}
+	for ifaceName, l := range m.listeners {
+		if _, ok := desiredByIface[ifaceName]; !ok {
+			m.stopListener(l)
+			delete(m.listeners, ifaceName)
+		}
+	}
```
mazdakn: You can merge this look with the first one, something like:

```
for ifaceName, l := range m.listeners {
  _, desired := desiredByIface[ifaceName]
  if desired && !l.failed.Load() {
    continue
  }
  m.stopListener(l)
  delete(m.listeners, ifaceName)
}
```

This way you can also remove `hasFailedListener` function.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,802 @@
...
+	hwAddr  net.HardwareAddr
+	cancel  context.CancelFunc
+	done    chan struct{}
+	failed  atomic.Bool
+}
+
+// proxyNeighManager automatically responds to ARP (IPv4) and NDP (IPv6) requests for
+// pod and LoadBalancer IPs that fall within the same L2 subnet as a host interface.
+// It listens on raw sockets and
+// responds directly in userspace.
```
mazdakn: the last two line looks like to be truncated.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,802 @@
...
+		nf = func(ifaceName string) (ndpConn, net.HardwareAddr, error) {
+			ifi, err := net.InterfaceByName(ifaceName)
+			if err != nil {
+				return nil, nil, err
+			}
+			conn, _, err := ndp.Listen(ifi, ndp.Unspecified)
+			if err != nil {
+				return nil, nil, err
+			}
+			// Set the IPMP6 filter to only deliver Neighbor Solicitations
```
mazdakn:
```suggestion
			// Set the ICMPv6 filter to only deliver Neighbor Solicitations
```

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,802 @@
...
+
+		if err := l.arpCli.SetReadDeadline(time.Now().Add(readDeadlineInterval)); err != nil {
+			logrus.WithError(err).WithField("iface", l.ifaceName).Warn("Failed to set ARP read deadline")
+		}
+
+		pkt, _, err := l.arpCli.Read()
+		if err != nil {
+			if err, ok := errors.AsType[net.Error](err); ok && err.Timeout() {
+				continue
+			}
```
mazdakn: We should return if the context is cancelled. That's an indication that `stopListener` is called, so we should try to recreate the listener again.

felix/dataplane/linux/proxy_neigh_mgr_test.go
```diff
@@ -0,0 +1,735 @@
...
+
+		It("should not respond for IPs in encapsulated pools", func() {
+			Expect(mgr.listeners).To(BeEmpty())
+		})
+	})
+})
+
+var _ = Describe("Proxy neighbor manager - LoadBalancer IPs", func() {
+
+	It("is claimed by exactly one node, which answers and GARPs for it", func() {
```
mazdakn, line 682: we need more tests, like removing a node and asserting expected behaviour.

felix/dataplane/linux/proxy_neigh_mgr_test.go
```diff
@@ -0,0 +1,735 @@
...
+	if d == nil {
+		return nil
+	}
+	return *d
+}
+
+// --- Tests ---
+
+var _ = Describe("Proxy neighbor manager (IPv4)", func() {
+	var (
```
mazdakn, line 416: This UT is really useful, and we should add a few more tests:
- Multi interface setup utilizing both eth0 and eth1.
- Tests that actually asserts ARP request/replies.
- Updating GARP tests to check that correct packet with expected information is sent.

felix/dataplane/linux/proxy_neigh_mgr_test.go
```diff
@@ -0,0 +1,735 @@
...
+				// Non-owning nodes don't open a listener for it at all.
+				Expect(mgr.listeners).To(BeEmpty())
+			}
+			mgr.cancel()
+		}
+		Expect(answering).To(Equal(1)) // exactly one node owns the VIP
+	})
+})
+
+var _ = Describe("Proxy NDP manager (IPv6)", func() {
```
mazdakn, line 715: Similar to IPv4, we should add some tests to verifies request/reply and also NA packet.

felix/dataplane/linux/proxy_neigh_mgr.go
MichalFupso: The hasFailedListener is still needed, when a listener fails the manager is not marked as dirty. hasFailedListener helps to determine if we should reconcile during completeDeferedWork

felix/dataplane/linux/proxy_neigh_mgr_test.go
MichalFupso: The node selection is driven by the lib/datastructures/hashring/hashring.go which contains test for adding/removing members. Might be an unnecessary test duplicate?

mazdakn: I'll check the existing test coverage.

felix/dataplane/linux/proxy_neigh_mgr_test.go
```diff
@@ -0,0 +1,735 @@
...
+		nl = newMockNetlinkForProxyNeigh()
+		arpClients = map[string]*mockARPClient{
+			"eth0": newMockARPClient(net.HardwareAddr{0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0x01}),
+			"eth1": newMockARPClient(net.HardwareAddr{0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0x02}),
+		}
+		mgr = newTestProxyNeighManager(nl, arpClients)
+	})
+
+	AfterEach(func() {
+		mgr.cancel()
```
mazdakn: I think calling `cancel` here does not stop go routines, so we are leaking them here. The listener loop relies on the 1s timeout to check the context. We should do two things:
- Add a timeout to the mocked version of arp/ndp client.
- Add a `stop` method to manager which basically stops all listeners.
- Call the `stop` method after each tests.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,807 @@
...
+		nf = func(ifaceName string) (ndpConn, net.HardwareAddr, error) {
+			ifi, err := net.InterfaceByName(ifaceName)
+			if err != nil {
+				return nil, nil, err
+			}
+			conn, _, err := ndp.Listen(ifi, ndp.Unspecified)
+			if err != nil {
+				return nil, nil, err
+			}
+			// Set the IPMPV6 filter to only deliver Neighbor Solicitations
```
mazdakn:
```suggestion
			// Set the ICMPV6 filter to only deliver Neighbor Solicitations
```

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,807 @@
...
+	WriteTo(ndp.Message, *ipv6.ControlMessage, netip.Addr) error
+	SetReadDeadline(time.Time) error
+	Close() error
+}
+
+type arpClientFactory func(ifaceName string) (arpClient, net.HardwareAddr, error)
+type ndpConnFactory func(ifaceName string) (ndpConn, net.HardwareAddr, error)
+
+// ifaceListener manages a raw socket listener for a single host interface.
+type ifaceListener struct {
```
mazdakn: nit: it's better to place the struct and it's method close to each other which  make reading it better.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,807 @@
...
+// startListener opens a raw socket on the given interface and starts a
+// listener goroutine.
+func (m *proxyNeighManager) startListener(ifaceName string) error {
+	l := &ifaceListener{
+		ifaceName: ifaceName,
+		done:      make(chan struct{}),
+	}
+
+	ctx, cancel := context.WithCancel(m.ctx)
+	l.cancel = cancel
```
mazdakn: The way it's shaping, I think it's better to add `ctx` field to the `ifaceListener` and initialize it here, instead of passing it as an argument to it's methods.

P.S.: I might have told before to remove it, but I think it make more sense now this way.

felix/dataplane/linux/proxy_neigh_mgr_test.go
```diff
@@ -0,0 +1,735 @@
...
+func (s netlinkshimStub) RuleDel(rule *netlink.Rule) error                    { return nil }
+func (s netlinkshimStub) Delete()                                             {}
+func (s netlinkshimStub) NeighAdd(neigh *netlink.Neigh) error                 { return nil }
+func (s netlinkshimStub) NeighList(linkIndex, family int) ([]netlink.Neigh, error) {
+	return nil, nil
+}
+func (s netlinkshimStub) NeighSet(a *netlink.Neigh) error         { return nil }
+func (s netlinkshimStub) NeighDel(a *netlink.Neigh) error         { return nil }
+func (s netlinkshimStub) SetSocketTimeout(to time.Duration) error { return nil }
+func (s netlinkshimStub) SetStrictCheck(b bool) error             { return nil }
```
mazdakn: You should try to use the existing netlink mock here : https://github.com/projectcalico/calico/blob/2676a70709bb6f54a349a44a30f85a0272a8f1e6/felix/netlinkshim/mocknetlink/netlink.go#L284

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,817 @@
...
+
+	desiredByIface := m.buildDesiredState()
+	err := m.reconcileListeners(desiredByIface)
+	m.publishDesiredIPs(desiredByIface)
+
+	return err
+}
+
+// Stop tears down every listener (cancelling its goroutine and closing its raw
+// socket) and cancels the manager's context. Used for clean shutdown by tests
```
mazdakn: These comments looks redundant. The method itself is self explanatory.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,817 @@
...
+func (m *proxyNeighManager) Stop() {
+	for ifaceName, l := range m.listeners {
+		l.stop()
+		delete(m.listeners, ifaceName)
+	}
+	m.cancel()
+}
+
+// hasFailedListener reports whether any listener goroutine has flagged itself
+// failed (and so needs dropping and recreating by reconcileListeners).
```
mazdakn: same about this comment.

felix/dataplane/linux/proxy_neigh_mgr_test.go
```diff
@@ -0,0 +1,902 @@
...
+
+	"github.com/projectcalico/calico/felix/netlinkshim/mocknetlink"
+	"github.com/projectcalico/calico/felix/proto"
+	"github.com/projectcalico/calico/felix/rules"
+	"github.com/projectcalico/calico/libcalico-go/lib/set"
+)
+
+type mockNetlinkForProxyNeigh struct {
+	*mocknetlink.MockNetlinkDataplane
+}
```
mazdakn: we don't need this struct anymore. You can use `MockNetlinkDataplane` directly.

felix/dataplane/linux/proxy_neigh_mgr_test.go
```diff
@@ -363,6 +379,25 @@ func getDesiredIPs(mgr *proxyNeighManager, ifaceName string) set.Set[string] {
 	return *d
 }
 
+// injectARPRequest pushes an ARP "who-has targetIP" request (from senderHW /
+// senderIP) onto the listener's read channel, simulating a request arriving on
+// the wire.
+func injectARPRequest(c *mockARPClient, senderHW net.HardwareAddr, senderIP, targetIP string) {
```
mazdakn: nit: make this a method of `mockARPClient`

felix/dataplane/linux/proxy_neigh_mgr_test.go
```diff
@@ -363,6 +379,25 @@ func getDesiredIPs(mgr *proxyNeighManager, ifaceName string) set.Set[string] {
...
+	pkt, err := arp.NewPacket(arp.OperationRequest, senderHW, netip.MustParseAddr(senderIP),
+		make(net.HardwareAddr, 6), netip.MustParseAddr(targetIP))
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	c.reads <- pkt
+}
+
+// injectNS pushes an IPv6 Neighbor Solicitation for targetIP, arriving from src,
+// onto the listener's read channel.
+func injectNS(c *mockNDPConn, src, targetIP string) {
```
mazdakn: same as this one.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,817 @@
...
+	ifaceName string
+	// desired is the set of IPs this listener answers ARP/NDP for. Built fresh
+	// and swapped atomically; never mutated after publishing.
+	desired atomic.Pointer[set.Set[string]]
+	arpCli  arpClient // IPv4 only
+	ndpCli  ndpConn   // IPv6 only
+	hwAddr  net.HardwareAddr
+	cancel  context.CancelFunc
+	done    chan struct{}
+	failed  atomic.Bool
```
mazdakn: We should add a `readTimeout` field to ifaceListener and to the manager itself. Then we should init the value when creating ifaceListener based on the value of manager. 

This would allow to set it to a lower value like 10ms for unit tests.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,817 @@
...
+		if pkt.Operation != arp.OperationRequest {
+			continue
+		}
+
+		targetIP := pkt.TargetIP.String()
+		desired := l.desired.Load()
+		if desired == nil || !(*desired).Contains(targetIP) {
+			continue
+		}
+
```
mazdakn, line 638: We need to ignore our own GARP and prevent responding to it. There are two options:
- Check ARP request details. If the address is in desired set and mac address is equal to device, we can skip it.
- Setting `PACKET_IGNORE_OUTGOING` on the socket, so all outgoing packets are filtered by socket, and not sent to userspace. This is a better option, since it has some performance gain. 

It would be great if we can add the second one. It's doable by using `arp.New` instead of `arp.Dial` and passing a `packet.Conn` object:

```
conn, err := packet.Listen(ifi, packet.Raw, etherTypeARP, nil)
  ...
  if err := setIgnoreOutgoing(conn); err != nil {
      logrus.WithError(err).WithField("iface", ifaceName).Debug(
          "Failed to set PACKET_IGNORE_OUTGOING; falling back to userspace filtering")
  }
  c, err := arp.New(ifi, conn)
```

```
  func setIgnoreOutgoing(conn *packet.Conn) error {
        rc, err := conn.SyscallConn()
        if err != nil {
                return err
        }
        var serr error
        if err := rc.Control(func(fd uintptr) {
                serr = unix.SetsockoptInt(int(fd), unix.SOL_PACKET, unix.PACKET_IGNORE_OUTGOING, 1)
        }); err != nil {
                return err
        }
        return serr
  }
```

mazdakn (changes requested):

felix/dataplane/linux/proxy_neigh_mgr.go
mazdakn: Actually, might worth adding both. Old kernels does not support `PACKET_IGNORE_OUTGOING`.
Also it does not hurt to check requests more before responding.

felix/fv/proxy_neigh_test.go
```diff
@@ -0,0 +1,346 @@
...
+// LB VIP allocation
+//
+// FV does not run kube-controllers, so we hand-pick a VIP from the LB pool
+// and write it directly into Service.Status.LoadBalancer.Ingress via the
+// status sub-resource. kubernetesServiceToProto (felix/calc/dataplane_passthru.go)
+// reads the VIP from the status field, so the calc graph emits a real
+// ServiceUpdate just as it would in a production cluster.
+
+const (
+	netshootImage = "docker.io/nicolaka/netshoot:v0.13"
```
mazdakn: we should not pull a new image during test. You can add any tool you need for the test to the felix image (which is used for fv) here: https://github.com/projectcalico/calico/blob/master/felix/docker-image/Dockerfile

felix/fv/proxy_neigh_test.go
```diff
@@ -0,0 +1,346 @@
...
+		)
+
+		BeforeEach(func() {
+			// Detect the docker bridge subnet so we can derive pool CIDRs
+			// that fall inside the host's L2 segment. A missing IPv4 subnet
+			// means the FV host's docker is misconfigured — fail loudly
+			// rather than silently skip.
+			bridgeV4, err := inspectDockerNetworkV4(dockerBridgeName)
+			Expect(err).NotTo(HaveOccurred(), "inspecting docker bridge")
+			Expect(bridgeV4).NotTo(BeNil(), "docker bridge network %q has no IPv4 subnet", dockerBridgeName)
```
mazdakn: We don't need to do this. Just use `ip` tool on any of felixes to get the address of eth0.

felix/fv/proxy_neigh_test.go
```diff
@@ -0,0 +1,346 @@
...
+			podIP, err := pickIPInPool(wlPoolCIDR, 10)
+			Expect(err).NotTo(HaveOccurred())
+
+			w := workload.New(tc.Felixes[0], "wl-pod", "default", podIP, "8080", "tcp")
+			Expect(w.Start(infra)).To(Succeed())
+			w.ConfigureInInfra(infra)
+
+			By(fmt.Sprintf("ARPing pod IP %s from %s", podIP, extL2.Name))
+			Eventually(func() error {
+				return arpingFrom(extL2, podIP)
```
mazdakn: why do we need arp ping? By using normal connectivity checker we should be able to verify the same behaviour. In the absence of static routes between a pod and external workload, connectivity normally fails. With this feature, it should just work. A normal ping initially send an arp to get next hop mac address, the new manager responds, and connectivity works.

mazdakn (changes requested):
I think the FV test can be simpler. Also we should run the FV in a dual stack setup to test both v4 and v6 mechanism.

node/tests/k8st/tests/proxy_neigh_test.go
```diff
@@ -0,0 +1,602 @@
...
+// proxy_neigh_test.go is a kind-only system test for Felix's
+// LocalSubnetL2Reachability feature (proxy ARP/NDP). It verifies that an
+// L2-adjacent client outside the cluster can reach a pod IP or LoadBalancer VIP
+// that lives in the host's L2 subnet — which only works if Felix answers the
+// ARP (IPv4) / NDP (IPv6) request on behalf of the hosting/owning node.
+//
+// This deliberately lives in the kind "k8st" system-test suite rather than the
+// cross-platform e2e suite: it requires direct access to the local docker
+// daemon (to attach an L2-adjacent peer to the "kind" docker network), which
+// the e2e suite (run against arbitrary clusters via kubeconfig) cannot provide.
```
mazdakn: unnecessary comment.

felix/fv/proxy_neigh_test.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	out := make(net.IP, len(ip))
+	copy(out, ip)
+	out[len(out)-1] += byte(idx)
+	return out.String(), nil
+}
+
+// createNoEncapPool creates an IPPool with the given CIDR, AllowedUses and no
+// encapsulation. proxy_neigh_mgr.isNoEncapPool requires both ipipMode and
+// vxlanMode to be Never.
```
mazdakn: The last line looks unnecessary.

node/tests/k8st/tests/proxy_neigh_test.go
```diff
@@ -0,0 +1,602 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	t.Cleanup(cancel)
+
+	cli := newClient(g)
+
+	// Find the docker network the cluster's nodes are on (kind defaults to "kind"
+	// but honours KIND_EXPERIMENTAL_DOCKER_NETWORK, so it isn't fixed) and derive
+	// the two pool CIDRs from its subnet for this family.
+	network, err := detectKindNetwork(t)
+	g.Expect(err).NotTo(HaveOccurred(), "detecting the cluster's docker network")
```
mazdakn: I think it's OK to hard code the name here. We don't expect a change, so would be better to not try to detect it at run time.

node/tests/k8st/tests/proxy_neigh_test.go
```diff
@@ -0,0 +1,602 @@
...
+)
+
+const (
+	// extL2Iface is the interface name inside the netshoot peer container.
+	extL2Iface = "eth0"
+
+	// netshootImage carries arping, ping6/ip (for NDP) and curl for the L2 peer.
+	netshootImage = "docker.io/nicolaka/netshoot:v0.13"
+	// echoImage serves /clientip over HTTP; agnhost is the standard k8s test image.
+	echoImage = "registry.k8s.io/e2e-test-images/agnhost:2.45"
```
mazdakn: we use nginx in other tests, defined as `NginxImage`. Probably better to just use it to reduce the number of image fetch during tests.

node/tests/k8st/tests/proxy_neigh_test.go
```diff
@@ -0,0 +1,602 @@
...
+		g.Eventually(func() error {
+			_, err := peer.Arping(parsed, 3)
+			return err
+		}, "60s", "2s").Should(Succeed(), "ARP probe for %s never got a reply", ip)
+	} else {
+		g.Eventually(func() error {
+			_, err := peer.NeighborSolicit(parsed)
+			return err
+		}, "60s", "2s").Should(Succeed(), "NDP probe for %s never got an advertisement", ip)
+	}
```
mazdakn: I think we should just use a simple connectivity check like curl, and remove these checks.

mazdakn (changes requested):

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,929 @@
...
+		if !ok {
+			continue
+		}
+		// Atomically swap in the new desired set; the previous pointer tells
+		// us which IPs are new and which are gone.
+		old := l.desired.Swap(&desired)
+		for ip := range desired.All() {
+			if old == nil || !(*old).Contains(ip) {
+				l.joinNDPGroup(ip)
+				l.sendGARP(ip)
```
mazdakn: we should probably call here based on ipVersion:

```
if m.ipVersion == 4 {
  m.sendGARP()
} else {
  m.joinNDPGroup()
  m.sendUNA()
}
```

Then change `sendGARP` to only send GARP for ipv4.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,929 @@
...
+			}
+		}
+		// Drop multicast subscriptions for IPs we no longer answer for.
+		if old != nil {
+			for ip := range (*old).All() {
+				if !desired.Contains(ip) {
+					l.leaveNDPGroup(ip)
```
mazdakn, line 535: It looks a bit confusing when you read this from ipv4 standpoint, and see NDP here.
better to check ipversion here, and remove `npdCli == nil` check in both `join/leaveNDPGroup`.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,929 @@
...
+// single socket read before waking to re-check for context cancellation.
+// A read that hits the deadline simply loops
+const readDeadlineInterval = 1 * time.Second
+
+// ipv6AllNodesMulticast is the IPv6 all-nodes link-local multicast address
+const ipv6AllNodesMulticast = "ff02::1"
+
+// etherTypeARP is the EtherType for ARP. We open the raw ARP socket directly
+// (rather than via arp.Dial) so we can set PACKET_IGNORE_OUTGOING on it.
+const etherTypeARP = 0x0806
```
mazdakn: no need to define this. You can use `ethernet.EtherTypeARP`.

node/tests/k8st/tests/proxy_neigh_test.go
```diff
@@ -0,0 +1,491 @@
...
+// L2-adjacent client outside the cluster can reach a pod IP or LoadBalancer VIP
+// that lives in the host's L2 subnet — which only works if Felix answers the
+// ARP (IPv4) / NDP (IPv6) request on behalf of the hosting/owning node.
+//
+// Like the rest of the k8st Go suite it is compiled into node/bin/k8st.test and
+// run against a live kind cluster by `make -C node kind-k8st-run-test` (select
+// it on its own with K8ST_GO_TO_RUN=TestProxyNeigh). Run directly against an
+// existing cluster with:
+//
+//	KUBECONFIG=... go test ./node/tests/k8st/tests -run TestProxyNeigh -v
```
mazdakn: These 6 lines are unnecessary.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,921 @@
...
+		if !ok {
+			continue
+		}
+		// Atomically swap in the new desired set; the previous pointer tells
+		// us which IPs are new and which are gone.
+		old := l.desired.Swap(&desired)
+		for ip := range desired.All() {
+			if old == nil || !(*old).Contains(ip) {
+				if m.ipVersion == 6 {
+					l.joinNDPGroup(ip)
```
mazdakn, line 524: You can remove `SendGARP` method and call the methods it's calling directly here.

```
if m.ipVersion == 6 {
  l.sendGARPv6(ip)
  l.jointNGPGroup(ip)
} else {
  l.sendGARPv4(ip)
}
```

this way you can remove `sendGARP` method, and also parse ip address once in this method and pass it on.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,921 @@
...
+	}
+	if l.arpCli != nil {
+		l.sendGARPV4(addr)
+	} else {
+		l.sendGARPV6(addr)
+	}
+}
+
+// sendGARPV4 sends a gratuitous ARP for ip using the listener's raw socket.
+func (l *ifaceListener) sendGARPV4(ip netip.Addr) {
```
mazdakn: nit: we should just name this `sendGARP`. GARP is only ipv4. For ipv6, we should rename the method to `SendUNA`

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,921 @@
...
+		logrus.WithFields(logrus.Fields{
+			"iface": l.ifaceName,
+			"ip":    ip,
+		}).Debug("Sent gratuitous ARP")
+	}
+}
+
+// sendGARPV6 sends an unsolicited Neighbor Advertisement for addr using the
+// listener's raw socket.
+func (l *ifaceListener) sendGARPV6(addr netip.Addr) {
```
mazdakn: nit: let's rename this to `sengUNA` to align with ipv6 term.

felix/dataplane/linux/proxy_neigh_mgr.go
mazdakn: @MichalFupso WDYT about this?

mazdakn (commented):
Some new nits.

mazdakn:
@MichalFupso you need to pick master changes. I merged a PR that renames `k8sutils` to just `utils`. It should be a easy to fix the conflict.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -0,0 +1,921 @@
...
+			logrus.WithError(err).WithField("iface", l.ifaceName).Warn("Failed to close NDP client")
+		}
+	}
+	<-l.done
+	logrus.WithField("iface", l.ifaceName).Info("Stopped proxy neighbor listener")
+}
+
+// runARPListener listens for ARP requests on a raw socket and replies for
+// IPs in the desired set.
+func (l *ifaceListener) runARPListener() {
```
mazdakn, line 616: both `runARPListener` and `runNDPListener` are the critical path in this manager. IMO, we should move the loop and common error handling to one method. For example:

```
func (l *ifaceListener) runListener(proto string, setDeadline func(time.Time) error, readAndRespond func() error) {
        defer close(l.done)

        logCtx := logrus.WithFields(logrus.Fields{"iface": l.ifaceName, "proto": proto})
        for {    
                if l.ctx.Err() != nil {
                        logCtx.Debug("Listener stopping: context cancelled")
                        return
                }
                 
                if err := setDeadline(time.Now().Add(l.readTimeout)); err != nil {
                        logCtx.WithError(err).Warn("Failed to set read deadline")
                }
         
                if err := readAndRespond(); err != nil {
                        if l.ctx.Err() != nil {
                                logCtx.Debug("Listener stopping: context cancelled during read")
                                return
                        }
                        if netErr, ok := errors.AsType[net.Error](err); ok && netErr.Timeout() {
                                continue
                        }
                        // Unrecoverable socket error: flag this listener so the manager
                        // drops it and recreates a fresh one on the next reconcile.
                        logCtx.WithError(err).Warn("Listener read failed; recreating listener")
                        l.failed.Store(true)
                        return
                }
        }
  }
```

and then `runARPListener` become:

```
func (l *ifaceListener) runARPListener() {
  l.runListener("ARP", l.arpCli.SetReadDeadline, func() {
       pkt, _, err := l.arpCli.Read() 
        if err != nil {
                return err
        }       
        if pkt.Operation != arp.OperationRequest {
                return nil
        }       
        // Ignore our own frames: the raw socket also receives what we send. The
        // kernel normally filters these via PACKET_IGNORE_OUTGOING; this guards
        // kernels that lack the option.
        if bytes.Equal(pkt.SenderHardwareAddr, l.hwAddr) {
                return nil
        }       
        targetIP := pkt.TargetIP.String()
        if !l.wantsIP(targetIP) {
                return nil
        }       
        if err := l.arpCli.Reply(pkt, l.hwAddr, pkt.TargetIP); err != nil {
                logrus.WithError(err).WithFields(logrus.Fields{
                        "iface": l.ifaceName, "ip": targetIP,
                }).Warn("ARP listener: failed to send reply")
        }
        return nil
  }
}
```

felix/config/config_params.go
```diff
@@ -375,6 +375,11 @@ type Config struct {
 	// IPs are always programmed, regardless of this setting.
 	FloatingIPs string `config:"oneof(Enabled,Disabled);Disabled"`
 
+	// LocalSubnetL2Reachability controls whether Felix programs per-IP proxy ARP/NDP
+	// neighbor entries on host interfaces for local pod IPs and LB VIPs that overlap the
+	// host subnet. [Default: Disabled]
+	LocalSubnetL2Reachability string `config:"oneof(Disabled,PodsAndLoadBalancers);Disabled"`
```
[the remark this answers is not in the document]
mazdakn: @MichalFupso this looks legit. The comments needs an update.

felix/dataplane/linux/proxy_neigh_mgr.go
MichalFupso: I think we should not switch them, currently the listeners are created in reconcileListeners and then we update the desiredIPSets and send garp. If we would have switched the order the listener is created after we check if we need to update IPs and it will take an extra cycle to advertise those

mazdakn (approved):
@MichalFupso I think we should read from/write to a raw socket from the listener goroutine. It seems writing to a socket from different go routines is not safe.
This is not an issue here, because we send GARP/UNA once before starting a listener. But for future improvements we should fix it.

mazdakn:
@MichalFupso please squash and merge this PR, since it already has 70+ commits.

PR #12455, opened 2026-04-14
Fix spurious Felix restarts from transient encapsulation changes

tomastigera:
## Description

When the calc graph's EncapsulationResolver reports an encapsulation change
(e.g., VXLAN enabled→disabled), the daemon now re-verifies by listing IP pools
directly from the datastore before restarting. If the fresh computation matches
the startup state, the change was transient (e.g., from a watcher resync race)
and is ignored.

This fixes a flaking FV test (`_BPF-SAFE_ VXLAN topology ... with mismatched
MTU interface pattern / should configure MTU correctly based on IP pools`) where
a spurious pool update with empty VXLANMode arrived in the calc graph, causing
Felix to restart without VXLAN even though the datastore pool had
VXLANMode=Always.

### Root cause from log analysis

1. A transient IP pool update arrives in the calc graph with `VXLANMode=""`
   (NO_ENCAP) even though the Kubernetes pool has `VXLANMode=Always`
2. The EncapsulationResolver computes `VXLANEnabled=false` and sends it to the
   daemon
3. The daemon sees a mismatch with its startup encapsulation and restarts Felix
4. The restarted Felix runs without VXLAN — never programs `vxlan.calico` in BPF
5. The test times out after 60s waiting for BPF program attachment

The spurious update coincides with a config_batcher flush, suggesting a
syncer/watcher resync race condition.

### Changes

- **`felix/daemon/daemon.go`**: Add `confirmEncapChanged()` method that re-lists
  IP pools from the datastore and recomputes encapsulation before committing to a
  restart. On error (e.g., datastore unreachable), falls back to restarting.
- **`felix/calc/encapsulation_resolver.go`**: Add pool counts to encapsulation
  change log messages for better diagnostics.

## Release Note
```release-note
Felix no longer restarts unnecessarily due to transient encapsulation state changes from watcher resyncs. 
```

### DIFF
--- a/felix/calc/encapsulation_resolver.go
+++ b/felix/calc/encapsulation_resolver.go
@@ -98,6 +98,9 @@ func (r *EncapsulationResolver) triggerCalculation() {
 			"newVXLANEnabled":   newEncap.VXLANEnabled,
 			"oldVXLANEnabledV6": r.config.Encapsulation.VXLANEnabledV6,
 			"newVXLANEnabledV6": newEncap.VXLANEnabledV6,
+			"numIPIPPools":      len(r.encapCalc.ipipPools),
+			"numVXLANPools":     len(r.encapCalc.vxlanPools),
+			"numVXLANPoolsV6":   len(r.encapCalc.vxlanPoolsv6),
 		}).Info("EncapsulationResolver: Encapsulation changed.")
 	}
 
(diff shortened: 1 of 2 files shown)


tomastigera:
replaced by https://github.com/projectcalico/calico/pull/12484

PR #12456, opened 2026-04-14
networking-calico: defensive fixes for silent elector death and status backlog

nelljerram:
For CORE-12651 and CI-1892

Three fixes for the issues investigated under CI-1892.  None of them changes the election wire protocol or the existing behaviour in the common case; they add extra checks that catch known failure modes.

1. Elector.healthy()

   self._master is a local Python flag that only gets cleared if the election greenlet exits normally via _attempt_step_down() or the refresh-loop finally clause.  If the greenlet dies silently - e.g. due to an eventlet-level issue that drops the frame without unwinding Python exceptions - _master stays True indefinitely and the rest of the driver keeps running master-only work (periodic resync, compaction, StatusWatcher) on a node whose etcd lease has long since expired, causing split-brain.

   Elector now stamps self._last_refresh on each successful lease refresh, and healthy() returns True only if the greenlet is still alive AND the lease has been refreshed within the last TTL. mech_calico.py callsites that decide whether to do master-only work switch from master() to healthy().

2. Elector.confirmed_master()

   Before each iteration of periodic_resync_thread, also re-read the election key from etcd and confirm the value matches our id_string.  If etcd disagrees, clear _master locally and skip the resync cycle.  This defends against any residual disagreement between our in-process flag and etcd's ground truth, at the cost of one extra etcd GET per resync tick (negligible vs. the resync itself).

3. StatusWatcher stale-update WARNING

   When an incoming Felix status update has a "time" field more than 5 mins in the past, log a rate-limited WARNING so operators see the backlog building up long before it grows to hours.  Skipped during initial-snapshot replay to avoid false positives.  Complements the ReportingIntervalSecs / agent_down_time config tuning that was recommended to the customer.

Also promote the surviving LOG.debug calls in election.py to LOG.info so election state transitions and refresh activity are visible in production logs without toggling debug logging.

Unit tests added for healthy(), confirmed_master() and the stale status check.

**Release note:**
```release-note
Bugfix: Make OpenStack driver periodic processing more resilient to an eventlet bug that can result in silent greenlet death.
```

### DIFF
--- a/networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
+++ b/networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
@@ -28,6 +28,7 @@ import contextlib
 from datetime import datetime, timedelta
 import os
 import re
+import sys
 import threading
 import uuid
 from functools import wraps
@@ -341,6 +342,11 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
         # Last resync completion time
         self.last_resync_time = datetime.now()
 
+        # List of (name, greenlet) for the long-running worker greenlets
+        # spawned by _post_fork_init.  Used by _check_greenlets_alive() to
+        # detect silent greenlet death.
+        self._greenlets = []
+
         # Tell the monkeypatch where we are.
         global mech_driver
         assert mech_driver is None
@@ -462,13 +468,33 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
                 # We deliberately do this last, to ensure that all of the setup
                 # above is complete before we start running.
                 self._epoch += 1
-                eventlet.spawn(self.resync_monitor_thread, self._epoch)
-                eventlet.spawn(self.periodic_resync_thread, self._epoch)
+                self._greenlets = []
+                self._greenlets.append((
+                    "resync_monitor",
+                    eventlet.spawn(self.resync_monitor_thread, self._epoch),
+                ))
+                self._greenlets.append((
+                    "periodic_resync",
+                    eventlet.spawn(self.periodic_resync_thread, self._epoch),
+                ))
                 if cfg.CONF.calico.etcd_compaction_period_mins > 0:
-                    eventlet.spawn(self.periodic_compaction_thread, self._epoch)
-                eventlet.spawn(self._status_updating_thread, self._epoch)
-                for _ in range(cfg.CONF.calico.num_port_status_threads):
-                    eventlet.spawn(self._loop_writing_port_statuses, self._epoch)
+                    self._greenlets.append((
+                        "periodic_compaction",
+                        eventlet.spawn(
+                            self.periodic_compaction_thread, self._epoch
+                        ),
+                    ))
+                self._greenlets.append((
+                    "status_updating",
+                    eventlet.spawn(self._status_updating_thread, self._epoch),
+                ))
+                for i in range(cfg.CONF.calico.num_port_status_threads):
+                    self._greenlets.append((
+                        "port_status_%d" % i,
+                        eventlet.spawn(
+                            self._loop_writing_port_statuses, self._epoch
+                        ),
+                    ))
             else:
                 LOG.info(
                     "PID %s: Not a voting participant; "
@@ -482,6 +508,29 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
                 "Calico mechanism driver initialisation done in process %s", current_pid
             )
 
+    def _check_greenlets_alive(self):
+        """Detect if any long-running worker greenlet has silently died.
+
+        Under eventlet, a greenlet can occasionally die without unwinding
+        its Python frames (e.g. due to a hub-level error), leaving no
+        traceback in the log and no state cleanup.  This method provides
+        mutual watchdogging: each of the driver's long-running loops
+        calls it periodically to verify that the others are still alive.
+
+        If a dead greenlet is found, we log an error and exit.  The
+        process manager (systemd) will restart neutron-server, which is
+        the safest recovery — the same approach the elector already uses
+        for its own unhandled-exception path.  See CI-1892.
+        """
+        for name, gt in self._greenlets:
+            if gt.dead:
+                LOG.error(
+                    "Worker greenlet %r has unexpectedly died; exiting so "
+                    "that the process manager can restart neutron-server.",
+                    name,
+                )
+                sys.exit(1)
+
     @logging_exceptions(LOG)
     def _status_updating_thread(self, expected_epoch):
         """_status_updating_thread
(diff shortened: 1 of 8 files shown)


networking-calico/networking_calico/plugins/ml2/drivers/calico/status.py
```diff
@@ -19,13 +19,26 @@
...
+
+# If a Felix status update we receive from etcd has a "time" field more than
+# this many seconds in the past, we are running behind and should warn the
+# operator.  Felix writes status updates every 30s by default, so anything
+# materially above that indicates a processing backlog.
+STALE_STATUS_WARN_SECS = 300
+
+# Rate-limit stale-status warnings to at most one per this many seconds, to
+# avoid flooding the log when every update in a large batch is stale.
+STALE_STATUS_WARN_INTERVAL_SECS = 300
```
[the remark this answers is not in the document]
nelljerram: Updated in PR description, and will also push updated commit message.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_status.py
```diff
@@ -0,0 +1,91 @@
...
+        with mock.patch.object(status.LOG, "warning") as m_warn:
+            self.watcher._check_for_stale_status("host1", {"time": fresh})
+        m_warn.assert_not_called()
+        self.assertEqual(0.0, self.watcher._last_stale_warn)
+
+    def test_stale_update_warns(self):
+        stale = self._fmt(datetime.now(tz=timezone.utc) - timedelta(hours=1))
+        with mock.patch.object(status.LOG, "warning") as m_warn:
+            self.watcher._check_for_stale_status("host1", {"time": stale})
+        m_warn.assert_called_once()
```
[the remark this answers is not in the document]
nelljerram: https://github.com/projectcalico/calico/pull/12456/commits/f34c2b3026bcccc0ed5229d789c783c4a11f99d6

networking-calico/networking_calico/plugins/ml2/drivers/calico/status.py
```diff
@@ -135,6 +154,57 @@ def _on_status_set(self, response, hostname):
...
+            # to a +00:00 offset for datetime.fromisoformat (which has only
+            # accepted the bare "Z" suffix since Python 3.11).
+            status_time = datetime.fromisoformat(status_time_str.replace("Z", "+00:00"))
+        except ValueError:
+            LOG.warning(
+                "Could not parse status time %r for host %s",
+                status_time_str,
+                hostname,
+            )
+            return
```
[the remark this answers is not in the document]
nelljerram: https://github.com/projectcalico/calico/pull/12456/commits/f34c2b3026bcccc0ed5229d789c783c4a11f99d6

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -482,6 +519,29 @@ def _post_fork_init(self, voting=False):
                 "Calico mechanism driver initialisation done in process %s", current_pid
             )
 
+    def _check_greenlets_alive(self):
```
zhanz1, line 522: With the current code (without this change), even though we might have multiple masters doing the same job, the cluster is actually still functional (e.g., it's ok to have two masters writing agent/status updates). My main worry here (with this change) is what if the elector threads on all neutron-servers die at roughly the same time? In this case, everyone will stop doing the job until we restart it here. This may take a LONG time, as this function is called only when we start to watch statuses or in resync monitor (which will sleep based on the configured time).

It's unfortunate that these threads are not reliable and none of the solutions on top my mind handles this well (e.g., have yet another thread to monitor this (or alter the existing monitor thread to be more general-purposed)), but lemme know what do you think?

nelljerram: @zhanz1 Thanks for looking at this.  I entirely agree that this is a theoretical problem:

> what if the elector threads on all neutron-servers die at roughly the same time? 

But do you think there is anything in this PR that has made that more likely?  I don't think the PR does that, but perhaps I have missed something.

zhanz1: Ah I didn't realize that the status update and resync monitor thread is constantly (every 5 secs) checking if the threads are alive if it's not the leader, my bad. I was thinking could there be a big gap between all three electors die till at least one the neutron-servers restarts.

No it doesn't make `the elector threads on all neutron-servers die at roughly the same time` more likely from my reading. I was simply thinking about edge cases because we don't have enough information on why exactly the threads die.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -482,6 +519,29 @@ def _post_fork_init(self, voting=False):
...
+        traceback in the log and no state cleanup.  This method provides
+        mutual watchdogging: each of the driver's long-running loops
+        calls it periodically to verify that the others are still alive.
+
+        If a dead greenlet is found, we log an error and exit.  The
+        process manager (systemd) will restart neutron-server, which is
+        the safest recovery — the same approach the elector already uses
+        for its own unhandled-exception path.
+        """
+        for name, gt in self._greenlets:
```
zhanz1, line 536: I would be more aggressive - not trusting eventlet at all. 

If an elector is not doing it's job (for master, this would be not updating `self._last_refresh`, for non-master, this would be the last time they watched the vote key change (so something like self._last_checked)), do a restart.

If the port status update worker thread is not consuming events from the queue, do a restart.

nelljerram:
**Re-assessing this PR after #12668 (the worker-split) merged.** The picture has changed enough that I'm splitting this into smaller, focused PRs and will close this one once they land.

**What was here:**

1. `Elector.healthy()` — defensive check against silent elector greenthread death.
2. `Elector.confirmed_master()` — etcd ground-truth re-read before each `periodic_resync_thread` iteration.
3. Mutual cross-greenlet watchdog.
4. `StatusWatcher` stale-update WARNING.

**What's being carried forward:**

- **Stale-update WARNING** → #13024. Stands on its own merits independent of the mastership story — operators benefit from early warning of a Felix status backlog either way.
- **Mastership documentation** → #13023. The de facto split since #12668 — static `[calico] startup_resync` config for the one-shot startup resync, dynamic election for the continuous compaction and status-watching loops — was never written down. Documenting it on `get_workers()` so future readers don't have to reverse-engineer it from `is_master()` callsites.

**What's being dropped, and why:**

- **`Elector.healthy()`**: largely superseded. `is_master()` already does a time-based stale-refresh check and warns if the elector hasn't refreshed within `MASTER_TIMEOUT`. The only thing `healthy()` added was an explicit `_greenlet.dead` check — but if the greenlet died, the refresh would stop, and the time check catches that.
- **`Elector.confirmed_master()`**: targeted at `periodic_resync_thread`, which no longer exists post-#12668. The resync is now one-shot at process start. The blast radius of stale mastership in the remaining loops is much smaller: compaction has its own etcd-coordinated lock, and stale status-watcher writes are caught by `is_master()`'s time check.
- **Mutual cross-greenlet watchdog**: designed for the old "one process runs many threads" layout. With #12668's per-worker process split, each worker has one job; if its main greenlet dies the process is effectively dead and Neutron's supervisor respawns it.

Closing this once #13023 and #13024 merge.

PR #12464, opened 2026-04-14
Return tombstone KVPair for malformed ClusterNetworkPolicy instead of error

mazdakn:
Previously, K8sClusterNetworkPolicyToCalico returned hard errors for invalid tiers and missing subject selectors. These errors surfaced as WatchError in the KDD watch converter, terminating and resyncing the watch stream. If the same malformed policy kept reappearing, this caused an infinite resync loop.

Instead, log a warning and return a KVPair with the correct Key/Revision but nil Value (a tombstone). This keeps the watch stream healthy and signals downstream consumers to remove any previously-cached version of the policy. Also add a nil-Value guard in the list path's convertFunc.

Original review comment: https://github.com/projectcalico/calico/pull/12355#discussion_r3041481508
Follow up on https://github.com/projectcalico/calico/pull/12354

**Release note:**
```release-note
TBD
```

### DIFF
--- a/libcalico-go/lib/backend/k8s/resources/kubeclusternetworkpolicy.go
+++ b/libcalico-go/lib/backend/k8s/resources/kubeclusternetworkpolicy.go
@@ -103,6 +103,10 @@ func (c *clusterNetworkPolicyClient) List(ctx context.Context, list model.ListIn
 		if err != nil && !errors.As(err, &e) {
 			return nil, err
 		}
+		// Skip malformed policies that returned a nil-Value tombstone KVPair.
+		if kvp == nil || kvp.Value == nil {
+			return nil, nil
+		}
 		return []*model.KVPair{kvp}, nil
 	}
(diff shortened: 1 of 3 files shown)


libcalico-go/lib/backend/k8s/resources/kubeclusternetworkpolicy.go
```diff
@@ -103,6 +103,10 @@ func (c *clusterNetworkPolicyClient) List(ctx context.Context, list model.ListIn
 		if err != nil && !errors.As(err, &e) {
 			return nil, err
 		}
+		// Skip malformed policies that returned a nil-Value tombstone KVPair.
+		if kvp == nil || kvp.Value == nil {
+			return nil, nil
+		}
```
mazdakn, line 109: This seems unnecessary.

caseydavenport (approved):

Issue #11750, opened 2026-01-30
nft segfault on debian trixie


jsmouret:
It looks like there is a version incompatibility with nftables on Debian Trixie.

## Expected Behavior

No errors

## Current Behavior

Reconciliation loops with errors:
```
$ kubectl logs -n calico-system calico-node-htqgt

2026-01-30 21:11:15.382 [WARNING][47] felix/table.go 587: Failed to load maps state error=error listing nftables maps: failed to run nft: signal: segmentation fault ipVersion=0x4 table="calico"
2026-01-30 21:11:15.408 [WARNING][47] felix/table.go 587: Failed to load maps state error=error listing nftables maps: failed to run nft: signal: segmentation fault ipVersion=0x6 table="calico"
2026-01-30 21:11:35.014 [WARNING][47] felix/table.go 587: Failed to load maps state error=error listing nftables maps: failed to run nft: signal: segmentation fault ipVersion=0x6 table="calico"
2026-01-30 21:11:35.015 [WARNING][47] felix/table.go 587: Failed to load maps state error=error listing nftables maps: failed to run nft: signal: segmentation fault ipVersion=0x4 table="calico"
2026-01-30 21:11:35.072 [INFO][47] felix/summary.go 100: Summarising 6 dataplane reconciliation loops over 1m1s: avg=118ms longest=197ms (resync-calico-v4,resync-calico-v6)
```
which matches errors on the host:
```
$ dmesg
[1453664.183356] nft[2663996]: segfault at 0 ip 0000000000000000 sp 00007ffc3cc5fab8 error 14 likely on CPU 1 (core 1, socket 0)
[1453664.187459] Code: Unable to access opcode bytes at 0xffffffffffffffd6.
[1453684.179986] nft[2664094]: segfault at 0 ip 0000000000000000 sp 00007ffc329f4958 error 14 likely on CPU 0 (core 0, socket 0)
[1453684.184930] Code: Unable to access opcode bytes at 0xffffffffffffffd6.
[1453684.199519] nft[2664097]: segfault at 0 ip 0000000000000000 sp 00007ffd00953388 error 14 likely on CPU 1 (core 1, socket 0)
[1453684.203321] Code: Unable to access opcode bytes at 0xffffffffffffffd6.
```

## Possible Solution

Maybe upgrade calico image with a fresher nftables package?

host: `nftables v1.1.3 (Commodore Bullmoose #4)`
calico node: `nftables v1.0.4 (Lester Gooch #3)`

[...]

PR #12471, opened 2026-04-15
Use ListAll() to reduce nft subprocesses during resync

caseydavenport:
Felix's nftables resync reads object names from the dataplane before reconciling. It did this with separate `List("map")` and `List("chain")` calls, each spawning its own `nft` subprocess. This replaces those two with a single `ListAll()` call, halving the per-resync object-name listing from two nft subprocesses to one. `ListAll()` was added in knftables v0.0.21 (already on master) for exactly this: it runs `nft --json --terse list table <family> <table>` once and returns all object names grouped by type.

Also replace the unscoped `nft list ruleset` debug dump (fired on transaction error) with a table-scoped `nft list table` call. The unscoped command parses objects from every kernel nftables table, which can crash older nft binaries when another table contains udata written by a newer nft. See #11750 for details.

Same optimization kube-proxy made in https://github.com/kubernetes/kubernetes/pull/137501.

Benchmarked with the nftables dataplane benchmark (`BenchmarkResync`): ~20% fewer allocations and ~20% less memory churn per resync, consistent across the 100/500/2000-chain scales.

```release-note
None
```

### DIFF
--- a/felix/nftables/maps.go
+++ b/felix/nftables/maps.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 Tigera, Inc. All rights reserved.
+// Copyright (c) 2024-2026 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -46,7 +46,7 @@ type MapsDataplane interface {
 
 	MapUpdates() *MapUpdates
 	FinishMapUpdates(updates *MapUpdates)
-	LoadDataplaneState() error
+	LoadDataplaneState(ctx context.Context, mapNames []string) error
 }
 
 var _ MapsDataplane = &Maps{}
@@ -278,10 +278,9 @@ func (s *Maps) filterAndCanonicaliseMembers(mtype MapType, members map[string][]
 	return filtered
 }
 
-// tryResync attempts to bring our state into sync with the dataplane.  It scans the contents of the
-// maps in the dataplane and queues up updates to any maps that are out-of-sync.
-func (s *Maps) LoadDataplaneState() error {
-	// Log the time spent as we exit the function.
+// LoadDataplaneState resyncs map state using the provided list of map names
+// (typically obtained from a ListAll call by the caller).
+func (s *Maps) LoadDataplaneState(ctx context.Context, maps []string) error {
 	resyncStart := time.Now()
 	defer func() {
 		s.logCxt.WithFields(logrus.Fields{
@@ -292,23 +291,7 @@ func (s *Maps) LoadDataplaneState() error {
 		}).Debug("Finished Maps resync")
 	}()
 
-	// Clear the dataplane metadata view, we'll build it back up again as we scan.
 	s.mapNameToProgrammedMetadata.Dataplane().DeleteAll()
-
-	// Load from the dataplane. Update our Dataplane() maps with the actual contents
-	// of the data plane.
-	//
-	// For any map that doesn't match the desired data plane state, we'll queue up an update.
-	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
-	defer cancel()
-	maps, err := s.nft.List(ctx, "map")
-	if err != nil {
-		if knftables.IsNotFound(err) {
-			// Table doesn't exist - nothing to resync.
-			return nil
-		}
-		return fmt.Errorf("error listing nftables maps: %s", err)
-	}
 
 	// We'll process each map in parallel, so we need a struct to hold the results.
 	// Once knftables is augmented to support reading many maps at once, we can remove this.
@@ -377,7 +360,7 @@ func (s *Maps) LoadDataplaneState() error {
 
 		memberTracker := s.getOrCreateMemberTracker(mapName)
 		numExtrasExpected := memberTracker.PendingDeletions().Len()
-		err = memberTracker.Dataplane().ReplaceFromIter(func(f func(k MapMember)) error {
+		err := memberTracker.Dataplane().ReplaceFromIter(func(f func(k MapMember)) error {
 			for item := range elemsSet.All() {
 				f(item)
 			}
--- a/felix/nftables/maps.go
+++ b/felix/nftables/maps.go
@@ -291,6 +291,7 @@ func (s *Maps) LoadDataplaneState(ctx context.Context, maps []string) error {
 		}).Debug("Finished Maps resync")
 	}()
 
+	// Clear the dataplane metadata view, we'll build it back up again as we scan.
 	s.mapNameToProgrammedMetadata.Dataplane().DeleteAll()
 
 	// We'll process each map in parallel, so we need a struct to hold the results.
--- a/felix/nftables/table.go
+++ b/felix/nftables/table.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2024 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2026 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


felix/nftables/maps.go
```diff
@@ -292,24 +291,8 @@ func (s *Maps) LoadDataplaneState() error {
 		}).Debug("Finished Maps resync")
 	}()
 
-	// Clear the dataplane metadata view, we'll build it back up again as we scan.
```
caseydavenport, line 295: Unnecessary comment removal?

felix/nftables/table.go
```diff
@@ -642,8 +642,25 @@ func (t *NftablesTable) decrefChain(chainName string) {
 }
 
 func (t *NftablesTable) loadDataplaneState() {
-	// Sync maps.
-	if err := t.LoadDataplaneState(); err != nil {
+	// Fetch all object names from the dataplane in a single nft invocation.
+	// This replaces separate List("map") and List("chain") calls, halving
+	// the number of nft subprocesses during resync.
```
caseydavenport:
```suggestion
```

felix/nftables/table.go
```diff
@@ -642,8 +642,25 @@ func (t *NftablesTable) decrefChain(chainName string) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), t.contextTimeout)
+	defer cancel()
+	allObjects, err := t.nft.ListAll(ctx)
+	if err != nil {
+		if knftables.IsNotFound(err) {
+			t.logCxt.Debug("Table not found in dataplane, nothing to load.")
+		} else {
+			t.logCxt.WithError(err).Warn("Failed to list all nftables objects")
+		}
```
caseydavenport, line 666:
```suggestion
		}
		
```

felix/nftables/table.go
```diff
@@ -656,7 +673,7 @@ func (t *NftablesTable) loadDataplaneState() {
 
 	t.lastReadTime = t.timeNow()
 
-	dataplaneHashes := t.getHashesFromDataplane()
+	dataplaneHashes := t.getHashesFromDataplane(allObjects["chain"])
```
caseydavenport: no constant we can use here for "chain"?

felix/nftables/table.go
```diff
@@ -790,7 +810,9 @@ func (t *NftablesTable) getHashesFromDataplane() map[string][]string {
 }
 
 // attemptToGetHashesFromDataplane reads nftables state and extracts our rule hashes.
-func (t *NftablesTable) attemptToGetHashesFromDataplane() (hashes map[string][]string, err error) {
+// If chainNames is non-nil, it is used directly (from a prior ListAll call). If nil,
```
caseydavenport, line 826: Do we ever need the "nil" fallback? Are there any callers who don't pass in chain names?

felix/nftables/table.go
```diff
@@ -1213,7 +1241,7 @@ func (t *NftablesTable) CheckRulesPresent(chain string, rules []generictables.Ru
 	features := t.featureDetector.GetFeatures()
 	hashes := CalculateRuleHashes(chain, rules, features)
 
-	dpHashes := t.getHashesFromDataplane()
+	dpHashes := t.getHashesFromDataplane(nil)
```
caseydavenport, line 1260: Could use a brief comment here explaining why nil is passed.

felix/nftables/table.go
```diff
@@ -642,8 +649,29 @@ func (t *NftablesTable) decrefChain(chainName string) {
 }
 
 func (t *NftablesTable) loadDataplaneState() {
-	// Sync maps.
-	if err := t.LoadDataplaneState(); err != nil {
+	// List all of our table's objects in a single nft invocation, giving us the
+	// map and chain names without a separate List call for each type.
+	listCtx, cancel := context.WithTimeout(context.Background(), t.contextTimeout)
+	allObjects, err := t.nft.ListAll(listCtx)
+	cancel()
```
mazdakn: nit: calling `defer cancel` right after `context.WithTimeout` looks more idiomatic.

felix/nftables/table.go
```diff
@@ -41,6 +41,11 @@ import (
 const (
 	MaxChainNameLength = knftables.NameLengthMax
 	defaultTimeout     = 30 * time.Second
+
+	// Object type names as returned by knftables' ListAll, used to index its
+	// result map.
+	objectTypeChain = "chain"
+	objectTypeMap   = "map"
```
mazdakn, line 48: Doesn't knftables define these constants?

felix/nftables/table.go
```diff
@@ -642,8 +649,29 @@ func (t *NftablesTable) decrefChain(chainName string) {
 }
 
 func (t *NftablesTable) loadDataplaneState() {
-	// Sync maps.
-	if err := t.LoadDataplaneState(); err != nil {
+	// List all of our table's objects in a single nft invocation, giving us the
+	// map and chain names without a separate List call for each type.
+	listCtx, cancel := context.WithTimeout(context.Background(), t.contextTimeout)
+	allObjects, err := t.nft.ListAll(listCtx)
```
mazdakn, line 657: we should either define new counters for `ListAll`, or increment `List` counter.

felix/nftables/table.go
```diff
@@ -808,18 +841,21 @@ func (t *NftablesTable) attemptToGetHashesFromDataplane() (hashes map[string][]s
...
-			err = nil
-			return
+	if chainNames == nil {
+		// List chains separately, as chains may exist without rules.
+		countNumListCalls.Inc()
+		chainNames, err = t.nft.List(ctx, objectTypeChain)
+		if err != nil {
+			if knftables.IsNotFound(err) {
+				err = nil
+				return
```
mazdakn, line 856: This part is only refactored in this change, so nothing new. But I wonder if it make sense to log at least a debug message instead of ignoring the error. WDYT?

felix/nftables/table.go
```diff
@@ -1138,13 +1174,16 @@ func (t *NftablesTable) applyUpdates() error {
 		}
 
 		if err := t.runTransaction(tx); err != nil {
-			// Let's just print out the entire ruleset for debugging purposes.
-			cmd := t.newCmd("nft", "list", "ruleset")
+			// Dump our table's state for debugging. We scope this to our
+			// own table rather than using "nft list ruleset" to avoid
+			// parsing objects from other tables that may contain udata
+			// written by a newer nft, which can crash older nft binaries.
+			cmd := t.newCmd("nft", "list", "table", string(t.family), t.name)
```
mazdakn: worth defining a utility function for dumping rules here. I think we do something similar here as well: https://github.com/projectcalico/calico/blob/347e14d026873685ece3778b9deb982165a66ac4/felix/nftables/table.go#L920

mazdakn (commented):
The change itself LGTM. Just left a few questions and nits.

mazdakn (approved):
LGTM

felix/nftables/table.go
caseydavenport: Yeah, switched both to deferred cancel. They needed distinct names (`listCancel`/`mapsCancel`) - the cancel var was being reused for `mapsCtx`, so a plain `defer cancel` would've ended up cancelling the wrong context.

felix/nftables/table.go
caseydavenport: Doesn't look like it - knftables exports constants for families, chain types, hooks etc., but the object-type strings it takes in `List()` and keys `ListAll()`'s result by are bare literals throughout (`"chain"`, `"map"`, `"set"`). So nothing to reuse, kept our own constants.

felix/nftables/table.go
caseydavenport: Bumped `countNumListCalls` before the `ListAll` (and `countNumListErrors` on a non-NotFound failure). It spawns an nft subprocess like `List`, so it should show up in the same metrics.

felix/nftables/table.go
caseydavenport: Added a debug log there.

felix/nftables/table.go
caseydavenport: Pulled both dump sites into a `dumpTableState()` helper.

PR #12474, opened 2026-04-15
Fix kind cluster deploy script ordering bugs

caseydavenport:
Fixes a kind-deploy flake when CLUSTER_ROUTING=FELIX. The patch to the Installation resource triggers a rollout of the Calico apiserver, and the subsequent pod wait was enumerating pods from the old ReplicaSet. When those old pods terminated mid-wait, `kubectl wait` failed with "not found" and the deploy failed.

This gets rid of the PATCH altogether, and just uses a new values.yaml file

It also removes some redundant checks / logs from the script. 

```release-note
None
```


hack/test/kind/deploy_resources.sh
```diff
@@ -169,15 +169,18 @@ echo
 echo "Install Calico using the helm chart"
 ${HELM} install calico ${CHART} -f ${VALUES_FILE} -n tigera-operator --create-namespace
 
+echo "Install calicoctl as a pod"
+${kubectl} apply -f ${INFRA_DIR}/calicoctl.yaml
+echo
+
+# Patch the Installation resource before waiting for tigerastatus. This avoids a
+# race where the patch triggers a new rollout after pods have already been
+# enumerated by a wait, causing "not found" errors on old ReplicaSet pods.
```
caseydavenport:
```suggestion
```

hack/test/kind/deploy_resources.sh
```diff
@@ -196,7 +196,8 @@ if [ "$CALICO_API_GROUP" != "projectcalico.org/v3" ]; then
 fi
 
 echo "Wait for Calico to be ready..."
-wait_pod_ready -n calico-system -l k8s-app
+# Calico-system pods are covered by the tigerastatus/calico and tigerastatus/apiserver
+# waits above. kube-dns and calicoctl are not operator-managed, so wait on them directly.
 wait_pod_ready -l k8s-app=kube-dns -n kube-system
 wait_pod_ready calicoctl -n kube-system
```
caseydavenport: @Copilot these should actually be moved all the way to the end of the file - we have another loop that ensures every TigeraStatus is ready.

hack/test/kind/deploy_resources.sh
```diff
@@ -196,7 +196,8 @@ if [ "$CALICO_API_GROUP" != "projectcalico.org/v3" ]; then
 fi
```
caseydavenport: @Copilot - I think these are unnecessary? They should be covered already by the check below. We can remove them.

hack/test/kind/deploy_resources.sh
```diff
@@ -169,15 +169,15 @@ echo
...
 ${HELM} install calico ${CHART} -f ${VALUES_FILE} -n tigera-operator --create-namespace
 
+echo "Install calicoctl as a pod"
+${kubectl} apply -f ${INFRA_DIR}/calicoctl.yaml
+echo
+
 if [[ "$CLUSTER_ROUTING" == "FELIX" ]]; then
   echo "Patching installation resource to Felix cluster routing mode"
   ${kubectl} patch installation default --type='merge' -p '{"spec": {"calicoNetwork": {"clusterRoutingMode":"Felix"}}}'
 fi
```
caseydavenport: @Copilot instead of patching the Installation, let's use a values file for this setup to avoid this two-step issue.

hack/test/kind/deploy_resources.sh
```diff
@@ -167,54 +175,24 @@ echo
...
-  echo "Patching installation resource to Felix cluster routing mode"
-  ${kubectl} patch installation default --type='merge' -p '{"spec": {"calicoNetwork": {"clusterRoutingMode":"Felix"}}}'
-fi
+# Build helm -f args: always include the base VALUES_FILE, then any overlays
+# in EXTRA_VALUES_FILES. Helm deep-merges the files, with later ones winning.
+helm_values_args=(-f "${VALUES_FILE}")
+for extra in ${EXTRA_VALUES_FILES:-}; do
+  helm_values_args+=(-f "${extra}")
+done
+${HELM} install calico ${CHART} "${helm_values_args[@]}" -n tigera-operator --create-namespace
```
mazdakn, line 186: nit: worth printing values files used for installation for sake of troubleshooting.

mazdakn (approved):
LGTM. Just left a nit.

hack/test/kind/deploy_resources.sh
caseydavenport: Good call — added in 28f02c21de.

PR #12477, opened 2026-04-15
E2E infra improvements + Maglev E2E code improvements:

aaaaaaaalex:
* Adds Prometheus metric scraper util
* ~~Adds UpdateFelixConfig util for serial tests~~
* Makes Maglev tests serial
* Adds BPF NAT-map-reading utils
* Adds ExitError formatter so stderr is printed in text (rather than bytes)
* Replaces 'context.TODO's in Maglev tests


e2e/pkg/utils/felix.go
```diff
@@ -0,0 +1,66 @@
...
+
+// UpdateFelixConfig applies a mutation to the default FelixConfiguration.
+// If the mutation results in no change, no update is performed and a no-op
+// cleanup function is returned. Otherwise, the returned cleanup function
+// restores the spec to its pre-mutation state.
+//
+// To avoid cleanups clobbering each-other, they should be applied from latest-to-oldest (e.g via DeferCleanup).
+// NOTE! Since Felixconfig is a global resource, mutations should only be done in serial tests. Cleanup order cannot
+// be guaranteed in parallel tests.
+func UpdateFelixConfig(cli ctrlclient.Client, mutate func(spec *v3.FelixConfigurationSpec)) (cleanup func() error, err error) {
```
caseydavenport: I'm actually intentionally removing this - there's a new generic version of this that we can / should use instead. 

https://github.com/projectcalico/calico/blob/1be532d3761c4570a9b493c502f0c085c7af6ec5/e2e/pkg/utils/config.go#L26-L76

This allows us to use this for other resources as well (e.g., BGPConfiguration, which prompted it)

e2e/pkg/utils/conncheck/conncheck.go
```diff
@@ -641,8 +641,8 @@ func buildFailureMessage(results []connectionResult) string {
 	return msg.String()
 }
 
-// createClientPod creates a long lived pod that sleeps, and can be used to execute connection tests as a client.
-func createClientPod(f *framework.Framework, namespace *v1.Namespace, baseName string, labels map[string]string, customizer func(pod *v1.Pod)) (*v1.Pod, error) {
+// CreateClientPod creates a long lived pod that sleeps, and can be used to execute connection tests as a client.
+func CreateClientPod(f *framework.Framework, namespace *v1.Namespace, baseName string, labels map[string]string, customizer func(pod *v1.Pod)) (*v1.Pod, error) {
```
caseydavenport, line 645: This is a breach of the intended API surface - we should talk about if this is needed and how we can avoid this.

e2e/pkg/utils/felix.go
aaaaaaaalex: Using the generic function now instead. Removed felix.go

caseydavenport (approved):

PR #12479, opened 2026-04-15
Add size limit for inbound Typha 

Behnam-Shobiri:
- Add a size limit for inbound Typha client gob frames to prevent oversized messages from consuming excessive memory.
- Reject oversized client frames before reading the payload, reducing Typha’s exposure to DoS from malformed or malicious connections.

**Release note:**
```release-note
Typha now rejects oversized inbound client gob frames before reading them, preventing a potential denial-of-service caused by excessive memory allocation.
```

### DIFF
--- a/typha/pkg/syncserver/sync_server.go
+++ b/typha/pkg/syncserver/sync_server.go
@@ -52,6 +52,7 @@ var (
 	ErrReadFailed               = errors.New("failed to read from client")
 	ErrUnexpectedClientMsg      = errors.New("unexpected message from client")
 	ErrUnsupportedClientFeature = errors.New("unsupported client feature")
+	errInboundMessageTooLarge   = errors.New("inbound message too large")
 )
 
 var (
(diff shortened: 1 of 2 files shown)


typha/pkg/syncserver/sync_server.go
```diff
@@ -872,6 +882,29 @@ func (h *connection) readFromClient(logCxt *log.Entry) {
...
+	n     int64
+}
+
+func (l *limitedReader) Read(p []byte) (int, error) {
+	n, err := l.r.Read(p)
+	l.n += int64(n)
+	if l.n > l.limit {
+		return n, fmt.Errorf("%w (read %d bytes, limit %d)", errInboundMessageTooLarge, l.n, l.limit)
+	}
+	return n, err
```
fasaxc, line 905: If `len(p)` is large (`saferio` uses 10MB) then you'll already have done the large upstream read before you return an error.  Better to limit the upstream read to the available limit:

```suggestion
	avail := l.limit - l.n
	if avail <= 0 {
		return 0, fmt.Errorf("%w (read %d bytes, limit %d)", errInboundMessageTooLarge, l.n, l.limit)
	}
	if avail < len(p) {
		p := p[:avail]
	}
	n, err := l.r.Read(p)
	l.n += int64(n)
	return n, err
````

Comment (Behnam-Shobiri) [reply typha/pkg/syncserver/sync_server.go:905]: I just changed these lines and add it
```
p := p[:avail] -> p = p[:avail]
avail < len(p) -> int64(len(p)) > avail
```

File: typha/pkg/syncserver/sync_server_message_limit_test.go [hunk 2]
```diff
@@ -0,0 +1,113 @@
...
+		n, err := lr.Read(buf[total:])
+		total += int64(n)
+		if err != nil {
+			if !errors.Is(err, errInboundMessageTooLarge) {
+				t.Fatalf("expected errInboundMessageTooLarge, got %v", err)
+			}
+			break
+		}
+	}
+	if total > limit {
```
fasaxc:
```suggestion
	if total != limit {
```
Should be exact?

typha/pkg/syncserver/sync_server_message_limit_test.go
```diff
@@ -0,0 +1,113 @@
...
+	var total int64
+	for {
+		n, err := lr.Read(buf[total:])
+		total += int64(n)
+		if err != nil {
+			if !errors.Is(err, errInboundMessageTooLarge) {
+				t.Fatalf("expected errInboundMessageTooLarge, got %v", err)
+			}
+			break
+		}
```
fasaxc, line 97:
```suggestion
		} else if n <= 0 { ...
		}
		if total > limit { t.Fatal(...) }
```
Makes the loop robust in case of an infinite loop bug in the reader.

typha/pkg/syncserver/sync_server_message_limit_test.go
```diff
@@ -0,0 +1,113 @@
...
+		t.Fatalf("expected reads to be clamped at limit, but read %d bytes (limit %d)", total, limit)
+	}
+}
+
+func TestLimitedReaderResetAllowsNextMessage(t *testing.T) {
+	data := bytes.Repeat([]byte{0x01}, 100)
+	lr := &limitedReader{r: bytes.NewReader(data), limit: 60}
+
+	buf := make([]byte, 50)
+	if _, err := lr.Read(buf); err != nil {
```
fasaxc: Also check the length that you got

fasaxc (changes requested):
Couple of test nits

fasaxc (approved):
I think this is fine, I can't see felix -> typha messages intentionally exceeding 1MB.  Only way that might happen is if the embedded type information was huge, but we should catch that in CI pretty easily.

Issue #10648, opened 2025-07-09
Calico Node is deleting pod's network interfaces in v3.30 when `FELIX_ROUTESYNCDISABLED` is `true`


cstorres:
Hi people! I'm using calico with mode network policies only, meaning, with the environment variable [`FELIX_ROUTESYNCDISABLED`](https://docs.tigera.io/calico/latest/reference/felix/configuration#process-feature-detectionoverrides) set to `true`. That way we avoid calico-node to delete the pod's network interfaces when it's deployed/restarted in the node (we have our own CNI).

The issue I've spotted when I was testing v3.30.0 was that, even with `FELIX_ROUTESYNCDISABLED` set to `true`, calico-node was deleting the pod's network interfaces in the node.

## Expected Behavior
When `FELIX_ROUTESYNCDISABLED` is set to `true`, calico-node should not delete the pod's network interfaces when it gets deployed/restarted in the node

## Current Behavior
When deploying calico-node v3.30.0 in a node, the network connectivity of the pods was gone so I couldn't reach out the pods running in that node.

### Symptoms

In a node running calico v3.29.4 I had the following pods:

```
kubectl get pods --field-selector spec.nodeName=sjc3-r14-u25 -A -o wide                                                                
NAMESPACE        NAME                       READY   STATUS    RESTARTS   AGE     IP              NODE           NOMINATED NODE   READINESS GATES
dsl              dsl-deployment-7r8rx       1/1     Running   0          4m47s   10.83.202.52    sjc3-r14-u25   <none>           <none>
infra            ipmi-exporter-tnhl7        1/1     Running   0          14d     10.83.2.49      sjc3-r14-u25   <none>           <none>
kube-system      calico-node-v3-29-xadsj    1/1     Running   0          21h     10.83.2.49      sjc3-r14-u25   <none>           <none>
kube-system      csi-dslplugin-crio-wxrfq   3/3     Running   0          14d     10.83.2.49      sjc3-r14-u25   <none>           <none>
kube-system      ipam-lnvlf                 1/1     Running   0          14d     10.83.2.49      sjc3-r14-u25   <none>           <none>
kube-system      kube-proxy-sjc3-r14-u25    1/1     Running   0          14d     10.83.2.49      sjc3-r14-u25   <none>           <none>
kube-system      node-local-dns-wbhsr       1/1     Running   0          14d     10.83.2.49      sjc3-r14-u25   <none>           <none>
[...]

PR #12489, opened 2026-04-15
fix: Introduce dummy interface for LinkAddressManagement to fix RouteSyncDisabled functionality

juanfresia:
# Changes

This PR fixes a regression introduced by #9875 that broke support for "network policy only" mode: `LinkAddrsManager` component alters the kernel route table even if the parameter `RouteSyncDisabled` is set to `true` (see feature #5454). The result is the bug reported in #10648.

I've created an interface for the LinkAddrsManager in order to be able to choose the implementation to be used. In this case, for the network-policy-only mode we are interested in disabling the feature altogether, so `DummyLinkAddrsManager` implements the interface but does basically does nothing.

When the data-plane managers are instantiated, the same `RouteSyncDisabled` flag is used to pick between the real and the dummy implementation of the `LinkAddrsManager`.

I've tested this in our staging environment, as well as a backport of this change into v1.31; and the behavior is as we'd expect: the network policies keep being implemented by calico-node but no routes are modified.

**Release note:**
```release-note
Fixed a regression introduced in v3.30 where `RouteSyncDisabled` flag was not being honored by `LinkAddressManager`.
```

### DIFF
--- a/felix/dataplane/linux/int_dataplane.go
+++ b/felix/dataplane/linux/int_dataplane.go
@@ -363,7 +363,7 @@ type InternalDataplane struct {
 	vxlanManagerV6      *vxlanManager
 	vxlanFDBs           []*vxlanfdb.VXLANFDB
 
-	linkAddrsManagers []*linkaddrs.LinkAddrsManager
+	linkAddrsManagers []linkaddrs.Interface
 
 	wireguardManager   *wireguardManager
 	wireguardManagerV6 *wireguardManager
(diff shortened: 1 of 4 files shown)


cstorres:
Thanks @juanfresia for the fix! 

@song-jiang your review could be really valuable here since you wrote the original PR, just to be sure we're not breaking anything :)

juanfresia:
Thanks @mazdakn, I've fixed the staticchecks errors which I believe were the only thing failing.

mazdakn:
@juanfresia thanks for providing the fix. I'll review the PR tomorrow.

felix/dataplane/linux/int_dataplane.go
```diff
@@ -1141,7 +1141,16 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		dp.arpTables = append(dp.arpTables, arpFilterTable)
 	}
 
-	linkAddrsManagerV4 := linkaddrs.New(4, config.RulesConfig.WorkloadIfacePrefixes, featureDetector, config.NetlinkTimeout)
+	var linkAddrsManagerV4 linkaddrs.Interface
+	if !config.RouteSyncDisabled {
+		log.Debug("Route management is enabled. Using default LinkAddrsManager")
+
```
mazdakn, line 1147: nit: please remove this blank line

mazdakn (commented):
@juanfresia The PR itself looks good to me. I just think we should add a test to make sure not regressing in future. Potentially a felix FV test covering this should be enough. Can you please look into adding one?

mazdakn (approved):
LGTM. Adding a FV test to cover this functionality: https://github.com/projectcalico/calico/pull/12601

mazdakn:
@juanfresia thanks for your contribution. Can you please pick this to relevant branches, like v3.32, and v3.31?

juanfresia:
Thank you @mazdakn , I'll cherry pick this to v3.32 and v3.31 as soon as I have some time.

PR #12494, opened 2026-04-15
Remove unused parameter from (*allocationBlock).deleteAttributes

djmitche:
In working on a larger patch, I noticed that this method does not use its `ordinals` argument, and in fact it's not clear what that would mean in the context of deleting attributes. So, I removed it.

There's no user-visible change here.

**Release note:**
```release-note
TBD
```


caseydavenport:
@djmitche thanks!

I'm curious what the larger patch you're working on is though! Is there an issue or something I can catch up on?

caseydavenport (approved):

djmitche:
Yeah, I've discussed in #networking but I will get an issue up soon - my bad for that!

djmitche:
https://github.com/projectcalico/calico/issues/12496

PR #12504, opened 2026-04-16
Force kind image rebuild when local image is missing

caseydavenport:
If a kind image is pruned or removed from the local Docker daemon while its Make stamp file remains, `kind-build-images` skips the rebuild because the stamp's mtime still looks fresh. This is a common footgun when iterating - `docker system prune`, then `make kind-reload` silently does nothing.

Each `.image.created-$(ARCH)` stamp now records the image ref it built. `hack/image-exists` runs at Make parse time via `$(shell)` and prints `MISSING-IMAGE` (a phony, always-out-of-date target) when the recorded image is gone, dragging the rule back in. The stamp is removed before the sub-make runs so the sub-make's own up-to-date check doesn't short-circuit the rebuild.

Applied to all four kind image stamps: `node`, `whisker`, `cmd/calico`, and `key-cert-provisioner`.

```release-note
None
```


lib.Makefile
```diff
@@ -1618,58 +1618,66 @@ KIND_IMAGE_MARKERS = \
...
+	echo "tigera/operator:$(KIND_TEST_BUILD_TAG)" > $@
 
 ## Build all component images needed for kind cluster testing, then tag them.
+## First validates that stamp files aren't stale (image pruned but stamp remains),
+## then runs the actual build in a sub-make so prerequisites are re-evaluated.
 .PHONY: kind-build-images
-kind-build-images: $(KIND_IMAGE_MARKERS) $(REPO_ROOT)/.stamp.operator
+kind-build-images:
+	@$(REPO_ROOT)/hack/test/kind/validate_stamps.sh \
+		$(KIND_IMAGE_MARKERS) $(REPO_ROOT)/.stamp.operator
```
caseydavenport:
```suggestion
	@$(REPO_ROOT)/hack/test/kind/validate_stamps.sh $(KIND_IMAGE_MARKERS) $(REPO_ROOT)/.stamp.operator
```

mazdakn (approved):

PR #12452, opened 2026-04-13
Refactor e2e test execution scripts

caseydavenport:
This PR refactors `body_standard.sh` into smaller composable phase scripts, and reworks how e2e binaries are sourced:

- **Per-PR CI** (`RUN_LOCAL_TESTS`): builds the e2e binary from source
- **Scheduled CI** (`TEST_TYPE=k8s-e2e`): downloads the e2e binary from the latest hashrelease
- **Other test types** (benchmarks, cert, etc.): falls back to `bz tests`

Also adds an `e2e-run` Makefile target and switches GCP e2e tests to use a config file instead of inline flags.

Supersedes #12507.

```release-note
None
```

PR #12507, opened 2026-04-16
Use hashrelease e2e binary in CI pipeline

caseydavenport:
Stacked on #12506 - merge that first.

Updates the CI pipeline script to download the pre-built e2e binary from the hashrelease instead of calling `bz tests`. The `RUN_LOCAL_TESTS` path (used by the GCP e2e block) still builds from source. Both paths now share a unified execution step using the ginkgo CLI in a `calico/go-build` Docker container.

This replaces the opaque `bz tests` call with direct binary execution, which makes test selection visible in the pipeline config (via `K8S_E2E_FLAGS`) and ensures the test binary matches the installed Calico version.

Part 2 of splitting #12351.

```release-note
None
```


caseydavenport:
Incorporated into https://github.com/projectcalico/calico/pull/12452

PR #12508, opened 2026-04-16
Add tests for allocationBlock, revealing two bugs

djmitche:
This adds a bunch of tests, revealing two bugs:
 - `release` under-counts occurrences of handles
 - `releaseByHandle` leaves the block in an inconsistent state if the sequence number check fails.

This also includes two trivial refactors to `ipam_block.go`, removing an unused function and inlining another to its only callsite.

I will make follow-up PRs to fix those two bugs, so that the fixes can be discussed separately from the tests.

**Release note:**
```release-note
TBD
```

### DIFF
--- a/libcalico-go/lib/ipam/ipam_block.go
+++ b/libcalico-go/lib/ipam/ipam_block.go
@@ -230,7 +230,16 @@ func (b allocationBlock) NumFreeAddresses(reservations addrFilter) int {
 // empty returns true if the block has released all of its assignable addresses,
 // and returns false if any assignable addresses are in use.
 func (b allocationBlock) empty() bool {
-	return b.containsOnlyReservedIPs()
+	for _, attrIdx := range b.Allocations {
+		if attrIdx == nil {
+			continue
+		}
+		attrs := b.Attributes[*attrIdx]
+		if attrs.HandleID == nil || strings.ToLower(*attrs.HandleID) != WindowsReservedHandle {
+			return false
+		}
+	}
+	return true
 }
 
 // inUseIPs returns a list of IPs currently allocated in this block in string format.
@@ -246,21 +255,9 @@ func (b allocationBlock) inUseIPs() []string {
 	return ips
 }
 
-// containsOnlyReservedIPs returns true if the block is empty excepted for
-// expected "reserved" IP addresses.
-func (b *allocationBlock) containsOnlyReservedIPs() bool {
-	for _, attrIdx := range b.Allocations {
-		if attrIdx == nil {
-			continue
-		}
-		attrs := b.Attributes[*attrIdx]
-		if attrs.HandleID == nil || strings.ToLower(*attrs.HandleID) != WindowsReservedHandle {
-			return false
-		}
-	}
-	return true
-}
-
+// release tries to release addresses matching the release options, on success, returns
+// slice of IPs that were released, map from handle ID to count of IPs released
+// for that handle.
 func (b *allocationBlock) release(addresses []ReleaseOptions) ([]cnet.IP, map[string]int, error) {
 	// Store return values.
 	unallocated := []cnet.IP{}
@@ -497,22 +494,6 @@ func (b allocationBlock) ipsByHandle(handleID string) []cnet.IP {
 		}
 	}
 	return ips
-}
-
-func (b allocationBlock) attributesForIP(ip cnet.IP) (map[string]string, error) {
-	// Convert to an ordinal.
-	ordinal, err := b.IPToOrdinal(ip)
-	if err != nil {
-		return nil, err
-	}
-
-	// Check if allocated.
-	attrIndex := b.Allocations[ordinal]
-	if attrIndex == nil {
-		log.Debugf("IP %s is not currently assigned in block", ip)
-		return nil, cerrors.ErrorResourceDoesNotExist{Identifier: ip.String(), Err: errors.New("IP is unassigned")}
-	}
-	return b.Attributes[*attrIndex].ActiveOwnerAttrs, nil
 }
 
 // allocationAttributesForIP returns the full AllocationAttribute for the given IP,
--- a/libcalico-go/lib/ipam/ipam_block_test.go
+++ b/libcalico-go/lib/ipam/ipam_block_test.go
@@ -1,0 +1,368 @@
+package ipam
+
+import (
+	"slices"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/projectcalico/calico/libcalico-go/lib/backend/model"
+	cnet "github.com/projectcalico/calico/libcalico-go/lib/net"
+)
+
+type testBlock struct {
+	allocationBlock
+}
+
+func makeTestBlock() *testBlock {
+	_, blockNet, err := cnet.ParseCIDR("100.64.0.0/24")
+	Expect(err).NotTo(HaveOccurred())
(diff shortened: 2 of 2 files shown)


djmitche:
Fixes are in 
* https://github.com/projectcalico/calico/compare/master...djmitche:calico:push-xllllxwswqqn?expand=1
* https://github.com/projectcalico/calico/compare/master...djmitche:calico:push-kvpssttzlsrz?expand=1

both include this commit as a parent, so I'll file them as PRs once this lands.

libcalico-go/lib/ipam/ipam_block_test.go
```diff
@@ -0,0 +1,374 @@
...
+
+		unallocated, countByHandle, err := block.release([]ReleaseOptions{{
+			Address: "100.64.0.13",
+		}})
+		Expect(err).To(Succeed())
+		Expect(unallocated).To(HaveLen(0))
+		Expect(countByHandle).To(Equal(map[string]int{"lucky": 1}))
+		Expect(block.allocatedOrdinals()).To(SatisfyAll(
+			HaveLen(2),
+			ContainElements([]int{26, 39}),
```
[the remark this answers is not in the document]
djmitche: This actually works anyway, but it's a good fix.

libcalico-go/lib/ipam/ipam_block_test.go
```diff
@@ -0,0 +1,374 @@
...
+		ip := cnet.ParseIP("100.64.0.32")
+		block.allocateAttrib([]int{32}, model.AllocationAttribute{
+			HandleID:            new("hand"),
+			ActiveOwnerAttrs:    map[string]string{"is": "active"},
+			AlternateOwnerAttrs: map[string]string{"is": "alternate"},
+		})
+		Expect(block.allocationAttributesForIP(*ip)).To(SatisfyAll(
+			HaveField("HandleID", new("hand")),
+			HaveField("ActiveOwnerAttrs", ConsistOf("active")),
+			HaveField("AlternateOwnerAttrs", ConsistOf("alternate")),
```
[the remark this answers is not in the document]
djmitche: Eh, I'm happy enough with this assertion

libcalico-go/lib/ipam/ipam_block_test.go
```diff
@@ -0,0 +1,374 @@
...
+	It("skips removal if the sequence numbers do not match", func() {
+		Skip("Currently failing: leaks the IP")
+
+		block := makeTestBlock()
+		block.allocate([]int{15}, "fifteen")
+		block.SequenceNumberForAllocation["15"] = 10
+
+		released := block.releaseByHandle(ReleaseOptions{
+			Handle:         "fifteen",
+			SequenceNumber: new(uint64(999)),
```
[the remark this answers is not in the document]
djmitche: I think copilot needs a newer Go version. Gemini's matrix multiplications got this wrong too. But it [works fine in go1.26](https://go.dev/doc/go1.26#language).

libcalico-go/lib/ipam/ipam_block_test.go
```diff
@@ -0,0 +1,374 @@
...
+
+	It("returns an error if the sequence numbers do not match", func() {
+		block := makeTestBlock()
+		block.allocate([]int{13}, "unlucky")
+		block.SequenceNumberForAllocation["13"] = 10
+
+		_, _, err := block.release([]ReleaseOptions{{
+			Address:        "100.64.0.13",
+			Handle:         "unlucky",
+			SequenceNumber: new(uint64(999)),
```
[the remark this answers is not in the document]
djmitche: I think copilot needs a newer Go version. Gemini's matrix multiplications got this wrong too. But it [works fine in go1.26](https://go.dev/doc/go1.26#language).

libcalico-go/lib/ipam/ipam_block_test.go
```diff
@@ -0,0 +1,368 @@
...
+		block.allocate([]int{10, 20, 210, 220}, "tens")
+		Expect(block.NumFreeAddresses(cidrSliceFilter([]cnet.IPNet{*wholeBlock}))).To(Equal(0))
+	})
+
+	It("correctly returns IPs by handle, whether zero, one, or many", func() {
+		block := makeTestBlock()
+		block.allocate([]int{33}, "singleton")
+		block.allocate([]int{10, 20}, "tens")
+		block.allocate([]int{33}, "thirties")
+		block.allocate([]int{34}, "thirties")
```
djmitche: whoops, using the same ordinal here twice was a mistake!

libcalico-go/lib/ipam/ipam_block.go
```diff
@@ -246,21 +255,6 @@ func (b allocationBlock) inUseIPs() []string {
...
-		}
-		attrs := b.Attributes[*attrIdx]
-		if attrs.HandleID == nil || strings.ToLower(*attrs.HandleID) != WindowsReservedHandle {
-			return false
-		}
-	}
-	return true
-}
-
 func (b *allocationBlock) release(addresses []ReleaseOptions) ([]cnet.IP, map[string]int, error) {
```
fasaxc, line 261:
```suggestion
// release tries to release addresses matching the release options, on success, returns 
// slice of IPs that were released, map from handle ID to count of IPs released
// for that handle.
func (b *allocationBlock) release(addresses []ReleaseOptions) ([]cnet.IP, map[string]int, error) {
```

libcalico-go/lib/ipam/ipam_block_test.go
```diff
@@ -0,0 +1,368 @@
...
+		Expect(released).To(Equal(1))
+		Expect(block.allocatedOrdinals()).To(Equal([]int{20}), "twenty not affected")
+		Expect(block.unallocatedTail(1)).To(ContainElement(15),
+			"Unallocated should have released block at the tail")
+
+		block.validate()
+	})
+
+	It("skips removal if the sequence numbers do not match", func() {
+		Skip("Currently failing: leaks the IP")
```
fasaxc, line 319: Can you clarify?  Do you mean that the IP _does_ get deallocated in this case?

IIUC, it should skip this IP.  The sequence number is used by the garbage collection routine to make sure that it doesn't race with something else that cleans up and re-uses the IP.  So it should skip non-matching IPs and then check them again on thee next GC loop.

djmitche: It gets leaked, in the sense that its slot in `b.Allocated` is set to `nil` but the ordinal is not added to `b.Unallocated`. This occurs because the attributes with the given handle are cleared despite this sequence-number issue, and elements of `Allocated` pointed to cleared attributes are set to `nil`.

The `Skip(..)` here is skipping the test since the expected behavior is not occurring.

The proposed fix in [this branch](https://github.com/projectcalico/calico/compare/master...djmitche:calico:push-xllllxwswqqn?expand=1) returns an error in this case without releasing anything, mirroring `release`'s behavior in this situation. If you'd prefer to release the non-conflicted IPs with the given handle, I can do so when that PR goes up for review.

fasaxc: Aha, so not just leaked but the block is corrupted

djmitche: Correct -- apologies for mis-using the term!

fasaxc (approved):
LGTM

fasaxc:
CI failure is just a code formatting check; `make fix-changed` should do the trick.

djmitche:
Hm, that failure doesn't look like it would be related?

caseydavenport:
Agreed, giving it a re-run.

PR #12518, opened 2026-04-17
fix(#12253): BPF, check failsafe for DHCP

haoadoreorange:
When `netplan apply` triggers DHCP renewal, packets use source 0.0.0.0 / dest 255.255.255.255 which are not in the route table as local host IPs. The existing failsafe checks are gated on local host IP recognition, so they are skipped and DHCP is denied by policy.

Add failsafe checks for:
- Outbound: when source IP is void (0.0.0.0)
- Inbound: when dest has no route

### DIFF
--- a/felix/bpf-gpl/tc.c
+++ b/felix/bpf-gpl/tc.c
@@ -700,6 +700,15 @@ syn_force_policy:
 	if (!dest_rt) {
 		CALI_DEBUG("No route for post DNAT dest " IP_FMT "", debug_ip(ctx->state->post_nat_ip_dst));
 		if (CALI_F_FROM_HEP) {
+			/* Check failsafe for inbound DHCP broadcast (dest 255.255.255.255).
+			 * https://github.com/projectcalico/calico/issues/12253
+			 */
+			if (ctx->state->post_nat_ip_dst == 0xffffffff &&
+			    is_failsafe_in(ctx->state->ip_proto, ctx->state->post_nat_dport, ctx->state->ip_src)) {
+				CALI_DEBUG("Inbound failsafe port: %d. Skip policy (DHCP broadcast).", ctx->state->post_nat_dport);
+				counter_inc(ctx, CALI_REASON_ACCEPTED_BY_FAILSAFE);
+				goto skip_policy;
+			}
 			/* Disable FIB, let the packet go through the host after it is
 			 * policed. It is ingress into the system and we do not know what
 			 * exactly is the packet's destination. It may be a local VM or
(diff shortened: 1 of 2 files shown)


felix/bpf-gpl/tc.c
```diff
@@ -700,6 +700,16 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 	if (!dest_rt) {
 		CALI_DEBUG("No route for post DNAT dest " IP_FMT "", debug_ip(ctx->state->post_nat_ip_dst));
 		if (CALI_F_FROM_HEP) {
+			/* Check failsafe before policy for packets with no route.
+			 * Handles DHCP responses (dest 255.255.255.255 or newly
+			 * assigned IP not yet in the route table).
+			 * https://github.com/projectcalico/calico/issues/12253
+			 */
+			if (is_failsafe_in(ctx->state->ip_proto, ctx->state->post_nat_dport, ctx->state->ip_src)) {
```
sridhartigera, line 708: I feel this check is too broad. We should probably look at narrowing down the check to DHCP packets with src=`0.0.0.0` outbound and dst=`255.255.255.255` in bound.

sridhartigera (changes requested):

felix/bpf-gpl/tc.c
haoadoreorange: Right, this will only allow DHCP broadcast. From what I read, DHCP unicast response has the assigned IP as dstIP, I don't know if it's used in practice ?
I pushed the change to only allow broadcast.

felix/bpf-gpl/tc.c
```diff
@@ -700,6 +700,15 @@ static CALI_BPF_INLINE void calico_tc_process_ct_lookup(struct cali_tc_ctx *ctx)
 	if (!dest_rt) {
 		CALI_DEBUG("No route for post DNAT dest " IP_FMT "", debug_ip(ctx->state->post_nat_ip_dst));
 		if (CALI_F_FROM_HEP) {
+			/* Check failsafe for inbound DHCP broadcast (dest 255.255.255.255).
+			 * https://github.com/projectcalico/calico/issues/12253
+			 */
+			if (ctx->state->post_nat_ip_dst == 0xffffffff &&
+			    is_failsafe_in(ctx->state->ip_proto, ctx->state->post_nat_dport, ctx->state->ip_src)) {
```
sridhartigera, line 707: This is allowing all broadcast packets with the failsafe (any failsafe). My suggestion is to narrow down the  check to DHCP. Ports 67,68 and broadcast dest for DHCP response and similarily src=0.0.0.0 for DHCP request. I would check against port 67,68 depending on the direction and proto UDP. Ideally would want to place the check at https://github.com/projectcalico/calico/blob/master/felix/bpf-gpl/tc.c#L788 and leave this code block as it is.

sridhartigera (changes requested):

felix/bpf-gpl/tc.c
haoadoreorange: a question on port 67,68: aren't they already in failsafe map when check with ctx->state->post_nat_dport ? I don't understand why an explicit check is needed

sridhartigera: @haoadoreorange Yes they are. Failsafe also has ports like 22 and failsafe in and out ports are configurable as well.

sridhartigera: @haoadoreorange Do you have a HEP and a policy on it?

PR #12523, opened 2026-04-17
Handle BGPConfiguration fields directly, not via confd "cache"

nelljerram:
I'd like eventually to remove all translations to the "v2" / "model" / legacy etcd paths, as well as all use of the confd "cache", and emit BIRD config based directly on our v3 resources.

This PR is one step in that direction, for the BGPConfiguration fields that we honour in the "default" (aka global) BGPConfiguration but not in a per-node BGPConfiguration.

(If this looks good for the BindMode field, I'll expand to other such fields.)

### DIFF
--- a/confd/pkg/backends/calico/client.go
+++ b/confd/pkg/backends/calico/client.go
@@ -1231,7 +1231,6 @@ func (c *client) updateBGPConfigCache(resName string, v3res *apiv3.BGPConfigurat
 	if resName == globalConfigName {
 		c.getPrefixAdvertisementsKVPair(v3res, model.GlobalBGPConfigKey{})
 		c.getListenPortKVPair(v3res, model.GlobalBGPConfigKey{}, updatePeersV1, updateReasons)
-		c.getBindModeKVPair(v3res, model.GlobalBGPConfigKey{}, updatePeersV1, updateReasons)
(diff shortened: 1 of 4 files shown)


confd/pkg/backends/calico/client.go
```diff
@@ -1251,13 +1250,18 @@ func (c *client) updateBGPConfigCache(resName string, v3res *apiv3.BGPConfigurat
 			c.serviceLoadBalancerAggregation = apiv3.ServiceLoadBalancerAggregationEnabled
 		}
 
-		// Check if normal route priority is changing.
+		// Check for changes in BGPConfiguration fields that impact GetBirdBGPConfig.  Note,
+		// the following changes do not affect the _set_ of BGP peers, so do not require
+		// setting updatePeersV1 and updateReasons.
```
nelljerram, line 1255: Perhaps we can just call `c.keyUpdated("/calico/bgpconfig")` unconditionally here, without checking individual fields, based on the fact that we're seeing _some_ kind of BGPConfiguration change.

song-jiang: Yeah, but it is not too bad to check individual fields.

song-jiang (approved):
LGTM

PR #12524, opened 2026-04-17
Handle more BGPConfiguration fields directly, not via confd "cache"

nelljerram:
- ServiceClusterIPs
- ServiceExternalIPs
- ServiceLoadBalancerIPs
- NodeMeshRestartTime
- NodeMeshPassword
- IgnoredInterfaces
- ProgramClusterRoutes

(Following the direction laid out in #12523)


confd/pkg/backends/calico/client.go
```diff
@@ -1191,29 +1191,16 @@ func (c *client) onUpdates(updates []api.Update, needUpdatePeersV1 bool) {
...
 
-		// Same for cluster CIDRs.
-		var clusterIPs []string
-		if len(c.cache["/calico/bgp/v1/global/svc_cluster_ips"]) > 0 {
-			clusterIPs = strings.Split(c.cache["/calico/bgp/v1/global/svc_cluster_ips"], ",")
+		// Same for cluster CIDRs - except those can be overridden by an env var, so only
+		// process cluster IPs in BGPConfiguration if that env var is not set.
+		if len(os.Getenv(envAdvertiseClusterIPs)) == 0 {
+			c.onClusterIPsUpdate(getServiceClusterIPs(c.globalBGPConfig))
 		}
```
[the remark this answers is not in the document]
nelljerram: Good catch, but:

1. This was equally a concern in the code before this PR.  In other words, it was already the case that a use of CALICO_ADVERTISE_CLUSTER_IPS would ignore whether the local node has the `node.kubernetes.io/exclude-from-external-load-balancers` label.  This PR doesn't change that.

2. Our docs have been clear for a long time that CALICO_ADVERTISE_CLUSTER_IPS is deprecated, and explain how to migrate to the non-deprecated approach.  So in case this interaction is actually important to someone, they can fix it by using BGPConfiguration instead of  the environment variable.

On balance, therefore, I don't think this merits additional work.

confd/pkg/backends/calico/ippools_test.go
```diff
@@ -459,128 +458,3 @@ func filterExpectedStatements(statements []string, filterAction string) (filtere
...
 	return
 }
-
-// programClusterRoutesCacheKey is the confd cache key that mirrors
-// BGPConfiguration.Spec.ProgramClusterRoutes into BIRD config.
-const programClusterRoutesCacheKey = "/calico/bgp/v1/global/program_cluster_routes"
-
-func strPtr(s string) *string { return &s }
-
-func TestGetProgramClusterRoutesKVPair(t *testing.T) {
```
mazdakn, line 469: Can we replace these tests with new ones? The intention was basically to verify the expected behavior during switching back and forth the config value. It would be great if we can get the same coverage in UTs, otherwise, we need to think about adding new tests in some other way.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -198,23 +199,17 @@ func (c *client) populateNodeConfig(pc *processorContext, config *types.BirdBGPC
...
-		// Default pattern with explanatory comment
-		config.DirectInterfaces = `-"cali*", -"kube-ipvs*", "*"`
+	var ignoredInterfaces []string
+	if pc.globalBGPConfig != nil {
+		ignoredInterfaces = pc.globalBGPConfig.Spec.IgnoredInterfaces
 	}
+	var patterns []string
+	for _, iface := range ignoredInterfaces {
+		patterns = append(patterns, fmt.Sprintf(`-"%s"`, iface))
+	}
```
mazdakn, line 226: nit: I think moving these processing logics to `pc` would make the code cleaner. For example:

```
func (pc *ProcessorContext) GetValueX() ...
```

Then we can move all values checks and defaulting to that method.

confd/pkg/backends/calico/ippools_test.go
mazdakn: hmm but we might not actually need these tests as suggested. The original issue was that the default BGPCofing changes were not triggering generation of BIRD configs due to caching, and now we read it directly.

mazdakn (approved):
LGTM. I just have a concern about removing tests, but not major issue.

confd/pkg/backends/calico/ippools_test.go
nelljerram: I've added a test in d631e06be9 - WDYT?

confd/pkg/backends/calico/bgp_processor.go
nelljerram: I've moved a couple of functions in 1e11df14b6, but I'm not confident I understood what you really meant here - PTAL.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -992,30 +1000,22 @@ func (c *client) processIPPool(
 	}
 
 	// IPIP encapsulation or No-Encap.
-	if ippool.IPIPMode == encap.Always || ippool.IPIPMode == encap.CrossSubnet ||
-		ippool.IPIPMode == encap.Never || ippool.VXLANMode == encap.Never {
```
nelljerram, line 996: I removed this condition because I believe it was 100% true, because it is known at this point in the code that `ippool.VXLANMode == encap.Never`.

song-jiang (approved):
LGTM

confd/pkg/backends/calico/ippools_test.go
mazdakn: Look great, thanks for adding it.

confd/pkg/backends/calico/bgp_processor.go
mazdakn: This is exactly what I meant. I just think this way we can move all processing of BGPConfig fields to ProcessoContext which makes the client code simpler. Anyway this is just a small improvement for sure.

confd/pkg/backends/calico/bgp_processor.go
mazdakn: Make sense. We don't expect any other values here.

PR #12533, opened 2026-04-17
[release-v3.31] fix(app-policy): normalize HTTP request-target before ALP path matching

electricjesus:
## Description

Bug fix in `app-policy/checker`.

`matchHTTPPaths` compared the HTTP request-target from the ext_authz Check call byte-for-byte against policy path rules, with no intervening normalisation. Upstream HTTP servers routinely resolve `.` / `..`, decode percent-escapes and collapse repeated slashes before dispatch, so the path actually served could differ from the one authorised.

This PR reshapes `matchHTTPPaths` to apply RFC 3986 / RFC 7230 style normalisation to both request-target and each rule's path before comparison, and to anchor prefix matches to path-segment boundaries. The normaliser also rejects shapes whose resolved form depends on upstream-specific decoding behaviour (paths whose decoded form still contains percent-encoded path separators, or paths containing a null byte).

Behaviour summary:
- Strip query and fragment.
- Percent-decode the path once.
- Reject decoded paths that still contain `%2e` / `%2f` / `%5c` (case-insensitive) or a null byte.
- Strip `;...` matrix parameters per segment.
- Collapse repeated slashes, fold `\` to `/`, and resolve `.` / `..` segments via `path.Clean`.
- Anchor prefix matches to path-segment boundaries (`prefix: /pub` no longer matches `/public-leak-endpoint`).
- Rule paths go through the same normalisation as the request.
- Rule paths that cannot be interpreted as absolute paths are skipped with a warning rather than matching every request.

Components affected: `app-policy` (Dikastes). Dataplane-mode matrix does not apply — Dikastes is a pod-namespace sidecar evaluating L7 policy in user-space.

Testing:
- `go test ./app-policy/checker/...` — green.
- New table-driven unit coverage for both `matchHTTPPaths` and the normaliser.
- Go fuzz canary asserting structural invariants on the normaliser.

## Related issues/PRs

Further detail will be published in a separate security advisory.

## Todos

- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
app-policy (Dikastes): normalize HTTP request-target before evaluating Application Layer Policy path rules, and reject shapes whose resolved form depends on upstream-specific decoding. Request paths are now RFC 3986 / RFC 7230 normalized (decode percent-escapes once, resolve dot-segments and repeated slashes, fold backslashes, strip matrix parameters per segment) and prefix matches are anchored to path-segment boundaries. Paths whose decoded form still contains percent-encoded path separators (%2e / %2f / %5c), or contains a null byte, are rejected.
```

## Reminder for the reviewer


Behnam-Shobiri (approved):

stevegaossou:
Let's hold merge for this PR. There's a bunch of consistent CI job failures for Felix BPF tests and I am thinking it would be a good opportunity to use this PR to work through those. 

The next v3.31 patch release won't be for a while, so there's no rush right now.

PR #12535, opened 2026-04-17
Sanitize calicoctl log output

Behnam-Shobiri:
**Release note:**
```release-note
`calicoctl` no longer logs raw client config on startup, which previously included `K8sAPIToken`, inline kubeconfig, `EtcdPassword`, and inline etcd key/cert material. The replacement log entry reports only non-sensitive fields and boolean "set" indicators for each credential.
```


anthonytwh (approved):

Behnam-Shobiri:
This is the fix for CVE-2026-6720

PR #12538, opened 2026-04-17
felix/fv: test HEP AoF policy on tunnel-decap forwarded traffic

tomastigera:
Extend `felix/fv/apply_on_forward_test.go` with a `_BPF-SAFE_` DatastoreDescribe that loops over `{vxlan, ipip, wireguard}` tunnel modes. For each mode two `It` blocks:

- **baseline**: no HEP → cross-node pod-to-pod flows across the tunnel.
- **main**: wildcard HEP on both nodes + GlobalNetworkPolicy with `Ingress: Deny, Egress: Allow, ApplyOnForward: true` → cross-node pod-to-pod is blocked by the AoF ingress-deny on the receiving node's tunnel interface (`vxlan.calico` / `tunl0` / `wireguard.cali`) after the kernel decapsulates the outer frame.

Guards against dataplane regressions that break HEP+AoF enforcement on tunnel-decapped traffic. The existing tests in this file cover the complementary "HEP without AoF does not block forwarded traffic" case; together they pin the documented semantics at https://docs.tigera.io/calico/latest/reference/host-endpoints/forwarded on both iptables and BPF dataplanes.

**Test plan:**
- [x] `make -C felix fv GINKGO_FOCUS="apply on forward tests; with tunnel encap"` — 6/6 pass in 61 s (iptables)
- [x] `make -C felix fv-bpf GINKGO_FOCUS="apply on forward tests; with tunnel encap"` — 6/6 pass in 101 s (BPF)

**Release note:**
```release-note
None
```

### DIFF
--- a/felix/fv/apply_on_forward_test.go
+++ b/felix/fv/apply_on_forward_test.go
@@ -203,3 +203,162 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ apply on forward tests; wit
 		})
 	})
 })
+
+// These tests exercise HostEndpoint policy with applyOnForward=true on
+// forwarded pod-to-pod traffic that traverses a tunnel. The tunnel is
+// decapsulated on the receiving node and the inner packet arrives on the
+// tunnel interface (vxlan.calico, tunl0, or wireguard.cali) as forwarded
+// traffic to a local workload.
+//
+// Per Calico docs
+// (https://docs.tigera.io/calico/latest/reference/host-endpoints/forwarded),
+// a wildcard HEP with an applyOnForward=true deny-ingress policy must block
+// this traffic; without applyOnForward=true, forwarded traffic through a HEP
+// is allowed by default and is covered by the tests above.
+//
+// These tests guard against encap-specific regressions where HEP policy
+// enforcement breaks on tunnel-decapped traffic. Runs in both iptables and
+// BPF modes (_BPF-SAFE_).
+var _ = infrastructure.DatastoreDescribe(
+	"_BPF-SAFE_ apply on forward tests; with tunnel encap",
+	[]apiconfig.DatastoreType{apiconfig.Kubernetes},
+	func(getInfra infrastructure.InfraFactory) {
+		for _, tunnel := range []string{"vxlan", "ipip", "wireguard"} {
+			tunnel := tunnel
+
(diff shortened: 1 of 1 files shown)


felix/fv/apply_on_forward_test.go
```diff
@@ -203,3 +203,142 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ apply on forward tests; wit
...
+				AfterEach(func() {
+					tc.Stop()
+					infra.Stop()
+				})
+
+				It("should allow pod-to-pod traffic when no AoF policy applies (baseline)", func() {
+					cc.ExpectSome(w[0], w[1])
+					cc.ExpectSome(w[1], w[0])
+					cc.CheckConnectivity()
+				})
```
tomastigera, line 297: no value, it could be run just before you set the policy to first show that the traffic is allowed and then it fails after the policy changes. It needs checker reset befor the next test.

tomastigera:
Force-pushed `fcc037a3c6` with the review fixes:

- Merged the standalone baseline `It` into the main test (addresses the CI flake where the WireGuard baseline raced the handshake — `test-033-FAILED.log` showed it timed out after 5 tries in 10s).
- Added `cc.StaggerStartBy = 100ms` for WireGuard to avoid simultaneous bidirectional handshake races (@Copilot).
- Create the `GlobalNetworkPolicy` *before* the `HostEndpoint`s so there's no transient default-deny window on the HEP (@Copilot).
- Skip the WireGuard case when `FELIX_FV_WIREGUARD_AVAILABLE != true`, matching `wireguard_test.go` (@Copilot).
- Call `ensureRoutesProgrammed(tc.Felixes)` for VXLAN/IPIP. WireGuard is excluded because that helper matches VXLAN iface name or the Felix/Bird default-route proto, neither of which is on the WG route path; the main test's 30s retry window handles WG convergence instead (@Copilot).

Re-tested:
- `make -C felix fv-bpf GINKGO_FOCUS="apply on forward tests; with tunnel encap" FELIX_FV_WIREGUARD_AVAILABLE=true` — 3/3 pass in 71 s.
- `make -C felix fv GINKGO_FOCUS="apply on forward tests; with tunnel encap" FELIX_FV_WIREGUARD_AVAILABLE=true` — 3/3 pass in 48 s.

sridhartigera (approved):

PR #12553, opened 2026-04-20
confd: fix bgp_local_pref overwrite on iBGP route re-export

song-jiang:
## Summary

When a node re-exports an iBGP-learned route (as a route reflector, or via
eBGP to a TOR), the export filter unconditionally overwrites `bgp_local_pref`
with a value derived from the default `krt_metric` (1024). This destroys the
original priority information from the originating node, causing:

- **BGPFilter Priority matching to fail** — community tags are never added to
  elevated-priority routes on eBGP export
- **Route reflector clients to lose elevated priority** — `bgp_local_pref` is
  reset to the default before re-advertisement

The root cause is that `krt_metric` (a kernel-only attribute) is undefined for
BGP-learned routes, so it defaults to 1024 and overwrites `bgp_local_pref`.

The fix checks `bgp_local_pref` first: if already defined (iBGP-learned
route), it is the authoritative source of priority — keep it and derive
`krt_metric` from it. Otherwise (locally-originated route), derive
`bgp_local_pref` from `krt_metric` as before.

**Before:**
```bird
if (!defined(krt_metric)) then { krt_metric = 1024; }
bgp_local_pref = 2147483647 - krt_metric;
```

**After:**
```bird
if (defined(bgp_local_pref)) then {
  krt_metric = 2147483647 - bgp_local_pref;
} else {
  if (!defined(krt_metric)) then { krt_metric = 1024; }
  bgp_local_pref = 2147483647 - krt_metric;
}
```


nelljerram:
> **Before:**
> 
> ```
> if (!defined(krt_metric)) then { krt_metric = 1024; }
> bgp_local_pref = 2147483647 - krt_metric;
> ```
> 
> **After:**
> 
> ```
> if (defined(bgp_local_pref)) then {
>   krt_metric = 2147483647 - bgp_local_pref;
> } else {
>   if (!defined(krt_metric)) then { krt_metric = 1024; }
>   bgp_local_pref = 2147483647 - krt_metric;
> }
> ```

I would prefer:
```
if (!defined(bgp_local_pref)) then {
  if (defined(krt_metric)) then {
    bgp_local_pref = 2147483647 - krt_metric;
  } else {
    bgp_local_pref = 2147483647 - 1024;
  }
}
```
because (having now had a bit more experience!) I don't think there is any point in setting `krt_metric` in a BGP export filter, and it looks a bit confusing to do that.  If `bgp_local_pref` is already defined, we can just leave it as is, and don't need to do anything.

song-jiang:
> I don't think there is any point in setting krt_metric in a BGP export filter

@nelljerram We still need setting krt_metric because `calico_aggr()` checks "krt_metric < 1024" to let elevated-priority /32s escape aggregation into /26 blocks.

nelljerram:
> > I don't think there is any point in setting krt_metric in a BGP export filter
> 
> @nelljerram We still need setting krt_metric because `calico_aggr()` checks "krt_metric < 1024" to let elevated-priority /32s escape aggregation into /26 blocks.

Oh, fine.  In that case I think your fix is the best choice for now.

confd/pkg/backends/calico/bgp_processor.go
```diff
@@ -746,18 +746,40 @@ func (c *client) buildExportFilter(
...
-	)
-
-	// Convert from krt_metric to BGP LOCAL_PREF.  Higher LOCAL_PREF = higher priority, but
-	// lower krt_metric = higher priority, so we invert: bgp_local_pref = INT_MAX - krt_metric.
-	// BGP LOCAL_PREF will only be propagated to iBGP peers; however it's helpful for us to set
-	// the bgp_local_pref attribute for both eBGP and iBGP peers, because then we can implement
-	// the Priority field as a match against bgp_local_pref.
-	filterLines = append(filterLines,
-		fmt.Sprintf("bgp_local_pref = %d - krt_metric;", template.BirdIntMaxValue),
+		"if (source = RTS_BGP && defined(bgp_local_pref)) then {",
```
skoryk-oleksandr: This check reads source without first verifying it’s defined. If source is ever undefined on a route, BIRD will raise a filter error and drop the route.

The same check in felix/etc/bird/calico-bird.conf.template:72 already uses the safe pattern:

`defined(source) && (source = RTS_BGP)
`
Suggest updating this to:

`if (defined(source) && source = RTS_BGP && defined(bgp_local_pref)) then { ... }
`

skoryk-oleksandr (commented):
Added one comment. Also not sure about this one : 
`felix/etc/bird/calico-bird.conf.template
`
Lines 57–58 of the static BIRD template still use the old pattern this PR is fixing:

```
if (!defined(krt_metric)) then { krt_metric = 1024; }
bgp_local_pref = 2147483647 - krt_metric;
```

This template is used by non-confd / OpenStack BIRD deployments. 
Leaving it unchanged means the same iBGP re-export issue can still affect those users?

confd/pkg/backends/calico/bgp_processor.go
song-jiang: Good point. Update my PR now.

skoryk-oleksandr: @song-jiang what about **IPV6** `calico-bird6.conf.template` ?

skoryk-oleksandr: committed ipv6 change

skoryk-oleksandr (approved):

PR #12554, opened 2026-04-20
Fix incorrect handle counting

djmitche:
Map indexing returns `found` not `ok`, with the opposite boolean sense. But indexing also returns the zero value when the map key does not exist, which happens to be useful here.

Fixes a bug identified in #12508.

**Release note:**
```release-note
Fix incorrect IPAM handle counting on release. 
```


caseydavenport (approved):
🤦‍♂️

PR #12555, opened 2026-04-20
Skip allocs with seq number conflicts in releaseByHandle

djmitche:
The existing implementation deleted the affected allocations but did not add them to `Deallocated`. Note that no current callers of this method pass a non-nil ReleaseOptions#SequenceNumber, and that this behavior is different from allocationBlock#release, which returns an error if any sequence numbers differ.

Fixes a bug identified in #12508. This is an alternative fix to the [one I proposed initially](https://github.com/projectcalico/calico/compare/master...djmitche:calico:push-xllllxwswqqn?expand=1), based on Slack conversations.

```release-note
IPAM: Fix potential early release of IP attributes.
```

### DIFF
--- a/libcalico-go/lib/ipam/ipam_block.go
+++ b/libcalico-go/lib/ipam/ipam_block.go
@@ -438,6 +438,10 @@ func sanitizeHandle(handleID string) string {
 	return strings.Split(handleID, "\r")[0]
 }
 
+// Release all allocations with the handle in opts.
+//
+// If opts.SequenceNumber is set, and any affected allocations' sequence numbers
+// do not match, those allocations are not released.
 func (b *allocationBlock) releaseByHandle(opts ReleaseOptions) int {
 	handleID := opts.Handle
 	attrIndexes := b.attributeIndexesByHandle(handleID)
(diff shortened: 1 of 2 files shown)


libcalico-go/lib/ipam/ipam_block.go
```diff
@@ -457,23 +461,33 @@ func (b *allocationBlock) releaseByHandle(opts ReleaseOptions) int {
...
-	// Clean and reorder attributes.
-	b.deleteAttributes(attrIndexes)
+	// Clean and reorder now-unused attributes.
+	deleteIndexes := make([]int, 0, len(attrIndexes))
+	for _, attrIndex := range attrIndexes {
+		if !usedAttrs[attrIndex] {
+			deleteIndexes = append(deleteIndexes, attrIndex)
+		}
+	}
+	b.deleteAttributes(deleteIndexes)
```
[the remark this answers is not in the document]
djmitche: This is an unusual circumstance and not worth the additional check.

libcalico-go/lib/ipam/ipam_block.go
```diff
@@ -457,23 +461,33 @@ func (b *allocationBlock) releaseByHandle(opts ReleaseOptions) int {
 
 	// There are addresses to release.
 	ordinals := []int{}
+	usedAttrs := make([]bool, len(b.Attributes))
```
caseydavenport: Mostly a nit, but I would call this something like "inUseAttrs" or "keepAttrs" to make this a bit more clear - "usedAttrs" implies these have been consumed somehow (to me at least). 

```suggestion
	keepAttrs := make([]bool, len(b.Attributes))
```

caseydavenport (commented):
Just a naming nit, but otherwise this LGTM

libcalico-go/lib/ipam/ipam_block.go
djmitche: Done -- good point!

caseydavenport (approved):

PR #12573, opened 2026-04-21
Fix incorrect comment on IPAM release behavour.

fasaxc:
Fix incorrect comment on IPAM release behavour.

`release()` returns the IPs that were skipped, not the IPs that were released.


caseydavenport (approved):

djmitche:
I should have read this more carefully when accepting the suggestion :)

PR #12579, opened 2026-04-21
Wait for CRDs to be Established in kube-controllers FV ApplyCRDs

caseydavenport:
The `policy name migration tests (kdd mode)` have been flaking with `no matches for kind "StagedNetworkPolicy" in version "crd.projectcalico.org/v1"`. The same race affects every kdd-mode test that uses `testutils.ApplyCRDs`.

`kubectl apply -f /crds/` returns as soon as the apiserver accepts each CRD object, but discovery registration happens asynchronously. The controller-runtime `crdClient` is constructed before `ApplyCRDs` runs, so when a test immediately calls `Create` on the new resource, its RESTMapper races the apiserver and returns `NoKindMatchError`.

Wait for all CRDs to report `Established` before `ApplyCRDs` returns. Same approach as in https://github.com/projectcalico/calico/pull/12225, pulled out as a targeted fix.

```release-note
None
```

### DIFF
--- a/kube-controllers/tests/testutils/utils.go
+++ b/kube-controllers/tests/testutils/utils.go
@@ -108,13 +108,13 @@ func ApplyCRDs(apiserver *containers.Container) {
 	// see a NoKindMatchError from the controller-runtime RESTMapper. Wait for all
 	// CRDs to report Established before returning.
 	waitEstablished := func() error {
-		out, err := apiserver.ExecOutput("kubectl", "wait", "--for=condition=Established", "--all", "crds", "--timeout=30s")
+		out, err := apiserver.ExecOutput("kubectl", "wait", "--for=condition=Established", "--all", "crds", "--timeout=5s")
 		if err != nil {
 			return fmt.Errorf("%s: %s", err, out)
 		}
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


kube-controllers/tests/testutils/utils.go
```diff
@@ -101,6 +101,20 @@ func ApplyCRDs(apiserver *containers.Container) {
...
 	}
 	EventuallyWithOffset(1, apply, 10*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
+
+	// kubectl apply returns as soon as the API server accepts each CRD object, but
+	// the apiserver needs additional time to register the CRD's API in discovery.
+	// Tests that hit the new CRD immediately race against that registration and can
+	// see a NoKindMatchError from the controller-runtime RESTMapper. Wait for all
+	// CRDs to report Established before returning.
+	waitEstablished := func() error {
+		out, err := apiserver.ExecOutput("kubectl", "wait", "--for=condition=Established", "--all", "crds", "--timeout=30s")
```
MichalFupso: This will retry 2 times, might be worth it to drop the timeout to 5s which might speed up the detection time as well?

MichalFupso (approved):

PR #12586, opened 2026-04-22
Sanitize log output

Behnam-Shobiri:
**Release note:**
```release-note
Sanitize log output
```


electricjesus (approved):
Approving only the app-policy changes

PR #12588, opened 2026-04-22
[release3.32] Sanitize log output (pick 12604)

Behnam-Shobiri:
Sanitize log output

**Release note:**
```release-note
Removed sensitive material (auth tokens, kubeconfig contents, etcd credentials, and inline certificates/keys) from log output. Logs that previously included full client-config or environment-variable dumps now log structured non-secret fields instead.
```
pick https://github.com/projectcalico/calico/pull/12604

### DIFF
--- a/app-policy/checker/match.go
+++ b/app-policy/checker/match.go
@@ -161,8 +161,6 @@ func computeNamespaceMatch(
 // matchRequest checks if the request part of the Rule matches the request. It returns true if the
 // Rule matches, false otherwise.
 func matchRequest(rule *proto.Rule, req *requestCache) bool {
-	// Do not log the request object, it may contain sensitive HTTP headers and bodies.
-	log.Debug("Matching request.")
 	return matchHTTP(rule.GetHttpMatch(), req.GetHttpMethod(), req.GetHttpPath())
 }
 
--- a/app-policy/checker/match.go
+++ b/app-policy/checker/match.go
@@ -161,7 +161,8 @@ func computeNamespaceMatch(
 // matchRequest checks if the request part of the Rule matches the request. It returns true if the
 // Rule matches, false otherwise.
 func matchRequest(rule *proto.Rule, req *requestCache) bool {
-	log.WithField("request", req).Debug("Matching request.")
+	// Do not log the request object, it may contain sensitive HTTP headers and bodies.
+	log.Debug("Matching request.")
 	return matchHTTP(rule.GetHttpMatch(), req.GetHttpMethod(), req.GetHttpPath())
 }
 
--- a/app-policy/checker/server.go
+++ b/app-policy/checker/server.go
@@ -86,8 +86,13 @@ func (as *authServer) RegisterGRPCServices(gs *grpc.Server) {
 func (as *authServer) Check(ctx context.Context, req *authz.CheckRequest) (*authz.CheckResponse, error) {
 	hostname, _ := os.Hostname()
 	logCtx := log.WithContext(ctx).WithField("hostname", hostname)
+	// Do not log req.Attributes.String() or req.String(), they contain sensitive HTTP
+	// headers (Authorization, Cookie) and request bodies.
 	if logCtx.Logger.IsLevelEnabled(log.DebugLevel) {
-		logCtx.Debug("Check start: ", req.Attributes.String())
+		logCtx.WithFields(log.Fields{
+			"method": req.GetAttributes().GetRequest().GetHttp().GetMethod(),
+			"path":   req.GetAttributes().GetRequest().GetHttp().GetPath(),
+		}).Debug("Check start")
 	}
 
 	resp := &authz.CheckResponse{Status: &status.Status{Code: INTERNAL}}
@@ -152,11 +157,12 @@ func (as *authServer) Check(ctx context.Context, req *authz.CheckRequest) (*auth
 		}
 	})
 
+	// Do not log the full CheckRequest, it may contain sensitive HTTP headers and bodies.
 	if logCtx.Logger.IsLevelEnabled(log.DebugLevel) {
 		logCtx.WithFields(log.Fields{
 			"code": code.Code(resp.Status.Code),
 			"msg":  resp.Status.Message,
-		}).Debug("Check complete: ", req.String())
+		}).Debug("Check complete")
 	}
 
 	return resp, nil
--- a/app-policy/checker/server.go
+++ b/app-policy/checker/server.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	core_v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
@@ -88,10 +89,16 @@ func (as *authServer) Check(ctx context.Context, req *authz.CheckRequest) (*auth
 	logCtx := log.WithContext(ctx).WithField("hostname", hostname)
 	// Do not log req.Attributes.String() or req.String(), they contain sensitive HTTP
 	// headers (Authorization, Cookie) and request bodies.
+	// Strip query string from the path — Envoy's :path pseudo-header includes it,
+	// and query parameters may carry tokens or credentials.
 	if logCtx.Logger.IsLevelEnabled(log.DebugLevel) {
+		httpPath := req.GetAttributes().GetRequest().GetHttp().GetPath()
+		if i := strings.IndexByte(httpPath, '?'); i >= 0 {
+			httpPath = httpPath[:i]
+		}
 		logCtx.WithFields(log.Fields{
 			"method": req.GetAttributes().GetRequest().GetHttp().GetMethod(),
-			"path":   req.GetAttributes().GetRequest().GetHttp().GetPath(),
+			"path":   httpPath,
 		}).Debug("Check start")
 	}
 
--- a/cni-plugin/pkg/k8s/k8s.go
+++ b/cni-plugin/pkg/k8s/k8s.go
@@ -299,8 +299,13 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 
 				newData, err := json.Marshal(stdinData)
 				if err != nil {
-					// Do not log stdinData, it may contain sensitive credentials (K8sAuthToken, K8sClientKey).
-					logger.WithError(err).Error("Error marshaling CNI stdin data")
+					// Do not log the full stdinData map, it may contain sensitive credentials
+					// (K8sAuthToken, K8sClientKey, K8sCertificateAuthority). Log only the keys.
+					keys := make([]string, 0, len(stdinData))
+					for k := range stdinData {
+						keys = append(keys, k)
+					}
+					logger.WithError(err).WithField("keys", keys).Error("Error marshaling CNI stdin data")
 					return nil, err
 				}
 				args.StdinData = newData
--- a/cni-plugin/pkg/k8s/k8s.go
+++ b/cni-plugin/pkg/k8s/k8s.go
@@ -300,7 +300,7 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 				newData, err := json.Marshal(stdinData)
 				if err != nil {
 					// Do not log stdinData, it may contain sensitive credentials (K8sAuthToken, K8sClientKey).
-				logger.WithError(err).Error("Error marshaling CNI stdin data")
+					logger.WithError(err).Error("Error marshaling CNI stdin data")
 					return nil, err
 				}
 				args.StdinData = newData
--- a/cni-plugin/pkg/k8s/k8s.go
+++ b/cni-plugin/pkg/k8s/k8s.go
@@ -299,7 +299,8 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 
 				newData, err := json.Marshal(stdinData)
 				if err != nil {
-					logger.WithField("stdinData", stdinData).Error("Error Marshaling data")
+					// Do not log stdinData, it may contain sensitive credentials (K8sAuthToken, K8sClientKey).
+				logger.WithError(err).Error("Error marshaling CNI stdin data")
 					return nil, err
 				}
 				args.StdinData = newData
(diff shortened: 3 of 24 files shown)


app-policy/checker/match.go
```diff
@@ -161,7 +161,8 @@ func computeNamespaceMatch(
 // matchRequest checks if the request part of the Rule matches the request. It returns true if the
 // Rule matches, false otherwise.
 func matchRequest(rule *proto.Rule, req *requestCache) bool {
-	log.WithField("request", req).Debug("Matching request.")
+	// Do not log the request object, it may contain sensitive HTTP headers and bodies.
+	log.Debug("Matching request.")
```
caseydavenport: This is a pretty useless log now.. we can probably just delete it.

app-policy/checker/server.go
```diff
@@ -86,8 +86,13 @@ func (as *authServer) RegisterGRPCServices(gs *grpc.Server) {
...
 	hostname, _ := os.Hostname()
 	logCtx := log.WithContext(ctx).WithField("hostname", hostname)
+	// Do not log req.Attributes.String() or req.String(), they contain sensitive HTTP
+	// headers (Authorization, Cookie) and request bodies.
 	if logCtx.Logger.IsLevelEnabled(log.DebugLevel) {
-		logCtx.Debug("Check start: ", req.Attributes.String())
+		logCtx.WithFields(log.Fields{
+			"method": req.GetAttributes().GetRequest().GetHttp().GetMethod(),
+			"path":   req.GetAttributes().GetRequest().GetHttp().GetPath(),
+		}).Debug("Check start")
```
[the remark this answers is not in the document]
caseydavenport: @Behnam-Shobiri WDYT about this?

cni-plugin/pkg/k8s/k8s.go
```diff
@@ -299,7 +299,13 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 
 				newData, err := json.Marshal(stdinData)
 				if err != nil {
-					logger.WithField("stdinData", stdinData).Error("Error Marshaling data")
+					// Do not log the full stdinData map, it may contain sensitive credentials
+					// (K8sAuthToken, K8sClientKey, K8sCertificateAuthority). Log only the keys.
+					keys := make([]string, 0, len(stdinData))
+					for k := range stdinData {
+						keys = append(keys, k)
```
caseydavenport: I don't think this is worth it - the goal of this log was to log out the bad input data so we can diagnose why it failed. If we're just logging the keys, we're not getting that information.

Just log the error, if we're not prepared to remove the sensitive data from the CNI configuration struct itself.

cni-plugin/pkg/k8s/k8s.go
```diff
@@ -986,7 +992,8 @@ func NewKubeVirtClient(conf types.NetConf, logger *logrus.Entry) (kubevirt.VirtC
 
 func getK8sNSInfo(client *kubernetes.Clientset, podNamespace string) (annotations map[string]string, err error) {
 	ns, err := client.CoreV1().Namespaces().Get(context.Background(), podNamespace, metav1.GetOptions{})
-	logrus.Debugf("namespace info %+v", ns)
+	// Do not log the full Namespace object, it may contain sensitive data in annotations.
```
caseydavenport: To be honest, this feels too aggressive to me. What sensitive data are you thinking about here? Namesapce annotations aren't inherently secret - end users can put sensitive data in plenty of inappropriate places... I don't think we can account for that.

cni-plugin/pkg/k8s/k8s.go
```diff
@@ -986,7 +992,8 @@ func NewKubeVirtClient(conf types.NetConf, logger *logrus.Entry) (kubevirt.VirtC
...
+	logrus.Debugf("namespace info name=%s", podNamespace)
```
caseydavenport:
```suggestion
	logrus.WithField("namespace", podNamespace).Debugf("Pod namespace")
```

We should use withFields.

cni-plugin/pkg/k8s/k8s.go
```diff
@@ -995,7 +1002,8 @@ func getK8sNSInfo(client *kubernetes.Clientset, podNamespace string) (annotation
 
 func getK8sPodInfo(client *kubernetes.Clientset, podName, podNamespace string) (labels map[string]string, annotations map[string]string, ports []internalapi.WorkloadEndpointPort, profiles []string, generateName, serviceAccount string, err error) {
 	pod, err := client.CoreV1().Pods(string(podNamespace)).Get(context.Background(), podName, metav1.GetOptions{})
-	logrus.Debugf("pod info %+v", pod)
+	// Do not log the full Pod object, it may contain sensitive data in env vars.
+	logrus.Debugf("pod info name=%s namespace=%s", podName, podNamespace)
```
caseydavenport: Should use WithFields here.

confd/pkg/backends/calico/client.go
```diff
@@ -1824,7 +1824,8 @@ func isValidCommunity(communityValue string) bool {
 // ParseFailed is called from the BGP syncer when an event could not be parsed.
 // We use this purely for logging.
 func (c *client) ParseFailed(rawKey string, rawValue string) {
-	log.Errorf("Unable to parse datastore entry Key=%s; Value=%s", rawKey, rawValue)
+	// Do not log rawValue, it may contain sensitive datastore entries (e.g. passwords).
+	log.Errorf("Unable to parse datastore entry Key=%s; ValueLen=%d", rawKey, len(rawValue))
```
caseydavenport, line 1830: Can we please not obfuscate ALL of our debug information in the case of errors like this?

If there are specific rawKey entries we shouldn't log, let's account for those and continue to log out important information needed for identification of errors?

confd/confd.go
```diff
@@ -26,7 +26,9 @@ func main() {
 	if err != nil {
 		log.Fatal(err.Error())
 	}
-	log.Infof("Config: %#v", c)
+	// Do not log the full config struct, Typha config contains TLS key/cert file paths.
+	log.Infof("Config: confDir=%s onetime=%t syncOnly=%t typhaAddr=%s typhaTLS=%t",
```
caseydavenport: Please take a copy and nil out the sensitive fields instead.

felix/config/param_types.go
```diff
@@ -653,7 +653,8 @@ func (p *EndpointListParam) Parse(raw string) (result any, err error) {
 		if u.Opaque != "" || u.User != nil || u.Path != "/" ||
 			u.RawPath != "" || u.RawQuery != "" ||
 			u.Fragment != "" {
-			log.WithField("url", fmt.Sprintf("%#v", u)).Error(
+			// Do not log the full URL, it may contain credentials in userinfo.
+			log.WithField("url", u.Host).Error(
 				"Unsupported URL part")
```
caseydavenport:
```suggestion
			log.WithField("url", u.Host).Error("Unsupported URL part")
```

felix/wireguard/wireguard.go
```diff
@@ -1775,7 +1775,8 @@ func (w *Wireguard) getNodeFromKey(key wgtypes.Key) *nodeData {
 
 // applyWireguardConfig applies the wireguard configuration.
 func (w *Wireguard) applyWireguardConfig(wireguardClient netlinkshim.Wireguard, c *wgtypes.Config) error {
-	w.logCtx.Debugf("Apply wireguard config update: %#v", c)
+	// Do not log the wgtypes.Config struct, it may contain the WireGuard PrivateKey.
+	w.logCtx.Debug("Apply wireguard config update")
```
caseydavenport: Same here - we can make a copy, nil the sensitive value(s), and log that so we don't lose all of our debug information. The log is more or less useless as it is.

goldmane/pkg/daemon/daemon.go
```diff
@@ -121,7 +121,12 @@ func newGRPCServer(cfg *Config) (*grpc.Server, error) {
 }
 
 func Run(ctx context.Context, cfg Config) {
-	logrus.WithField("cfg", cfg).Info("Loaded configuration")
+	// Do not log the full cfg struct, PushURL may contain sensitive credentials.
```
caseydavenport: Why do you say PushURL may contain sensitive credentials? I don't agree.

caseydavenport: And even so, we should follow the pattern of niling out the sensitive config value and continuing to log the whole struct.

guardian/pkg/server/server.go
```diff
@@ -77,7 +77,12 @@ func New(shutdownCtx context.Context, tunnelCert *tls.Certificate, dialer tunnel
 	}
 
 	for _, target := range srv.targets {
-		logrus.Infof("Will route traffic to %s for requests matching %s", target.Dest, target.Path)
+		// Do not log target.Dest directly, it may contain sensitive credentials in URL userinfo.
```
caseydavenport: Do you have any examples of this actually containing sensitive URL info?

typha/pkg/config/env_var_loader.go
```diff
@@ -25,6 +25,37 @@ import (
 // An environment entry of "TYPHA_FOO=bar" is translated to "foo": "bar".
 func LoadConfigFromEnvironment(environ []string) map[string]string {
 	result := make(map[string]string)
+	// isSensitiveParam checks whether a config parameter name suggests its value
+	// may contain credentials. Matching params have their values redacted in logs.
+	// Note: params ending in "file" (e.g. "etcdkeyfile") are file paths, not secrets.
+	sensitiveSubstrings := []string{
```
caseydavenport: This looks like duplicated logic from the Felix code above? We should use a common implementation.

typha/pkg/config/env_var_loader.go
```diff
@@ -37,8 +68,14 @@ func LoadConfigFromEnvironment(environ []string) map[string]string {
 		if strings.HasPrefix(key, "typha_") {
 			splits = strings.SplitN(key, "_", 2)
 			paramName := splits[1]
-			log.Infof("Found typha environment variable: %#v=%#v",
-				paramName, value)
+			// Redact values for env vars that may contain sensitive credentials.
+			if isSensitiveParam(paramName) {
+				log.Infof("Found typha environment variable: %s=<redacted>",
+					paramName)
```
caseydavenport:
```suggestion
				log.Infof("Found typha environment variable: %s=<redacted>",	paramName)
```

No need to truncate these lines - way too aggressive.

typha/pkg/config/env_var_loader.go
```diff
@@ -37,8 +68,14 @@ func LoadConfigFromEnvironment(environ []string) map[string]string {
...
+			} else {
+				log.Infof("Found typha environment variable: %s=%q",
+					paramName, value)
```
caseydavenport:
```suggestion
				log.Infof("Found typha environment variable: %s=%q", paramName, value)
```

typha/pkg/syncserver/sync_server.go
```diff
@@ -279,7 +279,11 @@ func (c *Config) requiringTLS() bool {
 
 func New(caches map[syncproto.SyncerType]BreadcrumbProvider, config Config) *Server {
 	config.ApplyDefaults()
-	log.WithField("config", config).Info("Creating server")
+	// Do not log the full config struct, it contains TLS key/cert file paths.
+	log.WithFields(log.Fields{
+		"port":       config.Port,
+		"tlsEnabled": config.KeyFile != "",
+	}).Info("Creating server")
```
caseydavenport: Same config logging pattern here

caseydavenport: Might even be good to have a helper Fields() function.

```go
func (c Config) Fields() logrus.Fields {
}
```

It can use reflection to build a logrus.Fields instance, that skips known sensitive entries. Seems like a better pattern to use throughout so we don't lose important debug information.

caseydavenport (commented):
@Behnam-Shobiri added some thoughts / comments - overall I am quite worried about the direction this is taking as we seem to be losing a lot of debug information with these changes, and many of these feel like they may not be taking into account what actual values are used - e.g., the URLs in question don't have sensitive information in them? 

Feel free to ping me in Slack too if you want to talk more.

confd/pkg/backends/calico/client.go
caseydavenport: @Behnam-Shobiri any chance we can do a similar IsSensitiveParam check here?

goldmane/pkg/emitter/emitter.go
```diff
@@ -326,3 +329,14 @@ func (e *Emitter) loadCachedState() error {
 	e.latestTimestamp = ts
 	return nil
 }
+
+// redactURL parses a URL string and returns its Redacted() form, which masks
+// any userinfo password.  If the URL cannot be parsed, returns the host or
+// a placeholder.
+func redactURL(raw string) string {
```
caseydavenport: We seem to have a few different ways of doing the same URL redaction - could we write a common helper in `libcalico-go` that can be called on all of the URLs rather than reimplementing in each component?

libcalico-go/lib/backend/etcdv3/etcdv3.go
```diff
@@ -156,7 +156,8 @@ func (c *etcdV3Client) Create(ctx context.Context, d *model.KVPair) (*model.KVPa
 	// Take a copy of the key before we default any policy names, we use this key for error returns to return the same name that user provided for create
 	keyCopy := d.Key
 
-	logCxt := log.WithFields(log.Fields{"model-etcdKey": d.Key, "value": d.Value, "ttl": d.TTL, "rev": d.Revision})
+	// Do not log d.Value, it may contain sensitive resource specs.
+	logCxt := log.WithFields(log.Fields{"model-etcdKey": d.Key, "ttl": d.TTL, "rev": d.Revision})
```
caseydavenport, line 160: Ideally we'd also use a "RedactSensitiveValues" helper model here.

typha/pkg/syncserver/sync_server.go
```diff
@@ -160,6 +160,30 @@ type Config struct {
 	DebugLogWrites bool
 }
 
+// LogFields returns a logrus.Fields map with all config values included,
+// but with TLS key/cert file paths redacted.
+func (c Config) LogFields() log.Fields {
+	return log.Fields{
```
caseydavenport, line 173: I think we can use reflection here - the main thing that I want to avoid is devs needing to remember to update this when a new field is added.

Behnam-Shobiri: I add a comment on structure so no one forget.

caseydavenport (approved):

PR #12592, opened 2026-04-22
Capture pod diagnostics on winfv-felix setup failure

caseydavenport:
When the winfv-felix setup times out waiting for pods, the existing trap only runs `kubectl get pod -A -o wide` - we know what failed but not why. A recent enterprise run timed out waiting on a Windows pod and there was no describe, no events, and no container logs in the job output to diagnose it from.

Adds a `collect_pod_diagnostics` helper in `process/testing/util/utils.sh` that writes describe + current-container and previous-container logs for every non-Ready pod into `/home/semaphore/fv.log/diagnostics/` (so Semaphore picks it up as a job artifact), with a `/tmp` fallback for hand runs. Wired into the `install-aso.sh` and `install-calico.sh` traps and the `setup-fv.sh` failure path in `run-win-fv.sh`. Also echoes the tail of each non-Ready container's logs straight to stdout so the failure is diagnosable from the job log alone without downloading artifacts.

Port of https://github.com/tigera/calico-private/pull/11644 - the four files touched are byte-identical between OSS and calico-private, and the helper is product-agnostic.

```release-note
None
```

### DIFF
--- a/process/testing/aso/install-aso.sh
+++ b/process/testing/aso/install-aso.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2024 Tigera, Inc. All rights reserved.
+# Copyright (c) 2024-2026 Tigera, Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


nelljerram:
Thanks @caseydavenport for this.

PR #12595, opened 2026-04-22
node: refresh CNI kubeconfig promptly on SA token rotation

skoryk-oleksandr:
## Summary

Fixes a multi-hour blind spot in `cni-config-monitor` where an externally-invalidated CNI kubeconfig token goes undetected between scheduled refreshes, leading to `plugin type="calico" failed (add): error getting ClusterInformation: connection is unauthorized: Unauthorized` until a calico-node pod restart.

## What this changes

`TokenRefresher.Run` now installs an `fsnotify` watcher on the directory containing the pod's projected ServiceAccount token. When kubelet rotates that token, the refresh loop wakes up immediately, mints a fresh token via `TokenRequest`, and rewrites `/etc/cni/net.d/calico-kubeconfig`. If watcher setup fails the refresher falls back to the original timer-only behaviour. Both `Events` and `Errors` channels are drained per fsnotify's contract.

The watcher is disabled at runtime on Windows (`runtime.GOOS == "windows"`); on Windows, kubelet's projected-volume layout differs and the fsnotify backend (`ReadDirectoryChangesW`) reports a different event shape. Windows still gets the original timer-driven refresh.

`Stop()` is idempotent via `sync.Once` so repeated cancellations are safe.

## Why now

The `TokenRefresher.Run` loop sleeps for `(tokenExp - now) / 4` + up to 100% jitter between refreshes. With the default 24h token validity that is 6-12 hours. During that sleep there is no mechanism that notices a token has become invalid externally. The two common triggers in practice are:

- ServiceAccount signing-key rotation (reproduced on kind; the failure is immediate and persistent until calico-node restarts).
- Any event that makes kubelet re-project the SA token before the scheduled refresh — picking it up promptly keeps the CNI kubeconfig as fresh as possible.

## End-to-end verification on a kind cluster

Brought up a kind cluster running this branch and ran through the scenario the fix targets.

**Fix activation confirmed** — on pod startup, calico-node logs show the fsnotify watcher is installed:
```
[INFO] node-services/token_watch.go: Watching service account token directory for rotation
```

**fsnotify fast-path demonstration** — triggered a projected-token change in the watched directory on a healthy cluster:

| | Before | After |
|---|---|---|
| CNI token `jti` | `59030a44-…` | `45043310-…` (new token) |
| CNI kubeconfig mtime | 20:56:07 | 21:01:47 |
| Latency trigger → new kubeconfig written | | **~30 ms** |

Log line, verbatim:
```
[INFO] node-services/token_watch.go: Projected service account token changed; refreshing CNI kubeconfig immediately
```

Same trigger on master (without this branch) produced **no response** — no log, no kubeconfig rewrite, no token change. That's the blind spot the fix closes.

**Underlying bug reproduction** — also reproduced the original failure end-to-end on the kind cluster by regenerating the SA signing key and restarting the apiserver. Pod creation fails with the exact error from the original report:
```
Failed to create pod sandbox: ... plugin type="calico" failed (add):
  error getting ClusterInformation: connection is unauthorized: Unauthorized
```

## Honest scope

This is a necessary-but-not-sufficient fix. On its own it changes worst-case automatic recovery from "up to 6-12 hours (or manual pod restart)" to "at most one kubelet projection cycle". With default Kubernetes settings the projected SA token lifetime is ~1 hour and kubelet refreshes at 80 % of that, so the practical upper bound after this fix is **~48 minutes** (average ~24 minutes, best case seconds).

That is still slow for time-sensitive workloads. Operators who need a tighter bound should additionally:
- Shorten the projected SA token's `expirationSeconds` on the calico-node DaemonSet (e.g. 600s → refresh at ~8 minutes). That change belongs in the operator/Helm chart, not here.
- Optionally add a health check that fails liveness after N minutes of `UpdateToken` errors so kubelet restarts the pod. Not included in this PR.

What this fix does NOT do:
- Eliminate transient `Unauthorized` errors when the SA signing key rotates. Every workload holding a projected SA token sees 401s until kubelet re-projects.
- Speed up kubelet's projected-token refresh cadence. That is controlled by the pod spec's `expirationSeconds`, set by the operator.

## Cross-platform validation

calico-node ships on Linux/amd64, Linux/arm64, and Windows/amd64, so the fix had to be safe on all three.

- **Linux/amd64** — full unit-test suite passes inside `calico/go-build:1.26.2-llvm20.1.8-k8s1.35.4` (the image and Go toolchain pinned in this repo).
- **Linux/arm64** — same suite re-run natively on a real Ampere arm64 GCP VM running the multi-arch go-build container. Build and `go vet` both clean.
- **Windows/amd64** — cross-compiled the unit-test binary (`GOOS=windows go test -c`), copied it to a real Windows Server 2022 GCP VM, executed it. The runtime guard is exercised in production code — the Windows test log shows `fsnotify-based CNI token fast path is disabled on Windows; refresh will rely on timer only` for every `startTokenFileWatcher` call. Tests that assert Linux/POSIX-specific kubelet atomic-writer behaviour are gated with `skipIfNotKubeletPosix` / `skipIfFsnotifyFastPathDisabled` and correctly skip on Windows; the rest of the suite passes there.

## Tests

Added `node/pkg/cni/token_watch_internal_test.go` with:

- `TestTokenRefresher_WakesUpOnTokenFileChange` — configures a 24h token (normally 6-12h sleep), writes a file in the watched dir, asserts a second `UpdateToken` fires within 3s. Without the fsnotify fast path this would time out. Skipped on Windows where the fast path is disabled.
- `TestTokenRefresher_FallsBackWhenWatcherSetupFails` — verifies the graceful timer-only fallback when the watched directory doesn't exist.
- `TestDrainEvents_IsNonBlockingAndDrainsBurst` — exercises the `drainEvents` helper on both full and empty channels.
- `TestTokenRefresher_DegradesGracefullyWhenWatcherChannelCloses` — uses an injected watcher factory returning a pre-closed channel, verifies the timer continues to drive refreshes.
- `TestTokenRefresher_RecoversAfterTransientUpdateTokenError` — first `UpdateToken` call fails, second succeeds; verifies the existing 5-10s retry path delivers a token after recovery.
- `TestFsnotifyDetectsKubeletAtomicWriterRotation` — uses a faithful port of kubelet's atomic-writer (kubernetes/kubernetes `pkg/volume/util/atomic_writer.go` @ `v1.35.4` — the version pinned in this repo's `go.mod`) to validate that the kubelet-style two-phase `Symlink` → `Rename` rotation actually fires `fsnotify` events that `drainEvents` picks up. Linux-only.
- `TestTokenRefresher_WakesUpOnAtomicWriterRotation` — full `TokenRefresher` driven through the same kubelet atomic-writer rotation; asserts the loop wakes up on the rotation. Linux-only.
- `TestTokenRefresher_HandlesBurstRotations` — issues N back-to-back kubelet-style rotations and asserts the refresher does not wedge or drop wake-ups. Linux-only.
- `TestTokenRefresher_NoGoroutineLeakOnRunStopCycle` — 30 successive Run/Stop cycles asserting goroutine count returns to baseline.
- `TestTokenRefresher_StopIsIdempotent` — verifies repeated `Stop()` calls do not panic (sync.Once).

All pass under `-race` with `-count=20` and stable under `-count=50` on Linux/amd64. Linux/arm64 and Windows/amd64 confirmed independently as described above.

**Release note:**
```release-note
calico/node now refreshes the CNI plugin's kubeconfig immediately when the pod's projected ServiceAccount token is rotated, closing a 6-12h window where an externally-invalidated token could cause CNI ADD to fail with "Unauthorized" until the calico-node pod was restarted.
```


node/pkg/cni/token_watch.go
```diff
@@ -39,13 +53,28 @@ type TokenRefresher struct {
...
 	namespace          string
 	serviceAccountName string
 
+	// tokenFilePath is the path to the in-pod projected service account token
+	// that this TokenRefresher uses (via the in-cluster client) to authenticate
+	// to the API server. The directory containing this file is watched with
+	// fsnotify so that the refresh loop wakes up immediately when kubelet
+	// rotates the projected token — otherwise an externally-invalidated CNI
+	// kubeconfig token can sit on disk for up to 12 hours before we notice.
+	// Overridable for tests.
```
caseydavenport:
```suggestion
	// tokenFilePath is the path to the in-pod projected service account token
	// that this TokenRefresher uses.
```

^ Claude comments can get a bit too verbose - it's actually counterproductive towards readability.

node/pkg/cni/token_watch.go
```diff
@@ -119,11 +151,27 @@ func (t *TokenRefresher) TokenChan() <-chan TokenUpdate {
...
 
 func (t *TokenRefresher) Run() {
+	// Watch the directory containing our own projected service account token
+	// so that we wake up as soon as kubelet rotates it. Two common cases make
+	// this important: (a) a signing-key rotation invalidates every
+	// previously-issued token while the exp claim is still in the future, so
+	// the timer alone wouldn't notice until the next scheduled refresh (up to
+	// ~12 hours later); (b) kubelet may re-project the token faster than our
+	// refresh cadence, and picking up the new token promptly keeps the CNI
+	// kubeconfig as fresh as possible.
```
caseydavenport:
```suggestion
	// Watch the directory containing our own projected service account token
	// so that we wake up as soon as kubelet rotates it (e.g., signing-key rotation or any 
	// refresh faster than our periodic check).
```

Same over-commenting.

node/pkg/cni/token_watch.go
```diff
@@ -145,14 +193,127 @@ func (t *TokenRefresher) Run() {
...
 		select {
-		case <-time.After(sleepTime):
+		case <-timer.C:
+		case ev, ok := <-tokenRotated:
+			if !timer.Stop() {
+				<-timer.C
+			}
+			if !ok {
+				// Watcher was closed — fall back to timer-only behaviour.
+				tokenRotated = nil
```
caseydavenport: Worth a log on this line? This isn't expected in mainline usage right?

node/pkg/cni/token_watch.go
```diff
@@ -145,14 +193,127 @@ func (t *TokenRefresher) Run() {
...
+			} else {
+				logrus.WithField("event", ev.String()).Info("Projected service account token changed; refreshing CNI kubeconfig immediately")
```
caseydavenport: This doesn't include the token value, right?

node/pkg/cni/token_watch.go
```diff
@@ -119,11 +151,27 @@ func (t *TokenRefresher) TokenChan() <-chan TokenUpdate {
...
 func (t *TokenRefresher) Run() {
+	// Watch the directory containing our own projected service account token
+	// so that we wake up as soon as kubelet rotates it. Two common cases make
+	// this important: (a) a signing-key rotation invalidates every
+	// previously-issued token while the exp claim is still in the future, so
+	// the timer alone wouldn't notice until the next scheduled refresh (up to
+	// ~12 hours later); (b) kubelet may re-project the token faster than our
+	// refresh cadence, and picking up the new token promptly keeps the CNI
+	// kubeconfig as fresh as possible.
+	tokenRotated, cleanupWatcher := t.watcherFactory()
```
caseydavenport: I'm not really sure the factory pattern makes sense here, given we only ever create a single watcher?

I think this would be cleaner if we used a distinct type / interface pattern:

```go
type FileNotifier interface {
  Events() <- Event
  Close() 
}

type fileNotifier struct {
  ... 
}

func NewFileNotifier(...) FileNotifier

func NewFakeFileNotifier(...) FileNotifier\
```

or something like that.

node/pkg/cni/token_watch.go
```diff
@@ -145,14 +193,127 @@ func (t *TokenRefresher) Run() {
...
+				// Watcher was closed — fall back to timer-only behaviour.
+				tokenRotated = nil
+			} else {
+				logrus.WithField("event", ev.String()).Info("Projected service account token changed; refreshing CNI kubeconfig immediately")
+				// Coalesce a kubelet atomic-writer rotation burst into a
+				// single UpdateToken so cluster-wide signing-key rotation
+				// does not multiply at kube-apiserver. coalesceRotationBurst
+				// returns false if Stop was requested while waiting, in
+				// which case we exit Run cleanly.
+				if !t.coalesceRotationBurst(tokenRotated) {
```
caseydavenport: For a clean split of concerns, I think this coalesce logic belongs on the other side of the channel - it's a bit more complicated though since it means a channel to receive the fsnotify events and another channel to send the coalesced event. 

Up to you if you want to change / think it's worth the complexity, but I think it would make the interface between these two components nicer.

node/pkg/cni/token_watch.go
```diff
@@ -145,14 +193,127 @@ func (t *TokenRefresher) Run() {
...
+// behaviour — this preserves the original semantics on platforms where
+// fsnotify isn't supported.
+//
+// On Windows the watcher is intentionally skipped. fsnotify on Windows uses
+// ReadDirectoryChangesW, whose event delivery can vary depending on how the
+// writer updates the file (rename vs. truncate vs. atomic-symlink-swap),
+// and projected SA token volumes on Windows host-process containers do not
+// follow the same kubelet atomic-writer pattern this watcher was designed
+// against. Forcing timer-only on Windows keeps behaviour identical to the
+// pre-fsnotify implementation — no regression risk on Windows nodes.
```
caseydavenport: This comment is also a mouthful

The "no regression risk on Windows nodes" is only relevant in the context of this PR - a year from now this code will be the norm and "regression" will be meaningless, so we should phrase the comments such that they apply to the code, not the code diff

Much of this comment is evident from reading the code immediately below (e.g., it's pretty obvious this what the Windows behavior is, and it's a private method so it's not like we're exporting this for arbitrary consumers that need details docstrings)

node/pkg/cni/token_watch.go
```diff
@@ -198,6 +359,15 @@ func tokenUpdateFromFile() (TokenUpdate, error) {
...
 		return TokenUpdate{}, err
 	}
+	return parseTokenUpdate(tokenBytes)
+}
+
+// parseTokenUpdate decodes a JWT-formatted service account token into a
+// TokenUpdate. Split out from tokenUpdateFromFile so tests can exercise
+// the parsing path (including malformed/missing claims) without writing
+// real files. Returns a typed error for any malformed claim — never panics
+// — so a corrupt token cannot crash the calico-node process.
```
caseydavenport:
```suggestion
// parseTokenUpdate decodes a JWT-formatted service account token into a
// TokenUpdate and returns an error if it is invalid or doesn't meet our requirements. 
```

Again comment is describing mostly what changed in this PR, and not long-term useful documentation.

node/pkg/cni/token_watch.go
```diff
@@ -145,14 +193,127 @@ func (t *TokenRefresher) Run() {
...
+// coalesceWindow, draining any events that arrive during that period. This
+// collapses a kubelet atomic-writer rotation burst (~5 events emitted within
+// a few milliseconds) into one UpdateToken call so cluster-wide SA signing
+// key rotation does not turn into a thundering herd at kube-apiserver.
+//
+// Returns true once the window is quiet, or false if Stop was requested
+// while waiting (in which case the caller must exit Run).
+//
+// If events is nil (watcher was previously closed), the function returns
+// once the window elapses with no events.
```
caseydavenport: I think this would be better represented as inline commenting - it's complicated enough that having breadcrumbs throughout would be better than a big block of text followed by a big block of code.

caseydavenport (changes requested):
Mostly comments on the Claude-generated comments and some nits - LMK what you think. Otherwise logic looks correct as far as I can tell!

caseydavenport (approved):

PR #12607, opened 2026-04-23
fix(ci): free up disk space for (hash)release runs

radTuti:
Followup to #12585 as the hashrelease CI pipeline still failed due to disk space.
This cleans up existing files in `f1-standard-4`

**Release note:**
```release-note
N/A
```

### DIFF
--- a/.semaphore/release/hashrelease.yml
+++ b/.semaphore/release/hashrelease.yml
@@ -29,6 +29,7 @@ global_job_config:
 
   prologue:
     commands:
+      - sudo rm -rf ~/.phpbrew ~/.kerl ~/.sbt ~/.nvm ~/.npm ~/.kiex /usr/local/golang/ /usr/lib/jvm /opt/* /opt/az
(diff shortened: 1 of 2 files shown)


danudey (approved):

.semaphore/release/hashrelease.yml
```diff
@@ -29,6 +29,7 @@ global_job_config:
 
   prologue:
     commands:
+      - sudo rm -rf ~/.phpbrew ~/.kerl ~/.sbt ~/.nvm ~/.npm ~/.kiex /usr/local/golang/ /usr/lib/jvm /opt/* /opt/az
```
hjiawei, line 32: How much space does this cleanup step actually save? This might save a few MBs, but I’m not sure it’s worth the added clutter. Was this simply carried over from the e1 era?

It might be fine for hashrelease but deleting .nvm, .npm, or /opt/* will also break builds and unregister semaphore agents on self-hosted runners.

PR #12612, opened 2026-04-24
Use user's centralized go-cache if available

danudey:
This PR replaces the `.go-pkg-cache` in the repository root with the user's existing `GOCACHE` directory (from `go env GOCACHE`) if it exists.

Resolution order, first match wins:

1. `LOCAL_GO_PKG_CACHE` from the environment — the opt-out; point it anywhere, including back at `$(REPO_ROOT)/.go-pkg-cache`
2. `go env GOCACHE`, when it resolves to an absolute path (otherwise it is the literal `off`)
3. `$(REPO_ROOT)/.go-pkg-cache`, as before

Benefits:

1. Saves a lot of disk space: every checkout used to carry its own copy of the compiled build output (~1.5G after building one component). The *module* cache was already shared, so this is only about build artifacts.
2. Saves compilation time across checkouts: building `kube-controllers` in a second worktree against a warm shared cache took 40s → 5s, with no cache growth.
3. The cache no longer lives inside the repo, so `git clean -xdf` no longer destroys it.

CI is unaffected: the Semaphore prologue pins `LOCAL_GO_PKG_CACHE` to `$HOME/${CALICO_DIR_NAME}/.go-pkg-cache`, the path the epilogue already tars and restores. The argoci e2e lane resolves the cache the same way, so its host-side build and its container share one.

Knock-on for the in-repo controller-gen, which is built into that cache. Now that checkouts share the directory: its stamp includes a hash of `build.sh` and the patches (`build.sh` is an order-only prerequisite, since mtimes are meaningless across checkouts); it is written via `mktemp` + `mv` so concurrent builds cannot truncate a binary another container is executing; and stamps untouched for 30 days are pruned, as `go clean -cache` leaves `bin/` alone.

Drawbacks:

1. Containers running as root could leave root-owned entries in the user's cache. The `RUN_AS_ROOT=true` targets (all in `felix/Makefile`) only run pre-built test binaries and `go tool test2json`, neither of which populates the build cache, so this does not currently bite.
2. `go clean -cache` now discards CI-shaped build state along with the user's own.
3. Host-run `go build`/`go test` write to the same directory but mostly miss the containers' entries, because containers set `GOAMD64=v2` while the host default is `v1`; `export GOAMD64=v2` closes the gap. Pre-existing, but previously invisible because the two caches were separate directories.

**Release note:**
```release-note
None
```

**AI assistance:** Claude Code — reviewed the change and applied the resulting fixes (temp-file collision, order-only prerequisite, `LOCAL_GO_PKG_CACHE` override handling, stamp pruning, argoci lane).

### DIFF
--- a/lib.Makefile
+++ b/lib.Makefile
@@ -361,7 +361,10 @@ DOCKER_BUILD=docker buildx build --load --platform=linux/$(ARCH) $(DOCKER_PULL) 
 	--build-arg GIT_VERSION=$(GIT_VERSION) \
 	--build-arg UBI_IMAGE=$(UBI_IMAGE)
 
-DOCKER_RUN_PRIV_NET := mkdir -p $(REPO_ROOT)/.go-pkg-cache bin $(GOMOD_CACHE) && \
+# Re-use the local user's GOCACHE directory, if there is one; otherwise, use the old .go-pkg-cache dir
+LOCAL_GO_PKG_CACHE ?= $(if $(shell go env GOCACHE),$(shell go env GOCACHE),$(REPOROOT)/.go-pkg-cache)
+
+DOCKER_RUN_PRIV_NET := mkdir -p $(LOCAL_GO_PKG_CACHE) bin $(GOMOD_CACHE) && \
 	docker run --rm \
 		--init \
 		$(EXTRA_DOCKER_ARGS) \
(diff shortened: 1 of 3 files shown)


lib.Makefile
```diff
@@ -361,7 +361,39 @@ DOCKER_BUILD=docker buildx build --load --platform=linux/$(ARCH) $(DOCKER_PULL)
...
+DEFAULT_GO_CACHE_PATH := $(REPOROOT)/.go-pkg-cache
+
+# Detect what path to use as the user's go cache path.
+# 
+# Setting LOCAL_GO_PKG_CACHE in your environment overrides this
+# unconditionally; setting GOCACHE overrides it if the value
+# you provide passes the sanity checks below.
+
+# Set the path (outside of containers) to mount as the go cache path
+GOENV_GOCACHE := $(strip $(shell go env GOCACHE 2>/dev/null))
```
radTuti, line 377: i think we want the mod cache
```suggestion
GOENV_GOCACHE := $(strip $(shell go env GOMODCACHE 2>/dev/null))
```

caseydavenport:
@danudey nice! this is something I was hoping to get done in the near future, thanks for picking it up!

lib.Makefile
danudey: `GOMODCACHE` is for downloaded go modules; `GOCACHE` is for built golang code. Honestly doing both might be the best call.

caseydavenport (approved):

PR #12613, opened 2026-04-24
Fix multi-arch build

caseydavenport:
**Release note:**
```release-note
TBD
```


docker/calico/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG CALICO_BASE
 ARG UBI_IMAGE
 
-FROM ${UBI_IMAGE} AS builder
+FROM --platform=$BUILDPLATFORM ${UBI_IMAGE} AS builder
```
hjiawei: Do you need `ARG BUILDPLATFORM` before the builder stage? Alternatively, you could use the architecture-specific image since you’ve already registered the emulator.

PR #12614, opened 2026-04-24
fix(e2e): build libbpf for amd64 multi-arch e2e binary target

radTuti:
The multi-arch pattern rule `bin/k8s/e2e-linux-%.test` (added in #12506) dropped the `$(LIBBPF_A)` prereq and `DOCKER_RUN_CGO` wiring. Only amd64 needs them — other arches have CGO off and compile the `//go:build !cgo` stub. amd64 fails to link on a clean tree with `undefined reference to libbpf_strerror_r`.

Add an explicit `bin/k8s/e2e-linux-amd64.test` rule (CGO + libbpf) and have `bin/k8s/e2e.test` rename it for local use (`mv`, not `ln`).

**Release note:**
```release-note
None
```

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -23,18 +23,17 @@ SRC_FILES=$(shell find pkg cmd -name '*.go')
 
 build: bin/k8s/e2e.test bin/clusternetworkpolicy/e2e.test $(SRC_FILES)
 
-bin/k8s/e2e.test: $(LIBBPF_A) $(SRC_FILES)
-	mkdir -p bin
-	$(DOCKER_RUN_CGO) $(CALICO_BUILD) go test ./cmd/k8s -c -o $@
+bin/k8s/e2e.test: bin/k8s/e2e-linux-$(ARCH).test
+	mv $< $@
 
 bin/clusternetworkpolicy/e2e.test: $(LIBBPF_A) $(SRC_FILES)
 	mkdir -p bin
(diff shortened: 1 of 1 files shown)


e2e/Makefile
```diff
@@ -31,10 +31,13 @@ bin/clusternetworkpolicy/e2e.test: $(LIBBPF_A) $(SRC_FILES)
...
-# Multi-arch e2e binary targets. Each architecture gets its own binary, built
-# via cross-compilation inside the calico/go-build container. Uses a pattern
-# rule so that any arch in VALIDARCHES (including ppc64le, s390x) works.
-# The recursive make sets ARCH so that DOCKER_RUN passes the correct GOARCH.
+# Multi-arch e2e binary targets. amd64 reuses bin/k8s/e2e.test (CGO + libbpf);
+# non-amd64 arches default to CGO off and use the bpf_stub.go path.
+bin/k8s/e2e-linux-amd64.test: $(SRC_FILES)
+	mkdir -p bin/k8s
+	$(MAKE) bin/k8s/e2e.test ARCH=amd64
+	ln -f bin/k8s/e2e.test $@
```
caseydavenport:
```suggestion
bin/k8s/e2e.test: bin/k8s/e2e-linux-amd64.test
  cp bin/k8s/e2e-linux-amd64.test $@
```

I think this is how I'd format it, but I am curious why we need this at all? Can't we just use the properly named per-arch binary?

caseydavenport (approved):

PR #12619, opened 2026-04-24
[CNI] Create netkit L2 pair for pod interface when device_type=netkit

tomastigera:
Adds a new top-level `device_type` field to the CNI configuration. When set to `netkit`, the CNI creates a netkit L2 pair (Linux 6.7+) instead of a veth pair for the pod interface; on older kernels the LinkAdd fails with `EOPNOTSUPP` and we silently fall back to veth so a mixed cluster keeps working. Other errno values propagate so we don't mask real failures.

Netkit lets BPF programs run inside `ndo_start_xmit()` via `BPF_NETKIT_PRIMARY` / `BPF_NETKIT_PEER`, which (per ByteDance) gives ~10% throughput and a major P99 latency improvement under contention with the BPF dataplane. Felix builds without explicit netkit awareness still attach via TC/TCX since netkit accepts the same `BPF_PROG_TYPE_SCHED_CLS` programs as veth, so this PR can ship independently of the Felix-side native netkit work.

For non-BPF dataplanes netkit L2 is Ethernet-compatible (iptables/nftables work unchanged); functionally fine, performance-neutral.

Related: [CORE-10672](https://tigera.atlassian.net/browse/CORE-10672). The Installation-spec / operator plumbing of this knob is a follow-up under the same epic.

**Release note:**
```release-note
Calico CNI plugin now supports a `device_type` configuration option to create netkit (Linux 6.7+) virtual devices instead of veth for the pod interface. Defaults to veth.
```

[CORE-10672]: https://tigera.atlassian.net/browse/CORE-10672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
+++ b/cni-plugin/pkg/dataplane/linux/dataplane_linux.go
@@ -16,6 +16,7 @@ package linux
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -47,6 +48,7 @@ type LinuxDataplane struct {
 	allowIPForwarding bool
 	mtu               int
 	queues            int
+	deviceType        string
 	logger            *logrus.Entry
 }
 
@@ -55,6 +57,7 @@ func NewLinuxDataplane(conf types.NetConf, logger *logrus.Entry) *LinuxDataplane
 		allowIPForwarding: conf.ContainerSettings.AllowIPForwarding,
 		mtu:               conf.MTU,
 		queues:            conf.NumQueues,
+		deviceType:        conf.DeviceType,
 		logger:            logger,
 	}
 }
@@ -141,17 +144,16 @@ func (d *LinuxDataplane) DoWorkloadNetnsSetUp(
 		la.MTU = d.mtu
 		la.NumTxQueues = d.queues
 		la.NumRxQueues = d.queues
-		veth := &netlink.Veth{
-			LinkAttrs:     la,
-			PeerName:      hostVethName,
-			PeerNamespace: netlink.NsFd(int(hostNS.Fd())),
-		}
 
 		var expectedHostSideIPv6Addr net.IP
-		if err := netlink.LinkAdd(veth); err != nil {
-			d.logger.Errorf("Error adding veth %+v: %s", veth, err)
+		createdType, err := d.addWorkloadLink(la, hostVethName, hostNS)
+		if err != nil {
 			return err
 		}
+		d.logger.WithFields(logrus.Fields{
+			"type":      createdType,
+			"requested": d.deviceType,
+		}).Info("Created pod interface")
 
 		hostVeth, err := hostNlHandle.LinkByName(hostVethName)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


cni-plugin/pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -141,17 +146,16 @@ func (d *LinuxDataplane) DoWorkloadNetnsSetUp(
...
-		}
 
 		var expectedHostSideIPv6Addr net.IP
-		if err := netlink.LinkAdd(veth); err != nil {
-			d.logger.Errorf("Error adding veth %+v: %s", veth, err)
+		createdType, err := d.addWorkloadLink(la, hostVethName, hostNS)
+		if err != nil {
 			return err
 		}
+		d.logger.WithFields(logrus.Fields{
```
caseydavenport, line 155: I think Info level logging is too verbose for this.

cni-plugin/pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -591,6 +595,77 @@ func (d *LinuxDataplane) configureContainerSysctls(hasIPv4, hasIPv6 bool) error
...
+		}
+		if addErr == nil {
+			return types.DeviceTypeNetkit, nil
+		}
+		if !isNetkitUnsupported(addErr) {
+			d.logger.WithError(addErr).Error("Error adding netkit link")
+			return "", addErr
+		}
+		d.logger.WithError(addErr).Info(
+			"netkit not supported on this kernel; falling back to veth")
```
caseydavenport:
```suggestion
		d.logger.WithError(addErr).Info("netkit not supported on this kernel; falling back to veth")
```

nit, but we can keep this on a single line.

cni-plugin/pkg/dataplane/linux/dataplane_linux_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+// TestAddWorkloadLinkIntegration exercises addWorkloadLink against real
+// temporary netnses. Both the "host" and "container" sides live in fresh
+// unshared netnses so the test is fully isolated from the environment it
+// runs in (avoiding collisions with leftover/real interfaces in CI which may
+// run with --net=host). Requires root (CAP_SYS_ADMIN to unshare netns) and,
+// for the netkit subtest, kernel 6.7+.
+func TestAddWorkloadLinkIntegration(t *testing.T) {
+	if os.Geteuid() != 0 {
+		t.Skip("skipping integration test: requires root (unshare netns)")
```
caseydavenport: I think we should fail explicitly. I would be worried about subtle bugs where we don't notice these tests aren't running.

cni-plugin/pkg/dataplane/linux/dataplane_linux_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}{
+		{"default-is-veth", "", types.DeviceTypeVeth, false},
+		{"explicit-veth", types.DeviceTypeVeth, types.DeviceTypeVeth, false},
+		{"netkit-on-67", types.DeviceTypeNetkit, types.DeviceTypeNetkit, true},
+	}
+
+	for _, tc := range subtests {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.needs67 && !kernelAtLeast(t, 6, 7) {
+				t.Skip("skipping: requires kernel 6.7+ for netkit")
```
caseydavenport: Same here

cni-plugin/pkg/dataplane/linux/dataplane_linux_test.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+
+			hostNS, err := ns.TempNetNS()
+			if err != nil {
+				t.Fatalf("TempNetNS (host): %v", err)
+			}
+			defer func() { _ = hostNS.Close() }()
```
caseydavenport: Probably should at least log these errors so we have the diags.

cni-plugin/pkg/plugin/plugin.go
```diff
@@ -208,6 +208,17 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
...
 	}
 
+	// Validate and normalize DeviceType. Unknown values fall back to veth to
+	// keep old and misconfigured CNI configs working.
+	switch conf.DeviceType {
+	case "", types.DeviceTypeVeth, types.DeviceTypeNetkit:
+		// OK.
+	default:
+		logrus.WithField("device_type", conf.DeviceType).Warn(
+			"Unknown device_type in CNI config, falling back to veth.")
```
caseydavenport:
```suggestion
		logrus.WithField("device_type", conf.DeviceType).Warn("Unknown device_type in CNI config, falling back to veth.")
```

Should we just error out properly here, forcing the user to fix their (broken) config?

cni-plugin/pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -591,6 +595,77 @@ func (d *LinuxDataplane) configureContainerSysctls(hasIPv4, hasIPv6 bool) error
...
+func (d *LinuxDataplane) addWorkloadLink(la netlink.LinkAttrs, hostName string, hostNS ns.NetNS) (string, error) {
+	if d.deviceType == types.DeviceTypeNetkit {
+		primaryAttrs := la
+		primaryAttrs.Name = hostName
+
+		var addErr error
+		err := hostNS.Do(func(contNS ns.NetNS) error {
+			peerAttrs := netlink.NewLinkAttrs()
+			peerAttrs.Name = la.Name
+			peerAttrs.Namespace = netlink.NsFd(int(contNS.Fd()))
```
caseydavenport, line 625: Container-side netkit peer MTU is not propagated. `peerAttrs` is created via `netlink.NewLinkAttrs()` with only `Name`/`Namespace` set, so the peer MTU defaults to the kernel's 1500. Unlike veth (where the netlink helper falls back to the primary's MTU when the peer is zero), `addNetkitAttrs` only emits `IFLA_NETKIT_MTU` when `peerLinkAttrs.MTU > 0`. With overlay/VXLAN/Wireguard MTUs (often <1500), the container-side iface will be mismatched against the host side. Same bug was just fixed in the FV test workload by [70648ce31f](https://github.com/projectcalico/calico/commit/70648ce31f) — needs `peerAttrs.MTU = la.MTU` here.

caseydavenport, line 625: Same pattern for queue counts: `la.NumTxQueues`/`la.NumRxQueues` get set on the host-side primary attrs upstream, but `peerAttrs` is constructed fresh and zero queues do not propagate through netkit (veth does propagate via the netlink helper's fallback). Deployments that set `NumQueues > 1` will silently get a single-queue container-side iface for netkit pods. Copy `NumTxQueues`/`NumRxQueues` onto `peerAttrs` alongside the MTU.

tomastigera: great catch, fixed

tomastigera: netlink does not propagate queues and new kernels even hardwire it to 1 and reject anything else.

caseydavenport (approved):

PR #12625, opened 2026-04-25
release: always produce calicoctl binaries to fix hashrelease packaging

caseydavenport:
The hashrelease packaging step copies \`calicoctl/bin/\` into the staged release output and currently fails with:

\`\`\`
failed to copy calicoctl/bin/ to .../bin/calicoctl: exit status 1: cp: cannot stat 'calicoctl/bin/': No such file or directory
\`\`\`

\`buildBinaries\` returns early when \`buildImages=true\`, on the assumption that the calicoctl binaries fall out of the calicoctl image build. After the calico binary consolidation in https://github.com/projectcalico/calico/pull/12225 the calicoctl image was removed, so nothing produces those binaries on the hashrelease path.

Build calicoctl unconditionally; cni-plugin and felix stay on the existing \`!buildImages\` branch since their binaries are still produced by their image release-build targets.

Targeted minimal fix to unblock master. A broader cleanup (including consolidating calicoctl into the cmd/calico binary) is in https://github.com/projectcalico/calico/pull/12624.

\`\`\`release-note
NONE
\`\`\`

### DIFF
--- a/release/pkg/manager/calico/manager.go
+++ b/release/pkg/manager/calico/manager.go
@@ -1114,15 +1114,13 @@ func (r *CalicoManager) buildE2EBinaries() error {
 }
 
 func (r *CalicoManager) buildBinaries() error {
-	// Skip building binaries if we are building images
-	// binaries are built as part of "release-build" target.
-	if r.buildImages {
-		return nil
-	}
-	m := map[string]string{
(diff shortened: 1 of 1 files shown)


caseydavenport:
Going to cheekily merge this to try and expedite getting hashreleases back on track.

PR #12629, opened 2026-04-26
release: per-step log files and per-push timing for hashrelease

caseydavenport:
## Description

The release tool buffers each component's `make` output and dumps it through a single `logrus.Info(out)` after the make finishes, so every line in the publish log carries the same `INFO[NNNN]` timestamp. That makes it impossible to tell which docker push inside a 30-minute publish was actually slow, which has been a real problem for diagnosing recent hashrelease regressions.

This PR makes the publish phase debuggable from the Semaphore artifact UI:

- `release/internal/command`: drop the debug-level gate in `RunInDir` and always tee stdout/stderr line-by-line through logrus, so each line gets its own timestamp in the live job log. Add `RunInDirToFile(...)` for callers that also want a per-call log file written alongside.
- `release/pkg/manager/calico`: add a `logsDir` (default `<repoRoot>/release/_logs`) and a `makeInDirectoryToFile` helper. Switch `release-build`, `image-windows`, `release-publish`, and `release-windows` to write per-component files like `_logs/publish/node.log`, `_logs/publish/third_party-envoy-gateway.log`.
- `lib.Makefile`: add a `log_step` macro and bracket each `docker push` (per-arch) and `docker manifest create` / `docker manifest push` with grep-able records:
  ```
  ==> step=start label="push docker.io/calico/node:v3.33.0-...-amd64" epoch=1777212804 at=2026-04-26T14:13:24Z
  ```
  Per-push duration falls out of an `awk` over those pairs.
- `.semaphore/release/hashrelease.yml`: set `LOGS_DIR` in the prologue and add an epilogue that runs `artifact push job ${LOGS_DIR} --destination semaphore/hashrelease-logs` so the per-step files end up in the Semaphore artifact tree, matching the existing pattern in `.semaphore/semaphore.yml.d/blocks/20-felix.yml:166`.

After this lands, `release/_logs/publish/node.log` from any hashrelease run gives the full make output for node's publish in real time, with start/end timing markers around every `docker push`. That should answer "why did node take 27 minutes?" directly.

```release-note
None
```


release/pkg/manager/calico/manager.go
```diff
@@ -102,6 +102,11 @@ func NewManager(opts ...Option) *CalicoManager {
 		logrus.Fatal("No repo root specified")
 	}
 	logrus.WithField("repoRoot", b.repoRoot).Info("Using repo root")
+
+	if b.logsDir == "" {
+		b.logsDir = filepath.Join(b.repoRoot, "release", "_logs")
+	}
+	logrus.WithField("logsDir", b.logsDir).Info("Using logs directory")
```
radTuti: can this be part of config and passed with options?

ideally logs will be `release/_logs/VERSION` when passed in.
CI can the override to preferred location

release/pkg/manager/calico/manager.go
```diff
@@ -1241,21 +1252,17 @@ func (r *CalicoManager) buildContainerImages() error {
 	}
 
 	for _, dir := range releaseDirs {
-		out, err := r.makeInDirectoryWithOutput(filepath.Join(r.repoRoot, dir), "release-build", env...)
-		if err != nil {
-			logrus.Error(out)
+		fullDir := filepath.Join(r.repoRoot, dir)
+		if _, err := r.makeInDirectoryToFile(fullDir, "release-build", "build", env...); err != nil {
```
radTuti, line 1254: prefer to have this condition on if the logsDir exists

release/internal/command/commands.go
```diff
@@ -34,27 +35,66 @@ type CommandRunner interface {
...
+			logrus.WithError(cerr).Warnf("Failed to close log file %s", logPath)
+		}
+	}()
+	return r.runInDir(dir, name, args, env, f, f)
+}
+
+// runInDir is the shared implementation. extraOut and extraErr (if non-nil) receive
+// a copy of the child's stdout / stderr in addition to the captured buffer and the
+// live os.Stdout / os.Stderr stream.
+func (r *RealCommandRunner) runInDir(dir, name string, args, env []string, extraOut, extraErr io.Writer) (string, error) {
```
radTuti, line 72: why is this not using the log level?

radTuti (changes requested):
with how CI has been running out of space lately, I think we should be mindful of having more files

caseydavenport:
@radTuti thanks! Resolved

radTuti (approved):

PR #12632, opened 2026-04-27
[EV-6581] Cross-compile ztunnel for arm64 instead of QEMU emulation

radixo:
**MERGE ON HOLD** depending on new Rust Build from: https://github.com/projectcalico/toolchain/pull/817

Switch the ztunnel build from running the Rust toolchain inside a QEMU-emulated arm64 container to native cross-compilation on amd64 using cargo's --target. This mirrors what was done for felix/node CGO builds in 618537a8ee, avoids rustc segfaulting under qemu-user, and collapses the dedicated arm64 CI runner for ztunnel into the regular amd64 cross-build job.

lib.Makefile:
- Add RUST_TARGET_<arch> map (amd64 -> x86_64-unknown-linux-gnu, arm64 -> aarch64-unknown-linux-gnu).
- DOCKER_RUST_BUILD: drop --platform=linux/$(ARCH) (no more QEMU), pin --platform linux/amd64, inject CARGO_BUILD_TARGET=$(RUST_TARGET) so cargo cross-compiles transparently — analogous to how DOCKER_GO_BUILD injects GOARCH for Go.

istio/Makefile:
- Drop the | register order-only prereq on the ztunnel target (binfmt registration no longer needed).
- Pick the binary out of out/rust/$(RUST_TARGET)/release/, since cargo writes to a target-prefixed subdir when CARGO_BUILD_TARGET is set.

.semaphore/:
- Merge the dedicated "Istio Ztunnel - native arm64 runner" block back into the standard amd64-runner Istio job; it now runs `make build ARCH=arm64` (full istio + ztunnel) instead of splitting build-istio from build-ztunnel.
- Publish istio images: combine the amd64 and native-arm64 publish jobs into a single amd64 job that pushes both arches, since cross-compilation removes the need for an arm64 runner.

**Release note:**
```release-note
None
```


.semaphore/push-images/istio.yml
```diff
@@ -29,26 +29,15 @@ blocks:
     skip:
       when: "branch !~ '.+'"
     task:
+      # ztunnel cross-compiles for arm64 via calico/rust-build's aarch64
+      # toolchain, so both arches publish from the same amd64 runner.
       jobs:
-        - name: Linux amd64
+        - name: Linux amd64 + arm64
```
hjiawei: Use "Linux multi-arch" to be consistent with other job names (like whisker which is amd64 and arm64 only)?

.semaphore/push-images/istio.yml
```diff
@@ -29,26 +29,15 @@ blocks:
...
-    dependencies: []
-    task:
-      agent:
-        machine:
-          type: s1-aws-arm64-2
-      jobs:
-        - name: Linux arm64
-          commands:
-            - if [ -z "${SEMAPHORE_GIT_PR_NUMBER}" ]; then make -C istio cd ARCHES=arm64 CONFIRM=true; fi
+            - if [ -z "${SEMAPHORE_GIT_PR_NUMBER}" ]; then make -C istio cd ARCHES="amd64 arm64" CONFIRM=true; fi
```
hjiawei: Shall we use `... then make -C istio cd push-manifests-with-tag CONFIRM=true; fi`?

.semaphore/push-images/istio.yml
hjiawei: This block can be merged into the first one because we don't need to wait for the native arm64 running to complete.

hjiawei: Change to "Publish istio images and multi-arch manifests"?

.semaphore/semaphore.yml.d/blocks/20-istio.yml
```diff
@@ -10,31 +10,14 @@
...
           - ../.semaphore/run-and-monitor ci.log make ci
       - name: "Istio: build arm64"
         matrix:
           - env_var: ARCH
             values:
               - arm64
         commands:
-          # Skip istio ztunnel in this arm64 cross-build job.
-          # ztunnel is built separately in the native arm64 runner block below.
-          - ../.semaphore/run-and-monitor build-$ARCH.log make build-istio ARCH=$ARCH
```
hjiawei, line 30: The `build-istio` target can be removed from the istio/Makefile. We can simplify it after the cross-compilation changes.

lib.Makefile
```diff
@@ -132,6 +132,13 @@ CALICO_BUILD    = $(GO_BUILD_IMAGE):$(GO_BUILD_VER)
 RUST_BUILD_IMAGE ?= calico/rust-build
 CALICO_RUST_BUILD = $(RUST_BUILD_IMAGE):$(RUST_BUILD_VER)
 
+# Map Calico's ARCH to Rust target triple. CARGO_BUILD_TARGET is then injected
+# into DOCKER_RUST_BUILD so cargo cross-compiles transparently — analogous to
+# how DOCKER_GO_BUILD injects GOARCH for Go.
+RUST_TARGET_amd64 = x86_64-unknown-linux-gnu
+RUST_TARGET_arm64 = aarch64-unknown-linux-gnu
+RUST_TARGET = $(RUST_TARGET_$(ARCH))
+
```
hjiawei: Can we move this to the cross-compilation section below, around line 151? Also, could we make the variable naming consistent with the Clang setup and merge the variable defined in https://github.com/projectcalico/toolchain/pull/817/changes#diff-8b72bfc3f9b20e8709d9a03a1425562b09f0d3bef3063f95697656b95c6ea79aR163?

.semaphore/semaphore.yml.d/blocks/20-istio.yml
radixo: I will remove build-istio and build-ztunnel so, and leave only build, WDYT?

lib.Makefile
```diff
@@ -382,12 +389,31 @@ DOCKER_RUN := $(DOCKER_RUN_PRIV_NET) --net=host
...
+	-e BINDGEN_EXTRA_CLANG_ARGS_aarch64_unknown_linux_gnu="--target=aarch64-linux-gnu --sysroot=/usr/aarch64-linux-gnu/sys-root -I/usr/aarch64-linux-gnu/sys-root/usr/include"
+endif
+
 DOCKER_RUST_BUILD := mkdir -p bin && \
 	docker run --rm \
 		--init \
-		--platform=linux/$(ARCH) \
 		--user $(LOCAL_USER_ID):$(LOCAL_GROUP_ID) \
 		$(EXTRA_DOCKER_ARGS) \
+		--platform linux/amd64 \
```
hjiawei: Drop this. Hardcoding amd64 forces arm64 hosts (Apple Silicon dev, future native arm64 CI) to run the amd64 image under QEMU — the exact regression this PR is trying to eliminate.

Let Docker pick the host's native arch and the rest just works:

- amd64 host → amd64 image (has the aarch64 sysroot) → cross-compiles via `--target`
- arm64 host → arm64 image → builds natively

lib.Makefile
```diff
@@ -382,12 +389,31 @@ DOCKER_RUN := $(DOCKER_RUN_PRIV_NET) --net=host
 
 DOCKER_GO_BUILD := $(DOCKER_RUN) $(CALICO_BUILD)
 
+# Cross-compile env for Rust + cc-rs / bindgen when building for arm64.
+# The CARGO_TARGET_* vars replace cargo's [target.aarch64-unknown-linux-gnu]
+# config table (linker + rustflags); the *_aarch64_unknown_linux_gnu vars
+# cover the C-compilation side that cc-rs / bindgen read at build time.
+# All arch-specific config lives here so the rust-build image stays neutral.
+ifeq ($(ARCH),arm64)
```
hjiawei: Gate this on actually cross-compiling, not just `ARCH=arm64`. On an arm64 host this currently points cargo at `/usr/aarch64-linux-gnu/sys-root`, which only exists in the amd64 variant of the rust-build image.

Mirror the Go side at L167–L176:

```makefile
ifeq ($(BUILDARCH),amd64)
ifneq ($(ARCH),amd64)
RUST_CROSS_ENV := ...
endif
endif
```

`CARGO_BUILD_TARGET` itself is harmless on amd64-native (just adds a target-prefixed output dir, which `istio/Makefile` already handles), so that one can stay unconditional. Only the cross env block needs the host≠target gate.

While you're here: once #3 (the AR env value) lets us drop the symlinks in 817, the hardcoded `/usr/aarch64-linux-gnu/sys-root` in this block can become `$(CROSS_SYSROOT)` and the triple `$(CROSS_TRIPLE)` — both already defined at L170–L171 for the Go side. Keeps Rust and Go in lock-step and makes adding ppc64le trivial later. (This is what the earlier inline suggestion about merging variable definitions was getting at.)

lib.Makefile
```diff
@@ -382,12 +389,31 @@ DOCKER_RUN := $(DOCKER_RUN_PRIV_NET) --net=host
...
+RUST_CROSS_ENV := \
+	-e CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=clang \
+	-e CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUSTFLAGS="-C link-arg=--target=aarch64-linux-gnu -C link-arg=--sysroot=/usr/aarch64-linux-gnu/sys-root -C link-arg=-fuse-ld=lld" \
+	-e CC_aarch64_unknown_linux_gnu=clang \
+	-e CXX_aarch64_unknown_linux_gnu=clang++ \
+	-e AR_aarch64_unknown_linux_gnu=aarch64-unknown-linux-gnu-ar \
```
hjiawei: Use the EPEL-style name:

```diff
--e AR_aarch64_unknown_linux_gnu=aarch64-unknown-linux-gnu-ar
+-e AR_aarch64_unknown_linux_gnu=aarch64-linux-gnu-ar
```

The env var **key** has to match the Rust target triple (cc-rs convention), but the **value** is just the executable name — cargo/cc-rs exec whatever you point at. The RUSTFLAGS above already use the short GNU triple (`--target=aarch64-linux-gnu`, `--sysroot=/usr/aarch64-linux-gnu/sys-root`); making AR consistent lets us delete the entire binutils symlink block in toolchain#817.

Verification step worth doing after the symlinks go: `make build-ztunnel ARCH=arm64` and grep the build log for any `aarch64-unknown-linux-gnu-<tool>: not found`. If anything pops up, add a single targeted env var or symlink rather than carrying all nine defensively.

hjiawei (commented):
Three changes — mostly around making this work on non-amd64 hosts and enabling the symlink block in toolchain#817 to be dropped. See inline.

lib.Makefile
hjiawei: @radixo Have you moved these Rust target triples into the cross-compilation section?

lib.Makefile
radixo: - grep "aarch64-unknown-linux-gnu-<tool>: not found" → **zero hits**
- Total occurrences of aarch64-unknown-linux-gnu- anywhere as a binutils prefix (excluding the legitimate uses in the Rust target triple, CARGO_TARGET env vars, and the output dir path target/aarch64-unknown-linux-gnu/release/) → **zero**

hjiawei (approved):

PR #12635, opened 2026-04-27
OpenStack: delete 'mech_driver' global variable

nelljerram:
We only needed this for a monkey-patching hack that was removed in https://github.com/projectcalico/calico/pull/10926


aaaaaaaalex (approved):
I can't speak to the impact of doing this but

PR #12643, opened 2026-04-28
EV-6583 Add calico-controller-gen wrapper with custom CRD markers

radixo:
## What

Replace the per-component
`0001-Add-nullable-to-IPAM-block-allocations-field.patch` files with
`hack/cmd/calico-crd-transform`, a small Go tool that patches
controller-gen-produced CRDs from rules in `transforms.yaml`.

Vanilla `controller-gen` runs first; the post-processor then applies:

- `numorstring.{Port,Protocol}` → `x-kubernetes-int-or-string` union
  (the struct walk would otherwise emit a
  `{minPort,maxPort,portName}`-style object schema).
- `IPAMBlock.allocations[*]` → `nullable: true` (controller-gen does
  not auto-detect that `[]*T` elements should be nullable).

## How

Rules use a single action — `apply.patch` — with JSON-Merge-Patch
semantics: a value sets a key, an explicit `null` deletes it, omitted
keys are preserved (so field-level metadata like `description` and
`default` survive a shape swap).

Rules with **zero matches across the input directory cause a non-zero
exit**, so config drift (renamed field, upstream behavior change)
fails loud instead of silently no-oping.

## Test plan

- [x] `go test ./hack/cmd/calico-crd-transform/` — patch set/delete,
  shape vs path matchers, zero-match loud fail, config validation.
- [x] `make -B -C api gen-files` and `make -B -C libcalico-go gen-crds`
  regenerate CRDs with the expected int-or-string + nullable
  transforms applied.
- [x] `make -B -C felix docs/config-params.{json,md}` regenerates
  byte-identically — descriptions for `*numorstring.Port` config knobs
  (NATPortRange, BPFPSNATPorts) are preserved.

## Follow-ups

- Drop the patched `controller-gen` binary from `projectcalico/toolchain`'s
  `calico-go-build` image once this lands (no longer used).

**Release note:**
```release-note
None
```


hack/cmd/calico-crd-transform/main.go
```diff
@@ -0,0 +1,277 @@
+// calico-crd-transform applies the rules in transforms.yaml to a directory
```
caseydavenport: We need copyright headers on all of our Go files.

kube-controllers/Makefile
```diff
@@ -47,7 +47,8 @@ check-copyright:
 ###############################################################################
 .PHONY: gen-files
 gen-files:
-	$(DOCKER_RUN) $(CALICO_BUILD) sh -c 'controller-gen crd paths=./pkg/controllers/migration/... output:crd:dir=pkg/controllers/migration/crd/'
+	$(MAKE) -C $(REPO_ROOT) build-calico-controller-gen
+	$(DOCKER_RUN) $(CALICO_BUILD) sh -c '$(CALICO_CONTROLLER_GEN) crd paths=./pkg/controllers/migration/... output:crd:dir=pkg/controllers/migration/crd/'
```
caseydavenport, line 51: I think this means we can remove the controller-gen tools from go-build after this merges? Or maybe we still need them in the operator repo?

lib.Makefile
```diff
@@ -435,6 +435,38 @@ RUST_CROSS_ENV := \
...
+# *.patch files under //hack/cmd/calico-controller-gen): NumOrString/Port/
+# Protocol/DSCP int-or-string union schemas, and nullable slice-of-pointer
+# elements. The projectcalico/toolchain repo bakes a (NumOrString-only) patched
+# binary into the calico/go-build image; we build our own patched binary
+# in-repo instead (download tarball -> apply patches -> go build), so CRD
+# generation owns its patches and does not depend on the image's controller-gen.
+#
+# CONTROLLER_TOOLS_VERSION MUST match the toolchain go-build image
+# (projectcalico/toolchain images/calico-go-build/Dockerfile).
+CONTROLLER_TOOLS_VERSION ?= v0.18.0
```
caseydavenport: Is there a way to get this from the go-build version, if we're keeping it there? Would be nice to avoid needing to keep two places in sync.

lib.Makefile
```diff
@@ -435,6 +435,38 @@ RUST_CROSS_ENV := \
...
+# the version OR editing/adding a patch produces a new path and triggers a
+# rebuild.
+CALICO_CONTROLLER_GEN_HASH := $(shell cat $(REPO_ROOT)/hack/cmd/calico-controller-gen/*.patch 2>/dev/null | sha256sum | cut -c1-12)
+CALICO_CONTROLLER_GEN := /go-cache/bin/calico-controller-gen-$(CONTROLLER_TOOLS_VERSION)-$(CALICO_CONTROLLER_GEN_HASH)
+
+# build-calico-controller-gen builds the patched controller-gen into the shared
+# cache. It must run with the repo root mounted (so build.sh and the patch are
+# visible); components in their own module (api/) reach it via
+# $(MAKE) -C $(REPO_ROOT) build-calico-controller-gen.
+.PHONY: build-calico-controller-gen
```
caseydavenport: would be nice if we could avoid a phony target here - if it's already build and none of its deps have changed, we should be able to just use the existing tool to save time.

kube-controllers/Makefile
radixo: Re-checking!

radixo: calico no longer uses the image's controller-gen, but operator still does, so let's keep it in go-build for now; dropping it is a follow-up coordinated with the operator repo.

lib.Makefile
radixo: Yeah, it is possible to get it from go-build version, but alternatively we can pin it on metadata.mk next to GOBUILD_VER, WDYT?

hack/cmd/calico-crd-transform/main.go
radixo: This one was removed

lib.Makefile
caseydavenport: I think it would be good to get it from go-build if it's not too complicated - that way we only have one thing to keep in sync (Or we could learn it from go.mod perhaps?)

Once it's removed from go-build (after the operator changes) then I think we can split it out to its own variable (or learn it from go.mod, or both)

caseydavenport (approved):

PR #12644, opened 2026-04-28
felix/.semaphore: capture stderr in batch test.log

tomastigera:
Merge stderr into `test.log` (`2>&1`) inside the `nohup` wrapper that runs the remote test command, so anything emitted by `make`/`docker`/cleanup after the test binary exits (e.g. a hang in `docker run --rm` tearing down a privileged container) lands in the uploaded artifact.

Previously the outer `nohup ... >& /dev/null` swallowed stderr because the inner `> test.log` only redirected stdout, leaving the orchestrator with no diagnostic clue when `test.rc` never appeared and the watchdog killed the batch.

**Release note:**
```release-note
TBD
```


lucastigera (approved):
LGTM

PR #12646, opened 2026-04-29
Drop calico/go-build wrapper around e2e test run

caseydavenport:
The e2e test phase script wraps `make e2e-run` in `calico/go-build` (~2GB pull every job) on both the local-build and hashrelease paths. The build toolchain is overkill for the run step — `make e2e-run` only needs Go to drive ginkgo against an already-existing `e2e.test` binary. Pick the runtime image based on the path:

- **Local-build path**: keep `calico/go-build`. The Makefile just pulled it to compile the binary (`e2e/Makefile:30-43`), so reusing it for the run step is free.
- **Hashrelease path**: switch to `golang:${GO_VERSION}-bookworm`. The path didn't compile anything, so there's no reason to drag in the LLVM/BPF toolchain. The official Go image is debian-bookworm-based, glibc-compatible with the binary, and runs ~800MB instead of ~2GB. `GO_VERSION` is parsed from `GO_BUILD_VER` in `metadata.mk` so the pin tracks the rest of the toolchain.

Also widens the change-detection filter on the GCP e2e block to fire on `.semaphore/end-to-end/` changes, so this PR (and future runner-script changes) actually exercises the GCP e2e job.

The binary is CGO=1 with libbpf (likely static)/libelf/libz/libc dynamically linked. `golang:bookworm` carries libc6/libelf1/zlib1g, so dynamic linkage and glibc symbol versions should resolve cleanly. CI will tell us if not.

```release-note
None
```

### DIFF
--- a/.semaphore/semaphore.yml.d/blocks/20-e2e-gcp.yml
+++ b/.semaphore/semaphore.yml.d/blocks/20-e2e-gcp.yml
@@ -1,6 +1,6 @@
 - name: "E2E tests on GCP kubeadm (local binary)"
   run:
-    when: "${FORCE_RUN} or change_in(['/e2e/**', '/.semaphore/end-to-end/'], {pipeline_file: 'ignore', exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
+    when: "${FORCE_RUN} or change_in(['/e2e/**', '/.semaphore/end-to-end/scripts/'], {pipeline_file: 'ignore', exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
   dependencies:
     - Prerequisites
   task:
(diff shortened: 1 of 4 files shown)


.semaphore/semaphore.yml.d/blocks/20-e2e-gcp.yml
```diff
@@ -1,6 +1,6 @@
 - name: "E2E tests on GCP kubeadm (local binary)"
   run:
-    when: "${FORCE_RUN} or change_in(['/e2e/**'], {pipeline_file: 'ignore', exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
+    when: "${FORCE_RUN} or change_in(['/e2e/**', '/.semaphore/end-to-end/'], {pipeline_file: 'ignore', exclude: ['/**/.gitignore', '/**/README.md', '/**/LICENSE']})"
```
lwr20: IIUC, this is so we get runs when the scripts change?  Can we make it `/.semaphore/end-to-end/scripts/` instead?  I'd prefer not to have this run when I update a pipeline, for example.

lwr20 (approved):

PR #12656, opened 2026-04-30
[BPF] Fix node-unreachable on Felix restart with no Node (CORE-12744 / #12642)

### LINKED ISSUE #12642: BPF dataplane: kubectl delete node without graceful Calico shutdown causes node to lose connectivity

## Description

## Steps to Reproduce

## Expected Behavior

## Current Behavior

## Context

## Environment

tomastigera:
## Symptom

In BPF mode, after `kubectl delete node <self>` followed by a Felix restart, the node loses all network connectivity (no SSH, no etcd/apiserver) and cannot recover on its own.

## Mechanism (verified end-to-end via FV)

1. New Felix runs `repinJumpMaps()` to move the previous Felix's `cali_progs_*` and `cali_jump_*` maps under `old_jumps/<tmp>/` and creates fresh empty maps at the original pin paths. The previous Felix's preambles (still attached to host interfaces) reference the OLD maps' kernel IDs.
2. Without a Node, `attachDataIfaceProgram` early-returns on `hostIP == nil`; new preambles are never installed and the new (empty) jump maps are never populated.
3. The apply loop still discards each interface from `dirtyIfaceNames` because `doApplyPolicy*` clears the missing-host-IP error at the bottom (the dual-stack "don't fail v4 if only v6 is enabled" trick swallows it).
4. With `dirtyIfaceNames.Len() == 0`, the cleanup gate at `bpf_ep_mgr.go:1966` fires `os.RemoveAll(old_jumps)`. The OLD maps lose their pin; program references from BPF programs alone do not keep `prog_array` entries alive, so the kernel reclaims the slots.
5. The OLD preambles are still attached, but every `bpf_tail_call` into the now-empty maps fails. Every packet through `eth0` is dropped.

## Fix

Tighten the cleanup gate. `os.RemoveAll(old_jumps)` only runs when both `dirtyIfaceNames` is empty AND every enabled IP family has its host IP known. While the host IP is unknown, the OLD jump maps stay pinned, the previous Felix's preambles keep working, and the node remains reachable. When the Node returns, `updateHostIP` marks all interfaces dirty, the next apply pass attaches new preambles via `link.Update` (atomic per-interface), and only then does the cleanup fire.

Cascading restarts are safe: each new Felix coming up without a Node simply nests another `old_jumps/<tmp>/` generation; the load-bearing first one is preserved by the gate; the eventual successful apply pass clears the entire `old_jumps/` tree at once.

Pure-IPv4, pure-IPv6, and dual-stack are all handled by the `m.vX != nil &&` guards in `allHostIPsKnown`.

## Test

`felix/fv/bpf_node_delete_test.go` adds two scenarios:

- **Node delete only**: confirms host SSH stays reachable when only the Node is removed (pre-existing behaviour, baseline).
- **Node delete + Felix restart + recovery**: the actual bug repro. Captures the Node spec, deletes the Node, restarts Felix without waiting for `/readiness` (Felix can't reach ready without a Node — that's a separate symptom), verifies SSH is still reachable, verifies the OLD jump maps are still pinned under `old_jumps/`, re-adds the Node, waits for `old_jumps/` to be cleaned up by Felix's recovery pass, and verifies SSH is still reachable post-recovery.

Both tests pass.

**Release note:**
``` release-note
Fix BPF dataplane node becoming unreachable after `kubectl delete node` followed by a Felix restart.
```

Fixes [#12642](https://github.com/projectcalico/calico/issues/12642)

### DIFF
--- a/felix/fv/bpf_node_delete_test.go
+++ b/felix/fv/bpf_node_delete_test.go
@@ -1,0 +1,251 @@
+// Copyright (c) 2026 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fv_test
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	api "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+
+	. "github.com/projectcalico/calico/felix/fv/connectivity"
+	"github.com/projectcalico/calico/felix/fv/containers"
+	"github.com/projectcalico/calico/felix/fv/infrastructure"
+	"github.com/projectcalico/calico/felix/fv/workload"
+	"github.com/projectcalico/calico/libcalico-go/lib/apiconfig"
+	client "github.com/projectcalico/calico/libcalico-go/lib/clientv3"
+	"github.com/projectcalico/calico/libcalico-go/lib/options"
+)
+
+// Repro for https://github.com/projectcalico/calico/issues/12642 — when the
+// Calico Node object is deleted while Felix is running in BPF mode, the host
+// must remain reachable on its failsafe ports.  In the broken state, Felix
+// reacts to the loss of host metadata by reprogramming the BPF dataplane in a
+// way that drops host management traffic, and the node becomes unreachable.
+var _ = infrastructure.DatastoreDescribe(
+	"_BPF-SAFE_ Felix node delete should preserve host connectivity",
+	[]apiconfig.DatastoreType{apiconfig.EtcdV3},
+	func(getInfra infrastructure.InfraFactory) {
+		if !infrastructure.BPFMode() {
+			// The bug is BPF-specific.
+			return
+		}
+
+		var (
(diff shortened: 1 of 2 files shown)


felix/fv/bpf_node_delete_test.go
```diff
@@ -0,0 +1,251 @@
...
+				context.Background(),
+				tc.Felixes[0].Hostname,
+				options.DeleteOptions{},
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Give Felix a moment to react to the deletion event.  We're
+			// looking for a behavioural change, not a transient flap, so a
+			// short fixed wait keeps the test deterministic.
+			time.Sleep(5 * time.Second)
```
fasaxc, line 150: Would be good to find a way to do a proper wait.  If all else fails, could use `Consistently()` instead of a sleep + Expect.  Then it fails fast if the expectation is not met.

felix/fv/bpf_node_delete_test.go
```diff
@@ -0,0 +1,251 @@
...
+			cc.CheckConnectivity()
+
+			By("Verifying the OLD jump maps are still pinned under old_jumps/")
+			// repinJumpMaps() moved the previous Felix's maps here; the
+			// cleanup gate must NOT have fired (host IP is still unknown),
+			// otherwise the old preambles' tail-call targets would be gone.
+			out, _ := felix.ExecOutput("sh", "-c",
+				"ls /sys/fs/bpf/tc/globals/old_jumps/*/cali_progs_ing2 2>&1 | head -1")
+			Expect(out).To(ContainSubstring("/sys/fs/bpf/tc/globals/old_jumps/"),
+				"expected at least one old jump map pin under old_jumps/, got: %q", out)
```
fasaxc, line 221: Could you avoid the specific sleep with `Eventually/Consistently(<bpf dir>).should(Contain(<old maps>))`

fasaxc (commented):
Not keen on sleeps in the tests but this is a very awkward thing to test; made some suggestions

fasaxc (changes requested):

tomastigera:
Closing in favor of https://github.com/projectcalico/calico/pull/12810

PR #12582, opened 2026-04-21
[PoC] networking-calico: Use a separate process & thread for resync

zhanz1:
## WHY
When running `networking-calico` on a large OpenStack cluster, the amount of workload endpoint updates, agent health updates, and entries to walk through during periodic resyncs could be quite huge. Currently, all of the status updates and the periodic jobs run in one single process, which is not scalable. Therefore, at the first stage, we should separate the work into different processes.

## WHAT
A PoC on moving the periodic resync job into its own process.

## HOW
Launch `neutron-server` with `calico` as the plugin and set up an OpenStack cluster.

## TEST

Tested on a OpenStack cluster, the periodic resync is able to run in its own process:

```sh
# neutron-server.log
2026-04-21 21:00:46.847 3571950 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:1:RESYNC req-596fe484-b680-453b-9030-7b85a96c93b7 - - - - - -] The periodic resync finished after 0:01:18.736173
2026-04-21 21:03:05.459 3571950 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:1:RESYNC req-596fe484-b680-453b-9030-7b85a96c93b7 - - - - - -] The periodic resync finished after 0:01:18.611338
2026-04-21 21:05:22.461 3571950 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:1:RESYNC req-596fe484-b680-453b-9030-7b85a96c93b7 - - - - - -] The periodic resync finished after 0:01:17.000493

# processes - Note the PID=3571950 is a standalone worker process dedicated for syncing
    PID USER      PRI  NI△ VIRT   RES   SHR S CPU% MEM%   TIME+  Command
3571926 neutron    20   0  235M  140M 12928 S  0.0  1.8  0:07.52 └─ /usr/bin/python3 /usr/bin/neutron-server --config-file=...
3571947 neutron    20   0  181M  153M  7808 S  0.0  1.9  0:18.09    ├─ neutron-server: api worker
3571948 neutron    20   0  162M  132M  5376 S  0.0  1.7  0:03.68    ├─ neutron-server: RpcWorker
3571949 neutron    20   0  160M  129M  5248 S  0.0  1.6  0:01.76    ├─ neutron-server: RpcReportsWorker
3571950 neutron    20   0  266M  236M  7808 R 87.7  3.0 10:42.66    ├─ wep-syncer: WorkloadEndPointSyncer
[...]

PR #12658, opened 2026-04-30
OpenStack: replace periodic resync with one-shot startup resync and calico-resync CLI

nelljerram:
Remove the periodic 30-second resync thread that ran inside every neutron-server fork.  In its place:

* A one-shot startup resync from a dedicated neutron-worker process, gated by the new [calico] startup_resync option (always by default, never to disable).  Runs resync once and then sits idle for the remaining lifetime of the Neutron server process.

* A new calico-resync console script that drives the same resync out-of-band, with --network, --subnet, --port and --security-group scope flags (all-resource resync when none are given).  Network resync expands to its subnets and ports; subnet resync expands to its ports; port resync can optionally pull in associated SGs via --include-sgs-for-ports.  Defaults to /etc/neutron/neutron.conf for --config-file.  Output is a JSON ResyncResult with per-phase timing and an ok flag.

Both paths share the new networking_calico/resync/ package so the in-process and out-of-process resync are the same code.

resync_interval_secs and resync_max_interval_secs are deprecated (retained as no-op options to avoid breaking neutron.conf).

Notable supporting changes:

* New Scope object describing what to resync (sets of network / subnet / port / SG IDs) and a ResyncResult dataclass for the structured summary.

* The narrow-scope and all-scope code paths are unified through a single get_from_neutron / get_from_etcd interface on each syncer. All paths now produce the same correct/updated/deleted/created counters and per-step timings; the endpoint phase additionally reports lm_created / lm_updated / lm_deleted for LiveMigration activity.

* LiveMigration handling is integrated into the WorkloadEndpoint compare loop (names are prefixed 'wep ' / 'lm ' to disambiguate), rather than running as a separate pass.  This eliminates a window in which the destination WEP could be deleted by the WEP resync and then immediately recreated by a subsequent LM resync.

## Wider context

This is the first step in a series of work on the Calico / OpenStack resync function, namely:
1. Replace periodic resync with startup and on-demand resyncs.  (This PR.)
2. Implement large scale testing to measure resync performance - which we know, anecdotally, to be quite poor.
3. Implement testing to verify that dynamic updates are not blocked by a slowly running resync.
4. Make changes to improve performance.  We already have draft work indicating ~40x speedups available.
5. Other work to modernize our driver code for Neutron DB access and to eliminate associated occasional bugs.

**Release note:**
```release-note
Calico for OpenStack no longer runs periodic resyncs between the Neutron DB and etcd, and the associated config options, `resync_interval_secs` and `resync_max_interval_secs`, have been deprecated.  Periodic resyncs have never been necessary for normal operation are were only originally coded only for imagined defensive reasons.  The Calico driver still resyncs against the Neutron DB on startup, when the Neutron server is started or restarted.  In exceptional circumstances where additional resync is needed, or desired as a check on the state, there is now a `calico-resync` CLI tool, which can perform either a full resync, or a reconciliation only of specified Neutron networks, subnets, ports or security groups.
```


networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -132,23 +119,40 @@
         default=100,
         help="The maximum allowed size of our cache of project names.",
     ),
+    cfg.StrOpt(
+        "startup_resync",
```
zhanz1, line 130: Given this is one-shot, it doesn't really make sense to have a dedicated process for it, yet we still don't want the resync to affect others - so I would like to make this run in the main parent neutron-server process in my PR. Any thoughts? :D

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -1205,116 +1188,30 @@ def _update_port(self, plugin_context, port):
...
-
-    def periodic_resync_thread(self, launch_epoch):
-        """Periodic Neutron DB -> etcd resynchronization logic.
-
-        On a fixed interval, spin over relevant Neutron DB objects and
-        reconcile them with etcd, ensuring that the etcd database and Neutron
-        are in synchronization with each other.
+    @logging_exceptions(LOG)
+    def _run_startup_resync(self):
+        """Drive a one-shot start-of-day resync from this fork.
```
zhanz1: I'm a little bit confused by the "start-of-day" here: I thought this is just for the one-time resync that happens when we start the service, but not necessarily start of day?

zhanz1 (commented):
Have some minor comments. It would be nice that if networking-calico debian package comes with a systemd timer that runs the resync everyday, so we kinda build this feature into the package, instead of operators needing to configure manually (well, if they want different intervals, then they need to change it, but it should be just one line :)).

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: @zhanz1 Do you mean using a neutron-worker for the one-shot resync?  If so, I have reached the same conclusion.  When we spoke on Friday we said we would land this first and then your PRs, but in fact I think it will be better for me to include the logic of #12582 here, so that this resync runs in a neutron-worker.  Please do let me know if that is not what you meant!

Moving to the neutron-worker architecture is cleaner in a few ways:
1. It guarantees that the resync can only run once per Neutron server, subject to the "startup_resync" setting - instead of relying on leader election amongst the non-API workers, or not using leader election and running several times, depending on how many non-API workers exist.
2. It guarantees that the resync can run even if there is a configuration to set RPC workers to zero.
3. It's actually clearer and more predictable to rely only on the "startup_resync" setting and **not** on leader election.  When someone runs with multiple Neutron servers - as I believe you do - it means they can configure `startup_resync = always` on one control node, and `startup_resync = never` on the others, and then they will know where the resync can run, and which Neutron server to restart if they want to use that mechanism to do a fresh resync.  (Although they can also use the `calico-resync` CLI.)
4. We can use the callbacks mechanism and post_fork_initialize to kick off the resync, for the resync worker, and to do the _post_fork_init logic for all the other workers.  This means that we will no longer need STARTUP_DELAY_SECS or the `requires_state` decorator.

I hope that all aligns with your thinking.  I will push some extra commits shortly to make these changes.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: I have been using "start-of-day" to mean "one shot" after the Neutron starts or restarts.  But I'll change this to "one shot" in order to be more clear.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
zhanz1: I actually mean moving to the main non-API worker process, not a single neutron-worker process. From my understanding, the neutron-worker process will be restarted if it exits (at least due to some errors I was checking). Similarly, after the one-shot resync, the process will also exit, so we need to make sure it's not restarted.

nelljerram: In this PR I've achieved that with the use of a threading.Event here: https://github.com/projectcalico/calico/pull/12658/changes#diff-7265b1489d3e7ea1e08035c0a582df1e8e257ee95c178a41f56cd5ccd588479eR84-R88.  That works, without consuming any CPU, and I think it's cleaner for resync to have its own process.

zhanz1: Ah thanks, didn't read things carefully. One downside of this is that it might be more confusing to have a process but not really doing anything after the one-shot resync is done. Yet, it is not a good practice to create a separate process ourselves using multiprocessing... So I think it's fine to leave it as-is (technically, the main `neutron-server` process is also doing nothing after it spawns the child processes).

Maybe this can be something to be implemented upstream (e.g., running a one-shot job on a separate process) in either oslo.service or neutron-lib.

networking-calico/networking_calico/resync/scope.py
```diff
@@ -0,0 +1,411 @@
...
+    item counts and per-step timings), it is merged into the
+    summary so callers see both the wall-clock total and the
+    syncer-internal breakdown.
+    """
+    t0 = time.monotonic()
+    detail = fn()
+    summary = {"total_ms": int((time.monotonic() - t0) * 1000), "error": None}
+    if isinstance(detail, dict):
+        summary.update(detail)
+    return summary
```
coutinhop, line 283: Claude seems to think that `error` will always be `None` here, and when there is actually an error (exception) it will be thrown up to the caller and thus `summary` will never return with an actual error.

Wouldn't the following be better? (with some appropriate changes in `Scope.run()` in L124-181 above)

```suggestion
      t0 = time.monotonic()
      summary: Dict[str, Any] = {"error": None}
      try:
          detail = fn()
          if isinstance(detail, dict):
              summary.update(detail)
      except Exception as exc:
          LOG.exception("Resync phase failed")
          summary["error"] = "%s: %s" % (type(exc).__name__, exc)
      summary["total_ms"] = int((time.monotonic() - t0) * 1000)
      return summary
```

I think my main point here is we'd try to run all phases and return all errors instead of breaking at the first error, in addition to having it presented somewhat more clearly (inside the `error` value). But am I missing something here?

networking-calico/networking_calico/plugins/ml2/drivers/calico/policy.py
```diff
@@ -45,18 +46,44 @@ def __init__(self, db, txn_from_context):
         self.region_string = calico_config.get_region_string()
         self.namespace = datamodel_v3.get_namespace(self.region_string)
 
+    def get_from_neutron(self, context, scope):
+        if scope.all():
+            sgs = self.db.get_security_groups(context)
+        else:
+            sgs = self.db.get_security_groups(
+                context, filters={"id": list(scope.ids())}
```
coutinhop, line 54: This sounded plausible to me:

When `calico-resync --port <pid>` runs without `--include-sgs-for-ports`, `Scope.expand()` leaves `all_sg_ids = set()`. `policy_scope()` returns a `ResourceScope` with `limited_ids=set()`. The
`PolicySyncer` then runs:

```
sgs = self.db.get_security_groups(context, filters={"id": []})
```

SQLAlchemy 1.4+ produces a false predicate for column.in_([]) and returns nothing — which is what we want. Older SQLAlchemy versions warn and return all rows, which would cause the resync
to compare every NetworkPolicy in etcd against every Neutron SG and potentially issue spurious writes/deletes. Same hazard in SubnetSyncer.get_from_neutron (subnets.py:43-47).

Modern Neutron pins SQLAlchemy ≥ 1.4, so this is unlikely to bite in supported configurations — but it costs nothing to short-circuit. Suggest a one-liner in each get_from_neutron:

```
if not scope.all() and not scope.ids():
    return {}
```

(and skip the corresponding etcd reads.) Also avoids a needless DB round-trip per empty phase.

networking-calico/networking_calico/plugins/ml2/drivers/calico/policy.py
```diff
@@ -45,18 +46,44 @@ def __init__(self, db, txn_from_context):
         self.region_string = calico_config.get_region_string()
         self.namespace = datamodel_v3.get_namespace(self.region_string)
 
+    def get_from_neutron(self, context, scope):
+        if scope.all():
```
coutinhop, line 50: As per the comment below:

```suggestion
        if not scope.all() and not scope.ids():
            return {}

        if scope.all():
```

networking-calico/networking_calico/plugins/ml2/drivers/calico/subnets.py
```diff
@@ -32,28 +32,52 @@
...
 
-    def delete_legacy_etcd_data(self):
-        etcdv3.delete_prefix(datamodel_v1.SUBNET_DIR)
-
-    def get_all_from_etcd(self):
-        return etcdv3.get_prefix(datamodel_v2.subnet_dir(self.region_string))
-
-    def get_all_from_neutron(self, context):
+    def get_from_neutron(self, context, scope):
+        if scope.all():
```
coutinhop, line 44: As per the comments on policy.py:

```suggestion
        if not scope.all() and not scope.ids():
            return {}

        if scope.all():
```

networking-calico/networking_calico/plugins/ml2/drivers/calico/syncer.py
```diff
@@ -90,113 +91,169 @@ def __init__(self, db, txn_from_context, resource_kind):
...
-        # Get all resources of this type from etcd - as an array of (name,
-        # data, mod_revision) tuples.
-        etcd_resources = self.get_all_from_etcd()
+          * in Neutron, in etcd, data matches  -> no-op (correct)
+          * in Neutron, in etcd, data differs  -> update
+          * in Neutron, not in etcd            -> create
+          * not in Neutron, in etcd            -> delete
+
+        For the narrow case, deletes are only implemented to the extent that etcd keys
+        can be deterministically inferred from the resource IDs.
```
coutinhop: Another one that made sense to me:

The narrow port path can't synthesise the WEP etcd key from a `port_id` alone: the key is `host-openstack-device_id-port_id`, and you only have `port_id` from `--port`. Concretely:

- `WorkloadEndpointSyncer.get_from_neutron` calls `db.get_ports(filters={"id":[pid]})` (endpoints.py:115). For a deleted port this returns `[]`, so `neutron_map = {}`.
- `get_from_etcd` then iterates for name in `neutron_map`, which never enters the loop, so `etcd_map = {}`.
- The compare loop has nothing to compare; the stale WEP (and any LMs whose key contains `port_id`) is left in etcd.

Result: an operator who runs `calico-resync --port <pid>` to clean up after a Neutron port deletion sees no writes/deletes and a ok: True result. This conflicts with the operator's mental model of "I asked you to resync that port, please fix it."

syncer.py:111-112 already acknowledges this in code comments. Two things to fix:

a. Mention it in the --port help text in cli.py:86-92 ("Narrow scope cannot delete stale etcd entries for ports that no longer exist in Neutron; use a full resync for that.")

b. Optionally, for clean_live_migrations=True, you can infer LM keys from port_id because the LM names are name-without-host indexed and contain the port_id UUID. Worth a follow-up: when scope.ports is non-empty, also scan live_migrations_without_host for any entry whose name-without-host contains a scope port_id, and add to etcd_map even when the WEP is gone

networking-calico/networking_calico/resync/scope.py
```diff
@@ -0,0 +1,411 @@
...
+                lambda: policy_syncer.resync(self.admin_context, self.policy_scope())
+            )
+            phases["endpoints"] = _run_phase(
+                lambda: endpoint_syncer.resync(
+                    self.admin_context, self.endpoint_scope()
+                )
+            )
+            if self.all():
+                phases["felix_config"] = _run_phase(provide_felix_config)
+
```
coutinhop, line 168: If we do go on with the changes in `_run_phase()`, we'd need something like this here:

```suggestion
              failed = [n for n, p in phases.items() if p["error"]]
              if failed:
                  ok = False
                  error = "phases failed: " + ", ".join(failed)
                  
```

networking-calico/networking_calico/resync/scope.py
nelljerram: Good spot that it is pointless to always populate `"error": None` here.  But I think it is correct for the resync to bail overall as soon as any part of it hits an exception.  I'd rather bail and give a clear report of the problem, than try to run later phases that might confuse the situation further.  I've removed the `"error"` setting and improved commenting in commit ad251760ed.

networking-calico/networking_calico/plugins/ml2/drivers/calico/policy.py
nelljerram: Good thought.  Shortcut implemented in commit 99f48e59e3.

networking-calico/networking_calico/plugins/ml2/drivers/calico/policy.py
nelljerram: Ditto.

networking-calico/networking_calico/plugins/ml2/drivers/calico/subnets.py
nelljerram: Ditto.

networking-calico/networking_calico/resync/scope.py
nelljerram: Moot, as I decided not to take that specific change.

networking-calico/networking_calico/plugins/ml2/drivers/calico/syncer.py
nelljerram: Good thoughts.  Agree with the suggested commenting in (a).  As regards (b):

- I think some of detail here is garbled: "when scope.ports is non-empty, also scan live_migrations_without_host for any entry whose name-without-host contains a scope port_id," doesn't make sense because a port normally doesn't have any associated live_migration

- but I think the broader point is "given that there's a --clean-live-migrations flag, why isn't there also --clean-ports or --clean-workload-endpoints?"  Which I would agree with.

I've improved commenting and added `--clean-workload-endpoints` in commits ec086faa02 and fee53408ef.

nelljerram:
@coutinhop I've addressed all your comments above, and also fixed some bugs that were discovered when using `calico-resync` to implement a resync scale test (coming in the next PR...)  Please take another look when you have time.

coutinhop:
> @coutinhop I've addressed all your comments above, and also fixed some bugs that were discovered when using `calico-resync` to implement a resync scale test (coming in the next PR...) Please take another look when you have time.

@nelljerram LGTM, I think copilot might have made some good points, can you look at them? thanks!

networking-calico/networking_calico/plugins/ml2/drivers/calico/syncer.py
```diff
@@ -90,113 +91,177 @@ def __init__(self, db, txn_from_context, resource_kind):
...
+        else:
+            with self.txn_from_context(context, "get-all-" + self.resource_kind):
+                neutron_map = self.get_from_neutron(context, scope)
+        t_neutron_read = time.monotonic()
+
+        # Read etcd state.  When the scope is not "all", this means the etcd data
+        # corresponding to the Neutron resources that we just read, and possibly - when
+        # etcd keys can be directly inferred from scope IDs - the etcd data for any IDs
+        # that are in scope but not found in Neutron.
+        etcd_map = self.get_from_etcd(scope, neutron_map)
```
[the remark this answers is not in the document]
nelljerram: Agreed; fixed in https://github.com/projectcalico/calico/pull/12658/commits/a9a488b303e067fe0b5e5b7a0a419760bf60d8ee

networking-calico/networking_calico/datamodel_v3.py
```diff
@@ -147,6 +147,24 @@ def get(resource_kind, name):
...
+    """Read a single Calico v3 resource from etcdv3.
+
+    Returns either ``(spec, mod_revision)`` or ``((spec, labels, annotations),
+    mod_revision)`` depending on ``with_labels_and_annotations``.  This mirrors the
+    per-tuple shape that :func:`get_all` returns for the matching kind, so callers can
+    use the result interchangeably.  Raises :class:`etcdv3.KeyNotFound` if the key is
+    absent.
+    """
+    value, mod_revision = _get_with_metadata(resource_kind, namespace, name)
+    spec = value["spec"]
```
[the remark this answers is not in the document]
nelljerram: Good catch.

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
```diff
@@ -205,6 +309,16 @@ def neutron_to_etcd_write_data(self, port, context, reread=False):
...
 
+    def neutron_to_live_migration_etcd_write_data(
+        self, port, dest_port, context, reread
+    ):
+        if reread:
+            try:
+                port = self.db.get_port(context, port["id"])
+            except n_exc.PortNotFound:
+                raise ResourceGone()
+        return live_migration_spec(self.namespace, port, dest_port)
```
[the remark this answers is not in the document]
nelljerram: It doesn't matter if dest_port is stale, because the only thing we take from dest_port is its binding:host_id value.  But the coding is misleading as it stands (because dest_port is manufactured, not a real Neutron thing) - so I will update update the coding to use just the dest host_id instead of "dest_port".

networking-calico/networking_calico/datamodel_v3.py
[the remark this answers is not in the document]
nelljerram: Fixed in https://github.com/projectcalico/calico/pull/12658/commits/80b0534c6f1c3973fe38aefc5ed938710756c9a8

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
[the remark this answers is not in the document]
nelljerram: https://github.com/projectcalico/calico/pull/12658/commits/88ee773b277cee50f1600691d6bea4aad2570060

networking-calico/networking_calico/plugins/ml2/drivers/calico/syncer.py
```diff
@@ -90,113 +91,177 @@ def __init__(self, db, txn_from_context, resource_kind):
...
                     )
                     if not self.update_in_etcd(name, write_data, mod_revision):
                         LOG.warning(
                             "failed etcd write for %s %s; presume"
-                            + " data updated by another writer",
+                            " data updated by another writer",
+                            self.resource_kind,
+                            name,
+                        )
+                    n_updated += 1
```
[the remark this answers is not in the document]
nelljerram: Fixed in https://github.com/projectcalico/calico/pull/12658/commits/41e8d5dacd5b99e9213d138a0b95687abd20d09a

networking-calico/networking_calico/plugins/ml2/drivers/calico/endpoints.py
```diff
@@ -90,109 +94,209 @@ def __init__(self, db, txn_from_context, policy_syncer, keystone_client):
...
+            "LiveMigration",
             self.namespace,
-            name,
-            spec,
-            labels=labels,
-            annotations=annotations,
+            remove_prefix(name, "lm "),
+            write_data,
             mod_revision=0,
         )
```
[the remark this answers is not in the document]
nelljerram: Also fixed in https://github.com/projectcalico/calico/pull/12658/commits/41e8d5dacd5b99e9213d138a0b95687abd20d09a

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -330,13 +324,67 @@ def __init__(self):
...
+        We use this hook to subscribe to Neutron's process-AFTER_INIT callback
+        so that we get a chance to run code in each worker process we own (see
+        ``get_workers``) once it has been forked.
+        """
+        super(CalicoMechanismDriver, self).initialize()
+        registry.subscribe(
+            self.post_fork_initialize,
+            resources.PROCESS,
+            events.AFTER_INIT,
+            cancellable=True,
```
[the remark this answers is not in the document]
nelljerram: I believe this comment is wrong.  I've verified against `neutron_lib/callbacks/registry.py` on both `unmaintained/2024.1` (Caracal) and master: `def subscribe(callback, resource, event, priority=..., cancellable=False)`.  The kwarg is in the supported API.

networking-calico/networking_calico/plugins/ml2/drivers/calico/workers.py
```diff
@@ -0,0 +1,95 @@
...
+
+class CalicoStartupResyncWorker(worker.BaseWorker):
+    """Worker that runs the one-shot Neutron-DB-to-etcd resync.
+
+    Spawned (in its own OS process) by neutron-server when ``[calico] startup_resync``
+    is ``always`` (the default).  The mech driver's ``post_fork_initialize`` callback
+    recognises this trigger class, builds the per-process state (DB connection, syncers,
+    Keystone client) and invokes the same ``Scope().run()`` entry point that the
+    ``calico-resync`` CLI uses.  The worker process then idles for the lifetime of
+    neutron-server.
```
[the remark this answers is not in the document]
nelljerram: PR description updated.

networking-calico/networking_calico/plugins/ml2/drivers/calico/syncer.py
```diff
@@ -90,113 +91,216 @@ def __init__(self, db, txn_from_context, resource_kind):
...
+            with self.txn_from_context(context, "get-all-" + self.resource_kind):
+                neutron_map = self.get_from_neutron(context, scope)
+        t_neutron_read = time.monotonic()
+
+        # Read etcd state.  When the scope is not "all", this means the etcd data
+        # corresponding to the Neutron resources that we just read, and possibly - when
+        # etcd keys can be directly inferred from scope IDs - the etcd data for any IDs
+        # that are in scope but not found in Neutron.
+        etcd_map = self.get_from_etcd(scope, neutron_map)
+        t_etcd_read = time.monotonic()
```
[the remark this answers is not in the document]
nelljerram: I've addressed this by reverting to an etcd-first design.  See commit https://github.com/projectcalico/calico/pull/12658/commits/40ebe91d44c9fad8d089b16b839cb5520bac2125

networking-calico/networking_calico/resync/scope.py
```diff
@@ -0,0 +1,426 @@
...
+        # Resync of a network includes its subnets and ports.  Track the subnets we
+        # discover here so the subnet block below can skip them: their ports were
+        # already added via the network_id port query.
+        network_subnet_ids = set()
+        if self.networks:
+            for subnet in self.db.get_subnets(
+                self.admin_context, filters={"network_id": list(self.networks)}
+            ):
+                network_subnet_ids.add(subnet["id"])
```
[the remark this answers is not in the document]
nelljerram: Good point - but I will add in a following PR.

networking-calico/networking_calico/resync/scope.py
```diff
@@ -0,0 +1,426 @@
...
+                for s in self.db.get_subnets(
+                    self.admin_context, filters={"id": list(remaining_subnet_ids)}
+                )
+            }
+            if subnet_network_ids:
+                for port in self.db.get_ports(
+                    self.admin_context, filters={"network_id": list(subnet_network_ids)}
+                ):
+                    for fixed_ip in port.get("fixed_ips", []) or []:
+                        if fixed_ip.get("subnet_id") in remaining_subnet_ids:
```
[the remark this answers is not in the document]
nelljerram: Done in https://github.com/projectcalico/calico/pull/12658/commits/b51721962b35f67b986c635447a29d1d5bfd3c41

networking-calico/networking_calico/resync/scope.py
```diff
@@ -0,0 +1,426 @@
...
+                            self.all_port_ids.add(port["id"])
+                            break
+
+        if self.include_security_groups_for_ports and self.all_port_ids:
+            for port in self.db.get_ports(
+                self.admin_context, filters={"id": list(self.all_port_ids)}
+            ):
+                for sg_id in port.get("security_groups", []) or []:
+                    self.all_sg_ids.add(sg_id)
```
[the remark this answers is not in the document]
nelljerram: Ditto.

networking-calico/networking_calico/resync/cli.py
```diff
@@ -0,0 +1,232 @@
...
+            "harder to parse cleanly."
+        ),
+    )
+    return parser
+
+
+def main(argv=None) -> int:
+    """Entry point for the ``calico-resync`` console script."""
+    args = _build_parser().parse_args(argv)
+
```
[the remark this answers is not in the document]
nelljerram: Tests are coming in a following PR.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -330,13 +324,67 @@ def __init__(self):
...
+          ``CalicoStartupResyncWorker`` anyway.
+
+        * Anything else (RPC / state-report / similar) -> connection state
+          plus the elector and master-only background threads.
+        """
+        trigger_cls = _trigger_class(trigger)
+        if trigger_cls is CalicoStartupResyncWorker:
+            self._do_startup_resync()
+        elif trigger_cls is wsgi.WorkerService:
+            self._post_fork_init(voting=False)
```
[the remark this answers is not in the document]
nelljerram: Will cover this in a following PR at FV/DevStack level.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_plugin_etcd.py
```diff
@@ -443,16 +442,32 @@ def make_context(self):
...
+        If the driver has been initialised (post _post_fork_init) we reuse its syncers
+        so the same primed project cache is in play.  Otherwise we let the runner build
+        fresh syncers against the mocked DB and Keystone.
+        """
+        return resync.Scope(
+            self.db,
+            driver=self.driver if hasattr(self.driver, "endpoint_syncer") else None,
+            admin_context=mech_calico.ctx.get_admin_context(),
+            **scope_kwargs,
+        ).run()
```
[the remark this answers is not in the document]
nelljerram: https://github.com/projectcalico/calico/pull/12658/commits/8e963dbe0a2a4c840e23716f2231af29b8ba4586

networking-calico/networking_calico/resync/cli.py
```diff
@@ -0,0 +1,198 @@
...
+        "--output",
+        default=None,
+        metavar="PATH",
+        help=(
+            "Write the JSON ResyncResult to PATH instead of stdout.  "
+            "Useful for tooling — depending on the layered config, "
+            "oslo.log may emit log lines to stdout, making the result "
+            "harder to parse cleanly."
+        ),
+    )
```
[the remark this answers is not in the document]
nelljerram: Good catch, PR description updated.

networking-calico/networking_calico/plugins/ml2/drivers/calico/syncer.py
```diff
@@ -90,113 +91,196 @@ def __init__(self, db, txn_from_context, resource_kind):
...
-                # Translate the Neutron resource to what we would write into
-                # etcd.  Take a transaction here in case the subclass method
-                # needs more Neutron DB reads.
+
+        n_correct = 0
+        n_updated = 0
+        n_deleted = 0
+        n_created = 0
+
+        for name in set(etcd_map) | set(neutron_map):
```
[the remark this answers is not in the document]
nelljerram: https://github.com/projectcalico/calico/pull/12658/commits/55d9ca46b23b9f5251c996bab4ad411fb713b2f1

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: Yes, if upstream implements a way to specify a one-shot worker process, we can then adapt to take advantage of that.  (In future, I mean.)

coutinhop (approved):

nelljerram:
Docs at https://github.com/tigera/docs/pull/2800

PR #12654, opened 2026-04-29
Bump nftables to 1.1.4 in node and istio CNI install images

caseydavenport:
The nftables binary shipped in UBI 9 / AlmaLinux 9 BaseOS (1.0.x) segfaults when it has to list or validate a ruleset that contains udata-bearing sets created by newer nft (1.1.x+) in the same network namespace. Reports of this in calico/node show up as repeated dataplane resync failures and core dumps that fill node disks. The relevant upstream fix is in nftables 1.1.4.

This builds libnftnl 1.3.0 and nftables 1.1.4 from source in the existing almalinux build stage and overlays them on top of the BaseOS install in both `node/Dockerfile` and `istio/Dockerfile-install-cni`. We keep the BaseOS `nftables` package install in the runtime stage so its transitive runtime deps (jansson, gmp, libmnl, ...) stay satisfied.

The earlier knftables bump in #12472 fixed the list-side path (it scopes list calls to our own table), but transactional writes via `nft -f -` and `nft --check -f -` still walk full kernel state internally and trip the same parser bug — that's what this PR addresses.

Fixes #11750

```release-note
Fixes nft binary segfaults in calico/node and the Istio CNI install image when newer nftables (>= 1.1.x) is in use elsewhere on the host.
```

PR #12660, opened 2026-04-30
Build patched nftables 1.1.1 in node and istio CNI install images

### LINKED ISSUE #11750: nft segfault on debian trixie

caseydavenport:
The nft binary shipped in UBI 9 / AlmaLinux 9 BaseOS (1.0.x) segfaults when it has to list or validate a ruleset that contains udata-bearing sets created by newer nft (1.1.2+) in the same network namespace. Reports of this in calico/node show up as repeated dataplane resync failures and core dumps that fill node disks.

This is the alternative to #12654, which bumped nft to 1.1.4 inside our images. That fixes the read side but exposes a write-side regression: nft >= 1.1.2 stamps udata fields that older nft binaries elsewhere on the host (in sidecars, on the node OS) can't parse, so they crash when they list any ruleset in a shared netns. See [this comment](https://github.com/projectcalico/calico/issues/11750#issuecomment-4348157903) for the original report.

This PR builds patched nftables and libnftnl RPMs from upstream tarballs and installs them in calico/node and the Istio CNI install image. The nftables RPM is 1.1.1 with a backport of upstream commit [be737a1](https://git.netfilter.org/nftables/commit/?id=be737a1986bfee0ddea4bee7863dca0123a2bcbc) ("src: netlink: fix crash when ops doesn't support udata", first released in 1.1.4). 1.1.1 is the last release whose set-creation udata is parseable by older nft, so a patched 1.1.1 is safe in both directions: our reads tolerate newer rulesets, and our writes do not poison the kernel for older readers.

The RPMs (specs in `node/rpm/` and `istio/rpm/`) Provide/Obsolete the BaseOS nftables and libnftnl packages, so the runtime image's rpm db owns /usr/sbin/nft and the libnftables / libnftnl shared libraries directly. nftables's runtime deps (jansson, gmp, libmnl) come in via the spec's Requires: rather than via leaving the BaseOS nftables package half-installed.

Fixes #11750
Supersedes #12654

```release-note
Fixes nft binary segfaults in calico/node and the Istio CNI install image when newer nftables is in use elsewhere on the host.
```


mazdakn (approved):
LGTM

hjiawei:
An alternative shape worth considering: build via `rpmbuild` with a custom spec + our patch as `Patch0:`, instead of `curl | tar | configure | make install`.

Rough sketch — drop `node/rpm/nftables.spec` (and one for libnftnl) in-tree, point `Source0:` at the same upstream tarball, list our backport as `Patch0:`. Builder stage adds `rpm-build` and runs `rpmbuild -ba`. Runtime stage does `dnf install ./RPMS/x86_64/nftables-1.1.1-1.tigera.rpm ./RPMS/x86_64/libnftnl-1.2.8-1.tigera.rpm` (with matching `Obsoletes:`/`Provides:` so it cleanly supersedes the BaseOS package), instead of leaving BaseOS nftables installed and overlaying files on top.

Concrete benefits over the current build-from-source overlay:

- **Real package ownership.** `rpm -qf /usr/sbin/nft` returns our package; `rpm -V nftables` is meaningful; upgrades/uninstalls work normally. Today the overlaid `/usr/sbin/nft` and `libnftables.so*` are dangling files that BaseOS's `nftables` package still claims to own — confusing during incident response.
- **Patch maintenance scales.** Adding the next backport is one more `Patch1:` line in the spec, not another `COPY` + `patch -p1` step. The upstream-style `%autopatch` flow is what every netfilter maintainer already reads.
- **Dependency closure is automatic.** The RPM's `Requires:` pulls in jansson/gmp/libmnl; we no longer need the "leave BaseOS package installed so its transitive deps stay satisfied" hack, which is a subtle invariant the comments have to spell out.
- **Reusable outside the image.** If we ever want this nft on a host (debug pod, support bundle, customer hotfix), an RPM is the natural artifact; a tar of staged files is not.
- **Reproducibility/provenance.** Spec encodes `Version:`/`Release:`/`%changelog` and ties the patch to a release identifier we can cite in CVE/release notes.

Costs are real but small: `rpm-build` in the builder stage, one spec file per project to maintain, and the "do not bump past 1.1.1" intent gets dispersed across `Version:` + `%changelog` rather than a single bold Dockerfile comment (mitigable with a comment at the top of the spec).

For a single targeted patch the overlay is defensible — but if we expect more backports here (or want host-installable artifacts), the rpmbuild path ages better. Happy to put up a follow-up PR converting to that shape if there's appetite.

caseydavenport:
@hjiawei yeah, thanks for that - sounds like a better approach!

hjiawei:
Building on the rpmbuild idea above — CI can build the RPMs once per workflow (or skip entirely on cache hit) and have both `node` and `istio` consume the same artifact, mirroring the existing `Pull: go-build image` / `Build: calico image` pattern.

**Shape:**
- New prerequisite block builds `calico/nftables-rpms:<tag>` from the shared spec + patch.
- Tarball cached in GCS keyed by `${NFTABLES_VER}-${LIBNFTNL_VER}-${PATCH_SHA}` so it survives across workflows; cache hit is a ~5s `gcloud storage` copy.
- `node` and `istio` image-build blocks `dependencies: ["Build: nftables RPMs"]`, load it via the existing `load-cached-images` helper, and the Dockerfiles consume via `FROM ${NFT_RPMS_IMAGE} AS nft-rpms` + `dnf install /rpms/*.rpm`.
- Version pins promoted to `metadata.mk` so the cache key, build-arg, and spec all share one source of truth.

Net: warm workflows shave ~1–2 min off both image builds; cold ones add ~30s for the producer. Reuses the GCS cache infra that's already there. Happy to draft either in this PR or a follow-up.

sshambar:
FYI - I cherry-picked the nft 1.1.1 build into the 3.31.5 branch and calico appears to run well and doesn't cause any problems for older nft builds.

I tried adding a host nft set with nft 1.1.5, and calico continued to work well -- however, I started seeing coredumps in kube-proxy (1.35.4 in nft mode) during it's nftable sync...but that's not calico's issue :)

caseydavenport:
@hjiawei I've adjusted to the approach you suggested in the latest commit - LMK what you think! 

@sshambar excellent news, thank you for the confirmation!

hjiawei:
How long do the non-amd64 emulated builds take now? Release build all 4 supported arches on a single f1-4 semaphore runner. I want to make sure inline rpmbuild (3 emulated arches across node+istio) doesn't put us at risk of timing out during the release process.

The build toolchain (rpmbuild/bisin/flex/automake and *-devel haders) currently lives in the same Dockerfile as runtime image. It leaves a path for any of it to bleed into the shipped image. A separate prebuilt RPM step would keep the toolchain out of the runtime image entirely.

Would it be possible to decoupling the RPM build from the runtime image build?

caseydavenport:
@hjiawei ready for another look!

hjiawei (approved):

PR #12668, opened 2026-04-30
networking-calico: Separate leader jobs into different processes

zhanz1:
## WHY

This is an extension of #12582.

As clusters grow larger, it is hard for a single Python process to do resync, compaction, and status updating at the same time. To address this, let's separate the jobs into multiple processes.

## WHAT

This change will introduce four new worker processes, each in charge of:

* CalicoResourceSyncerWorker: Sync resources from Neutron to etcd.
* CalicoManagerWorker: Do leader election and periodic compaction.
* CalicoAgentStatusWatcherWorker: Watch agent status updates and report them to Neutron.
* CalicoEndpointStatusWatcherWorker: Watch endpoint status updates and report them to Neutron.

## HOW

Launch `neutron-server` with `calico` as the plugin and set up an OpenStack cluster.

## TEST

Tested on a virtual OpenStack cluster with 3 instance of `neutron-server`, all seems to be working:

<details>

```
# Leader election
### Node 1
2026-04-30 21:37:36.216 2219629 INFO networking_calico.plugins.ml2.drivers.calico.election [-] Successfully become master - key /calico/openstack/v2/no-region/neutron_election, value node-1:2219629
### Node 2
2026-04-30 21:37:38.576 2605430 INFO networking_calico.etcdv3 [-] etcdv3 get key=/calico/openstack/v2/no-region/neutron_election results=[(b'node-1:2219629', {'key': b'/calico/openstack/v2/no-region/neutron_election', 'create_revision': '5287496', 'mod_revision': '5287498', 'version': '2', 'lease': '8674386991848183953'})]

# Periodic resync
2026-04-30 21:45:49.593 2605431 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:2:RESYNC req-0fe39c86-ac77-4462-8dbe-1c863e1fc29c - - - - - -] I am master: doing periodic resync
2026-04-30 21:48:09.642 2605431 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:2:RESYNC req-0fe39c86-ac77-4462-8dbe-1c863e1fc29c - - - - - -] I am master: doing periodic resync
2026-04-30 21:50:29.484 2605431 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:2:RESYNC req-0fe39c86-ac77-4462-8dbe-1c863e1fc29c - - - - - -] I am master: doing periodic resync

# Resync monitor
2026-04-30 21:43:29.522 2605431 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [-] I am master: monitoring periodic resync

# Compaction
2026-04-30 21:43:28.665 2605430 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:2:COMPACTION req-51698c4a-54f4-4537-b982-aab35936ee78 - - - - - -] I am master: doing periodic compaction

# Agent status
2026-04-30 21:56:20.202 2605432 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:3:STATUS_ETCD_WATCHER req-03c102c2-fdd6-4cd8-acf6-48ba45829e9f - - - - - -] Felix on host host-10 is alive; fanning out status report

# Port status
2026-04-30 21:58:03.728 2605433 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:7:STATUS_ETCD_WATCHER req-85d1dc7e-2ef7-4e35-b8d5-fdb0c99b1b7c - - - - - -] Status of port ('host-10', 'ba683f56-d996-4ad8-be1a-f5f294cb620d') on host host-10 changed to up

# Live migration
2026-04-30 21:44:27.425 2605433 INFO networking_calico.plugins.ml2.drivers.calico.mech_calico [None CALICO:9:UPDATE_PORT_POSTCOMMIT req-dbed5163-87e4-4b0f-89da-9200fe223980 - - - - - -] Live migration e394c4b2-0043-4009-bea8-db2101fa75e8: destination port 5f69b7d1-9a75-4831-bb0e-2e3a5dbdc038 active on host-10, notifying Nova
```

</details>

Also, updated the unit tests according to the new design.

## MISC

**Release note:**
```release-note
Feature: Split OpenStack driver's leader-only tasks into multiple processes. 
```


nelljerram:
@zhanz1 I pushed a formatting fix to this to satisfy our linters (black and flake8), and to allow CI to run.

zhanz1:
> @zhanz1 I pushed a formatting fix to this to satisfy our linters (black and flake8), and to allow CI to run.

Thanks... I thought I passed `make fmtpy` but it seems something were missed.

nelljerram:
> > @zhanz1 I pushed a formatting fix to this to satisfy our linters (black and flake8), and to allow CI to run.
> 
> Thanks... I thought I passed `make fmtpy` but it seems something were missed.

It's a bit confusing.  `fmtpy` is an action that edits the files that need it but then returns success.  Perhaps we should add a `make check-dirty` after the black line.

nelljerram:
@zhanz1 Please could you rework this PR on top of https://github.com/projectcalico/calico/pull/12658 (given that the latter has now merged)?  Or let me know if you would prefer me to do that.

nelljerram:
> @zhanz1 Please could you rework this PR on top of #12658 (given that the latter has now merged)? Or let me know if you would prefer me to do that.

BTW my plan will be to look again at #12456 after this PR has merged.  I think we will find that a lot of #12456 is no longer needed, but there may still be some useful pieces there.

zhanz1:
> @zhanz1 Please could you rework this PR on top of #12658 (given that the latter has now merged)? Or let me know if you would prefer me to do that.

Yes, I'll work on it, no worries!

> BTW my plan will be to look again at #12456 after this PR has merged. I think we will find that a lot of #12456 is no longer needed, but there may still be some useful pieces there.

Yes, I think the most important part is how we will be detecting dead greenthread after we split work onto multiple processes.

zhanz1:
@nelljerram I've rebased and also modified relevant naming and comments a little bit (e.g., `voting` is no longer a thing), feel free to take a look, thanks!

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -377,137 +383,146 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
           elector and master-only background threads.
         """
         trigger_cls = _trigger_class(trigger)
-        if trigger_cls is CalicoStartupResyncWorker:
-            self._do_startup_resync()
-        elif trigger_cls is wsgi.WorkerService:
-            self._post_fork_init(voting=False)
-        else:
-            self._post_fork_init(voting=True)
+        self._post_fork_inititialize_common()
```
nelljerram:
```suggestion
        self._post_fork_initialize_common()
```

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -377,137 +383,146 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
...
 
-    @logging_exceptions(LOG)
-    def _post_fork_init(self, voting=False):
-        """_post_fork_init
+        worker_mapping = {
+            CalicoManagerWorker: self._init_and_start_calico_manager,
+            CalicoStartupResyncWorker: self._init_and_start_calico_resouce_syncer,
+            CalicoAgentStatusWatcherWorker: self._init_and_start_agent_status_watcher,
+            CalicoEndpointStatusWatcherWorker: (
+                self._init_and_start_endpoint_status_watcher
```
nelljerram:
```suggestion
            CalicoManagerWorker: self._init_start_manager,
            CalicoStartupResyncWorker: self._init_start_resync,
            CalicoAgentStatusWatcherWorker: self._init_start_agent_status_watcher,
            CalicoEndpointStatusWatcherWorker: (
                self._init_start_endpoint_status_watcher
```

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: I would prefer the startup_resync not to have any intialization apart from _get_db(), so that it is as similar as possible to running the `calico-resync` CLI.  Can we keep
```
        if trigger_cls is CalicoStartupResyncWorker:
            self._do_startup_resync()
```
as a special initial case and then keep your approach for the other cases?

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -292,29 +298,19 @@ def __init__(self):
...
-        self._port_status_cache = {}
-        # Queue used to fan out port status updates to worker threads.  Notes:
-        # * we don't recreate the queue in _post_fork_init() so that we can't
-        #   possibly lose updates that had already been queued.
-        # * the queue contains tuples (priority, <status key>); we use a
-        #   higher priority for events and a lower priority for snapshot
-        #   keys, so that current data skips the queue.
-        self._port_status_queue = PriorityQueue()
-        self._port_status_queue_too_long = False
+        # Process-shared variable for checking if the current instance of
```
nelljerram:
```suggestion
        # Variable shared across all processes that are forked for the current Neutron server.
        # Tracks whether or not this Neutron server is the master for its OpenStack region
```

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -377,137 +383,146 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
...
+            state_report_topic = topics.REPORTS
+        except AttributeError:
+            # Older versions of OpenStack share the PLUGIN topic.
+            state_report_topic = topics.PLUGIN
+        self.state_report_rpc = agent_rpc.PluginReportStateAPI(state_report_topic)
+
+    def _init_and_start_calico_resouce_syncer(self):
+        self.start_up_resync_thread = eventlet.spawn(self._do_startup_resync)
+
+    def _init_and_start_calico_manager(self):
```
nelljerram: Can we call this "elector" instead of "manager"?

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -377,137 +383,146 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
...
+        self.elector = Elector(
+            cfg.CONF.calico.elector_name,
+            datamodel_v2.neutron_election_key(calico_config.get_region_string()),
+            self._is_master,
+            old_key=datamodel_v1.NEUTRON_ELECTION_KEY,
+            interval=MASTER_REFRESH_INTERVAL,
+            ttl=MASTER_TIMEOUT,
+        )
+
+        self.election_thread = eventlet.spawn(self.elector.run)
```
nelljerram: Why do we need `eventlet.spawn` here when `self.election.run` will immediately do another `eventlet.spawn`?

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: Oh I see, you haven't called it "elector" because it also does compaction.  WDYT about making another separate worker process for compaction?  That's effectively what will happen when eventlet is removed anyway, and I think it would be cleaner already.

networking-calico/networking_calico/plugins/ml2/drivers/calico/election.py
```diff
@@ -99,9 +105,10 @@ def __init__(self, server_id, election_key, old_key=None, interval=30, ttl=60):
             raise ValueError("TTL %r is <= interval %r" % (ttl, interval))
 
         # Is this the master? To start with, no
-        self._master = False
+        self._is_master.value = 0
+        self._greenlet = None
 
-        # Keep the greenlet ID handy to ease UT.
+    def run(self):
```
nelljerram: I think `start` would be a more conventional name than `run`.  (`start` is for something that kicks something off but then returns immediately itself.  `run` is for something that does not return immediately.)

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -324,6 +320,10 @@ def __init__(self):
         # safe to compare this with other values returned by monotonic_time().
         self._last_status_queue_log_time = monotonic_time()
 
+        # Flag for telling workers to stop. Only applicable to the
+        # calico worker processes.
```
nelljerram: IIUC this is only for UT.  If that's right, please also comment that here.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -366,8 +372,8 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
 
         * ``CalicoStartupResyncWorker`` -> just the one-shot resync.
 
-        * ``neutron.wsgi.WorkerService`` -> connection state only, ``voting=False``.
-          Per PR #11580, API workers must never be elected master, because their
+        * ``neutron.wsgi.WorkerService`` -> connection state and worker threads only.
```
nelljerram:
```suggestion
        * ``neutron.wsgi.WorkerService`` -> indicates an API worker process.
```

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -377,137 +383,146 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
...
-                    "initialisation?  Reinitialising Calico driver.",
-                    self._my_pid,
-                    current_pid,
-                )
-            else:
-                LOG.info(
-                    "Doing Calico mechanism driver initialisation in process %s",
-                    current_pid,
-                )
+        # We were not elected. We are not the master.
```
nelljerram: This comment belongs after the following `if` line.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -377,137 +383,146 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
...
+        if self._is_master.value <= 0:
+            return False
 
-            # (Re)init the DB.
-            self.db = None
-            self._get_db()
+        # Else, let's check if we refresh the time within timeout.
+        time_till_last_refreshed = self._is_master.value - time.time()
```
nelljerram: Subtraction is wrong way round I think.  (Copilot also commented this.)

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/lib.py
```diff
@@ -574,12 +579,12 @@ def do_post_fork_actions(self, uuid_str=None):
         the one UT process, which means:
 
         - Do the same startup preparations - DB connection etc. - that the API worker
-          process would do.  These are all coded in ``_post_fork_init()``.  This allows
-          tests to later call driver entrypoints like ``update_port_postcommit()``,
-          similarly as production Neutron would.
+          process would do.  These are all coded in ``_post_fork_initialize_common()``.
```
nelljerram: Can we keep the name `_post_fork_init()` for the method that you have named `_post_fork_initialize_common()` ?   That would minimize changes overall in the PR, including here.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -377,137 +383,146 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
...
-        else:
-            self._post_fork_init(voting=True)
+        self._post_fork_inititialize_common()
 
-    @logging_exceptions(LOG)
-    def _post_fork_init(self, voting=False):
-        """_post_fork_init
+        worker_mapping = {
+            CalicoManagerWorker: self._init_and_start_calico_manager,
+            CalicoStartupResyncWorker: self._init_and_start_calico_resouce_syncer,
```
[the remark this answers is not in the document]
zhanz1: *sigh Me English bad :(

networking-calico/networking_calico/plugins/ml2/drivers/calico/workers.py
```diff
@@ -92,3 +92,78 @@ def stop(self):
...
+        super(CalicoManagerWorker, self).start(name, desc)
+
+    def stop(self):
+        """Stop service."""
+        super(CalicoManagerWorker, self).stop()
+
+    def wait(self):
+        """Wait for service to complete."""
+        super(CalicoManagerWorker, self).wait()
+
```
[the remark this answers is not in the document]
zhanz1: No the behaviors are defined in the `post_fork_initialize` function - we need to keep these worker classes.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_election.py
```diff
@@ -87,15 +101,23 @@ def _wait_and_stop(self, client, elector):
...
+        elector = election.Elector(
+            "test_basic",
+            "/bloop",
+            self.mock_is_master,
+            interval=5,
+            ttl=15,
+        )
+        elector.run()
         self._wait_and_stop(client, elector)
-        self.assertFalse(elector.master())
```
nelljerram, line 98: Hm, not sure I understand the point of this test.  However, if we want to keep it, with the same purpose as it seems to have had before, it needs an ending assertion that it hasn't been elected as master.

nelljerram (commented):
I haven't reviewed the test changes in full detail yet, but I think I have enough comments queued up to be worth releasing.

Overall, I really like the shape of this change, so thanks for proposing it.  Just some detailed comments...

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
zhanz1: I see the point, lemme break it out even further, because I think not all process need what's currently in `_post_fork_initialize_common`.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -377,137 +383,146 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
...
+        # If not, there is something wrong with elector!!
+        if not refreshed_in_time:
+            LOG.warning(
+                "The elector hasn't refreshed the lease in "
+                f"{time_till_last_refreshed}s."
             )
 
-            # Admin context used by (only) the thread that updates Felix agent status.
-            self._agent_update_context = ctx.get_admin_context()
+        return refreshed_in_time
```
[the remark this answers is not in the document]
zhanz1: Good catch... will fix this and add a test on this.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/lib.py
zhanz1: Sure, `_post_fork_init` was a big function until we split it up and that's why I renamed it :D.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_election.py
zhanz1: I think it's just testing that the function will accept good parameters. In terms of the assertion, I added it to `_wait_and_stop` as all tests need it to avoid writing it in every single test function.

networking-calico/networking_calico/plugins/ml2/drivers/calico/election.py
zhanz1: I agree, let's change this.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
zhanz1: Yeah indeed, since we don't ever want to stop :D. But probably will be used in the future when moving to threading, if we want to stop cleanly?

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
zhanz1: Yeah added a test to capture this. Cannot recall what I was thinking after the break :(.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
zhanz1: We can, but to me it feels a bit wasteful to spawn a standalone process that only does a check every X seconds. Even with eventlet removed, I feel like we should just turn this into two threads for the same reason. These two components (`Elector` and `compaction`) won't really scale as clusters grow larger.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
zhanz1: Good point. I think it was starting the thread in `__init__` and I forgot about it when I turned this into a function.

nelljerram:
@zhanz1 I have prepared https://github.com/zhanz1/calico/pull/1 to merge latest master and resolve the conflicts on this PR.  If you think that looks good, please merge it; then I'll re-approve workflows and re-review after that.

zhanz1:
> @zhanz1 I have prepared https://github.com/zhanz1/calico/pull/1 to merge latest master and resolve the conflicts on this PR. If you think that looks good, please merge it; then I'll re-approve workflows and re-review after that.

Much thanks! I have merged it.

networking-calico/networking_calico/plugins/ml2/drivers/calico/workers.py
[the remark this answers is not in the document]
nelljerram: @zhanz1 I agree that we need distinct _classes_ to be defined, but I also agree with Copilot that we don't need the `__init__`, `start`, `stop` and `wait` methods when they only call the superclass.  In other words this whole class definition can be simplified to:
```
class CalicoManagerWorker(worker.BaseWorker):
    """Service for doing election and compaction.

    The super class will trigger the post_fork_initialize in the mech driver.
    """
    def reset(self):
        config.reset_service()

```

(and the same for the next two classes)

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -442,135 +450,152 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
...
+        self._get_db()
 
-            self._my_pid = current_pid
+        # Create syncers.
+        self.subnet_syncer = SubnetSyncer(self.db)
+        self.policy_syncer = PolicySyncer(self.db)
+        self.endpoint_syncer = WorkloadEndpointSyncer(self.db, self.policy_syncer)
+
+    def _init_start_calico_resource_syncer(self):
+        self.start_up_resync_thread = eventlet.spawn(self._do_startup_resync)
```
nelljerram, line 519: Do we need the extra indirection via `eventlet.spawn` here?  It looks like we can revert to `post_fork_initialize` calling `_do_startup_resync` directly.  Just need to add `self.start_up_resync_thread = True` before that, so that UT can check it.

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -442,135 +450,152 @@ def post_fork_initialize(self, resource, event, trigger, payload=None):
...
-                )
+        if self._is_master.value <= 0:
+            # We were not elected. We are not the master.
+            return False
 
-            # (Re)init the DB.
-            self.db = None
-            self._get_db()
+        # Else, let's check if we refresh the time within timeout.
+        time_till_last_refreshed = time.time() - self._is_master.value
```
nelljerram: nit: a slightly better name would be `time_since_last_refreshed`.

nelljerram (changes requested):
Looking good.  I still need to review test_plugin_etcd.py, but the comments below cover everything else.  I'll also kick off CI now...

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
zhanz1: We can indeed. My major concern was that it could take the resync a long time to finish, and thus blocking other code after `post_for_initialize` to run (e.g., health checks? some other background tasks?), even though we create this process just for startup resync indeed. I read [OVN's code](https://github.com/openstack/neutron/blob/stable/2026.1/neutron/plugins/ml2/drivers/ovn/mech_driver/mech_driver.py#L451-L467) and believe we are doing similar things here, just that I don't know what would be a reasonable time that `post_fork_initialize` can block. Thus, I put `eventlet.spawn`, since it is just coroutine and shouldn't be expensive.

networking-calico/networking_calico/plugins/ml2/drivers/calico/workers.py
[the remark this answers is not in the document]
zhanz1: I see, lemme update and test.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_plugin_etcd.py
```diff
@@ -2172,7 +2178,8 @@ def test_vif_plug_notification(self):
             mock_db_port
         )
 
-    def test_vif_plug_no_notification_for_non_migration(self):
+    @mock.patch("eventlet.spawn")
```
nelljerram: I couldn't see why we need this added mock, and for me the UT is successful without it - so I think this can be reverted.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_plugin_etcd.py
```diff
@@ -2496,38 +2503,72 @@ def test_felix_agent_state(self):
...
+                m_sleep.side_effect = maybe_end_loop
+                self.driver.watch_status_updates(m_watcher)
+
+        m_watcher = m_watcher.return_value
+        self.assertEqual(
+            [
+                mock.call(mock.ANY),
+                mock.call(mock.ANY),
+            ],
+            [c for c in m_spawn.mock_calls if c[0] == ""],
```
nelljerram, line 2538: What does the `c[0] == ""` condition mean here?

nelljerram (commented):
Two small points for test_plugin_etcd.py

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
nelljerram: Oh I see, it might be expected that registry callbacks like `post_fork_initialize` are supposed to return quickly.  In that case I'm happy to leave this as you have it.

nelljerram:
@zhanz1 Thanks so much for your work on this.  CI is looking good, and there are just a few remaining small points:
- removing subclass methods that seem to be unnecessary in the worker classes
- var rename -> `time_since_last_refreshed`
- removing an `eventlet.spawn` mock that appears to be not needed
- understanding the `c[0] == ""` condition.

Happy with everything else, and looking forward to merging this!

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
zhanz1: Right, that's idea. It doesn't hurt to have an extra thread anyways :D.

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_plugin_etcd.py
zhanz1: Frankly this a direct copy-paste from the original test case here. I basically made two copies of it because we will have one for agent status and one for port status.

Reading the doc a little bit, the [`mock_calls`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.mock_calls) are essentially `(name, args, kwargs)`. So `c[0] == ""` essentially means that all mock calls as just `eventlet.spawn()`, not something like `eventlet.spawn.something_random()`:

```sh
>>> eventlet = MagicMock()
>>> eventlet.spawn = MagicMock()
>>> eventlet.spawn(123)
<MagicMock name='mock.spawn()' id='4401221744'>
>>> eventlet.spawn.something_random(456)
<MagicMock name='mock.spawn.something_random()' id='4401229136'>
>>> eventlet.spawn.mock_calls[0][0]
''
>>> eventlet.spawn.mock_calls[0][1]
(123,)
>>> eventlet.spawn.mock_calls[1][0]
'something_random'
```

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_plugin_etcd.py
zhanz1: Yes I dont think it's using eventlet.spawn so all good. Will remove and retest.

networking-calico/networking_calico/plugins/ml2/drivers/calico/workers.py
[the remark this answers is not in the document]
zhanz1: No we cannot get rid of this, these are [abstract methods](https://github.com/openstack/oslo.service/blob/unmaintained/2024.1/oslo_service/service.py#L93-L113):

```
2026-06-16 20:53:11.085 1955676 ERROR neutron.service   File "/usr/lib/python3/dist-packages/networking_calico/plugins/ml2/drivers/calico/mech_calico.py", line 423, in get_workers
2026-06-16 20:53:11.085 1955676 ERROR neutron.service     CalicoManagerWorker(),
2026-06-16 20:53:11.085 1955676 ERROR neutron.service TypeError: Can't instantiate abstract class CalicoManagerWorker with abstract methods stop, wait
2026-06-16 20:53:11.085 1955676 ERROR neutron.service
```
However, `__init__` is actually fine.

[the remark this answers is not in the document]
nelljerram: Thanks for trying and explaining this @zhanz1

networking-calico/networking_calico/plugins/ml2/drivers/calico/test/test_plugin_etcd.py
nelljerram: Sorry, I didn't realize this was a copy of existing code.  Thanks for explaining.

nelljerram (approved):

PR #12675, opened 2026-05-01
Fix: start-of-day syncer could apply state without host IPs (broke NP frontends)

aaaaaaaalex:
This was a latent bug which now seems to be exposed (v3.31 and later).

CT cleaner races with syncer to acquire a lock at start of day which, if CT cleaner acquired it, would block syncer from applying bad state at start of day for 10's of seconds. In 3.31 and later, it loses this race, and syncer applies its state before hostIPs are recv'd from Felix. This means the state it applies is missing NP frontend entries.

**Release note:**
```release-note
Fix a latent bug in eBPF proxy which led to ingress traffic for NodePorts getting temporarily refused during calico-node restart.
```


aaaaaaaalex:
Duplicate of: https://github.com/projectcalico/calico/pull/12667

PR #12683, opened 2026-05-02
Push kind images to a local registry instead of kind load

caseydavenport:
Run a registry:2 container on the kind docker network and have `kind-build-images` push there via the existing `make push` pipeline. Kubelet pulls on demand, which gets us a few wins:

- No per-node tar+load fan-out. Each node pulls only what it schedules, so kind workers stop holding every image (the elasticsearch case in tigera/calico-private was tipping disk).
- No hand-rolled change-detection. `docker push` is content-addressable, so unchanged layers naturally don't transfer. `load_images.sh` and its bash+jq comparison against crictl output go away.

The registry container is bound to `127.0.0.1:5000` on the host and joined to the `kind` docker network so nodes can reach it as `kind-registry:5000`. Image refs use `localhost:5000/...` consistently; after cluster create, `registry.sh configure-nodes` writes `/etc/containerd/certs.d/localhost:5000/hosts.toml` on each node redirecting that name to `http://kind-registry:5000`. (The older `containerdConfigPatches` `mirrors` syntax is rejected once `config_path` is set, hence the per-node file.) The registry container persists across `kind-cluster-destroy` so its layer cache survives reprovisioning - `kind-registry-destroy` is a separate target for a full reset.

Image refs use a dedicated hostname rather than mirroring `docker.io`. That keeps kubelet pulls deterministic (no silent fallback to upstream when a tag is missing) without breaking test pods that pull `nginx`, `busybox`, etc. from real upstream registries.

Operator `imagePullPolicy` flips to `Always` (in `build-operator.sh` for components, plus a small chart change so the operator pod itself can be set to Always via values). Required because the test-build tag is stable while the digest changes on every reload - `IfNotPresent` would not re-pull.

```release-note
None
```


hack/test/kind/registry.sh
```diff
@@ -0,0 +1,89 @@
...
+      *)
+        echo "Restarting existing ${KIND_REGISTRY_NAME} (was ${state})"
+        docker start "${KIND_REGISTRY_NAME}" >/dev/null
+        ;;
+    esac
+
+    # Connect the registry to the kind network if not already. The kind
+    # network is created on first `kind create cluster`, so this may need to
+    # run after cluster creation; calling it pre-create works once the
+    # network exists from a previous cluster, otherwise it's retried below.
```
mazdakn: To me, "otherwise it's retried below" is unclear. Worth adding more clarification.

lib.Makefile
```diff
@@ -1629,47 +1640,29 @@ $(REPO_ROOT)/key-cert-provisioner/.image.created-$(ARCH): $(call local-deps-go-f
...
+# inside the kind nodes mirrors localhost:5000 to http://kind-registry:5000.
 .PHONY: kind-build-images
-kind-build-images: $(KIND_IMAGE_MARKERS) $(REPO_ROOT)/.stamp.operator
-	@for img in $(KIND_CALICO_IMAGES); do \
-	  base=$${img%%:*}; \
-	  docker tag $$base:latest-$(ARCH) $$img; \
-	done
+kind-build-images: kind-registry-up
+	$(MAKE) -C $(REPO_ROOT) push \
+	    DEV_IMAGE_REGISTRY=localhost:5000 \
```
mazdakn, line 1653: Worth using the defined vars like `KIND_REGISTRY_PORT` to not hard code these values.

mazdakn (approved):
LGTM. Left a couple of nits.

lib.Makefile
caseydavenport: Don't think a var will help here - it's hard coded in our test manifests so it can't really be changed.

PR #12685, opened 2026-05-04
Fix flex volume driver in the consolidated calico image

caseydavenport:
Egress gateway and dikastes pods using `flexVolume.driver: nodeagent/uds` get stuck in `ContainerCreating` with `FailedMount: no volume plugin matched` on the consolidated image (#12225). Two bugs:

1. The flexvol cobra command is registered as `flexvoldrv`, so `calico component flexvol` (what the operator invokes from the calico-node DaemonSet flex-volume init container) falls through to help text. Renamed to `flexvol`.

2. The old `pod2daemon-flexvol` image had `flexvol.sh` as entrypoint that copied the driver onto the kubelet plugin dir. That step disappeared in the consolidation. Added `calico component flexvol install --target <path>` that copies the running binary at mode 0550 using a write-tmp-then-rename. Operator invokes this from the init container.

Plus an argv[0] dispatch case for `uds` in `cmd/calico/main.go`, since kubelet calls the installed driver as `<plugin-dir>/uds <init|mount|unmount>`. Same pattern as the existing `calicoctl` case.

Separate bug fix in the first commit: `felix/cmd/calico-bpf/commands/root.go` was registering `initConfig` and `setLogLevel` globally via `cobra.OnInitialize` in a package `init()`. In the consolidated binary that fires for every cobra subcommand, including kubelet flex/CSI driver execs where `HOME` is unset and `initConfig` calls `os.UserHomeDir(); os.Exit(1)`. Scoped to the calico-bpf rootCmd via `PersistentPreRun`.

Verified on a kind cluster: installed the new driver, then a pod using `nodeagent/uds` went from `ContainerCreating` to `Running 1/1`.

Operator-side change to invoke `install --target` is a follow-up against tigera/operator.

```release-note
Fixes flex volume driver registration in the consolidated calico image, which was rendering as a no-op so pods using the nodeagent/uds flex volume driver (egress gateway, application-layer policy) could not mount their policysync volume.
```

### DIFF
--- a/pod2daemon/pkg/flexvol/flexvol.go
+++ b/pod2daemon/pkg/flexvol/flexvol.go
@@ -24,9 +24,11 @@ package flexvol
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"log/syslog"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -126,7 +128,7 @@ var (
 // NewCommand returns the root cobra command for the flexvol driver.
 func NewCommand() *cobra.Command {
 	rootCmd := &cobra.Command{
-		Use:           "flexvoldrv",
+		Use:           "flexvol",
 		SilenceErrors: true,
 		SilenceUsage:  true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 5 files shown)


matthewdupre (approved):

pod2daemon/pkg/flexvol/flexvol.go
```diff
@@ -192,14 +194,80 @@ func NewCommand() *cobra.Command {
...
+	src, err := os.Executable()
+	if err != nil {
+		return fmt.Errorf("resolving running binary: %w", err)
+	}
+
+	in, err := os.Open(src)
+	if err != nil {
+		return fmt.Errorf("opening %s: %w", src, err)
+	}
+	defer func() { _ = in.Close() }()
```
mazdakn: Is it possible to spit out some sort of warning here instead of ignoring the error?
Same question for the rest.

mazdakn (approved):
LGTM. Left one comment.

PR #12687, opened 2026-05-05
confd: let UPDATE_EXPECTED_DATA create missing golden files

caseydavenport:
Both \`compareOutput\` and \`expectOutput\` in \`confd/tests/helpers_test.go\` call \`t.Fatalf\` the moment a golden file doesn't exist on disk, so the \`UPDATE_EXPECTED_DATA=true\` path only ever refreshes goldens that are already there. Anyone adding a new confd FV test still has to check in empty placeholder files for all six BIRD configs by hand before the harness will fill them — easy to miss, and not what the README suggests the flag does.

This change treats a missing golden under \`UPDATE_EXPECTED_DATA\` as a mismatch (so the existing write path runs), and \`MkdirAll\`s the parent so a brand-new golden directory works on first run.

No behavior change when \`UPDATE_EXPECTED_DATA\` is unset.

\`\`\`release-note
None
\`\`\`

### DIFF
--- a/confd/tests/helpers_test.go
+++ b/confd/tests/helpers_test.go
@@ -838,9 +838,21 @@ func compareOutput(t *testing.T, outputDir, goldenDir string) {
 
 		expectedPath := filepath.Join("compiled_templates", goldenDir, f)
 		want, err := os.ReadFile(expectedPath)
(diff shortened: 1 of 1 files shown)


mazdakn (approved):
LGTM.
Thanks. I've felt this pain several times.

PR #12688, opened 2026-05-05
Decouple BPF program attachment from host IP

fasaxc:
## Description

In Calico's BPF dataplane, Felix refuses to attach BPF TC programs to any
interface (workload or host) until it knows the local node's host IP. The
host IP arrives via `proto.HostMetadata*Update` messages, which originate
from the `Node` resource's BGP / Addresses fields.

If the Node is deleted, or its IP fields are patched out, no
`HostMetadata*Update` arrives. After a Felix restart in this state, the
normal cleanup path (`cleanupOldAttach`, jump-map deletes) tears down the
existing programs, but `wepApplyPolicyToDirection` /
`attachDataIfaceProgram` then refuse to reattach with `"unknown host IP"` —
breaking connectivity on every interface Felix manages, even though the
*only* missing piece is the local host IP.

The host IP is meaningfully consumed by very few BPF features:

- VXLAN return-path encap source (`tc.c:975, 1087, 1200` —
  `STATE->ip_src = HOST_IP`).
- An IPv4 RPF strict-mode link-local check (`rpf.h:134`) — naturally
  evaluates false when `HOST_IP=0`.
- A debug log in `vxlan_attempt_decap`. The decap decision itself uses a
  route-table lookup, not `HOST_IP`.

Everything else — policy, conntrack, NAT, ICMP, kube-proxy replacement,
non-VXLAN forwarding — works fine without `HOST_IP`.

## Approach

Stop using \"no host IP\" as a hard gate. Programs now attach as normal
with `HOST_IP=0` as a sentinel. The Go-side already handled `nil` host IP
in the globals plumbing (`copy(dst, (nil net.IP).To4())` is a no-op, so
`globalData.HostIPv*` stays at zero); the gates were the only thing
short-circuiting before that.

The three VXLAN encap-source sites in `tc.c` now check `ip_void(HOST_IP)`
and, if true, drop the packet with a new `CALI_REASON_NO_HOST_IP` /
`DroppedNoHostIP` counter (slot 26 in the counters map; `RESERVED1` is
preserved for downstream use).

`updateHostIP` is also fixed to handle the empty-IP transition properly:
previously it would silently no-op on an empty `Ipv4Addr`, leaving stale
state. It now transitions `m.v4.hostIP` / `m.v6.hostIP` back to `nil` and
triggers the same dirty-marking that a real IP change does, so programs
are re-attached with `HOST_IP=0` when the host IP goes away (and back to
the real value when it returns).

## Testing

- `go build ./felix/...` clean.
- `gofmt -l` clean on modified files.
- Existing \`must re-attach programs when hostIP changes\` test passes.
- Full \`TestIntdataplane\` suite passes (XDP-state and constructor
  failures observed locally are pre-existing and stem from missing
  Docker-built `bpf-apache/bin/filter.o` and libbpf cgo headers, not from
  these changes).

\`make build-bpf\` (BPF verifier check via Docker) and \`make fix-changed\`
need to run in CI; not run locally.

## Followups (not in this PR)

- BPF unit tests under \`felix/bpf/ut/\` exercising the three encap-drop
  paths with \`hostIP=0\`.
- Felix FV that strips the Node resource's IP, restarts Felix, and
  asserts intra-node pod connectivity survives.

**Release note:**
\`\`\`release-note
Calico BPF dataplane: Felix no longer refuses to program BPF interfaces when the local node's host IP is unknown. Programs attach as normal; only VXLAN return-path encapsulation is dropped (counted as \`Dropped: VXLAN encap when host IP unknown\`) until the host IP becomes available again. This prevents a Felix restart from cascading into a node-wide connectivity outage when the Node resource has been deleted or its IP cleared.
\`\`\`


felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -879,40 +879,49 @@ func (m *bpfEndpointManager) withIface(ifaceName string, fn func(iface *bpfInter
...
-		// but taking it now makes us robust to refactoring.
-		m.ifacesLock.Lock()
-		for ifaceName := range m.nameToIface {
-			m.withIface(ifaceName, func(iface *bpfInterface) (forceDirty bool) {
-				iface.dpState.v4Readiness = ifaceNotReady
-				iface.dpState.v6Readiness = ifaceNotReady
-				return true
-			})
+		if ip == nil {
+			logrus.WithField("ipAddr", ipAddr).Warn("Cannot parse hostip, no change applied")
```
fasaxc, line 894: Felix should treat this as IP missing, not ignore it.

fasaxc:
Closed in favour of Tomas' version

PR #12698, opened 2026-05-06
Install runtime deps for e2e binary in scheduled CI

caseydavenport:
The scheduled k8s-e2e CI matrix has been failing since the e2e test binary started CGO-linking against libbpf. libbpf.a is statically linked, but libelf and libz come along as dynamic deps, and the test scripts call uuidgen. The hashrelease run path uses `golang:${GO_VERSION}-bookworm` as the runtime image, which doesn't ship any of those, so the binary fails at startup with `error while loading shared libraries: libelf.so.1`.

This installs `libelf1`, `zlib1g`, and `uuid-runtime` on container start for the bookworm path (the per-PR `calico/go-build` path already has them), and adds `git config --global --add safe.directory '*'` to silence the dubious-ownership warnings showing up in the same logs.

Sticking with the slim base rather than switching to `calico/go-build` for both paths because some runners are disk-constrained.

```release-note
None
```

### DIFF
--- a/.semaphore/end-to-end/scripts/phases/run_tests.sh
+++ b/.semaphore/end-to-end/scripts/phases/run_tests.sh
@@ -63,7 +63,7 @@ if [[ -n "${E2E_BINARY:-}" ]]; then
   else
     GO_VERSION=$(grep '^GO_BUILD_VER=' ./metadata.mk | cut -d= -f2 | cut -d- -f1)
     RUN_IMAGE="golang:${GO_VERSION}-bookworm"
-    PRE_RUN="apt-get update -qq && apt-get install -y --no-install-recommends libelf1 zlib1g uuid-runtime >/dev/null"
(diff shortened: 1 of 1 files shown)


lwr20 (approved):
A couple of Nits:   
  - Quoting change for E2E_TEST_CONFIG. Changed from a Make-style positional arg to '${E2E_TEST_CONFIG}' interpolated into a bash -c string. The outer shell expands the variable, then bash sees a single-quoted literal — fine for typical config-file paths, but if the value ever contains a single quote it'll break. Not a concern today; flag it if config values become user-controlled.          
  - `>/dev/null` on the install step hides which package failed if a future apt resolve breaks. Minor — -qq plus default install verbosity is already low; you could drop the redirect and let it through for easier postmortems. Not blocking.                                                                                                                                                                 
  - No retry on apt-get update. Debian mirrors do occasionally 503. If you see flakes, wrapping in a small retry loop is the cheap fix.

PR #12704, opened 2026-05-06
Use kubebuilder markers for numorstring int-or-string types

caseydavenport:
Alternative to https://github.com/projectcalico/calico/pull/12643. Adds three kubebuilder markers to each of `numorstring.Port`, `numorstring.Protocol`, and `numorstring.Uint8OrString` so vanilla controller-gen produces the correct CRD output without needing the patched controller-gen baked into `calico/go-build`.

Mirrors the existing pattern on `numorstring.DSCP`, which has been using these markers on master already.

Switches the `api` and `libcalico-go` Makefiles to invoke controller-gen via `go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.18.0` so the upstream binary runs instead of the patched one in PATH.

The output collapses to `type: integer` (or `string`) plus `x-kubernetes-int-or-string: true`, instead of `anyOf: [integer, string]` plus `x-kubernetes-int-or-string: true`. The two forms are functionally identical at the API server, which special-cases `x-kubernetes-int-or-string` and accepts both regardless of declared type. DSCP fields on master already use this exact form.

The `IPAMBlock.allocations` nullable patch still applies cleanly with vanilla controller-gen and is left as-is.

## Follow-ups

- Drop the controller-gen patches from the `projectcalico/toolchain` `calico-go-build` image once this lands.

```release-note
None
```

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -63,7 +63,7 @@ build: gen-files examples
 .PHONY: gen-files
 gen-files .generate_files: lint-cache-dir clean-generated
 	# Generate CRDs
-	$(DOCKER_RUN) $(CALICO_BUILD) sh -c '$(GIT_CONFIG_SSH) controller-gen crd:allowDangerousTypes=true,crdVersions=v1,deprecatedV1beta1CompatibilityPreserveUnknownFields=false paths=./pkg/apis/... output:crd:dir=config/crd/'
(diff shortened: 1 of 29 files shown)


caseydavenport:
Closing — the marker approach loses the anyOf: [integer, string] form, which is a real regression for OpenAPI tooling that doesn't honor x-kubernetes-int-or-string. Sticking with the post-processor design in #12643.

PR #12709, opened 2026-05-07
IPIP manager: do not update parent iface to empty string

mazdakn:
**Release note:**
```release-note
TBD
```

### DIFF
--- a/felix/dataplane/linux/ipip_mgr.go
+++ b/felix/dataplane/linux/ipip_mgr.go
@@ -123,8 +123,10 @@ func (m *ipipManager) OnUpdate(protoBufMsg any) {
 	switch msg := protoBufMsg.(type) {
 	case *proto.HostMetadataUpdate:
 		m.logCtx.WithField("hostname", msg.Hostname).Debug("Host update/create")
-		if msg.Hostname == m.hostname && msg.Ipv4Addr != "" {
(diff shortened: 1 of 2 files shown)


mazdakn:
Closing: on a closer look this guard isn't the right fix.

- An empty parent address doesn't tear down the tunnel device — nothing deletes it, and `keepDeviceInSync` just parks on its quiet 10s branch. `parentDevice` isn't cleared either, so routes are untouched.
- The motivating case doesn't reach this code: a cleared BGP spec makes `DataplanePassthru` fire `OnHostMetadataRemove`, and `ipipManager` still clears the address there.
- In the case the guard does change, it makes things worse: `Ipv4Addr == ""` means the node has no v4 address at all (`extractNodeAddress` already falls back to InternalIP/ExternalIP), so keeping the old value leaves `detectParentIface` hunting for an address that is no longer on any link — a `LinkList` plus an `AddrList` per link, and a warning, every second for the life of the process.

The `""` skip in `noEncapManager`/`vxlanManager` that this mirrors is an IP-version filter (one message carries both families, one manager per family), which doesn't apply to `ipipManager` — it is v4-only by construction.

#13504 instead removes the equivalent skip from `noEncapManager`, where it was swallowing genuine address removals, and backs off the retry cadence in `keepDeviceInSync` so a permanently unresolvable parent device no longer polls netlink at 1Hz.

PR #12710, opened 2026-05-07
Build the calico/calico image with ko

caseydavenport:
Replaces the Dockerfile + multi-stage BuildKit pipeline for the calico/calico image with ko. Same for test-signer (key-cert-provisioner FVs) and csi-node-driver-registrar, which is now its own image instead of being bundled inside calico/calico.

Companion operator PR: tigera/operator#4805.

## What changed

Three images switched off Dockerfile builds and onto ko:

- `calico/calico` (cmd/calico)
- `csi-node-driver-registrar`, now its own image carved out of `calico/calico` (the binary used to be bundled inside it)
- `test-signer` (used by key-cert-provisioner FVs)

Supporting changes:

- Defaulted typha's `LogFilePath` to empty so containerized typha doesn't try to `mkdir /var/log/calico` at startup. Avoids needing `typha.cfg` in the image.
- Dropped the bundled registrar binary and `typha.cfg` from `docker/calico/Dockerfile`.
- Pulled the cni-plugin `-upgrade` flow into a `calico ipam-upgrade` subcommand so cmd/calico has a single user-facing entry point.
- Resolved binaries via PATH (`calico`, `calico-ipam`, `csi-node-driver-registrar`) instead of hardcoded `/usr/bin/...` in manifests, charts, probes, init containers, and the CNI install path. The ko image puts binaries at `/ko-app/<name>` and adds that to PATH; the legacy `/usr/bin/...` still resolves in the Dockerfile image, so both variants work during the transition.

## Measurements

`make -C cmd/calico image-all` on warm Go cache + warm BuildKit. Builds amd64, arm64, ppc64le, s390x (the legacy run also builds FIPS amd64; the ko run was measured without FIPS since the FIPS docker socket plumbing is a separate fix).

| Metric | Legacy (master Dockerfile) | ko build | Delta |
|---|---|---|---|
| Wall-clock `image-all` | 7m 46s | 5m 09s | -2m 37s (~1.5x faster) |
| Wall-clock `image` (amd64 only, warm) | 8.96s | 6.98s | -2.0s |
| Image size (uncompressed, amd64) | 250 MB | 161 MB | -89 MB (-36%) |
| Layers | 2 (base + COPY shim) | 3 (base + ko-meta + binary) | +1 |
| calico binary in image | 214 MB | 154 MB | -60 MB |
| Bundled registrar | 29 MB | 0 (separate image) | -29 MB |

Most of the `image-all` win comes from arm64: the legacy build runs `buildx --platform linux/arm64` under qemu, ko cross-compiles natively from the amd64 host. Single-arch amd64 is a much smaller difference.

The 89 MB image-size delta breaks down as:

- ~60 MB from `-s -w` strip flags (ko sets these via `.ko.yaml`; the legacy `build_binary` macro does not).
- ~29 MB from removing the bundled registrar.

If we added `-s -w` to `build_binary` and dropped the registrar from the legacy image, the image-size delta would close to roughly zero. The size win is mostly about stripping symbols and decoupling the registrar, not ko itself.

## Where ko actually wins

- Build time on multi-arch, mostly via arm64. ko cross-compiles directly on the host; the legacy build does the same work under qemu via buildx, which is slow.
- No Dockerfile maintenance. The `FROM scratch AS source` shim, `COPY --from=builder` juggling, `BIN_DIR` / `TARGETARCH` build args all go away.
- Reproducible image digests across machines for the same input.

## Where it loses

- No symlinks, no extra files. `/usr/bin/calicoctl` and `/usr/bin/calico-ipam` are gone in the ko image, and there's no way to drop arbitrary content (typha.cfg, manifests, etc). Worked around via PATH-resolved bare commands and the new `calico ipam-upgrade` subcommand, but it bit us in several places.
- One extra layer (a 0-byte ko metadata layer). Cosmetic.
- Tied to ko's release cadence. Less mature than buildx, and we'd own the upgrade pain.
- Base image must live in a registry. No "FROM local-image" support, which limits experimentation with ad-hoc base images.
- FIPS-style CGO+boringcrypto builds need ko to run inside calico/go-build (libpcap headers come from there). Doable, but it's extra Makefile machinery per FIPS-enabled image, and the current implementation has a docker-socket permission bug under `image-all` that's still being worked through.

## Bottom line

~1.5x faster on `image-all` (mostly arm64), Dockerfile maintenance goes away, image is smaller (though most of that you could also get with `-s -w` on the legacy build). ko's constraints (no symlinks, no arbitrary content) bit us in a few places but were workable for cmd/calico.

```release-note
None
```


caseydavenport:
Closing this. Re-ran benchmarks fairly (4 archs vs 4 archs, warm caches) and the original 11x build-time claim doesn't hold up.

Real picture:

Wins:
- ~1-3 min faster on multi-arch `image-all`, almost entirely from arm64 cross-compile vs qemu under buildx
- Dockerfile deletion

Costs:
- 89 MB image-size delta is misleading. ~60 MB is `-s -w` (one-line change to `build_binary`), ~29 MB is the registrar split (orthogonal to ko)
- Registrar becomes its own image: new release artifact, operator/manifest/csi coordination, FIPS variant for it
- Lost symlinks (`/usr/bin/calicoctl`, `/usr/bin/calico-ipam`): updates across manifests, charts, probes, CNI install, FV tests
- Lost "drop arbitrary content": had to default typha's `LogFilePath` empty just to ship without `typha.cfg`
- New `calico ipam-upgrade` subcommand entry point
- FIPS path needs docker-in-docker and a working-but-fragile pipeline
- ko's release cadence and "no FROM local-image" friction we'd own

Blast radius is large across two repos for a build-time win that's narrower than first claimed. Cheaper to address arm64-via-qemu directly if that's what we care about.

`-s -w` on `build_binary` and the registrar carve-out are both worth doing as standalone PRs. Tagging those as follow-ups.

PR #12712, opened 2026-05-08
[v3.32] Build patched nftables 1.1.1 in node and istio CNI install images

caseydavenport:
**Cherry-pick history**
- Pick onto **release-v3.32**: projectcalico/calico#12660

Backports the patched nftables 1.1.1 RPM build into calico/node and the Istio CNI install image to fix nft segfaults when a newer nftables is present elsewhere on the host. See #12660 for the full rationale.

Conflicts were in the Semaphore CI templates, `node/Makefile`, and `istio/Makefile`. The `10-prerequisites.yml` modifications to producer blocks (`Pull: go-build image`, `Build: calico/node/whisker image`) were dropped on this branch since those producer blocks don't exist on release-v3.32 — only the new `Build: nftables RPMs` block was added, with consumers wired up directly in the node and istio CI blocks.

**Release note:**
```release-note
Fixes nft binary segfaults in calico/node and the Istio CNI install image when newer nftables is in use elsewhere on the host.
```


sshambar:
@caseydavenport BTW, tried out the release-3.32 branch (as of this commit, head of tree fails to build), and deployed in a test environment and it appears to work on systems with nft udata sets present in the host namespace :)

PR #12278, opened 2026-03-29
GC orphaned IPAMHandle CRDs in kube-controllers

caseydavenport:
IPAMHandle CRDs can be left behind when the IPAM release process is interrupted (CNI plugin killed, kube-controllers crash, etc). The handle CRD persists but has no corresponding allocations in any block. These accumulate over time and pollute etcd — on long-lived clusters with high pod churn, tens of thousands of orphaned handles can accumulate (see #8548 for a real example with 30k+ orphaned handles).

kube-controllers previously had no visibility into orphaned handles at all. It only tracked handles that had active allocations in blocks.

This adds an `orphanHandleTracker` that watches IPAMHandle CRDs via the syncer DataFeed and cross-references them against block allocations tracked by `handleTracker`. When a handle has no allocations for longer than the leak grace period, it is deleted. The tracking is event-driven from both sides:

- Block allocation removal: when a handle loses its last allocation, check if the handle CRD still exists and mark it as an orphan candidate
- Handle CRD arrival: when a new/updated handle CRD arrives with no matching allocations, mark it as an orphan candidate
- Grace period: same as the existing leak grace period, to avoid deleting handles that are briefly empty during normal operation

Also includes the `handleTracker`/`confirmedLeaks` cleanup fix for `onBlockDeleted` that was identified in a separate investigation (#11115).

Fixes https://github.com/projectcalico/calico/issues/8548

```release-note
Garbage collect orphaned IPAMHandle CRDs that have no matching allocations in any IPAM block, preventing unbounded accumulation of stale handle resources in etcd.
```

PR #12713, opened 2026-05-08
kube-controllers: garbage-collect divergent IPAM handles

fasaxc:
## Summary

Calico's IPAM keeps allocations across two separate CRDs — `IPAMBlock` (the actual IP allocations) and `IPAMHandle` (an index used by `ReleaseByHandle` and, for KubeVirt VM pods, IP-persistence decisions). Because they're separate resources with no transactional update, a CNI-plugin / node crash mid-operation can leave them out of sync. kube-controllers already GCs leaked block allocations but never touched handles. This PR adds a handle reconciler to the IPAM GC loop that uses blocks as source-of-truth and repairs handles that have been divergent for several consecutive cycles.

Failure modes addressed:

- **Orphan handle** — handle exists, no block references it.
- **Skewed handle** (count > truth) — release crashed between `updateBlock` and `decrementHandle`. Critical for VM handles since IP-persistence reads `Spec.Block` totals.
- **Missing handle** — block has an allocation whose `Attributes[i].HandleID` references a handle that doesn't exist; CNI's `ReleaseByHandle` can't clean up.
- **Stuck soft-delete** — `Spec.Deleted=true` set by a CNI plugin that died before the immediately-following hard-delete.

## Approach

Per cycle: derive expected per-block reference counts from the block cache, diff against a new watcher-fed `c.allHandles` cache, and after a divergence has been stable for 3 consecutive cycles **and** the handle's ResourceVersion has not changed, repair via CAS through the backend client.

- `libcalico-go/lib/backend/k8s/resources/ipam_handle.go` — implement `Watch()` (was `OperationNotSupported`) using the same `cache.NewListWatchFromClient` pattern as `IPAMBlock`.
- `kube-controllers/pkg/controllers/utils/syncer.go` — register `IPAMHandleListOptions` so the watcher syncer picks up handles.
- `kube-controllers/pkg/controllers/node/ipam.go` — register `IPAMHandleKey` notifications, dispatch into a new `c.allHandles` cache via `handleHandleUpdate`. Acquire `bapi.Client` in the constructor via the existing `Backend()` accessor pattern (matches `networkpolicy/policy_name_migrator.go`).
- `kube-controllers/pkg/controllers/node/ipam_handle_gc.go` — the reconciler. Repairs go directly through `bapi.Client.Create/Update/DeleteKVP`; no thin wrappers on `ipam.Interface`.

### Safety rules (encoded in code)

- Never directly delete a non-empty handle. Skewed → rewrite Block to expected; only the follow-up cycle, where the handle is observed to be empty, issues the hard-delete. Avoids ever turning a counter mismatch into a permanent IP leak by removing a handle that some racing CNI op was about to use.
- Repair direction is "set to expected", not "delta-adjust" — eliminates drift.
- CAS-or-nothing: every write uses the cached Revision. CAS conflict and `AlreadyExists` are benign races that drop the diff and re-evaluate next cycle.
- No repair before `bapi.InSync`.

### Knob and metrics

- Defaults to **on**; disable via `DISABLE_IPAM_HANDLE_GC=true` env var on the kube-controllers deployment.
- Prometheus:
  - `ipam_handle_gc_repairs_total{action="overwrite|create|delete|undelete"}`
  - `ipam_handle_gc_diffs_observed_total{class="orphan|missing|skewed|stuck_deleted"}`
  - `ipam_handle_gc_cas_conflicts_total`
  - `ipam_handle_gc_divergent_handles` gauge

## Test plan

- [x] 27 unit tests covering classifier, stability tracker, syncer-dispatch path, end-to-end repair for orphan/missing/skewed/stuck-deleted, safety rule (no delete of non-empty), CAS-conflict re-evaluation, knob disable
- [x] Existing IPAM controller UTs still green
- [x] Watch contract: integration test in `ipam_handle_test.go` that opens a watch, creates a handle, asserts an `ADDED` event arrives
- [ ] Manual smoke on a kind cluster: inject each inconsistency type and verify repair within ~3 GC intervals via logs/metrics
- [ ] Specifically for VM handles: create `IPAMHandle` named `k8s-pod-network.vmi.<ns>.<vm>` with inflated `Spec.Block`, no real allocations; verify rewrite-to-empty then orphan-delete
- [ ] CI: full kube-controllers test suite + libcalico-go ipam suite

**Release note:**
```release-note
kube-controllers now garbage-collects divergent IPAMHandle objects (orphan, skewed reference counts, missing, or stuck-soft-deleted) by reconciling against block reality.
```

### DIFF
--- a/kube-controllers/pkg/config/runconfig.go
+++ b/kube-controllers/pkg/config/runconfig.go
@@ -83,6 +83,14 @@ type NodeControllerConfig struct {
 	// The grace period used by the controller to determine if an IP address is leaked.
 	// Set to 0 to disable IP address garbage collection.
 	LeakGracePeriod *v1.Duration
+
+	// IPAMHandleGCEnabled controls whether the IPAM handle reconciler runs as
+	// part of the IPAM GC loop. When true, the controller compares each
+	// IPAMHandle against block reality and repairs handles that have been
+	// divergent for several consecutive cycles. Defaults to true; can be
+	// disabled via the DISABLE_IPAM_HANDLE_GC env var as an emergency
+	// off-switch.
+	IPAMHandleGCEnabled bool
 }
 
 type AutoHostEndpointConfig struct {
@@ -382,6 +390,13 @@ func mergeConfig(envVars map[string]string, envCfg Config, apiCfg v3.KubeControl
 		if apiCfg.Controllers.Node != nil {
 			rc.Node.LeakGracePeriod = apiCfg.Controllers.Node.LeakGracePeriod
 			status.RunningConfig.Controllers.Node.LeakGracePeriod = apiCfg.Controllers.Node.LeakGracePeriod
+		}
+
+		// IPAM handle GC is on by default; allow operators to disable it via env
+		// var as an emergency off-switch (no API field yet).
+		rc.Node.IPAMHandleGCEnabled = true
+		if v, _ := strconv.ParseBool(os.Getenv("DISABLE_IPAM_HANDLE_GC")); v {
+			rc.Node.IPAMHandleGCEnabled = false
 		}
 
 		if envCfg.DatastoreType != "kubernetes" {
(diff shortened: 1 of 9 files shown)


caseydavenport:
@fasaxc haha https://github.com/projectcalico/calico/pull/12278

PR #12715, opened 2026-05-08
felix: don't drop wireguard pubkey publish after a transient datastore failure

caseydavenport:
`handleWireguardStatUpdateFromDataplane` tracked the in-flight update in a single `current` variable shared across IP versions. If a v6 reconcile failed (e.g. context deadline exceeded talking to etcd) and a v4 update arrived from the dataplane while the retry was pending, the v4 update overwrote `current` and the v6 retry was silently dropped. The peer node then never saw the v6 pubkey and v6 wireguard tunnels stayed broken until the next time felix restarted or the v6 key changed.

Track pending pubkeys per IP version instead, so a new update for one version doesn't displace a pending retry for the other.

Also adds a unit test that fault-injects a one-shot v6 reconcile failure followed by a v4 update. Fails on master, passes with the fix.

**Release note:**
```release-note
Fix a bug where felix could silently drop a wireguard public key publication after a transient datastore failure if an update for the other IP version arrived during the retry window, leaving wireguard tunnels broken until the next restart.
```

### DIFF
--- a/felix/daemon/daemon.go
+++ b/felix/daemon/daemon.go
@@ -1357,6 +1357,7 @@ func (fc *DataplaneConnector) reconcileWireguardStatUpdate(dpPubKey string, ipVe
 
 func (fc *DataplaneConnector) handleWireguardStatUpdateFromDataplane(
 	reconcile func(pubKey string, ipVersion proto.IPVersion) error,
+	done <-chan struct{},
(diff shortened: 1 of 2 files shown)


aaaaaaaalex (approved):
LGTM

PR #12719, opened 2026-05-08
Add docker buildx bake config for calico/node

caseydavenport:
Adds a `docker-bake.hcl` with a `calico-node` target. Reuses the existing `node/Dockerfile` unchanged - it's already buildkit-friendly so no Dockerfile work is needed.

The point of this PR is to enable **registry-backed layer cache** for fresh CI runners. Today, every Semaphore runner redoes the runit compile, dnf/microdnf installs, and AlmaLinux repo grafting from scratch even though those layers don't change between commits. With `cache-from=type=registry`, BuildKit pulls those layers from a master-populated cache.

Local prototype numbers (linux/amd64):

| Scenario | Make | Bake |
|---|---|---|
| Image rebuild, no source change | 4.4s | 1.0s |
| Image rebuild, calico binary changed | 24s | 21s |
| Fresh runner + registry cache, no source change | n/a | **8.3s** |
| Fresh runner + registry cache + binary change | n/a | **49s** |

The bottom two rows are the CI win; Make has no equivalent today.

This is **build config only** - no Dockerfile or source changes. Pure-Go images are built with ko (https://github.com/projectcalico/calico/pull/12710), not bake.

The `calico` and `mountns` binaries are pre-built by the existing Make path and copied in via `BIN_DIR`. BUILD_CACHE (the GCS-backed `.go-pkg-cache` restore) covers the Go-build side; this PR covers the image-assembly side. The two stack.

Status: draft. Lands the bake config but does not yet wire `cache-from`/`cache-to` in CI - that's a follow-up so we can review the build config in isolation.

Tracking: CORE-12789. Findings doc: `~/designs/calico/build-system-experiments/node-bake-prototype.md`.

## Release Note

```release-note
None
```


caseydavenport:
Closing this. Looked at the actual layer structure and the wins don't pencil out.

The image is one 418 MB squashed layer (`FROM scratch; COPY --from=ubi / /`, deliberately to hide cleaned-up files from CVE scanners). 275 MB of that is the calico binary itself, which churns every PR. Cold-cache build is ~48s; the ~30s of "system" work (dnf install, runit compile, etc.) is what BuildKit already caches intra-stage.

The real lever for "speed up node CI builds across fresh runners" is registry-backed cache (`--cache-from=type=registry,...`), which the existing `DOCKER_BUILD` macro can do without bake. Bake's parallel/group features are nice but most of our images don't share enough base layers to benefit.

Layer split (system/bpf/binary into separate squashed final layers) would save registry push/pull bytes on PRs that only touch Go code, but no build-time savings since BuildKit already caches the system stages.

Not worth pursuing.

PR #12729, opened 2026-05-11
Fix tiered RBAC e2e tests and fix e2e CI failures (#12271)

MichalFupso:
Cherry-pick of https://github.com/projectcalico/calico/pull/12271

Two things in this PR:

**Tiered RBAC cross-version compat fix**

The existing tiered RBAC e2e tests use bare policy names (setting `Spec.Tier` without including the tier prefix in the policy name), which is only supported in v3.32+. On v3.31 and earlier the API requires the `tier.policyname` format, so these tests fail.

Tags the existing tests with `NoTierPrefix` so older branches can skip them via `-skip=NoTierPrefix`, and adds a new `CalicoDescribe` with a tier-prefixed test that provides basic RBAC coverage on all versions.

**E2E CI fixes**

Fixes several e2e test failures from the real GCP kubeadm CI run:

- **QoS tests**: Update `iperf3` image tag from `3.16` (deleted from Docker Hub) to `latest`
- **Staged policy tests**: Fix tier cleanup ordering — the `AfterEach` was trying to delete the tier before `DeferCleanup` had removed the policies inside it. Moved tier deletion to `DeferCleanup` (registered before policies, so it runs after them in LIFO order) and wrapped in `Eventually` to handle propagation delay
- **Conncheck retry loop**: Fix bug where expected-failure checks retried for the full 30s timeout. The loop condition `err != nil || result != exp.ExpectedResult` kept retrying on `err` even when `result` already matched, since `err` is inherently non-nil for any FAILURE result
- **Istio ambient test**: `tcpdump` not found in the client pod (`alpine:3` doesn't have it). `WithCapture()` now also sets the image to `netshoot` which has tcpdump and other network tools
- **IPIP toggle test**: Connectivity check was running before route convergence. Moved the `Eventually` wait for tunl0 routes before the connectivity check
- **BGP peer/export tests**: Added `CalicoNodeStatus`-based diagnostics (BGP session state, peer info, learned routes) so we have better data next time these fail

**Release note:**
```release-note
TBD
```

### DIFF
--- a/e2e/pkg/tests/bgp/export.go
+++ b/e2e/pkg/tests/bgp/export.go
@@ -37,6 +37,7 @@ var _ = describe.CalicoDescribe(
 	describe.WithTeam(describe.Core),
 	describe.WithFeature("BGPPeer"),
 	describe.WithCategory(describe.Networking),
+	describe.RequiresBGPMesh(),
 	describe.WithSerial(),
 	"BGP export tests",
 	func() {
@@ -47,6 +48,7 @@ var _ = describe.CalicoDescribe(
 		var server1 *conncheck.Server
 		var client1 *conncheck.Client
 		var restoreBGPConfig func()
+		var bgpStatus *BGPStatusMonitor
 
 		// Create a new framework for the tests.
 		f := utils.NewDefaultFramework("bgp-export")
@@ -69,9 +71,13 @@ var _ = describe.CalicoDescribe(
 			Expect(installation.Spec.CalicoNetwork).NotTo(BeNil(), "CalicoNetwork is not configured in the Installation")
 			Expect(installation.Spec.CalicoNetwork.BGP).NotTo(BeNil(), "BGP is not enabled in the cluster")
 			Expect(*installation.Spec.CalicoNetwork.BGP).To(Equal(v1.BGPEnabled), "BGP is not enabled in the cluster")
+			requireNonVXLANCluster(cli)
 
 			// Ensure full mesh BGP is functioning before each test.
 			restoreBGPConfig = ensureInitialBGPConfig(cli)
+
+			// Set up CalicoNodeStatus resources for BGP diagnostics and convergence waits.
+			bgpStatus = NewBGPStatusMonitor(cli)
 
 			// Create an IP pool for the test.
 			pool := v3.NewIPPool()
@@ -112,6 +118,9 @@ var _ = describe.CalicoDescribe(
 			checker.AddClient(client1)
 			checker.Deploy()
 
+			// Wait for BGP to be fully established before baseline connectivity check.
+			bgpStatus.WaitForEstablished()
+
 			// Verify initial connectivity.
 			checker.ResetExpectations()
 			checker.ExpectSuccess(client1, server1.ClusterIPs()...)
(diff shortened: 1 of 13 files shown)


e2e/pkg/tests/bgp/export.go
```diff
@@ -37,6 +37,7 @@ var _ = describe.CalicoDescribe(
 	describe.WithTeam(describe.Core),
 	describe.WithFeature("BGPPeer"),
 	describe.WithCategory(describe.Networking),
+	describe.RequiresBGPMesh(),
```
caseydavenport, line 40: I'm not sure we need this? We already label this with `Feature:BGPPeer` which is effectively the same thing?

e2e/pkg/tests/bgp/utils.go
```diff
@@ -169,3 +190,148 @@ func waitForBGPEstablishedForNode(cli ctrlclient.Client, node string) {
...
+		}
+		for _, route := range status.Status.Routes.RoutesV4 {
+			logrus.Infof("[BGP]   Node %s route: %s via %s type=%s", name, route.Destination, route.Gateway, route.Type)
+		}
+	}
+}
+
+// WaitForEstablished waits until all nodes have at least one established BGP peer
+// and no non-established peers. Logs BGP state on each poll for diagnostics.
+func (m *BGPStatusMonitor) WaitForEstablished() {
```
caseydavenport, line 270: I think this looks largely duplicated with the local functions defined above - we should just have a single implementation of this, so let's consolidate around one or the other?

e2e/pkg/tests/policy/tiered_rbac.go
```diff
@@ -64,6 +66,7 @@ var _ = describe.CalicoDescribe(
 	describe.WithTeam(describe.Core),
 	describe.WithCategory(describe.Policy),
 	describe.WithFeature("Tiered-RBAC"),
+	describe.WithNoTierPrefix(),
```
caseydavenport, line 69: This is weird... I could have sworn I already made this fix...

e2e/pkg/utils/conncheck/conncheck.go
```diff
@@ -454,7 +454,7 @@ func (c *connectionTester) runConnection(ctx context.Context, exp *Expectation,
 	// those changes taking effect. So a short retry loop is helpful.
 
 loop:
-	for err != nil || result != exp.ExpectedResult {
```
caseydavenport, line 457: Is this related / intentional?

e2e/pkg/utils/images/images.go
```diff
@@ -26,7 +26,7 @@ const (
 	TestWebserver = "gcr.io/kubernetes-e2e-test-images/test-webserver:1.0"
 	Agnhost       = "registry.k8s.io/e2e-test-images/agnhost:2.47"
 	RapidClient   = "quay.io/tigeradev/rapidclient"
-	Iperf3        = "docker.io/networkstatic/iperf3:3.16"
+	Iperf3        = "docker.io/networkstatic/iperf3:latest"
```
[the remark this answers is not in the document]
caseydavenport: I think this is legit-ish?

PR #12742, opened 2026-05-12
WIP: Qos galera repro

nelljerram:
**Release note:**
```release-note
TBD
```


nelljerram:
The issue here turned out to be unrelated to the details of the DB implementation (including Galera).  Rather it was caused by other coding issues (wrong transaction wrapping and using DB APIs at the wrong level) that have now been fixed separately.

PR #12745, opened 2026-05-12
[BPF] Fix inverted jump-map selection in pol-prog FV helper

tomastigera:
## Summary

`NumPolProgramsByEntryPoint()` in `felix/fv/infrastructure/bpfpolprog/bpfpolprog.go` selects the wrong jump map: it queries the **Egress** map when the caller asks for `"ingress"` and the **Ingress** map when the caller asks for `"egress"`. Both the netkit and TC branches are inverted the same way.

The entry-point index passed in is already direction-specific (the caller looks up `IngressPolicyV4` or `EgressPolicyV4` from `BPFIfState`) and the programs are pinned in the matching map, so querying the opposite map silently returns counts from the wrong index space.

Introduced by #12167 and moved verbatim into the new sub-package by #12717. Reported by Copilot review on the EE auto-pick.

## Why CI didn't catch it

The only caller is `bpf_pol_scale_test.go`, which asserts the same value (1, then 0) for both directions. After the EP is removed/stopped both maps converge to 0 either way, and the "1 dummy drop-all" state also happens to match for both directions, so the inversion was masked.

## What this PR does

1. Moves the `(netkit, direction) -> map name` selection into a new `jump.PolicyJumpMapName()` helper, next to the map parameter definitions it returns. The helper has the corrected mapping.
2. Adds a table-driven unit test in `felix/bpf/jump` that pins the contract for all four `{tc,netkit} × {ingress,egress}` combinations. The test asserts on direction (`_ing` vs `_egr`) and netkit (`_nk_`) substrings of the map name independent of the implementation, so an inversion would fail it.
3. Puts the test under `felix/bpf/jump/` (rather than `felix/fv/infrastructure/bpfpolprog/`) so `make ut` picks it up — `felix/fv/` is excluded from the unit-test run.

## Test plan

- [x] `make -C felix ut` covers `TestPolicyJumpMapName` (passes).
- [ ] `make -C felix fv-bpf GINKGO_FOCUS="sub-program cleanup tests"` still green (no regression on the existing FV path).

## Notes for reviewers

Test-helper-only change; no shipping behaviour changes.

### DIFF
--- a/felix/bpf/jump/map.go
+++ b/felix/bpf/jump/map.go
@@ -96,6 +96,23 @@ var NetkitEgressMapParameters = maps.MapParameters{
 	Version:    2,
 }
 
+// PolicyJumpMapName returns the pinned name of the policy program jump map for
+// the given attach mode (netkit vs TC/TCX) and direction ("ingress"/"egress").
+// Each direction lives in its own prog_array, so picking the wrong one returns
+// indices from the wrong namespace.
+func PolicyJumpMapName(netkit bool, ingressOrEgress string) string {
+	switch {
+	case netkit && ingressOrEgress == "egress":
+		return NetkitEgressMapParameters.VersionedName()
+	case netkit:
+		return NetkitIngressMapParameters.VersionedName()
+	case ingressOrEgress == "egress":
+		return EgressMapParameters.VersionedName()
+	default:
+		return IngressMapParameters.VersionedName()
+	}
+}
+
 func NetkitMaps() []maps.Map {
 	return []maps.Map{
 		maps.NewPinnedMap(NetkitIngressMapParameters),
--- a/felix/bpf/jump/map_test.go
+++ b/felix/bpf/jump/map_test.go
@@ -1,0 +1,60 @@
+// Copyright (c) 2026 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


tomastigera:
Closing — the Copilot finding this PR was based on is a false positive. The original `NumPolProgramsByEntryPoint` selecting the Egress map when caller passes `"ingress"` is *not* inverted: it compensates for the workload kernel-vs-policy direction flip (`felix/dataplane/linux/bpf_ep_mgr.go:3353-3367`). Workload-ingress policy sub-programs are pinned in the kernel-Egress jump map.

Proof from a CI run with the change applied (felix 25.10 BPF netkit job, scale test failed): the ifstate dump shows `IngressPolicyV4 == EgressPolicyV4` for each workload (the per-hook allocators each start at 0, so by coincidence the same slot index is used in both maps). Under the original code, `(ingressOrEgress="ingress", entryPointIdx=IngressPolicyV4)` lands on Egress map slot N — the right slot for workload-ingress sub-programs. My change made it query the Ingress map at slot N instead, which contains the trivial workload-egress program (count 1), and the scale test which expects >5 timed out.

There is a latent fragility here — the helper relies on `IngressPolicyV4 == EgressPolicyV4` being a coincidence rather than a guarantee — but addressing that is a larger refactor of `BPFPolEntryPointIdx` and its callers, not what this PR was about. Closing.

PR #12752, opened 2026-05-13
[BPF] Don't BYPASS HEP egress for NAT-outgoing flows (CI-1988)

tomastigera:
Bug fix.

For NAT-outgoing flows the BPF dataplane uses two conntrack entries: a workload-side entry keyed by the pre-SNAT 5-tuple (with the `NAT_OUT` flag) and a host-side entry keyed by the post-SNAT 5-tuple. The host-side entry is what lets reply traffic at `FROM_HEP` skip policy on the source node.

`CALI_SKB_MARK_NAT_OUT` has been a member of the BYPASS family since dc3bb62b94 (Jul 2020). That commit's rationale was "NAT outgoing means the traffic is leaving the cluster so there can only be one workload program in play" — but the HEP egress program is the only place on the outbound path where the post-SNAT 5-tuple is observable to BPF. The bypass mark short-circuits HEP egress at the top of `calico_tc_main`, so the host-side entry is never written. The first reply packet at `FROM_HEP` therefore misses BPF conntrack and is evaluated against host-endpoint policy.

For TCP the mid-flow-fall-through-to-iptables path saves the reply; UDP has no equivalent, so a HEP deny-ingress policy drops the reply. This is the customer-reported behaviour in [CI-1988](https://tigera.atlassian.net/browse/CI-1988): on eBPF clusters with a HostEndpoint GNP, UDP responses to pod-originated egress are blocked while TCP/ICMP work.

Change: `CALI_SKB_MARK_NAT_OUT` is now a SEEN mark (not BYPASS). The iptables SNAT rule still triggers off the NAT_OUT bit; HEP egress now runs and writes the host-side conntrack entry. Effectively a revert of dc3bb62b94.

Added FV reproducer in `felix/fv/bpf_test.go` that applies a HEP ingress-Deny GNP on the source node, runs pod → externalClient with NAT-outgoing, and asserts the two expected BPF conntrack entries (pre-SNAT workload-side and post-SNAT host-side) are present.

**Release note:**
```release-note
ebpf: Fix HostEndpoint policy blocking UDP return traffic for SNAT'd pod egress.
```

[CI-1988]: https://tigera.atlassian.net/browse/CI-1988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/felix/fv/bpf_test.go
+++ b/felix/fv/bpf_test.go
@@ -1719,6 +1719,67 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 						bpfWaitForGlobalNetworkPolicy(tc.Felixes[0], "eth0", "egress", "host-0-1")
 					})
 
+					// CI-1988: With a HostEndpoint Deny-ingress policy on the source node,
+					// the SNAT'd return traffic for pod-to-off-cluster connections must
+					// still be matched by BPF conntrack and forwarded back to the pod —
+					// i.e. it must not be evaluated against the host-endpoint policy
+					// (which would deny it as host-bound).
+					//
+					// applyOnForward defaults to false on the deny GNP, so pod egress
+					// (forwarded through eth0) is unaffected. Only a packet that the
+					// dataplane classifies as host-bound hits this policy. In iptables
+					// mode the reply matches nf_conntrack uniformly across protocols and
+					// is classified as forwarded ESTABLISHED. The bug report (Visa) is
+					// that in BPF mode UDP replies miss conntrack and get denied, while
+					// TCP/ICMP replies are correctly matched.
+					It("should not block SNAT'd return traffic via HEP policy", func() {
+						denyPol := api.NewGlobalNetworkPolicy()
+						denyPol.Name = "ci-1988-hep-deny-ingress"
+						denyOrder := float64(10)
+						denyPol.Spec.Order = &denyOrder
+						denyPol.Spec.Selector = "ep-type == 'host' && node == '" + tc.Felixes[0].Name + "'"
+						denyPol.Spec.Types = []api.PolicyType{api.PolicyTypeIngress}
+						denyPol.Spec.Ingress = []api.Rule{{Action: "Deny"}}
+						_ = createPolicy(denyPol)
+
+						bpfWaitForGlobalNetworkPolicy(tc.Felixes[0], "eth0", "ingress", "ci-1988-hep-deny-ingress")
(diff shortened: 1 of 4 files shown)


felix/fv/bpf_test.go
```diff
@@ -1719,6 +1719,67 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+					// applyOnForward defaults to false on the deny GNP, so pod egress
+					// (forwarded through eth0) is unaffected. Only a packet that the
+					// dataplane classifies as host-bound hits this policy. In iptables
+					// mode the reply matches nf_conntrack uniformly across protocols and
+					// is classified as forwarded ESTABLISHED. The bug report (Visa) is
+					// that in BPF mode UDP replies miss conntrack and get denied, while
+					// TCP/ICMP replies are correctly matched.
+					It("should not block SNAT'd return traffic via HEP policy", func() {
+						denyPol := api.NewGlobalNetworkPolicy()
+						denyPol.Name = "ci-1988-hep-deny-ingress"
```
fasaxc: We should have a deny egress too; this traffic should bypass normal policy on the HEP.

fasaxc (changes requested):
If we don't bypass, what happens to the NATted packets?  Do we put them through host egress policy?  That'd be wrong, this traffic should bypass HEP normal policy because it's workload traffic, agree we probably need some sort of conntrack entry on the HEP to get the return path to work.

fasaxc:
PR description has wrong format

felix/fv/bpf_test.go
tomastigera: yes added in the test

tomastigera:
> If we don't bypass, what happens to the NATted packets?  Do we put them through host egress policy?  That'd be wrong, this traffic should bypass HEP normal policy because it's workload traffic, agree we probably need some sort of conntrack entry on the HEP to get the return path to work.

only if the policy has `applyOnForward: false` - in that case it is forwarded traffic and the policy should not apply. If it has `applyOnForward: true` the egress policy applies

fasaxc (approved):
As discussed, this isn't quite right because HEP egress policy shouldn't apply to the outgoing WEP flow.  However, it seems like a step in the right direction; at least it's not possible to let the traffic through in a conntracked way.

PR #12769, opened 2026-05-14
fix(felix): exclude LB-only IPPools from all-ipam-pools ipset

### LINKED ISSUE #12767: SNAT is skipped for traffic destined to LoadBalancer-only IPPools shared across clusters
When same IPPool is used for LoadBalancer use case, traffic is not SNATed when service is failed over from local cluster to remote cluster.

Image

## Expected Behavior

When a pod in cluster A sends traffic to a LoadBalancer VIP (e.g. `10.20.40.94:443`) that is hosted on cluster B, the traffic should be SNATed to the node IP (because the workload IPPool has `natOutgoing: true`), allowing the packet to be routed correctly between clusters.

## Current Behavior

Traffic from pods to LoadBalancer VIPs in a shared IPPool is **not** masqueraded. The pod's source IP (`100.90.14.133`) appears on the wire instead of the node IP, causing return traffic to be lost and breaking cross-cluster connectivity.

This happens because [masq_mgr.go](https://github.com/projectcalico/calico/blob/master/felix/dataplane/linux/masq_mgr.go#L126) unconditionally adds **all** IPPool CIDRs (including LoadBalancer-only pools) to the `cali40all-ipam-pools` ipset. The NAT-outgoing rule matches:

```
-m set --match-set cali40masq-ipam-pools src ! --match-set cali40all-ipam-pools dst -j MASQUERADE
```

Since the LB pool CIDR is in `cali40all-ipam-pools`, traffic destined to it is [excluded from masquerade](https://github.com/projectcalico/calico/blob/master/felix/rules/nat.go#L62-L65) — even though no pod workloads are assigned from that pool and the destination is on a different cluster.

## Possible Solution

Exclude IPPools with `allowedUses: [LoadBalancer]` (i.e. pools that are **not** used for workload/tunnel assignment) from the `cali40all-ipam-pools` ipset.

1. Create two Calico clusters sharing a common LoadBalancer IPPool:
   ```yaml
   kind: IPPool
   metadata:
     name: lb-pool
   spec:
     cidr: 10.20.40.64/27
     allowedUses: ["LoadBalancer"]
     assignmentMode: Manual
     natOutgoing: true
   ```
2. Create a workload IPPool (`100.90.0.0/16`, `natOutgoing: true`) on both clusters.
3. On cluster B, expose a service with a LoadBalancer VIP from the shared pool (e.g. `10.20.40.94`).
4. From a pod on cluster A, send traffic to `10.20.40.94:443`.
5. Observe on the wire that the source IP is the pod IP (no SNAT), and the connection fails.

## Context

This affects multi-cluster deployments that announce anycast services or shared LoadBalancer IP pools for inter-region/inter-cluster failover.

## Your Environment

* Calico version: v3.30.7

defo89:
IPPools with `allowedUses: [LoadBalancer]` were unconditionally added to the `cali40all-ipam-pools` ipset, causing traffic destined to LoadBalancer VIPs to be exempted from SNAT. This broke cross-cluster connectivity when a shared LB pool was used across clusters.

Only add pools that include Workload or Tunnel uses to the `all-ipam-pools` ipset. LoadBalancer-only pools are excluded so that the NAT-outgoing rule continues to masquerade traffic destined to those CIDRs.

BPF may have a similar issue but is out of scope for this PR.

**Release note:**
```release-note
Fix SNAT being skipped for traffic destined to LoadBalancer-only IPPools by excluding them from the all-ipam-pools ipset.
```

### DIFF
--- a/felix/dataplane/linux/masq_mgr.go
+++ b/felix/dataplane/linux/masq_mgr.go
@@ -18,9 +18,8 @@ import (
 	"slices"
 	"strings"
 
-	log "github.com/sirupsen/logrus"
-
 	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+	log "github.com/sirupsen/logrus"
 
 	dpsets "github.com/projectcalico/calico/felix/dataplane/ipsets"
 	"github.com/projectcalico/calico/felix/ipsets"
--- a/felix/dataplane/linux/masq_mgr.go
+++ b/felix/dataplane/linux/masq_mgr.go
@@ -15,6 +15,7 @@
 package intdataplane
 
 import (
+	"slices"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
@@ -105,7 +106,9 @@ func (d *masqManager) OnUpdate(msg any) {
 		// defers and coalesces the update so removing then adding the
 		// same IP is a no-op anyway.
 		logCxt.Debug("Removing old pool.")
-		d.ipsetsDataplane.RemoveMembers(rules.IPSetIDAllPools, []string{oldPool.Cidr})
+		if !isLoadBalancerOnly(oldPool) {
+			d.ipsetsDataplane.RemoveMembers(rules.IPSetIDAllPools, []string{oldPool.Cidr})
+		}
 		if oldPool.Masquerade {
 			logCxt.Debug("Masquerade was enabled on pool.")
 			d.ipsetsDataplane.RemoveMembers(rules.IPSetIDNATOutgoingMasqPools, []string{oldPool.Cidr})
@@ -123,8 +126,15 @@ func (d *masqManager) OnUpdate(msg any) {
 		}
 
 		// Update the IP sets.
-		logCxt.Debug("Adding IPAM pool to IP sets.")
-		d.ipsetsDataplane.AddMembers(rules.IPSetIDAllPools, []string{newPool.Cidr})
+		// Exclude pools that are exclusively for LoadBalancer use from the
+		// all-ipam-pools ipset. These pools don't contain workload or tunnel
+		// addresses, so traffic destined to them should still be masqueraded.
+		if !isLoadBalancerOnly(newPool) {
+			logCxt.Debug("Adding IPAM pool to all-pools IP set.")
+			d.ipsetsDataplane.AddMembers(rules.IPSetIDAllPools, []string{newPool.Cidr})
+		} else {
+			logCxt.Debug("Skipping LoadBalancer-only pool from all-pools IP set.")
+		}
 		if newPool.Masquerade {
 			logCxt.Debug("IPAM has masquerade enabled.")
 			d.ipsetsDataplane.AddMembers(rules.IPSetIDNATOutgoingMasqPools, []string{newPool.Cidr})
@@ -148,4 +158,17 @@ func (m *masqManager) CompleteDeferredWork() error {
 	m.dirty = false
 
 	return nil
+}
+
+// isLoadBalancerOnly returns true if the pool's AllowedUses contains only
+// "LoadBalancer" and no workload/tunnel uses. Such pools should not be
+// included in the all-ipam-pools ipset because their CIDRs do not represent
+// local workload addresses and traffic to them should still be masqueraded.
+func isLoadBalancerOnly(pool *proto.IPAMPool) bool {
+	uses := pool.GetAllowedUses()
+	if len(uses) == 0 {
+		// Empty means default (Workload + Tunnel), so include in all-pools.
+		return false
+	}
+	return !slices.Contains(uses, "Workload") && !slices.Contains(uses, "Tunnel")
 }
--- a/felix/dataplane/linux/masq_mgr.go
+++ b/felix/dataplane/linux/masq_mgr.go
@@ -131,11 +131,11 @@ func (d *masqManager) OnUpdate(msg any) {
 		// Exclude pools that are exclusively for LoadBalancer use from the
 		// all-ipam-pools ipset. These pools don't contain workload or tunnel
 		// addresses, so traffic destined to them should still be masqueraded.
-		if !isLoadBalancerOnly(newPool) {
+		if isLoadBalancerOnly(newPool) {
+			logCxt.Debug("Skipping LoadBalancer-only pool from all-pools IP set.")
+		} else {
 			logCxt.Debug("Adding IPAM pool to all-pools IP set.")
 			d.ipsetsDataplane.AddMembers(rules.IPSetIDAllPools, []string{newPool.Cidr})
-		} else {
-			logCxt.Debug("Skipping LoadBalancer-only pool from all-pools IP set.")
 		}
 		if newPool.Masquerade {
 			logCxt.Debug("IPAM has masquerade enabled.")
--- a/felix/dataplane/linux/masq_mgr.go
+++ b/felix/dataplane/linux/masq_mgr.go
@@ -20,6 +20,8 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
+	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+
 	dpsets "github.com/projectcalico/calico/felix/dataplane/ipsets"
 	"github.com/projectcalico/calico/felix/ipsets"
 	"github.com/projectcalico/calico/felix/proto"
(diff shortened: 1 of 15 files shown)


defo89:
If possible - it would be appreciated if this also could be added to `Calico v3.31.6` milestone.

caseydavenport:
> If possible - it would be appreciated if this also could be added to Calico v3.31.6 milestone.

Once merged it can be ported to v3.32 and v3.31 at which point it will automatically grab the milestone.

defo89:
Seems I should not have force-pushed - sorry for that.

fasaxc:
Is that the only use of that IP pool?  
* If so, we should rename it to say "cali-network-ip-pools".
* If not, need to check the other uses and maybe split into two IP pools.

libcalico-go/lib/backend/model/ippool.go
```diff
@@ -91,27 +88,27 @@ func (options IPPoolListOptions) KeyFromDefaultPath(path string) Key {
...
+	CIDR             net.IPNet             `json:"cidr"`
+	IPIPInterface    string                `json:"ipip"`
+	IPIPMode         encap.Mode            `json:"ipip_mode"`
+	VXLANMode        encap.Mode            `json:"vxlan_mode"`
+	Masquerade       bool                  `json:"masquerade"`
+	IPAM             bool                  `json:"ipam"`
+	Disabled         bool                  `json:"disabled"`
+	DisableBGPExport bool                  `json:"disableBGPExport"`
+	AssignmentMode   v3.AssignmentMode     `json:"assignment_mode"`
+	AllowedUses      []v3.IPPoolAllowedUse `json:"allowedUses,omitempty"`
```
mazdakn:
```suggestion
	AllowedUses      []v3.IPPoolAllowedUse `json:"allowed_uses,omitempty"`
```

felix/dataplane/linux/masq_mgr.go
```diff
@@ -123,8 +126,15 @@ func (d *masqManager) OnUpdate(msg any) {
 		}
 
 		// Update the IP sets.
-		logCxt.Debug("Adding IPAM pool to IP sets.")
-		d.ipsetsDataplane.AddMembers(rules.IPSetIDAllPools, []string{newPool.Cidr})
+		// Exclude pools that are exclusively for LoadBalancer use from the
+		// all-ipam-pools ipset. These pools don't contain workload or tunnel
+		// addresses, so traffic destined to them should still be masqueraded.
+		if !isLoadBalancerOnly(newPool) {
```
mazdakn: nit: I find `!isLoadBalancerOnly` condition a bit hard to reason, specially that it also follows an `else`. It's better to reverse the condition:
```
if isLoadbalancerOnly() {
 // Skip
} else {
 // Add ip to ipset
}
```

felix/dataplane/linux/masq_mgr.go
```diff
@@ -149,3 +159,16 @@ func (m *masqManager) CompleteDeferredWork() error {
...
+// "LoadBalancer" and no workload/tunnel uses. Such pools should not be
+// included in the all-ipam-pools ipset because their CIDRs do not represent
+// local workload addresses and traffic to them should still be masqueraded.
+func isLoadBalancerOnly(pool *proto.IPAMPool) bool {
+	uses := pool.GetAllowedUses()
+	if len(uses) == 0 {
+		// Empty means default (Workload + Tunnel), so include in all-pools.
+		return false
+	}
+	return !slices.Contains(uses, "Workload") && !slices.Contains(uses, "Tunnel")
```
mazdakn: We check for `workload` and `tunnel` values, but in future we might add new ones. So it's better to check if a value is not `LoadBalancer` instead. (we should also use the defined const instead: https://github.com/projectcalico/calico/blob/3df7ac6d4fca2e004e815781d2c213e10af3b8cb/api/pkg/apis/projectcalico/v3/ippool.go#L159-L164)

Given this, we can do this instead:

```
return slices.Contains(uses, string(api.IPPoolAllowedUseLoadBalancer)) && len(uses) == 1
```

mazdakn (changes requested):
@defo89 Thanks for the fix. I left a few comments and nits. 

As you also mentioned, this fix needs to be applied to bpf dataplane. Are you planning to add a follow up PR for it?

defo89:
> @defo89 Thanks for the fix. I left a few comments and nits.
> 
> As you also mentioned, this fix needs to be applied to bpf dataplane. Are you planning to add a follow up PR for it?

@mazdakn Yes, that seems to be the case. Would you prefer the BPF dataplane fix to be part of the same PR or a separate one?

Update - after checking this I'd suggest to have a follow-up PR for BPF dataplane fix to avoid exploding this one and touching completely different code paths.

felix/dataplane/linux/masq_mgr.go
defo89: That makes sense. I have noticed that import is done as `apiv3` in other places - hence reused same pattern.

defo89:
> Is that the only use of that IP pool?
> 
> * If so, we should rename it to say "cali-network-ip-pools".
> * If not, need to check the other uses and maybe split into two IP pools.

`IPSetIDAllPools = "all-ipam-pools"` is also used in:
- [calico/felix/rules/nat.go](https://github.com/projectcalico/calico/blob/a974a6d6c1100ffaddb89ad2a364c2bcba17b2ad/felix/rules/nat.go#L57) - used for NAT outgoing rule.
- [calico/felix/rules/static.go](https://github.com/projectcalico/calico/blob/a974a6d6c1100ffaddb89ad2a364c2bcba17b2ad/felix/rules/static.go#L1091) - used for DSCP egress rule.

Both uses share the same intent: "check whether the destination a Calico network address" — so I guess no split is needed.

In case of rename, we would need an upgrade path, so old rules are cleaned up on existing nodes. ref [LegacyV4IPSetNames](https://github.com/projectcalico/calico/blob/a974a6d6c1100ffaddb89ad2a364c2bcba17b2ad/felix/rules/rule_defs.go#L250-L252)

What do you think?

fasaxc:
Calico cleans up any old IP set that starts with `cali` that it doesn't need any more.  It also sequences iptables/ip set updates properly to avoid trying to delete an in-use IP set.  So, renaming an IP set should be fairly straightforward.

BPF IP sets are driven by the same input events, so names must agree there.  BPF has special handling for named IP sets like the all-pools one, it maps them to reserved numeric IDs.

mazdakn:
@defo89 it's fine to address BPF in a separate PR.

mazdakn (approved):
LGTM

mazdakn:
@defo89 you need to address the conflict. You just need to run 'make generate'.

felix/proto/felixbackend.pb.go
```diff
@@ -1,18 +1,17 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.36.11
-// 	protoc        v6.33.5
+// 	protoc        v7.34.1
 // source: felixbackend.proto
```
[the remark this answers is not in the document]
defo89: I have re-generated with correct protoc version.

felix/proto/felixbackend_grpc.pb.go
```diff
@@ -1,14 +1,13 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-grpc v1.6.1
-// - protoc             v6.33.5
+// - protoc-gen-go-grpc v1.6.2
+// - protoc             v7.34.1
 // source: felixbackend.proto
```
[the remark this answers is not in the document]
defo89: I have re-generated with correct protoc version.

felix/dataplane/linux/masq_mgr.go
```diff
@@ -149,3 +160,16 @@ func (m *masqManager) CompleteDeferredWork() error {
...
+// "LoadBalancer" and no workload/tunnel uses. Such pools should not be
+// included in the network-ip-pools ipset because their CIDRs do not represent
+// local workload addresses and traffic to them should still be masqueraded.
+func isLoadBalancerOnly(pool *proto.IPAMPool) bool {
+	uses := pool.GetAllowedUses()
+	if len(uses) == 0 {
+		// Empty means default (Workload + Tunnel), so include in network-ip-pools.
+		return false
+	}
+	return slices.Contains(uses, string(apiv3.IPPoolAllowedUseLoadBalancer)) && len(uses) == 1
```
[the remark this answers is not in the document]
defo89: I am not sure this is worth it since Typha and Felix bump usually ship together when upgrading Calico.

mazdakn:
@defo89 thanks for your contribution. Please continue with back ports and a follow up fix for BPF.

PR #12770, opened 2026-05-14
OpenStack: add Neutron->Calico resync scale benchmark for CI

nelljerram:
Adds a large-scale benchmark exercising the new ``calico-resync`` one-shot resync tool against a populated Neutron DB, run as part of the existing networking-calico Semaphore pipeline.  At each scale (default 100, 1000, 3000 ports) the test populates real Neutron resources via the REST API as admin, INSERTs fake "Calico per-host agent (felix)" rows into the agents table so ports can bind to arbitrary hosts, runs three steady-state resyncs and one cold-etcd resync, and prints both a grep-able ``RESYNC_SCALE_RESULT`` summary line and per-run JSON.  Nothing about Neutron, neutron-lib or SQLAlchemy is mocked.

Adds a generic perf-publishing pipeline under ``hack/perf/`` so the resync numbers (and future perf measurements from any test in the monorepo) can be landed in the Lens ElasticSearch cluster for long-term trend tracking:

* Producers write one JSON file per measurement (scenario scalars only) to ``artifacts/perf/<index_family>/``.
* ``hack/perf/cmd/send-perf-results`` (stdlib-only Go binary) walks that tree once at the end of a CI job, applies any ``hack/perf/index-templates/<family>.json`` templates (idempotent upserts so template changes auto-propagate), augments each doc with CI metadata (``@timestamp``, ``git_commit``, ``git_branch``, ``code_version``, ``ci_run_id``, ``pr_number``, ``env``), and POSTs to ``<family>_<UTC year>/_doc``.
* All failure paths log + exit 0 unless ``--require-creds`` is set; Lens outages must not fail the producing test.
* Local runs skip silently when ``ELASTICSEARCH_URL`` is unset, and the injected ``env=dev`` field lets dashboards filter out accidental dev pushes.
* ``hack/perf/README.md`` documents the full convention -- producer contract, schema rules (flat scalars only, one measurement per doc), ES/Kibana gotchas, index-naming convention, starter index template, verification queries, and operational hygiene.

The resync benchmark writes per-iteration docs to ``artifacts/perf/benchmark_data_neutron_resync/`` and bootstrap.sh invokes ``send-perf-results`` once after the test.

Setup steps that aren't code (called out in the README); these will be automated later, or must be done manually:
* Get the Lens ES endpoint + API key from ``#banzai`` and attach as Semaphore secrets to the OpenStack pipeline.
* Create a Kibana index pattern (``benchmark_data_neutron_resync_*``) with ``@timestamp`` as the time field, plus a starter Lens visualisation.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/networking-calico/devstack/resync_scale_test.py
+++ b/networking-calico/devstack/resync_scale_test.py
@@ -373,7 +373,18 @@ def create_ports(conn, num_ports, networks, subnets, sgs, hosts):
         created.extend(p for p in results if p is not None)
         if (batch_start // PORT_BULK_BATCH) % 10 == 0:
             LOG.info("  ...created %d/%d ports", len(created), num_ports)
+
     LOG.info("Port creation done: %d ports", len(created))
+    if len(created) != num_ports:
+        # A shortfall here is a real problem (quota, race, transient DB
+        # pressure, ...) -- not something to paper over.  Raise so the
+        # caller aborts this scale before any resync runs, and we don't
+        # ship a perf doc to Lens labelled at full scale that actually
+        # measured scale - N.
+        raise RuntimeError(
+            "Port creation shortfall: only %d/%d ports created"
+            % (len(created), num_ports)
+        )
     return created
 
 
@@ -696,6 +707,26 @@ def run_one_scale(scale, conn, etcd_client, db_args, neutron_conf, extra_conf):
         LOG.info("Cold-etcd run")
         cold = run_calico_resync(neutron_conf, extra_conf)
 
+        # Belt-and-braces: cross-check that the cold run actually saw the
+        # full expected set of endpoints.  ``created`` is what landed in
+        # etcd from a cold start (we just wiped it), so it should equal
+        # ``scale``; ``correct`` should be 0 because nothing pre-existed.
+        # Mismatches indicate either a populate-phase bug we missed
+        # above, or a bind/postcommit failure between port-create and
+        # the resync (port made it into Neutron but never into etcd).
+        endpoints = (cold.get("phases") or {}).get("endpoints") or {}
+        if endpoints.get("created") != scale or endpoints.get("correct") != 0:
+            raise RuntimeError(
+                "Cold resync mismatch at scale=%d: endpoints.created=%r "
+                "(want %d), endpoints.correct=%r (want 0)"
+                % (
+                    scale,
+                    endpoints.get("created"),
+                    scale,
+                    endpoints.get("correct"),
+                )
+            )
+
         summarise(scale, num_networks, num_sgs, num_hosts, cold, steady_runs)
         return True
     finally:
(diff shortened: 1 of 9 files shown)


lwr20 (approved):
LGTM

networking-calico/devstack/resync_scale_test.py
```diff
@@ -0,0 +1,785 @@
...
+    try:
+        # Provision fake agent rows first.  The mech driver checks
+        # agent presence and aliveness during bind_port; both queries
+        # hit MySQL directly.
+        LOG.info("Inserting %d fake agent rows...", num_hosts)
+        insert_fake_agents(db_args, hosts)
+
+        nets, subs = create_networks_and_subnets(conn, num_networks)
+        sgs = create_security_groups(conn, num_sgs)
+        create_ports(conn, scale, nets, subs, sgs, hosts)
```
lwr20, line 694: **`ports={scale}` records the requested count, not the number actually created**

`create_ports` returns the list of successfully-created ports, but the return value is discarded here, so the actual count never reaches `summarise` / `_make_perf_docs`. `_create_one` swallows any `create_port` exception and returns `None` (filtered out at line 373), so partial failures are silent — and the note about `QueuePool` timeouts at scale ≥ 1000 (lines 748-754) means the populate phase can hit the same DB pressure under 8-way concurrency.

Net effect: if N ports fail to create, the steady-state resync runs over `scale − N` endpoints but the doc shipped to Lens is still labelled `ports={scale}` / `scale_ports={scale}`, so a half-populated run enters the long-term trend series mislabelled at full scale.

`len(created)` is already computed (line 376) — suggest threading it through and recording it as a separate field (e.g. `ports_created`) alongside the requested `ports`, and warning (or failing the scale) when it falls short:

```python
created = create_ports(conn, scale, nets, subs, sgs, hosts)
if len(created) != scale:
    LOG.warning("Only %d/%d ports created", len(created), scale)
```

Separately worth confirming whether all *created* ports actually reach etcd via the bind/postcommit hook before the steady-state runs fire — `len(created)` is only an upper bound on what the resync sees. If `calico-resync`'s `endpoints` phase reports a synced count, recording that would make the actual measured scale unambiguous.

nelljerram: Addressed in https://github.com/projectcalico/calico/pull/12770/commits/04ea62c52e02cef1eda73a34ba285e488af664aa.  The key point here is that we do not expect any creation errors, so we do want to flag and fail the CI overall if the number of ports successfully created is different than what was intended.  But apart from that there is no extra value from saying, e.g., that we only got 997 ports when it should have been 1000.

lwr20: LGTM.

coutinhop (approved):

PR #12771, opened 2026-05-14
e2e: test to cover cluster routing mode

mazdakn:
Adding a test to cover cluster routing mode in e2e tests.

**Release note:**
```release-note
TBD
```


e2e/pkg/tests/bgp/export.go
```diff
@@ -42,24 +42,27 @@ var _ = describe.CalicoDescribe(
...
-		var cli ctrlclient.Client
-		var checker conncheck.ConnectionTester
-		var server1 conncheck.Server
-		var client1 conncheck.Client
-		var restoreBGPConfig func()
-		var bgpStatus *BGPStatusMonitor
+		var (
+			err              error
+			cli              ctrlclient.Client
+			checker          conncheck.ConnectionTester = conncheck.NewDummyConnectionTester()
```
mazdakn: We can instead use `nil` check in the `AfterEach`, but felt it would be cleaner to not include many conditions, and make the test as expressive as possible.

e2e/pkg/tests/bgp/export.go
```diff
@@ -42,24 +42,27 @@ var _ = describe.CalicoDescribe(
...
 
 		ginkgo.BeforeEach(func() {
-			// Create a connection tester for the test.
-			checker = conncheck.NewConnectionTester(f)
-
 			cli, err = client.New(f.ClientConfig())
 			Expect(err).NotTo(HaveOccurred())
 
+			if felixProgramsClusterRoutes(cli) {
+				ginkgo.Skip("Skipping test since it needs BGP to be responsible for cluster routing")
```
caseydavenport: We are trying to avoid having tests self-skip - this should be an explicit fail - i.e., "This test requires that BIRD be used for cluster routing" and a label `describe.WithRequiresBGP()` that can then be skipped / selected.

e2e/pkg/tests/bgp/export.go
caseydavenport: A cleaner alternative might be for the `checker.Stop()` call to be a nil pointer receive and bail early if nil?

```go
func (c *ConnectionTester) Stop() error {
  if c == nil {
    return nil
  }
  ...
}
```

WDYT?

mazdakn: Right! I don't know what I was thinking :-/

e2e/pkg/tests/bgp/export.go
mazdakn: hmm I see.

e2e/pkg/tests/bgp/export.go
mazdakn: See this https://github.com/projectcalico/calico/pull/12771#discussion_r3251002595
The issue is that `ConnectionTester` is an interface.

caseydavenport: ahhh yes, forgot that I did that.

e2e/pkg/tests/networking/cluster_routes.go
```diff
@@ -0,0 +1,139 @@
...
+		})
+
+		ginkgo.Context("with an IPIP-Always pool", func() {
+			ginkgo.BeforeEach(func() {
+				setUpPoolAndPods(func(p *v3.IPPool) {
+					p.Spec.IPIPMode = v3.IPIPModeAlways
+				})
+			})
+
+			framework.ConformanceIt("Felix programs the IPIP cluster route (proto 80)",
```
mazdakn: need to update the test description.

e2e/pkg/tests/networking/routes.go
```diff
@@ -0,0 +1,214 @@
...
+func expectedClusterRouteProto(cli ctrlclient.Client) RouteProto {
+	inst := &v1.Installation{}
+	Expect(cli.Get(context.Background(), ctrlclient.ObjectKey{Name: "default"}, inst)).To(Succeed(), "Error querying Installation")
+	if inst.Spec.CalicoNetwork != nil &&
+		inst.Spec.CalicoNetwork.ClusterRoutingMode != nil &&
+		*inst.Spec.CalicoNetwork.ClusterRoutingMode == v1.ClusterRoutingModeFelix {
+		return RouteProtoFelix
+	}
+	return RouteProtoBIRD
+}
```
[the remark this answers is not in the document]
mazdakn: hmm we use Operator for our e2e tests, but make sense to just read `FelixConfiguration` which should exists all the time. I'll switch to it.

e2e/pkg/tests/networking/cluster_routes.go
```diff
@@ -0,0 +1,139 @@
...
+
+			poolName := utils.GenerateRandomName("route-owner-pool")
+			pool := v3.NewIPPool()
+			pool.Name = poolName
+			pool.Spec.CIDR = routeOwnerPoolCIDR
+			pool.Spec.NATOutgoing = true
+			pool.Spec.BlockSize = 28
+			pool.Spec.DisableBGPExport = false
+			poolMutator(pool)
+			Expect(cli.Create(context.Background(), pool)).To(Succeed(), "Error creating IP pool")
```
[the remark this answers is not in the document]
caseydavenport: I think this one is right - we should use checks up front in the tests to fail explicitly when an environmental requirement isn't met.

e2e/pkg/tests/networking/routes.go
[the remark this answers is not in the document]
caseydavenport: I think it's fine to call this one operator-only - just make sure it's tagged as such so it gets skipped correctly.

e2e/pkg/tests/networking/cluster_routes.go
```diff
@@ -0,0 +1,139 @@
...
+
+// routeOwnerPoolPrefix is the substring matcher used against `ip route`
+// destinations. Any IPAM block carved out of routeOwnerPoolCIDR will start
+// with this prefix.
+const routeOwnerPoolPrefix = "203.0.113."
+
+var _ = describe.CalicoDescribe(
+	describe.WithTeam(describe.Core),
+	describe.WithFeature("IPPool"),
+	describe.WithFeature("IPIP"),
```
caseydavenport: I wonder about this - the feature we're testing here is "ClusterRoutingMode" - we're not actually testing IP pools / IPIP here - we're just using them to test that the routes are programmed correctly? 

To be honest, I am not sure if this test is really carrying its weight as an e2e - given the other test below also includes a proto check, and presumably we have FVs already that check this? 

But I could be swayed either way - it should run pretty fast at least and is so minimalist that it likely won't ever flake.

caseydavenport (approved):
Couple of minor comments, but otherwise LGTM - up to you if you want to include the test or leave it out.

e2e/pkg/tests/networking/cluster_routes.go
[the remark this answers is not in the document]
mazdakn: Switch to checking `ProgramClusterRoutes` in `FelixConfiguration` which always exists.

PR #12772, opened 2026-05-14
Remove redundant processing of node address

mazdakn:
Moving retrieving address from a Node resource and the fallback logic to a separate function. It's also used in [ipsec binding](https://github.com/tigera/calico-private/blob/545613c9a4712d7a4a6524089b395c7d624bdf25/felix/calc/ipsec_binding_calculator.go#L443) in calc graph. It was needed as part of this pick PR: https://github.com/tigera/calico-private/pull/11915

**Release note:**
```release-note
TBD
```

### DIFF
--- a/felix/calc/dataplane_passthru.go
+++ b/felix/calc/dataplane_passthru.go
@@ -33,9 +33,8 @@ import (
 // with the rest of the dataplane API.
 //
 // HostMetadataUpdate is sourced from the v3 Node resource: BGP IPv4/IPv6 plus labels and ASN.
-// IPv4 falls back to the Node's address list (Internal preferred, External as backup) whenever
-// BGP doesn't supply one — mirroring the old felixnodeprocessor logic that previously fed the
-// retired HostIPKey channel. IPv6 falls back to the address list only when BGP is absent.
+// IPv4/IPv6 falls back to the Node's address list (Internal preferred, External as backup) whenever
+// BGP doesn't supply one.
 type DataplanePassthru struct {
 	callbacks passthruCallbacks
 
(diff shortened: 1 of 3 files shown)


felix/calc/dataplane_passthru.go
```diff
@@ -132,62 +132,73 @@ func (h *DataplanePassthru) processKindNode(key model.ResourceKey, update api.Up
...
 		}
-		if addr := bgpSpec.IPv6Address; addr != "" {
+		if addr != "" {
 			if s, ok := parseCIDR(addr); ok {
-				info.ip6Addr = s
-			} else {
-				logrus.WithField("addr", addr).Warn("Ignoring Node BGP IPv6Address: not a CIDR")
+				return s
 			}
+			logrus.WithField("addr", addr).Warn("Ignoring Node BGP address: not a CIDR")
```
caseydavenport: Might be worth a context logger in here to include the ipVersion / ipv6Support fields throughout?

felix/calc/dataplane_passthru.go
```diff
@@ -132,62 +132,73 @@ func (h *DataplanePassthru) processKindNode(key model.ResourceKey, update api.Up
...
-			info.asnumber = node.Spec.BGP.ASNumber.String()
-		}
 	}
-	// If BGP didn't supply an IPv4 address, fall back to the Node's address
+
+	// Fallback path.
+	if ipVersion == 6 && (!ipv6Support || bgpSpec != nil) {
+		// IPv6 fallback only fires when BGP is absent, preserving the historical
+		// asymmetry: IPv4 had a backup channel (HostIPKey) but IPv6 did not, so
+		// the Address list was only consulted for IPv6 when BGP itself was nil.
```
caseydavenport: Huh, I don't really remember / know why this is the case - is it still relevant today actually? 

Now that we have a combined v4/v6 struct, do we have the flexibility to clean this up?

mazdakn: Removed this part.

mazdakn: I also updated it parse both IPv4 and IPv6 unconditionally. WDYT?

caseydavenport (approved):

PR #12783, opened 2026-05-15
Add lib/datastructures/hashring consistent hash ring

fasaxc:
Adds a generic consistent hash ring for use cases like cluster-wide service-IP advertisement ownership (each node independently decides which IPs it owns; all nodes agree). Lives in a new `lib/datastructures` module — sized to be the home for future hash-aware data structures that share the xxh3 dependency, while keeping `lib/std` strictly stdlib-only.

## API

```go
r := hashring.New[string](
    hashring.WithReplicas(1),   // virtual nodes per member
    hashring.WithProbes(10),    // multi-probe samples per Lookup
    hashring.WithHash(myHash),  // optional; default is XXH3-64
)
r.Insert(key, value)            // O(R)
r.Remove(key)                   // O(1); swept lazily on next Lookup
v, ok := r.Lookup(queryKey)     // O(P · log(N·R))
```

`Ring[V any]` is keyed by string with a generic value type. Insert/Remove/Len/Lookup, not safe for concurrent use (callers wrap if needed). All three options have sane defaults; bare `New[V]()` is bare CH.

## Two orthogonal imbalance knobs

`WithReplicas` (R) and `WithProbes` (P) both reduce imbalance; the budget is roughly `R · P`. They're independent levers on a memory ↔ Lookup-CPU trade-off:

- `R=100, P=1` — classic virtual-nodes. Single binary search per Lookup; ring is `N · R` entries (heavy at large N).
- `R=1, P=10` — pure multi-probe (Appleton & O'Reilly, 2015). Ring is just N entries; Lookup hashes the query P times and picks the probe whose nearest entry is closest.
- `R=10, P=10` etc — hybrids in between.

At K8s scale (N=50000) the difference is ~120 MB vs ~1.2 MB of ring memory for the same imbalance budget; this is why the package exposes both knobs and doesn't bake in v-nodes by default.

## Default hash: XXH3-64

`github.com/zeebo/xxh3`. Earlier iterations defaulted to FNV-1a, but the tester reported that with 100 nodes / 1024 consecutive IPs FNV's busiest node owned 33 keys and two nodes owned zero — a hash-quality issue, not a probes issue (probes don't fix FNV; weak avalanche between adjacent salts causes virtual nodes for the same member to cluster). XXH3 reaches the Poisson sampling floor of uniform distribution. `WithHash` swaps it for any deterministic `func([]byte) uint64`.

## Deferred Remove

`Remove(key)` is O(1) — adds the key to a `deletedKeys` set. The entry slice and members map are swept in one pass at the next `Lookup`. Bulk removals drop from O(K·N) to O(K) + one O(N) sweep. Re-inserting a queued-for-removal key un-queues it and preserves the original ring positions (no double-counting).

## What's in the PR

- `lib/datastructures/go.mod` — new module, depends on `github.com/zeebo/xxh3 v1.1.0` only.
- `lib/datastructures/hashring/hashring.go` — the implementation (~230 LOC).
- `lib/datastructures/hashring/hashring_test.go` — parametric unit tests, fuzz, benchmarks (~700 LOC).
- `lib/datastructures/hashring/cmd/hashring-eval/main.go` — CLI that prints distribution and per-op timing across an R/P/N sweep, with a Poisson-floor reference row so the reader can see when an algorithm config has hit the theoretical limit.

## Test plan

- [x] Parametric unit tests across `(R, P)` configs: empty ring, single member, determinism, idempotent Insert, exact Remove with deferred sweep, sort-deferral correctness, reinsert stability, ~1/N consistent-hashing key reassignment on removal, distribution sanity, hash-collision tiebreak, generic value type
- [x] Explicit ring-wrap-around tests using stubbed hashes that pin known positions, including multi-probe wrap-distance comparison both directions
- [x] Deferred-delete contract: stale entries cleared on next Lookup; re-Insert un-queues without entry churn
- [x] `FuzzRing` over random Insert/Remove/Lookup with `(R, P)` from fuzz input; invariants checked after every op (including `deletedKeys ⊆ members`, `len(entries) == len(members)·R`)
- [x] Benchmarks: `BenchmarkLookup` (hot path), `BenchmarkFirstLookup` (sort-included), `BenchmarkRemove` (O(1)), `BenchmarkBulkRemoveThenLookup` (amortised sweep)
- [x] `cmd/hashring-eval` reports distribution + timing across R × P × N sweep against the Poisson floor

```release-note
None
```


lib/std/hashring/cmd/hashring-eval/main.go
```diff
@@ -0,0 +1,125 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// hashring-eval reports load distribution and lookup cost for the
+// hashring package across a sweep of (members, replicas, probes)
+// configurations. Run with `go run ./lib/std/hashring/cmd/hashring-eval`.
```
MichalFupso:
```suggestion
// configurations. Run with `go run ./lib/std/hashring/cmd/hashring-eval/main.go`.
```

lib/std/hashring/hashring.go
```diff
@@ -0,0 +1,258 @@
...
+// newDefaultHash returns a Hash backed by a per-Ring stdlib
+// hash/fnv.New64a hasher. Each Ring gets its own hasher so the
+// closure has no shared state (no goroutine-safety pitfall beyond
+// the Ring itself).
+func newDefaultHash() Hash {
+	h := fnv.New64a()
+	return func(b []byte) uint64 {
+		h.Reset()
+		h.Write(b)
+		return h.Sum64()
```
MichalFupso: Testing with 100 nodes and 1024 continuous IPs with Sum64 the busiest node managed 33 ips and there were two nodes that have not managed a single ip. With sha256 (first 64bits), the busiest node had 17 ips and each node had at least on ip. What do you think about switching to sha256 for a better distribution, there is a risk of conflict with only 64bits, so another option would be to add a mixing finalizer to Sum64

fasaxc: Good catch! Sounds like a known problem with FNV :facepalm: 

I researched best practice for hash rings and it sounds like XXHash is the front-runner for non-crypto usage; it's orders of magnitude faster than SHA256 but it still has excellent avalanche properties.  I'll pull that in (rather than doing rolling my own hash with unknown edge cases).

lib/datastructures/hashring/cmd/hashring-eval/main.go
```diff
@@ -0,0 +1,236 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// hashring-eval reports load distribution and lookup cost for the
+// hashring package across a sweep of (members, replicas, probes)
+// configurations. Run with `go run ./lib/datastructures/hashring/cmd/hashring-eval`.
```
fasaxc: This command doesn't work, have to run from the right module.

fasaxc: Fixed in 8c710dfb58 — doc now shows the `cd lib/datastructures && go run ./hashring/cmd/hashring-eval` form.

MichalFupso (approved):

PR #12796, opened 2026-05-19
felix: add VXLANSrcPortMin/VXLANSrcPortMax to control VXLAN source port range

john7doe:
## What this does

Adds `vxlanSrcPortMin` and `vxlanSrcPortMax` to `FelixConfiguration` so operators can constrain the UDP source ports used for outgoing VXLAN traffic in both the kernel and eBPF dataplanes. This provides a control for environments affected by Azure network flow limits, similar to the problem described in [this Linux change](https://github.com/torvalds/linux/commit/e1f95b1992b8cc31e37505787806918b0447909b).

## How

**Apply the range to both dataplanes.** The kernel VXLAN device receives netlink `PortLow` and `PortHigh`. The eBPF encapsulation path receives the same bounds through TC globals and maps its per-flow port hash into the inclusive range with `min + hash % (max - min + 1)`.

**Validate the bounds in Felix.** Both fields default to zero. A configured range requires both bounds in `1..65535` and `min < max`. Single-port ranges remain rejected for consistency with the kernel VXLAN device. `VXLANPort` continues to control the destination port.

**Exercise the production source-port selector.** The BPF unit harness loads the packet's UDP header into its scratch buffer and calls the same inline selector as `tc.c`. Coverage includes the unchanged default hash, a narrow range and both inclusive endpoints. The kernel-device test uses the current logging package.

**Keep the generated schemas consistent.** API artifacts and configuration documentation have been regenerated. The v1beta1 manifest bundles now carry the same source-port fields as the other installation manifests.

## Notes

Removing the configured bounds does not by itself reset a range already present on a kernel VXLAN device, preserving the existing unset/don't-care comparison behavior.

Focused BPF and VXLAN tests and the precompiled-program verifier sweep pass locally on Linux/arm64. A fresh Semaphore run is still required. After merge, confirm the configured source-port range on a VXLAN canary before rollout.

**Release note.** Added optional `vxlanSrcPortMin` and `vxlanSrcPortMax` settings to constrain VXLAN UDP source ports.

**AI assistance.** This PR was written in part with the assistance of generative AI.


felix/docs/config-params.md
```diff
@@ -2553,6 +2553,37 @@ The UDP port number to use for VXLAN traffic.
 | `FelixConfiguration` schema | Integer |
 | Default value (YAML) | `4789` |
 
+### `VXLANPortMax` (config file) / `vxlanPortMax` (YAML)
```
matthewdupre: I think it's a little confusing to have the parameter name be VXLANPortMin/Max, since there's already a VXLANPort (which controls the dest port).  I think VXLANSourcePortMin/Max (or Src, don't mind) would be clearer?

felix/design/bpf-encap-fragments-icmp.md
```diff
@@ -87,6 +87,50 @@ interface. These routes are still needed because:
...
+  strictly `min < max`, the BPF code remaps the hash into
+  `[min, max]` via `min + (hash % (max - min + 1))`. The `+1` keeps
+  the range *inclusive* of the upper bound and matches the Linux
+  kernel's `udp_flow_src_port()` helper, which the non-BPF
+  dataplane goes through via the netlink-managed VXLAN device.
+  Dropping the `+1` would silently lose the top port on the BPF
+  side and diverge from the kernel-managed device. The macros are
+  populated from the `vxlan_port_min` / `vxlan_port_max` globals
+  in `cali_tc_global_data` (see `felix/bpf-gpl/globals.h`).
+- Equal ports (`min == max`) are rejected by Felix's config
```
matthewdupre: This second bullet is redundant - it explains why min<max is enforced above, but it only matters when you're changing this so a comment on the code is sufficient.

felix/docs/config-params.md
matthewdupre, line 2580: Would be nice to specify "UDP destination port" now we've added source port settings.  I'd leave the name of the option the same for back compatibility though.

felix/bpf-gpl/tc.c
```diff
@@ -1274,9 +1274,26 @@ static CALI_BPF_INLINE enum do_nat_res do_nat(struct cali_tc_ctx *ctx,
...
+	 * Linux kernel's `udp_flow_src_port()`, which the non-BPF dataplane
+	 * goes through via the netlink-managed device; dropping the `+1`
+	 * here would silently lose the top port. The strict `>` check
+	 * mirrors Felix's config validator: a degenerate range (min == max)
+	 * falls through to the raw hash rather than pinning every packet to
+	 * a single port.
 	 */
 	__u16 vxlan_src_port = STATE->sport ^ STATE->dport;
 
+	if (VXLAN_PORT_MIN != 0 && VXLAN_PORT_MAX != 0 && VXLAN_PORT_MAX > VXLAN_PORT_MIN) {
```
matthewdupre: I'd prefer to push this kind of validation further up rather than doing it in a BPF program which is harder to work on and can't log a warning as conveniently.

felix/design/bpf-encap-fragments-icmp.md
```diff
@@ -87,6 +87,50 @@ interface. These routes are still needed because:
...
+  `low == high` and falls back to its default. The BPF guard in
+  `tc.c` uses a strict `>` to mirror the validator: if a
+  degenerate range somehow slips through, BPF falls back to the
+  raw `sport ^ dport` hash rather than pinning every packet to
+  one port.
+- The same range is also pushed onto the kernel-managed VXLAN
+  device as `PortLow` / `PortHigh` (`vxlan_mgr.go`) so that the
+  iptables/nftables dataplane behaves consistently. The
+  `vxlanLinksIncompat` check only fires when both sides report a
+  non-zero range — the kernel sometimes echoes back a default
```
matthewdupre: I'm curious about this "sometimes"?  What did the kernel report and do you know when it happens (e.g. is it a timing window after making a change to the range?)?

felix/design/bpf-encap-fragments-icmp.md
```diff
@@ -87,6 +87,50 @@ interface. These routes are still needed because:
...
+  range that we did not configure, and treating that as a
+  mismatch would needlessly recreate the device.
+- Leaving either field unset (zero) preserves the historical
```
matthewdupre: Remove - we don't need to duplicate the config documentation in the design file

matthewdupre (commented):
Thanks for submitting this - seems useful!  I have a few suggestions around naming and design documentation primarily.

felix/docs/config-params.md
john7doe: Changing to VXLANSrcPortMin/Max

felix/design/bpf-encap-fragments-icmp.md
john7doe: removed

felix/docs/config-params.md
john7doe: changed to UDP destination port

felix/bpf-gpl/tc.c
john7doe: This is not as much a validation as it is checking the port range should be clamped between [VXLAN_PORT_MIN;VXLAN_PORT_MAX] or not. I've removed the VXLAN_SRC_PORT_MAX > VXLAN_SRC_PORT_MIN part and let the BPF program rely on validation ensuring that invariant.

felix/design/bpf-encap-fragments-icmp.md
john7doe: That was my interpretation/guess of the exiting guards for the other fields in `vxlanLinksIncompat`. Looking again I think its there to implement the unset -> don't care semantics. Updated here and removed comment in `vxlanLinksIncompat`

felix/design/bpf-encap-fragments-icmp.md
john7doe: removed

john7doe:
@matthewdupre thanks for the review. I _think_ i've addressed them all.

felix/docs/config-params.md
```diff
@@ -2553,34 +2553,34 @@ The UDP port number to use for VXLAN traffic.
...
-### `VXLANPortMin` (config file) / `vxlanPortMin` (YAML)
+### `VXLANSrcPortMin` (config file) / `vxlanSrcPortMin` (YAML)
 
 The lower bound (inclusive) of the UDP source port range used by Felix for outgoing
-VXLAN-encapsulated traffic. If both VXLANPortMin and VXLANPortMax are set, the kernel VXLAN tunnel
-device and the eBPF dataplane's VXLAN encap will pick source ports within [VXLANPortMin, VXLANPortMax].
-VXLANPortMin must be strictly less than VXLANPortMax.
+VXLAN-encapsulated traffic. If both VXLANSrcPortMin and VXLANSrcPortMax are set, the kernel VXLAN tunnel
+device and the eBPF dataplane's VXLAN encap will pick source ports within [VXLANSrcPortMin, VXLANSrcPortMax].
+VXLANSrcPortMin must be strictly less than VXLANSrcPortMax.
```
matthewdupre, line 2610: Since the source port max is inclusive, I think min==max should be a valid configuration which will always use that single permitted port.

felix/design/bpf-encap-fragments-icmp.md
```diff
@@ -101,35 +101,23 @@ firewall ACLs, ECMP-hash pinning, or to align with the kernel
 VXLAN driver's `srcport min max`), `FelixConfiguration` exposes
 two optional fields:
 
-- `vxlanPortMin` and `vxlanPortMax` — when both are non-zero and
+- `vxlanSrcPortMin` and `vxlanSrcPortMax` — when both are non-zero and
   strictly `min < max`, the BPF code remaps the hash into
```
matthewdupre, line 105: (Allowing min==max as commented below will mean this needs to be weakened)

matthewdupre (commented):
I'm happy with this with this one last change (allowing min==max as the field documentation suggests).  Thanks!

felix/docs/config-params.md
john7doe: > which will always use that single permitted port.

I recalled that during my testing a single port did not work for us, so I dig a bit deeper because of your comment, and while that would be the result for the the eBPF dataplane's VXLAN, it would not be that for the the kernel VXLAN tunnel
devices, instead it would overwrites your port with the system ephemeral range:

a single port would pass validation: https://github.com/torvalds/linux/blob/master/drivers/net/vxlan/vxlan_core.c#L3494, but when later calling `udp_flow_src_port` it would fallback to the default range: https://github.com/torvalds/linux/blob/master/include/net/udp.h#L349-L352 due to the min >= max comparison.

So if we want to support a single port the validation would have to factor in the dataplane chosen. I _think_ it would just be a matter of checking ` bpfEnabled: true` in the config? 

Please let me know what you prefer.

matthewdupre: Oh interesting - I'd prefer consistency between dataplanes, so let's leave min==max disallowed.

matthewdupre (approved):

john7doe:
@matthewdupre had to fix a test that was failing on master.

matthewdupre:
Looks like Semaphore got confused and didn't rerun; kicked it

matthewdupre:
Ugh I really don't understand what's wrong with Semaphore CI triggering on this PR

matthewdupre:
@john7doe Looks like the new tests are failing in BPF mode - is that something you could look at please?  Sorry for all the CI hassle.

john7doe:
Sorry, this completely slipped of my radar.

@matthewdupre I've merged master and fixed the BPF test. The standalone harness read `udp_hdr(ctx)` without loading the packet's UDP header into the scratch buffer, so it hashed leftover state.

The test now fills that buffer and calls the same inline source-port selector as the encapsulation path. It covers the default hash, a configured range and both inclusive endpoints. The production mapping and the agreed `min < max` validation are unchanged.

The focused BPF tests, precompiled-program verifier sweep and VXLAN configuration/device tests pass locally on Linux/arm64. The generated manifests are refreshed too. Could you approve a fresh CI run?

PR #12799, opened 2026-05-19
Collect kind cluster diags on e2e failure

caseydavenport:
Adds a small `.semaphore/collect-kind-diags` script and wires it into the e2e (KinD) block as an `on_fail` epilogue. On failure it dumps:

- `kubectl get nodes -o wide`
- `kubectl get pods -A -o wide`
- `kubectl get events -A` (sorted)
- `kubectl get apiservices`
- `kubectl describe pod` for any pod not in Running/Completed
- Current and previous logs for pods in `calico-system`, `tigera-operator`, `calico-apiserver`

Everything lands in `artifacts/kind-diags.tar.zstd` so the existing `publish-artifacts` epilogue picks it up. The script is best-effort: a missing kubeconfig or torn-down cluster is treated as a no-op.

Motivation: e2e jobs intermittently fail with calico-apiserver unavailable mid-suite and the CI logs alone don't tell us whether the apiserver pod restarted, panicked, or was OOM-killed. The `--previous` log capture in particular is meant to catch a crash before the cluster gets torn down.

### Release Note

\`\`\`release-note
None
\`\`\`

### DIFF
--- a/.semaphore/collect-kind-diags
+++ b/.semaphore/collect-kind-diags
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2026 Tigera, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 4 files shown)


.semaphore/collect-kind-diags
```diff
@@ -0,0 +1,82 @@
...
+        > "$diags_dir/describe-${ns}-${pod}.txt" 2>&1 || true
+    done
+
+# Collect logs from the control-plane components most likely to be relevant
+# when e2e tests fail. Grabs both current and previous logs so we can see
+# what a restart killed. Missing namespaces are tolerated.
+namespaces=(
+  "calico-system"
+  "tigera-operator"
+  "calico-apiserver"
```
caseydavenport:
```suggestion
```

caseydavenport:
Addressed Copilot review:
- Drop the size-based filter on `*.previous.log` files and check kubectl's exit code instead, so short-but-real previous logs aren't deleted.
- Make the tar/zstd step best-effort - on failure we keep the raw directory so publish-artifacts can still upload it.

coutinhop (approved):

PR #12807, opened 2026-05-20
Add calico-resync tool into our packaging

nelljerram:
(Release noting and docs requirement are covered by #12658 )


danudey (approved):
lgtm!

PR #12808, opened 2026-05-20
fix: handle combined fsnotify create events

immanuwell:
Fixes a small CNI wait edge case.

fsnotify Op is a bitmask. Current code checks e.Op == fsnotify.Create, so an event like Create|Write is ignored. That can make the status-file wait run until timeout even though the file event already showed up. Tiny footgun, but real.

Repro:
- with old code, the new unit test injects fsnotify.Create|fsnotify.Write
- go test ./cni-plugin/pkg/wait fails because found stays false
- with this change it passes, because we use e.Has(fsnotify.Create)

No linked issue found, afaict. Related PRs: #8469, #12609.

Testing:
- go test ./cni-plugin/pkg/wait
- go test ./cni-plugin/pkg/wait ./cni-plugin/internal/pkg/utils

**Release note:**
```release-note
Fix CNI endpoint status-file waiting when fsnotify reports create with another operation.
```

### DIFF
--- a/cni-plugin/pkg/wait/wait.go
+++ b/cni-plugin/pkg/wait/wait.go
@@ -65,7 +65,7 @@ func ForEndpointReadyWithTimeout(policyDir string, endpoint *internalapi.Workloa
 func waitUntilFileExists(ctx context.Context, directory, filename string) error {
 	log := logrus.WithFields(logrus.Fields{
 		"directory":   directory,
-		"disiredFile": filename,
+		"desiredFile": filename,
 	})
 
 	retryInterval := 500 * time.Millisecond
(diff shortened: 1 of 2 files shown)


skoryk-oleksandr (changes requested):
@immanuwell  Thanks for catching this. Fix looks good to me. 
One small ask: could you add a counterpart test that injects a Write-only event, with no Create bit, and asserts that the watcher does not return true?
That would lock down the other side of the contract and prevent a future regression where someone accidentally over-broadens the match.

immanuwell:
@skoryk-oleksandr done, addressed all required suggestions, PTAL

PR #12810, opened 2026-05-21
[BPF] Fix node-unreachable on Felix restart with no Node

### LINKED ISSUE #12642: BPF dataplane: kubectl delete node without graceful Calico shutdown causes node to lose connectivity

tomastigera:
## Summary

Fixes #12642 — in BPF mode, after `kubectl delete node <self>` followed by a Felix restart, the node loses all network connectivity (no SSH, no etcd/apiserver) and cannot recover on its own.

Supersedes #12656 (cleanup-gate band-aid) and #12688 (same approach as this PR but without an FV).

## Mechanism (verified via FV, see "Test" section)

1. New Felix runs `repinJumpMaps()` to move the previous Felix's jump maps under `old_jumps/<tmp>/` and creates fresh empty maps at the original pin paths. The previous Felix's preambles (still attached to host interfaces) reference the OLD maps' kernel IDs.
2. Without a Node, `wepApplyPolicyToDirection` / `attachDataIfaceProgram` short-circuit with `"unknown host IP"`; new preambles are never installed.
3. `doApplyPolicy*` clears the per-family `"unknown host IP"` error at the bottom (the dual-stack "don't fail v4 if only v6 is enabled" trick swallows it), so `dirtyIfaceNames` empties.
4. With `dirtyIfaceNames.Len() == 0`, the cleanup gate fires `os.RemoveAll(old_jumps)`. The OLD maps lose their pin; program references from BPF programs alone do not keep `prog_array` entries alive, so the kernel reclaims the slots.
5. The OLD preambles are still attached, but every `bpf_tail_call` into the now-empty maps fails. Every packet through `eth0` is dropped.

## Approach

The host IP is meaningfully consumed by very few BPF features:

- VXLAN return-path encap source (`tc.c:975, 1087, 1200` — `STATE->ip_src = HOST_IP`).
- An IPv4 RPF strict-mode link-local check (`rpf.h:134`) — naturally evaluates false when `HOST_IP=0`.
- A debug log in `vxlan_attempt_decap`. The decap decision itself uses a route-table lookup, not `HOST_IP`.

Everything else — policy, conntrack, NAT, ICMP, kube-proxy replacement, non-VXLAN forwarding — works fine without `HOST_IP`.

**Stop using "no host IP" as a hard gate.** Programs now attach as normal with `HOST_IP=0` as a sentinel. The Go-side already handled `nil` host IP in the globals plumbing (`copy(dst, (nil net.IP).To4())` is a no-op, so `globalData.HostIPv*` stays at zero); the gates were the only thing short-circuiting before that.

The three VXLAN encap-source sites in `tc.c` now check `ip_void(HOST_IP)` and, if true, drop the packet with a new `CALI_REASON_NO_HOST_IP` / `DroppedNoHostIP` counter (slot 26 in the counters map; `RESERVED1` is preserved for downstream use).

### `lastSeenHostIP`: unified "host IP gone" handling

The BPF endpoint manager caches the last-known host IP per family as `lastSeenHostIP`. All three "no host IP" inputs from the calc graph — `HostMetadataRemove` (Node deleted), `HostMetadataUpdate` with empty `Ipv4Addr` / `Ipv6Addr` (Node alive, IP patched out), and parse failure — go through one path (`updateOurHostIP(nil, …)`) that:

- **Preserves** the cached `lastSeenHostIP` so pinned BPF programs continue running with a usable HOST_IP for the few features that need it (VXLAN encap source, RPF link-local check).
- Updates the `BPFHostIP` readiness reporter so the missing IP is visible to operators.
- Does **not** mark any interface dirty — no mid-flight re-attach in the "host IP gone" path.

This matters most around `HostMetadataRemove`: when the local Node is deleted, kubelet shortly terminates the calico-node pod via SIGTERM, so a full mark-all-ifaces-dirty + re-attach pass in that window risks Felix being killed mid-apply and leaving the BPF dataplane in an inconsistent state (preambles tail-calling into freed jump-map slots). Validated on a real cluster: triggering the rebuild on `HostMetadataRemove` broke connectivity within ~4 seconds of `kubectl delete node`; the no-rebuild path holds connectivity intact.

The Felix-restart-with-no-Node case (the original bug) is handled by the gate removal above: a fresh Felix attaches programs with `HOST_IP=0` from the start, no mid-flight rebuild involved. Once a host IP arrives, `updateOurHostIP` marks everything dirty so programs re-attach with the real `HOST_IP`.

### Readiness

A new `BPFHostIP` health reporter reflects datastore state independently of the dataplane cache. Per-family presence is tracked; the reporter goes degraded with `Detail: "Host IP unknown: IPv4 unknown since <RFC3339 timestamp>..."` while a configured family is missing.

- IPv6 disabled: requires an IPv4 host IP.
- IPv6 enabled (always also implies v4 enabled — v4 cannot be disabled): requires at least one of v4 or v6.

After Node delete the dataplane keeps using the cached IP (so connectivity holds across a Felix restart) but `/readiness` returns `503`, so operators see the condition rather than discovering it weeks later after an unrelated restart.

Co-authored with Shaun Crampton (most of the C/Go fix is from #12688).

## Test

`felix/fv/bpf_node_delete_test.go` adds three scenarios — each also asserts the new `BPFHostIP` readiness signal:

- **Node delete only**: deletes the Calico Node and Consistently probes host SSH over 10s. The pinned BPF programs keep running, connectivity holds throughout. Asserts `/readiness` goes 503 with `BPFHostIP ... IPv4 unknown since <timestamp>` in the body.
- **Node delete + Felix restart + recovery**: deletes the Node, sends SIGHUP to Felix, and immediately starts a 15s Consistently — without waiting for the new PID. The probe loop spans Felix's exit, restart, startup, `repinJumpMaps`, and first apply pass. Asserts `/readiness` is 503 (initial "Host IP not yet known" state) after restart, re-adds the Node, watches for the `"Host IP changed"` log signal, Consistently probes for another 10s through the recovery re-attach, and asserts `/readiness` returns to 200.
- **IP removed from Node while Node still exists**: clears the Calico Node's `Spec.BGP` AND `Spec.Addresses` (both must be cleared — the calc graph's `extractNodeAddress` falls back from BGP to `Spec.Addresses`). Asserts that failsafe SSH keeps working (programs re-attach with `HOST_IP=0`), readiness goes 503 with the per-family `"IPv4 unknown since <timestamp>"` detail, and recovers to 200 when the Node addresses are restored.

`felix/fv/bpf_dual_stack_test.go` repurposes the `"with IPv*N addresses only"` tests to validate the new behaviour and adds a new `"with IPv6 host IP removed from one node only"` test that codifies the cross-node NodePort breakage when a node lacks a family's host IP — exercises the encap-source drop site via an external client, asserts both the connectivity drop and the `DroppedNoHostIP` counter increment.

**Empirically validated**:

| Run | Test 1 | Test 2 | Test 3 |
|---|---|---|---|
| Master (no fix) + test 2 | PASS | **FAIL @ 4.568s** — BPF trace shows `tc_preamble failed to call main 37`, exactly the bug mechanism | — |
| This PR (with rebuild on `HostMetadataRemove`) — earlier iteration | PASS (FV) | PASS (FV) — but broke real cluster | — |
| This PR (final) | PASS | PASS | PASS |

**Cluster validation** (banzai GCP kubeadm, kernel 6.8, BPF mode, 4 nodes):

Used `kubectl delete node <test-node>` to drive the calc graph `HostMetadataRemove` event. With an earlier iteration of this PR that included a rebuild on `HostMetadataRemove`, all traffic to the affected node dropped within ~4 seconds of the delete — the rebuild pass races kubelet's SIGTERM and leaves the BPF state inconsistent. With the rebuild dropped, `kubectl delete node` + pod GC consistently leaves connectivity intact: pinned BPF programs continue running with cached `HOST_IP`, and 120 consecutive 1-Hz probes from a peer node passed without a single drop. Independently exercised the "IPv4 removed but Node still alive" path on the same cluster: 90 probes passed throughout, dataplane cleared and re-attached with `HOST_IP=0`, traffic recovered when the address was restored.

**Release note:**
```release-note
Fixes a bug in the eBPF dataplane in which deleting and restoring the local Node resource and restarting Felix could leave the node unable to handle network traffic.
```

## Behavior change: per-family attach is now driven by config, not by host IP

Pre-PR, BPF programs for a given IP family (v4/v6) were skipped at attach time when that family's host IP was nil — the `"unknown host IP"` early-returns in `wepApplyPolicyToDirection` / `attachDataIfaceProgram` short-circuited the attach. This PR removes those gates (they're the root of the bug) and lets programs attach with `HOST_IP=0`.

Practical consequence: on a Felix configured for dual-stack (`BPFIpv6Enabled=true`) but running on a host that only has an IPv4 address, v6 BPF programs are now attached too (with `HOST_IPv6=0`). Pre-PR they were not.

This is the right behavior. Felix can't distinguish pure-IPv6, dual-stack, and "v6-enabled-but-host-has-no-v6" configurations from each other — they all look identical to the dataplane. The "v6 enabled" config flag is the operator's signal that v6 attach is wanted; the previous per-host-IP gate was second-guessing that. For the `BPFIpv6Enabled=false` case (`m.v6 == nil` from construction), no v6 programs are attached, same as before.

Cost: a few extra BPF programs attached on misconfigured hosts (v6 enabled in Felix but no v6 address present). Negligible at runtime — they only execute on v6 packets, which never arrive on such hosts.


tomastigera:
Pushed test updates addressing CI failures from the first run:

1. **`felix/bpf/ut/attach_test.go`** — `TestAttach` now expects `Family:6 Type:"host"` to be present in the loaded programs (and `FlgIPv6Ready` in the iface state map) when IPv6 is enabled, regardless of whether the v6 host IP has been provided. Matches the new attach semantic documented in the "Behavior change" section of the PR description.

2. **`felix/fv/bpf_dual_stack_test.go`** — repurposed the two `"with IPv*N addresses only"` tests and added one new test:
   - `"with IPv4 addresses only"` / `"with IPv6 addresses only"`: same-host other-family connectivity is now expected to **work** (BPF attaches with `HOST_IP=0` and policy/routing applies). Cross-host other-family connectivity still expected to **fail**, but for the *right* reason now (VXLAN encap source drops with `DroppedNoHostIP`, not "programs weren't loaded"). The pre-PR `None` expectation was an artifact of the bug, not an invariant.
   - Deleted `"Should connect to w[0][0] using IPv6 after IPv6 host IP is added"` — it was testing the add-v6-IP transition that's now exercised by the new `updateHostIP`/`HostMetadataRemove` re-attach logic; nothing distinct left.
   - Added `"with IPv6 host IP removed from one node only"` / `"v6 NodePort via the no-v6 node fails for external clients with DroppedNoHostIP"`. Codifies the user-visible degradation under the new semantic: cross-node NodePort traffic from external clients to a node missing the family's host IP fails, because BPF can't VXLAN-encap the return path. Asserts both the connectivity drop and the `DroppedNoHostIP` counter increment. Restricted to `!ctlbEnabled` because the BPF kube-proxy on the no-v6 node fails to initialize ("proxy is not initialized") when CTLB is on and a family's host IP is missing — a pre-existing fragility unrelated to this PR that would mask the encap-drop behaviour the test is trying to codify.

Empirically validated locally:
- `make FOCUS="^TestAttach$" ut-bpf`: 21 tests, all pass.
- `make fv-bpf GINKGO_FOCUS="dual stack basic in-cluster connectivity"`: 14 tests, all pass.
- `make fv-bpf GINKGO_FOCUS="Felix node delete should preserve host connectivity"`: 2 tests, all pass (the original repro test from this PR).

felix/fv/bpf_node_delete_test.go
```diff
@@ -0,0 +1,245 @@
...
+			externalClient = infrastructure.RunExtClient(infra, "ext-client")
+
+			cc = &Checker{}
+
+			// Wait for BPF programs to come up before doing the connectivity
+			// check, otherwise the first probe races the dataplane.
+			ensureBPFProgramsAttached(tc.Felixes[0])
+		})
+
+		JustAfterEach(func() {
```
tomastigera: there is some machinery for this in the fv//infra tha bpf_test.go uses so it does not need to be copy-pasted

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -879,40 +879,49 @@ func (m *bpfEndpointManager) withIface(ifaceName string, fn func(iface *bpfInter
...
-		// but taking it now makes us robust to refactoring.
-		m.ifacesLock.Lock()
-		for ifaceName := range m.nameToIface {
-			m.withIface(ifaceName, func(iface *bpfInterface) (forceDirty bool) {
-				iface.dpState.v4Readiness = ifaceNotReady
-				iface.dpState.v6Readiness = ifaceNotReady
-				return true
-			})
+		if ip == nil {
+			logrus.WithField("ipAddr", ipAddr).Warn("Cannot parse hostip, no change applied")
```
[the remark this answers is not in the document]
tomastigera: Done in dce3ebe — log message now uses `"Cannot parse host IP"` and the field set includes `ipFamily`.

felix/fv/bpf_node_delete_test.go
tomastigera: Done in dce3ebe — dropped the JustAfterEach. `infra.DumpErrorData()` in AfterEach (via `felix/fv/infrastructure/diagnostics.go`) already covers the BPF dumps.

felix/fv/bpf_node_delete_test.go
```diff
@@ -0,0 +1,219 @@
...
+				hostW.Stop()
+			}
+			if externalClient != nil {
+				externalClient.Stop()
+			}
+			tc.Stop()
+			if CurrentSpecReport().Failed() {
+				infra.DumpErrorData()
+			}
+			infra.Stop()
```
fasaxc: Think this is all automatic now?

fasaxc (changes requested):
Not sure about using the last known IP after the IP gets deleted.  When we get a failure report from the field, user will say "I restarted calico-node and everything broke"; they won't think to mention "and 3 weeks ago I deleted the IP from the Node because I wanted 14 bytes back in etcd".  Perhaps you could put a time limit on how long we'll use the last-known IP (say 90s)?

Should we also reflect the lack of host IP in Felix's readiness?

Might be worth auditing for potential flaky tests where the test was relying on felix delaying program application until the IP was set.  Mainly thinking of timing changes:

* Previously: felix comes up, does nothing, test adds IP, felix does 1 unit of programming, then connectivity is correct.
* Now: felix comes up, does 1 unit of programming, test adds IP, felix does second unit of programming, then connectivity is correct.  Might introduce a bit of extra delay?

tomastigera:
Quick recap of the current behaviour, after addressing the review feedback:

**BPF attach** — no host-IP gate. Programs attach with `HOST_IP=0` if unknown; VXLAN encap-source sites drop with `CALI_REASON_NO_HOST_IP` / `DroppedNoHostIP` counter.

**Cached `m.v[46].hostIP`:**

| Calc graph input | Cache | Re-attach |
|---|---|---|
| `HostMetadataUpdate{Ipv4Addr:A}` | set to A | yes, with `HOST_IP=A` |
| `HostMetadataUpdate{Ipv4Addr:""}` (Node alive, IP removed) | cleared | yes, with `HOST_IP=0` |
| `HostMetadataRemove` (Node gone / BGP emptied) | **preserved** | no (avoids SIGTERM race) |

A fresh Felix with no Node in the datastore attaches with `HOST_IP=0` straight away — the original `#12642` fix.

**New `BPFHostIP` readiness reporter** — independent from the dataplane cache, so the cached IP keeping traffic flowing doesn't mask the missing Node. Rule: v6-disabled → require v4; v6-enabled → require at least one. `Detail` carries `IPv4 unknown since <RFC3339>` so an operator restarting calico-node weeks later sees the underlying cause rather than a silently-happy dataplane.

**FV coverage** (3 scenarios, all pass): Node delete only; Node delete + Felix restart + recovery; IP removed while Node still exists (clears both `Spec.BGP` and `Spec.Addresses` — `extractNodeAddress` falls back from BGP to `Spec.Addresses`).

**Cluster validation** (banzai GCP, BPF mode, kernel 6.8): `kubectl delete node` + restore, 25+ consecutive SSH probes through the delete window, no drops; `BPFHostIP` reporting ready → degraded (verified pre-delete) → ready after restore.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -429,6 +430,14 @@ type bpfEndpointManagerDataplane struct {
 	hostIP   net.IP
 	mgr      *bpfEndpointManager
 
+	// hostIPPresent reflects whether the calc graph currently reports
+	// a host IP of this family for our Node. It is independent from
+	// hostIP, which may be a cached last-known value after the Node
+	// has been deleted; presence drives the BPFHostIP readiness
+	// reporter only.
+	hostIPPresent     bool
+	hostIPAbsentSince time.Time
```
fasaxc: Let's group these with `hostIP`, and maybe rename `hostIP` to `lastSeenHostIP`?

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -878,41 +897,59 @@ func (m *bpfEndpointManager) withIface(ifaceName string, fn func(iface *bpfInter
...
+		if err != nil {
+			ip = net.ParseIP(ipAddr)
 		}
-		m.ifacesLock.Unlock()
-		// We use host IP as the source when routing service for the ctlb workaround. We
-		// need to update those routes, so make them all dirty.
-		for svc := range m.services {
-			m.dirtyServices.Add(svc)
+		if ip == nil {
+			logrus.WithField("ipAddr", ipAddr).Warn("Cannot parse host IP, no change applied")
```
fasaxc: this case should be treated as a deletion

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1987,6 +2043,82 @@ func (m *bpfEndpointManager) reportHealth(ready bool, detail string) {
 	})
 }
 
+// setHostIPPresence records whether the calc graph currently reports a host
+// IP of the given family. It updates the BPFHostIP readiness signal but
+// does not touch the cached d.hostIP that the dataplane uses to program
+// BPF globals — the two are deliberately decoupled so that the dataplane
+// can keep using the last-known host IP after Node deletion while
+// readiness reflects the missing Node.
+func (m *bpfEndpointManager) setHostIPPresence(ipFamily int, present bool) {
```
fasaxc, line 2076: I suggest having one utility method `updateOurHostIP(ip ip.Addr)` where you pass `ip = nil` for a deletion or a parse failure.  `updateOurHostIP` then

- Checks for equality with the current IP and returns early if no change
- Updates the ready flag depending on nil/non-nil of the new ip.
- If new ip is non-nil, updates the stored IP and marks everything dirty.

Also suggest a "mark everything dirty" utility method.

felix/dataplane/linux/bpf_ep_mgr.go
```diff
@@ -878,41 +897,59 @@ func (m *bpfEndpointManager) withIface(ifaceName string, fn func(iface *bpfInter
...
-				iface.dpState.v4Readiness = ifaceNotReady
-				iface.dpState.v6Readiness = ifaceNotReady
-				return true
-			})
+	var ip net.IP
+	if ipAddr != "" {
+		var err error
+		ip, _, err = net.ParseCIDR(ipAddr)
+		if err != nil {
+			ip = net.ParseIP(ipAddr)
```
fasaxc: Don't we have a ParseCIDROrIP helper somewhere?

fasaxc (changes requested):

felix/fv/bpf_node_delete_test.go
tomastigera: Yes — handled in 41c7099c (pre-refactor): the test now relies on `DatastoreDescribe` for `infra.Stop()` and the failure-time `DumpErrorData` call, so only the topology-local resources (`hostW`, `externalClient`, `tc`) get explicit teardown.

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: Done in 6ab8b9d — `hostIP` → `lastSeenHostIP`, grouped with `hostIPPresent` / `hostIPAbsentSince` under a single comment block that calls out the dataplane/readiness decoupling.

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: Done in 6ab8b9d — parse failure now flows through `parseHostIP` returning `nil`, which `updateOurHostIP(nil, …)` treats identically to `HostMetadataRemove` / empty addr: keep cached IP, degrade readiness, no rebuild.

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: Done in 6ab8b9d. `updateOurHostIP(ip net.IP, ipFamily int)` is the single utility method — accepts `nil` for deletion / parse failure / empty addr; refreshes the BPFHostIP readiness signal; if `ip` is non-nil and differs from the cached value, updates `lastSeenHostIP` and calls the new `markEverythingDirty()` helper. The three callers (`HostMetadataUpdate` for v4 and v6, `HostMetadataRemove`) all funnel through it.

felix/dataplane/linux/bpf_ep_mgr.go
tomastigera: Yes — `cnet.ParseCIDROrIP` in `libcalico-go/lib/net`. Switched to that in 6ab8b9d, via a small `parseHostIP(addr, family)` helper that returns `nil` on empty/unparseable input and logs the failure with the family.

fasaxc (approved):
LGTM; nice to finally put this footgun to bed!

PR #12824, opened 2026-05-22
Wire up built-in tier deletion protection

caseydavenport:
The ValidatingAdmissionPolicy added in #12026 to block deletion of the built-in tiers (`default`, `kube-admin`, `kube-baseline`) was placed in `api/config/admission/` and never bundled into any chart or operator manifest. Nothing was applying it, so default tier deletion was succeeding in v3 CRD mode (caught by an e2e failure in the gcp-bpf job - see https://tigera.semaphoreci.com/workflows/ed1045c6-79db-430b-a653-2a0e6a76252c).

Move the policy into `api/admission/`, which the calico Helm chart already picks up via the `charts/calico/admission` symlink, and regenerate `manifests/calico-v3-crds.yaml`. The operator inherits this naturally since it consumes the v3 CRD bundle from here.

Also fix the CEL expressions to read `oldObject.metadata.name` instead of `object.metadata.name` - `object` is null on DELETE requests, so the original expressions wouldn't have matched anything even if the policy had been deployed.

Separately, skip `RequiresCalicoAPIServer` tests in the gcp-bpf e2e config since that job runs with `USE_API_SERVER=false`. Six other failures in the same run were tier RBAC tests that have no enforcement path in v3 CRD mode without the aggregated API server.

```release-note
Prevent deletion of built-in tiers in CRD mode. 
```

### DIFF
--- a/api/admission/protect-builtin-tiers.yaml
+++ b/api/admission/protect-builtin-tiers.yaml
@@ -15,8 +15,8 @@ spec:
         resources: ["tiers"]
         operations: ["DELETE"]
   validations:
-    - expression: "!(object.metadata.name in ['default', 'kube-admin', 'kube-baseline'])"
-      messageExpression: "'The built-in tier ' + object.metadata.name + ' cannot be deleted'"
+    - expression: "!(oldObject.metadata.name in ['default', 'kube-admin', 'kube-baseline'])"
+      messageExpression: "'The built-in tier ' + oldObject.metadata.name + ' cannot be deleted'"
 ---
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingAdmissionPolicyBinding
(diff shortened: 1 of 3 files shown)


caseydavenport:
Heads up on CI: the one red job (`gcp-kubeadm ... DATAPLANE=CalicoBPF`) fails a single test, `Tiered RBAC default tier should not allow deleting the default tier`. That's expected pre-merge.

That job installs Calico via the operator pulling a prebuilt hashrelease, not the chart in this PR. The operator only imports the new ValidatingAdmissionPolicy once this lands, so the policy isn't in the hashrelease yet and the test can't pass until after merge + a fresh hashrelease. The KinD e2e jobs (which do install this PR's chart) are all green, including BIRD conformance.

skoryk-oleksandr (approved):

PR #12829, opened 2026-05-25
Initial migration of node k8st tests to golang

mazdakn:
Initial step to move k8st python test to golang.

This PR include:
- Initial Makefile change to run golang test alongside python ones. A success indicate both setup passed.
- a new k8stutils package holding all necessary helper functions
- Migrating two simple tests (`test_simplt.py` and `test_health.py`) to new framework

**Release note:**
```release-note
TBD
```


node/tests/k8st/tests/simple_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	restart := func(state *restartChurnState) {
+		k8stutils.MustRun(t, "docker exec "+state.restartNode+" pkill bird")
+		err := k8stutils.RetryUntilSuccess(t, 15*time.Second, func() error {
+			_, err := k8stutils.Run(t, "docker exec "+state.restartNode+" pgrep bird",
+				k8stutils.RunOptions{AllowFail: true, SuppressErrLog: true})
+			return err
+		})
+		NewWithT(t).Expect(err).NotTo(HaveOccurred(), "BIRD did not restart within 15s")
+		time.Sleep(5 * time.Second)
```
caseydavenport, line 50: Wold be great to avoid sleeps like this - they can sneakily add hours of time to test suites if they get executed by multiple tests.

node/tests/k8st/tests/simple_test.go
```diff
@@ -0,0 +1,181 @@
...
+// pod name (which changes across pod replacements in the GR test).
+type restartChurnState struct {
+	restartNode    string
+	restartNodeIP  string
+	restartPodName string
+}
+
+func (s *restartChurnState) refreshRestartPodName(t testing.TB) error {
+	t.Helper()
+	out, err := k8stutils.Kubectl(t, fmt.Sprintf(
```
caseydavenport, line 94: Is there any reason to use kubectl instead of the native Go client for stuff like this? Seems like it would be cleaner now that we have that option. The python tests just didn't have a good client available I think?

node/Makefile
```diff
@@ -390,21 +394,41 @@ k8s-test:
 	$(MAKE) kind-k8st-run-test
 	$(MAKE) -C $(REPO_ROOT) kind-down
 
+# Selector for Go-side k8st tests (e.g. K8ST_GO_TO_RUN='TestReadiness').
+# Defaults to running every Go test in the suite.
+K8ST_GO_TO_RUN?=.
```
caseydavenport, line 399: I think we should just standardize on `WHAT` if we can

Or ideally just switch over to using `go test` commands natively instead of plumbing through the Makefile.

caseydavenport (approved):
A few minor thoughts but I think we should merge this as is.

mazdakn:
Will address the comments in the follow up PR.

PR #12833, opened 2026-05-26
Render MutatingAdmissionPolicy as v1 when the cluster serves it

caseydavenport:
K8s 1.36 promoted `MutatingAdmissionPolicy` to `admissionregistration.k8s.io/v1` and v1beta1 is scheduled for removal in 1.37, so installs against clusters that only serve v1 fail. This uses Helm's `.Capabilities.APIVersions.Has` to detect the served version and rewrite the apiVersion at template-render time. If neither is served we fall back to v1beta1 and the apply fails naturally, same as today.

The companion operator-side fix is https://github.com/tigera/operator/pull/4837.

Verified that `helm template` without `--api-versions` still produces v1beta1 (so the generated `manifests/calico-v3-crds.yaml` is unchanged), and that passing `--api-versions admissionregistration.k8s.io/v1/MutatingAdmissionPolicy` flips all four resources to v1.

```release-note
HELM: Render MutatingAdmissionPolicy and MutatingAdmissionPolicyBinding as admissionregistration.k8s.io/v1 when the cluster serves it (Kubernetes 1.36+), falling back to v1beta1 otherwise.
```

### DIFF
--- a/charts/calico/templates/admission-policies.yaml
+++ b/charts/calico/templates/admission-policies.yaml
@@ -1,6 +1,17 @@
 {{- if .Values.useV3CRDs }}
+{{- /*
+  MutatingAdmissionPolicy was promoted to admissionregistration.k8s.io/v1 in
+  Kubernetes 1.36 and v1beta1 is targeted for removal in 1.37. Prefer v1 when
+  the cluster serves it; otherwise emit v1beta1 (the source-of-truth version
+  in this chart). If neither is served the apply will fail naturally, which
+  matches what the operator would do.
+*/ -}}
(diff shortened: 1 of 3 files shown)


mazdakn (approved):

james-callahan:
Could you also update the static manifests, as they don't currently work with k8s 1.36 (which doesn't have the beta api)

caseydavenport:
@james-callahan the static manifests should be updated in master / v3.33, which is the next release. 

For v3.32, you'll have to simply swap the API version manually - doing that on a release branch would be a breaking change for anyone deploying on older clusters.

PR #12837, opened 2026-05-26
docs: add IPAM design under docs/design/ipam/

caseydavenport:
Adds a cross-component design index for IPAM under `docs/design/ipam/`. IPAM spans `libcalico-go/lib/ipam`, `cni-plugin`, `kube-controllers`, and `node`, so there is no single owning component - a top-level `docs/design/<topic>/` placement matches that better than putting it under one of them.

Follows the documentation conventions already in place for Felix (component DESIGN.md index plus per-topic sub-files) and Goldmane (path-scoped Copilot instructions).

Structure:
- `docs/design/ipam/DESIGN.md` - the index.
- Five sub-designs covering the core library, datastore, CNI, garbage collection, and other callers (calicoctl, tunnel allocator, Felix, LoadBalancer, Flannel migration, vmipam).
- Three discoverability stub `DESIGN.md` files in consumer subdirectories that point at the canonical design.
- `.github/instructions/ipam.instructions.md` for Copilot path-scoped instructions.
- An extension to `.claude/CLAUDE.md` documenting the cross-component design pattern for future subsystems.

Each sub-design carries "Review notes" blocks per section in the Felix style, with reviewer-known landmines from prior IPAM PRs distributed across them.

**Release note:**
```release-note
None
```

### DIFF
--- a/docs/design/ipam/DESIGN.md
+++ b/docs/design/ipam/DESIGN.md
@@ -22,21 +22,25 @@ reason.
 
 ### Data model
 
-IPAM state is stored as four CRDs under `crd.projectcalico.org/v1`,
-plus `IPReservation` for carve-outs:
+IPAM state is stored as four CRDs, plus `IPReservation` for
+carve-outs:
 
 | Resource | Role |
 |---|---|
 | `IPAMBlock` | Contiguous slice of a pool (default /26 IPv4, /122 IPv6). Holds the ordinal bitmap, per-allocation attributes, and per-ordinal sequence numbers. |
 | `BlockAffinity` | Per-host (or per-virtual-owner) claim on a block. State machine: `pending` -> `confirmed` -> `pendingDeletion`. |
 | `IPAMHandle` | Secondary index keyed by handle ID, so `ReleaseByHandle` doesn't scan blocks. Also enforces per-handle allocation caps (KubeVirt VM persistence). |
-| `IPAMConfig` | Singleton. Holds `StrictAffinity`, `MaxBlocksPerHost`, `AutoAllocateBlocks`, `KubeVirtVMAddressPersistence`. |
+| `IPAMConfig` / `IPAMConfiguration` | Singleton. Holds `StrictAffinity`, `MaxBlocksPerHost`, `AutoAllocateBlocks`, `KubeVirtVMAddressPersistence`. The v1 name is `IPAMConfig`; the v3 name is `IPAMConfiguration`. |
 | `IPReservation` | Carve-outs from the pool (tunnel addresses, externally-managed IPs). Filtered at block-skip and ordinal-skip granularity. |
 
-These resources still live on the internal `crd.projectcalico.org/v1`
-group rather than the public `projectcalico.org/v3`. The long-running
-discussion about promoting them is
-https://github.com/projectcalico/calico/issues/6412.
+The IPAM CRDs exist on both `crd.projectcalico.org/v1` and
+`projectcalico.org/v3`. The two groups are not symmetric - they
+differ in field shape, naming (`IPAMConfig` vs
+`IPAMConfiguration`), and which clients write through each. A
+separate design doc covering the CRD-group layout is owed; until
+it lands, treat any cross-group assumption as something to
+verify in code rather than infer from this doc. The promotion
+history is https://github.com/projectcalico/calico/issues/6412.
 
 ### Why blocks, and why per-host affinity
 
@@ -47,13 +51,13 @@ ordinals into a single object, so most allocations are CAS on one
 object owned by one node. Blocks also let the dataplane advertise one
 route per block, not one per pod.
 
-**Per-host affinity** turns cluster-wide CAS thrash into per-node CAS.
-A node owns its blocks; allocations are local writes. When a node
-runs out of affine blocks it can borrow ordinals from another node's
-block (unless `StrictAffinity=true`, which is required when IPs are
-statically routed per-node by downstream gear and borrowing would
-break routing). Windows nodes force `StrictAffinity=true` because
-Windows can't route /26 affinity blocks remotely.
+**Per-host affinity** turns cluster-wide CAS thrash into per-node
+CAS. A node owns its blocks; allocations are local writes. When a
+node runs out of affine blocks it can borrow ordinals from another
+node's block, unless `StrictAffinity=true`. Strict affinity is the
+mode for clusters where downstream routing is per-node and
+borrowing would strand traffic (Windows forces this; see
+[ipam-cni](./ipam-cni.md#platform-differences)).
 
 ### The three primary consumers
 
@@ -71,13 +75,10 @@ moving.
 ### Repo split
 
 - `projectcalico/calico` (OSS) - all the IPAM code described here.
-- `tigera/operator` - owns the `IPAMConfig` CR (reconciles it against
-  user-facing operator config) and the RBAC for the IPAM CRDs.
-  Operator can refuse to apply config changes when datastore state
-  looks corrupt.
-- `tigera/calico-private` (enterprise) - carries the L2 bridge /
-  VLAN, federation, and KubeVirt IPAM extensions. Mostly cherry-picks
-  from OSS; the core library lives upstream.
+- `tigera/operator` - reconciles `IPAMConfig` / `IPAMConfiguration`
+  against user-facing operator config and provisions RBAC for the
+  IPAM CRDs. End users can also edit the config CR directly (name
+  varies by API group); operator is one writer, not the only one.
 
 ## 2. Sub-design index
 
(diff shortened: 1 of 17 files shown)


docs/design/ipam/DESIGN.md
```diff
@@ -0,0 +1,143 @@
...
+
+| Resource | Role |
+|---|---|
+| `IPAMBlock` | Contiguous slice of a pool (default /26 IPv4, /122 IPv6). Holds the ordinal bitmap, per-allocation attributes, and per-ordinal sequence numbers. |
+| `BlockAffinity` | Per-host (or per-virtual-owner) claim on a block. State machine: `pending` -> `confirmed` -> `pendingDeletion`. |
+| `IPAMHandle` | Secondary index keyed by handle ID, so `ReleaseByHandle` doesn't scan blocks. Also enforces per-handle allocation caps (KubeVirt VM persistence). |
+| `IPAMConfig` | Singleton. Holds `StrictAffinity`, `MaxBlocksPerHost`, `AutoAllocateBlocks`, `KubeVirtVMAddressPersistence`. |
+| `IPReservation` | Carve-outs from the pool (tunnel addresses, externally-managed IPs). Filtered at block-skip and ordinal-skip granularity. |
+
+These resources still live on the internal `crd.projectcalico.org/v1`
```
caseydavenport: Not quite true - we support both modes now. 

Probably need a separate design document discussing the backing CRD group design.

docs/design/ipam/DESIGN.md
```diff
@@ -0,0 +1,143 @@
...
+moving.
+
+### Repo split
+
+- `projectcalico/calico` (OSS) - all the IPAM code described here.
+- `tigera/operator` - owns the `IPAMConfig` CR (reconciles it against
+  user-facing operator config) and the RBAC for the IPAM CRDs.
+  Operator can refuse to apply config changes when datastore state
+  looks corrupt.
+- `tigera/calico-private` (enterprise) - carries the L2 bridge /
```
caseydavenport: Can drop this.

docs/design/ipam/DESIGN.md
```diff
@@ -0,0 +1,143 @@
...
+
+The IPAM GC lives in the *node* controller
+(`kube-controllers/pkg/controllers/node`), not in
+`pkg/controllers/ipam`. Confusing, but historical and not worth
+moving.
+
+### Repo split
+
+- `projectcalico/calico` (OSS) - all the IPAM code described here.
+- `tigera/operator` - owns the `IPAMConfig` CR (reconciles it against
```
caseydavenport: Not necessarily true - end users can modify IPAMConfig / IPAMConfiguration (name varies by API group) as well.

docs/design/ipam/ipam-core-library.md
```diff
@@ -0,0 +1,172 @@
...
+| `MaxBlocksPerHost` | Per-host cap on the number of affine blocks. 0 means default (20). Once a host hits the cap, `allowNewClaim` is forced false; existing blocks still fill. |
+| `AutoAllocateBlocks` | When false, `AutoAssign` will never claim a new block - only allocate from blocks the host already owns. |
+| `KubeVirtVMAddressPersistence` | Default for whether KubeVirt VM addresses survive VM restart / migration. Auto-detection is on by default. |
+
+**Review notes**
+
+- The default-when-missing behavior is load-bearing. New required fields need a default plus heal-forward; don't add a field that crashes when absent.
+- `MaxBlocksPerHost > 0` only makes sense with `StrictAffinity=true`; the validator enforces this. If you relax the validator, you also need to define what "borrow blocks but cap
+  our own" means - it currently isn't defined.
+- `SetIPAMConfig` is the only validation point. End-user kubectl writes hit the CR directly and bypass library checks until next read - new validation rules must live in the
```
caseydavenport: Not quite true - there's kubebuilder / CEL validation on the CRDs in k8s mode.

docs/design/ipam/ipam-core-library.md
```diff
@@ -0,0 +1,176 @@
...
+
+| Caller | Handle ID format |
+|---|---|
+| CNI workload (default) | `<network-name>.<container-id>` via `cni-plugin/internal/pkg/utils.GetHandleID`. For the default network, `<network-name>` is `k8s-pod-network`. |
+| CNI workload (KubeVirt persistent) | `<network-name>.<namespace>-<vm-name>` so live-migrated VMs keep the same handle. |
+| IPIP tunnel | `ipip-tunnel-addr-<node>` |
+| VXLAN tunnel | `vxlan-tunnel-addr-<node>`; IPv6 variant is `vxlan-v6-tunnel-addr-<node>` (note the `-v6-` infix, not a suffix) |
+| WireGuard tunnel | `wireguard-tunnel-addr-<node>`; IPv6 variant is `wireguard-v6-tunnel-addr-<node>` |
+| Windows-reserved | literal `windows-reserved-ipam-handle` |
+| LoadBalancer | `<namespace>:<service>` |
```
tomastigera: The LoadBalancer handle isn't `<namespace>:<service>`. The code builds it as:

```go
// kube-controllers/pkg/controllers/loadbalancer/loadbalancer_controller.go:1007
handle := strings.ToLower(fmt.Sprintf("%s-%s-%s", svc.Name, svc.Namespace, svc.UID))
```

i.e. `<service>-<namespace>-<uid>`, lowercased, and hash-truncated past the DNS1123 limit (lines 1019-1021). Looks like it got conflated with the `virtual:load-balancer` *affinity* string (which is documented correctly elsewhere).

docs/design/ipam/ipam-other-callers.md
```diff
@@ -0,0 +1,108 @@
...
+
+**Review notes**
+
+- Don't introduce a path that mutates owner attrs without going through `vmipam`, or the `CompareAndSwap` precondition protecting against the CNI racing the swap goes away.
+- Trigger-path refactors have regressed live-migration more than once; exercise the GARP and syncer paths whenever touching the monitor.
+
+## LoadBalancer controller
+
+[`kube-controllers/pkg/controllers/loadbalancer/`](../../../kube-controllers/pkg/controllers/loadbalancer/). Allocates IPs for Calico's Service LoadBalancer abstraction. Uses
+`virtual:load-balancer` affinity (not host-anchored), filters pools on `allowedUses: LoadBalancer`, and handles are `<namespace>:<service>`. `AssignIP` for static `loadBalancerIP`,
```
tomastigera: Same handle-format issue as in `ipam-core-library.md`: the handle is `<service>-<namespace>-<uid>` (lowercased, hash-truncated), not `<namespace>:<service>` — see `loadbalancer_controller.go:1007`. The `virtual:load-balancer` affinity mentioned just before this is correct; only the handle format here is wrong.

docs/design/ipam/ipam-other-callers.md
```diff
@@ -0,0 +1,108 @@
...
+## Flannel migration
+
+[`kube-controllers/pkg/controllers/flannelmigration/ipam_migrator.go`](../../../kube-controllers/pkg/controllers/flannelmigration/ipam_migrator.go). One-time migration from
+Flannel's host-local IPAM to Calico IPAM. For each node:
+
+1. Read Flannel subnet.
+2. `ClaimAffinity` for the equivalent Calico block.
+3. Reassign the existing VXLAN tunnel address with Calico metadata.
+
+Handle convention for the migrated tunnel: `vxlan-<node>` (note the missing `-tunnel-addr-` infix - a quirk preserved from the original migration). Idempotent; runs during cluster
```
tomastigera: This "quirk" doesn't exist in the code. The migrated tunnel handle is the standard format, with the `-tunnel-addr-` infix present:

```go
// kube-controllers/pkg/controllers/flannelmigration/ipam_migrator.go:248
handle := fmt.Sprintf("vxlan-tunnel-addr-%s", nodeName)
```

So it's `vxlan-tunnel-addr-<node>`, not `vxlan-<node>` — there's no missing-infix special case to preserve. Suggest dropping the parenthetical.

tomastigera (commented):
A couple of handle-ID formats don't match the code. I verified the rest of the concrete claims (line numbers, constants, function names, tunnel handle formats, etc.) and they check out — just these two to fix. Inline below.

tomastigera:
## Layout: where should a cross-component design live?

Separate from the content nits above — this is about placement, and since it sets the precedent for the next cross-component subsystem, worth settling explicitly. The cross-component design itself is well-justified: IPAM has no single owning component, so a repo-level home is right. The question is just *which* repo-level home.

Today the repo has no top-level `docs/` — component designs are component-rooted (`felix/DESIGN.md` + `felix/design/`, `goldmane/DESIGN.md`), and user-facing docs live in the separate `tigera/docs` repo. So this introduces a brand-new top-level tree, and there are two reasonable shapes:

**A. `docs/design/ipam/` (what this PR does)**
- Conventional, discoverable, already written.
- Downside: `docs/` at the calico root currently means nothing, so a top-level `docs/` invites "is this published to the docs site?" Mild, but real — and it makes `docs/` mean two different things across our repos.

**B. top-level `design/ipam/`**
- Structural match to what we already have: the existing pattern is `<component>/design/<topic>.md` (e.g. `felix/design/bpf-*.md`). The cross-component case is the same shape with the component segment dropped because nobody owns it → `design/ipam/*.md` with `design/ipam/DESIGN.md` as the index.
- Reads unambiguously as "design material," no overlap with the published-docs meaning of "docs."

**My preference is B.** It keeps the new top-level directory's purpose unambiguous and mirrors the `felix/design/` convention rather than inventing a parallel one. The principle I'd like to lock in either way: *cross-component designs hoist the existing `<component>/design/` pattern to the repo root, rather than creating a new structure.*

If we go with B it's a straight directory rename plus fixing the relative links in the stubs / instructions file and the `.claude/CLAUDE.md` + `.github/copilot-instructions.md` maps. Happy to go with A if folks feel `docs/` is clearer — just want it to be a deliberate call. Thoughts?

docs/design/ipam/ipam-core-library.md
caseydavenport: Fixed - it's `lb-<hash>` from `createHandle` now, with the affinity-string confusion called out. Thanks for catching it.

docs/design/ipam/ipam-other-callers.md
caseydavenport: Fixed here too, same way.

docs/design/ipam/ipam-other-callers.md
caseydavenport: Yep, dropped the parenthetical - it's the standard `vxlan-tunnel-addr-<node>`, no quirk to preserve.

docs/design/ipam/ipam-core-library.md
caseydavenport: Fixed - the bullet now says the CRD schema enforces field-level bounds in k8s mode, and only the cross-field rules are library-only. Matches the IPAMConfig section.

PR #12839, opened 2026-05-27
Fix IPAM GC stuck in update-conflict loop on stale sequence number

### LINKED ISSUE #12838: IPAM garbage collector stuck in permanent "update conflict" loop on stale per-ordinal sequence number
## Expected Behavior

The IPAM garbage collector in calico-kube-controllers should clean up leaked IP allocations and make progress. If a release hits a conflict it should sort itself out on a later sync, not get wedged.

## Current Behavior

We've been chasing an intermittent issue for a while now. Every few months, on long-running clusters, one or two nodes start doing a lot of disk read I/O. iotop points at the `kube-controllers` process. Digging into the logs, the IPAM GC is stuck in a tight loop trying to release the same leaked IPs over and over, and every single attempt fails with `update conflict`:

```
[INFO]    kube-controllers/ipam.go: Pod rescheduled on new node. Allocation no longer valid handle="k8s-pod-network.<hash>" ip="10.0.4.36" new="nodeB" old="nodeA" pod="ns/statefulset-pod-1"
[INFO]    kube-controllers/ipam.go: Garbage collecting leaked IP address handle="k8s-pod-network.<hash>" ip="10.0.4.36" node="nodeA"
[ERROR]   kube-controllers/ipam.go: Error releasing IPs: update conflict: 10.0.4.36
[WARNING] kube-controllers/ipam.go: Failed to release leaked IP error=update conflict: 10.0.4.36
[WARNING] kube-controllers/ipam.go: error syncing IPAM data error=update conflict: 10.0.4.36
[INFO]    kube-controllers/ipam.go: Checking dirty nodes for leaks and redundant affinities
```

It never recovers on its own. Over a few minutes of logs there are hundreds of these conflicts, not a single successful release, and it's always the same handful of IPs going round and round (one IP showing up 30-50 times). The IPs are real leaks — they belong to pods that moved to another node (StatefulSet pods getting rescheduled, in our case), so the GC is right to want to release them, it just can't.

Restarting the calico-kube-controllers deployment fixes it immediately, and then it's fine for another few months. So we've been living with a "restart it when a node starts grinding" workaround, but I finally sat down to find out why.

On the I/O side: the pod runs with the kubespray default 256M memory limit and no swap. When it's stuck in this loop the Go heap sits right at the limit, the cgroup's page cache gets squeezed to zero, and the kernel ends up constantly evicting and re-faulting the binary's own pages — a few hundred major faults a second, which with readahead is the multi-MB/s of reads we saw in iotop. No OOM kill (failcnt stays 0), it just thrashes. Bumping the limit hides the I/O symptom but obviously doesn't fix the loop itself.

## Possible Solution

The `update conflict` is misleading — it's not a datastore/resourceVersion CAS conflict, it's the IPAM sequence-number check in `allocationBlock.release()` (`libcalico-go/lib/ipam/ipam_block.go`), where a mismatched sequence number returns `ErrorResourceUpdateConflict{Err: ErrorBadSequenceNumber{...}}`. That comes straight back out and skips the retry loop in `releaseIPsFromBlock`, so re-reading the block doesn't help — the controller just keeps handing it the same stale sequence number.

I traced where the stale value comes from: `onBlockUpdated()` in `kube-controllers/pkg/controllers/node/ipam.go`. Allocations are tracked by `handle/IP`, and the sequence number isn't part of that key. When a block update comes in for an allocation it already knows about, it just does:

```go
// Check if we already know about this allocation.
if _, ok := c.allocationsByBlock[blockCIDR][alloc.id()]; ok {
    continue
}
```

So it never refreshes the cached sequence number. The per-ordinal sequence number gets bumped every time the address is re-allocated, so if the syncer ever delivers a coalesced update where the same `(handle, IP)` is still present but the sequence number has moved (IP released and re-allocated between syncs — which happens under StatefulSet / IP-reuse churn), the cached value is now stale forever. From that point on every GC release of that IP fails the sequence-number check, and only a restart (which rebuilds the cache) clears it. That lines up exactly with the "wedged until restart, then fine for months" behaviour.

The fix that worked for me is to just refresh the cached sequence number in place on a block update instead of skipping it (keeping the leak-tracking state intact). I put it on a branch with a regression unit test that reproduces the coalesced sequence-number change — it passes with the fix and fails without it (times out waiting for the cached value to update):

https://github.com/andrei-marinache/calico/tree/fix-ipam-gc-stale-sequence-number
PR: https://github.com/projectcalico/calico/pull/12839

1. IP `X` is allocated to handle `H` at ordinal `O`; kube-controllers caches the allocation with sequence number `N`.
2. Between two syncs the controller sees, `X` is released and re-allocated, so the block is rewritten and `O` now carries sequence number `M` > `N`, but as far as the controller's cache is concerned the `(handle, IP)` pair `H/X` was there before and is still there.
3. That coalesced block update hits the "already know about this allocation" branch in `onBlockUpdated`, so the cached sequence number stays at `N`.
4. The allocation later gets confirmed as a leak. Every release attempt sends sequence number `N`, the block rejects it with `update conflict` (bad sequence number), and it loops forever until kube-controllers is restarted.

There's a deterministic unit test on the branch above that drives this through `onBlockUpdated` directly.

## Context

This has been a recurring operational papercut for us across a fleet of long-lived clusters — every couple of months a node starts hammering its disk and we restart calico-kube-controllers to clear it.

## Your Environment

* Calico version: v3.30.6 (same code path is still in master)
* Calico dataplane: iptables, VXLAN, Kubernetes datastore (KDD)
* Orchestrator version: Kubernetes v1.33.7 (kubespray)
* Operating System and version: Oracle Linux Server release 9.7, cgroup v1, swap disabled
* Link to your project (optional): n/a

andrei-marinache:
Fixes #12838.

  The IPAM garbage collector in calico-kube-controllers can get permanently stuck trying to release a leaked IP, failing every time with `update conflict` and never making progress until the deployment is restarted (after which it's fine again
  for weeks/months).

  The `update conflict` isn't a datastore/resourceVersion CAS conflict — it's the sequence-number check in `allocationBlock.release()`, which returns `ErrorResourceUpdateConflict{Err: ErrorBadSequenceNumber{...}}` when the sequence number it's
  handed doesn't match the block. That error skips the retry loop in `releaseIPsFromBlock`, so re-reading the block doesn't help — the controller keeps submitting the same stale sequence number from its cache.

  The stale value comes from `onBlockUpdated()`. Allocations are tracked by `handle/IP` (the sequence number isn't part of the key), and when a block update arrives for an allocation we already know about, we `continue` without refreshing the
  cached sequence number. The per-ordinal sequence number is bumped every time the address is re-allocated, so a coalesced block update where the same `(handle, IP)` is still present but the sequence number has moved (IP released and
  re-allocated between syncs, common under StatefulSet / IP-reuse churn) leaves the cached value stale for good. From then on every GC release of that IP fails the sequence-number check, which is exactly the "wedged until restart" behaviour.

  This refreshes the cached sequence number in place on each block update, preserving the leak-tracking state, so releases use the current value and succeed.

  **Testing:** added a regression unit test in `ipam_test.go` that drives a coalesced sequence-number change through `onBlockUpdated` and asserts the cached allocation is refreshed. It passes with the fix and fails without it (times out waiting
  for the cached sequence number to update). Full `kube-controllers` node controller UT suite passes.

  **Release note:**

  ```release-note
  Fix a bug where the IPAM garbage collector in calico-kube-controllers could get stuck in a loop failing to release a leaked IP with an "update conflict" error until the controller was restarted.
```

### DIFF
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -575,6 +575,7 @@ func (c *IPAMController) onBlockUpdated(kvp model.KVPair) {
 			// it has been reallocated. Update the allocation in place and mark it as valid.
 			if existing.sequenceNumber != alloc.sequenceNumber {
 				existing.sequenceNumber = alloc.sequenceNumber
+				existing.attrs = alloc.attrs
 				existing.markValid()
 			}
 			continue
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -570,7 +570,22 @@ func (c *IPAMController) onBlockUpdated(kvp model.KVPair) {
 		currentAllocations[alloc.id()] = true
 
 		// Check if we already know about this allocation.
-		if _, ok := c.allocationsByBlock[blockCIDR][alloc.id()]; ok {
+		if existing, ok := c.allocationsByBlock[blockCIDR][alloc.id()]; ok {
+			// We already track this (handle, IP). Refresh the sequence number
+			// from the latest block, preserving leak-tracking state.
+			//
+			// The per-ordinal sequence number is bumped by IPAM every time the
+			// address is (re)allocated. A coalesced block update can carry a new
+			// sequence number for an allocation whose (handle, IP) identity is
+			// unchanged. If we keep the stale value, every attempt to release
+			// this IP fails with a sequence-number conflict (ErrorBadSequenceNumber,
+			// surfaced as "update conflict"), which is not a retryable datastore
+			// conflict and never resolves on its own. The GC then spins forever
+			// re-attempting a release it can never complete, and only a restart
+			// (which rebuilds this cache from scratch) clears it. Refreshing in
+			// place keeps our cached view accurate so the release uses the
+			// current sequence number and succeeds.
+			existing.sequenceNumber = alloc.sequenceNumber
 			continue
 		}
 
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -571,21 +571,12 @@ func (c *IPAMController) onBlockUpdated(kvp model.KVPair) {
 
 		// Check if we already know about this allocation.
 		if existing, ok := c.allocationsByBlock[blockCIDR][alloc.id()]; ok {
-			// We already track this (handle, IP). Refresh the sequence number
-			// from the latest block, preserving leak-tracking state.
-			//
-			// The per-ordinal sequence number is bumped by IPAM every time the
-			// address is (re)allocated. A coalesced block update can carry a new
-			// sequence number for an allocation whose (handle, IP) identity is
-			// unchanged. If we keep the stale value, every attempt to release
-			// this IP fails with a sequence-number conflict (ErrorBadSequenceNumber,
-			// surfaced as "update conflict"), which is not a retryable datastore
-			// conflict and never resolves on its own. The GC then spins forever
-			// re-attempting a release it can never complete, and only a restart
-			// (which rebuilds this cache from scratch) clears it. Refreshing in
-			// place keeps our cached view accurate so the release uses the
-			// current sequence number and succeeds.
-			existing.sequenceNumber = alloc.sequenceNumber
+			// If the sequence number has changed for an existing allocation, it means
+			// it has been reallocated. Update the allocation in place and mark it as valid.
+			if existing.sequenceNumber != alloc.sequenceNumber {
+				existing.sequenceNumber = alloc.sequenceNumber
+				existing.markValid()
+			}
 			continue
 		}
 
--- a/kube-controllers/pkg/controllers/node/ipam_test.go
+++ b/kube-controllers/pkg/controllers/node/ipam_test.go
@@ -723,18 +723,13 @@ var _ = Describe("IPAM controller UTs", func() {
 	})
 
 	It("should refresh the sequence number of an already-tracked allocation on block update", func() {
-		// Regression test: a coalesced block update that carries a new sequence
-		// number for an allocation whose (handle, IP) identity is unchanged must
-		// refresh the cached sequence number. If it doesn't, the cached value
-		// goes stale and every attempt to release the IP fails with a
-		// sequence-number conflict (ErrorBadSequenceNumber, surfaced as "update
-		// conflict"), which is not a retryable datastore conflict - so the GC
-		// spins forever and only a restart clears it.
+		// Regression test: if a reallocation bumps an allocation's sequence number
+		// but we keep the stale one, every GC release of that IP fails with "update
+		// conflict" forever (until restart). See onBlockUpdated.
 		c.Start(stopChan)
 
-		// A /30 block affine to "cnode" with one address allocated to a handle.
-		// The block has been written 10 times, and the address was allocated at
-		// that sequence number.
+		// A /30 block affine to "cnode" with one address allocated to a handle,
+		// at the block's current sequence number (10).
 		cidr := net.MustParseCIDR("10.0.0.0/30")
 		aff := "host:cnode"
 		key := model.BlockKey{CIDR: model.PrefixFromIPNet(cidr)}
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: kube-controllers/pkg/controllers/node/ipam.go [hunk 1]
```diff
@@ -570,7 +570,22 @@ func (c *IPAMController) onBlockUpdated(kvp model.KVPair) {
...
+			// address is (re)allocated. A coalesced block update can carry a new
+			// sequence number for an allocation whose (handle, IP) identity is
+			// unchanged. If we keep the stale value, every attempt to release
+			// this IP fails with a sequence-number conflict (ErrorBadSequenceNumber,
+			// surfaced as "update conflict"), which is not a retryable datastore
+			// conflict and never resolves on its own. The GC then spins forever
+			// re-attempting a release it can never complete, and only a restart
+			// (which rebuilds this cache from scratch) clears it. Refreshing in
+			// place keeps our cached view accurate so the release uses the
+			// current sequence number and succeeds.
```
caseydavenport: I think we should also clear `leakedAt` / `confirmedLeak` here when the sequence number actually changed. A seq bump on a same-(handle, IP) ordinal means the prior allocation was released and re-handed out, so any leak-tracking state we accumulated belongs to a dead incarnation. Carrying it over could trip the new allocation into confirmed-leak earlier than it should (the final `allocationIsValid` re-validate in `garbageCollectKnownLeaks` would usually catch the false positive in practice, but the cache should still be honest).

Something like:
```go
if existing.sequenceNumber != alloc.sequenceNumber {
    existing.sequenceNumber = alloc.sequenceNumber
    existing.markValid()
}
continue
```

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -570,7 +570,22 @@ func (c *IPAMController) onBlockUpdated(kvp model.KVPair) {
 		currentAllocations[alloc.id()] = true
 
 		// Check if we already know about this allocation.
-		if _, ok := c.allocationsByBlock[blockCIDR][alloc.id()]; ok {
+		if existing, ok := c.allocationsByBlock[blockCIDR][alloc.id()]; ok {
+			// We already track this (handle, IP). Refresh the sequence number
```
caseydavenport: This comment is a bit heavy for inline code - let's minimize it:

```go
// If the sequence number has changed for an existing allocation, it means 
// it has been reallocated. Update the allocation in place and mark it as valid.
```

kube-controllers/pkg/controllers/node/ipam_test.go
```diff
@@ -722,6 +722,85 @@ var _ = Describe("IPAM controller UTs", func() {
 		}, 1*time.Second, 100*time.Millisecond).Should(BeNil())
 	})
 
+	It("should refresh the sequence number of an already-tracked allocation on block update", func() {
```
caseydavenport, line 725: Two things on the test:

- We should call `assertConsistentState(c)` at the end - that's the convention for any test that mutates the controller and it catches map-divergence bugs.
- Paired with the suggestion above: worth adding a case that marks the cached allocation as a candidate leak before the second `onUpdate`, then asserts the leak state is cleared after.

caseydavenport (changes requested):
Thanks for fixing this @andrei-marinache!

A few suggestions for you.

kube-controllers/pkg/controllers/node/ipam.go
andrei-marinache: Done in e728624. Used your snippet (gate on the seq change + markValid()).
 
I deliberately didn't also delete from confirmedLeaks here: markValid() mutates the same *allocation the index holds, so the entry gets reconciled on the next checkAllocations pass (the else if … delete(c.confirmedLeaks, …) branch), and garbageCollectKnownLeaks re-validates before releasing anyway — so the cache stays honest without a second write.

kube-controllers/pkg/controllers/node/ipam.go
andrei-marinache: Right, sorry. Done in e728624.

kube-controllers/pkg/controllers/node/ipam_test.go
andrei-marinache: Done in e728624 — added assertConsistentState(c) at the end, plus a case that marks the allocation a leak candidate before the reallocating onUpdate and asserts the leak state is cleared after. (The final reads are in their own paused scope since assertConsistentState pauses the controller internally. Also trimmed the test's own comments to match the surrounding style.)

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -570,7 +570,13 @@ func (c *IPAMController) onBlockUpdated(kvp model.KVPair) {
 		currentAllocations[alloc.id()] = true
 
 		// Check if we already know about this allocation.
-		if _, ok := c.allocationsByBlock[blockCIDR][alloc.id()]; ok {
+		if existing, ok := c.allocationsByBlock[blockCIDR][alloc.id()]; ok {
+			// If the sequence number has changed for an existing allocation, it means
+			// it has been reallocated. Update the allocation in place and mark it as valid.
+			if existing.sequenceNumber != alloc.sequenceNumber {
+				existing.sequenceNumber = alloc.sequenceNumber
```
caseydavenport, line 577: I think we should probably also do this:

```go
existing.attrs = alloc.attrs
```

To make sure these stay in sync - right now, I don't think it's possible for these to get out of sync because the id() check makes sure we are dealing with the same allocation, but if the contents of `attrs` or the logic around this changes in the future, we might introduce a latent bug.

caseydavenport: From what I can tell, the rest of the fields are safe - we learn those (e.g., knode) down below so we don't risk acting on stale data.

caseydavenport (commented):
One last thing, but otherwise I think this LGTM!

kube-controllers/pkg/controllers/node/ipam.go
andrei-marinache: Done in cdcd9b3 — refresh existing.attrs inside the reallocation gate.

caseydavenport (approved):

caseydavenport:
Thanks @andrei-marinache !

PR #12840, opened 2026-05-27
Speed up the "make fix-changed" formatting pipeline

fasaxc:
Speeds up the `make fix-changed` formatting pipeline, which had two main costs.

**1. coalesce-imports compile cost.** The helper imported `libcalico-go/lib/logutils` solely for `ConfigureFormatter` (a cosmetic log prefix), which dragged in `prometheus/client_golang` and its transitive tree — 254 packages, 237 of them from Prometheus alone. On a cold build cache, `go run` recompiled all of it (~14s) on every invocation.

- Split the custom logrus `Formatter` and its helpers into a new `libcalico-go/lib/logutils/logformat` sub-package that depends only on the standard library and logrus. `logutils` re-exports the moved symbols as aliases, so existing callers are unchanged; the Prometheus-using code stays in `logutils`.
- coalesce-imports now imports `logformat`, dropping its dependency tree from 254 → 90 packages (0 Prometheus).
- Net: cold `go run` of the tool ~14s → ~7s, and the warm up-to-date check (go content-hashing the dep graph) roughly halves too.

**2. Parent-branch detection.** `find-parent-release-branch.sh` runs `git merge-base` + `git rev-list` per candidate release ref (~1s), and the make target ran it via `$(shell …)` unconditionally — even with `SKIP_FIX_CHANGED=true` or no changed files.

- Cache the result in `.dev-stamps/fix-changed-parent-branch`, alongside the repo's other build-tracking stamp files (gitignored, removed by `make clean`), keyed by HEAD and the candidate ref tips so it is recomputed only when something moves. Cache hit: ~1s → ~0.03s.
- Defer the call into the `SKIP_FIX_CHANGED` guard so it no longer runs when formatting is skipped, and pass the configured `release_prefix`/`git_repo_slug` through (the old `$(shell …)` form omitted them).


fasaxc:
Thanks @copilot — addressed:

- **Exported vars vs functions** (`logutils.go`): converted `FilterLevels`, `ConfigureFormatter`, `AppendTime`, and `FormatForSyslog` from `var`-pointing-at-`logformat` into real wrapper functions, so the public API stays as immutable functions and matches the prior godoc/semantics.
- **Misleading debug log** (`find-parent-release-branch.sh`): the "Found best result" line now reports `${best_count}` (the selected match's distance) instead of `${count}` (the last ref scanned).
- **`bin/` may not exist before `go build`** (format scripts): no longer applicable — the `go build`/`bin/` build step was reverted in an earlier revision (measurements showed `go run` is equally fast now that the helper's dependency tree is slim), so the scripts are unchanged from `master`.

hack/find-parent-release-branch.sh
```diff
@@ -44,14 +44,41 @@ fi # -v CI
...
-for ref in $(git for-each-ref --format='%(refname:short)' refs/remotes/${remote} | \
-             grep --perl "${remote}/master$|${remote}/${release_prefix}[3-9]\.[2-9].*" ); do
+# The merge-base scan below runs git merge-base + git rev-list for every
+# candidate ref, which is noticeable when it runs on every "make fix-changed".
+# Cache the result, keyed by HEAD and the candidate ref tips, so we only redo
+# the scan when something has actually moved.  The cache lives in the repo's
+# .dev-stamps/ directory alongside the other build-tracking stamp files (it is
+# gitignored and removed by "make clean").
+candidate_refs=$(git for-each-ref --format='%(refname:short) %(objectname)' refs/remotes/${remote} | \
+                 grep --perl "${remote}/master |${remote}/${release_prefix}[3-9]\.[2-9].*" )
```
aaaaaaaalex, line 54: Just flagging on the off-chance is was unintended that this line has changed; I can't intuitively parse the meaning of the perl expression so I can't speculate any further :)

aaaaaaaalex (approved):
LGTM with one comment

hack/find-parent-release-branch.sh
fasaxc: `--perl` means "use perl regex format", which is nicer than grep's default regex format. The change seems to be to swap a $ for a space after "master"; this matches the change to the format above,
 
- from `%(refname:short)<end of line>`
- to `%(refname:short) %(objectname)`

So now the refname has a space after it instead of the end of line (which is matched by `$`)

PR #12842, opened 2026-05-27
ci: add optional sem-ai CI triage workflow for Semaphore failures

tomastigera:
## Description

Adds a GitHub Actions workflow that, when Semaphore CI reports a failed
build on a PR's HEAD commit, calls an external triage service and posts
(or upserts) the result as a PR comment.

The triage service inspects the failed Semaphore workflow's diagnostics,
buckets each distinct failure (`flake` / `infra` / `test-bug` /
`genuine`) and correlates them with the PR description, separating
failures the PR likely caused from pre-existing / unrelated ones. The
PR comment surfaces a one-line recommendation ("dig deeper before
merging" vs "looks safe to merge despite CI") plus a short table of
failures and evidence, saving the PR author a few minutes per failed
CI run.

## Triggers

- `status` event with `context: ci/semaphoreci/pr: Calico` and
  `state ∈ {failure, error}`.
- `workflow_dispatch` for manual re-runs (takes `workflow_id` +
  `pr_number` inputs).

The job's `if:` filters tightly so pending/success statuses and other
contexts do not trigger triage.

## Gated by secrets — safe to merge before activation

The workflow's first step short-circuits if `SEM_AI_TRIAGE_URL` and
`SEM_AI_TRIAGE_TOKEN` repo secrets are absent (emits a warning,
exits 0). This change can be merged as a no-op; an admin then
provisions the secrets when ready.

## Comment behavior

- Single comment per PR, identified by an HTML marker comment.
  Re-runs (manual or automatic) edit the existing comment rather than
  stacking new ones.
- Skipped entirely when the triage service reports "PASS" — no comment
  is posted or updated.
- Concurrency-coalesced per PR so rapid re-runs cancel in-flight
  triage instead of stacking.

## Permissions

`pull-requests: write` (upsert the comment) and `contents: read`
(read PR metadata). Commenting uses the built-in `${{ github.token }}`
— the triage service does not hold any GitHub credentials.

## Known limitations (planned follow-ups, out of scope for this PR)

- Triage costs LLM tokens per failed PR run; the service budget is
  managed by its operator. Until secrets are set on this repo, the
  workflow remains a no-op, so no production risk.
- Bucketing is heuristic; the agent is calibrated to lean
  "genuine + dig deeper" when uncertain — false reassurance is the
  worst outcome.

## Test plan

- E2E exercised against a real failing Semaphore workflow from PR #12810
  (the agent correctly flagged BPF-related test failures as PR-caused
  and surfaced an unrelated test-harness panic separately).
- PASS path verified against a passing workflow (workflow exits without
  modifying the existing comment).
- Both `workflow_dispatch` and synthetic `status: failure` events
  verified to produce identical output.

### DIFF
--- a/.github/workflows/sem-ai-ci-triage.yml
+++ b/.github/workflows/sem-ai-ci-triage.yml
@@ -22,6 +22,10 @@ on:
         required: true
         type: number
 
+# SHA-keyed for status events: a force-push gets a new SHA and runs in a
+# separate group, so the old SHA's run is not cancelled. Acceptable here —
+# Semaphore typically fires one status per (SHA, context), so the race is
+# rare and a stale comment would be overwritten on the next failure anyway.
 concurrency:
   group: sem-ai-ci-triage-${{ inputs.pr_number || github.event.sha || github.run_id }}
   cancel-in-progress: true
@@ -133,11 +137,15 @@ jobs:
             exit 0
           fi
 
+          # Tolerate non-JSON / unexpected payloads — fall through to the
+          # empty-file skip below rather than aborting the step.
+          set +e
           jq -r '
             .result.artifacts[0].parts[0].text //
             .result.message.parts[0].text //
             empty
           ' resp.json > triage.md
+          set -e
 
           if [[ ! -s triage.md ]]; then
             echo "::warning::empty or malformed triage response — skipping comment for this run"
@@ -165,7 +173,7 @@ jobs:
           PR=${{ steps.resolve.outputs.pr }}
 
           EXISTING=$(gh api "repos/$REPO/issues/$PR/comments" --paginate \
-            --jq "[.[] | select(.body | startswith(\"$MARKER\"))][0].id")
+            --jq "[.[] | select(.user.login == \"github-actions[bot]\" and (.body | startswith(\"$MARKER\")))][0].id")
 
           PAYLOAD=$(jq -n --arg b "$FULL_BODY" '{body: $b}')
 
--- a/.github/workflows/sem-ai-ci-triage.yml
+++ b/.github/workflows/sem-ai-ci-triage.yml
@@ -40,10 +40,10 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     env:
-      SEM_AI_TRIAGE_URL:   ${{ secrets.SEM_AI_TRIAGE_URL }}
+      SEM_AI_TRIAGE_URL: ${{ secrets.SEM_AI_TRIAGE_URL }}
       SEM_AI_TRIAGE_TOKEN: ${{ secrets.SEM_AI_TRIAGE_TOKEN }}
-      GH_TOKEN:            ${{ github.token }}
-      REPO:                ${{ github.repository }}
+      GH_TOKEN: ${{ github.token }}
+      REPO: ${{ github.repository }}
     steps:
       - name: Skip if secrets are not configured
         id: gate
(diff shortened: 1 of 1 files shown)


.github/workflows/sem-ai-ci-triage.yml
```diff
@@ -0,0 +1,178 @@
...
+          if [[ "$BODY" == "PASS — no comment needed" ]]; then
+            echo "PASS — no comment to post"
+            exit 0
+          fi
+
+          MARKER="<!-- sem-ai-ci-triage -->"
+          FULL_BODY="$MARKER"$'\n'"$BODY"
+          PR=${{ steps.resolve.outputs.pr }}
+
+          EXISTING=$(gh api "repos/$REPO/issues/$PR/comments" --paginate \
```
skoryk-oleksandr, line 175: Marker-only matcher can overwrite a human comment that happens to start with `<!-- sem-ai-ci-triage -->` (e.g. someone quoting the bot). Add an author check:

```bash
--jq "[.[] | select(.user.login == \"github-actions[bot]\" and (.body | startswith(\"$MARKER\")))][0].id"
```

.github/workflows/sem-ai-ci-triage.yml
```diff
@@ -0,0 +1,178 @@
...
+            --data-binary @req.json)
+          CURL_RC=$?
+          set -e
+          if [[ $CURL_RC -ne 0 || "$HTTP" != "200" ]]; then
+            echo "::warning::triage service unreachable or errored (curl_rc=$CURL_RC, http=$HTTP) — skipping comment for this run"
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          jq -r '
```
skoryk-oleksandr, line 143: `set -e` is restored here, so a non-JSON response from the triage service makes `jq` fail the whole step instead of falling through to the empty-file skip below. Either append `|| true` or wrap with `set +e` / `set -e` like the curl call above.

.github/workflows/sem-ai-ci-triage.yml
```diff
@@ -0,0 +1,178 @@
...
+          else
+            TARGET="${{ github.event.target_url }}"
+            SHA="${{ github.event.sha }}"
+            WID=$(printf '%s' "$TARGET" | sed -nE 's|.*/workflows/([0-9a-fA-F-]{36}).*|\1|p')
+            if [[ -z "$WID" ]]; then
+              echo "::warning::could not extract workflow_id from target_url='$TARGET' — skipping"
+              echo "skip=true" >> "$GITHUB_OUTPUT"
+              exit 0
+            fi
+            PR=$(gh api "repos/$REPO/commits/$SHA/pulls" \
```
skoryk-oleksandr, line 79: All `gh api` calls in the script (this one plus ~90, ~167, ~174, ~177) are single attempts; a 5xx / 429 / secondary rate-limit fails the job. Wrap them in a small retry helper (3 attempts, 2s flat backoff is plenty in practice).

.github/workflows/sem-ai-ci-triage.yml
```diff
@@ -0,0 +1,178 @@
...
+        description: "Semaphore workflow_id (UUID)"
+        required: true
+        type: string
+      pr_number:
+        description: "PR number to comment on"
+        required: true
+        type: number
+
+concurrency:
+  group: sem-ai-ci-triage-${{ inputs.pr_number || github.event.sha || github.run_id }}
```
skoryk-oleksandr, line 30: Non-blocking: SHA-keyed group means a force-push leaves the old SHA's run going in parallel with the new SHA's run. Semaphore usually only fires one status per (SHA, context) so the race rarely matters; worth either keying on PR number after the resolve step or adding a one-line code comment noting the trade-off.

skoryk-oleksandr (commented):
LGTM. Added few improvement ideas.

skoryk-oleksandr (approved):

PR #12845, opened 2026-05-27
Default kind/image build parallelism to 4

caseydavenport:
The `kind-up` / `image` / `kind-reload` targets defaulted to `-j\$(nproc)`, which spawns N docker go-build containers concurrently. On a 32G workstation with an editor and AI session co-resident, that's enough to thrash into swap.

Cap the outer parallelism at `NUM_BUILD_JOBS` (default 4). Raise via `NUM_BUILD_JOBS=N make ...` on a bigger machine. CI agents in this repo are typically 4-core so this matches existing behavior on those; bigger CI agents can crank `NUM_BUILD_JOBS` if needed.

```release-note
None
```


fasaxc (approved):
Seems like a good stop-gap

PR #12861, opened 2026-05-29
fix(felix): exclude LB-only IPPools from BPF in-pool route flag

### LINKED ISSUE #12767: SNAT is skipped for traffic destined to LoadBalancer-only IPPools shared across clusters

defo89:
## Description

IPPools with `allowedUses: [LoadBalancer]` caused `poolTypeForPool()` to return `NO_ENCAP`, which set `FlagInIPAMPool` on BPF routes and exempted traffic to LB VIPs from SNAT masquerading. This broke cross-cluster connectivity when a shared LB pool was used across clusters.

### Changes

1. **Return `IPPoolType_NONE` for LB-only pools** in `poolTypeForPool()` via a new `isLoadBalancerOnlyPool()` helper, preventing `FlagInIPAMPool` from being set on BPF routes.

2. **Remove the stale `PoolType != NONE` guard** from the pool update path. This was previously dead code (a non-nil pool never resolved to `NONE`). With the new LB-only check it must be removed to keep LB-only pools in the trie so `NATOutgoing` and `CrossSubnet` continue to propagate correctly.

3. **Unit tests** covering:
   - LB-only pool → `IpPoolType == NONE` and `NatOutgoing` still propagated
   - Mixed-use pool (Workload + LoadBalancer) → `IpPoolType == VXLAN` (not excluded)

### Context

Complements PR #12769 which fixed the iptables/ipset path (`network-ip-pools` ipset). This PR fixes the BPF dataplane path (route map `FlagInIPAMPool` flag).

### DIFF
--- a/felix/calc/l3_route_resolver.go
+++ b/felix/calc/l3_route_resolver.go
@@ -645,9 +645,6 @@ func (c *L3RouteResolver) poolTypeForPool(pool *model.IPPool) proto.IPPoolType {
 // to them should still be masqueraded (not exempted from SNAT via the
 // FlagInIPAMPool BPF route flag or the network-ip-pools ipset).
 func isLoadBalancerOnlyPool(pool *model.IPPool) bool {
-	if len(pool.AllowedUses) == 0 {
-		return false
-	}
 	return len(pool.AllowedUses) == 1 && slices.Contains(pool.AllowedUses, apiv3.IPPoolAllowedUseLoadBalancer)
 }
 
--- a/felix/calc/l3_route_resolver.go
+++ b/felix/calc/l3_route_resolver.go
@@ -644,6 +644,8 @@ func (c *L3RouteResolver) poolTypeForPool(pool *model.IPPool) proto.IPPoolType {
 // their CIDRs do not represent workload/tunnel addresses and traffic destined
 // to them should still be masqueraded (not exempted from SNAT via the
 // FlagInIPAMPool BPF route flag or the network-ip-pools ipset).
+//
+// NOTE: keep in sync with isLoadBalancerOnly in felix/dataplane/linux/masq_mgr.go.
 func isLoadBalancerOnlyPool(pool *model.IPPool) bool {
 	return len(pool.AllowedUses) == 1 && slices.Contains(pool.AllowedUses, apiv3.IPPoolAllowedUseLoadBalancer)
 }
--- a/felix/calc/l3_route_resolver.go
+++ b/felix/calc/l3_route_resolver.go
@@ -17,9 +17,11 @@ package calc
 import (
 	"fmt"
 	"reflect"
+	"slices"
 	"sort"
 	"time"
 
+	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	"github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/calico/felix/dispatcher"
@@ -588,12 +590,12 @@ func (c *L3RouteResolver) OnPoolUpdate(update api.Update) (_ bool) {
 	if update.Value != nil {
 		newPool = c.getPoolInfo(update)
 	}
-	if newPool != nil && newPool.PoolType != proto.IPPoolType_NONE {
+	if newPool != nil {
 		logrus.WithFields(logrus.Fields{
 			"oldType": oldPool.PoolType,
 			"newType": newPool.PoolType,
 			"newPool": *newPool,
-		}).Info("Pool is active")
+		}).Info("Pool update")
 		c.allPools[poolKey] = *newPool
 		c.trie.UpdatePool(newPool.CIDR, newPool.PoolType, newPool.NATOutgoing, newPool.CrossSubnet)
 	} else if oldPoolExists {
@@ -625,6 +627,9 @@ func (c *L3RouteResolver) poolTypeForPool(pool *model.IPPool) proto.IPPoolType {
 	if pool == nil {
 		return proto.IPPoolType_NONE
 	}
+	if isLoadBalancerOnlyPool(pool) {
+		return proto.IPPoolType_NONE
+	}
 	if pool.VXLANMode != encap.Never {
 		return proto.IPPoolType_VXLAN
 	}
@@ -632,6 +637,18 @@ func (c *L3RouteResolver) poolTypeForPool(pool *model.IPPool) proto.IPPoolType {
 		return proto.IPPoolType_IPIP
 	}
 	return proto.IPPoolType_NO_ENCAP
+}
+
+// isLoadBalancerOnlyPool returns true if the pool's AllowedUses contains only
+// "LoadBalancer". Such pools should not contribute to route pool type because
+// their CIDRs do not represent workload/tunnel addresses and traffic destined
+// to them should still be masqueraded (not exempted from SNAT via the
+// FlagInIPAMPool BPF route flag or the network-ip-pools ipset).
+func isLoadBalancerOnlyPool(pool *model.IPPool) bool {
+	if len(pool.AllowedUses) == 0 {
+		return false
+	}
+	return len(pool.AllowedUses) == 1 && slices.Contains(pool.AllowedUses, apiv3.IPPoolAllowedUseLoadBalancer)
 }
 
 // routesFromBlock returns a list of routes which should exist based on the provided
--- a/felix/calc/l3_route_resolver_test.go
+++ b/felix/calc/l3_route_resolver_test.go
@@ -17,6 +17,7 @@ package calc
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	v3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 
 	"github.com/projectcalico/calico/felix/ip"
 	"github.com/projectcalico/calico/felix/proto"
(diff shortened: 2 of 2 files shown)


defo89:
Reproducer to confirm current behaviour in BPF dataplane.

### 1. Create a kind cluster

```bash
cat <<EOF | kind create cluster --name repro-bpf-12767 --config -
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  disableDefaultCNI: true
  podSubnet: "192.168.0.0/16"
  serviceSubnet: "10.96.0.0/12"
nodes:
- role: control-plane
- role: worker
EOF
```

### 2. Install Calico operator

```bash
kubectl create -f https://raw.githubusercontent.com/projectcalico/calico/v3.30.7/manifests/tigera-operator.yaml --namespace=tigera-operator
kubectl wait --for=condition=Available deployment/tigera-operator -n tigera-operator --timeout=120s
```

### 3. Install Calico with BPF dataplane

```bash
kubectl apply -f - <<EOF
apiVersion: operator.tigera.io/v1
kind: Installation
metadata:
  name: default
spec:
  calicoNetwork:
    ipPools:
    - cidr: 192.168.0.0/16
      encapsulation: VXLANCrossSubnet
      natOutgoing: Enabled
    linuxDataplane: BPF
EOF

kubectl wait --for=condition=Ready pods -n calico-system -l k8s-app=calico-node --timeout=300s
```

### 4. Create a LoadBalancer-only IPPool

```bash
kubectl apply -f - <<EOF
apiVersion: crd.projectcalico.org/v1
kind: IPPool
metadata:
  name: lb-pool
spec:
  cidr: 10.20.40.64/27
  ipipMode: Never
  vxlanMode: Never
  natOutgoing: false
  allowedUses:
  - LoadBalancer
  assignmentMode: Manual
EOF
```

### 5. Check BPF route map for LB pool CIDR

```bash
CALICO_NODE=$(kubectl get pods -n calico-system -l k8s-app=calico-node -o jsonpath='{.items[0].metadata.name}')
kubectl exec -n calico-system "${CALICO_NODE}" -c calico-node -- calico-node -bpf routes dump | grep "10.20.40"
```

## Current Result

The LB pool has `in-pool` flag (`FlagInIPAMPool`) set, which means traffic to LB VIPs in `10.20.40.64/27` will skip masquerading.

```
$ kubectl exec -n calico-system "${CALICO_NODE}" -c calico-node -- calico-node -bpf routes dump
 10.20.40.64/27: remote in-pool   <<<<
  172.18.0.3/32: local host
  172.18.0.4/32: remote host
 192.168.0.0/16: remote in-pool nat-out
```

## Expected Result

The route for `10.20.40.64/27` should show `remote` **without** the `in-pool` flag, or the CIDR should not appear in the BPF route map at all. 

`nat-out` flag is not relevant here.

felix/calc/l3_route_resolver.go
```diff
@@ -634,6 +639,18 @@ func (c *L3RouteResolver) poolTypeForPool(pool *model.IPPool) proto.IPPoolType {
...
 
+// isLoadBalancerOnlyPool returns true if the pool's AllowedUses contains only
+// "LoadBalancer". Such pools should not contribute to route pool type because
+// their CIDRs do not represent workload/tunnel addresses and traffic destined
+// to them should still be masqueraded (not exempted from SNAT via the
+// FlagInIPAMPool BPF route flag or the network-ip-pools ipset).
+func isLoadBalancerOnlyPool(pool *model.IPPool) bool {
+	if len(pool.AllowedUses) == 0 {
+		return false
+	}
```
mazdakn: we can remove this check, since it's covered with the next one.

felix/calc/l3_route_resolver.go
```diff
@@ -634,6 +639,18 @@ func (c *L3RouteResolver) poolTypeForPool(pool *model.IPPool) proto.IPPoolType {
 	return proto.IPPoolType_NO_ENCAP
 }
 
+// isLoadBalancerOnlyPool returns true if the pool's AllowedUses contains only
+// "LoadBalancer". Such pools should not contribute to route pool type because
+// their CIDRs do not represent workload/tunnel addresses and traffic destined
+// to them should still be masqueraded (not exempted from SNAT via the
+// FlagInIPAMPool BPF route flag or the network-ip-pools ipset).
+func isLoadBalancerOnlyPool(pool *model.IPPool) bool {
```
mazdakn, line 649: Please add a comment that this should be in sync with the one defined in `masq_mgr.go` added in the previous PR.

felix/calc/l3_route_resolver_test.go
```diff
@@ -287,6 +288,111 @@ var _ = Describe("L3RouteResolver", func() {
...
+				Key: model.IPPoolKey{CIDR: model.PrefixFromIPNet(net.IPNet{IPNet: lbPoolCIDR.ToIPNet()})},
+				Value: &model.IPPool{
+					CIDR:        net.IPNet{IPNet: lbPoolCIDR.ToIPNet()},
+					Masquerade:  true,
+					AllowedUses: []v3.IPPoolAllowedUse{v3.IPPoolAllowedUseLoadBalancer},
+				},
+			},
+		})
+
+		remoteAffinity := "host:remote-host"
```
mazdakn: nit:

```suggestion
		remoteAffinity := "virtual:loadbalancer"
```

mazdakn: does not matter, just make more sense.

mazdakn (commented):
Overall LGTM. Left a few nits.

mazdakn:
@defo89 have you managed to verify the fix actually works in bpf? Asking since you posted steps for re-producing the issue in bpf.

defo89:
> @defo89 have you managed to verify the fix actually works in bpf? Asking since you posted steps for re-producing the issue in bpf.

Fair point, I plan to perform some testing using local setup tomorrow.

mazdakn:
@defo89 thanks. Please let me know the result. Otherwise, we should be good to merge this one.

defo89:
Here is the testing scenario using local setup from above.

### Create Test Pod

```yaml
kubectl apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: snat-test
  namespace: default
spec:
  containers:
  - name: net
    image: nicolaka/netshoot
    command: ["sleep", "infinity"]
EOF

kubectl wait --for=condition=Ready pod/snat-test --timeout=60s
```

### Start tcpdump on the worker node to see SNAT

 ```bash
docker exec -it repro-bpf-12767-worker apt-get update && docker exec -it repro-bpf-12767-worker apt-get install -y tcpdump
docker exec -it repro-bpf-12767-worker tcpdump -nn -i eth0 'dst net 10.20.40.64/27'
 ```

### Generate traffic to an IP in the LB pool range

Pick an IP from the LB pool CIDR (10.20.40.64/27), e.g. 10.20.40.65:

```bash
# From the test pod, send traffic to an IP in the LB pool
kubectl exec snat-test -- curl -s --connect-timeout 3 http://10.20.40.65:80
```

### Results - Bug present (current)

Source IP is the pod IP (192.168.x.x) — no SNAT happened because in-pool flag exempts this traffic from masquerading.

```
10:19:31.721909 IP 192.168.82.199.39328 > 10.20.40.65.80: Flags [S], seq 2258391946, win 64860, options [mss 1410,sackOK,TS val 2700109389 ecr 0,nop,wscale 7], length 0
```

### Results - node and typha images build with this PR

Source IP is the node IP (172.18.0.x) — SNAT applied correctly.

```
12:03:01.320169 IP 172.18.0.3.3691 > 10.20.40.65.80: Flags [S], seq 3567599558, win 64860, options [mss 1410,sackOK,TS val 2706318987 ecr 0,nop,wscale 7], length 0
```

There is a small diff from the expectation - the ippool is not even present in the bpf route map
```
bash-5.1$    kubectl exec -n calico-system "${CALICO_NODE}" -c calico-node -- calico component node bpf routes dump
  172.18.0.2/32: remote host
  172.18.0.3/32: local host
 192.168.0.0/16: remote in-pool nat-out
...
```

And from what I read this is expected since the LB CIDR 10.20.40.64/27 has:
 - No host entries
 - No blocks allocated
 - No workload/tunnel refs
 - PoolType == NONE (doesn't set in-pool)

So it looks ok to me, wdyt?

mazdakn:
@defo89 thanks for testing it out. Look correct.

mazdakn (approved):
LGTM

mazdakn:
@defo89 thanks for the fix. Please pick this one as well to v3.32 and v3.32 release branches.

PR #12874, opened 2026-05-29
[v3.32] [BPF] Fix node-unreachable on Felix restart with no Node

### LINKED ISSUE #12642: BPF dataplane: kubectl delete node without graceful Calico shutdown causes node to lose connectivity

tomastigera:
## Cherry-pick history
- Pick onto **release-v3.32**: projectcalico/calico#12810

Cherry-pick of #12810. Conflict resolution notes:
- `felix/design/bpf-tests.md` is omitted — the `felix/design/` doc split does not exist on `release-v3.32`.
- `felix/dataplane/linux/bpf_ep_mgr.go`: on `release-v3.32` host metadata arrives as three separate proto messages (`HostMetadataUpdate` v4-only, `HostMetadataV6Update` v6-only, `HostMetadataV4V6Update` combined) rather than the single consolidated `HostMetadataUpdate` on master. All three call sites were updated to the refactored `updateOurHostIP(parseHostIP(...))` form.

Verified `make -C felix build` and `make -C felix build-bpf` both pass.

---

Fixes #12642 — in BPF mode, after `kubectl delete node <self>` followed by a Felix restart, the node loses all network connectivity (no SSH, no etcd/apiserver) and cannot recover on its own. See the original PR #12810 for full mechanism, approach, and test details.

**Release note:**
```release-note
Fixes a bug in the eBPF dataplane in which deleting and restoring the local Node resource and restarting Felix could leave the node unable to handle network traffic.
```


tomastigera:
## On-cluster verification (v3.32, eBPF dataplane)

Built `calico/node` from this PR's tip (`93ea476`), deployed to a 4-node v3.32 cluster and reproduced the #12642 scenario. **Result: fix confirmed.**

**Environment**
- Operator-managed, eBPF dataplane (`bpfEnabled: true`), IPv4-only, 1 control-plane + 3 workers
- kubeadm on GCP, kernel `6.8.0-1060-gcp` (matches the PR's original validation kernel)
- All 4 `calico-node` pods rolled onto the PR image cleanly and went Ready

**Healthy on the new image**
- New `BPFHostIP` readiness reporter present and ready; `/readiness` = 200
- New `DroppedNoHostIP` BPF counter present
- Dataplane functional (pod + host connectivity)

**Reproduction of #12642** — on a worker: `kubectl delete node <worker>` (GCs the calico-node pod → fresh Felix restarts with no Node / no host IP — the exact bug trigger). A 1 Hz probe from a pod on another node hit the target node's host IP (TCP/22 + ICMP) throughout.

- **Host connectivity preserved: 0 failures / 150 samples** across the whole window (delete → fresh Felix → recovery). Pre-fix, this state drops *all* traffic through the node's `eth0` (preambles tail-calling into freed jump-map slots).
- During the no-host-IP window: `/readiness` = **503** with

  ```
  | BPFHostIP           | reporting non-ready | Host IP not yet known. |
  | BPFEndpointManager  | reporting ready     |                        |
  ```

  i.e. programs attached with `HOST_IP=0` (gate removed) while the new reporter correctly flags the missing IP.
- Fresh-Felix log shows the full arc:
  ```
  felix/maps.go 611:   Repinning BPF map ... cali_v4_mglv2 -> ..._old
  felix/health.go 203: Health of component changed name="BPFHostIP" newReport="non-ready"
  ... (Node re-registered) ...
  felix/bpf_ep_mgr.go 969: Host IP changed: 10.128.0.215/32   HostMetadataV4V6Update=...
  felix/bpf_ep_mgr.go 945: Host IP changed: 10.128.0.215       HostMetadataUpdate=...
  felix/health.go 203: Health of component changed name="BPFHostIP" newReport="ready"
  ```
- **Recovery:** once the host IP returned, `updateOurHostIP` marked everything dirty, programs re-attached with the real `HOST_IP`, `BPFHostIP` flipped back to ready and `/readiness` returned 200.

The two `Host IP changed` call sites above (`bpf_ep_mgr.go:945` `HostMetadataUpdate` v4-only and `:969` `HostMetadataV4V6Update`) are the v3.32-specific three-way host-metadata split this backport had to adapt to vs. master's single consolidated case — both exercised and working at runtime.

**Not exercised:** `DroppedNoHostIP` counter *increment* via the cross-node VXLAN encap-source drop (the `bpf_dual_stack` path); the counter itself is confirmed present.

PR #12879, opened 2026-05-30
Make whisker's flow source pluggable

hjiawei:
## Description

Lets whisker-backend serve flows from a source other than Goldmane. Nothing changes for a Calico install: Goldmane is still the only source, wired as before.

`FlowsBackend` (`List`/`FilterHints`, plus `StreamingFlowsBackend` for watch) returns the `FlowResponse` wire type directly, so an upstream converts its own types with no round trip through proto. The Goldmane client and converter move to `pkg/upstream/goldmane`; the handlers no longer know where flows come from. `pkg/auth` holds the flow-filter contract the handlers apply when a factory is configured; none is wired here.

`FlowResponse` gains a `service` field, filled from the Goldmane flow key.

## Testing

Unit tests: handlers against a mock backend, the Goldmane backend's parameter conversion and filter rejection, proto-to-flow conversion, and the filter plumbing.

First consumer is Calico Enterprise, in tigera/calico-private#13686.

**Release note:**
```release-note
Whisker flow responses include the destination service when the flow targets one.
```

**AI assistance:** Claude Code helped write the abstraction, the tests and this description; every change was reviewed and validated by the author.


whisker-backend/cmd/whiskerbackend/app.go
```diff
@@ -75,3 +96,30 @@ func Run(ctx context.Context, cfg *config.Config) {
 		logrus.WithError(err).Fatal("An unexpected error occurred while waiting for shutdown.")
 	}
 }
+
+// newGoldmaneBackend builds the Goldmane flow backend: the only upstream in OSS
```
caseydavenport: Comment is on the wrong function, and shouldn't be referring to enterprise code.

whisker-backend/pkg/apis/v1/backend.go
```diff
@@ -0,0 +1,61 @@
...
+//
+// Streaming (watch) is an optional capability: upstreams that can stream flows
+// additionally implement StreamingFlowsBackend. The handler detects support at
+// runtime and rejects watch requests for upstreams that lack it, rather than
+// every backend being forced to provide a no-op Stream.
+type FlowsBackend interface {
+	// List returns one page of flows and the page and result totals across all
+	// pages. Pages are 0-based; a non-positive page size means everything in one
+	// page.
+	List(ctx context.Context, params ListFlowsParams) (meta apiutil.ListMeta, flows []FlowResponse, err error)
```
caseydavenport, line 33:
```suggestion
	List(ctx context.Context, params ListFlowsParams) (meta apiutil.ListMeta, flows []FlowResponse, err error)
	
```

whisker-backend/pkg/apis/v1/backend.go
```diff
@@ -0,0 +1,61 @@
...
+package v1
+
+import (
+	"context"
+
+	"github.com/projectcalico/calico/lib/httpmachinery/pkg/apiutil"
+)
+
+// FlowsBackend abstracts the flow data source so the handler is independent of
+// the upstream (Goldmane, Linseed, etc.). Each upstream implements this interface
```
caseydavenport: Probably overcommented, and referring to enterprise components is a no-no.

whisker-backend/pkg/apis/v1/flows.go
```diff
@@ -268,6 +268,22 @@ type FlowResponse struct {
 	PacketsOut      int64       `json:"packets_out"`
 	BytesIn         int64       `json:"bytes_in"`
 	BytesOut        int64       `json:"bytes_out"`
+
+	// Service is the destination service the flow targets, if any. Populated by
+	// both the Goldmane (OSS) and Linseed upstreams.
```
caseydavenport:
```suggestion
	// Service is the destination service the flow targets, if any.
```

whisker-backend/pkg/upstream/goldmane/convert.go
```diff
@@ -163,6 +163,23 @@ func protoToFlow(flow *proto.Flow) whiskerv1.FlowResponse {
...
+			Port:      flow.Key.DestServicePort,
+			PortName:  flow.Key.DestServicePortName,
+		}
+	}
+
+	// NOTE: SourceType/DestType are intentionally left unset — the Goldmane proto
+	// flow does not carry endpoint types. These fields are only consumed by the
+	// RBAC FlowFilter, which is currently wired only on the Linseed path. If RBAC
+	// is ever enabled for the Goldmane upstream, the empty types will be treated
+	// as deny-and-skip by auth.rbacFlowFilter (see whisker-backend/pkg/auth/rbac.go).
```
caseydavenport: Goldmane does carry them - [`source_type` and `dest_type`](https://github.com/projectcalico/calico/blob/6ae3dedbfaa2649af3c35b301c975fab1b4dcf3f/goldmane/proto/api.proto#L343-L360) are on `FlowKey`, and Felix populates both from the flow log's endpoint metadata in [`ConvertFlowlogToGoldmane`](https://github.com/projectcalico/calico/blob/6ae3dedbfaa2649af3c35b301c975fab1b4dcf3f/felix/collector/goldmane/client.go#L119-L136). We should fill them in here rather than leave the Goldmane path with empty values that the RBAC filter treats as a deny.

On the values themselves: the field doc says they match the linseed constants (`wep`, `hep`, `ns`, `net`), but those are Felix's internal names, which [`convertType`](https://github.com/projectcalico/calico/blob/6ae3dedbfaa2649af3c35b301c975fab1b4dcf3f/felix/collector/goldmane/client.go#L79-L94) already maps into the Goldmane enum on the way in. We should use the Goldmane names here (`WorkloadEndpoint`, `HostEndpoint`, `NetworkSet`, `Network`) and convert on the linseed side instead.

whisker-backend/pkg/auth/filter.go
```diff
@@ -0,0 +1,126 @@
...
+	"context"
+
+	whiskerv1 "github.com/projectcalico/calico/whisker-backend/pkg/apis/v1"
+)
+
+// FlowFilter decides whether a flow should be visible to the current user, and
+// redacts the parts of an admitted flow the user is not authorized to view.
+type FlowFilter interface {
+	IncludeFlow(flow *whiskerv1.FlowResponse) (bool, error)
+	// RedactPolicies obfuscates, in place, the policy hits on an admitted flow
```
caseydavenport, line 28:
```suggestion

	// RedactPolicies obfuscates, in place, the policy hits on an admitted flow
```

whisker-backend/pkg/auth/filter.go
```diff
@@ -0,0 +1,126 @@
...
+			if err := filter.RedactPolicies(&flows[i]); err != nil {
+				return nil, err
+			}
+			result = append(result, flows[i])
+		}
+	}
+	return result, nil
+}
+
+// FilterFlowsFromUser is a convenience that extracts the user from the context,
```
caseydavenport: Comment seems wrong - this function isn't doing anything with the given context.

whisker-backend/pkg/handlers/v1/flows.go
```diff
@@ -85,35 +107,37 @@ func (hdlr *flowsHdlr) ListOrStream(ctx apictx.Context, params whiskerv1.ListFlo
...
+	if err != nil {
+		logger.Error("failed to filter flows by RBAC", "error", err)
+		return apiutil.NewListOrStreamResponse[whiskerv1.FlowResponse]().SetStatus(http.StatusInternalServerError).SetError("Internal Server Error")
+	}
+
+	// A single-page result the RBAC filter emptied is no pages at all —
+	// backends already report 0 for an empty result, and the UI must not be
+	// told there is a page to render with nothing in it. Multi-page counts are
+	// left alone: they are real pagination state from the backend.
+	if len(flows) == 0 && meta.TotalPages == 1 {
```
caseydavenport, line 135: Just curious, do we need the `meta.TotalPages` check here at all? Doesn't this hold true for any `len(flows) == 0` ?

whisker-backend/pkg/handlers/v1/flows.go
```diff
@@ -85,35 +107,37 @@ func (hdlr *flowsHdlr) ListOrStream(ctx apictx.Context, params whiskerv1.ListFlo
...
+		meta = apiutil.ListMeta{}
 	}
+
+	return apiutil.NewListOrStreamResponse[whiskerv1.FlowResponse]().SetStatus(http.StatusOK).
+		SendList(meta, flows)
```
caseydavenport:
```suggestion
	return apiutil.NewListOrStreamResponse[whiskerv1.FlowResponse]().SetStatus(http.StatusOK).SendList(meta, flows)
```

whisker-backend/pkg/upstream/goldmane/backend.go
```diff
@@ -0,0 +1,120 @@
...
+func NewBackend(cli client.FlowsClient) *Backend {
+	return &Backend{cli: cli}
+}
+
+func (b *Backend) List(ctx context.Context, params whiskerv1.ListFlowsParams) (int, []whiskerv1.FlowResponse, error) {
+	flowReq := &proto.FlowListRequest{
+		SortBy:       toProtoSortByOptions(params.SortBy),
+		Filter:       toProtoFilter(params.Filters),
+		StartTimeGte: params.StartTimeGte,
+		StartTimeLt:  params.StartTimeLt,
```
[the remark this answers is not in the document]
caseydavenport: Seems legit?

whisker-backend/pkg/upstream/goldmane/backend.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return &goldmaneFlowStream{stream: stream}, nil
+}
+
+// FilterHints delegates to Goldmane's own FilterHints API, which returns hint
+// values directly rather than source flows, so a flow-level filter predicate
+// cannot be applied here. Rather than silently returning unfiltered hints, an
+// error is returned if one is ever supplied — the RBAC flow filter is only
+// wired for the (enterprise) Linseed upstream today, so includeFlow is nil in
+// practice.
```
caseydavenport:
```suggestion
```

whisker-backend/pkg/upstream/goldmane/backend.go
```diff
@@ -0,0 +1,133 @@
...
+func (b *Backend) FilterHints(ctx context.Context, params whiskerv1.FlowFilterHintsRequest, includeFlow whiskerv1.FlowFilterFunc) (apiutil.ListMeta, []whiskerv1.FlowFilterHintResponse, error) {
+	if includeFlow != nil {
+		return apiutil.ListMeta{}, nil, fmt.Errorf("the Goldmane upstream does not support flow filtering for filter hints")
```
caseydavenport, line 92:
```suggestion
        // Goldmane doesn't support filter functions, so error explicitly. 
		return apiutil.ListMeta{}, nil, fmt.Errorf("the Goldmane upstream does not support flow filtering for filter hints")
```

caseydavenport (changes requested):
A few thoughts, LMK what you think!

whisker-backend/pkg/handlers/v1/flows.go
hjiawei: Not quite — it holds for a single page, but not for every empty result.

Two multi-page cases need the totals kept. A page past the end: Goldmane returns the real totals there with `items: null`, because `calculateListMeta` keys off the total match count, and its `0/0` means "no matches at all" rather than "this page is empty". And a page a filter emptied while other pages still have visible flows. Zeroing on any `len(flows) == 0` would tell the UI there are no results in either case.

Only the single-page case means there is nothing to show at all. I've rewritten the comment to say that, since the old wording is what invited the question.

whisker-backend/pkg/upstream/goldmane/convert.go
hjiawei: Agreed on both counts, fixed.

The converter now fills `SourceType`/`DestType` from `FlowKey`, and the field doc names Goldmane's enum values (`WorkloadEndpoint`, `HostEndpoint`, `NetworkSet`, `Network`) rather than the Linseed ones. A type the upstream did not report stays empty, so it is not mistaken for a real one. `TestProtoToFlow_EndpointTypes` pins the mapping.

Since that makes the Goldmane names the contract, the Linseed upstream in the downstream fork converts on its own side in both directions — into these names when it builds the response, and back to its own values when it hands an endpoint to the visibility check. Both switch on `proto.EndpointType_*.String()` rather than string literals, so they cannot drift from the enum, and each direction has a test pinning it.

whisker-backend/pkg/upstream/goldmane/backend.go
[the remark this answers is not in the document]
hjiawei: Partly. `TestBackendListParameterConversion` and `TestBackendStreamParameterConversion` already assert the proto requests for the list and stream paths, so those two are covered. FilterHints was not — it only had the test for rejecting a filter predicate, and nothing checked the request it builds. Added `TestBackendFilterHintsParameterConversion` for it.

whisker-backend/pkg/apis/v1/flows.go
```diff
@@ -268,6 +268,22 @@ type FlowResponse struct {
...
 	BytesIn         int64       `json:"bytes_in"`
 	BytesOut        int64       `json:"bytes_out"`
+
+	// Service is the destination service the flow targets, if any.
+	Service *ServiceRef `json:"service,omitempty"`
+
+	// Endpoint types for source and destination, named as Goldmane's EndpointType
+	// enum does: WorkloadEndpoint, HostEndpoint, NetworkSet, Network. Empty when
+	// the upstream did not report one.
+	SourceType string `json:"-"`
```
caseydavenport: Any reason not to have json tags on these?

whisker-backend/pkg/handlers/v1/flows.go
```diff
@@ -57,21 +69,31 @@ func (hdlr *flowsHdlr) ListOrStream(ctx apictx.Context, params whiskerv1.ListFlo
...
 	if params.Watch {
 		logger.Debug("Watch is set, streaming flows...")
 		// TODO figure out how we're going to handle errors.
-		flowReq := &proto.FlowStreamRequest{
-			Filter:       filter,
-			StartTimeGte: params.StartTimeGte,
+
+		// Streaming is optional. Backends that cannot stream (e.g. Linseed)
+		// implement only FlowsBackend; reject the watch request with a clear
+		// error instead of a generic 500.
```
caseydavenport:
```suggestion
		// Streaming is optional. Backends that cannot stream
		// implement only FlowsBackend; reject the watch request with a clear
		// error instead of a generic 500.
```

caseydavenport (changes requested):
@hjiawei two more trivial comments, and some merge conflicts - but otherwise LGTM!

caseydavenport (approved):
Approving so I don't need to look again 😁

PR #12884, opened 2026-06-01
Fix CI flakes from Windows build downloads and built-in tier protection

caseydavenport:
A handful of CI flakes that showed up on https://github.com/projectcalico/calico/pull/12794, all pre-existing and unrelated to that PR's changes. The last two are fallout from #12824 landing on master.

**Windows archive build.** `make build-windows-archive` intermittently fails downloading `nssm.zip` (and friends) with `curl: (92) HTTP/2 stream 0 was not closed cleanly`. These downloads already had `--retry 5`, but plain `--retry` doesn't cover a mid-transfer HTTP/2 stream reset. Force HTTP/1.1 and add `--retry-all-errors`.

**libcalico-go tier watch test.** The "Tier watch functionality" test calls `Clean()` then asserts the datastore is empty. Since #12026 added a ValidatingAdmissionPolicy protecting the built-in tiers (default, kube-admin, kube-baseline), `Clean()` can't delete them, so they persist once any earlier spec creates them via `EnsureInitialized`. Ginkgo randomizes top-level container order, so whether they're present depends on the seed, which is the flake. The test now filters built-in tiers out of its list and snapshot-watch assertions, and `Clean()` skips them rather than burning its full retry budget failing to delete them.

**libcalico-go tier validation tests.** Same root cause. The validation tests create the built-in tiers and expect success. Now that they can't be deleted (#12824), `Clean()` leaves them around between tests, so the create comes back as `AlreadyExists`. That still means validation admitted the spec, so the test tolerates it for the built-in tiers.

**calicoctl datastore migrate import.** Same root cause. The import precondition refuses to run when the target datastore already has Calico resources, and the built-in tiers now count even though they're created automatically and can't be deleted. In practice the apiserver is usually installed after the migration so a real import target won't have them yet, but the check shouldn't treat them as user data regardless. Import now skips the built-in tiers, the same way it already skips Kubernetes-backed network policies.

```release-note
None
```

### DIFF
--- a/libcalico-go/lib/backend/k8s/client.go
+++ b/libcalico-go/lib/backend/k8s/client.go
@@ -46,6 +46,7 @@ import (
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/k8s/resources"
 	"github.com/projectcalico/calico/libcalico-go/lib/backend/model"
 	cerrors "github.com/projectcalico/calico/libcalico-go/lib/errors"
+	"github.com/projectcalico/calico/libcalico-go/lib/names"
 	"github.com/projectcalico/calico/libcalico-go/lib/set"
 	"github.com/projectcalico/calico/libcalico-go/lib/winutils"
 )
@@ -433,6 +434,14 @@ func (c *KubeClient) Clean() error {
 					return nil // Problems are reported through kindsWithProblems set.
 				} else {
 					for _, r := range rs.KVPairs {
+						// The built-in tiers cannot be deleted - they are protected by a
+						// ValidatingAdmissionPolicy. Skip them, otherwise every Clean() burns
+						// its full retry budget failing to delete them and logs spurious errors.
+						if k == apiv3.KindTier {
+							if rk, ok := r.Key.(model.ResourceKey); ok && names.TierIsStatic(rk.Name) {
+								continue
+							}
+						}
 						delEG.Go(func() error {
 							if _, err := c.DeleteKVP(ctx, r); err != nil {
 								log.WithError(err).WithField("Key", r.Key).Warning("Failed to delete entry from KDD")
--- a/libcalico-go/lib/backend/k8s/client.go
+++ b/libcalico-go/lib/backend/k8s/client.go
@@ -434,11 +434,13 @@ func (c *KubeClient) Clean() error {
 					return nil // Problems are reported through kindsWithProblems set.
 				} else {
 					for _, r := range rs.KVPairs {
-						// The built-in tiers cannot be deleted - they are protected by a
-						// ValidatingAdmissionPolicy. Skip them, otherwise every Clean() burns
-						// its full retry budget failing to delete them and logs spurious errors.
+						// The protected built-in tiers cannot be deleted - a
+						// ValidatingAdmissionPolicy blocks it. Skip them, otherwise every Clean()
+						// burns its full retry budget failing to delete them and logs spurious
+						// errors. Don't skip the deletable static tiers (anp/banp), or they leak
+						// into later specs.
 						if k == apiv3.KindTier {
-							if rk, ok := r.Key.(model.ResourceKey); ok && names.TierIsStatic(rk.Name) {
+							if rk, ok := r.Key.(model.ResourceKey); ok && names.TierIsProtected(rk.Name) {
 								continue
 							}
 						}
--- a/libcalico-go/lib/clientv3/tier_e2e_test.go
+++ b/libcalico-go/lib/clientv3/tier_e2e_test.go
@@ -38,6 +38,27 @@ func init() {
 	format.MaxLength = 0
 }
 
+// staticTierNames are the built-in tiers that always exist and cannot be deleted
+// (they are protected by a ValidatingAdmissionPolicy and recreated by
+// EnsureInitialized). Tests that snapshot all tiers must filter these out, since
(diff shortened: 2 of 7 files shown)


electricjesus (commented):
Heads up, CI's red here but I don't think it's your diff's fault.

The big one: the kube-admin/kube-baseline `Tier validation tests` fail with `tiers "kube-admin" already exists`. Those entries create a built-in tier, but per your own new comment those always exist and can't be deleted, so the create hits `already exists` and the BeforeEach Clean() can't reset them. It's the validation-side version of the same thing the watch test was hitting. They're enabled in OSS (we keep them commented out in enterprise), so they'll block the merge here even though they're not really about your change. Could have them tolerate `already exists`, or just create under a throwaway name.

The other failure is calicoctl TestDatastoreMigrationIPAM. Is that the flake you were after, or a separate thing?

The watch fix itself looks right to me, fwiw. I hit the same red from the calico-private side (the be.Clean() leaving the default tier) and traced it to the same place. Mostly just keen to get it merged so it picks into cp.

libcalico-go/lib/backend/k8s/client.go
```diff
@@ -433,6 +434,14 @@ func (c *KubeClient) Clean() error {
 					return nil // Problems are reported through kindsWithProblems set.
 				} else {
 					for _, r := range rs.KVPairs {
+						// The built-in tiers cannot be deleted - they are protected by a
+						// ValidatingAdmissionPolicy. Skip them, otherwise every Clean() burns
+						// its full retry budget failing to delete them and logs spurious errors.
+						if k == apiv3.KindTier {
+							if rk, ok := r.Key.(model.ResourceKey); ok && names.TierIsStatic(rk.Name) {
```
electricjesus: `names.TierIsStatic` matches five tiers, including `adminnetworkpolicy` and `baselineadminnetworkpolicy`. Those two aren't covered by the ValidatingAdmissionPolicy and are deletable, so skipping them here means `Clean()` leaves them behind and they leak into later specs. I'd scope this to just the three protected tiers (default, kube-admin, kube-baseline) rather than `TierIsStatic`.

libcalico-go/lib/backend/k8s/client.go
```diff
@@ -433,6 +434,14 @@ func (c *KubeClient) Clean() error {
 					return nil // Problems are reported through kindsWithProblems set.
 				} else {
 					for _, r := range rs.KVPairs {
+						// The built-in tiers cannot be deleted - they are protected by a
```
electricjesus: Small thing on the comment: it says the static tiers are protected by the ValidatingAdmissionPolicy, but that's only true for three of the five `TierIsStatic` names. Worth tightening the wording, or narrowing the check to match it.

libcalico-go/lib/clientv3/tier_e2e_test.go
```diff
@@ -622,7 +648,7 @@ var _ = testutils.E2eDatastoreDescribe("Tier tests", testutils.DatastoreAll, fun
 			By("Starting a watcher not specifying a rev - expect the current snapshot")
 			w, err = c.Tiers().Watch(ctx, options.ListOptions{})
 			Expect(err).NotTo(HaveOccurred())
-			testWatcher3 := testutils.NewTestResourceWatch(config.Spec.DatastoreType, w)
+			testWatcher3 := testutils.NewTestResourceWatch(config.Spec.DatastoreType, w, staticTierNames...)
```
electricjesus: The watcher ignore list uses `staticTierNames` (three names), but `filterStaticTiers` above uses `TierIsStatic` (five). If anp or banp ever land in this no-rev snapshot they won't get filtered, which makes the test order-dependent. Probably want both paths driven off the same set. (Same applies to testWatcher4 at :674.)

libcalico-go/lib/clientv3/tier_e2e_test.go
```diff
@@ -38,6 +38,27 @@ func init() {
 	format.MaxLength = 0
 }
 
+// staticTierNames are the built-in tiers that always exist and cannot be deleted
+// (they are protected by a ValidatingAdmissionPolicy and recreated by
+// EnsureInitialized). Tests that snapshot all tiers must filter these out, since
+// whether they are present depends on which other specs in the suite have run.
+var staticTierNames = []string{
```
electricjesus: `staticTierNames` (3) and `TierIsStatic` (5) overlap but don't line up. Might be cleaner to define one and derive the other so they can't drift.

node/Makefile
```diff
@@ -306,7 +306,7 @@ $(BIRD_SOURCE): .bird-source.created
 .bird-source.created:
 	rm -rf filesystem/included-source/bird*
 	mkdir -p filesystem/included-source/
-	curl -sSf -L --retry 5 -o $(BIRD_SOURCE) https://github.com/projectcalico/bird/tarball/$(BIRD_VERSION)
+	curl -sSf -L --retry 5 --retry-all-errors --http1.1 -o $(BIRD_SOURCE) https://github.com/projectcalico/bird/tarball/$(BIRD_VERSION)
```
electricjesus, line 309: This curl retry change for the bird/Windows downloads is unrelated to the tier fix. No objection to it, just worth keeping as its own commit so cherry-picks stay clean.

electricjesus (approved):

PR #12887, opened 2026-06-02
BPF FV: make host-IP tests robust to health/counter table wrapping

tomastigera:
Makes two BPF host-IP FV tests robust to ASCII-table line wrapping.

The node-delete and dual-stack host-IP FV tests (added in #12810) assert on substrings that the health/CLI tables can wrap across lines depending on the rendered column widths:

- `bpf_node_delete_test.go` matched the contiguous `"IPv4 unknown since"`, but the `/readiness` health table can wrap the long `BPFHostIP` detail between `"IPv4 unknown"` and `"since <timestamp>"`. Now matched with a regex that tolerates the intervening padding and `|` borders.
- `readDroppedNoHostIPCounter` required an exact match of the 32-char `"VXLAN encap when host IP unknown"` counter name, which wraps in the `calico-bpf counters dump` TYPE column. Now matched on the unique prefix that survives on the first line (where the counts are); this also matches the unwrapped form.

No product behaviour change — only the test parsing is made robust. These assertions fail deterministically on the release-v3.32 cherry-pick CI's table widths (projectcalico/calico#12874); this is the same fix, ported to master so the fragility can't resurface here.

Verified by building the FV test binary (`make -C felix fv/fv.test`); the underlying behaviour was confirmed correct by running the affected BPF FVs locally on the v3.32 branch.

**Release note:**
```release-note
None
```

### DIFF
--- a/.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
+++ b/.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
@@ -201,18 +201,29 @@
   # producer for the "calico" build-cache group — every Go component block on
   # master restores the module + build cache populated here.
   #
-  # cmd/calico's Go closure already covers felix/typha (registered as
-  # subcommands); the non-go paths close the gap for Felix Build/UT, which
-  # does untolerated gcloud cp of calico-cgo/calico-race/libbpf. Other
-  # consumers use .semaphore/load-cached-images, which falls back to a local
-  # build on cache miss.
+  # felix MUST be the primary (first) CHANGE_IN arg so this producer is a strict
+  # superset of its consumers. The Felix Build/UT/FV and Windows FV blocks all
+  # trigger on CHANGE_IN(felix,...) and then do an untolerated gcloud cp of
+  # calico-cgo/calico-race/libbpf (or load the cached calico image), so this
+  # producer must run whenever they do — including on felix-test-only PRs.
+  #
+  # The deps tool (hack/cmd/deps) emits `/<primary>/**` for the FIRST arg with no
+  # test-file exclusion, but for every *secondary*/dependency package it adds a
+  # `/<pkg>/**/*_test.go` EXCLUSION. With cmd/calico primary, felix is only a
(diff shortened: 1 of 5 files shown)


sridhartigera (approved):

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
```diff
@@ -201,18 +201,29 @@
...
+  # to cmd/calico and any subcommand's Go closure (goldmane, guardian, typha, …).
+  #
   # The process/testing paths trigger this so the win-fv-felix block (which now
   # installs the locally-built calico image) has it cached in GCS. The libcalico
   # ipam/stacktrace paths and the 20-felix.yml self-trigger keep this block a
   # superset of the win-fv-felix block's trigger (felix's full test closure plus
   # its own CI block), so win-fv-felix never runs without this image in GCS.
   run:
-    when: "${CHANGE_IN(cmd/calico,non-go:/felix/,non-go:/typha/,non-go:/.semaphore/vms/,non-go:/docker/calico/,non-go:/libcalico-go/lib/ipam/,non-go:/libcalico-go/lib/testutils/stacktrace/,non-go:/.semaphore/semaphore.yml.d/blocks/20-felix.yml,non-go:/process/testing/aso,non-go:/process/testing/util,non-go:/process/testing/winfv-felix)}"
+    when: "${CHANGE_IN(felix,cmd/calico,typha,non-go:/.semaphore/vms/,non-go:/docker/calico/,non-go:/libcalico-go/lib/ipam/,non-go:/libcalico-go/lib/testutils/stacktrace/,non-go:/.semaphore/semaphore.yml.d/blocks/20-felix.yml,non-go:/process/testing/aso,non-go:/process/testing/util,non-go:/process/testing/winfv-felix)}"
```
fasaxc, line 226: This'll work for felix, but I bet kube-controllers and maybe others have the same issue.  Perhaps we should add a deps tool tweak to allow multiple primaries?

```
${CHANGE_IN(cmd/calico,primary:felix,primary:typha, primary:kube-controllers,...
```

PR #12888, opened 2026-06-02
Add lib/kind module for in-process kind cluster lifecycle

Brian-McM:
## What this does

Introduces `lib/kind`, a pure-Go library (no `make`, no shell-out) that owns the kind cluster lifecycle for tests and tooling.

- **`kind.Up` / `Down` / `LoadExisting`** — create a cluster, write its kubeconfig, and tear it down idempotently. The default networking config is Calico-friendly (default CNI off, `/16` pod subnet) so Calico can install its own dataplane; callers can override it. `Up` deliberately does **not** wait for kube-system pods to go Ready, since that can't happen until a CNI is installed (the next step the caller owns).
- **`localregistry`** — runs a host Docker registry as a containerd mirror, so images load through the standard containerd pull pipeline with layer caching across runs instead of mutating `imagePullPolicy` to `Never`. Blobs persist on a named volume across container removal, and the registry can serve multiple concurrent clusters. Emits the containerd 2.x `config_path` + `hosts.toml` setup so newer `kindest/node` images (kind v0.27+) work.

Logging goes through the `lib/std/log` interface rather than logrus, so the library stays backend-agnostic and the consuming binary wires the concrete logger.

## Why

A pure-Go shape (rather than driving `make`/shell) keeps failure attribution clean and makes cluster bring-up behave identically in CI and on a developer machine. `lib/kind` is its own module so its kind/containerd dependency tree stays out of the main module.

## Testing

`go build ./...` and `go vet ./...` pass in the `lib/kind` module. No unit tests yet — the package spins up real kind clusters and Docker registries, so it's exercised by the consumers that build on it rather than in isolation.

**Release note:**
```release-note
None
```


lib/kind/kind.go
```diff
@@ -0,0 +1,222 @@
...
+
+	if err := prov.Delete(c.Name, c.KubeconfigPath); err != nil {
+		logger.Warn("Failed to delete pre-existing cluster, continuing with creation.", "error", err)
+	}
+
+	logger.Info("Creating cluster (this typically takes 60-120s).")
+	start := time.Now()
+	// Don't pass CreateWithWaitForReady — that waits on kube-system pods
+	// to go Ready, which can't happen until CNI is up, and CNI is the
+	// thing the next step (InstallCalicoOSS) installs. Create returns
```
rene-dekker: > InstallCalicoOSS
There is no code for InstallCalicoOSS ?

lib/kind/kind.go
```diff
@@ -0,0 +1,222 @@
...
+	// Build a clientset for in-cluster operations (webhook install, pod
+	// delete-and-recreate). Even with the puller disabled we want the
+	// clientset on Cluster for future tools.
+	restCfg, err := clientcmd.BuildConfigFromFlags("", c.KubeconfigPath)
+	if err != nil {
+		return cl, fmt.Errorf("load kubeconfig %s: %w", c.KubeconfigPath, err)
+	}
+	cs, err := kubernetes.NewForConfig(restCfg)
+	if err != nil {
+		return cl, fmt.Errorf("build clientset: %w", err)
```
caseydavenport: We should make these error returns consistent - both clientset failures return a non-nil `cl`, but the `prov.Create` failure above (L159) returns `nil`. A caller doing the usual `cl, err := Up(...); if err != nil { return err }` drops `cl` here and leaks the running cluster, since Create already succeeded by this point. I think it's cleaner to `Down` internally on these failures and return `nil`, so a non-nil error always means there's nothing left to clean up.

lib/kind/kind.go
caseydavenport: Yeah, I think the comment is referring to something that isn't in this PR?

caseydavenport:
This looks great.

If I'm reading it right, this covers the cluster + registry primitive (roughly `kind-cluster-create`, `registry.sh`, and teardown). To actually retire the shell/make tooling, are these the remaining pieces, or am I missing something?

- Locally-built image path. `Push` re-mirrors an upstream ref today, but the main reason people run `kind-up` is to load the images they just built - so we'd need a way to push a local docker-daemon image (and ideally skip it when the image ID hasn't changed).
- The install + wait-for-ready step. The `InstallCalicoOSS` you mention in the comments, plus CRD pre-install and the TigeraStatus/pod readiness waits (and the diagnostics the shell dumps on failure).
- A reload loop - rebuild changed images → `helm upgrade --reuse-values` → force the re-pull.
- The enterprise surface eventually: variant, license secret, pull secret, and the MCM mgmt/managed wiring.
- The specialized rigs on top (migration, VRF, gateway conformance).

We should update the pinned versions to match the ones in `metadata.mk`.

Brian-McM:
> This looks great.
> 
> If I'm reading it right, this covers the cluster + registry primitive (roughly `kind-cluster-create`, `registry.sh`, and teardown). To actually retire the shell/make tooling, are these the remaining pieces, or am I missing something?
> 
> * Locally-built image path. `Push` re-mirrors an upstream ref today, but the main reason people run `kind-up` is to load the images they just built - so we'd need a way to push a local docker-daemon image (and ideally skip it when the image ID hasn't changed).
> * The install + wait-for-ready step. The `InstallCalicoOSS` you mention in the comments, plus CRD pre-install and the TigeraStatus/pod readiness waits (and the diagnostics the shell dumps on failure).
> * A reload loop - rebuild changed images → `helm upgrade --reuse-values` → force the re-pull.
> * The enterprise surface eventually: variant, license secret, pull secret, and the MCM mgmt/managed wiring.
> * The specialized rigs on top (migration, VRF, gateway conformance).
> 
> We should update the pinned versions to match the ones in `metadata.mk`.

Yes there's still work to be done to actually replace the shell / make tooling. This is the very basics of what's needed just to bring the clusters up and down, referring specifically to kind. Some extra stuff might be in this module or might just be a part of the rigs on top (like installing calico in the clusters).

doucol (approved):

PR #12892, opened 2026-06-02
Add KubeVirt live migration e2e tests for KIND clusters

song-jiang:
## Summary
- Add KubeVirt live migration e2e tests that run on KIND clusters using MockVirt (KubeVirt simulation mode)
- Test cases cover iBGP/eBPG route convergence during migration, /32 route leak on migration timeout, NetworkPolicy enforcement across migration, IP persistence, and IPAM owner promotion
- Add Semaphore CI job that creates a KIND cluster, deploys MockVirt, and runs the `Feature:KubeVirt-KIND` tests


e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -77,97 +75,13 @@ var _ = describe.CalicoDescribe(
...
-			logrus.Infof("After promotion: Active pod=%s node=%s (was %s on %s)",
-				targetPodName, targetNode, sourcePod.Name, sourceNode)
-		})
-
-		// Test 2: TCP stream over iBGP must not lose any "seq=N" segments across two
+		// Test 1: TCP stream over iBGP must not lose any "seq=N" segments across two
 		// consecutive cross-node live migrations on a 3-worker cluster (server VM hops,
 		// client pod stays put).
 		It("should maintain TCP connection over iBGP across two consecutive live migrations", func() {
+			if isKINDCluster(f) {
```
skoryk-oleksandr: These `if isKINDCluster(f) { Fail(...) }` guards (here, line 184, line 394, and the inverted one in `live_migration_kind.go:232`) are self-skips dressed up as failures. The YAML configs already gate which lane picks up which suite via `Feature:KubeVirt` vs `Feature:KubeVirt-KIND`, so the runtime checks are redundant. If a test ever does land on the wrong cluster, the right fix is in the YAML labels, not a runtime guard that turns a CI misconfiguration into a polite test failure. Delete all four.

e2e/pkg/tests/kubevirt/live_migration.go
```diff
@@ -266,6 +180,9 @@ var _ = describe.CalicoDescribe(
 		// Requires EXT_IP, EXT_KEY, EXT_USER.
 		framework.Context("eBGP external client", describe.RequiresExternalNode(), func() {
```
skoryk-oleksandr: There's already `describe.WithExternalNode()` that attaches a label called `ExternalNode`, and every existing YAML excludes it. Adding `RequiresExternalNode` (different label name, same semantics) means lanes that don't exclude it still try to run the test, get nil from `externalnode.NewClient()`, and `Fail` with "External node not configured". Simplest fix: use the existing `describe.WithExternalNode()` here and drop the new helper.

e2e/pkg/tests/kubevirt/live_migration_kind.go
```diff
@@ -0,0 +1,633 @@
...
+// KubeVirt live migration route convergence tests for KIND clusters.
+// iBGP test: validates kernel route metric elevation (512) and reversion (1024)
+// on worker nodes during live migration — no external BIRD peer required.
+// eBGP test: validates BGP route priority (krt_metric, community tagging,
+// local_pref) using a local BIRD container on the Docker "kind" network.
+var _ = describe.CalicoDescribe(
+	describe.WithTeam(describe.Core),
+	describe.WithFeature("KubeVirt-KIND"),
+	describe.WithCategory(describe.Networking),
+	describe.WithSerial(),
```
skoryk-oleksandr: `WithSerial()` on the whole describe forces all four tests to run one at a time, but only the eBGP test owns cluster-scoped state (BIRD container, BGPPeer, BGPFilter). The iBGP, timeout, and netpol tests are namespace-scoped and parallel-safe. Move `WithSerial()` from the describe args to the eBGP `It` only — same shape as `live_migration.go`.

e2e/pkg/tests/kubevirt/live_migration_kind.go
```diff
@@ -0,0 +1,633 @@
...
+}
+
+// querySnapshot queries the BIRD routing table for both the /32 host route
+// and the /26 block route. Same pattern as queryTORSnapshot but via local
+// docker exec instead of SSH.
+func (p *kindBIRDPeer) querySnapshot(vmIP string) torRouteSnapshot {
+	ip := strings.Split(vmIP, "/")[0]
+
+	parsed := net.ParseIP(ip).To4()
+	blockIP := net.IPv4(parsed[0], parsed[1], parsed[2], parsed[3]&0xC0)
```
skoryk-oleksandr: `parsed := net.ParseIP(ip).To4()` returns nil for malformed input or IPv6, then `parsed[0]` panics with index-out-of-range. Today `vmIP` is always v4 from VMI status, but a dual-stack future or an unexpected upstream format takes the whole spec down with a stack trace instead of a clean failure, and may skip DeferCleanup so containers/BGPPeers leak. One-liner guard: `if parsed == nil { return torRouteSnapshot{}, fmt.Errorf("not an IPv4 address: %q", ip) }`.

e2e/pkg/tests/kubevirt/live_migration_mockvirt.go
```diff
@@ -0,0 +1,633 @@
...
+				if len(alternate) != 0 {
+					return fmt.Errorf("AlternateOwnerAttrs should be empty after failed migration cleanup, got %v", alternate)
+				}
+				return nil
+			}, 1*time.Minute, 2*time.Second).Should(Succeed())
+			logrus.Infof("IPAM ownership confirmed: Active pod=%s node=%s, Alternate=empty",
+				sourcePod.Name, sourceNode)
+		})
+
+		It("should enforce NetworkPolicy after live migration", func() {
```
skoryk-oleksandr, line 258: This is nearly identical to Test 4 in `live_migration.go`, just ICMP vs TCP. If we want this running on both lanes, simplest is one test in one file with both feature labels (same pattern `ip_persistence.go` uses). If there's a real reason to keep both, let's pull the shared body into a helper and parameterize on the target — otherwise a future change has to touch two places and they'll drift.

e2e/pkg/tests/kubevirt/live_migration_kind.go
```diff
@@ -0,0 +1,633 @@
...
+			}, 10*time.Second, 2*time.Second).Should(Succeed())
+		})
+
+		It("should converge eBGP routes after live migration", func() {
+			ctx, cancel := context.WithTimeout(context.Background(), eBGPDoubleMigrationTimeout)
+			defer cancel()
+			ns := f.Namespace.Name
+
+			// Start BIRD container on the kind Docker network.
+			bird := startKindBIRDPeer("kind-bird-ebgp")
```
skoryk-oleksandr: `kind-bird-ebgp` is a hardcoded cluster-scoped name, and `startKindBIRDPeer` does `docker rm -f` on it before creating a fresh one. Two overlapping runs on the same machine (parallel local + CI, two CI jobs sharing a runner, etc.) will silently kill each other's BIRD container mid-test. The BGPPeer/BGPFilter path in `setupEBGPPeeringCommon` already does this right: `utils.GenerateRandomName(prefix)` plus a sweep-stale-by-age helper. Same pattern would fit here.

e2e/pkg/tests/kubevirt/live_migration_kind.go
```diff
@@ -0,0 +1,633 @@
...
+				_, err := tester.Connect(allowed, vmTarget)
+				g.Expect(err).NotTo(HaveOccurred(),
+					"allowed client should still reach VM")
+				_, err = tester.Connect(denied, vmTarget)
+				g.Expect(err).To(HaveOccurred(),
+					"denied client should still be blocked by NetworkPolicy")
+			}, 10*time.Second, 2*time.Second).Should(Succeed())
+		})
+
+		It("should converge eBGP routes after live migration", func() {
```
skoryk-oleksandr: The iBGP test above, this eBGP test, and the eBGP test in `live_migration.go` all run the same eight-step sequence: create VM, migrate, assert elevated route metric on the new node, wait for revert, migrate again pinned to a third node, re-assert. Only the "ask the peer about the route" call varies (kernel route on a worker, BIRD on a kind container, BIRD on the TOR over SSH).

You already have the right abstraction in `ebgpBIRDPeer` — extending it with a `QueryRoute(vmIP) (RouteState, error)` method and pulling these steps into a helper like:

```go
func migrateAndVerifyConvergence(ctx, cli, f, vmName, sourceNode string, peer routePeer) (newNode string)
```

would collapse three ~80-line bodies into ~15 lines each. Right now if we ever change the elevated-metric value or the convergence-window budget, we have to update three places and not drift.

e2e/pkg/tests/kubevirt/live_migration_kind.go
```diff
@@ -0,0 +1,633 @@
...
+			vm.Create(ctx, cli)
+			DeferCleanup(func() { vm.Delete(cli) })
+
+			vmIP, node1 := vm.WaitForRunningWithIP(ctx, cli)
+			logrus.Infof("VM %s on %s with IP %s", vmName, node1, vmIP)
+
+			// Pre-migration: the source node should have a normal-priority /32 route.
+			By("Verifying pre-migration kernel route metric on source node")
+			Eventually(func() int {
+				return queryWorkerMetric(f, node1, vmIP)
```
skoryk-oleksandr: `queryWorkerMetric` (and `kindBIRDPeer.queryRoute` / `querySnapshot`) log a warning and return `-1` / zero-value on error. When this Eventually times out, the failure message says `expected 1024, got -1` — looks like a missing route, but the real cause (SSH broken, pod gone, parse failed) is hidden in a `Warnf` somewhere up in the log. Better to return `(value, error)` and use `Eventually(func(g Gomega) { ... })` so the actual error surfaces in the test report.

.semaphore/semaphore.yml.d/blocks/20-kubevirt-live-migration.yml
```diff
@@ -0,0 +1,45 @@
+- name: "KubeVirt live migration (KIND)"
+  run:
+    when: "${CHANGE_IN(felix,cni-plugin,confd,node,e2e,libcalico-go)}"
```
skoryk-oleksandr: Worth running `make gen-semaphore-yaml` and checking the expanded `when:` path list for this block in the generated `.semaphore/semaphore.yml`. I don't see `/e2e/pkg/tests/**` or `/e2e/pkg/tests/kubevirt/**` in the current expansion (only `/e2e/cmd/k8s/*.go`, `e2e/Makefile`, `e2e/deps.txt`, etc.). If that's right, a future PR editing one of these test files won't trigger this lane — the change merges with green CI but the lane that's supposed to verify it never ran. Likely fix is adding an entry to `e2e/deps.txt`.

skoryk-oleksandr (commented):
Overall LGTM. Added few comments.

e2e/pkg/tests/kubevirt/ip_persistence.go
```diff
@@ -49,6 +50,7 @@ import (
 var _ = describe.CalicoDescribe(
 	describe.WithTeam(describe.Core),
 	describe.WithFeature("KubeVirt"),
+	describe.WithFeature("KubeVirt-KIND"),
```
caseydavenport: I think the pattern we have been following elsewhere is defining "requirements" as separate labels -

e.g.,:

```
Feature:KubeVirt
RequiresGCP
```

Or whatever the requirement is that makes some of these tests unable to run on kind - that way we don't end up double-tagging tests as multiple features.

e2e/pkg/tests/kubevirt/live_migration.go
caseydavenport: I think this is wrong? 

If a test runs on the wrong cluster type, we want to fail fast with a very clear reason why. If anything I'd suggest making this error message much more explicit - 

```
Fail("This test requires XYZ which is not available on KIND clusters")
```

e2e/pkg/tests/kubevirt/live_migration_mockvirt.go
```diff
@@ -0,0 +1,633 @@
...
+		return nil
+	}, 5*time.Minute, 1*time.Second).Should(Succeed())
+}
+
+// KubeVirt live migration route convergence tests for KIND clusters.
+// iBGP test: validates kernel route metric elevation (512) and reversion (1024)
+// on worker nodes during live migration — no external BIRD peer required.
+// eBGP test: validates BGP route priority (krt_metric, community tagging,
+// local_pref) using a local BIRD container on the Docker "kind" network.
+var _ = describe.CalicoDescribe(
```
caseydavenport, line 45: I think we should move this into its own file separate from the helpers - also I think we want tests to go into `x_test.go`

e2e/pkg/tests/kubevirt/live_migration_mockvirt.go
```diff
@@ -0,0 +1,633 @@
...
+			}
+			// Double migration needs 3 workers: source, first target, second target.
+			utils.RequireNodeCount(f, 3)
+
+			var err error
+			cli, err = e2eclient.NewAPIClient(f.ClientConfig())
+			Expect(err).NotTo(HaveOccurred(), "failed to build controller-runtime client")
+		})
+
+		It("should converge iBGP routes after live migration", func() {
```
caseydavenport, line 68: Which part of this test is tightly coupled with kind?

e2e/pkg/tests/kubevirt/live_migration_kind.go
```diff
@@ -0,0 +1,633 @@
...
+// commands so no external node or SSH credentials are required.
+type kindBIRDPeer struct {
+	containerName string
+	containerIP   string
+}
+
+// startKindBIRDPeer starts a BIRD container on the kind Docker network and
+// returns a handle for interacting with it. The container runs in privileged
+// mode so BIRD can manipulate the routing table.
+func startKindBIRDPeer(name string) *kindBIRDPeer {
```
caseydavenport: I am a little bit skeptical of this - it sort of breaks the "infra" and "test code" boundary that we try to maintain in the e2e tests - the tests themselves should be able to run against a cluster without invasive modifications to that cluster.

We generally use the node k8s tests (historically in Python, but no reason we can't run a Go test in that suite too I think) when we want to do more things like this where we want to test "real" dataplane against a kind cluster in CI, but not run it anywhere else. 

I'm not necessarily against this... but it does feel against the purpose of these e2es (which is that they are designed primarily to test cross-platform features). WDYT?

e2e/pkg/tests/kubevirt/live_migration_mockvirt.go
song-jiang: Yeah, I think it is good enough to test network policy on KIND cluster. I'll remove the same test on the GCP cluster.

e2e/pkg/tests/kubevirt/live_migration_kind.go
song-jiang: Good point. This is closer in spirit to a node k8s test (running against PRs) than a traditional e2e. The main motivation for putting it in the e2e suite is to reuse the existing KubeVirt test cases and utilities rather than reimplementing them in a separate test group. The trade-off is that a few tests need KIND-specific infrastructure like the BIRD container.

e2e/pkg/tests/kubevirt/live_migration_mockvirt.go
song-jiang: Because of lack of a real guest OS, we can't live migrate an existing TCP connection. The iBGP test here (and the eBGP test as well) validates that Felix correctly elevates and reverts route metrics during migration, without relying on TCP connectivity.

e2e/pkg/tests/kubevirt/live_migration_mockvirt.go
song-jiang: Moved helper functions to utils.go. 

> I think we want tests to go into x_test.go

Are you sure? I see all the files under e2e directory do not follow that pattern.

e2e/pkg/tests/kubevirt/ip_persistence.go
song-jiang: Done. Added `RequiresRealKubeVirt` label

e2e/pkg/tests/kubevirt/live_migration_kind.go
song-jiang: I looked into this and the main hesitation is that the verification steps aren't as uniform as they appear: the iBGP test checks kernel route metrics on worker nodes, the eBGP KIND test queries a local BIRD container, and the eBGP GCP test queries a TOR over SSH. It is not easy to read if we put them into a shared function with different parameters. Collapsing them into a shared function would require enough parameters and callbacks that the result is harder to read than the duplication it removes. 

Agree we should refactor the code so that the magic numbers (metric values, convergence windows) shouldn't be repeated. I'll extract those as shared constants so a change is one-line.

skoryk-oleksandr (approved):
LGTM.

e2e/pkg/tests/kubevirt/live_migration_mockvirt.go
caseydavenport: Ah, yeah I seem to have been wrong on that!

e2e/pkg/tests/kubevirt/live_migration_kind.go
caseydavenport: I think we can still import the utilities and not include the tests in the e2e binary - it seems to me that we have a need for a way to interact with kubevirt from Golang and that library can be used in multiple places! Same with the bits that interact with kind / docker.

@Brian-McM is also working on a kind deployment library / tool that fits in this space, so we'll soon have a Golang library for deploying kind clusters on a per-test basis on-demand.

song-jiang: That's good to know. Agreed that long-term these should move to a separate test suite, especially once Brian's KIND library is available. I'd suggest getting this PR merged first and improving it with a follow-up once we know exactly the best approach forward.

song-jiang: @caseydavenport After second thoughts, I think we can improve this within the PR to address your concerns:  
 - Extracted shared BIRD/BGP utilities into a reusable package (e2e/pkg/utils/bgp/). This defines a BIRDPeer interface with two implementations — ContainerBIRDPeer (local Docker) and SSHBIRDPeer (SSH external node).                                                                            
  - Decoupled the BIRD container lifecycle from test code. Tests no longer create or destroy the container. Instead, infrastructure setup (the Semaphore CI pipeline) creates the container and passes its name via the `BIRD_BGPPEER_CONTAINER_NAME` env var. The test just discovers the pre-existing container. This restores the infra/test boundary you flagged.
- Replaced `isKINDCluster` with `isMockVirtDeployed` using the dynamic client to check the KubeVirt CR's simulationMode field. It checks whether MockVirt is actually deployed, so the same tests could run on any cluster with MockVirt.
- Renamed KIND references to MockVirt where they referred to the simulated KubeVirt environment (live_migration_kind.go → live_migration_mockvirt.go, setupKindEBGPPeering → setupMockVirtEBGPPeering, etc.). The CI block name stays "KubeVirt live migration (KIND)" since that describes where it runs in CI.

e2e/pkg/tests/kubevirt/live_migration_mockvirt.go
caseydavenport: Great, thanks! I think we should include that level of detail in the test file to explain this to future readers.

caseydavenport (approved):
Thanks @song-jiang - I think this is good to go now! 

I think we might still want to consider moving the MockVirt-only tests to a different suite longer term, but the framing / split as it exists now feels good.

PR #12898, opened 2026-06-03
networking-calico: add resync-concurrency test for CORE-12037

nelljerram:
Add a DevStack-side test that demonstrates a long-running resync does not meaningfully block dynamic port creation, plus the test-only driver knobs it needs to drive a deterministic test window.

Driver-side knobs (production code, default 0, never set in production):

* ``--inject-per-item-delay-ms`` hidden flag on ``calico-resync``.
* ``[calico] startup_resync_inject_per_item_delay_ms`` config option for the start-of-day resync.

Both flow through to ``Scope.__init__``, which passes the value to ``WorkloadEndpointSyncer``'s constructor.  ``ResourceSyncer`` then sleeps that many milliseconds after every iteration of its compare loop -- so the endpoints phase (which dominates real resync time anyway) is stretched out to a known duration of ``populate_scale * per_item_delay_ms``, giving the test a stable in-resync window.

Test script (``devstack/resync_concurrency_test.py``):

* Reuses ``resync_scale_test.py``'s populate / cleanup helpers via ``import resync_scale_test as scale``.
* Two scenarios:
  - ``on_demand`` runs ``calico-resync`` with the inject flag, polls its log file for the ``"Starting resync for WorkloadEndpoint"`` marker, then creates ``test_ports`` ports back-to-back through the Neutron REST API and times each API call.  Compares the in-resync median against a baseline median measured before the resync started.
  - ``startup`` writes the inject knob into ``[calico]`` via ``crudini``, restarts neutron-server with ``systemctl restart devstack@q-svc``, polls ``journalctl`` for the same endpoints-phase marker, then fires the same in-resync burst against the API workers while the resync worker is busy in its stretched endpoints phase.  Baseline runs after the resync completes.

* Each scenario asserts ``in_resync_median <= baseline_median * 2.0`` (configurable via ``RESYNC_CONCURRENCY_RATIO_LIMIT``).  The summary prints each in-resync port's timestamp between the resync's own ``started_at`` and ``finished_at``, so the overlap is visible directly to a reviewer.

* Tunable via ``RESYNC_CONCURRENCY_POPULATE_SCALE`` (default 1000), ``RESYNC_CONCURRENCY_TEST_PORTS`` (5), ``RESYNC_CONCURRENCY_PER_ITEM_DELAY_MS`` (50), ``RESYNC_CONCURRENCY_RATIO_LIMIT`` (2.0), and ``RESYNC_CONCURRENCY_SCENARIOS`` (default ``on_demand,startup``).

The startup scenario relies on the ``stack`` user having passwordless sudo for ``systemctl restart devstack@q-svc``, ``journalctl``, and ``crudini``, which DevStack's ``create-stack-user.sh`` already sets up.

### DIFF
--- a/networking-calico/devstack/bootstrap.sh
+++ b/networking-calico/devstack/bootstrap.sh
@@ -235,14 +235,16 @@ python3 ../calico/networking-calico/devstack/qos_responsiveness_tests.py -v
 EOF
 
 # Run resync concurrency test.  Prints one RESYNC_CONCURRENCY_RESULT
-# line per scenario.
+# line per scenario.  Non-zero exit (any scenario over the 2x ratio
+# gate) propagates out of the heredoc and fails the bootstrap, so
+# regressions block CI.
 sudo -u stack -H -E bash -x <<'EOF'
 cd /opt/stack/devstack
 . openrc admin admin
 
 export ETCD_HOST=${SERVICE_HOST}
 export RESYNC_CALICO_RESYNC=${DEVSTACK_VENV:-/usr/local}/bin/calico-resync
-python3 ../calico/networking-calico/devstack/resync_concurrency_test.py || true
+python3 ../calico/networking-calico/devstack/resync_concurrency_test.py
 EOF
 
 # Run resync scale benchmark.  Prints one RESYNC_SCALE_RESULT line per
--- a/networking-calico/devstack/bootstrap.sh
+++ b/networking-calico/devstack/bootstrap.sh
@@ -113,6 +113,12 @@ cd devstack-bootstrap
 sudo apt-get update
 sudo apt-get -y install git
 
+# crudini is used by resync_concurrency_test.py's `startup` scenario to
+# set/unset [calico] startup_resync_inject_per_item_delay_ms in neutron.conf
+# across a neutron-server restart.  Install it now so the test can use it
+# without needing a separate sudo block later.
+sudo apt-get -y install crudini
+
 # Enable IPv4 and IPv6 forwarding.
 sudo sysctl -w net.ipv4.ip_forward=1
 sudo sysctl -w net.ipv6.conf.all.forwarding=1
@@ -226,6 +232,17 @@ sudo pip install openstacksdk etcd3 pymysql
 
 export ETCD_HOST=${SERVICE_HOST}
 python3 ../calico/networking-calico/devstack/qos_responsiveness_tests.py -v
+EOF
+
+# Run resync concurrency test.  Prints one RESYNC_CONCURRENCY_RESULT
+# line per scenario.
+sudo -u stack -H -E bash -x <<'EOF'
+cd /opt/stack/devstack
+. openrc admin admin
+
+export ETCD_HOST=${SERVICE_HOST}
+export RESYNC_CALICO_RESYNC=${DEVSTACK_VENV:-/usr/local}/bin/calico-resync
+python3 ../calico/networking-calico/devstack/resync_concurrency_test.py || true
 EOF
 
 # Run resync scale benchmark.  Prints one RESYNC_SCALE_RESULT line per
--- a/networking-calico/devstack/resync_concurrency_test.py
+++ b/networking-calico/devstack/resync_concurrency_test.py
@@ -257,6 +257,7 @@ def scenario_on_demand(
     try:
         if not wait_for_endpoints_phase(log_file, deadline_secs=60):
             proc.kill()
+            proc.wait()
             raise RuntimeError(
                 "calico-resync did not reach the endpoints phase within 60s"
             )
@@ -289,6 +290,7 @@ def scenario_on_demand(
     finally:
         if proc.poll() is None:
             proc.kill()
+            proc.wait()
         try:
             os.unlink(output_file.name)
         except OSError:
--- a/networking-calico/devstack/resync_concurrency_test.py
+++ b/networking-calico/devstack/resync_concurrency_test.py
@@ -1,0 +1,644 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 Tigera, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied.  See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""Resync concurrency test: does a long-running resync block dynamic port creation?
+
+Resync runs in its own OS process - equally whether it is the
+``CalicoStartupResyncWorker`` for start-of-day resync within the Neutron server, or a
+``calico-resync`` invocation for on-demand - separate from the API workers.  So in
+principle a long resync should not block dynamic operations: the only ways it could
+would be through shared external resources -- Neutron DB lock contention, etcd's
+serialised transaction queue, or, on a single-host test box, CPU contention.
+
+This test demonstrates that, in practice, dynamic port creation latency does not regress
+meaningfully while a long resync is alive.  Two scenarios:
+
+* **on_demand**: run ``calico-resync`` with the ``--inject-per-item-delay-ms``
+  hidden flag so it stretches the endpoints phase to a known duration; create
+  test ports during that window; compare median latency to a baseline with no
+  concurrent resync.
+
+* **startup**: set ``[calico] startup_resync_inject_per_item_delay_ms`` in
+  ``neutron.conf``, restart neutron-server, and create test ports during the
+  ``CalicoStartupResyncWorker``'s stretched endpoints phase.
+
+Both scenarios assert that the in-resync median latency is at most ``2x`` the baseline
+median; both emit a timing summary that puts each port-create timestamp alongside the
+resync's own ``started_at`` / ``finished_at``, so a reviewer can see the overlap
+directly.
+
+Run as the ``stack`` user with the admin openrc sourced.  Environment vars:
+
+    ETCD_HOST=<ip>         (default localhost)
+    ETCD_PORT=<port>       (default 2379)
+    NEUTRON_CONF=<path>    (default /etc/neutron/neutron.conf)
+    RESYNC_CONCURRENCY_POPULATE_SCALE=<n>     (default 1000)
+    RESYNC_CONCURRENCY_TEST_PORTS=<n>         (default 5)
+    RESYNC_CONCURRENCY_PER_ITEM_DELAY_MS=<n>  (default 50)
+    RESYNC_CONCURRENCY_RATIO_LIMIT=<f>        (default 2.0)
+    RESYNC_CALICO_RESYNC=calico-resync        (path to the CLI)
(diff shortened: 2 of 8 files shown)


networking-calico/devstack/resync_concurrency_test.py
```diff
@@ -0,0 +1,644 @@
...
+
+
+def wait_for_log_text_in_journal(text, since_dt, deadline_secs):
+    """Poll ``journalctl -u devstack@q-svc`` until ``text`` appears.
+
+    ``since_dt`` is a naive ``datetime.datetime`` in UTC; we format it as
+    ``YYYY-MM-DD HH:MM:SS`` for ``journalctl --since``, which doesn't
+    accept ISO 8601's ``T``/``Z`` punctuation.
+    """
+    since_journal = since_dt.strftime("%Y-%m-%d %H:%M:%S")
```
lwr20, line 356: **`journalctl --since` timezone mismatch (likely to break the `startup` scenario off a UTC host).**

`since_dt` is a `datetime.utcnow()` value (lines 385/425) formatted here as a bare `YYYY-MM-DD HH:MM:SS` string. `journalctl --since` with no timezone interprets the timestamp in the system's **local** time, not UTC.

- On a host at e.g. UTC-5, the `--since` point lands ~5h in the future relative to the journal's local clock, so neither `Starting resync for WorkloadEndpoint` (line 393) nor `One-shot resync done` (line 421) is ever returned within the deadline → `RuntimeError`, and the `startup` scenario always fails.
- On a UTC+offset host the window reaches too far back and can match a *previous* run's marker, firing the burst prematurely.

It happens to work on a UTC CI box, but it's fragile. Suggest using `datetime.now()` (local, matching journald) or passing an explicit `UTC` to `journalctl` (e.g. set the timestamp with a `UTC` suffix / `--utc`).

networking-calico/devstack/resync_concurrency_test.py
```diff
@@ -0,0 +1,644 @@
...
+        output_file.name,
+        "--inject-per-item-delay-ms",
+        str(per_item_delay_ms),
+    ]
+    LOG.info("Starting calico-resync (per-item delay=%dms)", per_item_delay_ms)
+    LOG.info("  %s", " ".join(cmd))
+    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+    try:
+        if not wait_for_endpoints_phase(log_file, deadline_secs=60):
```
lwr20, line 258: **The in-resync burst may not actually overlap the stretched window it claims to measure.**

The burst is fired the instant `wait_for_endpoints_phase` sees `Starting resync for WorkloadEndpoint`. But that marker is logged at the *top* of `ResourceSyncer.resync()` (`syncer.py:132`) — **before** `get_from_etcd` (`syncer.py:146`) and the `CONTEXT_WRITER`-held `get_from_neutron` (`syncer.py:149-151`). The injected `time.sleep` is only in the compare loop afterwards (`syncer.py:250-251`).

So after the marker appears, the 6-port burst can run either:
- during the neutron read while a DB write lock is held — real contention the test then mis-attributes to the (claimed) compare phase, or
- entirely before the compare loop's sleeps even begin — measuring an un-stretched, lock-free window.

Either way `in_resync_median` doesn't necessarily reflect the stretched endpoints compare phase. Emitting a dedicated marker at the top of the compare loop (after both reads) would tie the burst to the actual delay window.

networking-calico/devstack/resync_concurrency_test.py
```diff
@@ -0,0 +1,644 @@
...
+        "--config-file",
+        extra_conf,
+        "--output",
+        output_file.name,
+        "--inject-per-item-delay-ms",
+        str(per_item_delay_ms),
+    ]
+    LOG.info("Starting calico-resync (per-item delay=%dms)", per_item_delay_ms)
+    LOG.info("  %s", " ".join(cmd))
+    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
lwr20: **`calico-resync` pipes are never drained → potential `proc.wait()` deadlock.**

`subprocess.Popen(..., stdout=PIPE, stderr=PIPE)` here, then `proc.wait()` at line 267 with no `communicate()`/read of the pipes. `use_stderr=False` routes oslo logging to the log file, but Python warnings, deprecation notices, or an unhandled traceback still go to stderr. If combined stdout+stderr exceeds the ~64KB OS pipe buffer, the child blocks on write and `proc.wait()` hangs indefinitely (there is no timeout).

Suggest `stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL` (logs already go to the file we tail), or read the pipes via `communicate()`.

networking-calico/devstack/resync_concurrency_test.py
```diff
@@ -0,0 +1,644 @@
...
+            proc.kill()
+            raise RuntimeError(
+                "calico-resync did not reach the endpoints phase within 60s"
+            )
+        LOG.info("Resync confirmed in endpoints phase -- starting in-resync burst")
+
+        # In-resync: measure while the resync is in the stretched endpoints
+        # phase.  We always discard the first sample as a warm-up (see
+        # _take_warmup_then_measure for rationale).
+        warmup, in_resync_main = _take_warmup_then_measure(
```
lwr20, line 269: **Asymmetric warm-up discard biases the `in_resync <= 2x baseline` gate toward passing.**

`_take_warmup_then_measure` discards the first in-resync sample (used at lines 256/405), but the baseline burst (`measure_port_create_burst`, lines 218/438) keeps all of its samples. Dropping the (often slowest) first in-resync sample lowers `in_resync_median` while `baseline_median` retains its first sample, deflating `ratio = in_resync/baseline`.

The docstring says this keeps the medians "directly comparable", but the asymmetric trimming actually makes them *not* computed the same way, and makes the `<= 2x` assertion more lenient — it could mask a genuine port-create regression. Suggest warm-up-trimming both bursts (or neither).

networking-calico/devstack/resync_concurrency_test.py
nelljerram: Addressed in https://github.com/projectcalico/calico/pull/12898/commits/b4f57b00ecfb4a4832db6dd62db5d6c4191ce5cd

networking-calico/devstack/resync_concurrency_test.py
nelljerram: Addressed in https://github.com/projectcalico/calico/pull/12898/commits/b4f57b00ecfb4a4832db6dd62db5d6c4191ce5cd

networking-calico/devstack/resync_concurrency_test.py
nelljerram: Addressed in https://github.com/projectcalico/calico/pull/12898/commits/b4f57b00ecfb4a4832db6dd62db5d6c4191ce5cd

networking-calico/devstack/resync_concurrency_test.py
nelljerram: Addressed in https://github.com/projectcalico/calico/pull/12898/commits/b4f57b00ecfb4a4832db6dd62db5d6c4191ce5cd

networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -158,6 +158,18 @@
         ),
         help="Deprecated and unused.  Retained to avoid neutron.conf errors.",
     ),
+    cfg.IntOpt(
+        "startup_resync_inject_per_item_delay_ms",
+        default=0,
+        help=(
```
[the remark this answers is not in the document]
nelljerram: Addressed in https://github.com/projectcalico/calico/pull/12898/commits/4743f3fab48cb1d10846e6c69c65d0110d6fd2c2

networking-calico/devstack/resync_concurrency_test.py
```diff
@@ -0,0 +1,664 @@
...
+            "Resync finished rc=%d, started_at=%s, finished_at=%s, total_ms=%d",
+            rc,
+            result["started_at"],
+            result["finished_at"],
+            result["total_ms"],
+        )
+    finally:
+        if proc.poll() is None:
+            proc.kill()
+        try:
```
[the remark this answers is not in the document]
nelljerram: Addressed in https://github.com/projectcalico/calico/pull/12898/commits/4743f3fab48cb1d10846e6c69c65d0110d6fd2c2

networking-calico/devstack/bootstrap.sh
```diff
@@ -228,6 +234,17 @@ export ETCD_HOST=${SERVICE_HOST}
...
 
+# Run resync concurrency test.  Prints one RESYNC_CONCURRENCY_RESULT
+# line per scenario.
+sudo -u stack -H -E bash -x <<'EOF'
+cd /opt/stack/devstack
+. openrc admin admin
+
+export ETCD_HOST=${SERVICE_HOST}
+export RESYNC_CALICO_RESYNC=${DEVSTACK_VENV:-/usr/local}/bin/calico-resync
+python3 ../calico/networking-calico/devstack/resync_concurrency_test.py || true
```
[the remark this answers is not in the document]
nelljerram: Addressed in https://github.com/projectcalico/calico/pull/12898/commits/4743f3fab48cb1d10846e6c69c65d0110d6fd2c2

networking-calico/networking_calico/plugins/ml2/drivers/calico/syncer.py
```diff
@@ -88,9 +88,13 @@ class ResourceSyncer(object):
...
     """
 
-    def __init__(self, db, resource_kind):
+    def __init__(self, db, resource_kind, inject_per_item_delay_ms=0):
         self.db = db
         self.resource_kind = resource_kind
+        # Test-only: when non-zero, sleep this many ms after each
+        # iteration of the compare loop below.  Used by the resync-
+        # concurrency test (CORE-12037).  Always 0 in production.
+        self._inject_per_item_delay_secs = inject_per_item_delay_ms / 1000.0
```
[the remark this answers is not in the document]
nelljerram: Addressed in https://github.com/projectcalico/calico/pull/12898/commits/4743f3fab48cb1d10846e6c69c65d0110d6fd2c2

networking-calico/devstack/resync_concurrency_test.py
```diff
@@ -0,0 +1,664 @@
...
+            raise RuntimeError("Start-of-day resync did not complete within the window")
+        t_resync_done_iso = datetime.datetime.utcnow().isoformat() + "Z"
+        LOG.info("Resync completed")
+    finally:
+        # Restore neutron.conf so future test runs aren't perturbed.
+        neutron_conf_unset(
+            neutron_conf,
+            "calico",
+            "startup_resync_inject_per_item_delay_ms",
+        )
```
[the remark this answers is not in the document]
nelljerram: Ignoring this one because there is no reason to believe that any other test will set this value.

lwr20 (approved):
Reviewed against head 1ff264. Production footprint is small and inert by default (guarded sleep, default-0 clamped config, one benign log line). The marker-race concern from the earlier revision is resolved by the dedicated compare-loop marker, and negative-delay inputs are rejected at three layers. CI green. LGTM.

Non-blocking nit: bootstrap.sh now hard-fails on the >2x ratio gate, which is sensitive to single-host CPU contention — worth keeping an eye on for flakes.

nelljerram:
Thanks @lwr20 .  I will keep on eye on flakes caused by the 2x hard boundary, and increase or remove if necessary.  But I haven't seen any case of those yet.

PR #12902, opened 2026-06-03
wait for Azure VM extensions in Windows FV bring-up

coutinhop:
vmss.sh waited for the Azure VirtualMachine to be Ready but not for the CustomScript extensions that actually install containerd (Linux) and configure OpenSSH / runtime (Windows). The harness then raced the extension and would time out 5+ minutes later deep inside setup-node.sh ("ERROR: containerd was not ready after 5 minutes") with no useful diagnostic — diagnose_aso_resources() only printed extension finalizers, not their Ready condition or provisioningState.

Add an explicit wait_for_aso_resource gate on
vm-linux-N-containerd / vm-windows-N-{openssh,customextension} after the VM-Ready gate, and extend the diagnostic dump to print each extension's status.conditions and provisioningState via jsonpath.

Harden the Linux containerd extension's inline bash against first-boot apt-lock contention with cloud-init / unattended-upgrades:
  - cloud-init status --wait before touching apt
  - stop and mask apt-daily{,-upgrade} timers + unattended-upgrades
  - raise DPkg::Lock::Timeout from 60 to 180
  - inline apt_retry() (3 attempts, 20s backoff) around all apt phases

No Windows-side bash changes (PowerShell, no apt). No setup-node.sh changes (its containerd polling is correct; the upstream gate makes its 5-min fallback moot in the happy path).

**Release note:**
```release-note
TBD
```

### DIFF
--- a/.semaphore/semaphore-scheduled-builds.yml
+++ b/.semaphore/semaphore-scheduled-builds.yml
@@ -342,8 +342,13 @@ blocks:
     # Publishing to Docker Hub happens in a separate post-build promotion
     # (push-images/nft-rpms.yml) so PR workflows never receive the push credential.
     #
-    # Consumers (untolerated gcloud cp): node, istio. Path forms cover the
-    # secondary-CHANGE_IN(istio) gap that misses istio/patches/*.
+    # Consumers (untolerated gcloud cp): "Build: node image" (10-prerequisites.yml),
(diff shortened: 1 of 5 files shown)


caseydavenport (approved):

marvin-tigera:

PR #12638, opened 2026-04-27
Prevent IP re-use from occurring too quickly

djmitche:
Fixes #12496; see that bug for motivation.

This
 * Distinguishes IP release (no longer tied to the workload) from deallocation (available for reallocation). Release occurs by marking the ordinal as released in the block, using a `ReleasedAt` timestamp. Deallocation occurs by garbage-collecting released ordinals to mark them as unallocated.
 * Adds `MinIPReclaimAgeSeconds` to the IPAMConfig struct
 * Passes IPAMConfig to all block allocations, so that they can perform block garbage collection
 * Performs block garbage collection in the kube controller, as a backstop in case it is missed on release or allocation
 * Verifies well-formedness of IPAM blocks, including a reasonable ReleasedAt, in `calicoctl ipam check`

**Release note:**
```release-note
The new MinIPReclaimAgeSeconds IPAM configuration parameter prevents IP addresses from being reused too quickly.
```

PR #12904, opened 2026-06-03
Make IPAM config read on read-only paths and create default at startup

caseydavenport:
GetIPAMConfig created the default IPAMConfiguration on a cache miss, so a plain read had the side effect of writing to the datastore. That races between nodes (hence the retry loop it carried), and it means read-only paths like IP release can end up materializing global config.

This makes GetIPAMConfig read-only: when no config exists it returns the default in memory. The default object is instead created explicitly at node startup, alongside the existing default FelixConfiguration and BGPConfiguration, so it still shows up for users. Unlike those two, the IPAM defaults are not the zero value, so AutoAllocateBlocks and the KubeVirt address persistence default are set explicitly.

Came up in #12638, which adds GetIPAMConfig calls on the release path and surfaced the write-on-read behavior. That can rebase on top of this.

Related: #12638

```release-note
None
```


libcalico-go/lib/ipam/ipam.go
```diff
@@ -2095,58 +2095,37 @@ func (c ipamClient) decrementHandle(ctx context.Context, handleID string, blockC
...
-				}
-				log.WithError(err).Errorf("Error creating IPAM config")
-				return nil, err
-			}
+		// No config has been set, so fall back to the default.
+		enabled := string(VMAddressPersistenceEnabled)
+		obj = &model.KVPair{
+			Key: model.IPAMConfigKey{},
+			Value: &model.IPAMConfig{
+				StrictAffinity:               false,
```
coutinhop, line 2116: interesting seeing `if detectOS(ctx) == "windows"` a few lines below... since windows requires strict affinity, could we set it conditionally here?

coutinhop (approved):

PR #12906, opened 2026-06-03
[release-v3.32] Allow server-side apply to create tiered policies

### LINKED ISSUE #12841: kubernetes server-side apply errors for calico network policies on calico 3.32.0
We deploy calico via the tigera-operator helm chart. After upgrading to 3.32.0 in our test environment, we noticed we are no longer able to apply `NetworkPolicy` or `GlobalNetworkPolicy` resources using server-side apply.

For example:

```
kind: GlobalNetworkPolicy
metadata:
  name: gnptest
spec:
  selector: fake == 'doesnotmatter'
  ingress:
  - action: Allow
  types:
  - Ingress
```

```
 $ kubectl apply -f gnp.yaml --server-side
Error from server (NotFound): globalnetworkpolicies.projectcalico.org "gnptest" not found
```

We initially noticed because we use Helm 4 to apply charts containing network policies, and by default Helm 4 uses server-side apply. Similar errors occur with our charts that contain calico network policies.

Client side apply of the same policy seems to be fine.

```
 $ kubectl apply -f gnp.yaml
globalnetworkpolicy.projectcalico.org/gnptest created
```

I'm not sure if I've missed something in the docs that would be needed to avoid this? We did not do the [native v3 CRD migration](https://docs.tigera.io/calico/latest/operations/crd-migration). I'm not sure if that would have fixed this? (Since the docs say it's currently a tech preview, we didn't interpret it as being required).

## Your Environment
* Calico version: 3.32.0
* Calico dataplane: iptables
* Orchestrator version: EKS 1.35

caseydavenport:
Cherry-pick of #12905 to release-v3.32.

Server-side apply of a NetworkPolicy, GlobalNetworkPolicy, or their staged variants failed with a NotFound error when the policy didn't already exist. This affects `kubectl apply --server-side` and Helm 4, which defaults to server-side apply. Client-side apply was unaffected.

The tier RBAC check on update fetched the existing policy before handing off to the store, so a create-on-apply (which arrives as an update with create allowed) returned NotFound instead of creating the policy. When there's no existing object, authorize the incoming object's tier and let the store create it.

Fixes #12841

```release-note
Fixes a NotFound error when using server-side apply (including Helm 4) to create Calico network policies that don't already exist.
```

### DIFF
--- a/apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
+++ b/apiserver/pkg/registry/projectcalico/globalpolicy/storage.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/google/uuid"
 	calico "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metainternalversion "k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


mazdakn (approved):
LGTM

PR #12909, opened 2026-06-03
Standardize Windows calico binary on the component subcommand tree

caseydavenport:
The operator recently moved the Windows node containers to drive the combined calico binary through `component ...` subcommands, matching Linux. The Windows binary and the CNI image hadn't caught up, so install-cni crashed ("The system cannot find the path specified") and the Windows node never came up.

This brings the Windows side in line with Linux:

- The Windows binary gets a `component` parent command, mirroring the Linux calico binary.
- The combined binary is named `calico.exe` everywhere on Windows (node image and cni image), matching Linux's `calico`. The node image previously shipped it as `calico-node.exe`.
- The cni-windows image ships `calico.exe`, so `calico.exe component cni install` runs the install and is also what gets staged onto the host as the CNI plugin.
- The felix, confd, and node service scripts call the binary through `component ...`.
- Drops the now-redundant standalone install binary.

Needs the matching operator change: https://github.com/tigera/operator/pull/4884

> [!NOTE]
> Docs follow-up: the manual Windows install now ships the binary as `calico.exe` instead of `calico-node.exe`. The install steps in tigera/docs need the same rename.

```release-note
NONE
```


node/windows-packaging/CalicoWindows/confd/confd-service.ps1
```diff
@@ -38,7 +38,7 @@ if($env:CALICO_NETWORKING_BACKEND -EQ "windows-bgp")
   rm blocks.ps1 -ErrorAction SilentlyContinue
 
   # Run the calico-confd binary.
-  & ..\calico-node.exe node confd --confdir="$PSScriptRoot"
+  & ..\calico-node.exe component confd --confdir="$PSScriptRoot"
```
coutinhop: are `calico-node.exe` and `calico.exe` different in any way? should we standardize `calico.exe` everywhere?

PR #12786, opened 2026-05-15
Stop auto-pushing per-component images on every master/release- merge

rene-dekker:
## Description

Removes the `auto_promote` on the 19 `push-images/*` promotions so they no longer fire on every merge. They remain as manual buttons so a single component can be re-pushed without rerunning the full hashrelease.

To push every per-component image at the current SHA, trigger the existing **Publish hashrelease** promotion with `BUILD_CONTAINER_IMAGES=true` and `PUBLISH_IMAGES=true`. This is what daily/scheduled hashrelease tasks already do.

Auto-promotion for **Cleanup** (`result = 'stopped'`) and **Run Fossa scans** is preserved.

## For PR author

- [x] Tests for change.
- [x] If changing pkg/apis/, run \`make generate\`
- [x] If updating Helm charts, set the version with prerelease identifier to \`0.0.0-need-set-on-cut-release\` so that release tooling will correctly set it on release.
- [x] Update user-facing documentation. Links to [docs](https://docs.tigera.io/calico/latest/about) can be added in the description.
- [x] If modifying API, ensure that the API change is also made in the [API repo](https://github.com/projectcalico/api).

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - \`kind/bug\` if this is a bugfix.
  - \`kind/enhancement\` if this is a a new feature.
  - \`release-note-required\` if this PR has user-facing changes.
  - Required labels for primarily docs-related changes.

\`\`\`release-note
The per-component "Push * images" Semaphore promotions no longer auto-fire on every master/release- merge. Use the "Publish hashrelease" promotion with PUBLISH_IMAGES=true to push images at the current SHA.
\`\`\`

PR #12910, opened 2026-06-04
Re-enable publication of OpenStack packages for master branch

nelljerram:
This was disabled as part of #12786, but the argument for needing packages for OpenStack is different from that for hashrelease images.  Our Calico/OpenStack ST (in tigera/calico-test) depends on those packages being up to date, and we don't have an analogue of the hashrelease build building its own images.


danudey (approved):

nelljerram:
Windows issue is known and being worked on, current advice is to ignore it.

PR #12913, opened 2026-06-04
Auto-approve operator CRD sync PRs

caseydavenport:
The operator CRD sync workflow opens a PR every hour, but it sat waiting for a manual approval before Marvin would merge it. This adds an auto-approve step (as marvin-tigera, since the bot can't approve its own PR and the Actions token can't approve at all) so the PR merges unattended once CI is green. Branch protection still gates the merge on the required Semaphore check, so nothing merges on red CI.

```release-note
None
```


skoryk-oleksandr (approved):
**LGTM**. 
Small non-blocking edge case:
Approval only runs when pull-request-operation is created or updated. If gh pr review approve fails once due to a transient error, and later runs have operation=none, the PR may stay unapproved because the approval step is skipped.

PR #12914, opened 2026-06-04
Add workflow to DM PR authors on Slack when CI finishes

tomastigera:
## Description

Adds `.github/workflows/ci-notify.yml`, which DMs a PR author on Slack when their PR's Semaphore CI finishes.

It piggybacks on the same GitHub `status` event that `sem-ai-ci-triage.yml` uses (Semaphore reports CI as commit statuses on the aggregate context `ci/semaphoreci/pr: Calico`), reusing its fork-aware PR/author resolution. Notifications fire on all terminal outcomes (success/failure/error).

**Opt-in** is the `CI_NOTIFY_MAP` repo variable — a comma-separated list of `<github-login>:<slack-member-id>` pairs (Slack member IDs aren't secret). Delivery uses the gh-pr-monitor Slack bot via `chat.postMessage`; its token is the `GH_PR_MONITOR_BOT_TOKEN` repo secret (needs `chat:write` + `im:write`).

On failure/error it briefly polls (~120s) for the `sem-ai-ci-triage` failure-analysis comment and deep-links to it when present, gated on the comment's `updated_at` being newer than the status event so it never links a previous run's stale analysis. A manual `workflow_dispatch` path (pr_number + simulated state) allows post-merge end-to-end testing. The workflow is a no-op (never fails noisily) when the variable/secret are unset, the author hasn't opted in, or Slack is unreachable. PR-derived free-form text (title) is kept off the `${{ }}`/`GITHUB_OUTPUT` interpolation path to avoid script injection.

### Also: clean up stale triage comments on green

`sem-ai-ci-triage.yml` only ever upserts its failure-analysis comment on failure/error and never removed it, so a PR that later passed kept showing an obsolete failure report. This PR adds a `cleanup` job to that workflow which, on a `success` status, deletes any lingering `sem-ai-ci-triage` comment. It is mutually exclusive with the existing triage job (success vs failure/error) and best-effort.

### Setup (one-time, by a maintainer)

```
gh variable set CI_NOTIFY_MAP --repo projectcalico/calico --body 'login:Uxxxx, ...'
gh secret   set GH_PR_MONITOR_BOT_TOKEN --repo projectcalico/calico --body 'xoxb-...'
```

## Release Note

```release-note
TBD
```

### DIFF
--- a/.github/workflows/ci-notify.yml
+++ b/.github/workflows/ci-notify.yml
@@ -149,12 +149,32 @@ jobs:
           env.STATE != 'success'
         env:
           PR: ${{ steps.resolve.outputs.pr }}
+          # When CI finished, per GitHub (status events only; empty on a
+          # manual workflow_dispatch). Used as the freshness reference below.
+          EVENT_CREATED_AT: ${{ github.event.created_at }}
         run: |
           MARKER="<!-- sem-ai-ci-triage -->"
           COMMENT_URL=""
-          # Bound the query with `since` (filters by updated_at): triage
-          # always *creates or PATCHes* its comment for the current run, so
-          # its updated_at is recent. A 2h window keeps the result set to
+
+          # Freshness gate: triage upserts a single comment per PR, so a
+          # comment from an *earlier* failure can already exist. Accept it
+          # only if it was updated after this CI run finished — i.e. triage
+          # (which is triggered by the same status event) has rewritten it
+          # for the current run. Compare GitHub-server timestamps on both
+          # sides (event created_at vs comment updated_at); ISO-8601 UTC
+          # sorts lexicographically, so a string '>' is a time comparison.
+          # If triage lags we keep polling; if it skips the run entirely the
+          # comment never goes fresh and we correctly omit the link. On a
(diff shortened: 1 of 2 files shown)


skoryk-oleksandr (approved):
**LGTM**. 
One minor comment : The only edge case is when triage has not rewritten the comment for the current run yet, or skips the run due to a backend error. In that case, the DM could point to a comment that still shows analysis from a previous failure.

If we want to close that gap, triage could stamp the analyzed SHA into the comment, and ci-notify could include the link only when that SHA matches the event SHA.

tomastigera:
Good catch — fixed in 542260b. Rather than stamping the SHA into the triage comment (which would couple the portable `sem-ai-ci-triage` workflow to this one, and wouldn't help on the `workflow_dispatch` test path where there's no SHA), I gated on the comment's `updated_at` vs the status event's `created_at`. Since triage is triggered by the same status event, only a comment rewritten *for the current run* is newer than the event — so we keep polling through triage lag, and if triage skips the run the link is omitted rather than pointing at a previous failure's analysis.

PR #12931, opened 2026-06-08
Make developer Makefiles work on SELinux-enforcing hosts

matthewdupre:
## Description

On Fedora (and other SELinux-enforcing hosts) with Docker's SELinux support enabled, unprivileged containers are denied access to bind-mounted repo files. Ad-hoc `docker run` invocations that mount host paths therefore fail with `permission denied` — for example the local kube-apiserver helper (`run-k8s-apiserver`) crashes on its mounted certs, and `make ut` hangs forever waiting for the apiserver to come up.

Relabelling the repo (`:z`) is slow and mutates host file labels, so instead this disables SELinux label confinement for our build/test containers:

- Define `DOCKER_SELINUX_ARGS` (`--security-opt label=disable`), gated on `/sys/fs/selinux` existing, and append it to `EXTRA_DOCKER_ARGS` so the `DOCKER_RUN`/`DOCKER_RUN_PRIV_NET`/`DOCKER_GO_BUILD` wrapper family is covered automatically.
- Add a `DOCKER_RUN_CMD` shared command for the ad-hoc `docker run` sites that bind-mount host files but don't go through those wrappers, and convert them to use it.

It is a no-op on non-SELinux hosts (the flag expands to nothing) and changes nothing in CI.

## Release Note

Not required; developer visible only

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ check-ocp-no-crds:
 	@CRD_FILES_IN_OCP_DIR=$$(grep "^kind: CustomResourceDefinition" manifests/ocp/* -l || true); if [ ! -z "$$CRD_FILES_IN_OCP_DIR" ]; then echo "ERROR: manifests/ocp should not have any CustomResourceDefinitions, these files should be removed:"; echo "$$CRD_FILES_IN_OCP_DIR"; exit 1; fi
 
 yaml-lint:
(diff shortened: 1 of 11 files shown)


coutinhop (approved):

matthewdupre:
I'm going to close this since it works fine with docker-ce installed from the docker repo, just not docker-cli from the main fedora repos.

PR #12944, opened 2026-06-09
Felix CI: recover BPF UT result when the test VM wedges after tests finish

tomastigera:
Felix BPF unit tests intermittently wedge their test VM via a kernel deadlock (`bpf_prog_free_deferred` vs `drain_vmap_area_work` on `vmap_purge_lock`, observed on `6.17.0-1018-gcp`) during post-test teardown — *after* the tests have already passed. The VM goes SSH-unreachable, the in-VM wrapper never writes `test.rc`, so `run_batch`'s completion poll blocked until the 55-minute watchdog killed the batch and reported a timeout, even though the tests passed.

This decouples the batch verdict from `test.rc`:

- `wait-for-test-completion` is now single-shot (`running` / `<rc>` / `66`); the runner owns the poll cadence and bounds each ssh with `timeout`, so a wedged VM (where even a trivial ssh never returns) can no longer block the monitor loop.
- When the poll is unresponsive, the verdict is recovered from the framework's end-of-run summary line already streamed to the runner host before the VM locked up (gotestsum `DONE N tests…` for `ut`, ginkgo `SUCCESS!/FAIL!` for `fv`), anchored so a mid-run `PASS` can't be mistaken for completion. The serial console is captured and a synthetic JUnit written so the wedge stays visible in test results.
- The SSH teardown steps (`collect-artifacts`, `scp`) and the VM delete are `timeout`-wrapped; `gcloud delete` reclaims a wedged VM via the control plane.

Only a clean end-of-run sentinel counts as a pass; a genuinely hung VM with no completion marker still fails (now in minutes instead of 55), so real failures are never masked.

The underlying kernel deadlock is unfixed upstream through Linux 7.0 / 7.1-rc7; this keeps a passing UT run from being reported as a CI timeout when it recurs.

**Release note:**
```release-note
TBD
```

### DIFF
--- a/felix/.semaphore/wait-for-test-completion
+++ b/felix/.semaphore/wait-for-test-completion
@@ -1,38 +1,45 @@
 #!/bin/bash
+# Single-shot status check for a detached test batch (see batches.sh:run_batch).
+# Prints exactly one token on stdout and exits:
+#   <number>  the batch finished; this is its exit code (read from test.rc)
+#   running   the batch process is still alive
+#   66        the batch process is gone but left no test.rc (crash/reboot)
+#
+# This script must NOT block or loop: the caller polls it repeatedly and owns
+# the overall timeout.  A loop here is exactly what used to wedge the whole job
+# when the VM locked up *after* the tests completed -- the loop would spin on
+# the VM forever, the caller's ssh would never return, and the batch would only
+# end at the 55-minute watchdog.  Returning promptly lets the caller fall back
+# to the streamed-log verdict when the VM stops responding.
 
-while true; do
-  if [ -f test.rc ]; then
-    # Test finished, output its return code and exit.
-    cat test.rc
-    break
-  fi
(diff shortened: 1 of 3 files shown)


felix/.semaphore/wait-for-test-completion
```diff
@@ -1,38 +1,45 @@
...
+  cat test.rc
+  exit 0
+fi
 
-  # Avoid tight loop.
-  sleep 1
+if [ ! -f test.pid ]; then
+  echo "ERROR: test.pid not found" >&2
+  echo "66"
+  exit 0
```
coutinhop, line 23: we should exit 1 (or anything != 0 really) in the error cases (same below)

coutinhop: actually, this would required changes in the caller file, might be fine to leave as is

coutinhop (approved):

PR #12953, opened 2026-06-10
[v3.32] fix(felix): exclude LB-only IPPools from BPF in-pool route flag

defo89:
**Cherry-pick history**
- Pick onto **release-v3.32**: projectcalico/calico#12861

## Description

IPPools with `allowedUses: [LoadBalancer]` caused `poolTypeForPool()` to return `NO_ENCAP`, which set `FlagInIPAMPool` on BPF routes and exempted traffic to LB VIPs from SNAT masquerading. This broke cross-cluster connectivity when a shared LB pool was used across clusters.

### Changes

1. **Return `IPPoolType_NONE` for LB-only pools** in `poolTypeForPool()` via a new `isLoadBalancerOnlyPool()` helper, preventing `FlagInIPAMPool` from being set on BPF routes.

2. **Remove the stale `PoolType != NONE` guard** from the pool update path. This was previously dead code (a non-nil pool never resolved to `NONE`). With the new LB-only check it must be removed to keep LB-only pools in the trie so `NATOutgoing` and `CrossSubnet` continue to propagate correctly.

3. **Unit tests** covering:
   - LB-only pool → `IpPoolType == NONE` and `NatOutgoing` still propagated
   - Mixed-use pool (Workload + LoadBalancer) → `IpPoolType == VXLAN` (not excluded)

### Context

Complements PR #12769 which fixed the iptables/ipset path (`network-ip-pools` ipset). This PR fixes the BPF dataplane path (route map `FlagInIPAMPool` flag).

### DIFF
--- a/felix/calc/l3_route_resolver.go
+++ b/felix/calc/l3_route_resolver.go
@@ -17,9 +17,11 @@ package calc
 import (
 	"fmt"
 	"reflect"
+	"slices"
 	"sort"
 	"time"
 
+	apiv3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
 	"github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/calico/felix/dispatcher"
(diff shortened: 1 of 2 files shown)


mazdakn (approved):

mazdakn:
@defo89 thanks for the pick. The build is failing, due to a function signature that you need to address.

PR #12960, opened 2026-06-11
felix: periodically re-announce proxy-neighbor IPs

mazdakn:
## Description

The proxy-neighbor manager (`felix/dataplane/linux/proxy_neigh_mgr.go`) sent a **single** gratuitous ARP (IPv4) / unsolicited NA (IPv6) when an IP was added to a listener's desired set, and never again for a stable IP. A one-shot announcement is lost if a switch CAM entry or a neighbor's cache later expires or is poisoned (e.g. a live-migration SOURCE→TARGET handover), which can black-hole traffic to a pod/LoadBalancer IP until the next solicitation.

Each listener goroutine now **periodically re-announces every IP it owns**, driven from its existing read loop:

- No new goroutine and no extra locking — the periodic check piggybacks on the loop that already wakes at least every `readTimeout`.
- The **single-owner-socket invariant is preserved**: all socket writes still happen on the listener goroutine. `applyDesiredState` and the new periodic path share one `announce()` helper.
- Multicast group join/leave still happens only on a desired-set delta; the refresh re-sends NAs **without** re-joining the solicited-node group.
- The initial announcement still fires immediately on add; the interval governs only the periodic refresh.

### Configurable interval

The cadence is exposed as a new `FelixConfiguration` field **`LocalSubnetL2ReachabilityRefreshInterval`** (env `FELIX_LocalSubnetL2ReachabilityRefreshInterval`):

- **Default 120s** — comfortably under typical switch CAM aging (~300s) so forwarding entries stay warm, while not below the kernel's own neighbor-cache refresh scale (~30–60s) where it would be redundant.
- **Set to 0 to disable** the periodic refresh entirely (restores the original one-shot-on-add behavior).

### Volume

This manager only runs in `LocalSubnetL2Reachability` mode (no-encap pools on the host L2 subnet). Per-node rate is bounded by `IPs / interval` — roughly 1–4 pps even at high pod density — i.e. far below switch broadcast storm-control thresholds.

## Tests

New specs in `proxy_neigh_mgr_test.go`:

- **IPv4** — repeated, well-formed gratuitous ARPs for the owned IP with no reconcile.
- **IPv6** — repeated unsolicited NAs *and* the solicited-node group joined exactly once (refresh must not re-join).
- **Config wiring** — `LocalSubnetL2ReachabilityRefreshInterval` flows through to the listener's announce interval, and a `0` interval disables periodic re-announcement.

Existing tests pass an interval of `0` (periodic disabled), so their exact write-count assertions are unchanged. Full `felix/dataplane/linux` and `felix/config` suites pass. Generated files (CRDs, OpenAPI, deepcopy, applyconfiguration, manifests, config-params docs) regenerated via `make generate`.

## Release Note

```release-note
Felix proxy-neighbor manager (LocalSubnetL2Reachability) now periodically re-announces (gratuitous ARP / unsolicited NA) the pod and LoadBalancer IPs it owns to keep neighbor caches and switch forwarding tables warm. The cadence is configurable via the new FelixConfiguration field LocalSubnetL2ReachabilityRefreshInterval (default 120s; 0 disables).
```

## Reminders

- [x] Includes tests for the change.
- [x] Release note added.


felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -811,6 +858,15 @@ func (l *ifaceListener) runListener(proto string, setDeadline func(time.Time) er
 		default:
 		}
 
+		// Periodically re-announce every owned IP so neighbor caches and switch
+		// forwarding tables stay warm even when the desired set is unchanged.
+		// The loop wakes at least every readTimeout, so this fires within one
+		// readTimeout of the interval elapsing.
+		if l.announceInterval > 0 && time.Since(l.lastAnnounce) >= l.announceInterval {
```
MichalFupso: Could we drop the lastAnnounce var from the ifaceListener and create a ticker in runListener

```
var tickerC <-chan time.Time
	if l.announceInterval > 0 {
		ticker := time.NewTicker(l.announceInterval)
		defer ticker.Stop()
		tickerC = ticker.C
	}
```

```
select {
		case <-tickerC:
			l.reannounceAll()
		default:
		}
```

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -682,6 +698,32 @@ func (l *ifaceListener) applyDesiredState() {
 	l.announced = desired
 }
 
+// announce sends a single gratuitous ARP (IPv4) or unsolicited NA (IPv6) for
+// addr. It does not touch multicast group membership, so it is safe to call
+// both for a newly desired IP (after joinNDPGroup) and for periodic refresh of
+// an already-joined IP.
+func (l *ifaceListener) announce(addr netip.Addr) {
```
MichalFupso: Looks like the helper method is reintroduced, I think we should add the parseIP to here and drop it from reannounceAll and applyDesiredState

felix/dataplane/linux/proxy_neigh_mgr.go
mazdakn: We can do that, but then we should refactor the whole for loop and use select inside it.

Does not make sense to just use a select for announcement.

MichalFupso: There is a select for reconcile, we can fold these two together.

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -789,45 +813,55 @@ func (l *ifaceListener) runNDPListener() {
 
 // runListener is the read loop shared by the ARP and NDP listeners, handling
 // context cancellation, read deadlines, timeouts and listener recreation on
-// unrecoverable errors, while readAndRespond does the protocol-specific read
-// and reply.
+// unrecoverable errors, while readAndRespond does the protocol-specific read and reply.
```
MichalFupso, line 816: nit: this comment is out of date, runListener now also manages periodic announcment

felix/dataplane/linux/proxy_neigh_mgr.go
```diff
@@ -487,46 +495,38 @@ func (m *proxyNeighManager) reconcileListeners(desiredByIface map[string]set.Set
...
-	// recreates any failed listener dropped above).
+	// Start a listener for each desired interface that doesn't have one, then
+	// publish the latest desired set to it and wake its goroutine.
 	var err error
-	for ifaceName := range desiredByIface {
-		if _, ok := m.listeners[ifaceName]; !ok {
+	for ifaceName, desired := range desiredByIface {
+		l, ok := m.listeners[ifaceName]
+		if !ok {
 			if err = m.startListener(ifaceName); err != nil {
```
MichalFupso, line 504: nit: if startListener returned (ifaceListener, err), we could simplify this to 

```suggestion
			l, err = m.startListener(ifaceName); if err != nil {
```

MichalFupso (approved):

PR #12968, opened 2026-06-12
lib/std/log, lib/logrus: add slog-shaped logging interface

Brian-McM:
Add an abstract logging package at lib/std/log defining an slog-shaped Logger interface (Debug/Info/Warn/Error/With/Enabled), with Level and Attr re-exported from log/slog so callers can build typed attributes without importing slog directly. Until a backend is registered via SetDefaultLogger, package-level calls go through a no-op default.

Add lib/logrus as its own module — a small adapter that wraps a *logrus.Logger and satisfies the lib/std/log.Logger interface. Variadic args to Debug/Info/Warn/Error and With are parsed as slog-style key/value pairs (with slog.Attr supported) and emitted as logrus fields; a non-string key or a dangling odd arg is filed under "!BADKEY", matching slog's convention.

Migrate lib/httpmachinery off direct logrus: per-request loggers are derived via log.With and stored in the request context. Callers use the new shape (e.g. logger.Error("msg", "error", err)) instead of WithError/WithField. The unexported ctxKey type closes the door on context key collisions from outside packages.

Wire whisker-backend to use the new pipeline by calling log.SetDefaultLogger(logrus.New(logrus.StandardLogger())) at startup, so that httpmachinery's per-request loggers (and any other lib code using log.Default) route through the existing logrus configuration. Whisker's own direct logrus calls are intentionally left in place; this PR is about the library interface, not a broad caller migration.

**Release note:**
```release-note
TBD
```


lib/logrusr/adaptor.go
```diff
@@ -0,0 +1,120 @@
...
+func slogToLogrusLevel(level log.Level) logrus.Level {
+	switch {
+	case level >= log.LevelError:
+		return logrus.ErrorLevel
+	case level >= log.LevelWarn:
+		return logrus.WarnLevel
+	case level >= log.LevelInfo:
+		return logrus.InfoLevel
+	default:
+		return logrus.DebugLevel
```
fasaxc, line 124: Think we use trace in a few places, should we define `<debug` -> trace

lib/std/log/logger.go
```diff
@@ -0,0 +1,101 @@
...
+func (n *noOpLogger) Warn(msg string, args ...any)  {}
+func (n *noOpLogger) Error(msg string, args ...any) {}
+
+func (n *noOpLogger) With(args ...any) Logger { return n }
+
+func (n *noOpLogger) Enabled(ctx context.Context, level Level) bool { return false }
+
+func SetDefaultLogger(log Logger) {
+	defaultLogger = log
+}
```
fasaxc, line 107: As written, this isn't thread safe and it's hard to guarantee no logging calls from `init()` scope so perhaps we need to use a mutex or atomics to set it

fasaxc (changes requested):
Couple of queries.  Seems like whisker-backend _has_ to be updated due to using the library, maybe that adds some weight to getting logrus -> slog (or lib/log) API rolled out sooner rather than later.

fasaxc:
Think I'd prefer the package to be logrusadapter or something, just to avoid auto-import annoyance.

lib/logrusr/adaptor.go
Brian-McM: I didn't add it because slog doesn't have it.

lib/std/log/logger.go
Brian-McM: I didn't bother making it thread safe because this is never supposed to be called more than once in an application and we don't run tests in parallel.

If you're swapping out the default logger at runtime then I think the thread safety of this would be the least of my concerns.

fasaxc (approved):

PR #12970, opened 2026-06-12
Ship corresponding source for the patched nftables RPMs

caseydavenport:
We build patched nftables (GPL-2.0-only) and libnftnl (GPL-2.0-or-later) RPMs in `hack/rpms/nftables/` and ship the binaries in calico/node and the istio install-cni image. The corresponding source wasn't in either image. This adds it: the producer stashes the pinned upstream tarballs plus our patches under `/src`, and both consumers copy that into `/included-source` alongside the BIRD and Felix GPL source already there.

Also corrects the libnftnl spec license tag - it's GPL-2.0-or-later upstream (COPYING and every `src/` header), not LGPL-2.1-or-later. The LGPL tag looks copied from libmnl, its build dependency.

```release-note
None
```


fasaxc (approved):
LGTM, hopefully not too big!

PR #12977, opened 2026-06-15
Add Felix calc-graph and dataplane design docs

fasaxc:
Adds two progressive-disclosure sub-designs under `felix/design/`, following the established pattern (the BPF design family + `felix/DESIGN.md` index): narrative architecture, per-section review notes, and a "keep in sync" tail, with thin path-scoped Copilot instruction files.

**Type of change:** documentation only — no code changes.

### What's added

- **`felix/design/calc-graph.md`** — the calculation graph "brain": the node contract (synchronous function-graph, **memoryless**, must handle the consistent→inconsistent transition), the syncer/upstream contract, inter-node ordering, the label-index edge-triggered refcounting, the `EventSequencer` flush order as the dataplane dependency contract, in-sync semantics, and the calc-graph FV testing framework (its five expanders and their blind spots).
- **`felix/design/dataplane.md`** — the Linux (non-BPF) dataplane: the manager (convert) / driver (reconcile) split, the `OnUpdate`/`apply()` event loop and its ordering, the failure philosophy, the **restart/resync mark-and-sweep** doctrine and per-subsystem "how Felix recognises its own state" table, the `*tables` Table abstraction, dispatch chains and mark bits, IP sets, the route drivers, dual-stack, a Windows contrast section, the calc-graph→dataplane proto contract, and a "how to write a manager" recipe.
- **`felix/DESIGN.md`** — marks the `calc-graph` and `dataplane` rows as written, broadens their applies-to globs, and updates the §1 cross-references.
- **`.github/instructions/{calc-graph,dataplane}.instructions.md`** — path-scoped Copilot pointers wiring the docs into automated review.

### Provenance / accuracy

Content was captured by Claude from a detailed design interview with me, then audited section-by-section against the actual code (with the concrete mechanism/symbol claims verified) before submission.


felix/design/calc-graph.md
```diff
@@ -0,0 +1,579 @@
...
+
+| Node (file) | Role |
+|---|---|
+| `ValidationFilter` (`validation_filter.go`) | Nils out invalid resources (treat-as-missing) before they reach the graph |
+| `AllUpdDispatcher` / `localEndpointDispatcher` (`dispatcher` pkg) + `endpointHostnameFilter` (`calc` pkg) | Type-based fan-out; the filter does the local-endpoint reduction |
+| `LiveMigrationCalculator` (`live_migration_calculator.go`) | Adjusts endpoint state for KubeVirt-style live migration |
+| `ActiveRulesCalculator` (`active_rules_calculator.go`) | Tracks which policies/profiles are active given local endpoint labels |
+| `RuleScanner` (`rule_scanner.go`) | Extracts selector/named-port references from active rules; drives the label index |
+| `PolicyResolver` / `PolicySorter` (`policy_resolver.go`, `policy_sorter.go`) | Computes the ordered per-endpoint policy list (tiers, order) |
+| `L3RouteResolver` (`l3_route_resolver.go`) | Computes routes from IP pools, WEPs, host IPs |
```
fasaxc:
```suggestion
| `L3RouteResolver` (`l3_route_resolver.go`) | Computes a generalized route map of Calico-known IP space from IP pools, WEPs, host IPs. Generalized in the sense that some entries are just "we know this useful information about this CIDR" rather than actual IP routes. |
```

felix/design/calc-graph.md
```diff
@@ -0,0 +1,579 @@
...
+   relies on.
+
+### Review notes for this section
+
+- A PR that proposes to "recompute" some derived state wholesale on
+  each update is fighting the architecture. The default is
+  incremental, edge-triggered processing. Wholesale recomputation
+  is acceptable only where the recomputed set is provably bounded
+  by *local* state (e.g. re-sorting one local endpoint's policy
+  list) and never by cluster-wide state.
```
fasaxc: Recomputing wholesale is actually valid for anything small or low churn, not just the example above.  For example, it's perfectly fine to recompute felix configuration from scratch when the handful of felix config resources change.

Full recomputation from cached state (with dedupe on output) is always the simplest option, so it should be used wherever it's the right trade-off.

felix/design/calc-graph.md
```diff
@@ -0,0 +1,579 @@
...
+- A PR that adds per-update work proportional to the *cluster-wide*
+  resource count (not the local count) is a scaling regression and
+  needs an explicit justification or a redesign. This is the
+  calc-graph analogue of the BPF "fast-path cost" rule.
```
fasaxc:
```suggestion
  needs an explicit justification or a redesign.
```
Not sure we need to ref the BPF quote.

felix/design/calc-graph.md
```diff
@@ -0,0 +1,579 @@
...
+- **(a) Synthesize a safe stand-in.** Best when a safe default
+  exists. Felix does this for profile rules: a missing profile is
+  resolved to a fail-safe **deny-all** rule set (the `DummyDropRules`
+  in `calc/active_rules_calculator.go`), so policy still resolves and
+  fails closed.
+- **(b) Buffer the dependent until the dependency arrives.** Usually
+  the **wrong** choice. You must **not** buffer endpoints or
+  policies — they are security-critical, most dependency chains
+  start at endpoints, and delaying them risks leaving traffic
+  unsecured. Reserve buffering for non-security-critical leaves.
```
fasaxc:
```suggestion
  start at endpoints, and delaying them risks leaving endpoints with stale policy/configuration, which could be a security hole. Reserve buffering for non-security-critical leaves.
```

felix/design/calc-graph.md
```diff
@@ -0,0 +1,579 @@
...
+- **(c) Make an explicit exception and handle it in the dataplane.**
+  Pass the inconsistency through to the dataplane as a signal rather
+  than resolving it in the graph, letting the dataplane fail closed
+  on its own terms. Use this sparingly, and only when neither a safe
+  stand-in (a) nor buffering (b) fits.
```
fasaxc: Not sure the advice s quite right, the trade off depends on how much work it is to handle in the dataplanes vs buffering.  If we have to do something in the dataplane anyway (for example, this is a WEP with a new security critical field, so a missing dependency means we need the endpoint to fail closed, buffering is not right and synthesizing a dummy resource might not make any sense for the particular type of resource.

felix/design/dataplane.md
```diff
@@ -0,0 +1,697 @@
...
+1. Clear `dataplaneNeedsSync`; it will be re-set if anything fails.
+2. **Resolution pass** — give every manager (via
+   `UpdateBatchResolver`) a chance to resolve cross-manager state
+   from the batch. This can produce an update from one manager to
+   another (e.g. an endpoint manager telling the BPF endpoint
+   manager about a HEP) that **must** land before either starts
+   programming.
+3. **Programming pass** — call `CompleteDeferredWork()` on every
+   manager.
+4. **XDP** — handled inline right after the programming pass (not a
```
fasaxc: This sounds like Felix's legacy XDP support, which runs as an extension to iptables mode.  This is no longer being enhanced; Felix now has a "proper" BPX dataplane, including XDP untracked policy support.

felix/design/dataplane.md
```diff
@@ -0,0 +1,697 @@
...
+   must be able to look at the live kernel state and decide "this
+   one is mine and unwanted" **without** the original datastore
+   state and **without** a delete edge-trigger.
+
+Therefore: **any feature that creates a new kind of kernel resource
+must, up front, design how a freshly-restarted Felix will recognise
+that resource as Calico's, for later cleanup.** This is a
+first-class design question, not an afterthought.
+
+### How each subsystem identifies "ours"
```
fasaxc, line 287: Not sure where it belongs, but there's a place where the "ours" calculation is important for security:
- Updates from the calc graph may be delayed
- CNI plugin creates a new WEP interface before that information gets written back to the datastore and round trips to felix
- We want to ffail closed, even for an unknown WEP interface and even if FElix is down.
Solution: the iptables / nftables dispatch rules drop any `cali*` interface that isn't allow-listed as part of the dispatch trie/map.  Always think about what happens if felix crashes / stops at a given point, how do we make sure that existing traffic keeps flowing and anything that we can't secure propoerly fails closed.

felix/design/dataplane.md
```diff
@@ -0,0 +1,697 @@
...
+  dataplane, BPF map entries; or, on Windows, HNS policy). This
+  conversion is the manager layer's defining job.
+
+- **Drivers** (`iptables/`, `nftables/`, `ipsets/`, `routetable/`,
+  `routerule/`, `vxlanfdb/`) own the job of **bringing the actual
+  kernel state into sync with the desired state** — reading back
+  what's there, computing a minimal delta, and applying it. Some
+  managers do this reconciliation themselves; others delegate it to
+  a driver and stay purely declarative.
+
```
fasaxc, line 66: One more minor aspect of the dataplane: status reporting back into the datastore.  this is mainly for openstack, but recently we added a file-based local status reporter that can signal to the CNI plugin "felix has programmed this workload's veth correctly" in order to make sure the CNI plugin delays pod startup until dataplane is ready.

felix/design/calc-graph.md
fasaxc: Applied your suggestion verbatim in 2d783ae.

felix/design/calc-graph.md
fasaxc: Good point — reworded the review note in 2d783ae: wholesale recomputation from cached state (with output dedup) is now described as the *simplest* option and fine for small/low-churn state (Felix config called out as an example). The anti-pattern is narrowed to cluster-wide-scaled work on high-churn paths.

felix/design/calc-graph.md
fasaxc: Done in 2d783ae — dropped the BPF analogue sentence.

felix/design/calc-graph.md
fasaxc: Applied your suggestion in 2d783ae.

felix/design/calc-graph.md
fasaxc: Reframed in 2d783ae: (a)/(b)/(c) are now presented as a trade-off (dataplane work vs buffering, and what fits the resource type) rather than a fixed preference order. Noted a dummy stand-in doesn't always make sense, and that (c) is the right call when the dataplane must act anyway (your new-security-critical-field example). The hard invariant is restated as 'fail closed; which mechanism is the judgement call'.

felix/design/dataplane.md
fasaxc: Added a note in 2d783ae clarifying this is Felix's *legacy* iptables-mode XDP (xdpState), no longer enhanced, and pointing at the BPF dataplane for the modern XDP path.

felix/design/dataplane.md
fasaxc: Added a new subsection 'Fail closed while Felix isn't running' in 2d783ae: the CNI-race / Felix-down scenario, the dispatch trie default-dropping unknown cali* interfaces, and the general 'what if Felix crashes at this exact point' doctrine. Also cross-linked it from the dispatch-chains section.

felix/design/dataplane.md
fasaxc: Added as dataplane job (e) in 2d783ae: status reporting into the datastore (WorkloadEndpointStatus, mainly OpenStack) plus the file-based local reporter that gates CNI pod start-up until the veth is programmed.

felix/design/dataplane.md
```diff
@@ -0,0 +1,697 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+
+# Felix dataplane — Design
```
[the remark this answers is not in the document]
fasaxc: The ambiguity is real, but "(non-BPF)" would be the wrong fix: Felix's Linux dataplane is a single `InternalDataplane` codebase switched between iptables/nftables/**eBPF** modes, and this doc deliberately owns the architecture **shared by all three** (manager/driver model, main loop, `apply()` cycle, resync doctrine) plus the `*tables`-specific parts. eBPF reuses that shared architecture and fans out to the `bpf-*` family only for its packet path/maps/managers. Retitled to "Felix Linux dataplane — Design" and rewrote the scope intro to make the shared-vs-`*tables`-specific split explicit, with reciprocal pointers between this doc and bpf-overview.md (commit 2d783ae).

.github/instructions/calc-graph.instructions.md
```diff
@@ -0,0 +1,70 @@
...
+Before writing code (Copilot coding agent) or reviewing a PR
+(Copilot code review) in any file matched by this instruction's
+`applyTo`:
+
+1. Read the relevant section(s) of
+   [`calc-graph.md`](../../felix/design/calc-graph.md) and apply
+   the review notes embedded there. The `labelindex` and
+   `dispatcher` packages are part of the calc graph for these
+   purposes.
+2. Pay particular attention to the **memoryless** node invariant
```
nelljerram: I initially found the "memoryless" term confusing, and not 100% sure about it even now.

My first thought was "oh, you mean stateless".  But of course that's wrong; the calc graph is very stateful.  Now I think you mean that it doesn't have any hysteresis.  Is "memoryless" the best term for that?  It feels like there should be a better one!

.github/instructions/calc-graph.instructions.md
```diff
@@ -0,0 +1,70 @@
...
+   (`calc/calc_graph_fv_test.go`) rather than only per-node unit
+   tests.
+3. The output boundary is the protobuf contract documented from
+   the consumer side in
+   [`dataplane.md`](../../felix/design/dataplane.md); follow that
+   link when a change affects what the graph emits or in what
+   order.
+
+Follow links — the design references siblings, code, and external
+resources.
```
nelljerram, line 39: Are we covering felix/docs/calc.dot anywhere?  I.e. ensuring that is kept up to date.  If not, would that be a good addition here?

felix/design/calc-graph.md
```diff
@@ -0,0 +1,594 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+
+# Felix calculation graph — Design
+
+This is the design doc for Felix's **calculation graph** (the
+"brain"): the code under [`felix/calc/`](../calc/) plus the
```
nelljerram: Does saying `the "brain"` really add anything?  I think it's better if one thing (the calc graph) has one name.

nelljerram: (If you agree, there are a few other occurrences in this PR.)

felix/design/calc-graph.md
```diff
@@ -0,0 +1,594 @@
...
+  when full recomputation would be too expensive.
+- The actual anti-pattern is per-update work whose cost scales with
+  the *cluster-wide* resource count on a *high-churn* path — e.g.
+  touching all 100k endpoints every time any one of them changes.
+  That is a scaling regression and needs an explicit justification
+  or a redesign.
+
+## The node model and the node contract
+
+Calculation nodes are **plain function objects wired into a call
```
nelljerram: Don't know what "plain function objects" means.  "Synchronous functions" perhaps?  But honestly I'm having to guess what the point here is.

felix/design/calc-graph.md
```diff
@@ -0,0 +1,594 @@
...
+and it must remain an invisible implementation detail behind the
+synchronous API.)
+
+Every node must honour the following contract.
+
+### Handle add, update and delete — in any order
+
+The syncer can deliver KVs in any order and can coalesce them
+(below). A node must react correctly to a create, an update, or a
+delete for any resource it tracks, arriving at any time.
```
nelljerram: But for a given single resource I believe we can assume seeing updates in the right order, right?  E.g. something is created before it is deleted.

felix/design/calc-graph.md
```diff
@@ -0,0 +1,594 @@
...
+The subtle, load-bearing half of this: a node must handle the
+transition **from consistent to inconsistent** and end up in
+exactly the same state as if the inconsistency had been present
+from start of day. If resource B disappears while A still
+references it, the node must reconcile to the "B is missing"
+output — it must **not** cling to the last-known-good value of B.
+
+### Be memoryless
+
+This is the single most important node invariant. **A node's output
```
nelljerram: Delete first sentence.  It adds nothing IMO.  Assuming this was mostly Claude-written, I think Claude itself needs a guideline: be less verbose when possible.

felix/design/calc-graph.md
```diff
@@ -0,0 +1,594 @@
...
+start of day with 100 local WEPs and 200 policies each, the naive
+version sorted 200 × 100 = 20,000 times instead of once per WEP =
+100 times. Suppression here was worth it; suppression on a leaf
+that does no downstream work is not worth the complexity.
+
+### Edge-triggered indexing must pair add/remove with identical keys
+
+Where a node maintains reference counts or membership indexes (the
+label indexes are the prime example, below), the add and remove
+operations are **edge-triggered** and must be keyed **identically**.
```
nelljerram: What does "edge-triggered" mean?

felix/design/calc-graph.md
```diff
@@ -0,0 +1,594 @@
...
+- **Non-nil value** means "this is the resource's current state;
+  update yourself to match." The node compares against whatever it
+  held before and reconciles.
+- **Coalescing.** The syncer may collapse a run of updates to the
+  same resource — `update → update → delete → update → update` can
+  arrive as just the final `update`. A node must never assume it
+  sees every intermediate state.
+- **Reversion is possible.** A datastore resync that lands on a
+  node with a stale cache can legitimately move a resource
+  *backwards* to an earlier version. "Compare current event against
```
nelljerram: But, to be clear, there will then be following update(s) to reach the actual latest version, right?

felix/design/calc-graph.md
```diff
@@ -0,0 +1,594 @@
...
+The one concrete load-bearing ordering in `NewCalculationGraph` is
+about **handler registration order on the same dispatcher**, not
+about the order the dispatchers are created (`dispatcher.Register`
+appends per key-type and `OnUpdate` then iterates handlers in
+registration order):
+
+- The `LiveMigrationCalculator` (`calc/live_migration_calculator.go`)
+  has its `OnUpdate` registered on the `localEndpointDispatcher`
+  **before** the `ActiveRulesCalculator` registers with that same
+  dispatcher, so the ARC sees migration-adjusted endpoint state.
```
nelljerram: I couldn't understand "so the ARC sees migration-adjusted endpoint state" here.  The actual comment says the right thing: "This ensures the LMC sees WEP updates first, so its wepData is already populated when the ARC fires computed selector match callbacks."

felix/design/calc-graph.md
```diff
@@ -0,0 +1,594 @@
...
+on deletion you will hear A and B removed *first*, then "A no longer
+matches B".
+
+### The principal nodes
+
+| Node (file) | Role |
+|---|---|
+| `ValidationFilter` (`validation_filter.go`) | Nils out invalid resources (treat-as-missing) before they reach the graph |
+| `AllUpdDispatcher` / `localEndpointDispatcher` (`dispatcher` pkg) + `endpointHostnameFilter` (`calc` pkg) | Type-based fan-out; the filter does the local-endpoint reduction |
+| `LiveMigrationCalculator` (`live_migration_calculator.go`) | Adjusts endpoint state for KubeVirt-style live migration |
```
nelljerram: To me "Adjusts endpoint state" means "modifying the provided model.WorkloadEndpoint" - which is absolutely not true.  What LMC actually does is correlate between input WEPs and LiveMigration resources and emit proto.WEPs to reflect that.  Oh, and also it's for OpenStack as well as KubeVirt.

nelljerram: It worries me that there's a bit too much "make something up in roughly the right ballpark" in this PR, when we have existing code comments that are correct, e.g.:
```
// LiveMigrationCalculator tracks local workload endpoints and LiveMigration resources and
// correlates between them, in order to correctly set the live_migration_role field in
// proto.WorkloadEndpoints that are sent to the dataplane.
```

felix/design/calc-graph.md
```diff
@@ -0,0 +1,594 @@
...
+- A PR that reorders node registration must justify it against the
+  "matches after members" pattern; most reorderings are inert, but
+  the one that isn't — `LiveMigrationCalculator`'s `OnUpdate`
+  registered on `localEndpointDispatcher` before the ARC's — is
+  silently load-bearing.
+
+## Label indexes and refcounting
+
+[`felix/labelindex/`](../labelindex/) holds the graph's most
+intricate machinery: `InheritIndex` (label inheritance from
```
nelljerram: This editorialisation is a tad annoying...  I hope I'm criticising Claude rather than you @fasaxc !

In general I would prefer:
- "show, don't tell" - i.e. just list what the thing does.  Cut the "this next part is amazing, you're going to love it" kind of thing.
- Across the board, more conciseness.

tomastigera:
### The doc-update rule is now repeated across four layers — worth DRYing up here

While reviewing this PR I went looking at how the "update the design doc when you change the code" rule is wired in, and it's stated in **four** places, three of which this PR adds to:

| Layer | Audience / entry point | What it states |
|---|---|---|
| `.claude/CLAUDE.md` → Documentation map | Claude Code | rule + full exemptions list |
| `.github/copilot-instructions.md` → Documentation map | Copilot (always injected) | rule + full exemptions list ("mirrored") |
| `.github/instructions/*.instructions.md` → "Update rule" + "Amending the PR" | Copilot, path-scoped (`applyTo`) | specialized "changes how it works" + `@copilot` boilerplate |
| `felix/design/*.md` → "Keep this document in sync" tail | reader who arrives via the `DESIGN.md` index | specialized rule again |

To be clear, **this PR didn't introduce the repetition — it faithfully follows the established pattern** (every existing instruction file and bpf topic doc already does this). So this is really a pattern-level question. But since this PR is the one extending the pattern to two more instruction files + two more design-doc tails, it's a natural place to fix it before it multiplies further.

There are two distinct kinds of duplication, and only one is worth removing:

1. **Intentional / keep it** — the rule restated per *audience*. The path-scoped Copilot file is auto-injected for matching files, the design-doc tail is read by someone who never sees the Copilot file, and `CLAUDE.md` is for Claude Code. "Say it where the reader is" is legitimate; I wouldn't collapse these.

2. **Accidental / verbatim — worth removing:**
   - The **"## Amending the PR" `@copilot` block** (~13 lines) is copied near-verbatim into *every* instruction file — 4 today (`bpf`, `goldmane`, `ipam`, `helm-charts`), 6 after this PR — differing only in the one-line example. Pure boilerplate, no audience-specific content. A wording change to the amendment mechanism is already an N-file edit.
   - The **exemptions list** (`(a) bug fix… (b) refactor… (c) comment edit… (d) dep bump`) is copy-pasted verbatim into `CLAUDE.md`, `copilot-instructions.md`, *both* new instruction files, and several design-doc tails.

There's also an existing inconsistency this PR inherits: `bpf.instructions.md`'s "Update rule" **references** the repo-wide rule rather than restating it (DRY), but the two new instruction files here **re-list the full exemptions inline**. So the instruction files aren't even consistent with each other yet.

This matters because the cost grows linearly: `felix/DESIGN.md` still lists `route-sync`, `flow-logs-collector`, `config-engine`, and `windows-dataplane` as unwritten. If each lands as its own instruction file with the full boilerplate (the path this PR is on — note it also chose *two separate* instruction files rather than one felix-wide one, unlike BPF which consolidated its whole family into a single `bpf.instructions.md`), the `@copilot` mechanism has to be maintained in lockstep across all of them.

**Suggested fix (small, low-risk, doable in this PR):**

1. Hoist the "Amending the PR" `@copilot` mechanism into `copilot-instructions.md` **once** (it's always injected, so path-scoped files can rely on it). Each `*.instructions.md` keeps only its one-line worked example + a pointer.
2. State the exemptions list once (it already lives in `copilot-instructions.md`) and reference it everywhere else — align the two new instruction files to `bpf.instructions.md`'s reference-don't-restate style.
3. Keep the short design-doc tails (they serve a reader who never sees the Copilot files), but trim to "the repo-wide rule applies; here 'changes how it works' means X" without re-deriving the exemptions — which the new `calc-graph.md`/`dataplane.md` tails already mostly do.

The design content itself is great and I spot-checked the concrete claims against the code (the `Manager` interface, `DummyDropRules` fail-closed, the five FV expanders + `DISABLE_TEST_EXPANSION`, the `LiveMigrationCalculator`-before-ARC registration) — all accurate. This is purely about the meta-rule plumbing.

nelljerram (commented):
Let me drop what I have so far ...  As mentioned out-of-band:
- I worry about building up any source of truth that isn't the code and that might be slightly wrong in comparison with it
- the material here feels about 2x longer than it needs to be, and would be easier to read and understand if it was more concise.

I wonder if it's possible to get a better result by re-prompting Claude with this feedback?

felix/design/calc-graph.md
fasaxc: I've seen new folks read "calc _graph_" and picture a graph of goroutines and channels, was trying to capture the fact that it's a DAG of nodes, but the links between them are simple function calls, and hence no locking or concurrency is implied.

felix/design/calc-graph.md
fasaxc: If you mean the KV type field then yes, but it's quite subtle.  

If a resource goes through states `Created -> A -> B -> C`, the only real guarantee is that, connectivity permitting, you'll eventually see C.  You might see

- `A -> B -> C`
- `A -> C` (common scenario: B skipped due to colaescing)
- `C` (common scenario: A and B skipped due to coalescing)
- `A -> C -> B -> C`  (rare but possible scenario: got to C then resynced against a stale replica, then catch up to C)
- `C -> Deleted -> A -> B -> C` (same but very stale replica that hasn't even reached A yet)

felix/design/calc-graph.md
fasaxc: Correct

.github/instructions/calc-graph.instructions.md
fasaxc: _Claude:_ Agreed the term was confusing (you're right it's *not* stateless — the graph is very stateful). Dropped the coined term throughout and reworded to plain language: a node's **output depends only on current state, not on history (no hysteresis)**. Done in ed17d72.

.github/instructions/calc-graph.instructions.md
fasaxc: _Claude:_ Good call — we weren't. Added a pointer to `felix/docs/calc.dot` in the wiring section and a "update calc.dot when you add/rewire a node" line in both the doc's keep-in-sync tail and the instruction file (in ed17d72). I've also logged a follow-up task to audit calc.dot itself for pre-existing omissions against the current node set.

felix/design/calc-graph.md
fasaxc: _Claude:_ Agreed — dropped the "brain" nickname in the title, intro and the instruction file (and there are no other occurrences left). ed17d72.

felix/design/calc-graph.md
fasaxc: _Claude:_ Deleted that sentence, and took the wider point: did a full concision pass on both docs (calc-graph.md down ~25%), cutting filler and editorialising. ed17d72.

felix/design/calc-graph.md
fasaxc: _Claude:_ Dropped the jargon — it now just says index **add/remove must be balanced and identically keyed** (one remove per add, same key, or it leaks). ed17d72.

felix/design/calc-graph.md
fasaxc: _Claude:_ Replaced my paraphrase with the accurate rationale from the code comment: the LMC's OnUpdate is registered before the ARC's so the LMC sees WEP updates first and its `wepData` is populated when the ARC fires computed selector-match callbacks. ed17d72.

felix/design/calc-graph.md
fasaxc: _Claude:_ Fixed. The row now tracks the godoc: the LMC **correlates local WEPs with LiveMigration resources to set the `live_migration_role` field on emitted `proto.WorkloadEndpoint`s (OpenStack and KubeVirt)** — no mutation of endpoint state. On your broader concern (make-something-up vs. the real comments): I regrounded every node row in that table on its actual godoc, and pulled the real comments before rewriting rather than paraphrasing from memory. ed17d72.

felix/design/calc-graph.md
fasaxc: _Claude:_ Fair — cut the editorialising and switched to show-don't-tell across both docs, and regrounded descriptions on the code comments rather than characterising them. ed17d72.

fasaxc:
_Claude:_ Thanks Neil — taken the overarching points on board in ed17d72:

- **Length / verbosity.** Full concision pass on both docs: dropped editorialising and "you'll love it" framing, cut filler, switched to show-don't-tell. calc-graph.md is ~25% shorter; dataplane.md is denser reference (tables, the ordered apply() list, the per-subsystem identification table) so it got a lighter trim of prose and flourishes rather than gutting the reference content — say the word if you want it cut harder.
- **Source-of-truth drift.** Regrounded the node/type descriptions on their actual godoc (pulling the real comments before rewriting, not paraphrasing from memory), and fixed the LMC description that had drifted. The docs deliberately capture the *why* and cross-cutting invariants that aren't in any single code comment; where a correct comment exists we now track and point at it.
- **Terminology.** Reworded the coined terms to plain language ("memoryless" → output-depends-only-on-current-state; "plain function objects" → a DAG whose edges are function calls; dropped "edge-triggered" and the "brain" nickname).

Per-thread replies above have the specifics.

fasaxc:
_Claude:_ Thanks Tomas — you're right that this is a pattern-level issue and the boilerplate duplication is real. For this PR we've done the **minimal alignment** (in ed17d72): the two new instruction files now *reference* the repo-wide doc-update rule and its exemptions (like `bpf.instructions.md`) instead of re-listing (a)-(d), so they're consistent with the existing files. 

We've deliberately **deferred the larger cleanup** — hoisting the "## Amending the PR" `@copilot` block + the exemptions list into `copilot-instructions.md` once, and the per-topic-vs-felix-wide instruction-file consolidation question — to a separate follow-up PR, since it touches the shared files and every existing instruction file, and is better reviewed on its own rather than expanding the scope of this docs PR. Agreed it's worth doing before the pattern multiplies across the still-unwritten route-sync / config-engine / flow-logs / windows topics.

.github/instructions/calc-graph.instructions.md
fasaxc: Looks like calc.dot is badly stale; I'll ask Claude for a new diagram

nelljerram: The embedded Mermaid diagram is a lovely improvement over the old docs/calc.dot.

.github/instructions/calc-graph.instructions.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Update rule
```
nelljerram:
```suggestion
## Doc update rule
```

.github/instructions/calc-graph.instructions.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Amending the PR
+
+The Copilot automated code-review step is read-only with respect
+to the PR branch — it cannot push the doc amendment itself. When
+the review flags a missing update per the rule above, its comment
+should include a ready-to-paste `@copilot` prompt naming the
+section and the new behaviour or invariant, for example:
+
+> `@copilot update felix/design/calc-graph.md "The EventSequencer (output stage)" to cover the new emitted message proto.FooUpdate — its place in the flush order, what it depends on, and what depends on it.`
```
nelljerram: But IIUC `@copilot` isn't available on almost all PRs - e.g. see attached for this PR.

Image

.github/instructions/dataplane.instructions.md
```diff
@@ -0,0 +1,91 @@
...
+embedded inline at the end of each section.
+
+Before writing code (Copilot coding agent) or reviewing a PR
+(Copilot code review) in any file matched by this instruction's
+`applyTo`:
+
+1. Read the relevant section(s) of
+   [`dataplane.md`](../../felix/design/dataplane.md) and apply the
+   review notes embedded there.
+2. **The headline question for any change that creates kernel
```
nelljerram: Is it really **the headline question**?  I agree it's an important point, but surely not the headline?  The headline must be how the component programs the correct corresponding kernel state.

nelljerram: (Editorialisation, again.)

.github/instructions/dataplane.instructions.md
```diff
@@ -0,0 +1,91 @@
...
+   rules — a BPF PR should load **both** this doc and the relevant
+   `bpf-*` sub-designs.
+
+The input boundary is the protobuf contract documented in
+[`dataplane.md` → The dataplane API](../../felix/design/dataplane.md)
+and from the producer side in
+[`calc-graph.md`](../../felix/design/calc-graph.md). Follow links —
+the design is a graph.
+
+## Update rule
```
nelljerram:
```suggestion
## Doc update rule
```

.github/instructions/dataplane.instructions.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Amending the PR
+
+The Copilot automated code-review step is read-only with respect
+to the PR branch — it cannot push the doc amendment itself. When
+the review flags a missing update per the rule above, its comment
+should include a ready-to-paste `@copilot` prompt naming the
+section and the new behaviour or invariant, for example:
+
+> `@copilot update felix/design/dataplane.md "How each subsystem identifies \"ours\"" to cover how the new fooManager recognises its kernel resources on restart, and what it sweeps.`
```
nelljerram: Same comment about `@copilot` not usually being available.

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+  (see the `EventSequencer`) instead of growing an unbounded backlog.
+- **Local filtering.** Only ~100 local WEPs need policy computed, not
+  the 100k in the cluster, so the graph reduces cluster state to the
+  local subset early (the `localEndpointDispatcher`, below).
+
+Centralising this keeps every dataplane simpler. The graph is the
+single place that handles:
+
+1. **Cross-resource inconsistency** from eventual consistency — a WEP
+   naming a not-yet-known profile, a rule selecting an unknown label.
```
nelljerram:
```suggestion
   naming a not-yet-known profile, a rule selecting a label that doesn't yet match any endpoints, or whose matches can change later.
```

nelljerram: (because there's no such thing as an "unknown" label in the sense of a missing resource for it)

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+2. **Transient duplicate state** — e.g. two WEPs briefly sharing an
+   IP after add/delete reordering; the graph resolves the conflict so
+   no dataplane has to.
+3. **Local relevance filtering** (problem above).
```
nelljerram: what "problem above" ?

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+- `A → B → C`
+- `A → C` — `B` coalesced away
+- `C` — `A` and `B` coalesced away
+- `A → C → B → C` — reached `C`, then a resync hit a stale replica
+  (back to `B`), then caught up
+- `C → Deleted → A → B → C` — resync hit a replica so stale the
+  resource didn't exist there yet, giving a **spurious `Deleted`**
+
+So a node must not assume ordering, monotonic versions, that it sees
+every transition, or even that a `Deleted` is real and final —
```
nelljerram: I'm not sure what this line is advising in practice.  How could code be wrong by making an incorrect "assumption that a `Deleted` is real and final"?  I'm pretty sure this isn't saying "ignore Deletes" - so what **is** it saying?

nelljerram: (I believe there are already many cases where the same resource can be Deleted and then re-Created.)

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+"compare the event against my state and reconcile" handles all of
+these; "assume values only move forward" does not.
+
+### The update-type side channel
+
+Each event also carries `Update.UpdateType` (`api.UpdateType`:
+`UpdateTypeKVNew`/`KVUpdated`/`KVDeleted`/`KVUnknown` in
+`libcalico-go/lib/backend/api`). Original purpose was stats
+without retaining objects (see `calc/stats_collector.go`).
+Treat it as advisory, and note that
```
nelljerram: Does "Treat it as advisory" mean "Usually ignore this UpdateType".

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+`UpdateTypeKVNew/Updated` can carry `nil` due to failed
+validation.
+
+### Review notes
+
+- A node that decides *existence* from the update type rather than
+  nil/non-nil is suspect; stats-style bookkeeping is the legitimate
```
nelljerram:
```suggestion
  nil/non-nil is suspect; stats-style bookkeeping is the only legitimate
```

nelljerram: (if that is correct.  If not, please state the other legitimate uses)

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+| `EventSequencer` (`event_sequencer.go`) | Output stage: buffers, coalesces, flushes in dependency order |
+
+### Review notes
+
+- A PR adding a consumer of "A matches B" events should state where it
+  sits relative to the producers of A and B, and handle the teardown
+  order. Any buffer must rely only on same-loop delivery.
+- A PR reordering node registration must justify it against the
+  matches-after-members pattern; most reorderings are inert, but the
+  `LiveMigrationCalculator`-before-ARC one is silently load-bearing.
```
nelljerram:
```suggestion
  `LiveMigrationCalculator`-before-ARC one is load-bearing.
```

nelljerram: (it's not "silent" if the code comments it - which is stated a few paragraphs above)

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+
+### Coalescing is back-pressure
+
+Coalescing is half the back-pressure mechanism. When the dataplane
+stalls on a big update while the datastore churns, coalescing
+**bounds memory** (at most one pending entry per resource, so ~the
+cluster's resource count) and **bounds the update size** handed to
+the dataplane. Worst case degrades gracefully: the dataplane runs a
+catch-up loop, each pass absorbing the previous pass's churn, rather
+than growing an unbounded queue.
```
nelljerram: I'm not sure I follow this.  I wasn't familiar with how the calc graph experiences back pressure, but I looked that up and now see that it's the `select` in `func (acg *AsyncCalcGraph) onEvent` - i.e. `onEvent` blocks if any of `acg.outputChannels` are full.  With that in mind, what exactly do the following parts of the above mean:

> at most one pending entry per resource

Does "resource" mean one of the `proto.` message types?  If not, what does it mean?  What does "pending" mean - i.e. whereabouts in the overall flow from the datastore to the dataplane?

> bounds the update size

Is that number of proto messages, or the size of each of those messages; or a bit of both?

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+- **(b) Pass the inconsistency through to the dataplane** as a signal
+  and let it fail closed on its own terms. The right choice when the
+  dataplane must act anyway — e.g. a WEP gains a security-critical
+  field, so a missing dependency means *that endpoint* must fail
+  closed.
+- **(c) Buffer the dependent until the dependency arrives.**
+  Requires sending a remove of any already-sent copy of the dependent
+  in order to maintain the calc-graph's "no hysteresis" invariant.
+  Never buffer endpoints/policies/profiles; these are security-critical
+  and part of the core feature set.
```
nelljerram: Is this last sentence part of bullet (c), or a more global instruction?  It reads more like the latter and so should be reformatted as such.

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+silently leave an endpoint or policy *open* — it must fail closed.
+
+### Review notes
+
+- A new message type must document its place in the flush order (and
+  why) and carry FV coverage — wrong-phase bugs only bite under
+  specific orderings, which FV expansion is built to catch.
+- Handling a missing referent by buffering an endpoint or policy is
+  almost always wrong; prefer (a) or (b).
+- Removing or weakening a coalescing path must argue it doesn't break
```
nelljerram: For me, this point is sufficiently obscure that it would help to point to an example.

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+an explicit asymmetric sequence when the teardown-with-live-referrer
+case matters (it usually does for indexes/refcounts).
+
+### Review notes
+
+- A calc-graph (or labelindex) change without an FV state is the
+  exception and needs justification; per-node unit tests aren't a
+  substitute — funnel coverage through the FV framework.
+- Check new states aren't accidentally symmetric (the
+  `[{A},{A,B},{A}]` trap); if so, add an asymmetric sequence.
```
nelljerram: This is a sequence, not a state, right?

In any case, it doesn't seem to me that a symmetric sequence would be necessarily wrong.  I think the important point here was already made under "Blind spot", and the bullet here can be deleted.

felix/design/calc-graph.md
```diff
@@ -0,0 +1,441 @@
...
+   broken, emit a fail-closed stand-in or pass-through signal rather
+   than withholding a security-critical update.
+2. **Buffering security-critical resources** (endpoints, policies) —
+   fail closed instead.
+3. **Recomputing work that scales with cluster-wide counts on a
+   high-churn path** (the historical 20,000-sorts bug).
+4. **Leaky/asymmetric refcounting** — add and remove keyed
+   differently.
+5. **Skipping the FV suite** — so the orderings and teardown paths
+   the expanders would catch go untested.
```
nelljerram, line 430: These points are slightly confusing, in that sometimes the part after the em-dash is "explanation about why this is wrong", and sometimes it is "this is what to do instead".  The "add and remove keyed differently" one risks being read either way, with opposite meanings.

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+```
+
+(Extended interfaces: `ManagerWithRouteTables`,
+`ManagerWithRouteRules` expose route syncers to the loop;
+`UpdateBatchResolver` lets a manager resolve cross-manager state
+before any programming begins.)
+
+**`OnUpdate` must be cheap and must not perform kernel I/O** (no
+netlink call, `iptables-restore`, subprocess fork, or BPF map
+write). The rule is narrower than "don't touch the dataplane": it
```
nelljerram:
```suggestion
write). It
```

nelljerram: (don't think the deleted phrase adds anything)

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+   [resync](#restart-resync-and-mark-and-sweep)).
+
+### Review notes for this section
+
+- Any kernel access (netlink call, `iptables-restore`, subprocess
+  fork, BPF map write) in an `OnUpdate` is a bug. It belongs in
+  `CompleteDeferredWork`. Flag it.
+- `OnUpdate` that does work proportional to anything other than the
+  size of the single message is suspect — the point is to be cheap
+  and defer.
```
nelljerram, line 133: Hm.  liveMigrationMonitor does a few not-obviously-non-blocking things in OnUpdate.  Might need to review that.

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+  insecure state.
+
+### Refresh timers vs error-triggered resync
+
+These are distinct mechanisms with overlapping effect:
+
+- **Error-triggered resync**: a failed `apply()` leaves work dirty
+  and re-runs soon. The re-run is paced by the leaky-bucket
+  `applyThrottle` (with a ~10s `retryTicker` as a backstop), not by
+  exponential backoff; true exponential backoff exists only *inside*
```
nelljerram:
```suggestion
  exponential backoff; exponential backoff is only used *inside*
```

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+| nftables (`nftables/`) | Inherited the iptables hash/prefix approach for porting ease, but doesn't strictly need it: **if it's in our table, it's ours.** A future simplification. |
+| ip rules (`routerule/`) | No marking support. Identified by **the tables they jump to being Felix-owned**. Imperfect — a config change can confuse it. |
+| routes (`routetable/`) | Heuristic, because the first implementation didn't uniformly use the route `proto` field (it should have): **in a Felix-owned table ⇒ ours; carries our proto ⇒ ours; points down a `cali`-owned veth ⇒ ours**; etc. The classifier is the `OwnershipPolicy` interface — `MainTableOwnershipPolicy.RouteIsOurs`/`IfaceIsOurs` in `routetable/ownershippol/`. (Not to be confused with `RouteClass`, which is a same-CIDR conflict tie-breaker among *desired* routes, not an ownership test.) |
+| IP sets, iptables chains, veths | Identified by **name prefix** (`cali`...). |
+
+### Review notes for this section
+
+- **The headline review question for any PR that creates new
```
nelljerram: Again, I agree this is important, but I don't think it's the headline.

nelljerram:
```suggestion
- **A key review question for any PR that creates new
```

nelljerram (commented):
Dropping what I have so far, as I need to switch from computer to phone...  Still need to finish about half of dataplane.md.

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+**security-critical**, because the dataplane has to keep doing the
+right thing in the windows when Felix is **down, restarting, or
+behind**. The motivating race:
+
+- Calc-graph updates can be delayed (Felix busy, restarting, or
+  crashed).
+- The CNI plugin creates a new workload's `cali*` veth and plugs in
+  the pod **before** that endpoint round-trips through the datastore
+  back to Felix — so for a moment the interface exists but Felix has
+  no policy for it.
```
nelljerram, line 326: Perhaps I am misunderstanding, but how can this possibly be a matter for Felix?  ISTM the CNI plugin has to ensure that the new workload is fail-closed when it creates it.

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+
+If the dataplane defaulted to "allow unknown interfaces" that pod
+would have open connectivity until Felix caught up. Instead it
```
nelljerram: Not following.  What do you imagine "allow unknown interfaces" to mean?  Presumably we also don't want Felix to delete a cali interface that it hasn't heard about yet.

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+**fails closed**: the iptables/nftables dispatch chains (the
+[dispatch trie/map](#rules-generation-dispatch-chains-and-mark-bits))
+**drop any `cali*` interface that isn't explicitly allow-listed**.
```
nelljerram: OK, this makes sense now.  The text would be clearer if everything from "If the dataplane" to "**fails closed**" was replaced by just "Therefore".

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+## Rules generation, dispatch chains and mark bits
+
+[`felix/rules/`](../rules/) is the `*tables` rule-rendering layer: it
+converts lists of Calico-internal rules/endpoints/etc into concrete
+`*tables` rules.
+
+- **Dispatch chains avoid O(n-endpoints) per-packet cost.**
+  Evaluating an iptables rule costs on the order of a microsecond
+  (rough estimate; no in-repo benchmark), so a flat chain of 100+
+  per-endpoint dispatch rules is a real per-packet tax. Felix builds
```
nelljerram: Is it really per-packet?  I would hope that our conntrack state shortcut occurs before any dispatch chain - so per-connection instead of per-packet.

nelljerram: More "per-packet"s below - please check and correct them all.

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+  interface that isn't in the trie is dropped, which is the
+  security-critical default that protects not-yet-known workloads
+  (see [Fail closed while Felix isn't running](#fail-closed-while-felix-isnt-running)).
+- **Mark-bit allocation.** Marks are a scarce, shared resource.
+  `MarkBitsManager` (`felix/markbits/`, e.g. `NextSingleBitMark`)
+  allocates bits for `*tables` modes from the configured range. (BPF
+  uses a *fixed*, congested range whose individual bits are managed
+  in a BPF header file — see the BPF design family.) An allocation
+  that exhausts the range, or collides with bits another subsystem
+  expects, is a startup/runtime failure.
```
nelljerram, line 468: Worth a note here about more mark bits being used in Enterprise - so check those in advance to avoid issues at merge time.

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+  `MarkBitsManager` against the configured range, and must not
+  assume a specific bit is free.
+
+## IP sets
+
+[`felix/ipsets/`](../ipsets/) reconciles kernel IP sets against
+desired membership. Three behaviours look odd until you know the
+kernel constraints behind them:
+
+- **Temp-set-and-swap.** Atomically changing the size/metadata of an
```
nelljerram: size = number of entries, or capacity?

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+  (delete IP sets *after* `*tables`) guarantees.
+- **Deletions are rate-limited/queued** because each `ipset destroy`
+  is surprisingly slow (~40ms) and serialised in the kernel: Felix
+  caps deletions per iteration (`MaxIPSetDeletionsPerIteration = 1`,
+  rescheduling with a ~100ms floor), so a big policy teardown of
+  thousands of sets doesn't stall the whole dataplane on cleanup.
+
+**The cross-layer invariant: never reference an IP set before it is
+programmed.** This is enforced jointly across two layers, and a
+change to *either* alone breaks it silently:
```
nelljerram:
```suggestion
change to *either* alone breaks it:
```

nelljerram: (I assume it would be very much not silent - probably a panic in the logs)

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+  have every referencing rule removed — which the `apply()` ordering
+  (delete IP sets *after* `*tables`) guarantees.
+- **Deletions are rate-limited/queued** because each `ipset destroy`
+  is surprisingly slow (~40ms) and serialised in the kernel: Felix
+  caps deletions per iteration (`MaxIPSetDeletionsPerIteration = 1`,
+  rescheduling with a ~100ms floor), so a big policy teardown of
+  thousands of sets doesn't stall the whole dataplane on cleanup.
+
+**The cross-layer invariant: never reference an IP set before it is
+programmed.** This is enforced jointly across two layers, and a
```
nelljerram: Unclear to me if "two layers" means the calc graph and the dataplane, or the manager and a driver in the dataplane.  Please just say exactly what you mean.

felix/design/dataplane.md
```diff
@@ -0,0 +1,687 @@
...
+  on resync" rule (below) matters even more on Windows than on
+  Linux.
+
+## How to write a manager (and common failure modes)
+
+The recurring ways dataplane changes go wrong, and the recipe that
+avoids most of them.
+
+**Start simple: dirty-flag + start-of-day resync.** For a
+low-traffic manager, don't reach for edge-triggered cleverness.
```
nelljerram: What is "edge-triggered"?

nelljerram (commented):
Review complete.

I guess this wouldn't be your practice anyway, but please make sure not to force-push from now on - I definitely don't want to review from scratch again, but it will be nice to keep an eye on the deltas.

felix/design/dataplane.md
fasaxc: Right now, CNI plugin relies on felix having already programmed the "drop traffic to/from unknown veths" rule.

felix/design/dataplane.md
fasaxc: "allow traffic to/from unknown interfaces"

felix/design/dataplane.md
fasaxc: It really is per packet.  Originally, we short-circuited earlier but that caused problems for folks with non-Calico rules; our solution was to move the conntrack accept to the per-endpoint chain.

felix/design/dataplane.md
fasaxc: capacity

felix/design/dataplane.md
fasaxc: It's an analogy to electronics design, it means responding to every change in a signal rather than, say, polling a signal once in a while and responding to its current state.  Agree, better to be concrete and explain what is meant given the analogy doesn't seem to be well known.

nelljerram (approved):
Looks great, thanks for being patient with my reviews 😄

PR #12979, opened 2026-06-15
Add nftables dataplane programming micro-benchmark

caseydavenport:
First step towards measuring (and then improving) nftables dataplane performance: a micro-benchmark that drives the nftables `Table.Apply()` path directly against a real kernel, so we can see the cost of rendering rules and pushing the `nft` transaction without standing up a cluster.

Two benchmarks, each swept across a few scales (IP sets, members per set, policy chains, rules per chain):

- `BenchmarkResync` - full resync against an already-correct dataplane (read back, diff, reconcile).
- `BenchmarkDeltaUpdate` - incremental cost of churning a single IP set member.

Both report wall-clock, per-op CPU, allocations, and the scale dimensions, so the output can be tracked over time.

Modelled on the existing `felix/routetable` benchmarks. Programs a real "calico" table in the host network namespace, so it must run as root (skips otherwise):

    sudo go test ./felix/nftables/ -run '^$' -bench . -benchmem

```release-note
None
```


mazdakn (approved):
LGTM. Just the same nits in the other PR.

PR #12983, opened 2026-06-15
nftables: skip dataplane reload after a no-op Apply

caseydavenport:
Stacked on #12979 (nftables programming micro-benchmark) - the actual change here is the last commit ("skip dataplane reload after a no-op Apply"); the first two commits are that PR and will drop out of this diff once it merges.

---

`NftablesTable.Apply()` invalidated its in-memory dataplane cache after every call, including when the apply wrote nothing. That forced the next `Apply()` to do a full re-read of the table (`nft list` + a JSON parse of every chain, rule, set and map), which is O(total rules).

In steady state that's expensive. A sync that only changes IP set members programs them via the separate IPSets path, leaving the table's own `applyUpdates` a no-op - but it still triggered the reload. So every pod or policy churn on a large node paid a full ruleset re-read on the following sync.

The reload only exists to recover nftables object handles for the objects we just wrote. On a no-op there's nothing to recover, so we can skip it: only invalidate the cache when we actually ran a transaction.

Measured with the micro-benchmark from #12979 (nftables programming micro-benchmark), single IP set member churn at 2000 sets / 2000 chains: steady-state apply drops from ~2.9s and 6.1M allocations to ~17ms and ~2k allocations.

A smaller O(total) cost remains on the no-op path - `applyUpdates` unconditionally deep-copies `chainToFullRules` - worth a separate follow-up.

```release-note
Improved nftables dataplane performance by skipping a full table re-read on dataplane syncs that don't modify nftables rules (for example, IP set member churn).
```

### DIFF
--- a/felix/nftables/bench_test.go
+++ b/felix/nftables/bench_test.go
@@ -1,0 +1,241 @@
+// Copyright (c) 2026 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


caseydavenport:
Folding this into #12984, which rewrites chains wholesale and removes the need for the post-write reload entirely - so this no-op-only guard is subsumed. Closing in favor of that.

PR #12984, opened 2026-06-15
nftables: rewrite changed chains wholesale instead of reloading the table

caseydavenport:
Stacked on #12979 (nftables programming micro-benchmark) - the change here is the final commit; #12979's commits drop out of this diff once it merges.

---

When a policy chain changes, Felix patched it rule-by-rule using nftables rule handles, then reloaded the whole table after every write to recover the handles of what it had just written. That reload is an O(total rules) `nft list` + JSON parse, and at scale it dominated every sync.

Rewrite a changed chain wholesale instead: flush it and re-add all of its rules in the same transaction. nftables applies the flush and adds atomically, so there's no window where the chain is empty. Felix already recomputes the whole chain on any change, and the chains are small (one policy each), so rewriting is cheap.

With no per-rule handles to recover, an enabled table no longer needs to read the dataplane back after a write - and a no-op apply (e.g. one that only touched IP set members, programmed via the separate IPSets path) never needs to either. So the post-write reload goes away for enabled tables entirely, along with the handle bookkeeping (`chainToFullRules` and its per-apply deep copy of the table). Out-of-band drift is still caught by the periodic refresh; disabled tables still reload during cleanup.

This supersedes #12983, which only skipped the reload on no-op applies - folded in here.

Measured with `BenchmarkChainUpdate` from #12979 (single chain rule change, steady state):

| sets / chains | before | after |
|---|---|---|
| 100 / 100 | 151 ms, 291k allocs | 19 ms, ~200 allocs |
| 2000 / 2000 | 3.5 s, 5.8M allocs | 23 ms, ~195 allocs |

After the change the cost is flat in table size - just the one transaction to rewrite the changed chain.

```release-note
Improved nftables dataplane performance at scale by rewriting changed chains in place instead of re-reading the whole table after each programming update.
```


.semaphore/semaphore.yml.d/blocks/45-nft-bench.yml
```diff
@@ -0,0 +1,38 @@
+- name: "Felix: nftables dataplane benchmark"
```
mazdakn: nit: maybe we should just call this `Felix: dataplane benchmark`, so we can potentially add other dataplane benchmark in future.

felix/nftables/bench_test.go
```diff
@@ -0,0 +1,370 @@
...
+	proc      procfs.Proc
+	startCPU  float64
+	startWall time.Time
+	startMem  runtime.MemStats
+}
+
+func record(b *testing.B, s scaleSpec, phase string) *recorder {
+	proc, _ := procfs.NewProc(os.Getpid())
+	stat, _ := proc.Stat()
+	r := &recorder{b: b, spec: s, phase: phase, proc: proc, startCPU: stat.CPUTime(), startWall: time.Now()}
```
mazdakn: nit: would be more readable as multiple lines.

felix/nftables/bench_test.go
```diff
@@ -0,0 +1,370 @@
...
+		})
+	}
+}
+
+// BenchmarkChainUpdate measures the incremental cost of changing a single policy
+// chain's rules and reprogramming, in steady state (no forced resync).  This is
+// the path the flush-and-rewrite change targets: today a chain write invalidates
+// the cache and the next Apply() does a full O(total) reload to recover rule
+// handles; rewriting the chain wholesale removes the need for handles, so no
+// reload is required.
```
mazdakn:
```
// ... today a chain write invalidates
// the cache and the next Apply() does a full O(total) reload to recover rule
// handles;
```

This is not necessary.

felix/nftables/table.go
```diff
@@ -1022,79 +1009,56 @@ func (t *NftablesTable) applyUpdates() error {
...
-						Handle: t.chainToFullRules[chainName][i].Handle,
-					})
-				} else {
-					// currentHashes was longer.  Append.
-					t.logCxt.WithFields(logrus.Fields{
-						"chainName": chainName,
-					}).Debug("Appending rule to chain")
-					tx.Add(t.render.Render(chainName, currentHashes[i], chain.Rules[i], features))
-				}
+		if reflect.DeepEqual(currentHashes, previousHashes) {
```
mazdakn: nit: better to use `slices.Equal` instead.

mazdakn (approved):
LGTM. Left a few nits.

PR #12985, opened 2026-06-15
Generalize the CRD-sync workflow into an upstream-sync bot

caseydavenport:
Generalizes the operator-CRD sync workflow into an upstream-sync bot. It runs the full `make generate` (which already includes the operator CRD pull and manifest regen) so any drifted generated file gets fixed, and bumps the bundled CNI plugin pin (`CNI_VERSION`) to the latest commit on `projectcalico/containernetworking-plugins`. Whatever drifted lands in one combined PR per branch.

Runs hourly, but a cheap drift check up front skips the heavy regen unless something actually changed: a recent merge, an operator CRD update, or the CNI pin falling behind upstream. The operator check reuses `get-operator-crds` with `SKIP_FIX_CHANGED` so it stays Docker-free, so a no-op run is about a minute instead of the full generate.

Only `CNI_VERSION` is bumped, not `FLANNEL_VERSION` (that's a release tag with the Go toolchain baked into the name). A bad CNI bump can't merge - it's gated by the PR's normal CI.

Still needs a `workflow_dispatch` dry run to confirm `make generate` completes cleanly on a GitHub-hosted runner before I trust the schedule.

Related: [CORE-12960](https://tigera.atlassian.net/browse/CORE-12960)

```release-note
None
```

[CORE-12960]: https://tigera.atlassian.net/browse/CORE-12960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/sync-generated-files.yml
```diff
@@ -49,37 +55,107 @@ jobs:
...
+            force=true  # manual runs do the full work (handy for testing / forcing a sync)
+          fi
+
+          # recent: a merge could have changed generated output. The window must
+          # stay larger than the cron interval (4h) so we never skip across a gap
+          # between runs. Coarse on purpose -- we can't know what a merge touched,
+          # so any merge means regenerate.
+          if [ "$force" = true ] || [ -n "$(git log -1 --since='5 hours ago' --format=%H)" ]; then
+            recent=true
+          fi
```
stevegaossou, line 82: I think this is the concern that Shaun had [called out here](https://tigera.slack.com/archives/CC08CBB43/p1780662311610279?thread_ts=1780656659.214909&cid=CC08CBB43): 

- There's a github flag to say "PRs must be up to date with master before merge" but that would drive a lot of CI churn in order to fix occasional issues like this.

Doing it once every 4 hours, does take care of some of the churn (any new commits in that time window get collapsed into one batch), but I wonder if we should maybe even go with a larger time window (once per day)?

stevegaossou: Other approach I guess is to narrow the scope of the of the "recent changes" from any commit to only commits that have changes in specific subdirectories in the codebase. Basically only directories that contain generated files.

.github/workflows/sync-generated-files.yml
```diff
@@ -49,37 +55,107 @@ jobs:
...
+
+          # operator: reuse the same fetch the generate target uses, minus the
+          # Docker-based formatting (SKIP_FIX_CHANGED), then see if anything
+          # changed. Revert the probe's writes; the real run redoes them.
+          make get-operator-crds SKIP_FIX_CHANGED=true
+          if [ "$force" = true ] || ! git diff --quiet; then
+            operator=true
+          fi
+          git checkout -- .
```
stevegaossou, line 91: Question: 

- Does running `make get-operator-crds SKIP_FIX_CHANGED=true` create any untracked (intermediate) files? 

If yes, then running a checkout, will not cleaned up those untracked files.

.github/workflows/sync-generated-files.yml
```diff
@@ -49,37 +55,107 @@ jobs:
...
+
+          # cni: pin behind upstream. Capture the target SHA here so the bump
+          # step doesn't have to query again (and can't race a moving upstream).
+          repo=projectcalico/containernetworking-plugins
+          db=$(gh api "repos/$repo" -q .default_branch)
+          cni_sha=$(gh api "repos/$repo/commits/$db" -q .sha)
+          cur=$(grep '^CNI_VERSION=' metadata.mk | cut -d= -f2)
+          if [ "$force" = true ] || [ "$cni_sha" != "$cur" ]; then
+            cni=true
+          fi
```
stevegaossou, line 101: Looking at the [previous change](https://github.com/projectcalico/calico/pull/12794/changes#diff-50966dfe17854fb9f8269dbb0857f319217834248d7ab87215f153d64d34cb56R60-R65) around the CNI_VERSION value,  does the new change in this pull request lead to the CNI_VERSION updating too frequently?

stevegaossou (commented):
Added a few questions.

.github/workflows/sync-generated-files.yml
caseydavenport: Yeah, the `recent=` check is close to useless on master - 72 commits in the last 7 days, so every 4h window has a merge in it. But a daily window doesn't fix that (25h is still always true), and the PR/CI churn is already bounded: `create-pull-request` no-ops when `make generate` comes back with no diff, so a run that finds nothing costs runner minutes and nothing else. I'd rather keep 4h so real drift shows up the same day it lands.

caseydavenport: `make generate` reads from `api/`, `libcalico-go/`, `felix/`, `goldmane/`, `kube-controllers/`, `lib/`, `manifests/`, `charts/`, `.semaphore/`, `go.mod`, and every `.proto`. That's most of the repo, so the filter wouldn't skip much, and it goes silently stale the first time someone adds a generator input outside the list.

.github/workflows/sync-generated-files.yml
caseydavenport: No - the target loops over the `operator.tigera.io_*.yaml` files that already exist under `charts/` and `curl -o`s over them in place, and `SKIP_FIX_CHANGED=true` skips the Docker formatting step. Nothing untracked gets written, so `git checkout -- .` covers it.

Side effect of that same loop worth flagging: a brand-new operator CRD never gets pulled at all, since there's no local filename to iterate over. That one goes away for free once the operator is in the monorepo.

.github/workflows/sync-generated-files.yml
caseydavenport: Not in practice - the last commit on `projectcalico/containernetworking-plugins` master is from Jan 2025, so there's nothing to chase.

PR #12997, opened 2026-06-16
Include third-party-cni-plugins in operator build

caseydavenport:
**Release note:**
```release-note
TBD
```


fasaxc (dismissed):
Think you need to wire the 3rd party image into the KIND/make push images list too?

PR #13013, opened 2026-06-17
Import third-party-cni-plugins image in Windows FV local-build flow

hjiawei:
The operator now renders the upstream CNI plugins as a separate cni-plugins init container on the calico-node DaemonSet (using the third-party-cni-plugins image) instead of collapsing them onto calico/calico. The win-fv local-build flow only imported calico/calico, calico/node, and calico/operator onto the Linux node, so the new init container hit ImagePullBackOff, calico-node never became Ready, and install-calico.sh timed out before any Felix Windows FV spec ran.

Build the third-party-cni-plugins image from the commit under test and ctr-import it onto the Linux node alongside the others, mirroring the existing in-job Windows image builds and matching how the kind flow was wired up for the same operator change.

**Release note:**
```release-note
TBD
```


hjiawei:
Close in favor of https://github.com/projectcalico/calico/pull/13014.

PR #13034, opened 2026-06-19
lib/logrusr: consolidate logrus utilities

Brian-McM:
## Summary

Move all generic logrus utility code out of `libcalico-go/lib/logutils` and into `lib/logrusr` (the adapter module added in #12968). Component-specific orchestration (`felix/logutils`, `typha/pkg/logutils`, `confd/pkg/log`) is renamed / relocated so that no package is called `logutils` afterwards.

**Moved to lib/logrusr:**
- `Formatter` / `ConfigureFormatter` and the Calico log line shape (timestamp, level, PID, component/file:line, message, sorted KVs).
- Background-dispatch destinations (`NewStreamDestination`, `NewSyslogDestination`) with optional log dropping and Prometheus counters.
- `BackgroundHook` and `WithDebugFileRegexp`.
- `RateLimitedLogger`, `NewRateLimitedLogger`, `OptInterval`, `OptBurst`, `OptLogger`.
- `Summarizer` / `OpRecorder` (from `felix/logutils`).
- `SafeParseLogLevel`, `IsSensitiveParam`, `RedactURL`.
- testing.TB hookup helpers (`RedirectLogrusToTestingT/TB`, `ConfigureLoggingForTestingT/TB`).

The old `logformat` sub-package collapses into `lib/logrusr` — its only external caller (`hack/cmd/coalesce-imports`) is updated.

**Package renames / deletions (all `logutils` names go away):**
- `felix/logutils` → `felix/logging` (Felix-specific `ConfigureEarlyLogging`, `ConfigureLogging`, `Tracef`, `DumpHeapMemoryProfile`, `DumpCPUProfile`, `RegisterProfilingSignalHandlers`).
- `typha/pkg/logutils` → `typha/pkg/logging` (Typha-specific Configure helpers).
- `confd/pkg/log` — deleted. `SetLevel` is inlined into `confd/pkg/config` (the only caller location); `ConfigureFormatter("confd")` moves to `confd.go`'s `main`.

**Custom caller detection (extra frame skipping).** `GetFileInfo` now walks the goroutine stack itself, skipping both `github.com/sirupsen/logrus` and `github.com/projectcalico/calico/lib/logrusr` frames, so wrappers like `RateLimitedLogger`, `Summarizer`, and the `logrusr.New` adapter report the *user's* file (e.g. `endpoint_mgr.go`) rather than the wrapper's file (e.g. `ratelimited.go`). Logrus's own `getCaller` only skips logrus itself, so before this change the reported file was `ratelimitedlogger.go` / `summary.go` in the old locations — the walker is strictly better than either. `SetReportCaller(true)` is turned off since we no longer rely on it; a synthetic `entry.Caller` (e.g. injected by a test) is still honoured as an override. A new test in `destination_test.go` locks the behaviour in.

**Caller updates:** 70+ files across apiserver, app-policy, calicoctl, cni-plugin, confd, felix, goldmane, guardian, hack, kube-controllers, libcalico-go, node, pod2daemon, release, test-tools, typha, webhooks, and whisker-backend.

**Test conversion:** the moved Ginkgo/Gomega tests in `lib/logrusr` are rewritten as raw `go test` (table-driven `t.Run`, select-with-timeout channel helpers, typed field-check helpers). `lib/logrusr/suite_test.go` is deleted; formatter bootstrap moves into a `TestMain`. `lib/logrusr/go.mod` no longer requires `ginkgo/v2` or `gomega`.

`libcalico-go/lib/logutils` is deleted. `cmd/deps.txt`, per-component `deps.txt`, and `.semaphore/semaphore.yml` are regenerated.

## Test plan

- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test ./lib/logrusr/...` passes (raw go tests + fuzz + adapter tests from #12968)
- [x] CI (semaphore) — 63/63 jobs green


confd/pkg/config/config.go
```diff
@@ -7,12 +7,24 @@ import (
...
-	logutils "github.com/projectcalico/calico/confd/pkg/log"
 	"github.com/projectcalico/calico/confd/pkg/resource/template"
 	"github.com/projectcalico/calico/libcalico-go/lib/winutils"
 	"github.com/projectcalico/calico/typha/pkg/syncclientutils"
 )
 
+// SetLevel sets the global logrus level from a string. On parse failure
+// it falls back to InfoLevel and logs a warning. Valid levels are panic,
+// fatal, error, warn, info, debug.
+func SetLevel(level string) {
```
fasaxc: SetLogLevel (given it's in a general config package?)

felix/collector/flowlog/types.go
```diff
@@ -492,12 +492,12 @@ func NewFlowStatsByProcess(
 
 func (f *FlowStatsByProcess) aggregateFlowStatsByProcess(mu *metric.Update) {
 	if stats, ok := f.statsByProcessName[FieldNotIncluded]; ok {
-		logutil.Tracef(f.displayDebugTraceLogs, "Process stats found %+v for metric update %+v", stats, mu)
+		logging.Tracef(f.displayDebugTraceLogs, "Process stats found %+v for metric update %+v", stats, mu)
```
fasaxc: Doesn't logrus support Trace natively now?

lib/logrusr/summarizer.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package logutils
```
fasaxc, line 15: The summariser feels like it should be log-impl independent.  Could come in a later PR, of course.

lib/logrusr/formatter.go
```diff
@@ -225,13 +236,70 @@ func FormatForSyslog(entry *log.Entry) string {
...
 		return FileNameUnknown, 0
 	}
-	return path.Base(entry.Caller.File), entry.Caller.Line
+	frames := runtime.CallersFrames(pcs[:n])
+	for {
+		frame, more := frames.Next()
+		pkg := getPackageName(frame.Function)
+		if pkg != logrusPackage && pkg != logrusrPackage {
+			return path.Base(frame.File), frame.Line
+		}
```
fasaxc: Do we need to extract the package?
```suggestion
		fn := frame.Function
		if !strings.HasPrefix(fn,logrusPackage+".") && !strings.HasPrefix(fn,logrusrPackage+".") {
			return path.Base(frame.File), frame.Line
		}
```

lib/logrusr/formatter.go
```diff
@@ -43,10 +54,10 @@ const (
...
 )
 
-func init() {
-	// We need logrus to record the caller on each log entry for us.
-	log.SetReportCaller(true)
-}
+// SetReportCaller is left off intentionally: our own GetFileInfo walks
+// the stack from Format() and knows to skip both logrus and lib/logrusr
+// frames, so logrus's own caller detection (which only skips logrus)
+// would land on wrapper types in this package instead of the user code.
```
fasaxc: Setting SetReportCaller is now explicitly wrong for this adapter code.  Claude caught some places that still set it to on, which will silently break. 

> key-cert-provisioner/pkg/keycert/run.go:32 still calls logrus.SetReportCaller(true)

lib/logrusr/formatter.go
```diff
@@ -225,13 +236,70 @@ func FormatForSyslog(entry *log.Entry) string {
...
+// code, so the wrapper frames (RateLimitedLogger, log.Logger adapter,
+// Summarizer, ...) are still on the stack — we walk past them to reach
+// the user code that actually asked for the log.
+//
+// If entry.Caller is already set (e.g. a test injecting a synthetic
+// frame, or an integration that populated it directly), we honour it
+// instead of walking.
 func GetFileInfo(entry *log.Entry) (string, int) {
-	if entry.Caller == nil {
+	if entry.Caller != nil {
```
fasaxc, line 343: Now that we don't rely on `entry.Caller`, GetFileInfo can get called 3-4x instead of once, it's also slower by 3x so overall this can add 30us per log (Claude benched it).

Claude suggests:
- Memoizing the result in `entry.Caller`.  This is safe if we do it in an early hook because logrus `Dup()`s the Entry once before sending it to all the hooks (so we can't race with code that is re-using the same Entry in another goroutine).
- Raise the initial skip, we're several layers deep on all code paths.

The benchmarks all set `SetReportCaller(True)` so instead of catching a performance regression with the benchmarks, it's continuing to test the old path.

lib/logrusr/formatter.go
```diff
@@ -17,13 +17,14 @@
 // standard library and logrus so that lightweight command-line tools can
 // import it to get consistent log formatting without pulling in the heavier
 // dependency tree (Prometheus, etc.) of the parent logutils package.
-package logformat
+package logrusr
```
fasaxc, line 26: logformat was _deliberately_ a separate package in order to allow it to be used without pulling in the prometheus client see  #12840

lib/logrusr/formatter.go
fasaxc: Note: you also need to skip `lib/std/log`, or the adapter doesn't get skipped.

lib/logrusr/formatter.go
fasaxc: Stale doc.

fasaxc (changes requested):
Some comments inline, mostly around performance, I think this will regress some previous optimisation work.

Claude also spotted that DESIGN.md references the old logutils package and that filenames stayed the same in the move so they carry the old package name.

lib/logrusr/formatter.go
fasaxc: Might be good if logrusr didn't depend on prom, TBH, it's an unexpected dep, what value do we get from prom in this package?

lib/logrusr/formatter.go
fasaxc, line 241: Drvie-by from Claude: there's a missing `return` here.

lib/logrusr/summarizer.go
Brian-McM: Yes I'd prefer another PR for that, but I agree.

lib/logrusr/formatter.go
Brian-McM: Yes I didn't mean to bring in this dependency, previous iterations just used an interface for the counters but I missed that this time around.

Brian-McM: It's gone now.

lib/logrusr/formatter.go
Brian-McM: Fixed.

fasaxc (approved):

PR #13037, opened 2026-06-22
ci: cut GCS egress for go-build image and master build cache

fasaxc:
Storing the build cache to GCS bucket
(`calico-transient-build-artifacts-europe-west3`) is driving high egress costs,
because Semaphore agents download them on nearly every job. This trims the two
biggest contributors with no change to other branches.

- **Move master's branch-keyed caches to Semaphore's free built-in `cache`
  command.** The ~1.8 GB Go build cache (`build-cache-<group>`) and the
  working-copy tarball now use `cache store`/`cache restore` on master. Other
  branches keep using GCS for now. 
- **Use the cached `calico/go-build` image in GCS as fallback only.** Try a (free!)
  pull from dockerhub first and only fall back to GCS on failure.

### DIFF
--- a/.semaphore/semaphore.yml.d/02-global_job_config.yml
+++ b/.semaphore/semaphore.yml.d/02-global_job_config.yml
@@ -37,8 +37,7 @@ global_job_config:
           # use GCS. (For PRs, SEMAPHORE_GIT_BRANCH is the base branch, so PRs
           # targeting master restore master's Semaphore cache.) Note: `cache
           # restore` exits 0 even on a miss, so we test for the restored file.
-          use_sem_cache=false
-          [[ "${SEMAPHORE_GIT_BRANCH}" == "master" ]] && use_sem_cache=true
+          if [[ "${SEMAPHORE_GIT_BRANCH}" == "master" ]]; then use_sem_cache=true; else use_sem_cache=false; fi
           rm -f /tmp/working-copy.tar.zstd
           if ${use_sem_cache}; then
             cache restore "working-copy-${SEMAPHORE_GIT_BRANCH}" || true
--- a/.semaphore/semaphore.yml.d/02-global_job_config.yml
+++ b/.semaphore/semaphore.yml.d/02-global_job_config.yml
@@ -33,12 +33,30 @@ global_job_config:
         if [[ -n "${SEMAPHORE_GIT_BRANCH}" && -z "$SKIP_CACHE_RESTORE" ]]; then
           echo "Looking for cached working copy for branch ${SEMAPHORE_GIT_BRANCH}"
           gcs_branch_cache_dir="${GCS_CACHE_DIR}/${SEMAPHORE_GIT_BRANCH}"
-          if gcloud storage cp "$gcs_branch_cache_dir/working-copy.tar.zstd" /tmp/working-copy.tar.zstd; then
+          # On master, use Semaphore's free built-in cache; other branches still
+          # use GCS. (For PRs, SEMAPHORE_GIT_BRANCH is the base branch, so PRs
+          # targeting master restore master's Semaphore cache.) Note: `cache
+          # restore` exits 0 even on a miss, so we test for the restored file.
+          use_sem_cache=false
+          [[ "${SEMAPHORE_GIT_BRANCH}" == "master" ]] && use_sem_cache=true
+          rm -f /tmp/working-copy.tar.zstd
+          if ${use_sem_cache}; then
+            cache restore "working-copy-${SEMAPHORE_GIT_BRANCH}" || true
+          else
+            gcloud storage cp "$gcs_branch_cache_dir/working-copy.tar.zstd" /tmp/working-copy.tar.zstd || true
+          fi
+          if [[ -f /tmp/working-copy.tar.zstd ]]; then
             echo "Restoring cache for branch ${SEMAPHORE_GIT_BRANCH}"
             tar --use-compress-program=zstd -xf /tmp/working-copy.tar.zstd & tar_pid=$!
             downloaded_build_cache=false
             if [[ -n "$BUILD_CACHE_GROUP" && -n "${SEMAPHORE_GIT_PR_NUMBER}" ]]; then
-              if gcloud storage cp "$gcs_branch_cache_dir/build-cache-${BUILD_CACHE_GROUP}.tar.zstd" /tmp/build-cache.tar.zstd; then
+              rm -f /tmp/build-cache.tar.zstd
+              if ${use_sem_cache}; then
+                cache restore "build-cache-${SEMAPHORE_GIT_BRANCH}-${BUILD_CACHE_GROUP}" || true
+              else
+                gcloud storage cp "$gcs_branch_cache_dir/build-cache-${BUILD_CACHE_GROUP}.tar.zstd" /tmp/build-cache.tar.zstd || true
+              fi
+              if [[ -f /tmp/build-cache.tar.zstd ]]; then
                 echo "Found build cache for group ${BUILD_CACHE_GROUP}"
                 downloaded_build_cache=true
               fi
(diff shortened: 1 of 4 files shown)


fasaxc:
Thanks @copilot — good catch. Fixed in the latest push.

The "Store working copy" job runs on every branch (`when: "branch =~ '.*'"`), and on a PR `SEMAPHORE_GIT_BRANCH` is the base branch (`master`), so the master/Semaphore-cache path was being taken on PRs — `cache delete` + `cache store` would have thrashed the shared `working-copy-master` key. (I can't simply drop the `delete`: Semaphore cache keys are *not* overwritten, so a refresh genuinely needs delete-then-store.)

The store on the master path is now gated on `-z "${SEMAPHORE_GIT_PR_NUMBER}"`, matching how the build-cache store is already gated. So the cache flows **branch → PR only**: master branch builds populate the cache, PRs only restore it and never write it. Other branches' GCS path is unchanged.

Also rebased onto current master and regenerated — the earlier "Check SemaphoreCI files" failure was stale generated YAML (new `/lib/logrusr` and `/lib/std/log` package triggers landed on master since the branch was cut), not the cache change.

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
```diff
@@ -124,8 +124,22 @@
...
+              # master uses Semaphore's free cache. Only refresh it on real
+              # branch builds, never on PRs: on a PR SEMAPHORE_GIT_BRANCH is the
+              # base branch (master), so storing here would thrash the shared
+              # master cache for concurrent PRs and master builds. PRs only
+              # restore (in the prologue) the copy populated by master builds.
+              # Semaphore cache keys are not overwritten, so delete-then-store.
+              if [[ -z "${SEMAPHORE_GIT_PR_NUMBER}" ]]; then
+                cache delete "working-copy-${SEMAPHORE_GIT_BRANCH}" || true
+                cache store "working-copy-${SEMAPHORE_GIT_BRANCH}" /tmp/working-copy.tar.zstd
+              fi
```
nelljerram: Should there be an inner else block here?  IOW, what happens for a master PR?

.semaphore/semaphore.yml.d/02-global_job_config.yml
```diff
@@ -33,12 +33,30 @@ global_job_config:
         if [[ -n "${SEMAPHORE_GIT_BRANCH}" && -z "$SKIP_CACHE_RESTORE" ]]; then
           echo "Looking for cached working copy for branch ${SEMAPHORE_GIT_BRANCH}"
           gcs_branch_cache_dir="${GCS_CACHE_DIR}/${SEMAPHORE_GIT_BRANCH}"
-          if gcloud storage cp "$gcs_branch_cache_dir/working-copy.tar.zstd" /tmp/working-copy.tar.zstd; then
+          # On master, use Semaphore's free built-in cache; other branches still
+          # use GCS. (For PRs, SEMAPHORE_GIT_BRANCH is the base branch, so PRs
+          # targeting master restore master's Semaphore cache.) Note: `cache
+          # restore` exits 0 even on a miss, so we test for the restored file.
+          use_sem_cache=false
+          [[ "${SEMAPHORE_GIT_BRANCH}" == "master" ]] && use_sem_cache=true
```
nelljerram: This feels like a weird way to write
```
if [[ "${SEMAPHORE_GIT_BRANCH}" == "master" ]]; then use_sem_cache=true; fi
```
Any reason for the above form?  Also I believe the above would fail if any `set -e` logic was in place.

.semaphore/semaphore.yml.d/02-global_job_config.yml
```diff
@@ -78,35 +96,23 @@ global_job_config:
           mkdir -p artifacts
         fi
       - echo $DOCKERHUB_PASSWORD | docker login --username "$DOCKERHUB_USERNAME" --password-stdin
-      # Best-effort: load calico/go-build from the GCS cache populated by the
-      # "Pull: go-build image" prerequisite job. Falls back to a Docker Hub pull
-      # on cache miss. Skipped on non-x86_64 / cross-compile agents since the
-      # cache is amd64-only.
```
nelljerram, line 84: Is there any replacement for the removed GCS use here?  It reads like we introduced the use of GCS to mitigate a problem with pulling calico/go-build from Docker - perhaps rate limits?  If that's right, are we now going to start seeing those problems again?

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
fasaxc: just noticed that this block is guarded by a 
```
when: "branch =~ '.*'"
```
so it only runs on branches, the SEMAPHORE_GIT_PR_NUMBER check is redundant.

The goal is to store the working copy on branch builds only.  On a PR build, it
- retrieves the working copy and build cache from PR's target branch
- does a fetch and checks out the PR commit; the key thing here is that the check out updates mtimes on files that were changed by the PR so that the build cache and working copy correctly relate to each other.

I chose to only store the build cache off on branch builds so that I didn't have to worry about ever-growing build cache.  master builds it clean and stores it off, PRs use it and it's "close enough" for 99% of the PRs that PR builds get good benefit.

.semaphore/semaphore.yml.d/02-global_job_config.yml
fasaxc: I guess we might :-/ it's about 1GB so I'm not sure where else we can stash it

.semaphore/semaphore.yml.d/02-global_job_config.yml
fasaxc: Good point — switched to the plain `if [[ ... ]]; then use_sem_cache=true; else use_sem_cache=false; fi` form. The old `[[ ... ]] && use_sem_cache=true` was just habit, no real reason — and you're right it returns non-zero when the test is false, which would bite under `set -e`. Fixed in e448e1e.

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
fasaxc: Right — as Shaun noted, the block's `when: "branch =~ '.*'"` already limits it to branch builds, so on a master PR this block doesn't run and nothing is stored (the PR only restores, in the prologue). I kept the inner `-z "${SEMAPHORE_GIT_PR_NUMBER}"` guard as cheap belt-and-suspenders and reworded the comment to say so — `.*` would also match an empty `branch` string if a PR pipeline is ever evaluated that way. Happy to drop it entirely if you'd prefer.

.semaphore/semaphore.yml.d/02-global_job_config.yml
fasaxc: Update: reinstated the GCS cache as a fallback in 0c9cb86, rather than relying on Docker Hub alone. Jobs now `docker pull` first (free) and only load the image from GCS if the pull fails — so we keep the robustness against the occasional pull failure you're worried about, but GCS egress stays near zero in the common case. The "Pull: go-build image" prerequisite job is back to keep that fallback populated (free uploads, no-op on a cache hit).

.semaphore/semaphore.yml.d/blocks/10-prerequisites.yml
nelljerram: I would prefer to drop the inner check; it definitely looks wrong with it.  (Even though I understand that it technically is OK.)

fasaxc: Dropped the inner `-z "${SEMAPHORE_GIT_PR_NUMBER}"` guard in b5e8fe — agreed it's better to rely on the one load-bearing `when: "branch =~ '.*'"` than to carry a guard that's redundant most of the time. The store is now just the `master` test + delete/store.

nelljerram (approved):

PR #13044, opened 2026-06-22
Harden the e2e projectcalico.org/v3 client against a rolling calico-apiserver

caseydavenport:
## Description

e2e specs fail outright when calico-apiserver is rolling: cluster config tuning in a setup hook dies with "the server is currently unable to handle the request", and specs reaching the v3 API die with "no matches for projectcalico.org/v3".

- Retry budget in the config helpers looked like two minutes but gave up after about 13 seconds, because the backoff drops its remaining attempts once the interval would pass the cap. It's now a fixed 4 second interval, with a test pinning it.
- Every v3 request retries transient aggregation-layer failures (503, 500, timeouts, an API group that has dropped out of discovery), but not conflicts, where the caller has to re-read first.
- Client setup still picks between the v3 API and calicoctl using discovery, but a single failed or empty discovery call no longer means calicoctl: transient failures retry, and a v3 API that is registered but missing from discovery is waited for.
- The e2e unit tests, which turned out to never run in CI, now do.

Related: [CORE-13228](https://tigera.atlassian.net/browse/CORE-13228)

```release-note
None
```

[CORE-13228]: https://tigera.atlassian.net/browse/CORE-13228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aaaaaaaalex (approved):
LGTM.
The only comment I can think to make is that e2e-UT sounds slightly self-contradictory IMO 😅

caseydavenport:
Yeah, agreed on that one! But oh well.

PR #13046, opened 2026-06-22
node k8st: migrate BGP v4 advertisement to Go

mazdakn:
## Description

Ports the BGP service-IP advertisement system test from Python to Go, continuing the k8st pytest→`go test` migration on the `node-k8st-tests` lineage.

`test_bgp_advert.py` becomes `bgp_advert_test.go` with two top-level tests, each running the former Python test methods as `t.Run` subtests under a shared fixture:

- **`TestBGPAdvert`** (full node-to-node mesh): `cluster_ip_advertisement`, `node_exclusion`, `external_ip_advertisement`, `fully_qualified_service_ips`, `loadbalancer_ip_advertisement`, `many_services`, `bgp_filter_ip_advertisement`
- **`TestBGPAdvertRR`** (route-reflector topology): `rr`, `single_ip_lb_rr`

Each top-level test performs the Python `setUpClass` inline (start the external BIRD router, create the `node-extra.peer` BGPPeer + `bgp-secrets` secret), and subtests get the per-test `setUp`/`tearDown` (namespace create, then ns-delete + mesh-restore + RR-config-clear).

### Conventions
- Calico v3 resources (BGPConfiguration / BGPPeer / BGPFilter) use the controller-runtime client; the v3 **Node** (route-reflector config) isn't in the api scheme, so it is read/applied via `calicoctl` + JSON map mutation.
- Shell-out only for docker/calicoctl; everything else native client-go.
- New shared helpers in the `utils` package: `external_node.go` (`StartExternalNodeWithBGP`, `Curl`, `ExternalNodeRoutes`) and `services.go` (the `test_base.py` deploy/service helpers).

Deletes the migrated `test_bgp_advert.py`.

### Testing
`go build`, `go vet ./node/tests/k8st/...`, `go test -c`, and `gofmt` all pass. Runtime behavior requires a live kind cluster (`make kind-k8st-run-test`), which can't be exercised locally; the port is a 1:1 translation of the Python logic.

**Release note:**
```release-note
None
```


node/tests/k8st/tests/bgp_advert_test.go
```diff
@@ -0,0 +1,1049 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/projectcalico/calico/node/tests/k8st/utils"
+)
+
+const (
+	// bgpTestNS is the namespace each advertisement subtest deploys into.
+	bgpTestNS = "bgp-test"
```
caseydavenport, line 52: This has long been a source of flakes - would be nice to be able to generate a random namespace (and random for any global resources) per-test execution.

caseydavenport (approved):

node/tests/k8st/tests/bgp_advert_test.go
mazdakn: Right. make sense. Will address it.

mazdakn: Will address this in the next PR.

Issue #12192, opened 2026-03-19
bpf - connection to service refused on calico-node pod startup


juliantaylor:
In calico 3.31.4 using ebpf and vxlan we are seeing a lot of connection refused errors for services receiving traffic via nodeports when calico-node pods on the nodes where the nodeport traffic arrives are restarted.

The problem does not occur under the same conditions in calico 3.30.6 and if calico-node pods are restarted on nodes where no nodeport traffic is arriving on (but with workload pods present)

The interrupt occurrs shortly before the calico-node pod becomes ready and affects almost all traffic for about 5 seconds for a simple 2.000req/s small http get service from around 5 source ips (without keepalive).

The felixconfiguration is:

```
spec:
  bpfConnectTimeLoadBalancing: Disabled
  bpfEnabled: true
  bpfExternalServiceMode: DSR
  bpfHostNetworkedNATWithoutCTLB: Enabled
  bpfKubeProxyIptablesCleanupEnabled: true
  bpfKubeProxyMinSyncPeriod: 1s
  bpfLogLevel: "Off"
  bpfMapSizeConntrack: 4096000
  bpfMapSizeConntrackCleanupQueue: 500000
  interfaceExclude: lo,docker0
  ipipEnabled: false
  logSeverityScreen: warning
  reportingInterval: 0s
```

This is if rather small impact for us as we drain all nodes of traffic before we restart calico pods for upgrades so I currently cannot spend a lot of time trying to narrow down the issue further.

I have not tested if other forms of traffic between nodes/pods is affected.

## Expected Behavior
restarting a calico node pod does cause traffic interruptions.

tomastigera:
@juliantaylor what kernel do you use? I just observed some issues that may be related with ubuntu 25.10 and kernel 6.17 in gcp

juliantaylor:
flatcar linux 4459.2.3 using 6.12.74, bare metal

skoryk-oleksandr:
@juliantaylor Thanks. A couple of things that would help with debugging:

1. `calico-node` pod logs during a restart  
   Temporarily set `logSeverityScreen: info` in your FelixConfiguration, restart a `calico-node` pod on an affected node, and capture the logs:

   `kubectl logs -n calico-system <calico-node-pod> -c calico-node --since=2m`

2. Does the issue reproduce with `bpfExternalServiceMode: Tunnel` instead of `DSR`?

juliantaylor:
[...]

PR #13065, opened 2026-06-23
[v3.31] felix: preserve API server service NAT backend across transient endpoint loss

lucastigera:
**Cherry-pick history**
- Pick onto **release-v3.31**: projectcalico/calico#13060

## Description

In the BPF dataplane with `bpfNetworkBootstrap` enabled, Felix reaches the API server **through** the `default/kubernetes` ClusterIP service's NAT entry (its `KUBERNETES_SERVICE_HOST` is the ClusterIP, and the `ebpf-bootstrap` init container seeds the frontend/backend from `KUBERNETES_SERVICE_IPS_PORTS` / `KUBERNETES_APISERVER_ENDPOINTS` before Felix starts).

This creates a self-dependency the generic kube-proxy model doesn't have. When the `kubernetes` service transiently loses all of its ready endpoints — e.g. a single control-plane apiserver restarting (service/process restart, cert rotation, upgrade, node reboot, OOM) removes the sole master endpoint — the syncer wrote `count=0` and deleted the backend. With no backend, new connections to the k8s api server cannot start, so **Felix severs its own connection to the API server** and can no longer learn the restored endpoints. The NAT stays empty until calico-node is restarted (which re-seeds it from the init container).

This only manifests in bootstrap mode, because that is the only configuration where Felix routes itself through the ClusterIP. It also requires the outage to last long enough that Felix's existing watch socket is torn down and a new connect (which goes through the CTLB `connect4` hook) is needed.

## Fix

Retain the last-known-good backend for the API server service whenever an update would leave it with zero ready endpoints, sourced from `prevEpsMap` (rebuilt from the BPF maps on restart by `startupBuildPrev`). The API server's backend is the control-plane node IP, which is stable across such an outage, so the retained backend is correct; a later update with real ready endpoints overwrites it. This is the same defensive shape as #12192 — don't let a transient, incomplete view erase a pre-existing NAT entry that traffic (here, Felix's own) depends on.

The protection is scoped to `default/kubernetes` only. Other services must still drop to zero backends so clients get a connection refusal rather than NAT to a dead backend.

### Unit test

New unit test in `felix/bpf/proxy/syncer_test.go`: programs the API server backend, empties its endpoints and asserts the backend is preserved, asserts a regular service emptied in the same apply still drops to zero (scoping), and asserts convergence when real endpoints return. Verified the test fails without the fix (the API server backend is deleted) and passes with it.

`go test ./felix/bpf/proxy/...`, `go vet`, `gofmt` clean.

### Validated on a live cluster

Reproduced and verified on a single-control-plane kubeadm cluster running the eBPF dataplane with `bpfNetworkBootstrap: Enabled` (calico-node confirmed to reach the API server through the ClusterIP — `KUBERNETES_SERVICE_HOST` is the service VIP).

Repro: move the `kube-apiserver` static-pod manifest out of `/etc/kubernetes/manifests/`, wait ~40s, move it back. The API server's own endpoint reconciler removes the sole control-plane endpoint on the way down, so the `kubernetes` EndpointSlice transiently empties.

| | Stock build (no fix) | Fixed build (this PR) |
|---|---|---|
| `kubernetes` VIP NAT after bounce (`calico component node bpf nat dump`) | drops to `count 0`, backend deleted | stays `count 1`, backend preserved |
| Connect to the VIP from calico-node | `dial tcp <vip>:443: connect: operation not permitted` (EPERM — CTLB `connect4` hard-deny), on every node, persistent | no errors |
| Recovery | required a manual `kubectl rollout restart daemonset/calico-node` | self-recovered, no restart, calico-node pods stayed up with 0 restarts |

In both cases the API server itself was healthy and its `kubernetes` endpoint repopulated after the bounce — confirming the failure was the dataplane retaining an empty NAT, not the API server, and that the fix lets Felix ride out the outage and re-converge on its own.

**Release note:**
```release-note
eBPF dataplane: fix loss of connectivity to the Kubernetes API server service after the API server is unavailable for a period, when using bpfNetworkBootstrap. Felix no longer clears the API server service's NAT backend when its endpoints transiently empty, so connectivity recovers without a calico-node restart.
```

### DIFF
--- a/felix/bpf/proxy/syncer.go
+++ b/felix/bpf/proxy/syncer.go
@@ -423,6 +423,49 @@ func (s *Syncer) applySvc(skey svcKey, sinfo Service, eps []k8sp.Endpoint) error
 	return nil
 }
 
+// isKubernetesAPIServerService reports whether sname is the default Kubernetes
+// API server service (default/kubernetes).
+func isKubernetesAPIServerService(sname k8sp.ServicePortName) bool {
+	return sname.Namespace == "default" && sname.Name == "kubernetes"
+}
+
+// countReadyEndpoints returns the number of ready endpoints in eps.
+func countReadyEndpoints(eps []k8sp.Endpoint) int {
+	n := 0
+	for _, ep := range eps {
+		if ep.IsReady() {
+			n++
+		}
+	}
+	return n
+}
+
+// apiServerFallbackEps returns the API server service's last-known-good
+// endpoints, marked ready.
+func (s *Syncer) apiServerFallbackEps(sname k8sp.ServicePortName) []k8sp.Endpoint {
+	prev := s.prevEpsMap[sname]
+	if len(prev) == 0 {
+		return nil
+	}
+
+	src := make([]k8sp.Endpoint, 0, len(prev))
+	for _, ep := range prev {
+		if ep.IsReady() {
+			src = append(src, ep)
+		}
+	}
+	if len(src) == 0 {
+		src = prev
+	}
+
+	out := make([]k8sp.Endpoint, 0, len(src))
+	for _, ep := range src {
+		out = append(out, NewEndpointInfo(ep.IP(), ep.Port(),
+			EndpointInfoOptIsReady(true), EndpointInfoOptIsServing(true)))
+	}
+	return out
+}
+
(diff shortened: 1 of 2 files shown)


lucastigera:
Suppressed by https://github.com/projectcalico/calico/pull/13066

PR #13067, opened 2026-06-24
Add release-backport skill for cherry-picking master PRs to release branches

skoryk-oleksandr:
## What

Adds a Claude Code skill at `.github/skills/`-style location `.claude/skills/release-backport/SKILL.md` that guides an agent through cherry-picking a merged master PR onto an active release branch, or normalising an existing cherry-pick PR to match the project's title / body / label conventions.

This complements the `cherry-pick-candidate` auto-labeler workflow already merged in `.github/workflows/`: the workflow signals *which* master PRs should be backported, and the skill helps execute the backport. Both are no-ops for contributors who don't use Claude Code; the skill file is only picked up by Claude Code sessions that open this repo.

## Two modes

**Mode A: pick from scratch.** Create a branch off the release branch, cherry-pick the merge commit with `--mainline=1 -x`, resolve conflicts inline (copyright year, import ordering), run codegen if any matching paths changed (`api/`, `*.proto`, `charts/`, `.semaphore/semaphore.yml.d/`), push, open the release PR with the right title prefix (`[v3.32] <original-title>`), body sections (cherry-pick history, original body, conflict notes, metadata expander), and labels (master PR labels minus `cherry-pick-candidate`).

**Mode B: fix up an existing cherry-pick PR.** When the user already ran `hack/cherry-pick-pull` or made the pick by hand, the skill reads the existing release PR, identifies the source master PR from the branch name, the body's `Original PR ID` metadata, or the cherry-pick `-x` SHA, runs the candidacy check on it, computes the right title / body / labels, shows a concrete unified diff, and applies via `gh pr edit`. Mode B is explicitly additive: if the existing body does not appear verbatim in the proposed body, treat it as custom content and inject the cherry-pick-history block above and metadata expander below instead of overwriting.

## Candidacy gate

Both modes run a candidacy check first. If the `cherry-pick-candidate` label is present on the originating master PR, proceed. If absent, apply a heuristic table (bug-fix / crash / race / leak / security keywords lean backport; feature / refactor / test / docs / release-prep / dep-bump lean skip) and ask the user to confirm before picking. The cherry-pick-candidate label is applied by the existing classifier workflow under `.github/workflows/cherry_pick_candidate*.yml`.

## Scope

- Claude Code skill only. No other AI tooling reads `.claude/skills/`.
- Read-only review of the master PR; the only mutating actions are `git`, `gh pr create`, and `gh pr edit` (Mode B). The skill explicitly says never to force-push without asking.
- No new dependencies, no workflow changes, no CI impact.

## Test plan

After merging:
- Use Claude Code in this repo and ask: "backport projectcalico/calico#NNNN to release-v3.32". Confirm the skill triggers, runs the candidacy check, and creates a release PR matching the convention.
- Use Claude Code and ask: "fix up cherry-pick PR #MMMM" (where `MMMM` is an existing release PR created via `hack/cherry-pick-pull`). Confirm the skill enters Mode B, shows a diff, and only mutates on confirmation.

## Release note

\`\`\`release-note
None
\`\`\`

### DIFF
--- a/.claude/skills/release-backport/SKILL.md
+++ b/.claude/skills/release-backport/SKILL.md
@@ -1,0 +1,443 @@
+---
+name: release-backport
+description: >
+  Cherry-pick a merged master PR onto an active release branch in this
+  repo (e.g. master → release-v3.32), or normalise an existing
+  cherry-pick PR that was created by `hack/cherry-pick-pull` or by hand
+  so its title, body, metadata, and labels match the project
+  convention. Use whenever the user asks to cherry-pick, backport, or
+  pick a master PR onto a release branch, or asks to fix up / normalise
+  an existing cherry-pick PR. Triggers: "backport #NNNNN to
+  release-vX.Y", "cherry-pick this to release-vX.Y", "pick #NNNNN onto
+  release-vX.Y", "fix up my cherry-pick PR #NNNNN", "normalise this
+  cherry-pick PR". Single or multiple PRs.
+---
+
+# Release backport
+
+Cherry-pick merged master PRs onto active release branches in this
+repo, or normalise an existing cherry-pick PR. Equivalent to running
+`hack/cherry-pick-pull` but with you resolving conflicts inline
+instead of being prompted interactively, plus a follow-up mode for
+fixing up PRs created by other means.
+
+## Backport candidacy
+
+Release backports happen only for changes that *should* ship in an
+existing release, bug fixes and the like, not new features. This is
+signalled by the `cherry-pick-candidate` label on the master PR,
+applied automatically by a GitHub Action that runs an AI classifier
+and adds (or omits) the label based on its decision. A maintainer can
+still add or remove it by hand.
+
+Before starting any release backport, double-check candidacy, don't
+blindly backport whatever you're handed:
+
+1. **Label check.** Fetch the master PR's labels. If
+   `cherry-pick-candidate` is present, proceed.
+2. **Heuristic check if the label is absent.** Don't silently proceed.
+   Apply the classifier's signals to the PR title/body and surface a
+   recommendation, then ask the user to confirm before picking:
+
+   | Lean **backport** (candidate) | Lean **skip** (not a candidate) |
+   |---|---|
+   | Bug fix: `fix`, `regression`, `incorrect`, `broken`, `stuck`, `hang`, `deadlock`, `should be` | New feature: title starts `Add`/`Implement`/`Introduce`/`Enable`/`Support` |
+   | Crash: `panic`, `nil pointer`, `segfault`, `crash`, `assertion` | Refactor/cleanup: `Refactor`/`Move`/`Rename`/`Drop`/`Remove unused`, "no behavior change" |
+   | Race: `race`, `data race`, `race condition` | Test-only (`*_test.go`, `*fv*`, `*e2e*`) or docs/comment-only |
+   | Leak: `memory leak`, `goroutine leak`, `fd leak`, `unbounded`, `accumulates` | Release prep / version bump |
+   | Security/hardening: `sanitize`, `harden`, `DoS`, `credential`, `token`, `expose` | Dependency update (Renovate/Dependabot/module bump) |
+
+   Tie-breaks: a fix mixed with feature work leans backport (fixes
+   win); if both title and body are vague, lean skip and let the
+   human decide. A false backport is more annoying than a missed one.
+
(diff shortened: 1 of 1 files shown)


.claude/skills/release-backport/SKILL.md
```diff
@@ -0,0 +1,443 @@
...
+| `release-v3.31` | `[v3.31]` |
+| `release-v3.32` | `[v3.32]` |
+
+The prefix is derived by stripping `release-` from the branch name.
+
+Multiple PRs: join stripped titles with `; `.
+
+### Body
+
+Three sections, in this order:
```
lucastigera, line 365: Currently, we can have up to **four** sections. Specifying **three** sections can cause the skill to omit `Section 3: Metadata` when a merge conflict occurs. In that scenario, the skill produces exactly three sections:

Section 1: Cherry-pick history
Section 2: Original PR body
Section 2.5: Conflict resolution notes

As a result, Section 3: Metadata will be missed.

lucastigera: We might want to make this a bit more generic, something like "Produce the following sections in the order shown below." ?

tomastigera:
_Claude:_ Feedback from actually using this skill end-to-end to backport #13135 to release-v3.32 (opened #13137). It worked well for the straightforward pick — the `cherry-pick-candidate` label check fired correctly and the conflict / PR-convention steps were followable. A few things surfaced that are worth folding in, ranked by how much they'd bite:

**1. No version-applicability check (the important one).** The candidacy section gates on bug-vs-feature but never reads the PR body for version scoping — statements like "works on v3.31, breaks on v3.32", version tables, or "regression introduced in vX.Y" — and never compares them to the requested target branch. #13135's body said explicitly (with a table) that v3.31 is unaffected and only v3.32 has the regression. The skill has no step that would stop a "backport to v3.31" request in that case. In this instance the human caught it (the ask started as "v3.31 and v3.32", then was corrected to v3.32-only); the skill did not. Since v3.31 lacks the `WG_PORT` drop block that the PR modifies, the pick onto v3.31 would have hit a conflict on absent context — inviting a "resolution" that fabricates code that shouldn't exist there. The "empty commit → skip" handling does *not* catch this: the change isn't already-present, it's *inapplicable because the bug never existed on that branch*. Suggested fix: add a candidacy step — "if the PR body scopes the bug to specific versions, confirm the target branch is in range; otherwise stop and report."

**2. `git checkout -b` in the main checkout vs. the worktree convention.** Prerequisites ("note the current branch so you can return to it") and Step 3 (`git checkout -b … <remote>/<branch>`) operate in the main checkout. The `~/dev/CLAUDE.md` convention is "never make changes directly in a project's main checkout … work in a git worktree under `.claude/worktrees/`". I deviated to a worktree. Consider having Step 3 do `git worktree add .claude/worktrees/<name>` so the skill and the convention agree.

**3. The Labels section omits the `docs-pr-required` auto-reinstatement trap.** The skill says preserve the master PR's docs/release-note labels, but the repo automation auto-adds `docs-pr-required` on creation. On #13137 the PR ended up with *both* `docs-pr-required` and `docs-not-required` and I had to remove the former by hand. The repo CLAUDE.md documents the ordering ("add the not-required label before removing the required one to avoid the automation reinstating it"). Worth a one-line note in the Labels section.

**4. Conflict guidance misses the dominant diverged-branch case.** The listed "common conflicts" are copyright-year and import-ordering. The conflict I actually hit was: a hunk conflicts because *surrounding, unrelated master-only code* doesn't exist on the release branch (here, the `IPFragTimeout` test machinery interleaved with this PR's `WgPort` additions), so you must keep only the PR's own additions and drop the master-context lines. A naive resolver that takes the whole incoming side would pull in the master-only lines and break compilation. Worth calling out.

**5. modify/delete conflicts not covered.** `felix/design/bpf-overview.md` came through as a modify/delete because `felix/design/` doesn't exist on v3.32 (the design-doc restructure is master-only). The skill doesn't address modify/delete resolution — the right move is to drop the file when the release branch never had it, rather than recreate an orphaned fragment.

**Minor / possibly not worth changing:** the skill uses `gh pr create` / `gh pr edit`; on the Calico repos I used `gh api` REST instead because certain `gh pr` operations hit the "Projects (classic) is being deprecated" error. I didn't verify that `create` specifically fails, so this may be fine as-is — flagging only in case a `gh api` fallback note is worth adding for robustness.

Overall the skill is solid for a same-line-of-development pick; #1 is the one gap I'd want closed before relying on it unattended, since it's precisely the case that almost went wrong here.

PR #13069, opened 2026-06-24
networking-calico: clean up stale election keys on both ends

nelljerram:
Two complementary fixes for the failure mode where the election key is left in etcd after the previous master has gone away.  Without them, a restarted neutron-server has to wait out the full lease TTL (60s) before any node can win the next election.

1. Stale-key cleanup on read (`election._check_master_process`).  When `_vote` reads the current election key and the value parses as `<server_id>:<pid>`, look up `/proc/<pid>`; if the process is no longer running on this host, CAS-delete the key against the observed value and restart the election.  This is the equivalent of the pre-12668 _check_master_process, restored verbatim aside from docstring and wrapping tidy-ups.  Covers SIGKILL / silent greenlet death of a previous master on this host.

2. Clean step-down on graceful shutdown (`CalicoManagerWorker.stop`). `CalicoManagerWorker` now keeps a back-reference to the mech driver and calls `driver.elector.stop()` before chaining to the base `BaseWorker.stop()`.  The elector's stop() blocks until its greenlet has exited, which is what runs the `finally: _attempt_step_down()` that deletes the election key.  Without this hook the elector greenlet was just killed, the finally never ran, and the key lingered.  Covers SIGTERM-initiated shutdown of the master.

Unit tests in test_election.py cover the /proc cleanup paths: live PID (no-op), dead PID with successful CAS-delete, dead PID with CAS-mismatch (restart), dead PID with etcd exception (restart), different host (no /proc check), and unparseable key value (warn only).

Resolves CORE-13033.


nelljerram:
@zhanz1 It has turned out that we need to reinstate the `_check_master_process` logic; otherwise there can be too long a gap in agent status watching when the Neutron server restarts - which in turn can lead to a Felix being considered as dead, and Neutron refusing to bind a port on that Felix's hypervisor.

Please could you review this PR and let me know your thoughts?

networking-calico/networking_calico/plugins/ml2/drivers/calico/workers.py
```diff
@@ -98,14 +98,40 @@ class CalicoManagerWorker(worker.BaseWorker):
...
+        """
+        elector = getattr(self._driver, "elector", None) if self._driver else None
+        if elector is not None:
+            try:
+                elector.stop()
+            except Exception:
+                # Best-effort -- we are already on the shutdown path, and
+                # the key will expire with its lease either way.
+                LOG.exception("Error stopping elector during worker shutdown")
         super(CalicoManagerWorker, self).stop()
```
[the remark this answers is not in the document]
nelljerram: Good catch — added in a5a679cdf2 (`networking_calico/plugins/ml2/drivers/calico/test/test_workers.py`). The new `TestCalicoManagerWorkerStop` covers all three branches:

- **happy path** (`test_stop_steps_down_elector_then_chains_to_super`) — `driver.elector.stop()` called once, then `super().stop()`.
- **elector absent** (three sub-cases) — `_driver` is `None` / `driver` has no `elector` attribute / `driver.elector` is `None`. In each case, `stop()` must not raise and must still chain to `super().stop()`.
- **`elector.stop()` raises** (`test_stop_swallows_elector_stop_exception`) — exception is logged via `LOG.exception`, swallowed, and `super().stop()` still called.

`super().stop()` is verified by patching `worker.BaseWorker.stop` (with `create=True`, since `BaseWorker` itself doesn't define `stop` — it inherits the abstract method from `oslo_service.ServiceBase`) and asserting it's called once per scenario.

networking-calico/networking_calico/plugins/ml2/drivers/calico/workers.py
```diff
@@ -98,14 +98,40 @@ class CalicoManagerWorker(worker.BaseWorker):
...
+        super(CalicoManagerWorker, self).__init__(*args, **kwargs)
+        self._driver = driver
+
     def start(self, name="calico-manager", desc=None):
         """Start service."""
         super(CalicoManagerWorker, self).start(name, desc)
 
     def stop(self):
-        """Stop service."""
+        """Stop service.
```
zhanz1, line 119: This makes sense. IIUC (correct me if I'm wrong), when we do something like `systemctl stop neutron-server`, it will send a `SIGINT`  to all `neutron-server` processes, which each process's handler will catch that and call `stop`?

networking-calico/networking_calico/plugins/ml2/drivers/calico/election.py
```diff
@@ -211,6 +220,44 @@ def _vote(self):
                 )
                 self._become_master()
 
+    def _check_master_process(self, master_id):
```
zhanz1, line 223: The "master" process over here is really the manager process right? If we run this function in the manager process itself (and it somehow died/malfunctioned), then presumably this function will not run either?

This was working before, IIUC, because each process can become master and they can check the master process is alive or not. Now with just one process can be master, we kinda lose this ability. Therefore, for checking malfunctioning/failures//crashes, I would think we should rely on TTL? It would probably make sense to make it configurable.

zhanz1 (commented):
Thanks for the fix and I think overall make sense, have some comments.

networking-calico/networking_calico/plugins/ml2/drivers/calico/workers.py
nelljerram: Yes, I believe so.  So when a shutdown is "graceful" in this way, the master elector deletes its etcd key.  However, in case that doesn't happen (for any reason), we also reinstate the check_master_process logic so that a stale etcd key can be immediately identified when the neutron server starts again.

networking-calico/networking_calico/plugins/ml2/drivers/calico/election.py
nelljerram: Yes, the master process is the "manager" process, because that is the one that includes the Elector.

> If we run this function in the manager process itself (and it somehow died/malfunctioned), then presumably this function will not run either?

I'm afraid I don't understand your query here.  The specific system test scenario here was the neutron-server being restarted in order to pick up a config change.  So the manager process in the old neutron-server is killed, and then a new manager process runs in the new neutron-server.

> This was working before, IIUC, because each process can become master and they can check the master process is alive or not. Now with just one process can be master, we kinda lose this ability.

No, I don't think that's right.  It was working before because of the `_check_master_process` logic.  That was removed in #12668 , and this PR now reinstates it.

> Therefore, for checking malfunctioning/failures//crashes, I would think we should rely on TTL?

What TTL do you have in mind here?  I'm afraid I don't understand your suggestion.

zhanz1: Ah I see what you are referring to here, I believe I misunderstood the intention of this function. Before Chao & I made any changes, IIUC, all neutron-server processes will run `_post_fork_init` and therefore run `Elector`. One of them will be elected as leader, and the rest will just do this `_check_master_process`. With this design, if the old leader process somehow died, then other processes can quickly delete the key and restart the election. When I was writing #12668, I thought this was the intention and hence removed it (because there will only be one process running Elector). I did not think about the restarts. 

In terms of the TTL, I mean the lease TTL (`MASTER_TIMEOUT`). I think it would be helpful to make it configurable so that in case of a machine failure, where no one will delete the election key, we can shrink the time it takes for other machines to step in and become master.

nelljerram: I agree it might be useful for MASTER_TIMEOUT and MASTER_REFRESH_INTERVAL to be configurable.  But I think that is independent of the current PR, isn't it?  10s is already quite low for MASTER_REFRESH_INTERVAL.

zhanz1: Yeah ofc, this can be a separate PR. While the refresh interval is 10s, I think it's the 60s timeout who's blocking longer for other neutron-server to step up?

zhanz1 (commented):
Make sense to me and thanks for fixing this!

networking-calico/networking_calico/plugins/ml2/drivers/calico/election.py
nelljerram: Yes, but (in case it's not clear) the current PR already fixes that.  When the Neutron server restarts:
1. It immediately creates the ManagerWorker process, -> `_init_start_calico_manager()` -> `self.elector.start()`
2. Elector calls `_vote`, which finds the key and calls `_check_master_process`
3. `_check_master_process` parses successfully, finds that host matches its own, but PID no longer running, and so deletes the key.
4. `_vote` now either sees the delete, or `KeyNotFound`, and so calls `_become_master`.

The important point is that that all happens immediately without waiting for any timeout or refresh interval.

zhanz1: Yes, this is happy path where the `neutron-server` is restarted and I think we are on the same page here :D. I'm more referring to when there is a machine failure (i.e., the machine crashed) and thus the `neutron-server` is never restarted (because it can't), and it would then take this 60s for other machines running `neutron-server` to become master - and in that case if we want to reduce the downtime, we'll need to reduce the 60s - but this should be another PR as we discussed.

nelljerram: Thanks, that makes sense.  Would you like to prepare that PR?

zhanz1: Sure, can do when I get a chance.

coutinhop (approved):

PR #13075, opened 2026-06-24
Only garbage collect IPAM blocks with elapsed cooldown IPs

caseydavenport:
Follow-up to #12638. The cold-IP GC backstop in kube-controllers walked every IPAM block on every sync trigger - each batched pod-delete, node-delete, syncer update, plus the periodic tick. That's a lot of work on a loop that's critical for IPAM GC in some scenarios, and we've been bitten before by doing too much in it.

This tracks which blocks have IPs in cooldown as block updates stream in, so the backstop only visits those blocks, and only once their oldest cooldown IP has actually finished cooling down. When nothing is cooling down, it does no work.

The design doc for the cold-IP GC backstop is missing entirely (the cooldown feature merged without one), so I'm handling that in a separate follow-up rather than partially documenting it here.

```release-note
None
```

### DIFF
--- a/kube-controllers/pkg/controllers/node/ipam.go
+++ b/kube-controllers/pkg/controllers/node/ipam.go
@@ -1368,10 +1368,13 @@ func (c *IPAMController) garbageCollectColdIPs() error {
 	for cidr, earliestReleasedAt := range c.coldBlocks {
 		if earliestReleasedAt.Add(cooldown).After(now) {
 			// The oldest cooldown IP in this block hasn't finished cooling down yet.
+			log.WithField("block", cidr).Debug("Block has IPs in cooldown but none are ready to deallocate yet")
 			continue
 		}
 		kvp, ok := c.allBlocks[cidr]
 		if !ok {
+			// coldBlocks should always be a subset of allBlocks; see assertConsistentState.
+			log.WithField("block", cidr).Warn("Block tracked as having cooldown IPs is missing from the block cache")
 			continue
 		}
 		if err := c.client.IPAM().GarbageCollectColdIPs(ctx, ipamConfig, &kvp); err != nil {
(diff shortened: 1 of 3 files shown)


kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1336,7 +1362,18 @@ func (c *IPAMController) garbageCollectColdIPs() error {
...
+	cooldown := time.Duration(ipamConfig.IPCooldownSeconds) * time.Second
+	now := time.Now()
+	for cidr, earliestReleasedAt := range c.coldBlocks {
+		if earliestReleasedAt.Add(cooldown).After(now) {
+			// The oldest cooldown IP in this block hasn't finished cooling down yet.
+			continue
+		}
+		kvp, ok := c.allBlocks[cidr]
+		if !ok {
+			continue
```
caseydavenport, line 1374: Probably warrants a warning log.

kube-controllers/pkg/controllers/node/ipam.go
```diff
@@ -1336,7 +1362,18 @@ func (c *IPAMController) garbageCollectColdIPs() error {
 	if err != nil {
 		return err
 	}
-	for _, kvp := range c.allBlocks {
+
+	cooldown := time.Duration(ipamConfig.IPCooldownSeconds) * time.Second
+	now := time.Now()
+	for cidr, earliestReleasedAt := range c.coldBlocks {
+		if earliestReleasedAt.Add(cooldown).After(now) {
+			// The oldest cooldown IP in this block hasn't finished cooling down yet.
```
caseydavenport, line 1366: Probably warrants a debug log.

fasaxc (dismissed):
Looks correct as it stands but maybe we should refactor releaseUnusedBlocks to share cache clean up with onBlockDeleted so that, when we delete a block, we always clean up all the cache maps, including coldBlocks?  They can't get out of sync right now, but only because releaseUnusedBlocks shouldn't be called for a cold block.

caseydavenport:
Yeah, good call. Pulled both paths through a `forgetBlock` helper so every block teardown cleans up all the maps, `coldBlocks` included. `releaseUnusedBlocks` only ever sees empty blocks so the allocation cleanup is a no-op there, but it keeps the two paths from drifting if that ever stops being true.

Pushed in 43829e7.

fasaxc (approved):

PR #13078, opened 2026-06-24
[BPF] support ip6 protochain in pcap log filters

tomastigera:
## What

Teaches the BPF pcap log-filter compiler (`felix/bpf/filter`) to compile
`ip6 protochain`, and adds IPv6 coverage for the filter compiler.

`BPFLogFilters` lets an operator target debug logging with a tcpdump-style
expression that Felix compiles to eBPF at runtime (`cBPF2eBPF`). libpcap
compiles `ip6 protochain` — which walks the IPv6 extension-header chain to
reach the L4 protocol — into a loop with a **backward `BPF_JA`**. Classic
BPF only ever runs such filters in libpcap's userspace interpreter, and the
eBPF verifier rejects the unbounded back-edge, so the expression failed to
load as a `BPFLogFilter`.

## How

- **Loop unrolling.** `cBPF2eBPF` detects a single backward-jump loop region
  and unrolls it a fixed `maxLoopUnroll` times: a forward jump stays in its
  copy, a back-edge becomes a forward jump into the next copy, and the last
  copy's back-edge falls through to `miss` (chain longer than the bound ⇒
  no match). This mirrors the bounded `for (i = 0; i < 8; i++)` extension-
  header walk in `felix/bpf-gpl/parsing6.h`. Anything more complex than a
  single reducible loop is rejected at compile time rather than mis-compiled.
- **Verifiable indirect loads.** Indirect loads now bound their scalar
  offset against the loaded length before forming the pointer, so the
  verifier can prove the load stays inside the scratch buffer even when the
  offset register grows across loop iterations. (A pointer-vs-pointer bounds
  check does not bound the offset for the verifier once it can exceed the
  buffer size.)
- Wires up the cBPF instruction classes the loop body uses (ALU ops, scratch
  memory `M[]`, `TAX`/`TXA`, immediate/scratch `LDX`) and fixes `BPF_JA`
  target computation, which previously ignored the signed `K` offset.
- Documented in `felix/design/bpf-observability.md`.

## Testing

`TestFilterIPv6` (new) covers IPv6 protocol / host / port / length /
portrange and `protochain` expressions over both Ethernet and raw-IPv6 link
types, plus negations — mirroring the existing IPv4 `TestFilter`. Verified
locally with `make -C felix FOCUS=TestFilter ut-bpf` (102 subtests pass; the
compiled programs, including `protochain`, pass the kernel verifier).

## Notes

Supersedes #11301 (Copilot-authored), carrying over its IPv6 tests and
instruction-translation work and adding the loop unrolling that makes
`protochain` actually load. Extension headers beyond the first `maxData`
bytes of the packet remain out of reach, as the filter matches against the
pre-loaded scratch buffer.

**Release note:**
```release-note
Felix BPF debug log filters (BPFLogFilters) now support the `ip6 protochain` pcap primitive.
```

### DIFF
--- a/felix/bpf/filter/filter.go
+++ b/felix/bpf/filter/filter.go
@@ -148,9 +148,8 @@ func programHeader(b *asm.Block, maxData int, stateMapFD maps.FD) {
 	// Check if bpf_skb_load_bytes succeeded (returns 0 on success)
 	b.JumpNEImm64(asm.R0, 0, "exit")
 
-	// Set up R8 to point to the end of actually loaded data (R7 + actual length)
-	b.Mov64(asm.R8, asm.R7)
-	b.Add64(asm.R8, asm.R9)
+	// R9 holds the number of bytes loaded into the scratch buffer (R7);
+	// indirect loads bound their offset against it.
 
 	b.LabelNextInsn("filter")
 	// Zero R1 (A) and R2 (X)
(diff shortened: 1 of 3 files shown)


tomastigera:
Thanks @copilot — addressed in the latest push (force-pushed into the single commit):

- **Indirect-load width** — the bounds check now guards the *last* byte read (`X+K+width-1 >= len → exit`) before forming the pointer, so multi-byte loads can't match on bytes past the loaded data. (Note: this was never an actual verifier OOB — the scratch buffer is 512B and the offset stays well under that — but matching on stale bytes was a real nuance worth closing. `protochain` only does byte loads, so its path is unchanged.)
- **`BPF_LD` / `BPF_LDX` missing `default`** — both switches now return a "unsupported mode" error instead of falling through to the generic `b.Instr` and silently miscompiling. Also wired up `LDX|LEN` (packet length into X).
- **`jumpTargets` reporting fall-through** — fixed to only report non-zero `Jt`/`Jf` offsets, so a conditional jump's fall-through (offset 0) is no longer treated as an explicit jump target in the loop-entry validation.

- **`maxLoopUnroll = 4` vs the C bound of 8** — intentional, not drift. A debug log filter doesn't need to follow every extension-header chain to the end, and headers past the first `maxData` bytes are unreachable anyway (the filter matches the pre-loaded scratch buffer). Reworded the comment/design note so they state this explicitly rather than implying parity with `parsing6.h`.

Re-ran `make -C felix FOCUS=TestFilter ut-bpf`: 102 subtests pass, no regressions.

felix/bpf/filter/filter.go
```diff
@@ -205,129 +204,463 @@ func fromBE(b *asm.Block, size uint8) {
...
+			target = i + int(cbpf.Jt) + 1
+		}
+		code = (code & 0xf) | op
+
+		b.InstrWithOffsetFixup(asm.OpCode(code), asm.R1, srcR, label(target), K)
+
+		return nil
+	}
+
+	dstR := asm.R1
```
sridhartigera: Do we need this now. switch (class) handles all 8 cases.

sridhartigera (commented):
I think its better to add an UT with ipv6 extension header to test the loop unrolling code.

tomastigera:
Thanks @sridhartigera — both addressed:

- **UT with extension headers** — added `TestFilterIPv6Protochain`, which builds IPv6 packets carrying hop-by-hop extension headers before the UDP header (one and two of them) and runs `ip6 protochain 17`/`6` against them. The existing `TestFilterIPv6` packet has UDP directly after the IPv6 header, so protochain matches at the loop header and never enters the body; these packets force the walk through one and two unrolled iterations. All four sub-tests pass:
  ```
  TestFilterIPv6Protochain/1ext_ip6_protochain_17  PASS
  TestFilterIPv6Protochain/1ext_ip6_protochain_6   PASS
  TestFilterIPv6Protochain/2ext_ip6_protochain_17  PASS
  TestFilterIPv6Protochain/2ext_ip6_protochain_6   PASS
  ```
- **Dead generic emit (filter.go)** — you're right, with the `default` cases now in the `LD`/`LDX` switches every one of the 8 `bpfClass` cases returns, so the trailing `b.Instr(...)` was unreachable. Removed it (and the now-unused `offset`); the end of `emitInsn` returns an "unhandled instruction class" error as a backstop.

sridhartigera (approved):
LGTM

PR #13093, opened 2026-06-25
felix/bpf: add per-rule Match summary to policy debug dump

sridhartigera:
### Description

The plain eBPF policy debug dump (`calico-bpf policy dump <iface> <hook>`) now prints a single positive **`Match:`** line directly under each rule header, summarising the rule's L3/L4 criteria on one line: protocol, source/destination nets, ports and ICMP type/code, with negated fields prefixed by `!`.

Selector-resolved IP sets are folded into the matching `src=`/`dst=` token: the builder emits them as hex IDs and the dump resolves those to live member IPs. This **replaces the previous standalone `IP sets:` line**, so the complete match (literal nets *plus* selector members) reads on a single line, e.g.:

```
Rule: policy-tcp  Action: allow
Match: proto=tcp src={11.0.0.8/32,10.0.0.8/32,10.65.0.3/32} dst={12.0.0.8/32,13.0.0.8/32} sports={8055,100-105} dports={9055,200-205}
Hit count: 0
```

The summary is shown **only in the plain dump**. The asm (`-a`) dump already prints the equivalent per-criterion `// If ... skip to next rule` comments inline with the instructions, so the summary is omitted there to avoid redundancy.

### Testing done

- Unit tests:
  - `TestFormatRuleMatch` — match-line formatting incl. folded selector IP sets, negation, named-port markers, ICMP.
  - `TestResolveMatchIPSets` — hex-ID → live-member-IP resolution (full / partial / no-map / no-op).
  - `TestPolicyDumpMatchComment` — end-to-end builder wiring: `Match:` emitted with folded hex IDs, old `IPSets` summary comment gone, ordered between the rule header and the hit count.
- FV: `felix/fv/bpf_policy_dump_test.go` updated — asserts the folded/resolved `Match:` line in the plain dump and that it is absent from every `-a` dump.

**Release note:**
```release-note
None
```

### DIFF
--- a/felix/bpf/polprog/pol_prog_builder.go
+++ b/felix/bpf/polprog/pol_prog_builder.go
@@ -544,9 +544,22 @@ func (p *Builder) writePolicyRules(policy Policy, actionLabels map[string]string
 	for ruleIdx, rule := range policy.Rules {
 		log.Debugf("Start of %s rule %d", policy.NamespacedName(), ruleIdx)
 		p.b.AddCommentF("Start of rule %s %s", policy.NamespacedName(), rule)
-		ipsets := p.printIPSetIDs(rule)
-		if ipsets != "" {
-			p.b.AddCommentF("IPSets %s", p.printIPSetIDs(rule))
+		// Emit the positive match summary directly under the rule header so it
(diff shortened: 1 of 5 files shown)


tomastigera (approved):
cool!

PR #13106, opened 2026-06-26
ci(deps): add CHANGE_IN_WITH_DEPENDENTS macro for artifact producers

fasaxc:
Producer CI blocks that save off artifacts for later blocks to use must run if any of the dependent blocks run.  The recent changes to add a monobinary image have introduced a lot of such transitive dependencies that we've ben handling in ad-hoc fashion.

This adds a `CHANGE_IN_WITH_DEPENDENTS(<own-spec>)` macro to `hack/cmd/deps` that derives the superset automatically from the existing `dependencies:` graph, so a producer can never fall behind its consumers again.

- Expands to a **single** `change_in(...)` clause (the folded-scalar formatting is unchanged), merging the producer's own intrinsic spec with every dependent's resolved trigger.
- Merge rule: **union inclusions, intersect exclusions** — a provable superset that never under-fires (it may slightly over-fire, which is harmless for these content-addressed, cache-cheap producers).
- The tool now YAML-parses blocks into a dependency graph, resolves **chained** producers recursively (with a cycle guard), accepts `<const> or ${CHANGE_IN(...)}` dependents, and validates that every `dependencies:` entry names a real block.

Migrated producers: `Build: nftables RPMs`, `Build: calico image`, `Build: node image`, `Felix: Build`, `Felix: Build Windows binaries`. This also closes a latent gap where `Felix: Build Windows binaries` (`CHANGE_IN(felix)`) was narrower than its consumer `Felix: Windows FV` (which also triggers on `process/testing/{aso,util,winfv-felix}`).


fasaxc:
this seem to do the trick when I merge it into PRs like https://github.com/projectcalico/calico/pull/13131, which previously didn't run the right blocks.

coutinhop (approved):

PR #13115, opened 2026-06-29
Honor CALICO_IPV6POOL_VXLAN when the IPv4 pool is disabled

### LINKED ISSUE #12896: CALICO_IPV6POOL_VXLAN env var is silently ignored when CALICO_IPV4POOL_CIDR=none
## Bug Description
In IPv6-only clusters with VXLAN enabled, the `CALICO_IPV6POOL_VXLAN` environment variable is silently ignored, causing the IPv6 IPPool's `vxlanMode` to default to `Never` instead of the user-specified value (e.g., `Always`). This breaks cross-node Pod communication in IPv6-only VXLAN environments.

## Root Cause
In `node/pkg/lifecycle/startup/startup.go`, the `configureIPPools()` function reads `CALICO_IPV6POOL_VXLAN` **inside** the `if ipv4PoolEnabled {}` block:
// node/pkg/lifecycle/startup/startup.go - configureIPPools()
var (
    ipv4IpipModeEnvVar, ipv4VXLANModeEnvVar, ipv4BlockSizeEnvVar string
    ipv6VXLANModeEnvVar, ipv6BlockSizeEnvVar string               // declared here
    ...
)
if ipv4PoolEnabled {
    ipv4IpipModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV4POOL_IPIP"))
    ipv4VXLANModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV4POOL_VXLAN"))
    ipv6VXLANModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV6POOL_VXLAN"))  // [表情] WRONG: inside ipv4 block
    ipv4BlockSizeEnvVar = os.Getenv("CALICO_IPV4POOL_BLOCK_SIZE")
    ...
}
if ipv6PoolEnabled {
    ipv6BlockSizeEnvVar = os.Getenv("CALICO_IPV6POOL_BLOCK_SIZE")
    // [表情] CALICO_IPV6POOL_VXLAN is NOT read here
    ipv6NodeSelector = os.Getenv("CALICO_IPV6POOL_NODE_SELECTOR")
    ...
}

## Expected Behavior
When `CALICO_IPV6POOL_VXLAN=Always` is set in the calico-node DaemonSet for an IPv6-only cluster, the IPv6 IPPool should be created with `vxlanMode: Always`, enabling VXLAN encapsulation for cross-node Pod communication.

## Current Behavior
In an IPv6-only cluster where `CALICO_IPV4POOL_CIDR=none` is set (to disable the IPv4 IPPool), the `CALICO_IPV6POOL_VXLAN` environment variable is silently ignored. The IPv6 IPPool is created with `vxlanMode: Never`, breaking cross-node Pod communication over VXLAN.
Verification after calico-node starts:
calicoctl get ippool default-ipv6-ippool -o yaml
kind: IPPool
metadata:
  name: default-ipv6-ippool
spec:
  cidr: fd00:177:177::/112
  vxlanMode: Never          # ← Should be "Always", env var ignored!
  ipipMode: Never
  natOutgoing: true
Cross-node Pod communication fails because no VXLAN tunnel is established.
Possible Solution
In node/pkg/lifecycle/startup/startup.go, the configureIPPools() function reads CALICO_IPV6POOL_VXLAN inside the if ipv4PoolEnabled {} block. It should be moved to the if ipv6PoolEnabled {} block where it logically belongs.
Current code (buggy):
// node/pkg/lifecycle/startup/startup.go - configureIPPools()
var (
    ipv4IpipModeEnvVar, ipv4VXLANModeEnvVar, ipv4BlockSizeEnvVar string
    ipv6VXLANModeEnvVar, ipv6BlockSizeEnvVar string
    // ...
)
if ipv4PoolEnabled {
    ipv4IpipModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV4POOL_IPIP"))
    ipv4VXLANModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV4POOL_VXLAN"))
    ipv6VXLANModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV6POOL_VXLAN"))  // [表情] inside ipv4 block
    ipv4BlockSizeEnvVar = os.Getenv("CALICO_IPV4POOL_BLOCK_SIZE")
    // ...
}
if ipv6PoolEnabled {
    ipv6BlockSizeEnvVar = os.Getenv("CALICO_IPV6POOL_BLOCK_SIZE")
    // [表情] CALICO_IPV6POOL_VXLAN is NOT read here
    ipv6NodeSelector = os.Getenv("CALICO_IPV6POOL_NODE_SELECTOR")
    // ...
}
When CALICO_IPV4POOL_CIDR=none is set: ipv4PoolEnabled = false, the entire if ipv4PoolEnabled {} block is skipped, ipv6VXLANModeEnvVar remains empty string "".
Then the empty string is passed to createIPPool():

if ipv6PoolEnabled && !ipv6Present && ipv6Supported() {
    createIPPool(ctx, client, ipv6Cidr, DEFAULT_IPV6_POOL_NAME,
        string(api.IPIPModeNever), ipv6VXLANModeEnvVar, ...)  // empty string!
}
In createIPPool(), empty string maps to VXLANModeNever:

func createIPPool(..., vxlanModeName string, ...) {
    switch strings.ToLower(vxlanModeName) {
    case "", "off", "never":       // ← empty string hits this case
        vxlanMode = api.VXLANModeNever
    case "crosssubnet", "cross-subnet":
        vxlanMode = api.VXLANModeCrossSubnet
    case "always":
        vxlanMode = api.VXLANModeAlways
    }
}
Suggested fix — move the env var reading into if ipv6PoolEnabled {}:

if ipv6PoolEnabled {
    ipv6VXLANModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV6POOL_VXLAN"))  // [表情] moved here
    ipv6BlockSizeEnvVar = os.Getenv("CALICO_IPV6POOL_BLOCK_SIZE")
    ipv6NodeSelector = os.Getenv("CALICO_IPV6POOL_NODE_SELECTOR")
    // ...
}
And remove ipv6VXLANModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV6POOL_VXLAN")) from the if ipv4PoolEnabled {} block.
Steps to Reproduce (for bugs)
Set up a Kubernetes IPv6-only single-stack cluster (KUBE_DNS_IP is an IPv6 address)
Deploy Calico (tested with v3.29.4) with the following environment variables in the calico-node DaemonSet:

env:
  - name: IP
    value: "none"
  - name: IP6
    value: "autodetect"
  - name: CALICO_IPV4POOL_CIDR
    value: "none"
  - name: CALICO_IPV6POOL_CIDR
    value: "fd00:177:177::/112"
  - name: CALICO_IPV6POOL_VXLAN
    value: "Always"
  - name: FELIX_IPV6SUPPORT
    value: "true"
Set calico_backend: "vxlan" in the calico-config ConfigMap
Set CNI IPAM config in the ConfigMap:

"ipam": {
    "type": "calico-ipam",
    "assign_ipv4": "false",
    "assign_ipv6": "true"
}
Wait for calico-node to start, then check the IPv6 IPPool:

calicoctl get ippool default-ipv6-ippool -o yaml
Observe vxlanMode: Never instead of the expected vxlanMode: Always
Cross-node Pod communication is broken. Verify with:

 Pod on node A cannot reach Pod on node B
kubectl exec -it <pod-on-node-a> -- ping6 <pod-ip-on-node-b>
Workaround — after manually patching, communication is restored:

calicoctl patch ippool default-ipv6-ippool --patch '{"spec":{"vxlanMode": "Always"}}'
 Cross-node Pod communication now works
Note: This bug does NOT affect IPv4-only clusters. When CALICO_IPV4POOL_CIDR is a valid CIDR (not none), ipv4PoolEnabled = true, the if ipv4PoolEnabled {} block executes normally and CALICO_IPV6POOL_VXLAN is read correctly (though it's irrelevant in that case since IPv6 is disabled). The bug only manifests when IPv4 pool is explicitly disabled with CALICO_IPV4POOL_CIDR=none.

## Context
We are deploying Calico in an IPv6-only single-stack Kubernetes cluster with VXLAN encapsulation (single-subnet VXLAN mode). Setting CALICO_IPV4POOL_CIDR=none is the documented approach to prevent Calico from creating an unnecessary IPv4 IPPool in an IPv6-only environment. However, this causes the IPv6 VXLAN mode to be silently ignored, resulting in broken cross-node Pod connectivity. This took significant debugging effort to trace back to the source code, as there is no error or warning logged — the env var is simply never read.

## Your Environment
Calico Version: V3.29.4
Kubernetes Version: V1.32.6
Calico dataplane: iptables
Operating System and Version: TencentOS Server 3.1 (Final)
Calico Deployment Method: Calico was deployed using the standard YAML manifests.

caseydavenport:
In IPv6-only clusters that disable the IPv4 pool with `CALICO_IPV4POOL_CIDR=none`, `CALICO_IPV6POOL_VXLAN` was silently ignored and the IPv6 pool came up with `vxlanMode: Never`, breaking cross-node pod connectivity over VXLAN. The env var was being read inside the IPv4-pool branch of `configureIPPools`, which is skipped when the IPv4 pool is disabled. This moves the read into the IPv6-pool branch where it belongs.

Added a regression test that disables the IPv4 pool, sets `CALICO_IPV6POOL_VXLAN=Always`, and checks the resulting IPv6 pool. It fails without the fix.

Fixes #12896

**Release note:**
```release-note
Fixes CALICO_IPV6POOL_VXLAN being ignored in IPv6-only clusters where the IPv4 pool is disabled (CALICO_IPV4POOL_CIDR=none), which left the IPv6 pool without VXLAN encapsulation.
```

### DIFF
--- a/node/pkg/lifecycle/startup/startup_test.go
+++ b/node/pkg/lifecycle/startup/startup_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016,2021 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016,2021,2026 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -445,6 +445,30 @@ var _ = Describe("FV tests against a real etcd", func() {
 			[]EnvItem{{"CALICO_IPV4POOL_DISABLE_BGP_EXPORT", "false"}, {"CALICO_IPV6POOL_DISABLE_BGP_EXPORT", "false"}},
 			"192.168.0.0/16", randomULAPool, "Off", "Off", "Off", true, false, 26, 122, "all()", "all()", false, false),
 	)
+
+	It("should honor CALICO_IPV6POOL_VXLAN when the IPv4 pool is disabled", func() {
+		cfg, err := apiconfig.LoadClientConfigFromEnvironment()
+		Expect(err).NotTo(HaveOccurred())
+		c, err := client.New(*cfg)
+		Expect(err).NotTo(HaveOccurred())
+		be, err := backend.NewClient(*cfg)
+		Expect(err).NotTo(HaveOccurred())
+		err = be.Clean()
+		Expect(err).NotTo(HaveOccurred())
+
+		defer temporarilySetEnv("CALICO_IPV4POOL_CIDR", "none")()
+		defer temporarilySetEnv("CALICO_IPV6POOL_VXLAN", "Always")()
+
+		configureIPPools(ctx, c, kubeadmConfig)
+
+		poolList, err := c.IPPools().List(ctx, options.ListOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(poolList.Items).To(HaveLen(1))
+
+		pool := poolList.Items[0]
+		Expect(pool.Name).To(Equal(DEFAULT_IPV6_POOL_NAME))
+		Expect(pool.Spec.VXLANMode).To(Equal(apiv3.VXLANModeAlways))
+	})
 
 	It("should properly clear node IPs", func() {
 		cfg, err := apiconfig.LoadClientConfigFromEnvironment()
--- a/node/pkg/lifecycle/startup/startup.go
+++ b/node/pkg/lifecycle/startup/startup.go
@@ -971,7 +971,6 @@ func configureIPPools(ctx context.Context, client client.Interface, kubeadmConfi
 	if ipv4PoolEnabled {
 		ipv4IpipModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV4POOL_IPIP"))
 		ipv4VXLANModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV4POOL_VXLAN"))
-		ipv6VXLANModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV6POOL_VXLAN"))
 
 		ipv4BlockSizeEnvVar = os.Getenv("CALICO_IPV4POOL_BLOCK_SIZE")
 		if ipv4BlockSizeEnvVar != "" {
@@ -1005,6 +1004,8 @@ func configureIPPools(ctx context.Context, client client.Interface, kubeadmConfi
 	}
 
 	if ipv6PoolEnabled {
+		ipv6VXLANModeEnvVar = strings.ToLower(os.Getenv("CALICO_IPV6POOL_VXLAN"))
+
 		ipv6BlockSizeEnvVar = os.Getenv("CALICO_IPV6POOL_BLOCK_SIZE")
 		if ipv6BlockSizeEnvVar != "" {
 			ipv6BlockSize = parseBlockSizeEnvironment(ipv6BlockSizeEnvVar)


coutinhop (approved):

node/pkg/lifecycle/startup/startup_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016,2021 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016,2021,2026 Tigera, Inc. All rights reserved.
```
mazdakn, line 1: I think we use `2016-2026` format these days.

PR #13116, opened 2026-06-29
Don't tag a host IP as a workload when it falls within an IPAM block

### LINKED ISSUE #12847: Host IPs inside containing IPAM block get extra route programmed, breaking networking to the k8s node
## Expected Behavior

When a node's host IP `10.0.4.1` falls within an IPAM block `10.0.4.0/24`, networking to the node doesn't break.

## Current Behavior

In my homelab cluster running on `v3.30.2`, when I try to set up the IPPools such that the pods created on each node are in the `/24` range for that node's IP, I see an extra route programmed via `vxlan.calico` for the host IP:

```
10.0.4.0/32   via vxlan.calico                       metric 0      # VTEP/tunnel-endpoint route (expected)
10.0.4.1/32   via end0                               metric 1024   # operator-installed link-scoped peer route
10.0.4.0/24   via vxlan.calico  gateway 10.0.4.0     metric 0      # block CIDR route (expected)
10.0.4.1/32   via vxlan.calico  gateway 10.0.4.0     metric 0      # ← THIS ROUTE IS THE BUG
```

This extra route seems to break networking to that node, causing it to become unready in k8s and unreachable via talosctl.

## Possible Solution

This bug seems very similar to #12180 and could be fixed by adding another bool to `l3_route_resolver` `flush` to track if the host IP was in the trie and not apply the type flags to the route if it's a host IP. https://github.com/projectcalico/calico/blob/f70e559d5710a8a566d53208f14c477e9293c4ab/felix/calc/l3_route_resolver.go#L767-L777 

I'm happy to contribute a failing test for this bug modeled after the one in #12179 but I didn't want to do so before seeing if the current behavior is expected.

### Cluster setup

| Node     | InternalIP | Per-node IPPool       | blockSize |
|----------|-----------:|-----------------------|----------:|
| node-1   | 10.0.1.1   | 10.0.1.0/24           |        24 |
| node-2   | 10.0.2.1   | 10.0.2.0/24           |        24 |
| node-3   | 10.0.3.1   | 10.0.3.0/24           |        24 |
| node-4   | 10.0.4.1   | 10.0.4.0/24           |        24 |

- Encapsulation: VXLANCrossSubnet
- Each per-node IPPool has `nodeSelector` binding it to its respective node.
- The IPAMBlock for node-4's pool is `10.0.4.0/24` (one block per pool because `blockSize == pool prefix`).
- The block's `allocations[0]` is the VTEP (10.0.4.0). 10.0.4.1 is **unallocated** in IPAM and has no WorkloadEndpoint.

## Context

As shown in the reproduction section, I'm trying to set up a per-node IPPool so that each node and it's workloads have a dedicated `/24` to work with, making the pods easy to identify which node they're on from the IP.

## Your Environment

* Calico version `3.30.2`
* Calico dataplane (bpf, nftables, iptables, windows etc.)
* Orchestrator version (e.g. kubernetes, openshift, etc.): kubernetes `1.33.3`
* Operating System and version: `talos linux 1.12.7`

caseydavenport:
When a node's host IP falls inside one of its own Calico IPAM blocks (for example a per-node /24 pool where the node IP is the .1), Felix programs a route for the node's own IP via the VXLAN/IPIP tunnel device. That shadows the real link route and breaks connectivity to the node.

The route resolver was tagging the host IP as a remote workload because it sits inside the containing block. Same root cause as #12180 (tunnel IPs), so this extends the same guard to host IPs: a host IP that merely falls within a parent block no longer picks up the workload route type.

Note that overlapping the node network with a Calico IP pool isn't a supported configuration, but the fix is small and low risk.

Fixes #12847

```release-note
Fixes a loss of connectivity to a node when its host IP falls within a Calico IPAM block, in setups where the node network overlaps a Calico IP pool.
```

### DIFF
--- a/felix/calc/l3_route_resolver.go
+++ b/felix/calc/l3_route_resolver.go
@@ -846,17 +846,19 @@ func (c *L3RouteResolver) flush() {
 			}
 		}
 
-		// Apply the accumulated block workload type flags unless the route is a tunnel
-		// or host IP that merely falls within a parent block. Such an IP inside a larger
-		// block (e.g. a /32 inside a /24) should not get REMOTE_WORKLOAD — for a tunnel
-		// IP that makes isRemoteTunnelRoute() program a spurious /32 route, and for a host
-		// IP it makes the route manager program the node's own IP via the tunnel device,
-		// which breaks connectivity to the node.
+		// Apply the accumulated block workload type flags, with two exceptions.
 		//
-		// If the block is at the same CIDR as the route (e.g. a /32 block from a dedicated
-		// pool), the workload type is still needed so the route manager programs the
-		// correct directly-connected route.
-		if blockSeen && (blockMatchesRoute || (!hasTunnelRef && !hasHostRef)) {
+		// A host IP never inherits the workload type from a containing block, even one at
+		// its exact CIDR. It's reachable via its own host route; tagging it REMOTE_WORKLOAD
+		// makes the route manager program the node's own IP via the tunnel device, which
+		// breaks connectivity to the node.
+		//
+		// A tunnel IP that merely falls within a larger parent block (e.g. a /32 inside a
+		// /24) also skips the workload type, since REMOTE_WORKLOAD makes
+		// isRemoteTunnelRoute() program a spurious /32 route. A tunnel IP whose block is at
+		// the same CIDR (a dedicated /32 tunnel pool) still needs it, so the route manager
+		// programs the correct directly-connected route.
+		if blockSeen && !hasHostRef && (blockMatchesRoute || !hasTunnelRef) {
 			rt.Types |= blockTypes
 		}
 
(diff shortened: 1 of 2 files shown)


felix/calc/l3_route_resolver.go
```diff
@@ -845,14 +847,16 @@ func (c *L3RouteResolver) flush() {
...
+		// which breaks connectivity to the node.
 		//
-		// However, if the block is at the same CIDR as the route (e.g., a /32 block from
-		// a dedicated tunnel pool), the workload type is still needed so the route manager
-		// programs the correct directly-connected route.
-		if blockSeen && (!hasTunnelRef || blockMatchesRoute) {
+		// If the block is at the same CIDR as the route (e.g. a /32 block from a dedicated
+		// pool), the workload type is still needed so the route manager programs the
+		// correct directly-connected route.
+		if blockSeen && (blockMatchesRoute || (!hasTunnelRef && !hasHostRef)) {
```
fasaxc: Are we sure we need blockMatchesRoute to override a host IP?  A host IP should be reachable with no extra tunnel route

tomastigera (approved):
LGTM

Test nit: It only asserts the negative. Consider also asserting the positive/regression guard isn't broken — i.e. a genuine remote workload IP inside the same block still gets REMOTE_WORKLOAD. Without it, someone could "fix" a future issue by unconditionally suppressing block types and this test wouldn't catch it.

Issue #8934, opened 2024-06-21
Felix should configure iptables rules such that VXLAN UDP Flows are not tracked in conntrack, when in VXLAN mode


adkafka:
We have a workload that manages many (10,000s) TCP connections per node. Traffic is sent between nodes in a Kubernetes cluster. We use Calico in a VXLAN configuration as our CNI. Felix manages the `iptables` entries as expected. Additionally, we have `kube-proxy` running on our cluster in a standard configuration (using `iptables` not `ipvs`).

The issue we are noticing is that our Conntrack tables are unexpectedly full. Some of the entries in the Conntrack table are expected (the TCP connections responsible for our application traffic), but to my surprise, almost half of the entries in our Conntrack table are UDP "connections" responsible for the VxLAN tunnels between nodes. All of these connections are in the "UNREPLIED" state.

Here are some commands to illustrate this:

```
$ cat /proc/sys/net/nf_conntrack_max
131072
$ cat /proc/net/nf_conntrack | wc -l
129605
$ cat /proc/net/nf_conntrack | grep "udp" | wc -l
59262
$ cat /proc/net/nf_conntrack | grep "udp" | grep "UNREPLIED" | grep "dport=4789" | wc -l
59258
$ cat /proc/net/nf_conntrack | grep "tcp" | wc -l
70343
```

Of the 129,605 flows tracking in Conntrack, 59,258 (~46%) of them are UDP flows corresponding to VXLAN. This limits how many connections each node in our cluster can handle significantly. Luckily, when one of these nodes Conntrack table *does* fill up, after dropping a couple packets, it will enter "early_drop" mode, and remove many `UNREPLIED` connections from the Conntrack table (which in our case, are the UDP VXLAN flows). This prevents having significant application impact, but it does make monitoring our Conntrack usage much more difficult.

After some discussion in the Calico slack (#networking https://calicousers.slack.com/archives/CPEPF833L/p1718663125404499), we decided to experiment with adding `iptables` rules such that these VXLAN UDP flows were **not** tracked in Conntrack. We found that it had the desired effect and caused no impact to our application traffic. Therefore, we are proposing that Calico automatically add these rules when in VXLAN mode. It may be worth putting this behind a configuration flag and defaulting to "off" to prevent accidentally breaking any workloads.

The `iptables` rules I added to each of these node to configure it **not** to track VXLAN UDP traffic was:
```
$ iptables --table raw --append OUTPUT --protocol udp --dport 4789 --jump NOTRACK
[...]

PR #13118, opened 2026-06-29
Avoid masquerading tunnel traffic onto the VXLAN port

caseydavenport:
The static NAT postrouting chain masquerades host-sourced traffic that egresses a tunnel interface. The masquerade rule had no `--to-ports`, so the kernel picked the source port from its default SNAT range (1024-65535). That range includes the VXLAN port (default 4789): the kernel prefers to keep a flow's original port but remaps into the range on collision, and nothing stopped it from landing on 4789. When it did, the reply, whose dest port is then 4789, got dropped by the "Drop VXLAN packets from non-allowed hosts" filter rule. That rule matches any UDP packet to the VXLAN port regardless of ingress interface, so it bites replies to perfectly normal flows (e.g. a node sending UDP to a CoreDNS pod across the VXLAN tunnel).

Conntrack would normally hide this, since the reply belongs to an established flow and un-NAT would restore the original dest port before the filter rules run. But we set NOTRACK on all UDP to the VXLAN port (#8934, since VXLAN uses different source ports in each direction and tracking left flows unreplied), so there's no conntrack entry to un-NAT the reply or accept it as established, and it reaches the drop rule with dest port 4789 intact. That's why this only shows up on 3.29+; older versions tracked the flow and never saw the collision.

The fix constrains those masquerade rules to a `--to-ports` range that excludes the VXLAN port. iptables/nftables only allow a single contiguous range, so we take whichever side of the port is wider (`4790-65535` for the default). Only applied when VXLAN is enabled, since the drop rule only exists then.

```release-note
Fix VXLAN traffic being dropped when the kernel masqueraded a tunnel flow's source port onto the VXLAN port (default 4789).
```

Fixes #12244

### DIFF
--- a/felix/fv/vxlan_test.go
+++ b/felix/fv/vxlan_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2025 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020-2026 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@ package fv_test
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


hengqiali:
Hi  @caseydavenport , thanks for your MR, do you think we will backport this MR to like 3.31 old version, or just let it as a new feature in verion >= 3.33.0? Thanks

caseydavenport:
@hengqiali I think this is probably OK to backport to v3.32 / v3.31.

tomastigera (commented):
Add an FV tests that makes sure the port limiting works?

matthewdupre (approved):
I'm good with this subject to suitable testing (i.e. make sure Tomas's comment is satisfied before merging).

The magic port range is a bit spooky, but we can always just move this to a config option if we need to or split between the two halves of the range with weighted rules in future.

hengqiali:
Hi @caseydavenport @danudey , I'm testing it and it works now. Kindly ask that if this bugfix will be released into 3.31.7? Do we have an approximate time for 3.31.7 release? Because our production calico version is blocked at 3.28 and cannot upgrade to new version because this bug. Thank you

hengqiali:
> @hengqiali I think this is probably OK to backport to v3.32 / v3.31.

Hi @caseydavenport @danudey I still do not see the new released 3.31.7 version containing this bugfix, we have waited this bugfix for a long time, I'm not sure why calico open source community is so not friendly for users.  Especially there have been a known bug and many communication with this thing..

matthewdupre:
@hengqiali You're welcome to create backport PRs of this if you'd like to.  I think people often don't look at comments on closed PRs.

hengqiali:
@matthewdupre I would like to know if incoming backport PR can be merged into 3.31.7 version or not.

-> BTW, when I comment the bugfix, the PR is not closed.
image

matthewdupre:
@hengqiali 3.31.7 came out yesterday: https://github.com/projectcalico/calico/releases/tag/v3.31.7, so that release is closed.  I'm not sure if there'll be another 3.31 patch (.8) - 3.31 is getting towards its end of regular patch releases.

As Casey said before, if there are backport PRs we'll review them as normal - I'm guessing this should apply cleanly so it should be smooth.

hengqiali:
Hi @matthewdupre thanks for your response, how about I create MR for v3.32.1, so that v3.32.2 could carry this backport, do you know that when the v3.32.2 will be released? From my inspection, calico release each tag every 2 months, so I guess v3.32.2 is the most recent release. To achieve this, I just need to cherry pick the patch to release-v3.32 branch, then it's done right? Highly appreciated.

matthewdupre:
@hengqiali Yeah, if you raise a cherry pick PR and @ me for review that's enough.  3.32.2 is aiming for this week so we might or might not make that.  If not 3.32.3 is aiming for October as you observed.

hengqiali:
Cherry-pick onto `release-v3.32` for v3.32.2: https://github.com/projectcalico/calico/pull/13628

@matthewdupre as discussed — please review when you have a chance.

PR #13122, opened 2026-06-30
watchersyncer: add NewMultiClient + ResourceType.ClientID

pasanw:
## Description

Port the multi-client constructor and per-resource `ClientID` field from Enterprise into OSS. Pure infrastructure: the mechanism allows a single watcherSyncer to draw events from multiple backend clients, with each `ResourceType` naming which client it uses via the new `ClientID` field.

The existing single-client API is preserved by making `New()` a thin wrapper that calls `NewMultiClient(map[string]api.Client{"": client}, ...)`. `ResourceType` values with no explicit `ClientID` get the empty-string default, which matches the map key. Existing callers are unaffected.

No OSS consumer currently uses the multi-client form. Bringing the plumbing into OSS removes the structural divergence between the trees.


fasaxc (changes requested):
Looks like some good comments from Copilot

fasaxc (approved):

PR #13123, opened 2026-06-30
node k8st: use Gomega assertions and Eventually for retries

mazdakn:
Test-only refactor of the Go k8st system tests (`node/tests/k8st/`) to standardize on Gomega, matching the pattern the BGP advertisement tests already established.

**What changed**
- Replaced the remaining raw `t.Fatalf`/`t.Errorf` assertions with Gomega `Expect`/`Eventually`.
- Replaced the bespoke `utils.RetryUntilSuccess` retry helper with Gomega's `Eventually` (1s polling) at every call site, and removed the now-unused helper along with the dead `ErrTimeout` var.
- Dot-imported Gomega in all k8st files so matchers are used unqualified.

One deliberate exception: `bgp_advert_test.go`'s `teardownTest` keeps non-fatal cleanup semantics by using a `NewGomega` handler that reports via `t.Errorf` rather than failing the test (a fatal there would skip sibling cleanups).

**Testing**
- `go vet ./node/tests/k8st/...` passes.
- `gofmt -l` clean.
- No production code or behaviour changes; these are system tests that run against a live kind cluster.

**Release note:**
```release-note
NONE
```

### DIFF
--- a/node/tests/k8st/utils/utils.go
+++ b/node/tests/k8st/utils/utils.go
@@ -20,7 +20,6 @@ package utils
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -29,6 +28,7 @@ import (
 	"testing"
 	"time"
 
+	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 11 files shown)


caseydavenport (approved):

node/tests/k8st/utils/utils.go
```diff
@@ -162,59 +163,6 @@ func mergeRunOptions(opts []RunOptions) RunOptions {
...
-// ----------------------------------------------------------------------------
-// Retry.
-
-// RetryUntilSuccess invokes fn until it returns nil or the timeout
-// elapses. It uses exponential backoff starting at 0.5s and capped at 10s.
-// The time taken by fn counts toward the wall-clock deadline
-// so the overall budget is predictable.
-//
-// Returns the last error from fn on timeout, or nil on success.
-func RetryUntilSuccess(t testing.TB, timeout time.Duration, fn func() error) error {
```
matthewdupre, line 174: @mazdakn is there a particular reason this was removed in favour of Eventually()?  There are a few things about Eventually that are not good that this tries to address.

PR #155, opened 2016-11-04
Combine FAQ pages, add TOC

caseydavenport:
Fixes #34

PR #13124, opened 2026-07-01
argoci: OSS Calico e2e scaffold + generated nftables cron (.argoci/)

lwr20:
## Description

Adds the `.argoci/` directory ArgoCI runs the OSS Calico e2e suites from, as part of moving OSS e2e off Semaphore's scheduled builds onto ArgoCI (Argo Workflows). Scoped to **only `.argoci/`** — the converter tooling that generates/verifies these lives in the `semaphore-to-argoci` skill (tigera/claude-plugin-marketplace#117), not in this repo.

- `.argoci/scripts/` — e2e lifecycle ported from `.semaphore/end-to-end/scripts/`, adapted for ArgoCI.
- `.argoci/cron/e2e-nftables-master.yaml` — first generated **condensed** ArgoCI workflow (`cc-argoci-handler` expands it). nftables is the proving case, verified job-for-job against its Semaphore source by the parity harness.
- `.argoci/README.md` — what the dir is and how crons are (re)generated.

> Note: squashed and narrowed to `.argoci/` after the converter moved to the skill; the DESIGN.md earlier reviewers commented on now lives in the skill PR.

**Release note:**
```release-note
NONE
```


.argoci/DESIGN.md
```diff
@@ -0,0 +1,495 @@
...
+
+---
+
+## Goal
+
+Migrate **all** OSS Calico *scheduled* e2e pipelines from Semaphore CI to
+ArgoCI (Argo Workflows on `argoci.dev.calicocloud.io`), across **all active
+branches** (`master`, `release-v3.32`, `release-v3.31`), **starting with
+`nftables.yml`** as the proving case. This is **time-critical: Semaphore is
+being retired soon**, so the deliverable is a **repeatable Go converter +
```
nelljerram: I guess you mean that Tigera is going to stop using Semaphore, right?  As opposed to Semaphore themselves are shutting down.  Please clarify wording accordingly.

.argoci/DESIGN.md
```diff
@@ -0,0 +1,495 @@
...
+  history limits, `workflowMetadata.labels`.
+- Each template: `source checkout` → `source global_prologue.sh` → `<body>`
+  → `source global_epilogue.sh` → `exit $CI_EXIT_CODE`; `onExit: exit-handler`
+  guarantees cleanup. A `dind` sidecar provides Docker.
+
+## Approach
+
+### Central insight
+
+The **lifecycle is constant** across pipelines and streams. Only the
```
nelljerram: I don't know what this means.

.argoci/DESIGN.md
```diff
@@ -0,0 +1,495 @@
...
+**block/job/matrix/env tree** and the **selection mechanism** differ. So:
+
+1. **Foundational, authored once by hand**: shared lifecycle in `.argoci/`.
+2. **Per-(pipeline × branch), automated** by the Go converter: read
+   `.semaphore/end-to-end/pipelines/<name>.yml` from a given branch → emit
+   `.argoci/cron/e2e-<name>-<stream>.yaml`.
+
+### Runtime/branch strategy (DECIDED: per-branch checkout)
+
+**Each cron checks out its own target branch** (`branch: master` |
```
nelljerram: Why is cron in the picture?  I expect CI to run when a PR is pushed or changes, and when a branch changes.  I can see a role for cron for our scheduled daily builds, but otherwise not.

nelljerram (commented):
This feels like way too many words.  I haven't made it more than 30% through, but I guess it's because it's a mix of stuff that is blindingly obvious, with a few spots of stuff that is genuinely subtle and needs to be carefully reviewed.  Unfortunately that mix means that it's very likely the subtle parts will be missed.

Is there a way you can re-prompt AI to present this much more concisely and to call out the parts that really need review?

.argoci/DESIGN.md
lwr20: i.e. the structure of all e2es is identical, across release streams and every pipeline file.

.argoci/DESIGN.md
lwr20: This is not about CI, this is about migrating the e2es only.

.argoci/DESIGN.md
```diff
@@ -0,0 +1,495 @@
...
+  dual-mode `K8S_E2E_FLAGS` path then becomes vestigial.
+- **Decommissioning Semaphore itself** (the act of turning it off) — this
+  work makes ArgoCI ready; cutover is the trigger.
+
+## Glossary & assumptions (read first)
+
+- **`bz`** — the "banzai" CLI that provisions/installs/destroys test clusters
+  (`bz provision|install|destroy|tests`). **Pre-installed in the ArgoCI base
+  image**; not installed in the prologue. `PROVISIONER=local-kind|gcp-kubeadm|
+  aws-eks|aws-talos|…` selects the backend; provider specifics live in `bz`.
```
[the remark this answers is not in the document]
lwr20: Fixed in 51e51dd — `PROVISIONER` and its values are now separate single-line code spans.

.argoci/DESIGN.md
```diff
@@ -0,0 +1,495 @@
...
+  workflow finishes (success/failure/termination). Our cleanup safety net.
+- **dind** — a `docker:*-dind` sidecar (`DOCKER_HOST=tcp://127.0.0.1:2375`).
+  Required: `run_tests.sh` runs the e2e binary via `docker run`, and
+  `local-kind` runs the cluster in-pod via Docker.
+- **skill** — a Claude Code skill: the human-in-the-loop wrapper that drives
+  the Go converter and hand-tunes irregular cases. Uses the repo's
+  `e2e-tests` skill for config-file authoring.
+- **Secrets** (`envFrom.secretRef`, must exist in `argoci` ns):
+  `banzai-secrets` (GITHUB_ACCESS_TOKEN, cloud creds), `marvin-github-ssh-
+  private-key`. Materialized to files in the prologue via `createLocalSecret`.
```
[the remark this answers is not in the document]
lwr20: Fixed in 51e51dd — the `marvin-github-ssh-private-key` code span is on one line now.

.argoci/DESIGN.md
```diff
@@ -0,0 +1,495 @@
...
+
+The definition of a **faithful** conversion: the generated ArgoCI crons
+enumerate the **same set of test jobs, with the same resolved parameters**, as
+the Semaphore pipelines. We prove it by enumerating both sides into a common
+canonical form and diffing — the converter's CI acceptance gate *and* the
+pre-cutover audit before any Semaphore pipeline is switched off.
+
+**Prior art, already in-repo:** `.semaphore/end-to-end/report/
+generate_e2e_report.py` (mirrored from `tigera/banzai-calico/report`) already
+enumerates Semaphore jobs. It merges `default_env_vars` (the
```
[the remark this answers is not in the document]
lwr20: Fixed in 51e51dd — the report path is now a single-line code span.

.argoci/DESIGN.md
```diff
@@ -0,0 +1,495 @@
...
+metadata: {name: e2e-nftables-v3.31, namespace: argoci, labels: {repo: calico, branch: release-v3.31}}
+spec:
+  schedules: ["0 3 * * 2"]; timezone: UTC; concurrencyPolicy: Forbid
+  failedJobsHistoryLimit: 3; successfulJobsHistoryLimit: 3
+  workflowMetadata: {generateName: e2e-nftables-v3.31-, labels: {repo: calico, type: Nightly, branch: release-v3.31, dashboard: calico-oss}}
+  workflowSpec:
+    entrypoint: pipeline; onExit: exit-handler
+    # branch = the target stream's branch; checkout + RELEASE_STREAM derive from it.
+    arguments: {parameters: [{name: reponame, value: calico}, {name: repoURL, value: git@github.com:projectcalico/calico.git}, {name: branch, value: release-v3.31}]}
+    nodeSelector: {role: argoci}; tolerations: [{key: argoci, operator: Exists}]
```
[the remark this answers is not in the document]
lwr20: Fixed in 51e51dd — replaced the semicolon-compressed (invalid-YAML) skeleton with a pointer to the committed `.argoci/cron/e2e-nftables-master.yaml`, which is real, valid YAML.

.argoci/DESIGN.md
lwr20: Clarified: this is an internal goal to stop using Semaphore for OSS e2e by end of July 2026 — Semaphore the platform is not going away. Wording updated.

.argoci/DESIGN.md
lwr20: Reworded per your note: every e2e job has the identical lifecycle/structure across all release streams and every pipeline file; only the block/job/matrix/env tree and the test-selection mechanism differ.

lwr20:
Thanks — good push. Condensed the doc from ~530 to ~137 lines (c6ef4239) and led with a "Decisions that need careful review" section calling out the three genuinely non-obvious calls: per-branch checkout, dual-mode test selection (forced by E2E_TEST_CONFIG being master-only), and teardown-on-cancel. Settled findings and the standard ArgoCI onboarding steps are moved out of the review path into a Findings section and an Ops-dependencies checklist, so the subtle parts stand out. PTAL.

caseydavenport:
What's @cc-github-bot doing here?

hack/cmd/argo-jobreport/cron.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// cronDoc is the subset of a CronWorkflow this tool reads.
+type cronDoc struct {
```
caseydavenport: This is a bit of a nightmare to read 😬

hack/cmd/semaphore2argo/main.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/projectcalico/calico/hack/argoci/convert"
+)
+
+func main() {
+	in := flag.String("in", "", "path to a Semaphore end-to-end pipeline YAML (required)")
```
caseydavenport: We're already importing Cobra into this repo, so might be worth using that for new Go CLI tools as well?

I personally find it a lot easier to use than manual flag handling, but YMMV

.argoci/cron/e2e-nftables-master.yaml
```diff
@@ -0,0 +1,189 @@
+# Generated by hack/cmd/semaphore2argo from a Semaphore e2e pipeline. DO NOT EDIT.
```
Glen-Tigera: Stale references after the tooling move: `hack/cmd/semaphore2argo` no longer exists in this repo (the converter now lives in the `semaphore-to-argoci` skill), and `.argoci/DESIGN.md` (line 2) moved to the skill too. Suggest updating the header to point at where the generator actually lives so anyone reading a generated file can find it.

.argoci/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Do not hand-edit `cron/*.yaml`
+
+They are generated from `.semaphore/end-to-end/pipelines/*.yml` by the
+`semaphore-to-argoci` skill and verified job-for-job against the Semaphore
+source by its parity harness. To change a pipeline, edit the Semaphore source
+(or the scripts here) and regenerate:
+
+```
+semaphore2argo --in .semaphore/end-to-end/pipelines/<name>.yml \
```
Glen-Tigera: `semaphore2argo` is only in the marketplace `semaphore-to-argoci` skill, not on an OSS contributor's PATH — so the regeneration instructions here aren't runnable by a projectcalico/calico contributor. Worth noting explicitly that the generator lives in tigera-de-tools (and is Tigera-internal).

.argoci/scripts/phases/run_tests.sh
```diff
@@ -0,0 +1,118 @@
...
+  echo "[INFO] hashrelease URL: ${HASHREL_URL}"
+  ARCH=$(uname -m); [[ "$ARCH" == "x86_64" ]] && ARCH=amd64; [[ "$ARCH" == "aarch64" ]] && ARCH=arm64
+  mkdir -p "${CI_HOME}/${CI_GIT_DIR}/e2e/bin/k8s"
+  curl --retry 9 --retry-all-errors -fsSL "${HASHREL_URL}/files/e2e/e2e-linux-${ARCH}.test" -o "${CI_HOME}/${CI_GIT_DIR}/e2e/bin/k8s/e2e.test"
+  chmod +x "${CI_HOME}/${CI_GIT_DIR}/e2e/bin/k8s/e2e.test"
+  echo "[INFO] downloaded e2e binary to ${CI_HOME}/${CI_GIT_DIR}/e2e/bin/k8s/e2e.test"
+  E2E_BINARY=/go/src/github.com/projectcalico/calico/e2e/bin/k8s/e2e.test
+fi
+
+if [[ -n "${E2E_BINARY:-}" && -n "${E2E_TEST_CONFIG:-}" ]]; then
```
Glen-Tigera, line 54: `E2E_BINARY` is assigned a path in both branches above (lines 35/47) but only its emptiness is tested here — it's really a boolean sentinel and isn't passed into `make e2e-run`. Reads a bit misleadingly; a comment (or a plain boolean flag) would make the intent clearer.

.argoci/scripts/global_epilogue.sh
```diff
@@ -0,0 +1,44 @@
...
+  curl --retry 3 -fsSL -H "Authorization: token ${GITHUB_ACCESS_TOKEN}" \
+    -H "Accept: application/vnd.github.v3.raw" \
+    -o /tmp/run-lens.sh \
+    https://raw.githubusercontent.com/tigera/banzai-lens/main/uploader/run-lens.sh && \
+    chmod +x /tmp/run-lens.sh && /tmp/run-lens.sh || true
+fi
+
+# Tear the cluster down.
+echo "[INFO] destroying cluster ${CLUSTER_NAME}"
+bz destroy |& tee "${BZ_LOGS_DIR}/destroy.log" || true
```
Glen-Tigera, line 42: `bz` (and `gsutil`) are relied on here, but `bz` is installed to `${HOME}/.local/bin` and added to PATH in the prologue. This is only safe if the handler runs prologue + body + epilogue in the same shell process (the epilogue runs at EXIT). Worth a one-line confirmation against the cc-argoci-handler contract that PATH/exports survive into the exit handler.

Glen-Tigera (commented):
Reviewed the `.argoci/` scaffold together with the converter skill (tigera/claude-plugin-marketplace#117). Solid, well-commented work and the generated-artifacts-here / generator-in-the-skill split is the right boundary. A few things worth resolving before cutover — inline comments below, plus these PR-level points:

- **ArgoCI preprocessor error.** `cc-github-bot` reports `Failed to get file .argoci/ciworkflow.yaml … 404` on each commit (this is what @caseydavenport spotted). Introducing `.argoci/` seems to activate a preprocessor that then fails on a file this PR doesn't add — worth confirming with ops whether `cron/*` alone is a valid state, otherwise the error may start firing on unrelated PRs after merge.
- **`ci/semaphoreci/pr: Calico` is currently failing** — worth confirming it's unrelated before merge.
- **Orphaned-cluster janitor.** The design notes the janitor cron is "to build". An `activeDeadlineSeconds`-killed pod leaks a cloud cluster (cost + quota), so I'd treat the janitor as a prerequisite for enabling anything beyond the nftables proving pipeline rather than a follow-up.
- **Per-branch cherry-pick (design).** The rejected alternative (scripts on master only + `RELEASE_STREAM` param) looks stronger than the DESIGN credits — dual-mode already absorbs the main master-vs-release difference, so `scripts/` are largely branch-invariant and only `cron/*` genuinely needs to be per-branch. Might be worth re-litigating to avoid 3-way drift.

.argoci/cron/e2e-nftables-master.yaml
```diff
@@ -0,0 +1,189 @@
...
+# Condensed ArgoCI workflow; cc-argoci-handler expands it. See .argoci/DESIGN.md.
+version: condensed
+generateName: e2e-nftables-master-
+schedule: 0 3 * * 2
+globalPrologue: |
+    cd "${CI_HOME}/${CI_GIT_DIR}"
+    export DATAPLANE="${DATAPLANE:-CalicoNftables}"
+    export ENABLE_AUTOHEP="${ENABLE_AUTOHEP:-true}"
+    export FUNCTIONAL_AREA="${FUNCTIONAL_AREA:-nftables.yml}"
+    export INSTALLER="${INSTALLER:-operator}"
```
Glen-Tigera: I think these could be defined in the `env` section or put into global_prologue.sh instead if we want them to be overridable

lwr20:
Thanks @Glen-Tigera — addressed in the latest commits:

- **Stale header refs** (`hack/cmd/semaphore2argo`, `.argoci/DESIGN.md`) — fixed in the generator; regenerated crons now point at the `semaphore-to-argoci` skill.
- **README regeneration instructions** — added a note that `semaphore2argo` ships in the Tigera-internal `semaphore-to-argoci` skill (not on an OSS contributor's PATH); external contributors edit the Semaphore source/scripts and ask a Tigera maintainer to regenerate.
- **`E2E_BINARY` sentinel** — added a comment clarifying it's a set/unset sentinel (value unused; `make e2e-run` locates the binary), and that regex (`K8S_E2E_FLAGS`) / non-e2e jobs fall through to `bz tests`.
- **`bz`/PATH survival into the exit handler** — confirmed. `cc-argoci-handler` wraps our `globalEpilogue` as an `EXIT` trap **inside the same `bash -e` step script** (not a separate container), so the prologue's `PATH`/exports survive into it. Empirically verified in a real ArgoCI run: the epilogue's `bz destroy` succeeded and tore down all 18 cluster resources.
- **Overridable env** (`DATAPLANE` etc.) — they already are: emitted as `export VAR="${VAR:-default}"`, and a step's own `env` is a real container var set *before* the prologue runs, so it wins over the default. So overriding via the cron step `env` (or the Semaphore source) works as-is. Happy to move them to `env` if you'd prefer that surface.

Also (not a review comment, but related): the nftables cron now uses the real Semaphore schedule (`10 21 * * 1`, from the `calico-monorepo-e2es` task) instead of the earlier placeholder.

.argoci/cron/e2e-nftables-master.yaml
lwr20: Fixed. Regenerated with the updated emitter — the header now points at the `semaphore-to-argoci` skill (no more `hack/cmd/semaphore2argo` or `.argoci/DESIGN.md`).

.argoci/README.md
lwr20: Fixed. Added a note in the README that `semaphore2argo` ships in the Tigera-internal `semaphore-to-argoci` skill (not on an OSS contributor's PATH) — external contributors edit the Semaphore source/scripts and ask a Tigera maintainer to regenerate.

.argoci/scripts/phases/run_tests.sh
lwr20: Fixed. Added a comment: `E2E_BINARY` is a set/unset sentinel (its value isn't used — `make e2e-run` locates the binary itself); the structured path runs only when a k8s-e2e binary was acquired AND `E2E_TEST_CONFIG` selects specs, otherwise regex (`K8S_E2E_FLAGS`) / non-e2e jobs fall through to `bz tests`.

.argoci/scripts/global_epilogue.sh
lwr20: Confirmed. `cc-argoci-handler` wraps our `globalEpilogue` as an `EXIT` trap inside the *same* `bash -e` step script (not a separate container), so the prologue's PATH/exports survive into it. Empirically verified in a real ArgoCI run: the epilogue's `bz destroy` succeeded and tore down all 18 cluster resources.

.argoci/cron/e2e-nftables-master.yaml
lwr20: They already are overridable: emitted as `export VAR="${VAR:-default}"`, and a step's own `env` is a real container var set *before* the prologue runs, so it wins over the default. So overriding via the cron step `env` (or the Semaphore source) works as-is. Left in `globalPrologue` as `${VAR:-default}`; happy to move them to the `env` section if you'd prefer that surface.

.argoci/README.md
lwr20: Revised fix: rather than annotate it, I removed the regeneration section (and the converter mention) from the README entirely. `semaphore2argo` is a one-off migration tool — the skill exists because there are many pipelines to move — so once migrated these are just the ArgoCI e2e workflows and the in-repo README no longer references the converter or a regeneration workflow.

.argoci/cron/e2e-nftables-master.yaml
lwr20: Updated further: since these crons are hand-maintained going forward (not regenerated), the header no longer says "Generated by … DO NOT EDIT" or references semaphore2argo at all. It's now just: `# Condensed ArgoCI e2e workflow; cc-argoci-handler expands it into a CronWorkflow at runtime.` + `# Migrated from the Semaphore e2e pipeline of the same name — maintained here by hand.`

.argoci/cron/e2e-nftables.yaml
```diff
@@ -0,0 +1,189 @@
+# Condensed ArgoCI e2e workflow; cc-argoci-handler expands it into a CronWorkflow at runtime.
```
Glen-Tigera, line 1: **This file fails `make yaml-lint`** — the cause of the red `ci/semaphoreci/pr: Calico` check (block "Check YAML files", job `d1b31bf9`). ~20 errors like `wrong indentation: expected 10 but found 8` and `expected 14 but found 12`.

Calico's `.yamllint.yaml` sets:
```yaml
indentation:
  spaces: consistent
  indent-sequences: consistent
```
The generated cron mixes indentation levels — top-level sequences (`defaults.secretBundles`, `steps`) are indented 4 spaces, but sequences nested inside a step (`services:`, `env:`, matrix `env:`) are indented only 2 — so `consistent` mode rejects it.

It comes from the converter's `render.go` using `yaml.Marshal`, whose yaml.v3 default (indent 4) produces this mixed nesting. Fix is in the emitter (see tigera/claude-plugin-marketplace#117 `convert/render.go`): emit with `yaml.NewEncoder` + `enc.SetIndent(2)` for a uniform +2 step, then regenerate this file. Verify locally with what CI runs:
```
docker run --rm -v $PWD:/data cytopia/yamllint:latest .
```
Only this file is affected (scripts are `.sh`, README is `.md`), and the indentation change is cosmetic to the YAML model — parsing/parity are unaffected.

Glen-Tigera (commented):
Dug into the red `ci/semaphoreci/pr: Calico` check — it's a **real, deterministic failure introduced by this PR**, not a flake. The failing block is **"Check YAML files"** → `make yaml-lint` (`cytopia/yamllint`), and it's rejecting this one generated file. Root cause + fix in the inline comment below. `master`'s push pipeline is green, so this is specifically the new `.argoci/` YAML.

Glen-Tigera (approved):
Approving. 👍 Thanks for working through the review — every finding is addressed: all-10-pipeline parity, Cobra + cron.go readability, the honest round-trip-vs-Semaphore-faithfulness framing, the empirically-verified teardown (18 resources torn down in a real ArgoCI run), and the cleaner hand-maintained-crons model.

One fix to land before merge, which I'm trusting you to resolve: the red `ci/semaphoreci/pr: Calico` is the `make yaml-lint` failure on `.argoci/cron/e2e-nftables-master.yaml` (root cause + the `render.go` `SetIndent(2)` fix are in my inline comments / tigera/claude-plugin-marketplace#117). Please get that green before merging.

Non-blocking, already scoped in the DESIGN for cutover (not this scaffold PR): the orphaned-cluster janitor cron, and a final ops confirmation that `.argoci/cron/*` alone is a valid ArgoCI state (the preprocessor errors were all on rebased-away commits).

.argoci/cron/e2e-nftables.yaml
lwr20: Fixed. Root cause exactly as you diagnosed — `render.go` used `yaml.Marshal` (yaml.v3 default indent 4), which nests sequences inconsistently vs top-level ones. Switched the emitter to `yaml.NewEncoder` + `enc.SetIndent(2)` (tigera/claude-plugin-marketplace#117, `8c04553`) for a uniform +2 step, and regenerated this cron (`ed0280cf`) plus the other 9 in #13133. Verified locally against the same config CI uses: `yamllint -c .yamllint.yaml .argoci/cron/*.yaml` → rc=0. Cosmetic to the YAML model, so parity is unaffected (`TestParityAllPipelines` still green). Thanks for the precise repro + the `SetIndent(2)` pointer.

.argoci/DESIGN.md
lwr20: I've moved the migration tool, including this design into our skills repo.  It doesn't belong in the monorepo because its a one-off migration.

lwr20:
🚧 **Hold for merge order.** These crons are now branchless (`e2e-<suite>.yaml`, `generateName: e2e-<suite>-`); the branch is appended to the live CronWorkflow name by the handler. That requires **cc-argoci-handler v0.12** (tigera/cc-utils#155) to be **deployed first** — merging before then would make the old handler name master's crons without a branch qualifier. Do not merge until #155 is merged, tagged v0.12, and rolled out.

lwr20:
@caseydavenport `@cc-github-bot` is the **ArgoCI pre-processor** (`cc-argoci-handler`) — it reacts to this repo's webhook events and turns `.argoci/` config into Argo workflows on our internal ArgoCI cluster. It's roughly the ArgoCI equivalent of Semaphore reading `.semaphore/`.

The `Failed to get file .argoci/ciworkflow.yaml` message came from an **earlier commit** of this PR that included a `.argoci/config.yaml` declaring a *per-PR* CI workflow (`ciFile: ciworkflow.yaml`, triggered on `.argoci/*` / `e2e/config/*` changes). OSS Calico has **no per-PR ArgoCI workflow** — the e2e suites are all **scheduled crons** (`.argoci/cron/*.yaml`) — so there was no `ciworkflow.yaml` for the pre-processor to fetch, and it logged the miss.

It's benign: it doesn't block this PR, gate the merge, or touch anything on Semaphore.

I've since **removed that `config.yaml`** from the PR (the current head has no `.argoci/config.yaml`), so the repo no longer advertises a per-PR CI workflow and the pre-processor won't attempt per-PR CI. The scheduled suites are registered from the cron files separately, not per-PR. If we ever want a per-repo ArgoCI config, it'll be a schedule-only one with no `ciFile`, so this won't recur.

hack/cmd/argo-jobreport/cron.go
lwr20: Fair 😅 — and this file is no longer in the PR. The generator **and** the parity harness (`argo-jobreport`, `semaphore2argo`) have moved out of the calico repo into the converter skill (`tigera/claude-plugin-marketplace`, under `tigera-de-tools/skills/semaphore-to-argoci`). What lands in calico now is just the hand-maintained `.argoci/` YAML + scripts — no generator Go in-tree.

So this `cron.go` (the cron-enumerator behind the parity diff) lives in the skill now, not here. The readability point still stands for that repo — I'll take a pass at tidying it there.

hack/cmd/semaphore2argo/main.go
lwr20: Good call — agreed Cobra reads better than manual `flag`. These CLIs (`semaphore2argo`, `argo-jobreport`) have also moved out of calico into the converter skill (`tigera/claude-plugin-marketplace`), so they're no longer part of this PR. They still use `flag` over there, so the suggestion carries over to that repo — I'll switch them to Cobra in the skill for consistency.

PR #13126, opened 2026-07-01
felix/fv: drain neighbours before device teardown to avoid kernel panic

fasaxc:
Felix FV jobs intermittently **panic and reboot their GCE test VM during teardown**, flaking the suite. The watchdog-captured serial console (job `f347a9ad`, kernel `6.17.0-1020-gcp`, Ubuntu 24.04 BPF-iptables FV) shows a kernel **use-after-free**:

```
BUG: unable to handle page fault for address: fffffffffffffff0
RIP: netdev_master_upper_dev_get_rcu+0x2d          ← walks a freed netdevice's adj_list.upper
 l3mdev_master_ifindex_rcu / icmp_route_lookup / __icmp_send
 ipv4_send_dest_unreach / ipv4_link_failure / arp_error_report
 neigh_invalidate / neigh_timer_handler / run_timer_softirq
Kernel panic - not syncing: Fatal exception in interrupt   → reboot
```

### Root cause

An **in-use device being torn down**, exactly as suspected:

1. A packet is queued behind an **unresolved ARP/ND neighbour**.
2. The packet's **device is freed** (container/veth teardown).
3. The neighbour's retransmit **timer then fires**, declares failure, and tries to send an ICMP "destination unreachable" for the queued packet — walking the **already-freed netdevice** and faulting in **interrupt context**, which is always fatal.

`Comm: docker` is incidental (the timer interrupted docker's teardown). The crashing device has a `docker0` master, i.e. the host-side veth of a Calico FV "node" container. This is a kernel bug on the GCP 6.17 kernel that we cannot fix from Calico, so we work around it by not triggering the path.

> Note: `net.ipv4.neigh.default.unres_qlen_bytes=0` is **not** sufficient — the kernel still queues one skb unconditionally (`__skb_queue_tail` after the drop loop in `__neigh_event_send`), and one queued skb is enough to panic.

### Change

Drain neighbour state (which drops queued packets **and** cancels the timers) before the devices are freed, and stop racing the teardown:

- **`felix/fv/infrastructure/felix.go` `Felix.Stop()`** — new `drainNeighbours()`: flush the host neighbour entries for the container's own IPs (scoped with `ip neigh flush to <ip>` so parallel local `make fv` batches are undisturbed) and all neighbours inside the container, before `Container.Stop()`.
- **`felix/fv/workload/workload.go` `Workload.Stop()`** — flush the workload's neighbour state, then tear the workload down **sequentially** instead of deleting the veth and its namespace concurrently with killing the process that uses them — that concurrency (`kill & ip link del & ip netns del & wait`) was itself the race.

Both are best-effort (cannot fail a test).

### DIFF
--- a/felix/fv/infrastructure/felix.go
+++ b/felix/fv/infrastructure/felix.go
@@ -341,6 +341,24 @@ func (f *Felix) Stop() {
 		_ = f.ExecMayFail("rmdir", path.Join("/run/calico/cgroup/", f.Name))
 	}
 	f.FlowServerStop()
+
+	// Quiesce the neighbour tables before we free this container's network
+	// devices.  This works around a kernel use-after-free panic (observed on
+	// 6.17.0-1020-gcp) that reboots the test VM during teardown and flakes the
+	// FV suite.  When an unresolved ARP/ND neighbour times out, the kernel's
+	// neigh_invalidate() generates an ICMP "destination unreachable" for every
+	// packet still queued behind that neighbour
+	// (neigh_timer_handler -> arp_error_report -> ipv4_link_failure ->
+	// __icmp_send).  If the input device of a queued packet has already been
+	// torn down, __icmp_send -> icmp_route_lookup -> l3mdev_master_ifindex_rcu
+	// -> netdev_master_upper_dev_get_rcu dereferences the freed netdevice and
+	// panics in interrupt context.  Removing this container frees its veth (the
+	// docker0-bridged host side and the in-container side), so flushing the
+	// neighbour entries first drops any queued packets and cancels the
+	// retransmit timers, ensuring no timer can fire against a device we are
+	// about to free.  Best effort: teardown must not fail if these don't run.
+	f.drainNeighbours()
+
 	f.Container.Stop()
 
 	if ginkgo.CurrentSpecReport().Failed() {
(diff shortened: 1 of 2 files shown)


fasaxc:
_Claude:_ Addressed the Copilot review comments — both spots used the IPv4 `ip neigh flush` for the IPv6 address, which wouldn't flush ND state. Fixed to use `ip -6 neigh flush to <addr>` for the IPv6 case in both `Felix.Stop()` (`drainNeighbours`) and `Workload.Stop()`. Amended and force-pushed.

fasaxc:
_Claude:_ Flake-confidence check: ran the full pipeline 3× on this PR (initial CI + 2 reruns) and **`Felix: FV` passed all 3 times** with no teardown kernel panic / VM reboot. Given the prior flake was an intermittent kernel UAF on device teardown, this isn't statistical proof, but it's a good signal the neighbour-drain workaround holds. If it ever recurs, the escalation path is pinning the FV GCE image off the 6.17-gcp kernel.

felix/fv/infrastructure/felix.go
```diff
@@ -350,6 +368,26 @@ func (f *Felix) Stop() {
...
+// the kernel panic this avoids.
+//
+// On the host we flush only the entries for this container's own addresses
+// (rather than the whole table) so that batches running in parallel on the
+// same host - as `make fv` does locally - are not disturbed.  Inside the
+// container, which is about to be destroyed, we flush everything.  All calls
+// are best effort.
+func (f *Felix) drainNeighbours() {
+	if f.IP != "" {
+		_ = utils.RunMayFail("ip", "neigh", "flush", "to", f.IP)
```
tomastigera, line 382: what about the overlay devices like vxlan/ipip/wg ?

fasaxc: _Claude:_ Good question. They are covered, just not obviously: the two in-container calls (`ip neigh flush all` / `ip -6 neigh flush all`) are **not** device-scoped, so they flush the entire ARP/ND table in the felix container's netns — including entries on `vxlan.calico`, `tunl0`/ipip and `wireguard.cali`. Those overlay devices live in the container netns and are torn down when the container is removed, and we flush before `Container.Stop()`, so their neighbour timers are cancelled first.

The per-address `flush to <ip>` calls are only on the **host** side, and the host has no overlay devices — just the docker0-bridged veth (which is where the captured panic actually was). I scoped those to the container's own IPs so a parallel local `make fv` batch on the same host isn't disturbed.

One nuance worth calling out: this hook drains at container teardown. If Felix deletes a tunnel device at *runtime* (e.g. an encap-config change mid-test), that's a product-code path this test-harness hook doesn't touch — the captured flake was at container teardown, but happy to add a flush there too (or broaden the host flush) if you think the runtime path is worth guarding. I can also add a one-line comment clarifying that `flush all` intentionally covers the overlay devices.

tomastigera (approved):

PR #13127, opened 2026-07-01
Detect served MutatingAdmissionPolicy API version in the v3 CRDs chart

### LINKED ISSUE #12978: can't install projectcalico.org.v3 helm chart on kubernetes 1.36
As per the release notes of kubernetes 1.36:

> Promoted the MutatingAdmissionPolicy to GA (v1) in Kubernetes v1.36

the v1beta1 version of this resource does not exist on vanilla 1.36 kubernetes clusters including EKS.

Some resources, eg: https://github.com/projectcalico/calico/blob/v3.32.0/api/admission/networkpolicy.mutatingadmissionpolicy.yaml make the v3 CRDs chart fail to install.

caseydavenport:
The projectcalico.org.v3 helm chart couldn't install on Kubernetes 1.36, where MutatingAdmissionPolicy is only served at v1. The chart symlinked the raw api/admission YAML directly under templates/, so the hard-coded admissionregistration.k8s.io/v1beta1 apiVersion rendered verbatim and the install failed.

The admission policies are part of the projectcalico.org/v3 API surface (defaulting and validation that the aggregated apiserver would otherwise provide), so they belong wherever the v3 CRDs ship. This does two things:

- Renders the chart's admission policies through a template that detects the served API version (the same trick the calico chart already uses): v1 on 1.36+, v1alpha1 when only the alpha API is served, v1beta1 otherwise.
- Fixes the manifest generator, which was only rendering the chart's CRD templates and silently dropping the admission policies from the generated `v3_projectcalico_org.yaml`. It now renders them too, pinned to v1 to match `calico-v3-crds.yaml`.

Also drops a redundant leading document separator from protect-builtin-tiers.yaml so the policies render cleanly under `helm --show-only`.

Fixes #12978

```release-note
HELM: Fix installation of the projectcalico.org.v3 CRDs chart on Kubernetes 1.36+ by rendering the MutatingAdmissionPolicy resources at the API version the cluster actually serves.
```


mazdakn (approved):
LGTM

PR #13129, opened 2026-07-02
networking-calico: derive tap MAC from port MAC

nelljerram:
Older libvirt (before 9.5.0) had no explicit ``managed`` parameter and always overwrote the tap-side MAC by replacing the leading octet with ``0xfe``.  libvirt 9.5.0 split that hybrid behaviour behind ``managed``, and OpenStack must now use ``managed=no`` (because Nova creates the tap), which incidentally stops libvirt from touching the tap MAC.  The result: the tap-side MAC on a post-9.5.0 hypervisor is whatever Nova put in ``vif_details['mac_address']`` -- historically our static ``00:61:fe:ed:ca:fe`` -- and the VM's ARP cache goes stale after a live migration from a pre-9.5.0 source until the guest re-ARPs.  That's the tens-of-seconds packet-loss regression in CI-1936.

Reproduce libvirt's old derivation per-port in ``get_vif_details`` so Nova sets the tap MAC to ``fe:<last-5-octets-of-VM-MAC>``.  Existing bound ports keep their old ``vif_details`` value until Neutron re-runs ``bind_port`` for them (port create, hard reboot, live-migration destination rebind) -- the fix takes effect on the next bind, which is also the only moment when the tap MAC is actually consumed.

**Release note:**
```release-note
Bugfix for OpenStack: Set the host-side MAC address in the same way as libvirt used (< 9.5.0) to do for Nova-plugged TAP interfaces, but now does not (because of those interfaces having to be marked as `managed=no`).  This allows for live migration from a hypervisor with libvirt<9.5.0 to one with libvirt>=9.5.0.
```


networking-calico/networking_calico/plugins/ml2/drivers/calico/mech_calico.py
```diff
@@ -364,6 +364,42 @@ def get_logging_values(self):
...
+    maps VM MAC ``fa:16:3e:aa:bb:cc`` to tap MAC ``fe:16:3e:aa:bb:cc``.
+
+    libvirt 9.5.0 split those behaviours behind ``managed``: ``managed=yes`` (the new
+    default) creates the tap and errors out if one already exists, while ``managed=no``
+    accepts an externally-created tap but no longer sets its MAC.  OpenStack must use
+    ``managed=no`` because Nova creates the tap, and that side-effect leaves the tap MAC
+    at whatever Nova put in ``VIF_DETAILS_TAP_MAC_ADDRESS`` -- historically the static
+    value in ``DEFAULT_TAP_MAC``.  Across a live migration between a pre-9.5.0 source
+    and a post-9.5.0 destination that flips the on-wire tap MAC and the VM's ARP cache
+    goes stale until it re-ARPs, causing tens of seconds of packet loss.  See CI-1936.
```
fasaxc: Leaking internal ticket number into the code

fasaxc (approved):
One small nit

PR #13131, opened 2026-07-02
typha: fix flake in "grace period after reading snapshot" FV test

fasaxc:
The grace-period FV test set MaxFallBehind to exactly 1s and manufactured the client's "fall behind" gap with time.Sleep(time.Second) between breadcrumbs. The server only counts a grace period when it observes crumbAge > MaxFallBehind, so with crumbAge == MaxFallBehind the check was a boundary coin-flip. A run where the measured breadcrumb gap came out at 999.95ms left typha_connections_grace_used at 0 and failed the assertion.

Drop MaxFallBehind to 500ms so the ~1s breadcrumb spacing reliably clears the threshold, keeping the 1s grace window unchanged.

### DIFF
--- a/typha/fv-tests/server_test.go
+++ b/typha/fv-tests/server_test.go
@@ -896,8 +896,12 @@ var _ = Describe("With an in-process Server with short grace period", func() {
 		h.Config.PingInterval = 10000 * time.Second
 		h.Config.PongTimeout = 50000 * time.Second
 
-		// Short timeouts so we can hit them quickly.
-		h.Config.MaxFallBehind = time.Second
+		// Short timeouts so we can hit them quickly.  MaxFallBehind is kept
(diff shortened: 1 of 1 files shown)


hack/cmd/deps/deps_test.go
```diff
@@ -0,0 +1,363 @@
+// Copyright (c) 2026 Tigera, Inc. All rights reserved.
```
coutinhop, line 1: is this supposed to go into this PR? fine by me if it is...

coutinhop (approved):

hack/cmd/deps/deps_test.go
fasaxc: Sorry, I rebased on https://github.com/projectcalico/calico/pull/13106 to fix CI, still waiting for that to be approved

PR #13134, opened 2026-07-02
felix/bpf: treat bridge-enslaved NICs as L2-only

fasaxc:
## Description

When a host NIC that matches `BPFDataIfacePattern` is enslaved to a Linux bridge, Felix's BPF dataplane attached the L3 host-endpoint (TC/HEP) program to that physical port. That's the wrong device — the host's L3 (IP, routes) lives on the bridge or one of its VLAN sub-interfaces, not on the port.

With `BPFEnforceRPF=Strict` this dropped inbound traffic for a node whose IP is on a bridge sub-interface: the HEP program on the port runs the reverse-path check keyed on the port's ingress ifindex, but the return route resolves via the sub-interface, so the ifindex mismatch fails RPF.

This treats a bridge port as L2-only and never attaches the L3 program to it, mirroring the existing bond-slave handling. The L3-bearing device (the bridge or its VLAN sub-interface) must itself match `BPFDataIfacePattern` and is attached independently. Bridge ports are detected via their kernel `Protinfo`, since the netlink library does not synthesise a slave object for them (only bond/vrf).

The suppression lives in the L3/HEP attach decision (`applyProgramsToDirtyDataInterfaces`); the attach plumbing is left intact.

CORE-13089

## Changes

- `getIfaceTypeFromLink`: classify bridge ports as a new `IfaceTypeBridgeSlave` via `MasterIndex != 0 && Protinfo != nil`.
- `applyProgramsToDirtyDataInterfaces`: unconditionally suppress the TC program on a bridge port (XDP-only, redirected to the L3 device's host endpoint).
- `getEndpointType` / `getIfTypeFlags`: handle the new type; `ifstate` gains `FlgBridgeSlave` (`bridgeslave`) for introspection.
- Docs: `felix/design/bpf-overview.md` documents the L2-enslaved-device rule and the RPF rationale.

## Testing

- Unit tests (`bpf_ep_mgr_test.go`): a bridge-enslaved NIC gets no TC program while the bridge sub-interface does; the port keeps working (regains its program) when removed from the bridge; and the suppression holds even when the L3 device is not in the pattern. Full `bpf_ep_mgr` UT suite passes.
- FV (`bpf_bridge_test.go`): the node's own `eth0` is enslaved to a bridge (IP on `br0`, and IP on a VLAN sub-interface `br0.100`) before Felix starts, under `BPFEnforceRPF=Strict`. Asserts the program lands on the L3 device (not `eth0`), the ifstate flags are correct, and the node retains network — which under strict RPF would fail if the program were on the port.

**Release note:**
```release-note
Felix eBPF: if a Calico-managed host data interface is enslaved to a Linux bridge, point BPFDataIfacePattern at the bridge or sub-interface that holds the host IP; the BPF host-endpoint program no longer attaches to the enslaved port (consistent with bonded interfaces).
```

### DIFF
--- a/felix/bpf/ifstate/map.go
+++ b/felix/bpf/ifstate/map.go
@@ -39,32 +39,34 @@ const (
 )
 
 const (
-	FlgWEP        = uint32(0x1)
-	FlgIPv4Ready  = uint32(0x2)
-	FlgIPv6Ready  = uint32(0x4)
-	FlgHEP        = uint32(0x8)
-	FlgBond       = uint32(0x10)
-	FlgBondSlave  = uint32(0x20)
-	FlgVxlan      = uint32(0x40)
-	FlgIPIP       = uint32(0x80)
-	FlgWireguard  = uint32(0x100)
-	FlgL3         = uint32(0x200)
-	FlgNotManaged = uint32(0x400)
-	FlgMax        = uint32(0x7ff)
+	FlgWEP         = uint32(0x1)
+	FlgIPv4Ready   = uint32(0x2)
+	FlgIPv6Ready   = uint32(0x4)
+	FlgHEP         = uint32(0x8)
+	FlgBond        = uint32(0x10)
+	FlgBondSlave   = uint32(0x20)
+	FlgVxlan       = uint32(0x40)
+	FlgIPIP        = uint32(0x80)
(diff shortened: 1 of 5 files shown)


fasaxc:
Replaced by #13922

PR #11230, opened 2025-10-21
Enable fib for ipip, wireguard

sridhartigera:
## Description

This PR enables FIB lookup for L3 tunnels such as IPIP. Till date we have not enabled fib for L3 devices as there were issues around sending a L2 packet from WEPs to L3 devices and viceversa using bpf redirect  APIs. This can be solved by setting the tunnel key on the packet before redirecting similar to vxlan.

1. Have enabled FIB always
2. Fixed some tests so that felix gets Route updates about remote workloads and hosts.
3. Add support for `calico-node -bpf routes add` and `calico-node -bpf routes del`

## Todos

- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Enabled FIB for IPIP so that we can redirect packets in eBPF.
```

## Reminder for the reviewer

Make sure that this PR has the correct labels and milestone set.

Every PR needs one `docs-*` label.

- `docs-pr-required`: This change requires a change to the documentation that has not been completed yet.
- `docs-completed`: This change has all necessary documentation completed.
- `docs-not-required`: This change has no user-facing impact and requires no docs.

Every PR needs one `release-note-*` label.

- `release-note-required`: This PR has user-facing changes. Most PRs should have this label.
- `release-note-not-required`: This PR has no user-facing changes.

Other optional labels:

- `cherry-pick-candidate`: This PR should be cherry-picked to an earlier release. For bug fixes only.
- `needs-operator-pr`: This PR is related to install and requires a corresponding change to the operator.

PR #13135, opened 2026-07-02
[BPF] Do not drop non-Calico WireGuard traffic at host endpoints

### LINKED ISSUE #12900: Calico v3.32.0 in eBPF mode breaks wireguard overlay to other subnets
Hello,

i am using the cluster described in #11590, expanded by another local node:

Two nodes are located behind a NAT-Gateway at home, the third one is a vps with a public ip. The idea is to run the pods on the "internal" nodes at home and use the public ip as ingress and in a later step as egress. To preserve the source ip, i want to use the ebpf mode.

The nodes are directly connected with a full-meshed wireguard vpn, the needed ports are forwarded by the NAT-Gateway, including source NAT for outgoing wireguard packages. The nodes can ping each other wireguard interfaces and the cluster is setup in ebpf mode. VXLan is used on top of the wireguard interfaces to allow source ip preservetion, the kubernetes nodes are bound to the addresses of the VXLan interfaces.

On Calico 3.31.5 the cluster is working as expected.

When i upgrade the cluster to calico v3.32.0, the wireguard connectivity to the external node breaks. Removing eth0 from bpfDataIfacePattern restores the connection but the services with external ip on the vps are no longer reachable from external nodes outside of the cluster.

The issue is also reproducible on a single-node K3s cluster and occurs before any multinode Kubernetes communication is involved. After Calico starts and attaches the eBPF dataplane to eth0 , wireguard peers connected through the overlay become unreachable. Pings to wireguard clients in other subnets, like the vps, stop working. Net-Local Removing eth0 from bpfDataIfacePattern  restores connectivity.

Attaching a bridge to eth0 and using this as the main interface for K3s does not help, the connection also breaks as described above when using calico v3.32.0. Also removing the VXLan interfaces does not make a difference.

I also tried some changes to bpfDataIfacePattern and bpfL3IfacePattern but nothing helped. When eth0 is part of bpfL3IfacePattern i also lose the ssh connection to the node. 

I suspect something changed between version 3.31 and 3.32 in the handling of traffic on the physical interface.

## Expected Behavior

Calico eBPF programs attached to the physical interface should not interfere with a wireguard overlay network build outside of calico.

The wireguard traffic should continue to function regardless of whether the underlying physical interface is included in bpfDataIfacePattern .

## Current Behavior

The wireguard connection to nodes in other subnets breaks, when the Calico-Node-Pod starts and attaches the physical interface to the dataplane.

## Possible Solution

None found until now.

Minimal Setup:
1. Setup an one node cluster with calico 3.32 in ebpf mode and connect the node in another subnet with a wireguard tunnel. Use the tunnel ips as the k3s interfaces.
3. Try to ping the other nodes after the calico-node pod is running.

## Your Environment
Calico version: v3.32.0
Calico dataplane: bpf
Orchestrator version (e.g. kubernetes, openshift, etc.): k3s version v1.35.3+k3s1 (be38e884)
Operating System and version: AlpineLinux 3.23 with a self build Kairos image with the current kairos-init

k3s configuration:
```
node-ip: 192.168.201.10,fd05:2046:b984:a311::10
disable:
  - servicelb
  - traefik
  - local-storage
cluster-cidr: 10.42.0.0/18,fd05:2046:b984:a320::0/114
service-cidr: 10.43.0.0/17,fd05:2046:b984:a330::0/113
flannel-backend: none
disable-kube-proxy: true
disable-network-policy: true
cluster-init: true
kube-apiserver-arg:
  - default-not-ready-toleration-seconds=60
  - default-unreachable-toleration-seconds=60
  - enable-admission-plugins=NodeRestriction,DefaultTolerationSeconds
kube-controller-manager-arg:
  - node-cidr-mask-size-ipv4=25
  - node-cidr-mask-size-ipv6=121
  - node-monitor-grace-period=30s
  - node-startup-grace-period=30s
  - node-monitor-period=10s
```

tigera-operator.yaml

```
apiVersion: helm.cattle.io/v1
kind: HelmChart
metadata:
  name: tigera-operator
  namespace: kube-system
spec:
  # for values see https://docs.k3s.io/helm
  repo: https://docs.tigera.io/calico/charts
  chart: tigera-operator
  version: v3.32.0
  targetNamespace: tigera-operator
  createNamespace: true
  bootstrap: True
  valuesContent: |-
    kubernetesServiceEndpoint:
      host: 192.168.201.10
      port: 6443
    goldmane:
      enabled: false
    whisker:
      enabled: false
    defaultFelixConfiguration:
      enabled: true
      bpfL3IfacePattern:  bpfDataIfacePattern: ^((en|wl|ww|sl|ib)[Popsx].*|(eth|wlan|wwan|bond).*|tunl0$|vxlan.calico$|vxlan-v6.calico$|wireguard.cali$|wg-v6.cali$|vxlan_home$)
    installation:
      controlPlaneReplicas: 1
      calicoNetwork:
        mtu: 1350
        bgp: Enabled
        linuxDataplane: BPF
        nodeAddressAutodetectionV4:
          kubernetes: NodeInternalIP
        nodeAddressAutodetectionV6:
          kubernetes: NodeInternalIP
        ipPools:
          - blockSize: 25
            cidr: 10.42.0.0/18
            encapsulation: None
            natOutgoing: Enabled
            nodeSelector: all()
          - blockSize: 121
            cidr: fd05:2046:b984:a320::0/114
            encapsulation: None
            natOutgoing: Enabled
            nodeSelector: all()
```

tomastigera:
### Description

In eBPF mode, the host-endpoint TC programs treat **any** UDP packet on `WG_PORT` as Calico's own host-to-host WireGuard. Since #11308 such a packet is dropped (`CALI_REASON_UNAUTH_SOURCE`) unless the peer's *underlay* address is a known Calico host.

Two problems follow:

- `WG_PORT` defaults to `51820` — the stock WireGuard port — and is compiled in regardless of whether Calico's own WireGuard is enabled. A user-managed WireGuard mesh on `51820` therefore has all of its inbound traffic black-holed as soon as the eBPF dataplane attaches to the physical interface (the reporter's scenario in #12900: works on v3.31, breaks on v3.32).
- Even Calico's own WireGuard breaks when the underlay peer address differs from the node's Calico address (e.g. nodes behind NAT).

The adjacent VXLAN branch already handles the analogous case gracefully — it fast-allows known-host traffic but *falls through to policy* on a miss ("the user can be using VXLAN on a different VNI so we don't simply drop it"). This change makes the WireGuard branch do the same: keep the known-host fast-allow (so a user's HEP policy cannot break the cluster's own WireGuard mesh), but fall through to host-endpoint policy instead of dropping outright. Policy then governs the traffic like any other host-terminated flow.

Root cause and behavior verified against the compiled BPF program (new `bpf/ut` test loaded and run on a real kernel).

### What changed between v3.31 and v3.32

The regression was introduced by **#11308 ("[eBPF] - Fix host->host wireguard traffic", `b1d154380d`)**, which added the `if (dport == WG_PORT) { ... goto deny; }` block in `felix/bpf-gpl/parsing.h`. It builds on **#11230 ("Enable fib for ipip, wireguard", `d993149f60`)**, but #11230 only enabled FIB — the parse-time drop of unknown-source WireGuard traffic is #11308. Both landed in the v3.32 line and neither is in v3.31, which matches the report exactly:

| | WG_PORT drop in `parsing.h` | contains #11308 |
|---|---|---|
| v3.31.0 / v3.31.6 / release-v3.31 | absent | no |
| v3.32.0 / release-v3.32 | present | yes |

This PR relaxes that specific `goto deny` to a policy fall-through.

**Release note:**
```release-note
Fixed a regression in eBPF mode where traffic on the WireGuard port (51820 by default) arriving at a host interface was dropped unless it came from a known Calico node, breaking user-managed WireGuard overlays and Calico WireGuard between nodes with differing underlay addresses (e.g. behind NAT).
```

Fixes #12900

### DIFF
--- a/felix/bpf-gpl/parsing.h
+++ b/felix/bpf-gpl/parsing.h
@@ -168,24 +168,25 @@ static CALI_BPF_INLINE int tc_state_fill_from_nexthdr(struct cali_tc_ctx *ctx, b
 			}
 		}
 		if (ctx->state->dport == WG_PORT) {
-			/* Wireguard packet. Allow if to/from known Calico host. */
+			/* Wireguard packet. Fast-allow if to/from a known Calico host so
+			 * that a user's HEP policy cannot accidentally block the cluster's
+			 * own host-to-host WireGuard mesh. Otherwise fall through to
+			 * policy: WG_PORT defaults to 51820, which is also the stock
+			 * WireGuard port, so the traffic may belong to a user-managed
+			 * WireGuard unrelated to Calico - we must not drop it outright. */
 			if (CALI_F_FROM_HEP) {
 				if (rt_addr_is_remote_host(&ctx->state->ip_src)) {
 					CALI_DEBUG("Wireguard packet from known Calico host, allow.");
 					goto allow;
 				} else {
-					CALI_DEBUG("Wireguard packet from unknown source, drop.");
-					deny_reason(ctx, CALI_REASON_UNAUTH_SOURCE);
-					goto deny;
+					CALI_DEBUG("Wireguard packet from unknown source, fall through to policy.");
 				}
 			} else if (CALI_F_TO_HEP && !CALI_F_L3_DEV) {
 				if (rt_addr_is_remote_host(&ctx->state->ip_dst)) {
 					CALI_DEBUG("Wireguard packet to known Calico host, allow.");
 					goto allow;
 				} else {
-					CALI_DEBUG("Wireguard packet to unknown dest, drop.");
-					deny_reason(ctx, CALI_REASON_UNAUTH_SOURCE);
-					goto deny;
+					CALI_DEBUG("Wireguard packet to unknown dest, fall through to policy.");
 				}
 			}
 		}
--- a/felix/bpf/ut/bpf_prog_test.go
+++ b/felix/bpf/ut/bpf_prog_test.go
@@ -83,6 +83,7 @@ func init() {
 const (
 	natTunnelMTU      = uint16(700)
 	testVxlanPort     = uint16(5665)
+	testWGPort        = uint16(5666)
 	testMaglevLUTSize = uint32(31)
 )
 
@@ -899,6 +900,7 @@ func objLoad(fname, bpfFsDir, ipFamily string, topts testOpts, polProg, hasHostC
 					IfaceName:     setLogPrefix(ifaceLog),
 					MaglevLUTSize: testMaglevLUTSize,
 					IPFragTimeout: topts.ipfragTimeout,
+					WgPort:        topts.wgPort,
 				}
 				if topts.flowLogsEnabled {
 					globals.Flags |= libbpf.GlobalsFlowLogsEnabled
@@ -1314,6 +1316,7 @@ type testOpts struct {
 	istioDSCP                     int8
 	workloadSrcSpoofingConfigured bool
 	ipfragTimeout                 uint32
+	wgPort                        uint16
 }
 
 type testOption func(opts *testOpts)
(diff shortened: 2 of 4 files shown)


sridhartigera (approved):
LGTM

PR #13138, opened 2026-07-04
felix: support anycast failover for LoadBalancer VIPs

### LINKED ISSUE #13136: Support anycast failover for LoadBalancer VIPs when local endpoints are unavailable
## Expected Behavior

When a LoadBalancer Service VIP is advertised via BGP anycast from multiple clusters, traffic should fail over to another cluster when local endpoints are unavailable.

## Current Behavior

With `externalTrafficPolicy: Local`, traffic to the VIP is dropped locally on nodes without endpoints. Calico correctly withdraws the BGP route but packets are black-holed instead of being re-routed to another cluster announcing anycast site. 

1. `Cluster A` and `Cluster B` with Calico + BGP, same anycast LoadBalancer VIP.
2. Calico IPPool with `allowedUses: [LoadBalancer]`.
3. Same Service in both clusters, externalTrafficPolicy: Local.
4. Traffic from Pod on `Cluster A` to VIP → served by local `Cluster A`.
Output from the node in `Cluster A` with respective pod running:
```shell
# iptables-save | grep 1.2.3.4
-A KUBE-SERVICES -d 1.2.3.4/32 -p udp -m comment --comment "test/ntp-anycast:ntp-udp-123 loadbalancer IP" -m udp --dport 123 -j KUBE-EXT-IL6URECE27T65EPH
```
5. Scale deployment to zero on `Cluster A` → traffic dropped instead of failing over to `Cluster B`.
Output any node in `Cluster A`:
```shell
# iptables-save | grep 1.2.3.4
-A KUBE-EXTERNAL-SERVICES -d 1.2.3.4/32 -p udp -m comment --comment "test/ntp-anycast:ntp-udp-123 has no endpoints" -m udp --dport 123 -j REJECT --reject-with icmp-port-unreachable
```

See https://github.com/kubernetes/kubernetes/issues/139300 for more context.

## Context

We are trying to run anycast services with inter-cluster failover using Calico and BGP.

As per https://github.com/kubernetes/kubernetes/issues/139300 there are two possible solution areas — Kubernetes (proper API) and Calico. Happy to discuss the best approach which would fix both backends - iptables and BPF. 

## Your Environment

* Calico version: [v3.31.6](https://github.com/projectcalico/calico/releases/tag/v3.31.6) which includes https://github.com/projectcalico/calico/pull/12857 

Subhom1:
### Description
Adds opt-in anycast failover for LoadBalancer VIPs using:

```projectcalico.org/anycast: "true"```

When no local endpoint is available:

- iptables traffic accepted by Calico policy bypasses kube-proxy’s no-endpoint rejection and follows normal routing.
- The BPF dataplane falls through to routing instead of returning ICMP unreachable.
- Rules are limited to the Service’s LoadBalancer VIPs, protocols, and ports.

Fixes #13136.

### Testing

- ```go test ./felix/rules```
- ```go test ./felix/calc```
- ```go test ./felix/bpf/proxy```
- Linux dataplane compile-only test
- Linux dataplane suite: 982 passed; 12 unrelated failures due to missing generated XDP fixtures

### Release note

Support opt-in anycast failover for LoadBalancer VIPs when local endpoints are unavailable.

### DIFF
--- a/felix/bpf-gpl/nat_lookup.h
+++ b/felix/bpf-gpl/nat_lookup.h
@@ -157,7 +157,8 @@ static CALI_BPF_INLINE struct calico_nat_dest* calico_nat_lookup(ipv46_addr_t *i
 
 	if (count == 0) {
 		CALI_DEBUG("NAT: no backend");
-		*res = NAT_NO_BACKEND;
+		*res = nat_lv1_val->flags & NAT_FLG_NO_BACKEND_FORWARD ?
+			NAT_NO_BACKEND_FORWARD : NAT_NO_BACKEND;
 		return NULL;
 	}
 
--- a/felix/bpf-gpl/nat_types.h
+++ b/felix/bpf-gpl/nat_types.h
@@ -11,6 +11,7 @@ typedef enum calico_nat_lookup_result {
 	NAT_LOOKUP_ALLOW,
 	NAT_FE_LOOKUP_DROP,
 	NAT_NO_BACKEND,
+	NAT_NO_BACKEND_FORWARD,
 	NAT_EXCLUDE,
 	NAT_MAGLEV,
 } nat_lookup_result;
@@ -44,6 +45,8 @@ struct __attribute__((__packed__)) calico_nat_key {
 
 // This is used as a special ID along with count=0 to drop a packet at nat level1 lookup
 #define NAT_FE_DROP_COUNT  0xffffffff
+
+#define NAT_FLG_NO_BACKEND_FORWARD 0x10
 
 struct calico_nat_value {
 	__u32 id;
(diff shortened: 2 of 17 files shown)


fasaxc:
Haven't reviewed the code itself, but API-wise, I'm not sure `anycast` is the right framing; it's your use case, but it doesn't really explain what the flag _does_.  This should be something like `lb.projectcalico.org/no-endpoints-action: Drop | Forward`.  We already use the `lb.projectcalico.org` for Maglev annotations.

Subhom1:
> Haven't reviewed the code itself, but API-wise, I'm not sure `anycast` is the right framing; it's your use case, but it doesn't really explain what the flag _does_. This should be something like `lb.projectcalico.org/no-endpoints-action: Drop | Forward`. We already use the `lb.projectcalico.org` for Maglev annotations.

Thanks, agreed. I’ve updated the API to describe the behavior rather than the use case:
```lb.projectcalico.org/no-endpoints-action: Forward```

PR #13145, opened 2026-07-06
felix: strip trace-printk from main BPF programs under lockdown via .rodata flag

tomastigera:
### Description

Follow-up to #13142 (now merged). This is a clean standalone change on top of it.

#13142 stopped the *constant, per-interface-attach* `could not enable bpf_trace_printk events` spew under kernel `lockdown=confidentiality` by loading trace-printk-free preamble variants. The **main dataplane programs**, however, still reference the helper unconditionally: `skb_log` (the policy `Log` action) is gated at runtime (`CALI_ST_LOG_PACKET`), not by `BPFLogLevel`, so it is compiled into every program including the `no_log` builds. Loading the main program set therefore still emits a bounded, one-off burst on each Felix program-load cycle (start/restart/reload/reboot).

Rather than doubling the compiled program matrix with `_notrace` variants, this adds a small, general **`struct cali_rodata_flags`** in each program's own frozen `.rodata.cali_flags` section (separate from the main globals, so their layout is untouched). When Felix detects confidentiality lockdown it sets `no_trace_printk` before load (`bpf.SetNoTracePrintk` → `Map.SetRodataFlags`, wired through `loadObject`); the `bpf_log` macro and `skb_log` guard on it. Because the section is read-only and frozen, the verifier folds the flag to a constant and **dead-code-eliminates** the guarded branches, so the loaded program references no trace helper and its load is silent — same object serves locked-down and normal nodes. The struct is the general vehicle for future load-time feature flags.

### Tests

New `bpf/ut` load test (`TestNoTracePrintkFlagStripsTraceHelper`): loads a main program with the flag clear and set, dumps the loaded (verified) program via `bpftool prog dump xlated`, and asserts trace-printk *call instructions* are present only when the flag is clear. Passes locally alongside the existing `TestPrecompiledBinariesAreLoadable` / preamble-notrace tests.

End-to-end verification on a real `lockdown=confidentiality` cluster is pending.

### Release Note

```release-note
NONE
```

### DIFF
--- a/felix/bpf-gpl/globals.h
+++ b/felix/bpf-gpl/globals.h
@@ -86,6 +86,27 @@ struct cali_xdp_preamble_globals {
 	struct cali_xdp_globals v6;
 };
 
+/* prog_flags holds per-program, load-time feature flags, distinct from the
+ * node-wide globals above: it lives in its own frozen read-only section so it
+ * does not disturb their layout, and Felix sets it per program at load. Because
+ * the section is frozen, the verifier folds each field to a constant and
+ * dead-code-eliminates the guarded branches (e.g. a program can load with no
+ * reference to the bpf_trace_printk helper at all). New per-program load-time
+ * flags belong here — append fields as needed.
+ */
+struct prog_flags {
+	/* no_trace_printk, when set, drops the code paths that call
+	 * bpf_trace_printk/bpf_trace_vprintk. Felix sets it on nodes running with
+	 * kernel lockdown=confidentiality, where ftrace is disabled and loading any
+	 * program that references the helper spams the kernel log on every load. */
+	__u8 no_trace_printk;
+};
+
+/* Instance lives in its own .rodata section; only objects that reference it get
+ * the map. Declared volatile const so the compiler emits real loads that the
+ * verifier can fold against the frozen map. */
+__attribute__((section(".rodata.prog_flags"))) volatile const struct prog_flags PROG_FLAGS;
+
 struct cali_ct_cleanup_globals {
     __u64 creation_grace;
 
--- a/felix/bpf-gpl/log.h
+++ b/felix/bpf-gpl/log.h
@@ -24,9 +24,16 @@
 #define IPVER_PFX	""
 #endif
 
+/* The no_trace_printk guard lets Felix strip every bpf_trace_printk reference
+ * from a program at load time (see struct prog_flags): the flag lives in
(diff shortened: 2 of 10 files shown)


sridhartigera (approved):

tomastigera:
_Claude:_ CI is red, but the failures are flakes and infra, not this change. This change only gates `bpf_trace_printk` behind a `.rodata.prog_flags` flag that Felix sets only under `lockdown=confidentiality`. CI nodes are not locked down, so the flag is 0 and behavior is unchanged. It cannot affect policy, BGP, IPAM, or host endpoints.

Failures seen across runs:

**E2E tests (KinD), BPF** — flaky, and a different set each run:
- Run 1: 2 AutoHEP "should block" specs failed; **passed on rerun**.
- Latest run: `Ran 118 of 1092 Specs — 113 Passed, 5 Failed`. The 5: three AutoHEP specs failing the same `BeforeEach` (`auto_hostendpoints.go:118`, "expected host endpoints to be present, but none found" — one root cause, three specs), plus BGP password (`bgp_password.go:179`, peer not yet in CalicoNodeStatus) and IPAM GC (`ipam_gc.go:170`).
- 113/118 passed, so BPF programs load and enforce fine. The failing specs are all `[Serial]` timing-sensitive tests in kube-controllers / BIRD / IPAM paths, unrelated to BPF logging. Different specs fail each run — the signature of flakes, not a real bug.

**Felix: FV — BPF tests on Ubuntu 25.10 (nftables)** — repo-wide infra, not this PR. VM create fails: `The resource 'projects/ubuntu-os-cloud/global/images/family/ubuntu-2510-amd64' was not found`. Confirmed with gcloud: `ubuntu-os-cloud` no longer has any `ubuntu-25*` family (only `ubuntu-2404-lts` and `ubuntu-2604-lts`). Ubuntu 25.10 is an interim release past EOL, rotated out of GCP. The tests never run. This hits every Calico PR and master until `felix/.semaphore/fv-prologue` is bumped off `ubuntu-2510-amd64`.

**Felix: Windows FV** — separate common flake.

Local checks are green: `make -C felix build-bpf`, `make -C felix build`, and the `bpf/ut` load test that proves the flag strips every `bpf_trace_printk` instruction at load (`TestNoTracePrintkFlagStripsTraceHelper`), plus the existing precompiled-loadable and preamble tests. On-cluster e2e under real `lockdown=confidentiality` was also verified manually.

Re-running the E2E and Windows FV blocks should clear them.

PR #13150, opened 2026-07-07
Add s390x (big-endian) support to netfilter attribute deserialization

viddya673:
This PR solves 2 test-case failures in the felix test-suite: 

- `felix/nfnetlink/conntrack_test.go`
- `felix/nfnetlink/nflog_test.go`

**Reason of failure**: `DeserializeNfAttr()` currently reinterprets the raw `nlattr` bytes using `unsafe.Pointer` and native struct layout. This works on little-endian systems, but on big-endian systems the 16-bit `Len` and `Type` fields are read byte-swapped.
This change preserves the existing little-endian fast path and normalizes `Len` and `Type` on big-endian systems before returning the deserialized attribute.

**Fix**: Added `nativeBigEndian` helper and `swap16` helper and used them to check and swap `NfAttr` pointer to return the correct `Len` and `Type` values in-case of Big-Endian systems.

### DIFF
--- a/felix/nfnetlink/nfnl/nfnetlink_linux.go
+++ b/felix/nfnetlink/nfnl/nfnetlink_linux.go
@@ -94,12 +94,31 @@ func nfaAlignOf(attrlen int) int {
 	return (attrlen + syscall.NLA_ALIGNTO - 1) & ^(syscall.NLA_ALIGNTO - 1)
 }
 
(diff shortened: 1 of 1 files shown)


viddya673:
@mazdakn Hi, a reviewer has not been assigned yet for this PR. Can you please help me with that

PR #13151, opened 2026-07-07
argoci: add .argoci/config.yaml to onboard the repo

lwr20:
Adds the per-repo ArgoCI config. Without it, `cc-argoci-handler` finds no config, treats ArgoCI as not set up, and skips cron registration on merge — so the `.argoci/cron/` suites never register.

V2, cron-only: empty `workflows`, no `ciFile` (a `ciFile` makes the handler fetch a per-PR workflow that does not exist, erroring on every `.argoci/` PR).

## Release Note
```release-note
NONE
```

### DIFF
--- a/.argoci/config.yaml
+++ b/.argoci/config.yaml
@@ -1,0 +1,6 @@
+# ArgoCI config: scheduled e2e only. Suites are CronWorkflows in
+# .argoci/cron/*.yaml, registered on merge to the default branch. No per-PR
+# workflow (empty `workflows`, no `ciFile`). Required to onboard the repo —
+# without it the handler treats ArgoCI as not set up and skips cron registration.
+version: v2
+workflows: []


sujeet-kr (approved):

Glen-Tigera (approved):
LGTM. Verified the schema and behavior against `cc-argoci-handler`:

- `version: v2` + `workflows: []` unmarshals into `CiConfigV2`, and `GetCiConfig` nils out `V1` — so the v1 legacy `ciFile` per-PR fetch path is never taken (confirms the rationale for omitting `ciFile`).
- With an empty `Workflows` slice, the `for _, workflow := range configV2.Workflows` loop in `handlePullRequestEvent` never executes, so `.argoci/` PRs get no per-PR workflow and no error. The `workflowFile == ""` panic is only reachable for present entries.
- Before this file existed the repo fell into the `configs.V1 == nil && configs.V2 == nil` else-branch (just a Slack notice), which is exactly why the already-merged `.argoci/cron/e2e-nftables.yaml` never registered.

One operational heads-up (not a blocker): merging this config alone won't retroactively register the existing `e2e-nftables.yaml`, since `updateRequiredArgoFiles` only applies `.argoci/cron/` files *changed in the merging PR*, and this PR touches only `config.yaml`. Registering the existing cron will need an `[update-cron-workflows]` commit on `master` (or a comment command / a subsequent touch of the cron file).

PR #13153, opened 2026-07-07
argoci: make scaffold scripts bug-for-bug identical to Semaphore

lwr20:
Aligns the `.argoci/scripts/*` scaffold (from #13124) with `.semaphore/end-to-end/scripts/*`, and wires the runtime environment Semaphore's host provided implicitly so ArgoCI runs behave identically. **Merging registers the `e2e-nftables` CronWorkflow (schedule: Mondays 21:10 UTC).**

Commits:
1. Make scaffold scripts bug-for-bug identical to Semaphore. Most impactful: `run_tests.sh` keys the `make e2e-run` path on `TEST_TYPE==k8s-e2e`, not an `E2E_TEST_CONFIG` gate — so empty-config k8s-e2e jobs run the monorepo suite Semaphore runs, not `bz tests`.
2. Port `body_flannel-migration.sh`.
3. Runtime adaptations for the isolated runner: AWS creds (banzai-secrets), EKS in-container auth, `az login`, `GH_TOKEN`, `USER`, python3.10-venv stopgap, `ARGO_WORKFLOW_UID`. Also pins `GOOGLE_REGION`/`GOOGLE_ZONE` on the region-fixed gcp-kubeadm jobs (kubevirt, talos-native-gcp) whose reserved IPs live only in us-central1 — the prologue's load-spread region randomisation stays for jobs that run anywhere.

Behaviour-changing fixes to Semaphore-inherited bugs are held for DE-4390 (#15).

## Release Note
```release-note
NONE
```

### DIFF
--- a/.argoci/cron/e2e-nftables.yaml
+++ b/.argoci/cron/e2e-nftables.yaml
@@ -6,6 +6,7 @@ schedule: 10 21 * * 1
 globalPrologue: |
   cd "${CI_HOME}/${CI_GIT_DIR}"
   export ARGO_WORKFLOW_NAME="{{workflow.name}}"
+  export ARGO_WORKFLOW_UID="{{workflow.uid}}"
   export DATAPLANE="${DATAPLANE:-CalicoNftables}"
   export ENABLE_AUTOHEP="${ENABLE_AUTOHEP:-true}"
   export FUNCTIONAL_AREA="${FUNCTIONAL_AREA:-nftables.yml}"
@@ -74,6 +75,13 @@ steps:
     env:
       - name: CLUSTER_ROUTING_MODE
         value: Felix
+      # gcp-kubeadm uses the fixed semaphore-autotest VPC + reserved IPs, which
+      # exist only in us-central1; pin region/zone (the load-spread randomisation
+      # in the prologue can't apply to region-fixed jobs). See arm64-wg.
+      - name: GOOGLE_REGION
+        value: us-central1
+      - name: GOOGLE_ZONE
+        value: us-central1-a
       - name: K8S_VERSION
         value: stable
       - name: PROVISIONER
@@ -180,6 +188,13 @@ steps:
     env:
       - name: E2E_TEST_CONFIG
         value: e2e/config/gcp-kubevirt.yaml
+      # gcp-kubeadm uses the fixed semaphore-autotest VPC + reserved IPs, which
+      # exist only in us-central1; pin region/zone (the load-spread randomisation
+      # in the prologue can't apply to region-fixed jobs). See arm64-wg.
+      - name: GOOGLE_REGION
+        value: us-central1
+      - name: GOOGLE_ZONE
+        value: us-central1-a
       - name: K8S_VERSION
         value: stable-1.35
       - name: NUM_KUBEVIRT_VMS
--- a/.argoci/scripts/body_flannel-migration.sh
+++ b/.argoci/scripts/body_flannel-migration.sh
@@ -1,0 +1,142 @@
+#!/usr/bin/env bash
+# body_flannel-migration.sh - flannel-to-Calico migration test flow.
+#
+# Provisions a cluster, installs flannel + a CNI plugin helper, runs a basic
+# connectivity smoke test, applies Calico + the flannel-migration job, waits
+# for the migration to complete, then runs the full e2e suite on Calico.
+#
+# Uses the legacy `./bz.sh tests:run` test runner (not the in-repo binary).
+# When the in-repo binary reaches parity, this script can migrate to
+# `make e2e-run` like body_standard.sh's run_tests_local.sh phase.
+set -exo pipefail
+
+PHASES="$(cd "$(dirname "$0")" && pwd)/phases"
+
+echo "[INFO] starting job..."
+
+export CNI_VERSION=${CNI_VERSION:-"v1.1.1"}
+export DOCS_BASE=${DOCS_BASE:-"https://github.com/projectcalico/calico"}
+export DOWNLEVEL_MANIFEST=${DOWNLEVEL_MANIFEST:-"https://github.com/projectcalico/calico/raw/release-${RELEASE_STREAM}/manifests/canal.yaml"}
+export CALICO_MANIFEST=${CALICO_MANIFEST:-"manifests/flannel-migration/calico.yaml"}
+export MIGRATION_MANIFEST=${MIGRATION_MANIFEST:-"manifests/flannel-migration/migration-job.yaml"}
+
+if [ "${USE_HASH_RELEASE}" == "true" ]; then
+  echo "[INFO] Using hash release for flannel migration"
+  LATEST_HASHREL="https://latest-os.docs.eng.tigera.net/${RELEASE_STREAM}.txt"
+  echo "Checking ${LATEST_HASHREL} for latest hash release url..."
+  DOCS_URL=$(curl --retry 9 --retry-all-errors -sS ${LATEST_HASHREL})
+  echo "Using $DOCS_URL for hash release base url"
+else
+  if [[ "${RELEASE_STREAM}" == "master" ]]; then
+    echo "Cannot use latest release on master branch"
+    exit 1
+  else
+    echo "[INFO] Using latest release for flannel migration"
+    export DOCS_URL=$DOCS_BASE/raw/release-${RELEASE_STREAM}
+  fi
+fi
+
+export BZ_LOCAL=${BZ_HOME}/.local
+export KUBECONFIG=$BZ_LOCAL/kubeconfig
+export PATH=$PATH:$BZ_LOCAL/bin
+
+# Modern OSes no longer include br_netfilter by default, which breaks flannel.
+echo "[INFO] installing br_netfilter..."
+sudo modprobe br_netfilter
+
+mkdir -p "$BZ_LOGS_DIR"
+cd "${BZ_HOME}"
+source "${PHASES}/provision.sh"
+
+# Install bridge CNI plugin (needed by kube-flannel manifest).
+kubectl apply -f - <<EOF
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: cni-installer
+  namespace: kube-system
+  labels:
+    app: cni-installer
+spec:
+  selector:
+    matchLabels:
+      app: cni-installer
+  template:
+    metadata:
+      labels:
+        app: cni-installer
+    spec:
+      nodeSelector:
+        kubernetes.io/os: linux
+      hostNetwork: true
+      terminationGracePeriodSeconds: 0
+      tolerations:
+        - effect: NoSchedule
+          operator: Exists
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - effect: NoExecute
+          operator: Exists
+      terminationGracePeriodSeconds: 0
+      priorityClassName: system-node-critical
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
+      initContainers:
+      - name: cni-installer
+        image: quay.io/dosmith/cni-plugins:gen4
+        command: ["/bin/bash", "-c", "cp -f /usr/src/plugins/bin/* /opt/cni/bin"]
+        volumeMounts:
+        - name: bindir
+          mountPath: /opt/cni/bin
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            cpu: 10m
+            memory: 10Mi
+      containers:
+      - name: pause
+        image: registry.k8s.io/pause
+        resources:
+          requests:
+            cpu: 10m
+            memory: 10Mi
+      volumes:
+      - name: bindir
+        hostPath:
+          path: /opt/cni/bin
+EOF
+
+# Update flannel.yaml to use the podCIDR that CRC sets up.
+wget -O flannel.yaml "$DOWNLEVEL_MANIFEST"
+sed -i "s?10.244.0.0/16?192.168.0.0/16?g" ./flannel.yaml
+kubectl apply -f - < ./flannel.yaml
+sleep 30 # wait for flannel to come up
+kubectl get po -A -owide
+
+# Run a basic services test to check that flannel networking is working.
+K8S_E2E_FLAGS='--ginkgo.focus=should.serve.a.basic.endpoint.from.pods' \
+  ./bz.sh tests:run |& tee >(gzip --stdout > "${BZ_LOGS_DIR}/e2e-tests-pre.log")
+
+kubectl delete -n kube-system ds cni-installer || true  # remove the CNI installer daemonset
+kubectl apply -f "$DOCS_URL/$CALICO_MANIFEST"
+wget -O calico-migration.yaml "$DOCS_URL/$MIGRATION_MANIFEST"
+kubectl apply -f - < ./calico-migration.yaml
+sleep 5  # make sure the job has started before we check its status
+kubectl -n kube-system get jobs flannel-migration
+kubectl -n kube-system describe jobs flannel-migration
+kubectl get po -A -owide
+kubectl wait --for=condition=complete --timeout=600s -n kube-system job/flannel-migration
+kubectl -n kube-system get jobs flannel-migration
+kubectl -n kube-system describe jobs flannel-migration
+kubectl -n kube-system logs -l k8s-app=flannel-migration-controller
+kubectl get po -A -owide
+
+# Delete the migration job because the presence of a non-Running pod in
+# kube-system upsets the e2es.
+kubectl -n kube-system delete job/flannel-migration || true
+kubectl -n kube-system delete po -l k8s-app=flannel-migration-controller || true
+
+# Run e2e on uplevel calico.
+./bz.sh tests:run |& tee >(gzip --stdout > "${BZ_LOGS_DIR}/e2e-tests.log")
--- a/.argoci/scripts/global_prologue.sh
+++ b/.argoci/scripts/global_prologue.sh
@@ -31,9 +31,47 @@ if [[ -f "${GOOGLE_APPLICATION_CREDENTIALS}" ]]; then
 else
   echo "[WARN] GOOGLE_APPLICATION_CREDENTIALS missing: ${GOOGLE_APPLICATION_CREDENTIALS}"
 fi
+# Azure SP login for azr-* (bz's azure path needs an active az session).
+# Creds via banzai-secrets: AZ_SP_ID/AZ_SP_PASSWORD/AZ_TENANT_ID/AZ_SUBSCRIPTION_ID.
+if [[ "${PROVISIONER:-}" == azr-* ]]; then
+  if command -v az >/dev/null 2>&1 && [[ -n "${AZ_SP_ID:-}" ]]; then
+    az login --service-principal -u "${AZ_SP_ID}" -p "${AZ_SP_PASSWORD}" --tenant "${AZ_TENANT_ID}" >/dev/null \
+      && az account set --subscription "${AZ_SUBSCRIPTION_ID}" \
+      && echo "[INFO] az login OK (subscription ${AZ_SUBSCRIPTION_ID})" \
+      || echo "[WARN] az login failed; azr-* provisioning will fail"
+  else
+    echo "[WARN] azr-* provisioner but az CLI or AZ_SP_ID missing; azure auth skipped"
+  fi
+fi
+
 export DOCKER_AUTH_FILE="${DOCKER_AUTH_FILE:-${HOME}/.docker/config.json}"
 chmod 0600 "${HOME}"/.keys/* 2>/dev/null || true
 if [[ -f "${HOME}/.keys/marvin" ]]; then eval "$(ssh-agent -s)" >/dev/null 2>&1 || true; ssh-add "${HOME}/.keys/marvin" 2>/dev/null || true; fi
+
+# AWS creds for aws-* provisioners. banzai-secrets ships them as $credentials, a
+# ~/.aws/credentials ini blob; the SDK reads the file, so materialise it. Fall back
+# to AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY env.
+mkdir -p "${HOME}/.aws"
+if [[ -n "${credentials:-}" ]]; then
+  printf '%s\n' "${credentials}" > "${HOME}/.aws/credentials"
+  chmod 0600 "${HOME}/.aws/credentials"
+  echo "[INFO] wrote ~/.aws/credentials from banzai-secrets 'credentials' key"
+elif [[ -n "${AWS_ACCESS_KEY_ID:-}" && -n "${AWS_SECRET_ACCESS_KEY:-}" ]]; then
+  {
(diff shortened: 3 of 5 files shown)


.argoci/scripts/phases/run_tests.sh
```diff
@@ -48,10 +42,10 @@ elif [[ "${TEST_TYPE}" == "k8s-e2e" && -n "${E2E_TEST_CONFIG:-}" ]]; then
...
 
 # E2E_BINARY is a set/unset sentinel (its value is not used here -- make
-# e2e-run locates the binary itself). Take the structured path only when a
-# k8s-e2e binary was acquired above AND E2E_TEST_CONFIG selects specs; regex
-# (K8S_E2E_FLAGS) and non-e2e jobs fall through to bz tests below.
-if [[ -n "${E2E_BINARY:-}" && -n "${E2E_TEST_CONFIG:-}" ]]; then
+# e2e-run locates the binary itself). Take the structured path whenever a
+# k8s-e2e binary was acquired above; non-e2e test types fall through to bz
+# tests below. E2E_TEST_CONFIG may be empty (selects the default config).
+if [[ -n "${E2E_BINARY:-}" ]]; then
```
Glen-Tigera, line 48: This gate change (dropping `&& -n "${E2E_TEST_CONFIG:-}"`) correctly matches `.semaphore/end-to-end/scripts/run_tests.sh` — so as a parity change it's right. But it has two consequences worth confirming the `e2e-nftables` cron is migrated for in lockstep:

1. **`K8S_E2E_FLAGS` becomes dead config.** `make e2e-run` never receives `--ginkgo.focus/--ginkgo.skip` (root `Makefile` e2e-run target passes only `--calico.test-config`), yet the cron still sets an elaborate `K8S_E2E_FLAGS` on 8/9 jobs (globalPrologue L13, plus the EKS-calico-cni per-job value at L122–123 with extra **known-failure** DNS/proxy skips). Those skips are now ignored, so previously-skipped tests will actually run and likely fail.
2. **Empty `E2E_TEST_CONFIG` overrides the Makefile default.** Those same 8 jobs set no `E2E_TEST_CONFIG`, and this path passes `E2E_TEST_CONFIG=''` unconditionally — which overrides the Makefile's `?= e2e/config/kind.yaml` and hands the binary an empty `--calico.test-config=` rather than the kind.yaml default.

If historical Semaphore ran these nftables jobs through `make e2e-run` (not `bz`+`K8S_E2E_FLAGS`), then the fix belongs in the cron: drop the now-dead `K8S_E2E_FLAGS` and give each job a proper `E2E_TEST_CONFIG`. Flagging since the cron isn't in this PR.

.argoci/scripts/global_prologue.sh
```diff
@@ -44,12 +44,13 @@ RANDOM_TOKEN1=$(LC_ALL=C tr -dc 'a-z0-9' </dev/urandom | head -c 5 || true)
...
-export INSTALLER=${INSTALLER:-operator}
+export INSTALLER=${INSTALLER:-"manual"}
 export DATAPLANE=${DATAPLANE:-CalicoIptables}
 export TEST_TYPE=${TEST_TYPE:-k8s-e2e}
 export GOOGLE_PROJECT=${GOOGLE_PROJECT:-unique-caldron-775}
-export GOOGLE_REGION=${GOOGLE_REGION:-us-central1}
-export GOOGLE_ZONE=${GOOGLE_ZONE:-us-central1-a}
+export GOOGLE_REGIONS=("us-central1" "us-west1")
+export GOOGLE_REGION=${GOOGLE_REGION:-${GOOGLE_REGIONS[RANDOM%${#GOOGLE_REGIONS[@]}]}}
+export GOOGLE_ZONE=${GOOGLE_ZONE:-$(gcloud compute zones list --filter="region~'$GOOGLE_REGION'" --format="value(name)" | awk 'BEGIN {srand()} {a[NR]=$0} rand() * NR < 1 {zone=$0} END {print zone}')}
```
Glen-Tigera, line 94: `GOOGLE_ZONE` is derived from `gcloud … | awk` inside `${GOOGLE_ZONE:-$(…)}`. The prologue runs under `set -o pipefail` but deliberately **not** `set -e` (it's sourced), and this assignment returns rc=0 even when the substitution is empty. So if `gcloud` fails (auth/API/quota), `GOOGLE_ZONE` is silently set to `""` and GCP provisioning proceeds with an empty zone instead of failing fast with a clear error. Consider validating the result is non-empty (and erroring if not).

Minor, related: this `gcloud compute zones list` runs on every job including the AWS-provisioner ones (aws-talos/aws-kubeadm/aws-eks) that never use `GOOGLE_ZONE` — worth gating on a GCP provisioner.

.argoci/scripts/global_prologue.sh
```diff
@@ -44,12 +44,13 @@ RANDOM_TOKEN1=$(LC_ALL=C tr -dc 'a-z0-9' </dev/urandom | head -c 5 || true)
...
 export PROVISIONER=${PROVISIONER:-gcp-kubeadm}
-export INSTALLER=${INSTALLER:-operator}
+export INSTALLER=${INSTALLER:-"manual"}
 export DATAPLANE=${DATAPLANE:-CalicoIptables}
 export TEST_TYPE=${TEST_TYPE:-k8s-e2e}
 export GOOGLE_PROJECT=${GOOGLE_PROJECT:-unique-caldron-775}
-export GOOGLE_REGION=${GOOGLE_REGION:-us-central1}
-export GOOGLE_ZONE=${GOOGLE_ZONE:-us-central1-a}
+export GOOGLE_REGIONS=("us-central1" "us-west1")
+export GOOGLE_REGION=${GOOGLE_REGION:-${GOOGLE_REGIONS[RANDOM%${#GOOGLE_REGIONS[@]}]}}
```
Glen-Tigera, line 93: `GOOGLE_REGION` is now randomly chosen from `{us-central1, us-west1}`, but `GOOGLE_NETWORK` stays the fixed `semaphore-autotest` VPC (L54). Jobs that don't pin a region (e.g. `nftables-talos-native-v3-crds-gcp`, `kubevirt-e2e-tests`) will land in `us-west1` ~50% of the time — if `semaphore-autotest` is a custom-mode VPC with no `us-west1` subnet, `bz provision` will fail to place the instance. Worth confirming a `us-west1` subnet (and any region-specific machine types/images) exist before this rides.

.argoci/scripts/body_flannel-migration.sh
```diff
@@ -0,0 +1,142 @@
...
+# `make e2e-run` like body_standard.sh's run_tests_local.sh phase.
+set -exo pipefail
+
+PHASES="$(cd "$(dirname "$0")" && pwd)/phases"
+
+echo "[INFO] starting job..."
+
+export CNI_VERSION=${CNI_VERSION:-"v1.1.1"}
+export DOCS_BASE=${DOCS_BASE:-"https://github.com/projectcalico/calico"}
+export DOWNLEVEL_MANIFEST=${DOWNLEVEL_MANIFEST:-"https://github.com/projectcalico/calico/raw/release-${RELEASE_STREAM}/manifests/canal.yaml"}
```
Glen-Tigera, line 19: `DOWNLEVEL_MANIFEST` hardcodes `release-${RELEASE_STREAM}`. On `master` (the default stream, with `USE_HASH_RELEASE=true` also the default) this expands to `…/raw/release-master/manifests/canal.yaml` — there is no `release-master` branch, so `wget -O flannel.yaml` (L112) 404s (or truncates the file) and the flannel install breaks. The `DOCS_URL` guard above only `exit 1`s on master when `USE_HASH_RELEASE != true`, so the hashrelease path still reaches this. Only `release-vX.Y` streams work today.

.argoci/scripts/body_flannel-migration.sh
```diff
@@ -0,0 +1,142 @@
...
+      hostNetwork: true
+      terminationGracePeriodSeconds: 0
+      tolerations:
+        - effect: NoSchedule
+          operator: Exists
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - effect: NoExecute
+          operator: Exists
+      terminationGracePeriodSeconds: 0
```
Glen-Tigera, line 80: `terminationGracePeriodSeconds: 0` is set twice in this pod spec (also at the line above the tolerations block). Duplicate YAML map key — harmless value-wise (both 0), but strict kubectl field validation (default since ~1.25) can reject a duplicate field, and at best it's confusing. Suggest deleting this one.

```suggestion
      priorityClassName: system-node-critical
```

.argoci/scripts/body_flannel-migration.sh
```diff
@@ -0,0 +1,142 @@
...
+
+kubectl delete -n kube-system ds cni-installer || true  # remove the CNI installer daemonset
+kubectl apply -f "$DOCS_URL/$CALICO_MANIFEST"
+wget -O calico-migration.yaml "$DOCS_URL/$MIGRATION_MANIFEST"
+kubectl apply -f - < ./calico-migration.yaml
+sleep 5  # make sure the job has started before we check its status
+kubectl -n kube-system get jobs flannel-migration
+kubectl -n kube-system describe jobs flannel-migration
+kubectl get po -A -owide
+kubectl wait --for=condition=complete --timeout=600s -n kube-system job/flannel-migration
```
Glen-Tigera, line 130: Nit: `kubectl wait --for=condition=complete` only returns early when the Job reaches `complete` (or is deleted). A **failed** migration sets `Failed=true`, never `complete=true`, so this blocks the full `--timeout=600s` and only then errors — ~10 wasted minutes per broken run. Consider also racing a `--for=condition=failed` (or a jsonpath poll) to abort fast on failure.

.argoci/scripts/body_flannel-migration.sh
```diff
@@ -0,0 +1,142 @@
...
+        - effect: NoExecute
+          operator: Exists
+      terminationGracePeriodSeconds: 0
+      priorityClassName: system-node-critical
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
+      initContainers:
+      - name: cni-installer
+        image: quay.io/dosmith/cni-plugins:gen4
```
Glen-Tigera, line 87: `quay.io/dosmith/cni-plugins:gen4` is a personal quay namespace with a mutable tag baked into shared CI. If that account/repo/tag is deleted, made private, or repushed, every flannel-migration run silently breaks or changes behavior with no provenance. If this was copied verbatim from Semaphore for parity, fine to confirm as intended — otherwise worth moving to an org-owned registry (e.g. `quay.io/calico/…`) with a pinned digest.

.argoci/scripts/body_flannel-migration.sh
```diff
@@ -0,0 +1,142 @@
...
+# Update flannel.yaml to use the podCIDR that CRC sets up.
+wget -O flannel.yaml "$DOWNLEVEL_MANIFEST"
+sed -i "s?10.244.0.0/16?192.168.0.0/16?g" ./flannel.yaml
+kubectl apply -f - < ./flannel.yaml
+sleep 30 # wait for flannel to come up
+kubectl get po -A -owide
+
+# Run a basic services test to check that flannel networking is working.
+K8S_E2E_FLAGS='--ginkgo.focus=should.serve.a.basic.endpoint.from.pods' \
+  ./bz.sh tests:run |& tee >(gzip --stdout > "${BZ_LOGS_DIR}/e2e-tests-pre.log")
```
Glen-Tigera, line 120: Nit: these `tee >(gzip --stdout > …)` sinks write gzip'd bytes into files named `.log` (also `e2e-tests.log` at the end of the script), whereas `phases/provision.sh` uses `.log.gz` for the same pattern. Anyone/anything opening `e2e-tests*.log` gets binary garbage, and it defeats tooling keyed on `.gz`. Suggest `e2e-tests-pre.log.gz` / `e2e-tests.log.gz`.

.argoci/scripts/global_prologue.sh
```diff
@@ -59,6 +60,11 @@ export AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-us-west-2}
 # BRANCH then master. release-vX.Y -> vX.Y, else master.
 export RELEASE_STREAM=${RELEASE_STREAM:-$( _b="${CI_GIT_CLONED_BRANCH:-${BRANCH:-master}}"; [[ "${_b}" =~ ^release-(v[0-9]+\.[0-9]+)$ ]] && echo "${BASH_REMATCH[1]}" || echo "master" )}
 
+export K8S_VERSION=${K8S_VERSION:-stable-3}
+export K8S_E2E_EXTRA_FLAGS=${K8S_E2E_EXTRA_FLAGS:-" --e2ecfg.calicoctl-opensource-image=calico/ctl:release-${RELEASE_STREAM} "}
```
Glen-Tigera, line 105: Latent trap: when `RELEASE_STREAM=master` (the default), this default becomes `calico/ctl:release-master`, which is almost certainly not a published tag (master builds are `calico/ctl:master`). It happens to be inert on the `make e2e-run` path (this var isn't forwarded into the container or onto the make line), so it'd only bite the `bz tests` (non-e2e) path — but worth fixing the master casing regardless.

lwr20:
Thanks Glen — thorough as always. Confirming the framing: this PR's contract is bug-for-bug parity with `.semaphore/end-to-end/scripts/`, and every item here is verbatim from that source:

- **#1** — `TEST_TYPE` defaults to `k8s-e2e` (`global_prologue.sh` L57) and Semaphore's `run_tests.sh` passes `E2E_TEST_CONFIG=''` unconditionally (L96). So on Semaphore these nftables jobs *already* ran through `make e2e-run` with an empty config and already-dead `K8S_E2E_FLAGS` — this PR reproduces that rather than introducing it.
- GOOGLE_ZONE L125; GOOGLE_REGION L61-62; flannel DOWNLEVEL L19, dup `terminationGracePeriodSeconds` L72/80, `dosmith` image L87, gzip-into-`.log` L120/142; calicoctl `calico/ctl:release-${RELEASE_STREAM}` L80 — all verbatim.

They're all real latent issues worth fixing, but doing so here would diverge from Semaphore (and trip the scaffold-parity guard), so I'd rather not block parity on them. Tracked all nine as a follow-up ArgoCI-side hardening pass in Jira **DE-4390** — **#1 first**, since giving each k8s-e2e job a proper `E2E_TEST_CONFIG` (and dropping the dead flags) changes what runs and deserves a before/after test-count check.

Good to approve this as the faithful baseline?

.argoci/scripts/phases/run_tests.sh
Glen-Tigera: Agreed — I confirmed the same when checking against `.semaphore/end-to-end/scripts/run_tests.sh` (identical `[[ -n "${E2E_BINARY:-}" ]]` gate + `make e2e-run` with `E2E_TEST_CONFIG=''`), so this is a faithful reproduction of existing Semaphore behavior, not a regression introduced here. Deferring to DE-4390 makes sense, and sequencing #1 first with a before/after test-count check is exactly right — that's the one with real behavior change (dead `K8S_E2E_FLAGS` → proper per-job `E2E_TEST_CONFIG`). Thanks for filing all nine.

Glen-Tigera (approved):
Approving as the faithful Semaphore baseline. 👍

I re-checked the framing and it holds up: the scripts reproduce `.semaphore/end-to-end/scripts/` bug-for-bug — I'd independently verified the `run_tests.sh` gate + `make e2e-run` path against the Semaphore source, and the prologue/flannel items are all verbatim from there too. So none of my nine findings are regressions *introduced* by this PR; they're pre-existing latent issues coming along for parity. Fixing them here would diverge from Semaphore and trip the scaffold-parity guard, so tracking them in DE-4390 (with #1 first, behind a before/after test-count check) is the right split.

New commit since my review (`e5f36bb4e`, `ARGO_WORKFLOW_UID` export) is a benign one-liner. CI green. Good to merge as the baseline; the hardening lands in DE-4390.

sabags (commented):
Reviewed against the Semaphore counterparts to check the "bug-for-bug identical" claim. The core restores hold up: `phases/run_tests.sh` now keys selection on `TEST_TYPE == k8s-e2e` → `E2E_BINARY` exactly like `.semaphore/end-to-end/scripts/phases/run_tests.sh`, and the prologue defaults (`INSTALLER=manual`, `K8S_VERSION=stable-3`, `K8S_E2E_EXTRA_FLAGS`, `HELM_PATCH`, `CALICOCTL_INSTALL_TYPE`, `GOOGLE_REGIONS` randomisation) match the Semaphore prologue verbatim. The `aws-iam-authenticator` mount, `ocp-cert` always-diags branch, AWS/Azure/GH_TOKEN auth, and `USER` default all look correct. A few things worth a second look:

---

**1. `run_tests.sh` change makes the nftables cron's `K8S_E2E_FLAGS` inert — confirm the intended coverage.** (`.argoci/cron/e2e-nftables.yaml:14`, `.argoci/scripts/phases/run_tests.sh:31`)

This is the one behavioural change I want to flag loudly. `e2e-nftables.yaml` is the only cron, it sets `K8S_E2E_FLAGS` (the `[sig-calico]|[Conformance]` focus) but **no `E2E_TEST_CONFIG`**, and `TEST_TYPE` defaults to `k8s-e2e`. Tracing the new path:

- `body_standard.sh:21` sources `run_tests.sh`.
- `TEST_TYPE == k8s-e2e` → downloads the binary → `E2E_BINARY` set → the `if [[ -n "${E2E_BINARY:-}" ]]` branch → `make e2e-run`.
- `make e2e-run` (root `Makefile:281`) invokes ginkgo with only `--calico.test-config=$(abspath $(E2E_TEST_CONFIG))` — no `--ginkgo.focus`. `K8S_E2E_FLAGS` is read **nowhere** in the repo (grep confirms: only the prologue `export`s it and run_tests' INFO echo references it).

So after this PR the nftables cron runs `make e2e-run` with an **empty** `E2E_TEST_CONFIG` (default config) and its `K8S_E2E_FLAGS` focus is dropped. Before the PR the `&& -n E2E_TEST_CONFIG` gate sent it to `bz tests`, which *does* apply `K8S_E2E_FLAGS`. This exactly mirrors Semaphore's `nftables.yml` (same global `K8S_E2E_FLAGS`, no `E2E_TEST_CONFIG` on the main jobs, same `make e2e-run` route), so it is faithful — **but** it means the sole scheduled argoci e2e run changes what it executes, and the large `K8S_E2E_FLAGS` block in a cron this PR is actively editing is now dead. Two asks:
- Confirm an empty `E2E_TEST_CONFIG` actually exercises the intended nftables/sig-calico coverage under `make e2e-run` (vs. some default that skips it).
- If specific specs are wanted, select them via `E2E_TEST_CONFIG` (like the KubeVirt job in `nftables.yml` does with `e2e/config/gcp-kubevirt.yaml`) rather than the now-inert `K8S_E2E_FLAGS`.

**2. `body_flannel-migration.sh` is added but wired to nothing — and the PR body says it isn't in scope.** (`.argoci/scripts/body_flannel-migration.sh`)

The description states: *"Not covered here (follow-up… ): restoring `body_flannel-migration.sh` / `phases/hcp.sh` and wiring the flannel/HCP jobs to them."* — yet the file is added in this PR. No cron/template references it (grep for `body_flannel-migration` finds only the file itself), so it's dead until the follow-up. Either the file landed here by mistake or the description is stale; please reconcile so reviewers know whether it's in scope. (The content itself is a faithful port of the Semaphore script — only `PHASES=` differs, and that change is an improvement.)

**3. `ARGO_WORKFLOW_UID` is exported but consumed nowhere.** (`.argoci/cron/e2e-nftables.yaml:9`)

`{{workflow.uid}}` is exported only in this one cron and is referenced nowhere in `.argoci/` (or `.semaphore/`). The epilogue builds `ARTIFACT_DEST` from `ARGO_WORKFLOW_NAME`, not the UID. If the intent was to disambiguate artifact paths across workflow re-runs (name can repeat), the epilogue needs to use it; otherwise it's a dead export. Also, if it's meant to be a standard var it should live wherever the other crons get their common exports, not only here.

**4. Nit: duplicate `terminationGracePeriodSeconds: 0` in the `cni-installer` DaemonSet.** (`.argoci/scripts/body_flannel-migration.sh:72` and `:80`)

Same key twice in the same pod spec (harmless — same value, last wins — and it's faithful to the Semaphore original), but since this is a newly-added file it's a cheap copy-paste to drop.

.argoci/cron/e2e-nftables.yaml
```diff
@@ -6,6 +6,7 @@ schedule: 10 21 * * 1
 globalPrologue: |
   cd "${CI_HOME}/${CI_GIT_DIR}"
   export ARGO_WORKFLOW_NAME="{{workflow.name}}"
+  export ARGO_WORKFLOW_UID="{{workflow.uid}}"
```
sabags, line 9: This exports `ARGO_WORKFLOW_UID`, but I can't find anything that reads it — the epilogue builds the artifact path from `ARGO_WORKFLOW_NAME`, not the UID, and there are no other references in `.argoci/`. Was it meant to be used somewhere (e.g. to keep artifact paths unique across workflow re-runs, since the name can repeat)? If not, it looks like a dead export and can be dropped. Also, it's only set in this one cron — if it's meant to be a standard var, it should probably live wherever the other crons pick up their shared exports.

.argoci/scripts/body_flannel-migration.sh
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env bash
+# body_flannel-migration.sh - flannel-to-Calico migration test flow.
```
sabags, line 2: This script is added here, but nothing seems to invoke it yet — I don't see any cron or template referencing `body_flannel-migration.sh`. So as of this PR it's dead until something wires it up. Is it intentionally staged ahead of the job wiring, or should the wiring land in the same PR? Just flagging so it isn't left dangling by accident.

.argoci/cron/e2e-nftables.yaml
lwr20: Not dead — it's consumed by Lens, just invisibly to a code-grep. The `lens-report` ingest sweeps the whole runner environment into each case doc, so exporting `ARGO_WORKFLOW_UID` makes it a queryable field (`argo_workflow_uid`) in Lens. Confirmed in yesterday's run: `e2e-nftables-13153-pr-84bll`'s docs carry `argo_workflow_uid = 908e828f-…`. `run-lens.sh` never names it (it reads `TEST_TYPE`/`LENS_*`/etc.), which is why grepping the scaffold comes up empty.

Purpose: `ARGO_WORKFLOW_NAME` can repeat across re-runs, but the uid is unique — so it's the reliable key to pull one specific run's results out of Lens.

Good call on the second point though — it was only on this cron. I've now exported it across all the crons in #13133, so every suite gets the same per-run provenance.

.argoci/scripts/body_flannel-migration.sh
lwr20: Intentional staging. It's wired by the flannel-migration jobs in #13133 (`e2e-iptables.yaml` and `e2e-patch-verification.yaml` both invoke `body_flannel-migration.sh`). Ported here so the scaffold scripts land as one coherent unit; the jobs that call it come with the rest of the cron set in the follow-up. Thanks for flagging — agreed it'd be a smell if it were left dangling, but it isn't.

PR #13161, opened 2026-07-08
felix: abort parked live-migration timer pops when TimeWait is exited

nelljerram:
The elevated-routing timer's AfterFunc callback delivers its pop over the unbuffered timerC, whose only reader is the dataplane main loop. If the timer fires while the main loop is busy processing a batch, the callback parks in the send.  Unlike the GARP-detection case (fixed previously), nothing waits for this callback, so it cannot deadlock the main loop - but a parked pop could be delivered late, after the FSM had left TimeWait.  If a re-migration in the same batch had put the FSM back into TimeWait by then, OnTimerPop's staleness check ("FSM exists and is in TimeWait") would misattribute the stale pop to the new episode and cut its TimeWait short, reverting the workload's route to normal priority before routing had converged - the transient misrouting that TimeWait exists to prevent.

Apply the same abort pattern as GARP detection: the callback sends via a select over a per-timer stop channel that stopElevatedRoutingTimer() closes, so leaving TimeWait always aborts a parked pop before any new episode can begin.  The regression test parks an episode-1 pop, drives the FSM out of and back into TimeWait, and verifies the stale pop is never delivered; without the fix it reproduces the misattribution deterministically.

**Release note:**
```release-note
Fixed a rare race in OpenStack/KubeVirt live migration handling where, after back-to-back migrations of the same VM, Felix could revert the VM's route to normal priority before BGP routing had converged.
```

### DIFF
--- a/felix/dataplane/linux/live_migration.go
+++ b/felix/dataplane/linux/live_migration.go
@@ -241,8 +241,8 @@ type liveMigrationFSM struct {
 	monitor      *liveMigrationMonitor
 	currentState liveMigrationState
 	migrationUID string
-	timer        *time.Timer
 	timerStopC   chan struct{}
+	timerWG      sync.WaitGroup
 	pcapHandle   garpHandle
 	garpStopC    chan struct{}
 	garpWG       sync.WaitGroup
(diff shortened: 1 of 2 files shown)


felix/dataplane/linux/live_migration.go
```diff
@@ -384,16 +385,30 @@ func (f *liveMigrationFSM) stopGARPDetection() {
...
+	// put the FSM back into TimeWait by then, OnTimerPop would misattribute the stale pop to
+	// the new episode and cut its TimeWait short, reverting the workload's route to normal
+	// priority before routing has converged.
+	stopC := make(chan struct{})
+	f.timerStopC = stopC
 	f.timer = time.AfterFunc(f.monitor.convergenceTime, func() {
-		f.monitor.timerC <- id
+		select {
+		case f.monitor.timerC <- id:
+		case <-stopC:
```
fasaxc: This only reduces the race it doesn't eliminate it; if both `stopC` and `timerC` are unblocked, the choice is random so it might still send the timer pop.

The proper fix is to use a wait group to wait for the timer func to finish in stopElevatedRoutingTimer but it takes care because `timer.Stop()` may prevent the func() from running at all if it's called early enough so you either need to check `Stop()`'s return value, or refactor to use a full goroutine:

```
go func(){
  defer wg.Done()
  timerC := time.After(f.monitor.convergenceTime)
  var outputC chan  types.WorkloadEndpointID
  for {
    select {
      case <-timerC:
        // Timer popped, enable the output channel.
        outputC = f.monitor.timerC
      case <-stopC:
        return
      case outputC<-id:
        return
    }
  }
}
```

fasaxc (changes requested):

felix/dataplane/linux/live_migration.go
nelljerram: Thanks, agreed.  Have adopted your suggestion of using a full goroutine, plus a WG that `stopElevatedRoutingTimer` waits on — the wait runs on the main loop (sole `timerC` reader), so the send case can't fire during it and the goroutine must take the stop case.

felix/dataplane/linux/live_migration.go
```diff
@@ -384,16 +385,58 @@ func (f *liveMigrationFSM) stopGARPDetection() {
...
+	// to abort that send, an already-fired timer could deliver its pop late, after the FSM has
+	// left TimeWait - and if a re-migration has put the FSM back into TimeWait by then,
+	// OnTimerPop would misattribute the stale pop to the new episode and cut its TimeWait
+	// short, reverting the workload's route to normal priority before routing has converged.
+	//
+	// stopC alone is not enough to close that hole: if the goroutine reaches its select after
+	// stopC is closed and the main loop is back at its own select, both cases are ready and Go
+	// chooses between them at random.  The guarantee comes from stopElevatedRoutingTimer()
+	// waiting on timerWG: while the main loop is blocked in that Wait() it cannot be receiving
+	// from timerC, so the send case cannot fire and the goroutine must take the stop case.
```
fasaxc: Feel like this over-explains, the default assumption of the reader should be that the code is as complex as needed, but no more so.  If there's a wait group, then the intent is clear, stop should wait for the background goroutine to stop.  A brief note along the lines of "Must wait for the background goroutine to finish to prevent just-cancelled goroutine from sending a stale value." would be the right level IMO.

felix/dataplane/linux/live_migration.go
```diff
@@ -384,16 +385,58 @@ func (f *liveMigrationFSM) stopGARPDetection() {
...
+		f.logCtx.Warn("Elevated routing timer already running; stopping it before restart")
+		f.stopElevatedRoutingTimer()
+	}
+	stopC := make(chan struct{})
+	f.timerStopC = stopC
+	f.timerWG.Add(1)
+	go func() {
+		defer f.timerWG.Done()
+		timer := time.NewTimer(f.monitor.convergenceTime)
+		defer timer.Stop()
```
fasaxc: I'd reach for 
```
timerC := time.After(f.monitor.convergenceTime)
```
these days; in [recent Go versions](https://go.dev/wiki/Go123Timer), the GC is aware of timer channels and it cleans the timer up automatically when they are no longer reachable.

felix/dataplane/linux/live_migration.go
nelljerram: I'm not entirely following, because your "Must wait ..." suggestion feels like it belongs in `stopElevatedRoutingTimer`, and there is already a note like that there.  But I agree the note _here_ is over-explained and will look at making it more concise.

nelljerram: OK, I think both shorter and clearer now - WDYT?

fasaxc (approved):

PR #13175, opened 2026-07-09
Felix: fix rare start-of-day deadlock on restart-requiring config change

fasaxc:
Bug fix for a rare but permanent start-of-day deadlock in Felix, seen as an FV flake ("Spoof tests ... should drop spoofed traffic" timing out waiting for felix to restart after `NodeIP`/`IpInIpTunnelAddr` were written during topology setup).

The deadlock is a three-way cycle in `daemon.Run()`:

1. The calc graph was started before the opening `ConfigUpdate` was sent to the dataplane connector. Its first flushed event is always a `ConfigUpdate` (`EventSequencer.Flush()` flushes config first), so it could win the race onto the unbuffered `ToDataplane` channel.
2. If that `ConfigUpdate` carried a restart-requiring change, the connector pump parked in `shutDownProcess()` on an unconditional send to the unbuffered `failureReportChan` — whose only reader, `monitorAndManageShutdown()`, had not started yet.
3. The main goroutine could never start it: it was parked sending the opening `ConfigUpdate` that the pump would never consume. The panic backstop in `shutDownProcess()` sat *after* the blocking send, so it never fired; felix hung forever instead of restarting.

Changes:

- **Reorder start of day**: start the calc graph only after the opening `ConfigUpdate` has been handed to the connector. This removes the deadlock's enabling condition structurally (at the handoff, main is the only sender and has no further blocking channel ops before the shutdown monitor starts) and restores the documented protocol ordering — the dataplane driver receives its config before the stream of calc-graph updates ("ConfigUpdate is sent at start of day", `felixbackend.proto`).
- **Make the shutdown backstops reachable** (defence in depth): `shutDownProcess()` and the config-changed/fatal-error callbacks now time out their failure-report send, so a stuck send panics (and restarts felix) rather than deadlocking. `shutDownProcess()`'s backstop is extended from 5s to `gracefulShutdownTimeout` (30s) so it cannot beat a legitimately slow external-driver shutdown (~7s worst case in the monitor's kill path).
- **Regression test**: `felix/daemon/shutdown_test.go` asserts `shutDownProcess()` panics rather than blocking forever when the shutdown monitor never reads (hangs on the old code), and that the reason is still delivered when a reader exists.

Testing: new UTs plus the existing `felix/daemon` suite pass; `go vet` clean. The reordered graceful-restart path is exercised by the existing FV restart tests; the original failure is timing-dependent and not reproducible on demand.

**Release note:**
```release-note
Fixed a rare start-of-day deadlock that could prevent Felix from restarting when a restart-requiring configuration change (e.g. the node IP being set) arrived while Felix was still starting up.
```

### DIFF
--- a/felix/daemon/daemon.go
+++ b/felix/daemon/daemon.go
@@ -711,25 +711,13 @@ configRetry:
 	}
 
 	// Send the opening message to the dataplane driver, giving it its
-	// config.  Do this before starting the calculation graph so that the
-	// dataplane driver is guaranteed to receive its config before the
-	// stream of updates from the calculation graph.
-	//
-	// Starting the calc graph first used to allow a rare start-of-day
-	// deadlock: the graph's first flushed event is always a ConfigUpdate,
-	// and if it carried a restart-triggering change it could park the
-	// connector in shutDownProcess() — reporting on failureReportChan —
-	// before monitorAndManageShutdown(), the only reader of that channel,
-	// had started.  This opening ConfigUpdate can't itself trigger a
-	// restart: handleConfigUpdate diffs it against the very config object
-	// it was generated from.
+	// config.  Do this before starting the calculation graph so that we
+	// don't race with it to send the first message.  We used to start calc
+	// graph first but that could result in blocking or even deadlocking
+	// here.
 	dpConnector.ToDataplane <- configParams.ToConfigUpdate()
 
 	// Now the dataplane driver has its config, start the calculation graph.
-	// Every consumer of its output channels is already running (the
-	// connector pump, the policy-sync processor, the collector's
-	// DataplaneInfoReader); the syncer has meanwhile been accumulating
-	// updates in the syncerToValidator dedupe buffer.
 	go syncerToValidator.SendToSinkForever(validator)
 	asyncCalcGraph.Start()
 	log.Infof("Started the processing graph")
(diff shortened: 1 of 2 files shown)


felix/daemon/daemon.go
```diff
@@ -1448,12 +1485,26 @@ func (fc *DataplaneConnector) sendMessagesToDataplaneDriver() {
...
-	// Send a failure report to the managed shutdown thread then give it
-	// a few seconds to do the shutdown.
-	fc.failureReportChan <- reason
-	time.Sleep(5 * time.Second)
-	// The graceful shutdown failed, terminate the process.
+	timeout := time.After(fc.shutdownReportTimeout)
+	select {
+	case fc.failureReportChan <- reason:
+	case <-timeout:
+		log.Panic("Managed shutdown failed: timed out reporting shutdown reason. Panicking.")
```
fasaxc: Include the reason in the panic messages

felix/daemon/daemon.go
```diff
@@ -700,9 +711,29 @@ configRetry:
...
+	// stream of updates from the calculation graph.
+	//
+	// Starting the calc graph first used to allow a rare start-of-day
+	// deadlock: the graph's first flushed event is always a ConfigUpdate,
+	// and if it carried a restart-triggering change it could park the
+	// connector in shutDownProcess() — reporting on failureReportChan —
+	// before monitorAndManageShutdown(), the only reader of that channel,
+	// had started.  This opening ConfigUpdate can't itself trigger a
+	// restart: handleConfigUpdate diffs it against the very config object
+	// it was generated from.
```
fasaxc:
```suggestion
	// config.  Do this before starting the calculation graph so that we
	// don't race with it to send the first message.  We used to start calc
	// graph first but that could result in blocking or even deadlocking
	// here.
```

sridhartigera (approved):

PR #13176, opened 2026-07-09
felix: support a host capability fallback for NFTablesMode=Auto

hjiawei:
NFTablesMode=Auto resolves via kube-proxy detection, which gives no signal on a host that does not run kube-proxy: today such hosts always fall back to iptables. As more distros drop the iptables/ipset userspace in favor of nftables, hosts capable of running the nftables dataplane should be able to select it without per-distro pinned configuration.

Add a host capability check that useNftables consults in Auto mode, when supplied, as a fallback when kube-proxy is not detected: the host must have a usable nft binary (knftables enforces >= v1.0.1) and a kernel of at least 5.13. These floors are Calico's documented nftables requirement ("Linux kernel version 5.13 or later with nft >= 1.0.1"), which mirrors kube-proxy's own nftables mode requirement. That keeps e.g. RHEL 9+/Debian 12+/Ubuntu 22.04+ on nftables while RHEL 8 (4.18 kernel, nft 0.9.x) stays on iptables.

Nothing supplies the fallback yet, so behavior is unchanged: in a cluster, Auto must keep following kube-proxy - that signal is about running the same dataplane as kube-proxy, not about host capability. The caller for hosts that never run kube-proxy arrives separately.

Also document the existing Auto-mode resolution: on the NFTablesMode API field description (CRDs/docs regenerated) and in a paragraph in felix/DESIGN.md's "Dataplane backends" section.

**Release note:**
```release-note
TBD
```

### DIFF
--- a/api/config/crd/projectcalico.org_felixconfigurations.yaml
+++ b/api/config/crd/projectcalico.org_felixconfigurations.yaml
@@ -1120,9 +1120,10 @@ spec:
                   type: integer
                 nftablesMode:
                   default: Auto
-                  description:
-                    "NFTablesMode configures nftables support in Felix. [Default:
-                    Auto]"
+                  description: |-
+                    NFTablesMode configures nftables support in Felix. In Auto mode, Felix uses the
+                    nftables dataplane if kube-proxy is detected to be running in nftables mode.
+                    [Default: Auto]
                   enum:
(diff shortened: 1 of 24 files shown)


mazdakn (approved):
LGTM

PR #13185, opened 2026-07-09
fix(felix): expire flows when the rule trace verdict changes match index

radTuti:
When a policy change flips a flow's verdict mid-flow, the new verdict rule can land in an empty slot of the recorded rule trace, silently changing the flow's action without expiring the already-reported flow. The flow log aggregator then holds a live reference forever and re-exports the flow with zeroed counters on every flush, so Goldmane/Whisker keep showing a deny flow with zero packets in every query window until Felix restarts.

Treat a verdict arriving at a different match index as a rule trace change, so the existing flow is reported and expired before the trace is replaced. Regression tests drive the collector through deny→allow transitions and assert the deny aggregator drains.

**Release note:**

```release-note
Fix that Felix kept re-emitting stale deny flow logs with zero counters (shown in Whisker as a live deny) after a policy change allowed the traffic.
```


felix/collector/rule_trace_change_test.go
```diff
@@ -0,0 +1,262 @@
...
+	if data := c.epStats[tup]; data == nil || !data.Reported {
+		t.Fatalf("setup: deny flow not reported, data=%v", data)
+	}
+}
+
+func ageOut(t *testing.T, c *collector, tup tuple.Tuple) {
+	t.Helper()
+	time.Sleep(250 * time.Millisecond) // > AgeTimeout
+	c.checkEpStats()
+	if _, ok := c.epStats[tup]; ok {
```
caseydavenport, line 179: We should rewind the flow's timestamps rather than sleep. A 250ms sleep against a 200ms `AgeTimeout` will flake on a loaded CI agent, and these tests are in-package so they can age a flow directly:

```go
func rewind(d *Data, by time.Duration) {
	d.updatedAt -= by
	d.ruleUpdatedAt -= by
}
```

`denyThenReport` then calls `rewind(c.epStats[tup], time.Second)` before `checkEpStats()`, and `ageOut` calls `rewind(c.epStats[tup], 2*conf.AgeTimeout)`. The 5ms sleeps in the tests themselves go away with it.

felix/collector/rule_trace_change_test.go
```diff
@@ -0,0 +1,262 @@
...
+// re-exported forever.
+func assertDenyAggDrains(t *testing.T, cap *muCapture) {
+	t.Helper()
+	var last []*flowlog.FlowLog
+	for range 4 {
+		last = cap.denyAgg.GetAndCalibrate()
+	}
+	for _, fl := range last {
+		t.Errorf("leaked flow in deny aggregator: action=%v reporter=%v pktsIn=%d numFlows=%d policies=%v",
+			fl.Action, fl.Reporter, fl.PacketsIn, fl.NumFlows, fl.FlowEnforcedPolicySet)
```
caseydavenport: Only the fourth flush gets checked here, so a flow that is wrongly re-exported on flushes 1-3 still passes. The first flush legitimately carries the final deny record and every flush after it should be empty:

```go
cap.denyAgg.GetAndCalibrate() // final report for the expired deny flow
for i := range 3 {
	if fls := cap.denyAgg.GetAndCalibrate(); len(fls) > 0 {
		t.Errorf("leaked flow in deny aggregator on flush %d: %+v", i+2, fls)
	}
}
```

caseydavenport (changes requested):

caseydavenport (approved):

PR #13193, opened 2026-07-10
Harden run-and-monitor against Semaphore agent job hangs

fasaxc:
A CI job hung after `run-and-monitor` printed its final `SUCCEEDED ... exit with RC=0` message, until the job timed out. Root cause: the log tail the wrapper copies to the console contained binary test output including a raw 0x01 byte. The Semaphore agent detects command completion by scanning console output for a 0x01-prefixed end marker, and its scanner can permanently discard the real marker when stray 0x01 bytes arrive just before it (`flushInputAll()` in buffering mode flushes through a partially-received marker). The wrapped command had also left the job terminal with output post-processing disabled, which breaks the agent's CRLF-sensitive start-marker detection for the next command.

This hardens the wrapper on three fronts:

- Strip control characters (keeping tab/LF/CR/ESC, so ANSI colour still works) from everything copied from the log file to the console.
- Save terminal state at startup and restore it in `cleanup()` before the final output.
- Kill the monitor pipeline's console-facing end and wait for it, so nothing can write to the console after the final message.

Also fixes the line count in `output-tail`: `wc`'s leading padding made `cut -d' ' -f1` return an empty string, so the tail of a long log was never printed.

Testing: added `.semaphore/test-run-and-monitor.sh`, a manual containerised smoke test covering the success, failure, control-character and terminal-state paths (all pass on ubuntu:24.04; see its header for the invocation). Verified on the live hung runner that restoring the terminal state and re-emitting the marker unstuck the agent.

### DIFF
--- a/.semaphore/run-and-monitor
+++ b/.semaphore/run-and-monitor
@@ -81,7 +81,7 @@ grep_args+=( "-e" "${pattern:1}" )
 # Semaphore agent's command-completion marker detection, hanging the job.
 function sanitize()
 {
-  stdbuf -oL tr -d '\000-\010\013\014\016-\032\034-\037\177'
+  LC_ALL=C stdbuf -oL tr -d '\000-\010\013\014\016-\032\034-\037\177'
 }
 
 tail --pid $pid -n 1000000 -F "$log_file" | {
--- a/.semaphore/run-and-monitor
+++ b/.semaphore/run-and-monitor
@@ -24,6 +24,14 @@ log_monitor_regexps=(
   "Test batch"
 )
 
+# Save the terminal state so we can restore it before our final output.  The
+# wrapped command can leave the terminal in a bad state (for example with
+# output post-processing disabled); the Semaphore agent detects command
+# completion by scanning the terminal output for a marker, and a corrupted
+# terminal state can prevent that marker from being recognised, hanging the
+# job until it times out.
+saved_stty="$(stty -g < /dev/tty 2>/dev/null || true)"
+
 echo "Current dir: $(pwd)"
 my_dir="$(dirname $0)"
 repo_dir="${my_dir}/.."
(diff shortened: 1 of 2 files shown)


.semaphore/run-and-monitor
```diff
@@ -67,26 +75,48 @@ for r in "${log_monitor_regexps[@]}"; do
 done
 grep_args+=( "-e" "${pattern:1}" )
 
+# Strip control characters (keeping tab, newline, CR and the ANSI colour
+# escape) from everything we copy from the log to the console.  Test output
+# can contain binary junk, and a raw 0x01 byte in particular confuses the
+# Semaphore agent's command-completion marker detection, hanging the job.
+function sanitize()
+{
+  LC_ALL=C stdbuf -oL tr -d '\000-\010\013\014\016-\032\034-\037\177'
```
nelljerram, line 84: Is suppressing DEL and Backspace going to cause garbled output if we still have any jobs that try to do clever output, e.g. because they don't have `BUILDKIT_PROGRESS=plain` ?

nelljerram (approved):
One query, and I'm interested in the answer, but don't think I will need to re-review.

.semaphore/run-and-monitor
fasaxc: _Claude:_ Short answer: no — the two mechanisms "clever output" actually uses are both preserved.

The filter is `tr -d '\000-\010\013\014\016-\032\034-\037\177'`, which keeps TAB, LF, **CR (0x0D)** and **ESC (0x1B)**, and drops BS (0x08) and DEL (0x7F) among the rest. The live-update tricks rely on the ones we keep:

- **Carriage return** — the classic redraw-this-line-in-place. Kept.
- **ANSI escape sequences (ESC `[…`)** — cursor-up / clear-line / colour. This is exactly what BuildKit's TTY/`auto` progress renderer uses when `BUILDKIT_PROGRESS=plain` is *not* set. ESC is kept, so those pass straight through. (The command runs under `expect` in a PTY, so `isatty` is true and BuildKit picks the ANSI renderer; those bytes land in the log and survive `sanitize`.)

The only thing dropping BS/DEL affects is a tool that animates by emitting a literal backspace to erase characters (old-school `printf '\b\b\b%3d%%'` counters / spinners). Those are rare, and the failure mode is cosmetic — characters concatenate instead of overwriting (`10%20%30%`), not a hang. Two mitigations on top:

- It only affects the **console monitor view**; the full, unsanitized output is always in the artifact log, so nothing is lost.
- The console stream is already line-oriented (`stdbuf -oL`, `grep --line-buffered`), so intra-line backspace animation wouldn't render faithfully on Semaphore's non-interactive log regardless.

So CR + ESC preserved ⇒ progress bars and BuildKit `auto` progress still work; only literal-backspace animations degrade, cosmetically, and only in the filtered console view.

fasaxc: I'm inclined to fix only if we hit such an instance!

PR #13196, opened 2026-07-10
argoci: query GOOGLE_PROJECT in the zone lookup

lwr20:
## Problem
The prologue's zone lookup passed no `--project`:
```
gcloud compute zones list --filter="region~'$GOOGLE_REGION'" ...
```
On the ArgoCI runner, gcloud then uses the **runner's default project** (`tigera-cc-dev`, where the service account has no `compute.zones.list`) instead of `GOOGLE_PROJECT` (`unique-caldron-775`). The lookup fails:
```
ERROR: (gcloud.compute.zones.list) Some requests did not succeed:
 - Required 'compute.zones.list' permission for 'projects/tigera-cc-dev'
```
so `GOOGLE_ZONE` never resolves and **every gcp-kubeadm job that doesn't pin its zone fails to provision.** This was the dominant blocker across the iptables/bpf pipeline runs.

## Fix
Pass `--project "${GOOGLE_PROJECT}"` to the zones list, so it queries the e2e project the SA can actually read.

## Test
`bash -n` clean; validated on an ArgoCI run.

## Release Note
```release-note
NONE
```


lwr20:
Superseded by #13195 — folded the `--project` fix in there to keep the prologue migration fixes in one PR.

PR #13197, opened 2026-07-10
Fix EnsureBlock IPv6 branch using IPv4 pool selector

fasaxc:
## Summary

`ipamClient.EnsureBlock`'s IPv6 branch validates `args.IPv6Pools` but then passes `args.IPv4Pools` into the underlying `ensureBlock` call for the v6 allocation. As a result, a caller that supplies an explicit `IPv6Pools` selector alongside `HostReservedAttrIPv6s` gets the block allocated against the wrong pool list — either landing in an unintended IPv6 pool, or failing outright if the IPv4 CIDR isn't a valid IPv6 pool.

The only caller passing explicit `HostReservedAttrIPv6s`/pools is the Windows host-local block-ensure path in `node/pkg/lifecycle/startup/startup_windows.go`, so this has been latent rather than actively triggering failures.

Fix: pass `args.IPv6Pools` instead of `args.IPv4Pools` in the v6 branch.

## Test plan

- [x] Added a regression test in `libcalico-go/lib/ipam/ipam_test.go` (`should allocate an IPv6 block from IPv6Pools, not IPv4Pools`) that calls `EnsureBlock` with distinct `IPv4Pools`/`IPv6Pools` and asserts the returned v6 block CIDR falls inside the requested `IPv6Pools`.
- [x] Verified the test fails against the pre-fix code (etcd backend errors with `the given pool (10.0.0.0/24) does not exist, or is not enabled`, since the v4 CIDR is looked up against the v6 pool table).
- [x] Verified the test passes with the fix, and the full `libcalico-go/lib/ipam` suite still passes against the etcdv3 backend.

**Release note:**
```release-note
Fixed a latent bug where EnsureBlock could allocate an IPv6 block from the wrong IP pool when an explicit IPv6 pool selector was provided (Windows host-local IPAM block reservation).
```

### DIFF
--- a/libcalico-go/lib/ipam/ipam.go
+++ b/libcalico-go/lib/ipam/ipam.go
@@ -2465,7 +2465,7 @@ func (c ipamClient) EnsureBlock(ctx context.Context, args BlockArgs) (*net.IPNet
 				return nil, nil, fmt.Errorf("provided IPv6 IPPools list contains one or more IPv4 IPPools")
 			}
 		}
-		v6Net, err = c.ensureBlock(ctx, args.HostReservedAttrIPv6s, args.IPv4Pools, 6, affinityCfg)
+		v6Net, err = c.ensureBlock(ctx, args.HostReservedAttrIPv6s, args.IPv6Pools, 6, affinityCfg)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


lucastigera (approved):
LGTM

PR #13199, opened 2026-07-10
felix: answer IPv6 DAD probes in proxy-NDP

MichalFupso:
Felix's IPv6 proxy-NDP listener silently ignored Duplicate Address Detection probes, so it never defended a proxied pod/LB IP against another host claiming the same address. Bind the NDP socket to the interface link-local and strip the interface zone before the unspecified-source check, so DAD probes now reach the handler and are answered.

**Release note:**
```release-note
TBD
```


lucastigera (approved):
LGTM.

PR #13200, opened 2026-07-10
ci: more utility functions and edge case handling

danudey:


marvin-tigera:

release/internal/utils/gpg.go
```diff
@@ -45,7 +45,7 @@ func GetGPGPubKey(gpgKeyID string) (string, error) {
 func SignRPMFiles(gpgKeyID string, rpmFiles []string) error {
 	if len(rpmFiles) == 0 {
 		logrus.Warn("called with zero filenames, exiting")
-		return nil
+		return fmt.Errorf("provided RPM files list is empty")
```
radTuti: why the switch to return an error?

release/internal/utils/gpg.go
```diff
@@ -45,7 +45,7 @@ func GetGPGPubKey(gpgKeyID string) (string, error) {
 func SignRPMFiles(gpgKeyID string, rpmFiles []string) error {
 	if len(rpmFiles) == 0 {
 		logrus.Warn("called with zero filenames, exiting")
```
radTuti: not needed if an error gets returned
```suggestion
```

release/internal/utils/gpg.go
danudey: If a function is building a list of RPM files to sign, it should be validating that list is reasonable. It may be correct behavior to find zero RPMs to sign, but if so it shouldn't bother calling `SignRPMFiles()` just to be told that its own list is empty.

Put another way, it could be that passing zero files is reasonable or it could be that it's a mistake, but `SignRPMFiles()` has no way to know which it is and, if it returns `nil`, no way to indicate that nothing was done.

This does raise the issue that `CheckRPMSigs()` should have the same behavior, which I will implement.

release/internal/utils/gpg.go
danudey: The reason I tend to use this pattern is that in many cases, when debugging release tool, we inadvertently see a program exit with "main.go:99 main.main() Error running app" and then an error message which isn't explicit about what happened where; for example, getting an error like "branch name is not a valid release branch" doesn't help narrow down the code path where the actual error happened.

Putting in error or warning messages whenever we return an error means that we end up with a cascade of error messages that serve almost as a stack trace.

If that's not a useful pattern then I'm happy to remove it in favor of a better solution.

radTuti (approved):

PR #13201, opened 2026-07-10
calico-node: fix install-cni to call `component cni install` (monobinary regression)

lwr20:
## Description

The consolidation of Calico components into the single `calico` binary (#12225) registered the CNI plugin under `calico component cni` and updated the `calico-node` / `calico-kube-controllers` manifests to the `calico component <name>` form — but the **`install-cni` init container was left calling `calico cni install`**.

The combined binary has no top-level `cni` subcommand (it lives under `component`), so the init container exits immediately with:

```
Error: unknown command "cni" for "calico"
Did you mean this?  ctl
```

`calico-node` therefore never gets past its init containers — every pod sits in `Init:CrashLoopBackOff` — so **every manifest-based (non-operator) install of `master` fails to come up**. Operator installs are unaffected (the operator invokes the binary itself).

Found via e2e triage: all `calico.yaml`/`canal.yaml`-based install jobs (datastore, dual-stack, distro-support, wireguard, kind) fail at `manual:calico:verify-components` waiting for `calico-node` Ready; diags show the init container crash above.

## Fix

`charts/calico/templates/calico-node.yaml` — `install-cni` args `["cni", "install"]` → `["component", "cni", "install"]`, matching the sibling `node` (`command: ["calico","component","node","init",...]`) and `kube-controllers` (`args: ["component","kube-controllers"]`) invocations. Regenerated the 10 affected static manifests (`make gen-manifests`).

## Release Note

```release-note
Fixed calico-node install-cni init container crash-looping on manifest-based (non-operator) installs (regression from the single-binary consolidation), which prevented calico-node from starting.
```

### DIFF
--- a/charts/calico/templates/calico-node.yaml
+++ b/charts/calico/templates/calico-node.yaml
@@ -95,7 +95,7 @@ spec:
         - name: install-cni
           image: {{.Values.calico.registry}}/{{.Values.calico.image}}:{{ .Values.version }}
           imagePullPolicy: {{.Values.imagePullPolicy}}
-          args: ["cni", "install"]
+          args: ["component", "cni", "install"]
           envFrom:
             - configMapRef:
                 # Allow KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT to be overridden for eBPF mode.
--- a/manifests/calico-bpf.yaml
+++ b/manifests/calico-bpf.yaml
@@ -12458,7 +12458,7 @@ spec:
         - name: install-cni
           image: quay.io/calico/calico:master
           imagePullPolicy: IfNotPresent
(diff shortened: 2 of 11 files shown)


coutinhop (approved):

lwr20:
Adding `skip-bot-cherry-pick`: this fixes a **monobinary-only** regression (install-cni calling `calico component cni install`). The single-binary consolidation (#12225) is **master-only** — release-v3.32/v3.31/v3.30 are all pre-monobinary (separate `calico/cni` image, `command: ["/opt/cni/bin/install"]`), so the bug doesn't exist there and this doesn't apply. Future release branches are cut from master (which already has this), so no backport is ever needed.

PR #13205, opened 2026-07-13
kube-controllers: support annotations on auto host endpoint templates

### LINKED ISSUE #13192: Auto Host Endpoint QoS annotation
Is there a way to specify QoS for autogenerated host endpoints?

## Expected Behavior
It should be possible to specify qos.projectcalico.org/dscp annotation in host endpoint template

## Current Behavior
Auto host endpoint controller doesn't manage annotations

## Context
Our nodes have two network interfaces: one for application traffic and another for storage traffic. We would like to specify different QoS values for traffic leaving different interfaces.

Easily done with manually created HostEndpoints and we can patch the auto endpoints after creation, but these approaches are rather inconvenient.

MichalFupso:
Add an `annotations` field to the KubeControllersConfiguration auto-HEP Template, mirroring `labels`: annotations set on a template are applied to the host endpoints it generates and kept in sync.

**Release note:**
```release-note
 kube-controllers can now set annotations on automatically-created host endpoints via the annotations field of the KubeControllersConfiguration auto host endpoint template. Generated host endpoints are fully managed, so any annotation added directly to a generated host endpoint (outside the template) is removed on the next reconcile — set custom annotations on the template's annotations field instead.
```

### DIFF
--- a/api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
+++ b/api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
@@ -185,6 +185,9 @@ type Template struct {
 	// Labels adds the specified labels to the generated AutoHostEndpoint, labels from node with the same name will be overwritten by values from the template label
 	Labels map[string]string `json:"labels,omitempty" validate:"omitempty,labels"`
 
+	// Annotations adds the specified annotations to the generated AutoHostEndpoint.
+	Annotations map[string]string `json:"annotations,omitempty"`
+
 	// NodeSelector allows the AutoHostEndpoint to be created only for specific nodes
 	NodeSelector string `json:"nodeSelector,omitempty" validate:"omitempty,selector"`
 }
--- a/api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
+++ b/api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
@@ -155,7 +155,7 @@ type AutoHostEndpointConfig struct {
 
 	// Templates contains definition for creating AutoHostEndpoints
 	// +listType=atomic
-	Templates []Template `json:"templates,omitempty" validate:"omitempty"`
+	Templates []Template `json:"templates,omitempty" validate:"omitempty,dive"`
 }
 
 // DefaultHostEndpointMode controls whether a default host endpoint is created for each node.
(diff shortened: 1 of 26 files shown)


kube-controllers/pkg/controllers/node/auto_hep_fv_test.go
```diff
@@ -194,7 +194,7 @@ var _ = Describe("Auto Hostendpoint FV tests", Ordered, ContinueOnFailure, func(
 		}
 		expectedIPs := []string{"172.16.1.1", "fe80::1", "192.168.100.1"}
 		Eventually(func() error {
-			return testutils.ExpectHostendpoint(c, expectedHepName, expectedHepLabels, expectedIPs, autoHepProfiles, defaultInterfaceName)
+			return testutils.ExpectHostendpoint(c, expectedHepName, expectedHepLabels, nil, expectedIPs, autoHepProfiles, defaultInterfaceName)
```
mazdakn, line 197: It would be better to pass in a new `expectedHepAnnotions` similar to label and pass it to all or some of the existing tests.

api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
```diff
@@ -155,7 +155,7 @@ type AutoHostEndpointConfig struct {
 
 	// Templates contains definition for creating AutoHostEndpoints
 	// +listType=atomic
-	Templates []Template `json:"templates,omitempty" validate:"omitempty"`
+	Templates []Template `json:"templates,omitempty" validate:"omitempty,dive"`
```
mazdakn, line 158: Does adding this may cause validation to reject some of the already existing resource in datastore? 
it's valid to add this. My concern is if we might break some clusters with this new validation?

mazdakn (commented):
Overall LGTM. Just one question.
Also please update the copy-right years if needed.

kube-controllers/pkg/controllers/node/auto_hep_fv_test.go
MichalFupso: I don't think that would add much value to the tests. The reason why other tests have expectedLabels is because they inherit node labels. I think the new test that adds/removes the annotation spec from label is sufficient.

api/pkg/apis/projectcalico/v3/kubecontrollersconfig.go
MichalFupso: We validate other fields in the validator, this should be safe for current clusters as invalid values would have been rejected already

mazdakn (approved):

